PR #9503, opened 2021-12-30
[ROM] Calculate ROM basis output process

rubenzorrilla:
**📝 Description**
This PR adds a process to set up the `RomParameters.json` file, including the calculation of the ROM basis. I want to stress that, as this is routine outputting something, I derived it from the `OutputProcess` class. This makes possible to straightforwardly customize the snapshot frequency, something that might be interesting to limit the computational effort in large (potentially explicit) simulations. Unittesting has been added accordingly.

PR #9504, opened 2021-12-31
[ROM] Complement offline stage of ROM

Rbravo555:
#depends on #9503

**📝 Description**
This PR adds the possibility to modify the parameters of a ROM simulation by editing the RomParameters.json.

In Particular, to indicate that a ROM simulation should be used for training the HROM elements and weights, it was necessary to pass a string to RomAnalysis. Now, one needs to set  "train_rom": True in the RomParameters.json.

**🆕 Changelog**
- Getting rid of the ElementsAndWeights.json file. Instead, its content is appended to the RomParameters.json
- Getting rid of the CreateHRomAnalysisInstance of the rom_analysis
- Modifying parameters of the rom_output_process to include train_hrom and run_hrom


rubenzorrilla:
As the previous implementation was way intrusive, I created the `AuxiliaryRomUtilities`. In there I've defined two static methods, one for the creation of the HROM computing model part and the visualization one. Regarding the former, I've basically refactored the previous Python code to c++. Concerning the latter, I think that the solution we had so far is not very convenient, as it relies on the existent conditions to create the visualization mesh. Chances are that there are no conditions, so no visualization mesh is created. Instead, I suggest to detect the skin and create those entities manually (in this PR I'm only adding the interface, I'd add this capability in a different PR).

Another open issue is to add the capability to optionally add the condition together with its parent element in the HROM computing model part creation as this might be needed in some formulations. 

(I didn't test the implementation yet)

rubenzorrilla:
@Rbravo555 @SADPR I think we should start thinking about merging this (otherwise the PR will get bigger and bigger). On top of the HROM workflow enhancements, now we have the capability to add HROM condition parents and minimum conditions. Besides this, the skin of a volumetric body can be generated as a visualization mesh. To this regard, in a future PR we should create a process to do the solution projection onto this mesh (as well as the BC imposition) and the output. Note that this comes after the fact that the **visualization model parts are independent to the HROM one**.

rubenzorrilla:
#9608 must be merged first.

applications/RomApplication/custom_utilities/rom_auxiliary_utilities.cpp
```diff
@@ -0,0 +1,440 @@
...
+    NodesPointerSetType hrom_nodes_set;
+    std::vector<Element::Pointer> hrom_elems_vect;
+    std::vector<Condition::Pointer> hrom_conds_vect;
+
+    const auto& r_elem_weights = HRomWeights["Elements"];
+    hrom_elems_vect.reserve(rOriginModelPart.NumberOfElements());
+    for (auto it = r_elem_weights.begin(); it != r_elem_weights.end(); ++it) {
+        // Get element from origin mesh
+        const IndexType elem_id = stoi(it.name());
+        const auto p_elem = rOriginModelPart.pGetElement(elem_id + 1); //FIXME: WHY THIS +1?
```
SADPR, line 62: Is it going to remain +1?  For me, it is ok for now...

rubenzorrilla: I'd change the Ids 0 issue in a standalone PR to easily backtrack the changes if something gets messed up.

applications/RomApplication/python_scripts/rom_analysis.py
```diff
@@ -46,78 +86,67 @@ def ModifyAfterSolverInitialize(self):
...
-                    res_mat = self.ResidualUtilityObject.GetResiduals()
-                    np_res_mat = np.array(res_mat, copy=False)
-                    self.time_step_residual_matrix_container.append(np_res_mat)
-
+            # Call the HROM training utility to append the current step residuals
+            # Note that this needs to be done prior to the other processes to avoid unfixing the BCs
+            if self.train_hrom:
+                self.__hrom_training_utility.AppendCurrentStepResiduals()
+
+            # #FIXME: Make this optional. This must be a process
```
SADPR, line 117: I'm confused with this, isn't it always projected in ProjectToFineBasis?

rubenzorrilla: The point is that the visualization model parts must be independent to the HROM one. I'll take care of this in an upcoming PR. So far I only included some tools required for the creation of such model parts.

SADPR (commented):
It looks good to me. @Rbravo555 check it so we can merge it.

applications/RomApplication/python_scripts/rom_analysis.py
```diff
@@ -4,6 +4,7 @@
 import KratosMultiphysics.RomApplication as KratosROM
 from KratosMultiphysics.RomApplication import new_python_solvers_wrapper_rom
 from KratosMultiphysics.RomApplication.hrom_training_utility import HRomTrainingUtility
+from KratosMultiphysics.RomApplication.calculate_rom_basis_output_process import CalculateRomBasisOutputProcess
```
rubenzorrilla, line 7: Why is this required? The process must be imported by the factory when instantiating the content of the `output_process_list`?

Rbravo555: I noticed the ROM simulation runs well right after the training. But when launching it separately (for changing the number of modes to use for example) Kratos complained that the “calculate_rom_output_process” didn’t belong to the RomApplication. Importing it manually here solved the problem for me..

rubenzorrilla: OK. My bet is that if the process is not added in the process list it is not imported, then it cannot do the check below.

rubenzorrilla (approved):

PR #9501, opened 2021-12-30
[MathUtils] Add tolerance as parameter to GeneralizedInvertMatrix

miguelmaso:
**📝 Description**
Possible solution suggested in #9500 
The default behavior is the same.

**🆕 Changelog**
- Add `Tolerance` as an argument of `GeneralizedInvertMatrix`

Issue #9500, opened 2021-12-29
MathUtils::InvertMatrix throws an error if the matrix is not invertible


miguelmaso:
**Description**
Currently, `MathUtils::InvertMatrix` throws an error if the matrix is not invertible. Since I'm using this function in an iterative procedure, the `KRATOS_ERROR` is stopping the iterations. More or less I'm doing the following:
```c++
Matrix A(6,6); // this is 2D, size is 10 in a 3D case
Matrix invA(6,6);
bool is_invertible = false;
while (!is_invertible) {
    // do something and compute A
    double detA;
    is_invertible = MathUtils<double>::InvertMatrix(A, invA, detA);
    if (!is_invertible){
        // modify the topology
    }
}
```
Fortunately, the system is very simple and usually it is possible to get an invertible A.
For robustness I'm asking to change the error into a warning. Further, It would be useful to return a value indicating if the inverse is successful.

ddiezrod:
I think you can already do now what you want. If you pass a negative value as the fourth argument to InvertMatrix, it will not check the condition number, so no error is thrown. If you want to have a warning. you can call later `MathUtils::CheckConditionNumber` using ThrowError=false which will return false if the condition number is too high (matrix not invertible) and true otherwise.

loumalouomega:
You can change the ZeroTolerance to a large number:

https://github.com/KratosMultiphysics/Kratos/blob/260a6b4193f18a4f201d2f20478cb3facb91f12c/kratos/utilities/math_utils.h#L400

If you don't want to think about a big number, just use the STL maximum: https://en.cppreference.com/w/cpp/types/numeric_limits/max

miguelmaso:
> I think you can already do now what you want. If you pass a negative value as the fourth argument to InvertMatrix, it will not check the condition number, so no error is thrown. If you want to have a warning. you can call later `MathUtils::CheckConditionNumber` using ThrowError=false which will return false if the condition number is too high (matrix not invertible) and true otherwise.

Thanks @ddiezrod !

PR #9505, opened 2022-01-03
[Shallow] Derivatives recovery

miguelmaso:
**📝 Description**
Apply the solution suggested in #9500.

**🆕 Changelog**
- An iterative procedure to calculate the polynomial weights is possible since a custom tolerance is passed to `MathUtils::GeneralizedInvertMatrix` (#9501)
- Update the signature of the custom functions.


ipouplana (approved):
Good!

PR #9506, opened 2022-01-03
[Core][Wheel] Add missing module to Kratos Wheel

armingeiser:
**📝 Description**
Some python files where missing in the Kratos wheel

**🆕 Changelog**
- Fix missing python files in wheel

### DIFF
--- a/kratos/KratosMultiphysics.json
+++ b/kratos/KratosMultiphysics.json
@@ -1,6 +1,6 @@
 {
 	"wheel_name": "KratosMultiphysics",
-	"included_modules": [],
+	"included_modules": ["response_functions"],
 	"included_binaries": ["Kratos.*", "KratosCore.*", "KratosCore.*", "zlib.*", "libz.*"],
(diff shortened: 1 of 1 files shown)


kratos/KratosMultiphysics.json
```diff
@@ -1,6 +1,6 @@
 {
 	"wheel_name": "KratosMultiphysics",
-	"included_modules": [],
+	"included_modules": ["response_functions"],
```
armingeiser, line 3: There are also two other folders at this level: "testing" and "application_generator" but from my understanding they need not to be part of the wheel.

armingeiser:
The wheels are not tested before uploading to PyPi, all ShapeOptimization tests failed with the version currently online.

roigcarlo (commented):
:+1: I was not aware that subfolders weren't imported. That should not happen. 

anyway if this works ok until I fix it for all Kratos. Thx!

roigcarlo (approved):

PR #5825, opened 2019-10-22
Link on install

roigcarlo:
Cleanup from #5145 and #5170 

This is a draft, I am still testing everything, I just want some CI to trigger as well

List of changes:
- [x] New install mechanism is implemented for all applications
- [x] CMake files have been harmonized between all applications
- [x] Tests are now mandatory. Applications without tests now have the automatically generated ones with dummy tests.
- [x] **TrilinosApplication**: Some scripts are installed out of the new mechanism. Still testing those.
- [x] **CoSimulationApplication**: I did not understand the CMake.
- [x] Improve readme and put some doc into the configuration files.

Commits are done per application, so if any application needs to remain with the old mechanism ( you won't be able to install with the rest ) it still can be reverted.

PR #3217, opened 2018-10-28
making python-modules work

philbucher:
With this modification the python-scripts can be subdivided into subfolders 
=> now they behave as real python modules!
=> the changes necessary are very small and can be done app-by-app without breaking anything
i.e. these changes do NOT break the current mechanism, it is fully backwards compatible and can be adapted slowly

e.g. now one can do:
~~~py
import KratosMultiphysics
import KratosMultiphysics.StructuralMechanicsApplication as KSM

from KratosMultiphysics import new_linear_solver_factory

from KratosMultiphysics.StructuralMechanicsApplication import structural_mechanics_analysis

# for demonstration:
from KratosMultiphysics.StructuralMechanicsApplication.solvers import custom_solver
~~~

solves #2655

I will discuss this with @jcotela in the next days, afterwards you can discuss it in the comittee

I think this also works in python2, but I have not tried it

FYI @adityaghantasala @oberbichler

PR #9507, opened 2022-01-03
[ShapeOpt] add missing init files

armingeiser:
**📝 Description**
The subfolders of the python scripts folder in the ShapeOptApp had no `__init__.py` files. 

I don't know why it works without but its good practice to have those files in each folder. 

**🆕 Changelog**
- Added `__init__.py` to all submodules in ShapeOptimizationApplication


miguelmaso:
> I don't know why it works without but its good practice to have those files in each folder.

it was introduced in #3217

philbucher:
> > I don't know why it works without but its good practice to have those files in each folder.
> 
> it was introduced in #3217

actually the solution from #3217 was only an intermediate solution that was removed when we switched to the new installation mechanism (#5825). 
Basically #3217 enabled us to use the python-scripts properly (i.e. `from Kratos.AppName import py_script.py` instead of `import py_script.py`) This was a prerequisite for #5825

I am not sure why it works without the `__init__.py`. The CoSimApp has many subfolders but so far it was always working without

@roigcarlo do you know better?

sunethwarna:
Isn't it because of the following that we can use python imports eventhough we don't have `__init__.py` files?

https://docs.python.org/3/whatsnew/3.3.html#pep-420-implicit-namespace-packages

armingeiser:
Hm, maybe it works because of the namespace package thing as @sunethwarna says. 

But from what i understand we are not actually using namespace packages, but all our apps should rather be "standard" packages, thus the `__init__.py``files should be there.

roigcarlo (approved):
Exactly as @sunethwarna says. As we had support for 2.x when we make the change to use standard python modules it wouldn't have worked without. In any case I still think is a good idea to have the `__init__.py` file, among other reasons because not all our folders are sub-modules and this is a clean way to differentiate them.

PR #8384, opened 2021-02-27
[Core] Replace Cotire with CMake Unity Build

loumalouomega:
**Description**
From the official Cotire repo: https://github.com/sakra/cotire

> The functionality provided by cotire has been superseded by features added to CMake 3.16. Support for pre-compiling and unity builds is now built into CMake. Thus, there will not be any further updates or support for this project.

**Requires CMake 3.16** (which is available in many distros since long time ago).

As you can see in the changes the usage is much simpler (this PR removes lines instead of adding). Compilation is done in the standard way:

~~~sh
cmake --build "${KRATOS_BUILD}/${KRATOS_BUILD_TYPE}" --target install -- -j$(nproc)
~~~

instead of:

~~~sh
cmake --build "${KRATOS_BUILD}/${KRATOS_BUILD_TYPE}" --target all_unity    -- -j$(nproc) && \
cmake --build "${KRATOS_BUILD}/${KRATOS_BUILD_TYPE}" --target install/fast -- -j$(nproc)
~~~

And you don't need to explicitly cotire anything, you only need to add to your compilation the following flag: CMAKE_UNITY_BUILD=ON. CMake does the rest.

Last but not least, in order to exclude from unity it is necessary to change the property `SKIP_UNITY_BUILD_INCLUSION` instead of `COTIRE_EXCLUDED`.

But first let's see if in compiles in the CI...

**Changelog**
- refactor main CMake files
- Clean existing CMake application files
- Add missing compilation guards

PR #9509, opened 2022-01-03
Run tests using wheel

### LINKED ISSUE #9508: Add Workflow to create and test wheels based on the current master
**Description**
Currently the wheels are not tested before uploading to PyPi (see #9506). 
I suggest to add a workflow that builds Kratos as wheel instead of the "standard" build, and runs the tests by installing the local Kratos wheels using `pip`. 

Doing this for **one** python version and **one** operating system would already catch most of the potentially occurring errors. Also this should not be too difficult to implement. 

The reason i suggest to build based on latest master instead of using the uploaded packages from PyPi is that otherwise an update of the test cases on the master branch would break the test if the wheels are simply downloaded from PyPi.

At the same time, this would also "test" the whole wheel creation process.

armingeiser:
**📝 Description**
Proposal to test the Kratos wheels.

- A new CI job is added to build and test the Kratos wheels (only on Ubuntu, only for one Python version)
- The python testing scripts are added to the wheel to allow running the tests
- The `run_tests.py` script is extended to allow running the tests (in the source directory) using the pip-installed Kratos 
- Some apps where not creating a wheel because of a missing README - this has been fixed

- [ ] some tests for apps are failing here, this should definitely be looked at (see test result)

**🆕 Changelog**
- Add github action to build and test wheel for a single python version and operating system

closes #9508


.github/workflows/ci.yml
```diff
@@ -459,7 +459,7 @@ jobs:
       shell: bash
       run: |
         mkdir ${GITHUB_WORKSPACE}/wheels
-        docker run --rm -v ${GITHUB_WORKSPACE}/wheels:/data_swap_guest kratosmultiphysics/kratos-wheelbuilder-linux master 2
+        docker run --rm -v ${GITHUB_WORKSPACE}/wheels:/data_swap_guest kratosmultiphysics/kratos-wheelbuilder-linux master 4
```
philbucher: the runners only have 2 CPUs
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

philbucher:
Also i would disable the cpp-tests, these take a lot of the compilation time

armingeiser:
```
"KratosEigenSolversApplication==${KRATOS_VERSION}",
"KratosConstitutiveLawsApplication==${KRATOS_VERSION}",
"KratosDemStructuresCouplingApplication==${KRATOS_VERSION}",
"KratosCSharpWrapperApplication==${KRATOS_VERSION}",
"KratosRANSApplication==${KRATOS_VERSION}",
"KratosCompressiblePotentialFlowApplication==${KRATOS_VERSION}",
"KratosIgaApplication==${KRATOS_VERSION}",
"KratosChimeraApplication==${KRATOS_VERSION}",
"KratosRomApplication==${KRATOS_VERSION}"
```
no wheels are created for these apps, thats why they are removed for now from the build/test

philbucher:
you could use clang instead of gcc, in my experience it is ~30% faster with compiling Kratos

armingeiser:
I don't get why compilation is so slow. I mean its the same apps and settings(?) as in the standard CI runs. 
Could it be because the (supposedly old) compiler version in the manylinux image (for generating portable wheels)? Not sure if we can/should switch to clang there.

armingeiser:
Ok, so it finally worked. Before continuing on this i would like to hear the opinion of others on the topic.

philbucher:
- overall having a build/test for the wheels makes perfect sense to me, esp since we want to push this more and more as the distribution mechanism. Hence kudos for this contribution
- I am not sure if the build is much slower in your case. In the nightly build we use `Release`. This is fully optimized, gcc needs almost 2hrs for compiling it. Clang needs only bit more than 1hr, thats why I was suggesting it.The CI builds don't use any optimization and hence compile faster (at the cost of slower runtime)
- given the relatively slow build probably adding it to the nightly build makes more sense at first, lets see how it goes after collecting some experiences. I guess we can also add a manual trigger same as for the CI and the nightly build so that whoever wants to try sth can do so without modifying the CI file :)
- Can you please clean up this PR and make separate ones for the fixes that you added (and are independent of this PRs purpose)? I am loosing a bit the overview with the CI changes

kratos/python_scripts/run_tests.py
```diff
@@ -252,6 +255,15 @@ def main():
                 Usage()
                 sys.exit()
 
+        elif o in ('-p', '--path'):
+            try:
+                path = str(a)
+                print(f"path: {path}")
```
armingeiser:
```suggestion
                print("path: {}".format(path))
```

kratos/python_scripts/run_tests.py
```diff
@@ -252,6 +255,15 @@ def main():
...
+            except:
+                print(f'Error: Cannot parse applications directory path containing the tests {a}.')
```
armingeiser:
```suggestion
                print("Error: Cannot parse applications directory path containing the tests {}.".format(a))
```

philbucher:
BTW looking at the compiler output it seems like the unity build is not working, it is compiling the sources one by one

armingeiser:
> BTW looking at the compiler output it seems like the unity build is not working, it is compiling the sources one by one

the wheelbuilder image has cmake 3.15.3 installed. UnityBuild is supported starting from 3.16

.github/workflows/nightly_wheels_build_and_test.yml
```diff
@@ -0,0 +1,47 @@
+name: Nightly Wheels Build and Test
```
philbucher: Please note that putting this into an separate file will create another separate build (we have max 5 in parallel)
=> Each file counts as one workflow (i.e. CI, nightly and this too)
In the future if we want to include it in the regular CI then it might be moved to the `ci.yml`

.github/workflows/nightly_wheels_build_and_test.yml
```diff
@@ -0,0 +1,47 @@
...
+
+    - name: Build
+      shell: bash
+      run: |
+        python3 --version
+        mkdir ${GITHUB_WORKSPACE}/wheels
+        ls -al ${GITHUB_WORKSPACE}/wheels
+        echo "${GITHUB_HEAD_REF}"
+        docker run --rm -v ${GITHUB_WORKSPACE}/wheels:/data_swap_guest kratosmultiphysics/kratos-wheelbuilder-linux ${GITHUB_HEAD_REF}
+        ls -al ${GITHUB_WORKSPACE}/wheels
```
philbucher: I guess this has some debugging things leftover

.github/workflows/nightly_wheels_build_and_test.yml
```diff
@@ -0,0 +1,47 @@
...
+# for cancelling redundant runs
+concurrency:
+  group: nightly-wheels-build-test-${{ github.head_ref }}
+  cancel-in-progress: true
+
+jobs:
+  linux-wheel:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
philbucher:
```suggestion
```

can be removed as you don't use a strategy build

.github/workflows/nightly_wheels_build_and_test.yml
armingeiser: Ah, than i misunderstood you. I can simply put this job into the nightly file. We can then still move it to the regular ci file if wanted in the future.

kratos/KratosMultiphysics.json
```diff
@@ -1,6 +1,6 @@
 {
 	"wheel_name": "KratosMultiphysics",
-	"included_modules": [],
+	"included_modules": ["testing", "response_functions"],
```
armingeiser, line 3: contains change from #9506

.github/workflows/nightly_wheels_build_and_test.yml
armingeiser: removed them apart from the last line, since this a convenient way to see which wheels have been successfully build

armingeiser:
Summary of current status:

- build takes 4h
- cpp tests have to be built, since many apps are calling the cpp tests explicitly in their small test suite and fail if they are not compiled. (without the build takes 3.5h)
- numpy is required by a few apps to pass the tests
- some apps are still failing: 

```
CoSimulationApplication
TypeError: CreateDataTransferOperator() takes 1 positional argument but 2 were given
TypeError: CreateCouplingOperation() takes 3 positional arguments but 4 were given
TypeError: __init__() takes 3 positional arguments but 4 were given

MultilevelMonteCarloApplication
ModuleNotFoundError: No module named 'applications'

ShapeOptimizationApplication
ModuleNotFoundError: No module named 'KratosMultiphysics.ShapeOptimizationApplication.response_functions.total_volume'

SwimmingDEMApplication
ModuleNotFoundError: No module named 'h5py'


CPP
  File "kratos/python_scripts/run_tests.py", line 334, in <module>
    main()
  File "kratos/python_scripts/run_tests.py", line 326, in main
    commander.RunCppTests(applications)
  File "kratos/python_scripts/run_tests.py", line 153, in RunCppTests
    import_module("KratosMultiphysics." + application)
...
  File "/home/runner/.local/lib/python3.8/site-packages/KratosMultiphysics/ParticleMechanicsApplication/__init__.py", line 3, in <module>
    from KratosParticleMechanicsApplication import *
ImportError: generic_type: type "LinearElasticIsotropic3DLaw" is already registered
```

The error for ShapeOpt is strange, since the failing test exists on master but not yet on this branch.

philbucher:
> cpp tests have to be built, since many apps are calling the cpp tests explicitly in their small test suite and fail if they are not compiled. (without the build takes 3.5h)

this should be fixed/removed in the apps as it is legacy
can you list the problematic apps?

armingeiser:
> > cpp tests have to be built, since many apps are calling the cpp tests explicitly in their small test suite and fail if they are not compiled. (without the build takes 3.5h)
> 
> this should be fixed/removed in the apps as it is legacy can you list the problematic apps?

```
ContactStructuralMechanicsApplication
ConvectionDiffusionApplication
CSharpWrapperApplication
HDF5Application  - not part of wheel build
MeshingApplication
MeshMovingApplication
ParticleMechanicsApplication
RANSApplication
ShallowWaterApplication  - not part of wheel build
StatisticsApplication
```

could this be responsible for the error
```
ImportError: generic_type: type "LinearElasticIsotropic3DLaw" is already registered
```
?

armingeiser:
> build takes 4h

Building the wheels (including cpp tests) in `Custom` instead of `Release` takes 1h50min.

philbucher:
> > build takes 4h
> 
> Building the wheels (including cpp tests) in `Custom` instead of `Release` takes 1h50min.

for some reason the unity build is not working :/
@roigcarlo and idea why it is not working?

philbucher:
ah ok that explains it :)
cotire was removed in #8384, so unfortunately we can only use the unity build with a newer CMake.

Not sure if we can update the CMake. Lets hear what @roigcarlo has to say

roigcarlo:
Yep i can rebuild the containers so they use cmake 3.16.

Assign it to me please i will do it net week

armingeiser:
I made a few suggestions in #9522

roigcarlo:
Well I've been taking a look at all, I am ok with all the changes, as I said I am now preparing the new containers, but reached this point, wouldn't be better to just move al the current workflows to be wheel based?

So:
custom_gcc -> wheels_custom_gcc
nightly_gcc -> wheels_nightly_gcc
custom_clang -> wheels_custom_gcc
nightly_clang -> wheels_nightly_gcc
win -> wheels_win

These don't need to change:
fulldebug_gcc 
fulldebug_clang
centos
linux_core_without_unity (why is it called ubuntu-core?)
linux_intel
windows_core_without_unity

armingeiser:
> wouldn't be better to just move al the current workflows to be wheel based

I think that makes sense, as long as the standard workflow of the developers (build and run in-place instead through wheels) is also tested. 

A few things do not work with wheels yet however (afaik):
- pardiso solvers
- hdf5 app
- 
so they would not be tested.

armingeiser:
One advantage of having a wheel-based workflow is that we could test for multiple linux distributions with just one wheel build.

roigcarlo:
> I think that makes sense, as long as the standard workflow of the developers (build and run in-place instead through wheels) is also tested.

Well this is what I was trying to avoid. It's essentially the same process. What would be different to test is building without unity ( the server does that for wheels ) but is essentially the same problem we have now ( and the reason we have the core-without-unity builds)

As for the dots, the generation does work, it's just it cannot be distributed correctly but if you use it on the same instance it was generated should give no problem (HDF5 fix is on the way)

philbucher:
for me the unification makes sense as long as we don't loose coverage of some configurations

> linux_core_without_unity (why is it called ubuntu-core?)

because it uses ubuntu (and not e.g. CentOS)

philbucher:
another difference btw wheels and nightly build is also the building of the Cpp tests. As we are currently compiling them into the binaries I think it would be a good idea to avoid having wheels with the C++ tests

armingeiser:
Seems i found the issue of some tests failing because of code from master branch the tests:

https://github.community/t/github-sha-not-the-same-as-the-triggering-commit/18286

Apparently the Github action does not run on the actual commit of the branch in the PR, but on a version that is locally merged with master already... 
This means the wheel that is actually built for the current branch (because of a separate checkout inside the wheelbuilder image) is build based on different code compared to the code in the workspace of the Github action...

:confused:

armingeiser:
> This means the wheel that is actually built for the current branch (because of a separate checkout inside the wheelbuilder image) is build based on different code compared to the code in the workspace of the Github action...

@roigcarlo Is it possible to modify the wheelbuilder image in a way that we can use the code from the Github workflow instead of explicitly pulling from git inside the image?

roigcarlo:
Yes. We can tell the container to mount a directory containing the repo instead of downloading it. We can even add a flag or a switch to chose between one mode or the other

armingeiser:
Alternatively one could switch to checking out the actual PR commit: 
https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit

roigcarlo:
I had another approach in mind:

```yml
jobs:
  ubuntu:
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      matrix:
        build-type: [Custom, FullDebug]
        compiler: [gcc, clang]
    env:
      KRATOS_BUILD_TYPE: ${{ matrix.build-type }}
      OMPI_MCA_rmaps_base_oversubscribe: 1
      OMPI_MCA_btl_vader_single_copy_mechanism: none

    container:
      image: kratosmultiphysics/kratos-wheelbuilder-linux:latest
      options: --user 1001

    steps:
    - uses: actions/checkout@v2

    # Same as the current job
    - name: Build
      ...

    # Here we insert the wheel build from "scripts/wheels/linux/build.sh"
    - name: Build Wheel

    # Test now change, we install from wheels instead of setting the paths and run directly from python:
    - name: Install Wheels
      ...

    # And then run the tests (Note that here the sets are deleted.)
    - name: Running small tests
      shell: bash
      run: |
        source /opt/intel/oneapi/setvars.sh
        python3 kratos/python_scripts/run_tests.py -l small -c python3
```

So at the end, 
- The compilation process for the testing workflows with unitary builds (the 4 fast ones Custom/FullDebug + Gcc/Clang) remains the same
- An extra step is added to build wheels from the result of the output above (instead of using it directly)
- We install the wheels in the native python of the container.
- We tests normally but Kratos is used as a python package. 

On more detail my idea was to split the `build.sh` in parts and specifically separate the wheel creation process in a stand alone script so it can be directly be called by the CI in wheel builder image but without having to set up another container inside the one running by the CI ( I suspect this is also slowing down the compilation)

roigcarlo:
Actually thinking on this, we could share the compilation and then test the wheels and the results of the compilation in the same job (but I still think that doing just the wheels one is enough)

armingeiser:
So if i understand correctly:

- build as it is done now in the standard CI jobs
- still ev. run some tests using the build result (e.g. by setting the paths)
- new step: create wheel from the result of the compilation 
- new step: install wheel
- new step: test using the installed wheel

The container in container thing does not slow down a lot:
> With unity build and Custom build mode, the build time is down to 38min (still including the cpp tests)

5-10 min of these 38 minutes is actually for creating the wheels, not for building

This is fine for me, and i think fits well for a `on: pull_request` schedule.
Only thing to keep in mind is that the wheelbuilder image is a different linux version, so we should still establish some process to test the wheels that before uploading to PyPi

roigcarlo:
Exactly that would be the idea. And I would let a specific CI for the wheels image (but something more in line with the core-without-unity, in terms of workload). At the end it is only a subset of the features that we use in the regular Linux image (and the windows one is exactly the same).

Eventually when it works we will run it nightly in the server so we will have that covered there as well in more detail.

PR #9510, opened 2022-01-03
[Core] FindNodalNeighboursForEntitiesProcess - Avoid error clearing neighbours

ddiezrod:
**📝 Description**
This was added some time ago on our side by @jrubiogonzalez . If the output variable was not assigned to the nodes in the first place, calling `ClearNeighbours` will lead to a segmentation fault.

There might be other workarounds to fix this, but on some occasions, it might not so easy. I think its worthy in any case to add this check as it should be quite cheap and give us more confidence that this will not crash (again) in the future.  

**🆕 Changelog**
-Adding a check in ClearNeighBours from FindNodalNeighboursForEntitiesProcess to see if variable actually exists before clearing. If not an empty new object is assigned to the node.


jrubiogonzalez:
Thanks @ddiezrod, I created a PR more than one year ago and I completely forgot about it. Thanks for taking the lead!

sunethwarna (approved):

ddiezrod:
thanks @sunethwarna

PR #9512, opened 2022-01-04
[ROM] HROM settings in rom basis output process

rubenzorrilla:
**📝 Description**
Following the ideas in discussed in #9503 and #9504, this PR adds the required HROM settings to the `RomParameters.json` file. Note that its addition is optional according to the `append_hrom_settings` field, which I defaulted to `false`. I firstly set it to `true` but I think it's safer to do it in this way. Besides this, I want to point out that the settings are created such that the fields that are likely to be edited manually by the user remain at the top of the file to avoid browsing through it when we have large models (this is only ensured if Python 3.6 onward is used).


applications/RomApplication/python_scripts/calculate_rom_basis_output_process.py
```diff
@@ -109,6 +113,27 @@ def PrintOutput(self):
                     self.next_output += self.snapshots_interval
 
     def ExecuteFinalize(self):
+        # Initialize the Python dictionary with the default settings
+        # Note that this order is kept if Python 3.6 onwards is used
```
philbucher, line 113: Huhhhh this is dangerous 
Are you sure abt this? Do you have a reference?

in the CoSimApp we use OrderedDict for this purpose

rubenzorrilla: Well, more than dangerous in this case it would be annoying (if the order is not kept you'll need to browse through the file to manually edit something as the basis arrays might be way long).

In any case, [here](https://stackoverflow.com/questions/1867861/how-to-keep-keys-values-in-same-order-as-declared) it is explained.

philbucher: Cool, I didnt know!

applications/RomApplication/tests/test_calculate_rom_basis_output_process.py
```diff
@@ -35,14 +34,36 @@ def testCalculateRomBasisOutputProcess(self):
...
+        # Create a CalculateROMBasisOutputProcess instance
+        self.process_settings = KratosMultiphysics.Parameters("""{
+            "model_part_name": "MainModelPart",
+            "snapshots_control_type": "step",
+            "snapshots_interval": 1.0,
+            "nodal_unknowns": ["TEMPERATURE"],
+            "rom_basis_output_format": "json",
+            "rom_basis_output_name": "RomParametersWithHRom",
+            "svd_truncation_tolerance": 1.0e-6,
+            "append_hrom_settings": true
```
Rbravo555: I thought the `append_hrom_settings` was to be defalted to `false`

Rbravo555 (commented):
I think this modification to the output process should not be optional. One should have the extra fields in the `RomParameters.json`, even if only requiring to run simulations without hyper-reduction. Although probably I am missing something. Could you explain in which sense it is "safer", as you mentioned, to do it this way?

Moreover, this PR alone will add the fields to the RomParameters.json, but setting for example   
`"train_hrom": true` once the file is created, will not trigger any additional action. Should we 
 take care of using these flags within the workflow in this PR or probably it's better to do it in other ones?

rubenzorrilla:
IMO is safer because the HROM (as far as I know) is not working for all the physics yet. In any case, I agree that in the mid-term these should be always written by default. Maybe we can keep it simple and do what you suggest.

applications/RomApplication/tests/test_calculate_rom_basis_output_process.py
rubenzorrilla: I changed my mind while doing the PR. In any case, I'd follow your suggestion to always write these files.

rubenzorrilla:
About the usage of these new fields, yes, I'd add it in an upcoming PR (see advances in #9504).

applications/RomApplication/tests/test_calculate_rom_basis_output_process.py
rubenzorrilla: Done.

Rbravo555 (approved):

PR #9514, opened 2022-01-04
removing slow tests

GuillermoCasas:
**📝 Description**
Removing slow Night tests that were failng

**🆕 Changelog**
- For the moment, commenting slow tests


philbucher:
instead of removing I suggest to move them to the validation suite (this is also what I do with tests that are too slow for the CI)

this way they can still be executed locally

philbucher:
thx a lot for fixing btw :)

ipouplana (approved):
Thanks @GuillermoCasas

PR #9516, opened 2022-01-04
[ShapeOpt] split code into header source

### LINKED ISSUE #9513: [ShapeOpt] Split ShapeOptimizationApplication code into header/source files
**Description**
ShapeOptimizationApplication should be split into header/source. This has been pointed out several times in the past by @philbucher

armingeiser:
**📝 Description**
Split the cpp code into header and source files.

No other changes have been made, only copied the function implementation to the .cpp file. 

closes #9513

### DIFF
--- a/applications/ShapeOptimizationApplication/custom_utilities/damping/damping_utilities.cpp
+++ b/applications/ShapeOptimizationApplication/custom_utilities/damping/damping_utilities.cpp
@@ -1,0 +1,172 @@
+// ==============================================================================
+//  KratosShapeOptimizationApplication
+//
+//  License:         BSD License
+//                   license: ShapeOptimizationApplication/license.txt
+//
+//  Main authors:    Baumgaertner Daniel, https://github.com/dbaumgaertner,
+//                   Willer Matthias, https://github.com/matthiaswiller
+//
+// ==============================================================================
+
+// ------------------------------------------------------------------------------
+// System includes
+// ------------------------------------------------------------------------------
+#include <iostream>
+#include <string>
+#include <algorithm>
+
+// ------------------------------------------------------------------------------
+// Project includes
+// ------------------------------------------------------------------------------
+#include "custom_utilities/damping/damping_utilities.h"
+#include "includes/define.h"
+#include "includes/model_part.h"
+#include "spatial_containers/spatial_containers.h"
+#include "utilities/builtin_timer.h"
+#include "utilities/parallel_utilities.h"
+#include "custom_utilities/filter_function.h"
+#include "shape_optimization_application.h"
+
+// ==============================================================================
+
+namespace Kratos
+{
+
+DampingUtilities::DampingUtilities( ModelPart& modelPartToDamp, Parameters DampingSettings )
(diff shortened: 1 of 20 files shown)


philbucher:
a) seems like you are missing some `KRATOS_API` 
b) how much fast does it compile now?

applications/ShapeOptimizationApplication/custom_utilities/filter_function.cpp
```diff
@@ -0,0 +1,71 @@
...
+    else if (FilterFunctionType == "linear")
+        mFilterFunctional =  [](double radius, double distance) {return std::max(0.0, (radius - distance) / radius);};
+
+    // Type 3: Constant function
+    else if (FilterFunctionType == "constant")
+        mFilterFunctional = [](double radius, double distance) {return 1.0;};
+
+    // Type 4: Cosine function
+    else if (FilterFunctionType == "cosine")
+        mFilterFunctional = [](double radius, double distance) {return std::max(0.0, 1-0.5*(1-cos(3.1415926535897932384626433832795/radius*distance)));};
```
philbucher:
```suggestion
        mFilterFunctional = [](double radius, double distance) {return std::max(0.0, 1-0.5*(1-std::cos(Globals::Pi/radius*distance)));};
```

armingeiser:
>  b) how much fast does it compile now?

I tried to isolate the build of the ShapeOptApp.
Seems like its aktually 10s slower now (3 runs).

I did not expect a major improvement since all headers in the shapeopt app are included only once when exposing to python. But i did not expect a slow down...

philbucher:
hm good question
I guess you don't have many cross-includes

Probably it will make a noticeable difference when you change one cpp file now
previously you would have to recompile a lot more, now not any more :)

applications/ShapeOptimizationApplication/custom_utilities/damping/damping_utilities.cpp
```diff
@@ -0,0 +1,172 @@
...
+// Project includes
+// ------------------------------------------------------------------------------
+#include "custom_utilities/damping/damping_utilities.h"
+#include "includes/define.h"
+#include "includes/model_part.h"
+#include "spatial_containers/spatial_containers.h"
+#include "utilities/builtin_timer.h"
+#include "utilities/parallel_utilities.h"
+#include "custom_utilities/filter_function.h"
+#include "shape_optimization_application.h"
```
philbucher, line 29: this is only necessary as the Variables are declared in this file
Another cleanup would be to move the vars to a separate file as is done for the other apps :)

philbucher (approved):
I didn't review but if it passes the CI I would say it is fine

I saw that there are some loops that could easily be parallelized, but this is for the future :)

applications/ShapeOptimizationApplication/custom_utilities/damping/damping_utilities.cpp
armingeiser: you mean importing the application? Yes, variables can be moved to another file in another PR =)

PR #9522, opened 2022-01-09
[Wheel] Proposal of wheel build options for CI

armingeiser:
A few proposed updates to the wheel building process that would improve how wheels can be tested in CI (#9509).

### DIFF
--- a/scripts/wheels/linux/build.sh
+++ b/scripts/wheels/linux/build.sh
@@ -2,10 +2,15 @@
 # PYTHONS=("cp36" "cp37" "cp38" "cp39" "cp310")
 PYTHONS=("cp38")
 export KRATOS_VERSION="9.0.0"
+export KRATOS_BUILD_TYPE="Custom"
+
+export HASH=$(git show -s --format=%h) # Used in version number
+if [ "${KRATOS_BUILD_TYPE}" != "Release" ] then
+    export KRATOS_VERSION="${KRATOS_VERSION}.dev-${HASH}"
+fi
 
 BASE_LD_LIBRARY_PATH=$LD_LIBRARY_PATH
 export KRATOS_ROOT="/workspace/kratos/Kratos"
-export KRATOS_BUILD_TYPE="Custom"
 WHEEL_ROOT="/workspace/wheel"
 WHEEL_OUT="/data_swap_guest"
 CORE_LIB_DIR="/workspace/coreLibs"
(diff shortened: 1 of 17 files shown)


scripts/wheels/linux/build.sh
```diff
@@ -5,10 +5,21 @@ export KRATOS_VERSION="9.0.0"
 
 BASE_LD_LIBRARY_PATH=$LD_LIBRARY_PATH
 export KRATOS_ROOT="/workspace/kratos/Kratos"
+export KRATOS_BUILD_TYPE="Custom"
```
armingeiser: i suggest to allow calling the docker image with this as env variable instead of setting it here (or command option like the branch name) This would allow to run these tests with Custom mode for faster builds)

scripts/wheels/linux/build.sh
```diff
@@ -2,13 +2,29 @@
 # PYTHONS=("cp36" "cp37" "cp38" "cp39" "cp310")
 PYTHONS=("cp38")
 export KRATOS_VERSION="9.0.0"
+export KRATOS_BUILD_TYPE="Custom"
+
+export HASH=$(git show -s --format=%h) # Used in version number
+if [[ "${KRATOS_BUILD_TYPE}" != "Release" ]]; then
+    export KRATOS_VERSION="${KRATOS_VERSION}.dev.${HASH}"
```
armingeiser, line 9: alternatives e.g. `git describe --tag` -> `v9.0-<commits since tag>-<git hash> or also many python packages exist for this.

armingeiser: including `dev` in the version name explicitly distinguishes from the actual release and is e.g. not installed using standard `pip install` command - requires  `pip install --pre`

scripts/wheels/linux/build.sh
```diff
@@ -2,13 +2,30 @@
 # PYTHONS=("cp36" "cp37" "cp38" "cp39" "cp310")
 PYTHONS=("cp38")
 export KRATOS_VERSION="9.0.0"
+export KRATOS_BUILD_TYPE="Custom"
+
+export HASH=$(git show -s --format=%h) # Used in version number
+if [[ "${KRATOS_BUILD_TYPE}" != "Release" ]]; then
+    DATE=$(date '+%Y%m%d')
+    export KRATOS_VERSION="${KRATOS_VERSION}.dev${DATE}"
```
armingeiser, line 10: including `dev` in the version name explicitly distinguishes from the actual release and is e.g. not installed using standard pip install command - requires pip install flag`--pre`.

The date would still allow to have e.g. a nightly version that can be updated consistently using pip install.

PR #9524, opened 2022-01-10
[ConstitutiveLawsApp] Minor flag change in Parallel Rule of mixtures

atfard:
**📝 Description**
Minor change in P-RoM, from 6hrs to 10 min difference in calculation time

@AlejandroCornejo has supervised this also.


AlejandroCornejo (approved):
Necessary changes, it optimizes a lot the CL

SergioJimenezReyes (approved):

PR #9526, opened 2022-01-10
Adding debug check for nonshallow serialization of globalPointer in mpi

roigcarlo:
**📝 Description**
#9468

It's a draft, still not sure this is what I want, but we can discuss here.

**🆕 Changelog**
Adding debug warning for globalpointers

### DIFF
--- a/kratos/includes/global_pointer.h
+++ b/kratos/includes/global_pointer.h
@@ -239,6 +239,10 @@ public:
         rSerializer.save("D", mDataPointer);
       }
 #ifdef KRATOS_USING_MPI
+  #ifdef KRATOS_DEBUG
+      // Calling with a serializer not for which is not the MPI one or has SHALLOW_GLOBAL_POINTERS_SERIALIZATION disabled
+      KRATOS_WARNING_IF(ParallelEnvironment::GetDefaultDataCommunicator().IsDistributed() && !rSerializer.Is(Serializer::MPI) && !rSerializer.Is(Serializer::SHALLOW_GLOBAL_POINTERS_SERIALIZATION)) << "Calling non-shallow serialization of a globalPointer in an mpi context. This is probably an error." << std::endl;
+  #endif
       rSerializer.save("R", mRank);
 #endif
   }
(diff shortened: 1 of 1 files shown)


kratos/includes/global_pointer.h
```diff
@@ -239,6 +239,10 @@ class GlobalPointer {
         rSerializer.save("D", mDataPointer);
       }
 #ifdef KRATOS_USING_MPI
+  #ifdef KRATOS_DEBUG
+      // Calling with a serializer not for which is not the MPI one or has SHALLOW_GLOBAL_POINTERS_SERIALIZATION disabled
+      KRATOS_WARNING_IF(ParallelEnvironment::GetDefaultDataCommunicator().IsDistributed() && !rSerializer.Is(Serializer::MPI) && !rSerializer.Is(Serializer::SHALLOW_GLOBAL_POINTERS_SERIALIZATION)) << "Calling non-shallow serialization of a globalPointer in an mpi context. This is probably an error." << std::endl;
```
philbucher, line 244: maybe instead of asking the default data comm we could add a flag to the Serializer if it is distributed or not
or is this already the `Serializer::MPI` flag?

roigcarlo: Well sort of, that flag is actually never used but is set only if you construct the serializer with the MPI interface `Kratos/kratos/includes/mpi_serializer.h`.

The mpi serializer itself is used for things other than MPI (I know). If I am not wrong those usages are in the `MultiLevelMonteCarlo` and `ComprssiblePotentialFlowApplication`. At the end we are using MPISerializer as an alias for a Serializer with the SHALLOW_GLOBAL_POINTERS_SERIALIZATION flag set.

TLDR, MPI just indicates if the serializer was constructed using `Serializer` or `MPISerializer`. If we change the behavior of the existing usages (I think there just a couple) and we trust that nobody is gonna use the MPISerializer because they want we could use it.

philbucher: Hm I see

I am a bit concerned abt the proposed check as this will break the structure running in serial if Kratos runs in MPI (if we use e.g. the find_global_neighbors_process for the structure)

Any other way to check this?

roigcarlo: I always forget that case, you are right...

roigcarlo: We could do something like this, but not sure if we could relay in the DefaultComm, as you say it may change...

```C++
KRATOS_WARNING_IF(ParallelEnvironment::GetDefaultDataCommunicator().Rank() != mRank && !rSerializer.Is(Serializer::SHALLOW_GLOBAL_POINTERS_SERIALIZATION)) << "Calling non-shallow deserialization of a globalPointer in an mpi context. This is probably an error." << std::endl;
```

Issue #7877, opened 2020-11-25
[Core] PointOutputProcess is not an output process (no IsOutputStep nor PrintOutput)


frawahl:
**Description**
An update of the visualization_variables of elements, which are cut, of the visualization_model_part does not seem to happen every time step. E.g. if a PointOutputProcess (or one of the derived classes) creates output every time step for a cut element, the values in the output file of the PointOutputProcess won't change for a certain number of time steps (often until a GiD or VTK Output of the visualization model part happens). The output of works for the standard FluidModelPart.fluid_computational_model_part.

**Scope**
- EmbeddedSkinVisualizationProcess

**To Reproduce**
- Use `embedded_symbolic_navier_stokes_discontinuous` as element_type
- Add `ApplyEmbeddedSkinVisualizationProcess` in your ProjectParameters to have a visualization_model_part,( e.g. "visualization_variables": ["VELOCITY","PRESSURE"])
- Add a `PointOutputProcess` using the visualization_model_part for a point in an element, which is cut (e.g. "output_variables": [ "PRESSURE"])
- Add a `GiDOutputProcess`, which output_interval is bigger than 1 (e.g. "nodal_results"       : ["VELOCITY","PRESSURE"])
- Run a couple of time steps (more that output_interval of your GiDOutputProcess)
example see [visOutput.zip](https://github.com/KratosMultiphysics/Kratos/files/5599021/visOutput.zip) --> first output point of the LineOutputProcess is a cut element, the remaining points not

**Expected behavior**
The files created by a PointOutputProcess writes updated results for every time step, including elements, which are cut.

**Environment**
- OS: Ubuntu 20.04.1
- Branch: master
- Python 3.7.4

rubenzorrilla:
The `EmbeddedSkinVisualizationProcess` updates the variable values within the `ExecuteBeforeOutputStep` method. This means that these are only updated when the `IsOutputStep` of the output process is called. This is intentionally done to avoid the overhead of updating the output values when these are not to be printed.

The poblem in here is that the `PointOutputProcess` does not behave as an output process (even though it is). IMO, the problem in here is the `PointOutputProcess` which behaves as an standard process, not as an output one. Long story short, this process should be placed in the output processes list in the json an then implement the `IsOutputStep`. Then the `AnalysisStage` would call the `ExecuteBeforeOutputStep` of the visualization process to update the values.

[...]

PR #9529, opened 2022-01-11
[AnalysisStage] [OutputProcesses] Unify processes

miguelmaso:
**📝 Description**
Several output processes does not derive from `OutputProcess` and does not implement `IsOutputStep` nor `PrintOutput`. This situation can lead to some confusions (e.g. #7877) and and makes difficult using them when operations are done at `ExecuteBeforeOutputStep`.

However, modifying the interface of that "output processes" may break the current simulations, since they are constructed as regular processes.

In this draft I'm proposing a possible solution to ensure a smooth transition. The key is to detect if an `OutputProcess` is constructed as a regular process. When implementing these lines I've seen three possible choices:

1. Throw a warning and append the output process to the output processes list. (this is the implementation in this draft)
2. Throw a warning and do nothing. This will be more disruptive, since the modified process will stop printing output until the user construct it as an OutputProcess.
3. Don't throw a warning and append the output process to the list of output processes. In the future we could remove the `output_processes` block in the project parameters and unify the `processes`.

I'm liking option 3 more. Any opinions @KratosMultiphysics/technical-committee @KratosMultiphysics/implementation-committee ?

**🆕 Changelog**
- Detect if an `OutputProcess` is constructed as a regular process

### DIFF
--- a/kratos/python_scripts/analysis_stage.py
+++ b/kratos/python_scripts/analysis_stage.py
@@ -353,8 +353,8 @@ class AnalysisStage(object):
         deprecated_output_processes    = self._CheckDeprecatedOutputProcesses(self._list_of_processes)
         order_processes_initialization = self._GetOrderOfOutputProcessesInitialization()
         self._list_of_output_processes = self._CreateProcesses("output_processes", order_processes_initialization)
-        self._list_of_output_processes.extend(deprecated_output_processes)
         self._list_of_processes.extend(self._list_of_output_processes) # Adding the output processes to the regular processes
+        self._list_of_output_processes.extend(deprecated_output_processes)
 
     def __CheckIfSolveSolutionStepReturnsAValue(self, is_converged):
         """In case the solver does not return the state of convergence
--- a/kratos/python_scripts/analysis_stage.py
+++ b/kratos/python_scripts/analysis_stage.py
@@ -330,6 +330,16 @@ class AnalysisStage(object):
         """
         return []
 
+    def _CheckDeprecatedOutputProcesses(self, list_of_processes):
+        deprecated_output_processes = []
+        for process in list_of_processes:
+            if issubclass(type(process), KratosMultiphysics.OutputProcess):
+                deprecated_output_processes.append(process)
+                msg  = "{} is an OutputProcess. However, it has been constructed as a regular process.\n"
+                msg += "Please, define it as an 'output_processes' in the ProjectParameters."
+                IssueDeprecationWarning("AnalysisStage", msg.format(process.__class__.__name__))
+        return deprecated_output_processes
+
     def _GetSimulationName(self):
         """Returns the name of the Simulation
         """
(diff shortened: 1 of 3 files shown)


rubenzorrilla:
I'm very positive about this. Just mentioning that I don't think we can unify output and standard processes in a unique list as you point in option 3 since their API is slightly different. I also think it is good to keep separated to avoid polluting the "standard" processes with output stuff (note that we might eventually need to extend it in the future).

miguelmaso:
> Just mentioning that I don't think we can unify output and standard processes in a unique list as you point in option 3 since their API is slightly different.

What I'm trying to say in option 3 is to unify the dictionaries in the ProjectParameters and to separate the processes in the AnalysisStage using this mechanism. There should be a specific list for the output processes to impose an order of initialization. BTW its only an option and I don't have a strong opinion.

> I also think it is good to keep separated to avoid polluting the "standard" processes with output stuff (note that we might eventually need to extend it in the future).

I agree

philbucher:
Very cool, I hadn't thought of this before!
I really like option 1! If it is backward compatible it is very nice

Regarding the overall settings in the ProjectParams I though of even extending it
E.g. "initializaition_processes", "output_processes", "boundary_condition_processes" etc
=> Then the different process classes could have different APIs (e.g. "initialization_processes" only have "Execute" but not "ExecuteInitializeSolutionStep" etc
But this is besides the point of this PR
Hence I propose to make this one work backward compatible and then we can start the migration as e.g. pointed out in #7877

ddiezrod:
I also think option 1 is quite nice, I think its better to throw a warning so the user knows that they should modify their input file. As I see it, it should be possible to do it backward compatible, the only difference will be that it will print in PrintOutput, not in ExecuteFinalizeSolutionStep (or whatever function it was calling before to print).

philbucher:
from what I can see the currently implementation is already option 1 no?

miguelmaso:
ok, going for option 1

miguelmaso:
> from what I can see the currently implementation is already option 1 no?

yes, the current implementation is option 1

philbucher (dismissed):
:+1:

rubenzorrilla (dismissed):
I also think that this should go in as it is.

kratos/python_scripts/analysis_stage.py
```diff
@@ -330,6 +330,16 @@ def _GetOrderOfOutputProcessesInitialization(self):
         """
         return []
 
+    def _CheckDeprecatedOutputProcesses(self, list_of_processes):
+        deprecated_output_processes = []
+        for process in list_of_processes:
+            if issubclass(type(process), KratosMultiphysics.OutputProcess):
```
ddiezrod, line 336: Im not sure, is `issubclass` recursive? I mean, would this return true if a process is derived from another process which is finally derived from an output process?

miguelmaso: Is recursive. I've added a test to show that behavior

philbucher (approved):

kratos/python_scripts/analysis_stage.py
ddiezrod: thanks @miguelmaso !

Issue #9534, opened 2022-01-13
Discrepancy in the definition of variational redistance coefficients.


ddiezrod:
**Description**
Long time ago, @pablobecker did some tests with the variational redistance for our Molding solution in Altair. The problem they had is that this redistance was not very robust for them and in some cases it could completely ruin the case. The conclusion of these tests was that it was better to modify some of the coefficients that control the loads in the redistance, making it more robust but at the same time slower, needing more iterations to converge.
These are the differences that were introduced at that moment.
![image](https://user-images.githubusercontent.com/25667299/149331928-e2104397-69f9-4ab7-b947-28efacf8c59a.png)
https://github.com/KratosMultiphysics/Kratos/blob/dd7ee6320f8185188856331fb0ad21ba2d289884/kratos/elements/distance_calculation_element_simplex.h#L262

We've lived with this difference from github core for all this time, but it is a nightmare for maintenance, as many tests have slightly different results. We understand that you might prefer to maintain the current values even if solution is less robust. (At the end its quite hard to justify any value very rigorously)

I discussed this with @pooyan-dadvand and @jrubiogonzalez also and we though that maybe we could introduce this coefficients from element properties or from process info, so everybody can choose the coefficients that better work for them. What do you think @KratosMultiphysics/technical-committee ? @RiccardoRossi @rubenzorrilla @philbucher

philbucher:
Without having any clue abt the physics behind, this seems like a candidate for ProcessInfo (is it always the same within a ModelPart) or Properties (can it be different)

In any case this seems to be a „magic number“ so makes sense to have it variable

In StructuralMechanics we use sth similar for the damping coefficients, can look for it if it is interesting

rubenzorrilla:
Makes sense for me . As suggested by @philbucher this is should be in the `ProcessInfo`. For the moment we can add it to the solver defaults so we don't need to update the tests results. In the future we can change it to more robust values.

pooyan-dadvand:
Usually, it is constant during the calculation for this reason we thought `ProcessInfo` is not the best choice. But it is true that the solver knows this magic number rather than the users.

RiccardoRossi:
well, are we sure that the magic numbers suggested by Pablo aren't better than the ones we use? because i am not :-D

pablobecker:
[...]

PR #9535, opened 2022-01-13
[Core] Adding coefficients to variable redistance

ddiezrod:
**📝 Description**
This closes the problem exposed in https://github.com/KratosMultiphysics/Kratos/issues/9534. A couple of comments:
- I have no idea how to name these new variables, I thought it could be less confusing to just call them "coefficients" as its difficult to assign a very clear meaning to them, but Im very open to use more descriptive names.
- Variational redistance is not very clean right now. There are several things that could be improved: divide it in cpp/header, add constructor with settings, replace local flag with an option from settings... As its quite a lot I think it might be better to fix this, as its more urgent, and do the cleaning later.

**🆕 Changelog**
- Adding option to change some "magical" coefficients in variational redistance process from constructor.


kratos/includes/variables.h
```diff
@@ -470,6 +470,10 @@ namespace Kratos
     //for Geometry Variables
     KRATOS_DEFINE_3D_VARIABLE_WITH_COMPONENTS(PARAMETER_2D_COORDINATES)
 
+    // Variational redistance
+    KRATOS_DEFINE_VARIABLE(double, VARIATIONAL_REDISTANCE_COEFFICIENT_1)
```
loumalouomega: I don't like the names....

loumalouomega (commented):
I don't like the names. I think it is more elegant _FIRST and _SECOND. Not blocking, just saying

kratos/includes/variables.h
ddiezrod: Ok, I will rename them, if nobody has a better option

ddiezrod:

RiccardoRossi (dismissed):
ok

ddiezrod:
I renamed the coeffs, following @loumalouomega suggestion. Can someone reapprove?

rubenzorrilla (approved):

ddiezrod:
It seems failure in nightly tests come from other place. Merging.

PR #9537, opened 2022-01-14
Print info in file process updated

AlejandroCornejo:
**📝 Description**
I discovered that `double` entities where not printed correctly. Now it works. Initially was implemented supposing that tensors were printed.

### DIFF
--- a/kratos/python_scripts/print_info_in_file_process.py
+++ b/kratos/python_scripts/print_info_in_file_process.py
@@ -94,14 +94,20 @@ class PrintInfoInFileProcess(KratosMultiphysics.OutputProcess):
         else:
             for elem in self.model_part.Elements:
                 array_values = self.GetValueToPrint(elem)[self.integration_point]
-                for comp in range(len(array_values)):
-                    array_values[comp] = 0.0
+                if not isinstance(array_values, float): # Can be an scalar entity...
+                    for comp in range(len(array_values)):
+                        array_values[comp] = 0.0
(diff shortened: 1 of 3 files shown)


AlejandroCornejo:
> Can you add a test please?
hi, sure! One problem that I see of adding a test of IP variables is that I need some kind of `Element`... The elements that could work are inside the SMApp... What should we do?

philbucher:
In those cases you can add a `skiptest` when using elements from apps. E.g.:
https://github.com/KratosMultiphysics/Kratos/blob/dd7ee6320f8185188856331fb0ad21ba2d289884/kratos/tests/test_linear_constraints.py#L365-L367

AlejandroCornejo:
> In those cases you can add a `skiptest` when using elements from apps. E.g.:
> 
> https://github.com/KratosMultiphysics/Kratos/blob/dd7ee6320f8185188856331fb0ad21ba2d289884/kratos/tests/test_linear_constraints.py#L365-L367

That's magnificent! I'll do it

AlejandroCornejo:
@philbucher I've added a test in which I check the print of integration point variables: vectors (were working previously) and scalar entities (now bugfixed). THX!

philbucher (approved):
nice!
Also I recommend using the `pathlib` module instead of `os.path`. It is a "new" module that handles everything related to file paths much nicer and natively
https://docs.python.org/3/library/pathlib.html

PR #9541, opened 2022-01-17
defining -DBOOST_UBLAS_MOVE_SEMANTICS

RiccardoRossi:
this option should activate move semantics for ublas types

**📝 Description**
tentatively activating ublas semantics for ublas types

- improvement
-
**🆕 Changelog**
this option should activate move semantics for ublas types

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -517,6 +517,7 @@ include_directories( ${KRATOS_SOURCE_DIR}/external_libraries )
 
 # defines needed
 add_definitions( -DKRATOS_PYTHON )
+add_definitions( -DBOOST_UBLAS_MOVE_SEMANTICS )
 
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
   add_definitions( -fPIC )


roigcarlo (approved):
No problems on my side! :+1:

RiccardoRossi:
@KratosMultiphysics/technical-committee do you have any concern about  merging this?

philbucher:
Does it improve performance?
Are there any risks?

RiccardoRossi:
according to the docs it should (according to the docs, it removes the need for noalias in some cases)... however i don't have a benchmark showing it.

still, the capability would be very useful

RiccardoRossi:
just to add that the patch is there since about 2011 ... i just did not notice the option before

rubenzorrilla:
Just mentioning that this change might give some problems with `matrix` class implementation in old versions of Boost library (I was using 1.61 :older_adult:). Checking the implementation reveals that my ancient Boost library one was slightly different. Updating the library to the latest release (1.78) solves the issue.

AlejandroCornejo:
> Just mentioning that this change might give some problems with `matrix` class implementation in old versions of Boost library (I was using 1.61 🧓). Checking the implementation reveals that my ancient Boost library one was slightly different. Updating the library to the latest release (1.78) solves the issue.
> 
> EDIT: @KratosMultiphysics/all tagging everyone just in case this is also happening to you.

I am using 1.69 without any problem (as far as I know). Just for my curiosity, what problems did you have @rubenzorrilla ?

rubenzorrilla:
Basically it was not compiling  because of the assignment definition of `matrix` class (see `matrix.hpp`) when compiling with move semantics, which in my old version was
```c++
#ifdef BOOST_UBLAS_MOVE_SEMANTICS

        /*! @note "pass by value" the key idea to enable move semantics */
        BOOST_UBLAS_INLINE
        fixed_matrix &operator = (matrix m) {
            assign_temporary(m);
            return *this;
        }
#else
```
which in newer versions should be
```c++
#ifdef BOOST_UBLAS_MOVE_SEMANTICS

        /*! @note "pass by value" the key idea to enable move semantics */
        BOOST_UBLAS_INLINE
        fixed_matrix &operator = (fixed_matrix m) {
            assign_temporary(m);
            return *this;
        }
#else
```

PR #9426, opened 2021-12-03
[Fluid] Refactoring of compressible Navier-Stokes symbolic element generation

EduardGomezEscandell:
**📝 Description**
The necessity has arised to make the symbolic generator more flexible to allow for more types of geometry and variations of the formulation. 

**🆕 Changelog**

Deprecation:
- [x] Unused element `compressible_navier_stokes.(h|cpp)` and its symbolic generator are removed in favour of the new formulation.

Testing side:
- [x] Implemented tests to validate the producted element does not change
- [ ] Implemented tests to validate the producted element matches the one in `custom_elements`. [*Posponed to future PR*]

Symbolic side:
- [x] Moved the symbolic_generation directory to python_scripts (so it can be called from tests)
- [x] symbolic generator is now a class instead of a script. Parameters are passed as Kratos Parameters instead of being hardcoded
- [x] Definition of dimension, ngauss, nnodes, shape functions is moved to its own `symbolic_geometry` file
- [x] Addition of quadrilateral geometry to `symbolic_geometry`.

Element's side:
- [x] All non-specialized method templates are moved to the header file
- [x] The source file is split: a diferenent source file per class template specialization. This allows you to run each symbolic generator individually (remarkable time savings).

**🐘 Elephant in the room**
There are upwards of 40 files changed in this PR. I'm not proud of it. Either way, the majority of these changes are renamings and deletions, and the impact this has on `CMakeLists` and `#include`s.

**Directory changes**
The `symbolic_generation` directory is moved to `python_scripts`, and its internal folder structure is changed to:
```
symbolic_generation/
├── __init__.py
└── compressible_navier_stokes
    ├── __init__.py
    ├── compressible_navier_stokes_symbolic_generator.py
    ├── documentation_formulation.ipynb
    ├── generate_quadrilateral.py
    ├── generate_triangle.py
    ├── generate_tetrahedron.py
    ├── src
    │   └── dependencies called by compressible_navier_stokes_symbolic_generator.py
    └── templates
        ├── compressible_navier_stokes_explicit_template_2D3N.cpp
        ├── compressible_navier_stokes_explicit_template_2D4N.cpp
        └── compressible_navier_stokes_explicit_template_3D4N.cpp


```

PR #9544, opened 2022-01-17
[CI] Added sympy to CI dockers

EduardGomezEscandell:
**📝 Description**
There are some scripts within Kratos where sympy is imported (for instance [here](https://github.com/KratosMultiphysics/Kratos/blob/ci/add-sympy-to-docker/kratos/python_scripts/sympy_fe_utilities.py#L1)).

So far this hadn't been an issue because no file that used them was unit-tested. Now, however, the FluidDynamicsApplication will start testing its symbolic file generation in #9426.

**🆕 Changelog**
- Added sympy to ubuntu and centos dockers


scripts/docker_files/docker_file_ci_ubuntu_bionic/DockerFile
```diff
@@ -46,7 +46,8 @@ RUN apt-get update -y && apt-get upgrade -y && \
     python3 -m pip install --upgrade pip && \
     pip3 install \
         numpy \
-        scipy && \
+        scipy \
+        sympy && \
```
philbucher: this container is not use any more and also not automatically build
Hence I recommend not touching it any more, we will remove it eventually

roigcarlo:
@philbucher Is because of #9509 and #9522 or is there any other change I am not aware of?

philbucher:
> Edit: A ok, you meant the changes on the bionic container nvm

:)

philbucher (approved):

PR #9548, opened 2022-01-18
[Core] fix exposure of MemoryInfo

philbucher:
The exposure of `MemoryInfo` is broken since we moved to pybind many years ago :sweat_smile: 

this PR fixes this, also I exposed `HumanReadableSize`


rubenzorrilla (approved):
Thanks!

Issue #7713, opened 2020-10-30
[Core] Passing ProcessInfo in ModelPart check function


rubenzorrilla:
**Description**
Thanks to the warning that we fixed in #7712 @philbucher and I realised that the `Check` function of the `ModelPart` requires passing a `ProcessInfo` container.

Considering that the `ModelPart` owns the `ProcessInfo` this makes not so much sense. Having said so it probably makes sense to remove it. Even though this is an easy change, it would imply breaking the current API (silently for those applications that are not compiled in the CI).

philbucher:
I think that `ModelPart::Check` is not even called in our usual workflow, only the one of the `SolvingStrategy`, which doesn't call the one of the `ModelPart`

=> also we should decide who should do the checking of `Elements` etc (MP, Strategy, Scheme?)

pooyan-dadvand:
@KratosMultiphysics/technical-committee delegates completely this issue to the @KratosMultiphysics/implementation-committee

rubenzorrilla:
Ping @KratosMultiphysics/implementation-committee 

Summarizing things to be discussed:

- [ ] Remove `ProcessInfo` argument from `ModelPart` `Check` method.
- [ ] Who checks the `Elements`, `Conditions`, `Constraints`, etc? Through the strategy as we do know or should we start using the `ModelPart` check?

rubenzorrilla:
@KratosMultiphysics/implementation-committee is there any advance in this regard?

AlejandroCornejo:
Hi! On behalf of the @KratosMultiphysics/implementation-committee, We agree with @rubenzorrilla that the `ProcessInfo `has to be removed from the `check `method.

However, who performs the check is more complex. **One option** is to add the ModelPart->Check in the strategy and remove the Elements, conditions check from the Scheme.

Another option (easier one) is to modify the `Scheme `check implementation and call the `ModelPart->check` within the scheme. In this way we do not repeat code.

What is your opinion?

ddiezrod:
After some discussion in the @KratosMultiphysics/implementation-committee we think it might be safer to just go with the second option that @AlejandroCornejo proposed: modify the scheme to call the ModelPart::Check(), modifying the body of this function so it behaves exactly as it does right now in the scheme. If @KratosMultiphysics/technical-committee is with this, we'll proceed in the following days with the changes.

rubenzorrilla:
Should be solved in #9552. Thanks guys!

PR #9335, opened 2021-11-12
[Core][ParallelUtils] adding support for exceptions

philbucher:
This PR is a draft to add exception handling to the `ParallelUtilities`. With this the exceptions occurring inside shared memory parallel regions will be properly printed. Following [this article](https://docs.microsoft.com/en-us/cpp/parallel/concrt/convert-an-openmp-loop-that-uses-exception-handling?view=msvc-170) it was actually quite straight forward.

My solution works for OpenMP based parallelism as well as for C++11 based parallelism

I used a similar principle as `KRATOS_TRY` / `KRATOS_CATCH`

So far I only ported one loop, if the solution is accepted I can also port the other ones

closes #9259 

example of an error from within a parallel region:
~~~
    TestParUtilsExceptions Failed with message: 
        Error: The following errors occured in a parallel region!
Thread #2 caught exception: Error: inside parallel region

in kratos/tests/cpp_tests/utilities/test_parallel_utilities.cpp:424:auto Testing::TestParUtilsExceptions::TestFunction()::(anonymous class)::operator()(double &) const
Thread #0 caught exception: Error: inside parallel region

in kratos/tests/cpp_tests/utilities/test_parallel_utilities.cpp:424:auto Testing::TestParUtilsExceptions::TestFunction()::(anonymous class)::operator()(double &) const
Thread #4 caught exception: Error: inside parallel region

in kratos/tests/cpp_tests/utilities/test_parallel_utilities.cpp:424:auto Testing::TestParUtilsExceptions::TestFunction()::(anonymous class)::operator()(double &) const
Thread #1 caught exception: Error: inside parallel region

in kratos/tests/cpp_tests/utilities/test_parallel_utilities.cpp:424:auto Testing::TestParUtilsExceptions::TestFunction()::(anonymous class)::operator()(double &) const
Thread #3 caught exception: Error: inside parallel region

in kratos/tests/cpp_tests/utilities/test_parallel_utilities.cpp:424:auto Testing::TestParUtilsExceptions::TestFunction()::(anonymous class)::operator()(double &) const


[...]

PR #9552, opened 2022-01-18
[Core] moving check to model part and removing processinfo from arguments

ddiezrod:
**📝 Description**
As shown in https://github.com/KratosMultiphysics/Kratos/issues/7713 , right now there is a bit of confusion about who should Check the elements/conditions etc. Right this check is actually being called twice, once in the solving strategy and once again in the scheme. The solution proposed here is to call the model part check only from the strategy and add there the checks for the entities. I think this should not break anything unless there is a strategy somewhere that is not calling the base class (I did not find any)

**🆕 Changelog**
- Removed elements/conditions/constraints checks from solving_strategy and scheme base classes.
- Update model part Check so it does exactly what the scheme used to do.
- Call model part check from scheme.

### DIFF
--- a/kratos/sources/model_part.cpp
+++ b/kratos/sources/model_part.cpp
@@ -2207,7 +2207,6 @@ int ModelPart::Check() const
 {
     KRATOS_TRY
 
-    const int num_threads = ParallelUtilities::GetNumThreads();
     const ProcessInfo& r_current_process_info = this->GetProcessInfo();
 
     // Checks for all of the elements
--- a/kratos/sources/model_part.cpp
+++ b/kratos/sources/model_part.cpp
@@ -2203,27 +2203,36 @@ void ModelPart::SetBufferSizeSubModelParts(ModelPart::IndexType NewBufferSize)
 }
 
 /// run input validation
-int ModelPart::Check(const ProcessInfo& rCurrentProcessInfo) const
+int ModelPart::Check() const
 {
     KRATOS_TRY
-    int err = 0;
-    for (ElementConstantIterator elem_iterator = ElementsBegin(); elem_iterator != ElementsEnd(); elem_iterator++)
-    {
-        const auto& r_elem = *elem_iterator;
-        err = r_elem.Check(rCurrentProcessInfo);
-    }
-    for (ConditionConstantIterator condition_iterator = ConditionsBegin(); condition_iterator != ConditionsEnd(); condition_iterator++)
-    {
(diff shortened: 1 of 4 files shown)


kratos/sources/model_part.cpp
```diff
@@ -2203,27 +2203,36 @@ void ModelPart::SetBufferSizeSubModelParts(ModelPart::IndexType NewBufferSize)
...
-        const auto& r_constraint = *constraint_iterator;
-        err = r_constraint.Check(rCurrentProcessInfo);
-    }
-    return err;
+    //TODO: This is required for the exception handling. It can be removed once we move to the C++ parallelism
+    #ifdef KRATOS_SMP_CXX11
+        int num_threads = ParallelUtilities::GetNumThreads();
+    #else
+        int num_threads = 1;
+    #endif
```
philbucher: can be removed since #9335

philbucher (dismissed):
makes perfect sense to me :+1:

kratos/sources/model_part.cpp
ddiezrod: Oh yes, I forgot about that ;)

kratos/sources/model_part.cpp
```diff
@@ -2203,27 +2203,31 @@ void ModelPart::SetBufferSizeSubModelParts(ModelPart::IndexType NewBufferSize)
...
+
+    // Checks for all of the conditions
+    BlockPartition<const ModelPart::ConditionsContainerType>(this->Conditions(), num_threads).for_each([&r_current_process_info](const Condition& rCondition){
+        rCondition.Check(r_current_process_info);
+    });
+
+    // Checks for all of the constraints
+    BlockPartition<const ModelPart::MasterSlaveConstraintContainerType>(this->MasterSlaveConstraints(), num_threads).for_each([&r_current_process_info](const MasterSlaveConstraint& rConstraint){
+        rConstraint.Check(r_current_process_info);
+    });
```
philbucher:
```suggestion
    block_for_each(this->MasterSlaveConstraints(), [&r_current_process_info](const MasterSlaveConstraint& rConstraint){
        rConstraint.Check(r_current_process_info);
    });
```

same for the others

ddiezrod: done 👍

philbucher (approved):
perfect :+1:

ddiezrod:
merging ;)

PR #9556, opened 2022-01-18
[ROM] Clean up ROM B&S

rubenzorrilla:
**📝 Description**
This cleans up the ROM B&S. To be highlighted:
- The defaults mechanism is now properly used by adding and checking the base B&S settings (see get defaults and assignation methods)
- A `Create` method and `Name` have been added (working towards the factories...)
- There were many member variables that were not used since the base B&S are the ones that must be employed. The useless ones have been removed. Besides I renamed the misleading `mRomDofs` to `mNumberOfRomModes` (not in the input json, this is to be done in a future PR).
- Some code clean up has been done as well.
- I also took the chance to arrange the Doxygen.


Rbravo555:
What does `Doxygen` mean?

applications/RomApplication/custom_strategies/rom_builder_and_solver.h
```diff
@@ -371,150 +356,125 @@ class ROMBuilderAndSolver : public BuilderAndSolver<TSparseSpace, TDenseSpace, T
...
+        auto& r_dof_set = BaseType::GetDofSet();
+        const int dofs_number = r_dof_set.size();
+        const auto dofs_begin = r_dof_set.begin();
 
         #pragma omp parallel firstprivate(dofs_begin, dofs_number)
         {
             const Matrix *pcurrent_rom_nodal_basis = nullptr;
             unsigned int old_dof_id;
             #pragma omp for nowait
-            for (int k = 0; k < static_cast<int>(dofs_number); k++){
```
Rbravo555, line 386: I remember having done this to avoid errors when compiling in Windows. But I assumed it is compatible now, as it is passing the CI...

applications/RomApplication/custom_strategies/rom_builder_and_solver.h
```diff
@@ -371,150 +356,125 @@ class ROMBuilderAndSolver : public BuilderAndSolver<TSparseSpace, TDenseSpace, T
...
         KRATOS_ERROR_IF(!pScheme) << "No scheme provided!" << std::endl;
 
         // Getting the elements from the model
-        auto help_nelements = static_cast<int>(rModelPart.Elements().size());
-        auto help_el_begin = rModelPart.ElementsBegin();
+        // Only selected conditions and elements are used for the calculation in an H-ROM simulation
+        const auto el_begin = mHromSimulation ? mSelectedElements.begin() : rModelPart.ElementsBegin();
+        const int nelements = mHromSimulation ? mSelectedElements.size() : rModelPart.NumberOfElements();
+        const auto cond_begin = mHromSimulation ? mSelectedConditions.begin() : rModelPart.ConditionsBegin();
+        const int nconditions = mHromSimulation ? mSelectedConditions.size() : rModelPart.NumberOfConditions();
```
Rbravo555, line 436: nice!

Rbravo555 (commented):
It seems that in order for the workflow to incorporate these changes, we will also have to include the new fields in the KratosROM.RomResidualsUtility
I mean:
`    "echo_level": 1,
    "name": "rom_builder_and_solver",
`

applications/RomApplication/custom_strategies/rom_builder_and_solver.h
rubenzorrilla: As far as I know the issue with Windows parallelism is still there (`unsigned int` parallel for is not supported). This is why we used to cast to `int`. Note that in this case I'm implicitly casting to `int` when declaring the variable as `const int dofs_number = r_dof_set.size();`.

rubenzorrilla:
> What does `Doxygen` mean?

It is a tool for documentation (see [here](https://www.doxygen.nl/index.html)).

rubenzorrilla:
> It seems that in order for the workflow to incorporate these changes, we will also have to include the new fields in the KratosROM.RomResidualsUtility I mean: ` "echo_level": 1, "name": "rom_builder_and_solver",`

I guess this is because we're using the same parameters as input right? If this is the case I think we should manipulate them to pass what the `RomResidualsUtility` expects (we can do it in a future PR).

rubenzorrilla:
@Rbravo555 current workflow should work now.

Rbravo555 (approved):

PR #9558, opened 2022-01-19
[Fluid] Fixed error when modelpart is empty

EduardGomezEscandell:
**📝 Description**
PR #9426 introduces a bug that makes the program crash in `check_and_prepare_model_process_fluid.py` when a modelpart is empty (most typical case is in an unbalanced MPI). This slipped through the standard CI but got caught in nightly.

**🆕 Changelog**
- Added check to avoid getting element by Id when modelpart is empty

### DIFF
--- a/applications/FluidDynamicsApplication/python_scripts/check_and_prepare_model_process_fluid.py
+++ b/applications/FluidDynamicsApplication/python_scripts/check_and_prepare_model_process_fluid.py
@@ -67,7 +67,7 @@ class CheckAndPrepareModelProcess(KratosMultiphysics.Process):
             # Empty modelpart
             return
 
-        geometry = self.main_model_part.GetElement(1).GetGeometry()
+        geometry = self.main_model_part.Elements().__iter__().__next__().GetGeometry()
         is_simplex = geometry.LocalSpaceDimension() + 1 == geometry.PointsNumber()
         if not is_simplex:
             msg = "Geoemetry is not simplex. Orientation check is only available"
--- a/applications/FluidDynamicsApplication/python_scripts/check_and_prepare_model_process_fluid.py
+++ b/applications/FluidDynamicsApplication/python_scripts/check_and_prepare_model_process_fluid.py
@@ -35,7 +35,7 @@ class CheckAndPrepareModelProcess(KratosMultiphysics.Process):
     def _ElementsAreNotSimplex(self):
         "Checks whether the first element is non-simplex"
         if self.main_model_part.NumberOfElements() == 0:
-            return False
+            return True
         
         geometry = self.main_model_part.Elements.__iter__().__next__().GetGeometry()
         is_simplex = geometry.LocalSpaceDimension() + 1 == geometry.PointsNumber()
--- a/applications/FluidDynamicsApplication/python_scripts/check_and_prepare_model_process_fluid.py
+++ b/applications/FluidDynamicsApplication/python_scripts/check_and_prepare_model_process_fluid.py
@@ -67,7 +67,7 @@ class CheckAndPrepareModelProcess(KratosMultiphysics.Process):
             # Empty modelpart
             return
 
-        geometry = self.main_model_part.Elements().__iter__().__next__().GetGeometry()
+        geometry = self.main_model_part.Elements.__iter__().__next__().GetGeometry()
         is_simplex = geometry.LocalSpaceDimension() + 1 == geometry.PointsNumber()
         if not is_simplex:
             msg = "Geoemetry is not simplex. Orientation check is only available"
--- a/applications/FluidDynamicsApplication/python_scripts/check_and_prepare_model_process_fluid.py
+++ b/applications/FluidDynamicsApplication/python_scripts/check_and_prepare_model_process_fluid.py
@@ -32,6 +32,15 @@ class CheckAndPrepareModelProcess(KratosMultiphysics.Process):
         #self.list_of_inlets = Parameters["list_of_inlets"]
 
 
+    def _ElementsAreNotSimplex(self):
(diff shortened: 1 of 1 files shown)


rubenzorrilla (dismissed):

applications/FluidDynamicsApplication/python_scripts/check_and_prepare_model_process_fluid.py
```diff
@@ -63,6 +63,10 @@ def Execute(self):
         fluid_computational_model_part.AddConditions(list(list_of_ids))
 
         #verify the orientation of the skin (only implemented for tris and tets)
+        if self.main_model_part.NumberOfElements() == 0:
+            # Empty modelpart
+            return
+
         geometry = self.main_model_part.GetElement(1).GetGeometry()
```
philbucher: This will crash if you don’t have element with Id 1 in this partition
Please check the fluid solver how to get the first element (sth with __next__)

applications/FluidDynamicsApplication/python_scripts/check_and_prepare_model_process_fluid.py
```diff
@@ -63,6 +63,10 @@ def Execute(self):
         fluid_computational_model_part.AddConditions(list(list_of_ids))
 
         #verify the orientation of the skin (only implemented for tris and tets)
+        if self.main_model_part.NumberOfElements() == 0:
```
philbucher: Also what if there are no elements but there are conditions?

philbucher (changes requested):
Some things to clarify, this is sensitive for MPI

applications/FluidDynamicsApplication/python_scripts/check_and_prepare_model_process_fluid.py
EduardGomezEscandell: Then the check will fail here:

https://github.com/KratosMultiphysics/Kratos/blob/261342507dc747f540bd89d12a8b425a9e715ea5/kratos/processes/tetrahedral_mesh_orientation_check.cpp#L194-L196

In the future we will need to implement a `quadrilateral_mesh_orientation_check` but for now quadrilateral fluid elements' infrastrucure is still very much in development.

applications/FluidDynamicsApplication/python_scripts/check_and_prepare_model_process_fluid.py
EduardGomezEscandell: True. Fixed.

applications/FluidDynamicsApplication/python_scripts/check_and_prepare_model_process_fluid.py
philbucher: Since this is a detail of the `TetrahedralMeshOrientationCheck`, I would slightly modify the check:
~~~py
if self.main_model_part.NumberOfElements() > 0:
        geometry = self.main_model_part.Elements.__iter__().__next__().GetGeometry()
        is_simplex = geometry.LocalSpaceDimension() + 1 == geometry.PointsNumber()
        if not is_simplex:
            msg = "Geoemetry is not simplex. Orientation check is only available"
            msg += "for simplex geometries and hence it will be skipped"
            KratosMultiphysics.Logger.PrintWarning(type(self).__name__, msg)
            return
~~~

this way the overall flow is not changed

EduardGomezEscandell: Well this is going to modify the behaviour of the program beyond what this PR was trying to fix (i.e. the error introduced in #9426).

I don't know if the error thrown by `TetrahedralMeshOrientationCheck` was the expected behaviour or not. If the Fluid application required all conditions to be in the same rank as their parent element (which I think is the case), throwing the error would be good.

@rubenzorrilla thoughts?

philbucher: I would like to restore the original behavior, maybe I expressed myself badly
Your currently proposed solution modifies the original behavior regarding skipping the check if there are no elements in the partition

I propose to restore the old behavior by only skipping the check if it is not a simplex element

EduardGomezEscandell: Oh ok, agreed then.

rubenzorrilla:
Should be fine now right?

applications/FluidDynamicsApplication/python_scripts/check_and_prepare_model_process_fluid.py
philbucher: but you didn't push changes?

My concern is that now the `TetrahedralMeshOrientationCheck` is not called any more if there are no elements, hence my suggestion

EduardGomezEscandell: Yeah didn't have time to do it yesterday. They're up now.

applications/FluidDynamicsApplication/python_scripts/check_and_prepare_model_process_fluid.py
```diff
@@ -32,6 +32,15 @@ def __init__(self, main_model_part, Parameters ):
         #self.list_of_inlets = Parameters["list_of_inlets"]
 
 
+    def _ElementsAreNotSimplex(self):
+        "Checks whether the first element is non-simplex"
+        if self.main_model_part.NumberOfElements() == 0:
+            return False
```
philbucher:
```suggestion
            return True
```

otherwise again the check is skipped if there are no elements

philbucher (approved):

philbucher:
Thx

Issue #3070, opened 2018-10-16
[CORE] Enable calculation of "distance to skin" for user-defined DISTANCE variables


marcnunezc:
@rubenzorrilla @pooyan-dadvand @RiccardoRossi 

Dear all,

In my application, I am using two independent distance functions (one using the common DISTANCE variable and the other using an specific EXAMPLE_DISTANCE defined in the application). I would need to use the _calculate distance_ processes for both variables, which is currently not possible as the DISTANCE variable is directly defined inside the processes. For instance (in _calculate_signed_distance_to_3d_skin_process.h_ ): 

>     void CalculateNodeDistance(Node<3>& rNode)
>     {
>         double coord[3] = {rNode.X(), rNode.Y(), rNode.Z()};
>         double distance = DistancePositionInSpace(coord);
>         double& node_distance =  rNode.GetSolutionStepValue(DISTANCE);

I am opening this issue to discuss how we could redefine this processes in order to compute the distances for any user-defined variable (not only the default DISTANCE variable). 

Would it be OK to add two properties in these processes objects (e.g mrDistanceVariable and mrElementalDistancesVariable), which could be defined through the constructor as an input, or defaulted to DISTANCE/ELEMENTAL_DISTANCES otherwise?

Thanks,

philbucher:
Another way would be to use the process as is and then copy the computed to other variables
=> this can easily be done with the [VariableUtils](https://github.com/KratosMultiphysics/Kratos/blob/master/kratos/utilities/variable_utils.cpp#L209-L224)

pooyan-dadvand:
Not a bad idea, I would pass it in the constructor with default to be DISTANCE in `CalculateDistanceToSkinProcess` and ELEMENTAL_DISTANCES in `CalculateDiscontinuousDistanceToSkinProcess`

philbucher:
is this solved?

marcnunezc:
Hi!
Not yet!
Pooyan suggestion is not yet implemented, but I have been using the VariableUtils as you pointed out to do the workaround.

philbucher:
@marcnunezc was this part of your recent implementations?

marcnunezc:
No it's not yet done!

I started to implement it [here](https://github.com/KratosMultiphysics/Kratos/tree/core/user-defined-vars-skin-process) some weeks ago, but it's still pending

PR #9567, opened 2022-01-21
[Core] Use parameters in distance processes 

marcnunezc:
**📝 Description**
This is a first step to enable #3070 ( solved in  #9568). 

The use of flags as options is replaced and deprectaed (upon agreement) by the parameters.

Then the addition of user-defined variables is much easier. 

If you agree, the flag constructor would be deprecated. 

**🆕 Changelog**
E.g.
- Added new constructor with parameters for both discontinuous and continuous distance processes.
- Deprecated flag constructor. 
- Tests updated using the new constructor.


rubenzorrilla (commented):
I'm very very positive about this (the local flags export has been always a pain in the neck...). However, I don't see why we need to save the entire `mParameters` since we're saving the input values in the `mOptions` flag container. Alternatively, we could get rid of this `mOptions` by saving three `bool` type member variables.

marcnunezc:
> I'm very very positive about this (the local flags export has been always a pain in the neck...). However, I don't see why we need to save the entire `mParameters` since we're saving the input values in the `mOptions` flag container. Alternatively, we could get rid of this `mOptions` by saving three `bool` type member variables.

Actually, I started implementing your suggestion but I switched to the current one as I thought it was cleaner. 

Thinking about it again, I agree it's better to use bool member variables.

marcnunezc:
I defined bool member variables and updated the tests with the new constructor.

rubenzorrilla (commented):
Very good.  However I think that there are some more usages (at least in the `FSIApplication`) that will start throwing a warning. These should be amended (we can do it in a follow-up PR). Besides, I think that once this is merged, we can open another PR removing the deprecated constructors (this could be more tedious to be merged, @KratosMultiphysics/altair not sure if this will affect you).

marcnunezc:
> Very good. However I think that there are some more usages (at least in the `FSIApplication`) that will start throwing a warning. These should be amended (we can do it in a follow-up PR). Besides, I think that once this is merged, we can open another PR removing the deprecated constructors (this could be more tedious to be merged, @KratosMultiphysics/altair not sure if this will affect you).

Seems resonable.

I just checked the usage on the FSIApplication, and it appears to be here:
https://github.com/KratosMultiphysics/Kratos/blob/6c863e447d39b09c4ea9b0fc2bd2227009bc4ac2/applications/FSIApplication/python_scripts/partitioned_embedded_fsi_base_solver.py#L216-L229

However it does not seem to be covered by the tests (I put an Exception there and tests run fine), but we can take a look in a separate PR as you comment.

rubenzorrilla:
> > Very good. However I think that there are some more usages (at least in the `FSIApplication`) that will start throwing a warning. These should be amended (we can do it in a follow-up PR). Besides, I think that once this is merged, we can open another PR removing the deprecated constructors (this could be more tedious to be merged, @KratosMultiphysics/altair not sure if this will affect you).
> 
> Seems resonable.
> 
> I just checked the usage on the FSIApplication, and it appears to be here:
> 
> https://github.com/KratosMultiphysics/Kratos/blob/6c863e447d39b09c4ea9b0fc2bd2227009bc4ac2/applications/FSIApplication/python_scripts/partitioned_embedded_fsi_base_solver.py#L216-L229
> 
> However it does not seem to be covered by the tests (I put an Exception there and tests run fine), but we can take a look in a separate PR as you comment.

Might be not covered as the cases involving this solver are potentially big... In any case, lets do it in another one.

rubenzorrilla (approved):
:rocket:

Issue #7434, opened 2020-09-08
Classes VS Namespaces


RiccardoRossi:
**Description**

In PR #7039 a discussion started about exporting to python free functions as part of new user specified namespaces.

After discussion in @KratosMultiphysics/technical-committee we reached the consensus to keep the current usage, that is, to have static methods exported as members of classes and not as free functions within specific namespaces

The rationale of this is

- classes allow private members and methods while namespaces don't
- classes can be templated
- KRATOS_API has to be added only once

On the other hand the @KratosMultiphysics/technical-committee agrees in principle about adding a limited number of namespaces, on the line of "internal", "utilities", "processes" in order to better organize the code and to benefit of better code suggestions by modern editors. However to avoid the proliferation of such namespaces we require that this is agreed to by the @KratosMultiphysics/technical-committee

We shall also observe however that adding such namespaces for existing classes, for example for processes and utilities, would imply a very large effort, so we would need to consider this carefully before actually doing it.

This message should be ported to our coding style guides.

roigcarlo:
I've ported this to our wiki: https://github.com/KratosMultiphysics/Kratos/wiki/Namespaces-vs-Static-Classes

PR #9569, opened 2022-01-21
[Fluid] Improving symbolic generator tests

EduardGomezEscandell:
**📝 Description**
The current way of testing (purely comparing text files generated) is sub-optimal because sometimes `sympy` rearranges the order of operations when minor changes are made to the generator (`a+b` may become `b+a`). These would cause the text comparison to fail, even if the behaviour of the file is the same once compiled. Sometimes the rearangments are a bit more complex than the ones of the example and are hard to interpret even by reading the file.

A new test is introduced and performs a three-step test:
1. Calls the symbolic generator to create a `test.cpp` file.
2. Compiles this file to create binary `test.out` file.
3. Runs this binary, and uses its output to see if the test has passed or not.

**🆕 Changelog**
- Created new test that compiles the generated output from CompressibleNavierStokesSymbolicGenerator.
- Replaced text comparsion tests in nightly build
- Changed CMakeLists.txt so that the test header file is installed (by default `.h` are excluded from the python package, but this one is necessary to run the test).
- Changed nightly CI so that variable `CXX` is available during tests.

**Dependencies**
This PR depends upon #9583


EduardGomezEscandell:
Also running in nightly:

https://github.com/KratosMultiphysics/Kratos/actions/runs/1728948895

.github/workflows/nightly_build.yml
```diff
@@ -72,6 +72,14 @@ jobs:
...
         export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:${GITHUB_WORKSPACE}/bin/${KRATOS_BUILD_TYPE}/libs
         python3 kratos/python_scripts/run_tests.py -l nightly -c python3
+        if [ ${{ matrix.compiler }} = gcc ]; then
+          export CXX=/usr/bin/g++
+        elif [ ${{ matrix.compiler }} = clang ]; then
+          export CXX=/usr/bin/clang++
+        else
+          echo 'Unsupported compiler: ${{ matrix.compiler }}'
+          exit 1
+        fi
```
rubenzorrilla, line 82: This change must be done in a standalone PR.

applications/FluidDynamicsApplication/CMakeLists.txt
```diff
@@ -154,6 +154,9 @@ if(${INSTALL_TESTING_FILES} MATCHES ON )
         PATTERN "*.cpp" EXCLUDE
         PATTERN "*.hpp" EXCLUDE
       )
+
+    install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/tests/CompressibleSymbolicGeneration/compilable_symbolic_test.h
```
rubenzorrilla: By doing this we'd need to manually add any future test. Couldn't we make something similar to what we do above but for a `${CMAKE_CURRENT_SOURCE_DIR}/tests/symbolic_generation` generic folder?

PD: Note that our convention in the repo is to write subfolders in snake case (admittedly not always fulfilled...)

applications/FluidDynamicsApplication/tests/CompressibleSymbolicGeneration/compilable_symbolic_test.h
```diff
@@ -0,0 +1,312 @@
...
+#include <cmath>
+#include <ostream>
+#include <iostream>
+
+/**
+ * This namespace contains classes and function mimicking the necessary
+ * behaviour of Kratos in order to run the generated code without having
+ * to modify it.
+ */
+namespace TestCompressibleNavierStokesSymbolic {
```
rubenzorrilla: Addition of new namespaces must be technically justified (see #7434).

applications/FluidDynamicsApplication/tests/CompressibleSymbolicGeneration/compilable_symbolic_test.h
```diff
@@ -0,0 +1,312 @@
...
+
+/** Matrix-like class with only:
+ * - storage manipulation access
+ * - operator<<
+ * - almost equality check
+ */
+template<std::size_t TRows, std::size_t TCols>
+class Matrix
```
rubenzorrilla: Why you require your own matrix class? Can't you use Kratos bounded one?

EduardGomezEscandell: Two reasons:
- The Kratos source is not packaged with the Python package. This means that when running the test, the location of the kratos source is unknown. It may not even be in the machine at all if Kratos was installed with pip.
- This way compilation time is kept almost instantaneous, and the only cost of the test is the symbolic generation.

For these two reasons the cpp code in this test is completely isolated from Kratos.

rubenzorrilla (commented):
This adds the possibility of testing the automatic differentiation itself (not the already generated elements as we do now). However, it adds some extra complexities to our current testing mechanism so it needs to be discussed in the @KratosMultiphysics/technical-committee. Could you detail the test workflow for the next KTC meeting?

Besides, it would be nice to use our style guide also in here.

applications/FluidDynamicsApplication/CMakeLists.txt
EduardGomezEscandell: Well this was done kind of on purpose. When speaking with @roigcarlo he explained to me that `.h`, `.cpp`, etc. are blacklisted to avoid accidentally polluting the python package. I decided to keep it very restrictive in order to ensure this goal is still fulfilled.

I have no strong feelings one way or the other, just explaining my motivation.

Re: snake case. Not an issue, will change it.

EduardGomezEscandell:
> This adds the possibility of testing the automatic differentiation itself (not the already generated elements as we do now). However, it adds some extra complexities to our current testing mechanism so it needs to be discussed in the @KratosMultiphysics/technical-committee. Could you detail the test workflow for the next KTC meeting?

Do you mean joining the meeting or explaining it here/in a document? I'd be happy to do either.

> Besides, it would be nice to use our style guide also in here.

Yes, I was torn between trying to cram everything into a single file, or separating the`.h` into diferent files. I settled for the former because it isn't much code and I didn't want to add a lot of files to the test folder. In this regard I am open to sugestions.

I think this is the biggest concern style-wise. I can go over the file to fix some other issues as well.

rubenzorrilla:
> > This adds the possibility of testing the automatic differentiation itself (not the already generated elements as we do now). However, it adds some extra complexities to our current testing mechanism so it needs to be discussed in the @KratosMultiphysics/technical-committee. Could you detail the test workflow for the next KTC meeting?
> 
> Do you mean joining the meeting or explaining it here/in a document? I'd be happy to do either.
> 

I meant in here. Something like
1. we generate the elemtent
2. we do bla bla
3. etc...
will be enough.

> > Besides, it would be nice to use our style guide also in here.
> 
> Yes, I was torn between trying to cram everything into a single file, or separating the`.h` into diferent files. I settled for the former because it isn't much code and I didn't want to add a lot of files to the test folder. In this regard I am open to sugestions.
> 
> I think this is the biggest concern style-wise. I can go over the file to fix some other issues as well.

.github/workflows/nightly_build.yml
EduardGomezEscandell: Ok. Opened PR here #9583.

EduardGomezEscandell:
## Explanation of test
### Ingredients
- *C++ template file*: Code that tests the output of the generated code against a reference. The file contains sentinels left there for the symbolic generator to know where to write the generated code. There is one of such files for each test. I named them `*.template` since they may not be valid C++ due to the missing code.
- *Header file*: Provides some utilities that are common to various tests. Particulary, classes that mimic the behaviour of Kratos matrices. The generated code only uses `operator()(size_t, size_t)` and `operator[](size_t)`, so these matrix classes can be very simple.
- *Compiler* (CI runs only): The compiler must be known (e.g. via an environment variable). Otherwise it can be specified in the test.
- *Python test*. This file orchestrates the test and performs the tasks expaned below.

Note that the C++ code does not include Kratos, for the reasons explained earlier:
- The Kratos source is not packaged with the Python package. This means that when running the test, the location of the kratos source is unknown. It may not even be in the machine at all if Kratos was installed with pip.
- This way compilation time is kept almost instantaneous, and the only cost of the test is the symbolic generation.

So far there are two tests, one for each 2d geometry:
- Quadrilateral
- Triangle

Once it is better tested, tetrahedra will also be added.

### Step 1: Generation ###
The generator is called and run with default settings. It replaces the sentinels in the template file and generates a `.cpp` file.
```
Symbolic generator:   test.template --> test.cpp
```

### Step 2: Checking compiler exists ###
Simply calls `{compiler} --version` and checks the errorcode. If an error occurs, the test fails.
Arguably this could be moved to the beggining as step 0.

### Step 3: Compiling ###
Calls `{compiler} test.cpp -o test.out` to generate a test binary.
```
Compiler:   test.cpp, test.h  -->  test.out
```
I added some extra compiler flags (such as address sanitizer) so that some obvious bugs are caught and cause the test to fail with a more helpful message, but this is just an extra.

### Step 4: Running the test ###
Runs the binary, which returns an error code to indicate if it is succesful or not. Additionally it may print error information.

### Other considerations ###
*Why not generate and compile the code during the compilation of Kratos?*
While I suppose it would be possible to treat all of this as a `cpp_test`, and call the symbolic generator from CMake, this may significantly increase compilation time. The quadrilateral+triangle test takes about 2 minutes in my machine, and if a hexahedron is added at some point, it can take up to 5 minutes. This will take significantly longer on lower-spec'd machines. This seems quite negligible, since compiling all of Kratos may take half an hour, but keep in mind that unless some cool CMake code is written to conditionaly call the generator, we cannot exploit the incremental build, so even changing a single line in some random source file will trigger the symbolic generator, and take minutes instead of seconds to compile.

*Could we store a precompiled Kratos core to avoid creating our own classes here?*
Maybe. I don't know enough CMake, but it seems even more complicated and intrusive than having a faux matrix class.

*Some improvements*
If this is a one-off thing, I believe it is good as-is. If we want to consider adding more tests of this nature, we may want to make things cleaner by having a common header file/files for all tests with comonly needed utilis, such as the faux matrix class or some testing utils.

EduardGomezEscandell:
Just for reference, here is what a failed test looks like:
```
 |  /           |
 ' /   __| _` | __|  _ \   __|
 . \  |   (   | |   (   |\__ \
           Multi-Physics 9.0."Dev"-d005cde645-Release
Compiled with threading support.
Maximum number of threads: 4.
Importing    KratosFluidDynamicsApplication 
Initializing KratosFluidDynamicsApplication...

----------------Checking compiler exists----------------
clang version 10.0.0-4ubuntu1 
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin

--------------------Generating--------------------------
Computing geometry: quadrilateral (2D4N)
 - Compute Source Matrix
 - Compute Euler Jacobian matrix
 - Compute diffusive flux (shock capturing ON)
 - Compute stabilization matrix
 - Compute non-linear operator
 - Compute non-linear adjoint operator
 - Compute variational formulation
 - Calculate the projections of the residuals
   - Gauss point: 0
 - Subscales type: ASGS
 - Substitution of the discretized values at the gauss points
    Gauss point: 0
    - Compute stabilization matrix on Gauss pt.
    - Substitution in the variational formulation
    Compute RHS
    Substituting outstring from compilable_2D4N.template
 - Subscales type: OSS
 - Substitution of the discretized values at the gauss points
    Gauss point: 0
    - Compute stabilization matrix on Gauss pt.
    - Substitution in the variational formulation
    Compute RHS
    Substituting outstring from compilable_2D4N.template
 - Writing symbolic_test_2D4N.cpp
Geometry quadrilateral (2D4N) done

---------------------Compiling---------------------------

----------------------Testing----------------------------
Testing result of //substitute_rho_proj_2D: Failed
Testing result of //substitute_mom_proj_2D: Failed
Testing result of //substitute_tot_ener_proj_2D: Passed
Testing result of //substitute_rhs_2D_OSS: Passed
Testing result of //substitute_rhs_2D_ASGS: Passed
F
FAIL: test_Quadrilateral (__main__.CompressibleNavierStokesSymbolicGeneratorCompilationTest)
  File "/home/eduard/Work/Kratos/applications/FluidDynamicsApplication/tests/compressible_navier_stokes_symbolic_generator_compiled_test.py", line 154, in test_Quadrilateral
    self._GenerateCompileAndRun(**args)
  File "/home/eduard/Work/Kratos/applications/FluidDynamicsApplication/tests/compressible_navier_stokes_symbolic_generator_compiled_test.py", line 120, in _GenerateCompileAndRun
    self.assertEqual(errcode, 0, "Tests returned error code {}.\nStderr:\n{}\n".format(errcode, self._ReadTestStdErr()))
AssertionError: 256 != 0 : Tests returned error code 256.
Stderr:
Errors in result of //substitute_rho_proj_2D:
Mismatch in entry #  1:         0 !=         42 | Diff =        -42 | Rel diff = -1
Mismatch in entry #  3:         0 !=  6.022e+23 | Diff = -6.022e+23 | Rel diff = -1

Errors in result of //substitute_mom_proj_2D:
Mismatch in entry #  0:  323.4119 !=   3.141593 | Diff =   320.2703 | Rel diff = 0.9902861




Ran 1 test in 82.625s

FAILED (failures=1)
```

EduardGomezEscandell:
Closing because #9627 allows for much simpler testing without the need of compiling.

philbucher:
do you still need the branch?

EduardGomezEscandell:
> do you still need the branch?

No, you're right. Deleted.

PR #9489, opened 2021-12-27
[Mapping] Barycentric mapper

philbucher:
This PR adds the barycentric mapper to the MappingApp. It is a combination between the nearest neighbor and nearest element mapper, it reconstructs a geometry from the closest nodes and then interpolates by using the shape functions of this geometry

=> It hence only requires nodes as input and not geometries which makes it very suitable for particle methods where no geometries are available.

It is required to choose what kind of geometry should be used aka reconstructed for the interpolation: `line`, `triangle` or `tetrahedral`

PR #9576, opened 2022-01-24
[Shallow] 3d coupling with NS

miguelmaso:
**📝 Description**
minor changes. Leftover.

**🆕 Changelog**
- Print the `DOMAIN_SIZE` in the interface model part
- Apply the mapper settings according to the `DOMAIN_SIZE`
- Aesthetic changes in the custom output

### DIFF
--- a/applications/ShallowWaterApplication/python_scripts/coupling/depth_integration_input_process.py
+++ b/applications/ShallowWaterApplication/python_scripts/coupling/depth_integration_input_process.py
@@ -12,7 +12,7 @@ def Factory(settings, model):
         raise Exception("expected input shall be a Parameters object, encapsulating a json string")
     return DepthIntegrationInputProcess(model, settings["Parameters"])
 
-class DepthIntegrationInputProcess(KM.OutputProcess):
+class DepthIntegrationInputProcess(KM.Process):
     """DepthIntegrationInputProcess
 
     Read the depth integrated values from an HDF5 file and set them as boundary conditions.
@@ -40,7 +40,7 @@ class DepthIntegrationInputProcess(KM.OutputProcess):
     def __init__(self, model, settings):
         """The constructor of the DepthIntegrationInputProcess."""
 
-        KM.OutputProcess.__init__(self)
+        KM.Process.__init__(self)
         self.settings = settings
         self.settings.ValidateAndAssignDefaults(self.GetDefaultParameters())
 
@@ -73,6 +73,7 @@ class DepthIntegrationInputProcess(KM.OutputProcess):
     def ExecuteInitialize(self):
         '''Read the input_model_part and set the variables.'''
         self.hdf5_import.ExecuteInitialize()
+        self._CheckInputCoordinates()
         self._CheckInputVariables()
         self._CreateMapper()
         self._MapToBoundaryCondition()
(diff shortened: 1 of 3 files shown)


ipouplana (dismissed):

applications/ShallowWaterApplication/python_scripts/coupling/depth_integration_input_process.py
```diff
@@ -173,15 +182,15 @@ def _CreateMapper(self):
...
-        mapper_settings["search_settings"]["search_radius"].SetDouble(distance)
+        domain_size = self.input_model_part.ProcessInfo[KM.DOMAIN_SIZE]
+        if domain_size == 2:
+            min_point, max_point = KM.BoundingBox(self.interface_model_part.Nodes).GetPoints()
+            search_radius = 1.05 * (max_point - min_point).norm_2()
+        elif domain_size == 3:
+            search_radius = self.interface_model_part.Conditions.__iter__().__next__().GetGeometry().Length()
+        else:
+            raise Exception("DepthIntegrationInputProcess._CreateMapper: The domain size of the input_model_part is: ", domain_size)
+        mapper_settings["search_settings"]["search_radius"].SetDouble(search_radius)
```
philbucher: is it slow without specifying this?
I recently updated the search and made it a lot faster, but considering that you are using the new input I guess you are aware of it

philbucher: also I see that you are using the nearest neighbor, is this one sufficient for you in terms of accuracy?
Have you tried the nearest element?

or maybe even the [barycentric mapper](https://github.com/KratosMultiphysics/Kratos/pull/9489)

miguelmaso: Thanks for the comments @philbucher 
In some cases I have a cloud of points. Since accuracy is not crucial, its better to use the nearest neighbor in all the cases.
For the 3D case, now I'm using the default search radius, its cleaner

applications/ShallowWaterApplication/python_scripts/coupling/depth_integration_input_process.py
```diff
@@ -166,22 +175,26 @@ def _MapToBoundaryCondition(self):
...
+                "mapper_type": "nearest_neighbor",
+                "echo_level" : 0,
+                "search_settings" : {
+                    "search_radius" : 0.0
+                }
+            }""")
+            min_point, max_point = KM.BoundingBox(self.interface_model_part.Nodes).GetPoints()
+            search_radius = 1.05 * (max_point - min_point).norm_2()
+            mapper_settings.AddEmptyValue("search_settings")
+            mapper_settings["search_settings"]["search_radius"].SetDouble(search_radius)
```
philbucher, line 190: is this faster?
this seems larger than what I use by default so my guess is that it is slower

But ofc this is only a guess :)

philbucher (approved):

applications/ShallowWaterApplication/python_scripts/coupling/depth_integration_input_process.py
miguelmaso: Here I have a very specific configuration. In the 2D case I need to map from one node to all the BC, so I need a big search radius. The 3D case is more standard.

PR #9578, opened 2022-01-25
[MPI] enable AMGCL to use MPI_Comm other than MPI_COMM_WORLD

philbucher:
This PR enables the use of AMGCL with sub-MPI communicators
The other trilinos solvers already support this

Two open questions:
- I didn't find a better way of retrieving the `MPI_Comm` from the Trilinos Matrix. Does anyone know a better way? I read all the documentation, tutorials and their github issues but didn't find anything. I have not yet asked in their github as maybe @RiccardoRossi or @roigcarlo know the answer
- @ddemidov the `idrs` hardcodes the use of `MPI_COMM_WORLD`. Is this necessary? It also works without for my case. To me it seems like the rank is only used as seed for some random distribution. I checked your master branch, it is also hardcoded there.

Thx for your help :)

### DIFF
--- a/applications/TrilinosApplication/external_includes/amgcl_mpi_solver_impl.cpp
+++ b/applications/TrilinosApplication/external_includes/amgcl_mpi_solver_impl.cpp
@@ -22,6 +22,7 @@
 #include "Epetra_FECrsMatrix.h"
 #include "Epetra_FEVector.h"
 #include "trilinos_space.h"
+#include "custom_utilities/trilinos_solver_utilities.h"
 
 namespace Kratos
 {
@@ -45,10 +46,7 @@ void AMGCLScalarSolve(
     bool use_gpgpu
     )
 {
-    const Epetra_Comm& r_epetra_comm = rA.Comm();
-    const Epetra_MpiComm& r_epetra_mpi_comm = dynamic_cast<const Epetra_MpiComm&>(r_epetra_comm); // cannot use static_cast due to virtual inheritance
-
-    MPI_Comm the_comm = r_epetra_mpi_comm.Comm();
+    MPI_Comm the_comm = TrilinosSolverUtilities::GetMPICommFromEpetraComm(rA.Comm());
 
 #ifdef AMGCL_GPGPU
     if (use_gpgpu && vexcl_context()) {
@@ -112,10 +110,7 @@ void AMGCLBlockSolve(
     bool use_gpgpu
     )
 {
-    const Epetra_Comm& r_epetra_comm = rA.Comm();
-    const Epetra_MpiComm& r_epetra_mpi_comm = dynamic_cast<const Epetra_MpiComm&>(r_epetra_comm); // cannot use static_cast due to virtual inheritance
-
-    MPI_Comm the_comm = r_epetra_mpi_comm.Comm();
+    MPI_Comm the_comm = TrilinosSolverUtilities::GetMPICommFromEpetraComm(rA.Comm());
 
     if(amgclParams.get<std::string>("precond.class") != "amg")
         amgclParams.erase("precond.coarsening");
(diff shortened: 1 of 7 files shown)


ddemidov:
Thanks for pointing out the explicit use of `MPI_COMM_WORLD` in the IDRS solver. It is indeed used there as a repeatable random seed (so that consequent solves of the same problem would yield the same results). I think the IDRS solver needs the random vectors to be different on each MPI process, so simply using zero for pid is a bad idea. ddemidov/amgcl@b0c9618 should work better: it uses pid=0 for serial solves, and pid=comm.rank for parallel solves.

applications/TrilinosApplication/external_includes/amgcl_mpi_solver_impl.cpp
```diff
@@ -45,6 +45,11 @@ void AMGCLScalarSolve(
     bool use_gpgpu
     )
 {
+    const Epetra_Comm& r_epetra_comm = rA.Comm();
+    const Epetra_MpiComm& r_epetra_mpi_comm = dynamic_cast<const Epetra_MpiComm&>(r_epetra_comm); // cannot use static_cast due to virtual inheritance
+
+    MPI_Comm the_comm = r_epetra_mpi_comm.Comm();
```
ddemidov: Looks like `Epetra_Comm` inherits `Epetra_MpiComm`. If that is indeed the case, would not the following work?
```cpp
MPI_Comm the_comm = ra.Comm().Comm();
```

ddemidov: Sorry, it seems [the other way around](https://docs.trilinos.org/dev/packages/epetra/doc/html/classEpetra__Comm.html#details) (`Epetra_MpiComm` inherits `Epetra_Comm`).

philbucher: yes exactly thats why this doesn't work

In Kratos we have a fct for this:
https://github.com/KratosMultiphysics/Kratos/blob/b947b82c90dfcbf13d60511427f85990d36b90be/kratos/mpi/sources/mpi_data_communicator.cpp#L399-L410

Unfortunately I haven't found any equivalent in Trilinos

philbucher:
thx for the suggestion @ddemidov, I will use it!

@KratosMultiphysics/technical-committee should we apply this patch or update AMGCL altogether?

ddemidov:
> should we apply this patch or update AMGCL altogether?

Looks like the last update to amgcl has been made 13 months ago. The two potentially interesting (for Kratos) features implemented since then are:
1. partial rebuild of amgcl hierarchy (https://arxiv.org/abs/2108.02054), saves about 40% in setup time for non steady-state problems,
2. support for block arithmetic with aggregation AMG when null-space vectors or grid coordinates are provided (ddemidov/amgcl#215), saves about 40% in solve time.

rubenzorrilla:
> thx for the suggestion @ddemidov, I will use it!
> 
> @KratosMultiphysics/technical-committee should we apply this patch or update AMGCL altogether?

I'd update AMGCL altogether.

philbucher:
I now reverted the change in amgcl, as I think updating the library separately makes more sense

I now cleaned up the code a bit, from my side it is ready to be merged

ddemidov:
Will you update the amgcl code yourself, or should I make a PR?

philbucher:
If you could do it that would be great, thx!
You know best what needs to be updated

preferably in a separate PR
Thx!

ddemidov:
See #9589

ddemidov (approved):
This should probably be merged after #9589, since the use of `MPI_COMM_WORLD` in `amgcl/solvers/idrs.hpp` is fixed there.

applications/TrilinosApplication/external_includes/amgcl_mpi_solver_impl.cpp
philbucher: I asked in trilinos but no better answer:
https://github.com/trilinos/Trilinos/issues/10122

philbucher:
#9589 is merged now, hence also merging this :tada:

philbucher:
thx for the help @ddemidov !

PR #9579, opened 2022-01-25
[FSIApplication] Use parameter constructor to create discontinuous distance process

marcnunezc:
**📝 Description**
After #9567, this creates the discontinuous distance process using parameters instead of flags. 

Could we add a test that uses this or test this PR with a working example?

**🆕 Changelog**
- Use parameters to construct discontinuous distance process in partitioned embedded solver.


rubenzorrilla (approved):
Thanks! About the test... we could try to create a "fake" problem to only check the first step. I'll think about it.

marcnunezc:
No longer needed as it was added in #9881. Closing PR

loumalouomega:
Do I delete the branch?

marcnunezc:
> Do I delete the branch?

Sorry I forgot!

loumalouomega:
> > Do I delete the branch?
> 
> Sorry I forgot to delete it!

image

PR #9584, opened 2022-01-26
[Fluid] Parallelized gauss point loops in symbolic generator

EduardGomezEscandell:
**📝 Description**
I was motivated by Riccardo Rossi's class in the CIMNE Winter School to attempt to parallelize parts of the code in the symbolic generator. Particularly, the loops over the gauss points.

*Theoretical expectations*
The serial execution of `generate_triangle.py` takes 328 seconds*. The parallelizable parts take 174 seconds. If they were perfectly parallelized into three threads (one for each gauss point), the total execution time would come down to 212s, therefore an improvement ceiling of 35%.

\*These benchmarks are with profiling enabled, without profiling it takes 148 seconds.

*Results*
They're bad. It runs 20% slower on my old laptop. I will run a benchmark on my PC and see if it is also bad or if it improves. If it improves, the CI run will be the tiebreaker. If it doesn't improve, this branch will be removed.

I ran the benchmark with `htop` on and when run in serial, one of the cores is at 100% and the others idle.
When running in parallel, all four cores went to about 20-30%. If anyone can offer their expertise as to why this may happen I'll be glad to listen.

**🆕 Changelog**
- Loops over gauss points in symbolic generation are now parallel.

### DIFF
--- a/applications/FluidDynamicsApplication/python_scripts/symbolic_generation/compressible_navier_stokes/compressible_navier_stokes_symbolic_generator.py
+++ b/applications/FluidDynamicsApplication/python_scripts/symbolic_generation/compressible_navier_stokes/compressible_navier_stokes_symbolic_generator.py
@@ -2,6 +2,8 @@ import KratosMultiphysics
 import sympy
 import KratosMultiphysics.sympy_fe_utilities as KratosSympy
 
+import concurrent.futures as Futures
+
 from KratosMultiphysics.FluidDynamicsApplication.symbolic_generation.compressible_navier_stokes.src import generate_convective_flux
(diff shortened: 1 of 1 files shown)


RiccardoRossi (commented):
your problem is the GIL which is used heavily here

try to substitute
Futures.ThreadPoolExecutor(max_workers=ngauss)

by the ProcessPoolExecutor

RiccardoRossi:
> ProcessPoolExecutor

also note that the number of workers depends on the hardware...

EduardGomezEscandell:
Seems that this didn't fix it.

EduardGomezEscandell:
Closing for now since this was more of an experiment. More investigation is needed.

PR #9586, opened 2022-01-27
[Core] fix compilation with older versions of boost

philbucher:
following #9541

<s>@rubenzorrilla said that it didn't work with 1.61, @AlejandroCornejo said that it works with 1.69
I didn't find anything in the changelogs in the versions between, hence going with 1.62 for now

If we get more reports then we can update this number</s>

**EDIT:** I checked the source code of the boost releases, the fix is only included from 1.70. Hence now I am only enabling the feature in question if the version of boost is new enough, otherwise a warning is issues. 
This way it is fully backward compatible

FYI @mpentek

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -375,7 +375,7 @@ message(STATUS "Boost Include: ${Boost_INCLUDE_DIRS}")
 message(STATUS "Boost Linkdir: ${Boost_LIBRARY_DIRS}")
 
 if(Boost_VERSION_STRING VERSION_LESS 1.70)
-  message(STATUS "WARNING: Kratos requires at least boost version 1.70 to enable bost related performance improvements")
+  message(STATUS "WARNING: Kratos requires at least boost version 1.70 to enable boost related performance improvements")
 else()
   add_definitions( -DBOOST_UBLAS_MOVE_SEMANTICS )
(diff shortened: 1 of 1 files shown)


rubenzorrilla:
I required to update 1.62 this morning, so I can confirm 1.62 works neither (I'm late I think... :sweat:)

loumalouomega:
> I required to update 1.62 this morning, so I can confirm 1.62 works neither (I'm late I think... sweat)

When I started using Kratos I used 1.57 👴

philbucher:
> > I checked the source codes, the change only made it to the 1.70 release!!!
> > Since the boost in the CentOS container is super old (1.53 I think) I would go with the backward compatible solution, i.e. enable the move semantics only if the version of boost is new enough and issue a warning otherwise
> > @KratosMultiphysics/altair unless you manually install a new boost, I guess your compilation is also broken ...
> 
> Could it be that it works with some compilers? I checked the source code of 1.69 and shouldn't work. However it works fine for @AlejandroCornejo, who AFAIK mainly works in Windows.

no idea, I was asking myself the same
But at this point I would choose the safer path and go with 1.70 to enable the move semantics.

rubenzorrilla (approved):

PR #9588, opened 2022-01-31
[FastPR][Core] Missing const in ModelPart Has methods

rubenzorrilla:
**📝 Description**
`const` is missing in `ModelPart`'s `HasNode`, `HasElement` and `HasCondition` methods.


philbucher (approved):
👍

PR #9589, opened 2022-01-31
[core] Update AMGCL to the latest version

ddemidov:
**📝 Description**

Updates the AMGCL library to the latest upstream version. See #9578 for details.

**🆕 Changelog**

Fixes direct use of `MPI_COMM_WORLD` in `external_libraries/amgcl/solvers/idrs.hpp`.

The core AMGCL functionality should stay the same. New features that could be potentially useful in Kratos:

1. partial rebuild of amgcl hierarchy (https://arxiv.org/abs/2108.02054), saves about 40% in setup time for non steady-state problems,
2. support for block arithmetic with aggregation AMG when null-space vectors or grid coordinates are provided (ddemidov/amgcl#215), saves about 40% in solve time.

The full list of upstream changes: https://github.com/ddemidov/amgcl/compare/b4ab723...ec2ea4b


ddiezrod:
Its being quite a long time since our last merge from github master, so its not easy to test this right, so from @KratosMultiphysics/altair  side, I'd say that if all tests pass correctly here you can go ahead.

RiccardoRossi (approved):
all test pass and ...a few bugs were corrected within the library.
approving since anyway it would be possible to revert this since it is only touching one folder

PR #9267, opened 2021-10-29
[AnalysisStage] Initial time

miguelmaso:
**📝 Description**
Currently, the initial time is not set in the ProcessInfo. If the initial time is different from 0, the first time step may be computed wrong.
For example, if the initial time is 2, and the time step is 0.1, `ModelPart.CloneTimeStep(2+0.1)` will set a time step of 2.1, instead of 0.1

**🆕 Changelog**
- AnalysisStage: Set the initial time
- MeshMovingApp: update some test (start at time different from 0)
- Core: add a test (this test can't be hacked, since the reference is computed on the fly)

PR #9592, opened 2022-02-01
[Fluid] Standarize buffer filling

rubenzorrilla:
**📝 Description**
This PR (finally) makes the buffer filling of the `FluidDynamicsApplication` solvers properly, that is to say _à la_ `StructuralMechanicsApplication`. As you note, what we basically do is to go backwards such that we do enough "fake" `CloneSolutionStep` operations instead of dropping the initial steps until the buffer is filled as we've done so far (see the function `_SetAndFillBuffer` in the `fluid_solver.py`). Thanks to this we're starting the simulations at time step one, which is what the user expects. On top of this, this makes possible the usage of buffer > 2 time schemes in the CFD ROM.

(as we're solving an "extra" time step, this required to slightly modify our current test values...)

PD: `initialize_with_compressible_potential_flow_process_test`is the unique test failing. @EduardGomezEscandell will take care of it. Creating the PR to start triggering the CI.

PD2: @KratosMultiphysics/pfem2 and @KratosMultiphysics/shallow-water also use the dropping step(s) bad practice. I think it will be positive to do what we're doing in here.


philbucher:
hm this will probably also change many other test results
for sure the ones in CoSim and MeshMoving (=> ale-fluid solver)

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_solver.py
```diff
@@ -254,75 +254,73 @@ def InitializeSolutionStep(self):
...
+
+        # Prepare distance correction for next step
+        self._GetDistanceModificationProcess().ExecuteFinalizeSolutionStep()
+
+        # Finalize the solver current step
+        self._GetSolutionStrategy().FinalizeSolutionStep()
+        # Limit the obtained acceleration for the next step
+        # This limitation should be called on the second solution step onwards (e.g. STEP=3 for BDF2)
+        # We intentionally avoid correcting the acceleration in the first resolution step as this might cause problems with zero initial conditions
+        if self._apply_acceleration_limitation and self.main_model_part.ProcessInfo[KratosMultiphysics.STEP] >= self.min_buffer_size:
```
ddiezrod: I think this should be now
`if self._apply_acceleration_limitation and self.main_model_part.ProcessInfo[KratosMultiphysics.STEP] > 1:`

right? So it is only applied after the first time step is fully solved.

rubenzorrilla:
> hm this will probably also change many other test results for sure the ones in CoSim and MeshMoving (=> ale-fluid solver)

Potentially yes.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_solver.py
rubenzorrilla: You're right. Thanks for spotting!

rubenzorrilla: Done!

dcagritan:
Hello Rubén,
I'll deal with the PFEM2Application over the weekend. Thanks for your feedback.
Regards

miguelmaso:
Be careful. Some time ago I tried the shame in the ShallowWaterApp, but I faced some problems with the initial time. I think you're reading the initial time at the `PrepareModelPart` but the time is not set. #9267 is related to that

dcagritan:
> t t

Great! Thank you @miguelmaso!
Regards

applications/FluidDynamicsApplication/python_scripts/fluid_solver.py
```diff
@@ -223,6 +212,27 @@ def _ExecuteCheckAndPrepare(self):
...
+        required_buffer_size = self.GetMinimumBufferSize()
+        current_buffer_size = self.main_model_part.GetBufferSize()
+        buffer_size = max(current_buffer_size, required_buffer_size)
+        self.main_model_part.SetBufferSize(buffer_size)
+
+        # Cycle the buffer. This sets all historical nodal solution step data to
+        # the current value and initializes the time stepping in the process info.
+        step = -buffer_size
+        delta_time = self._ComputeInitialDeltaTime()
+        init_time = self.main_model_part.ProcessInfo[KratosMultiphysics.TIME]
```
miguelmaso: I think the initial time is not yet set and will return always 0. Convergence problems may appear if the initial time is different than 0

philbucher: if the implementation of @rubenzorrilla is done exactly in the same way as n the structrual app then it works fine.
The solution there has existed since forever and has always worked afaik

applications/FluidDynamicsApplication/python_scripts/fluid_solver.py
```diff
@@ -223,6 +212,27 @@ def _ExecuteCheckAndPrepare(self):
 
         check_and_prepare_model_process_fluid.CheckAndPrepareModelProcess(self.main_model_part, prepare_model_part_settings).Execute()
 
+    def _SetAndFillBuffer(self):
```
philbucher, line 217: @rubenzorrilla if you want to make it consistent with the structural app then I suggest to refactor this fct from there to the [auxiliary_solver_utilities](https://github.com/KratosMultiphysics/Kratos/blob/master/kratos/python_scripts/auxiliary_solver_utilities.py) and then use it from both apps

rubenzorrilla: Great.

philbucher: I checked, the impl is mostly the same
In this case I think it would be a good idea to refactor the fct first in a separate PR (an making it as flexible as you need, e.g. with regards to passing dT or the buffer sizes)

rubenzorrilla:
@miguelmaso @dcagritan My comment was only for letting you know. Non-core application changes must be done in a different PR.

dcagritan:
@rubenzorrilla, yes, I'm aware of that. Thanks for letting me know!
Good day

applications/FluidDynamicsApplication/python_scripts/fluid_solver.py
miguelmaso: have you checked with a different initial time? (especially with explicit strategies)

philbucher: In restarts the initial time is different from 0
Also the same is used for the explicit structural simulations

or what exactly do you mean?

miguelmaso: if it works with a non restarted simulation with initial time different from 0:
```json
    "problem_data"     : {
        "start_time"    : 100.0,
        "end_time"      : 100.1,
        ...
```

philbucher: yes this is what we always do for restarts

applications/FluidDynamicsApplication/python_scripts/fluid_solver.py
rubenzorrilla: Makes much sense. I'll proceed as suggested.

applications/FluidDynamicsApplication/python_scripts/fluid_solver.py
philbucher: actually I confused it, sorry my bad
In a restart this fct is not called at all, then the current time is gotten from the restart file
https://github.com/KratosMultiphysics/Kratos/blob/1a8bc639d77a054eb7f15c1f122d1ef7135e8fb4/kratos/python_scripts/analysis_stage.py#L104-L105

actually looking at your PR #9267 I am not sure what will happen with a `start_time` unequal to zero :/
in this case the comment from above is correct

rubenzorrilla: > In restarts the initial time is different from 0 Also the same is used for the explicit structural simulations
> 
> or what exactly do you mean?

I can confirm this is the current behavior.

applications/FluidDynamicsApplication/python_scripts/fluid_solver.py
rubenzorrilla: Done. See #9594.

applications/FluidDynamicsApplication/python_scripts/fluid_solver.py
```diff
@@ -95,8 +96,8 @@ def PrepareModelPart(self):
             self._ReplaceElementsAndConditions()
             ## Executes the check and prepare model process
             self._ExecuteCheckAndPrepare()
-            ## Set buffer size
-            self.main_model_part.SetBufferSize(self.min_buffer_size)
+            ## Set and fill buffer
+            self._SetAndFillBuffer()
```
miguelmaso, line 98: I have a doubt. The initial condition should be impsed after setting the buffer size and before filling the buffer. It is not relevant if the buffer size is 2, but with higher values there will be uninitialized steps (I have a scheme with buffer size 4)
Am I wrong?

rubenzorrilla: mmm... I think I don't get your point at all. Either if you impose your initial condition by using a custom process that fills the entire buffer or by customizing the `ModifyAfterSolverInitialize` in your `MainKratos.py` the `_SetAndFillBuffer` would have been already called.

rubenzorrilla:
Issues with Swimming DEM application are handled in #10179.
About the fluid adjoint analysis stuff, we're (@sunethwarna) are discussing the best way to get rid of the `nsteps` in `problem_data` section and the negative time step.

sunethwarna:
I will get rid of the `nsteps` in the ajoints and will have usual `start_time` and `end_time` in `problem_data`. We don't have a problem with the `problem_data` section and the negative time step ryt? It is set at the `solver_settings` part.

rubenzorrilla:
> I will get rid of the `nsteps` in the ajoints and will have usual `start_time` and `end_time` in `problem_data`. We don't have a problem with the `problem_data` section and the negative time step ryt? It is set at the `solver_settings` part.

As we discussed yesterday, I'd avoid a negative time step, as it is in my opinion a bit counterintuitive for the user (note that current CI error comes indeed from a check that ensures exactly this). Instead I'd modify the `_ComputeDeltaTime` in the `AdjointFluidSolver` to put the minus under the hood, so the user (or the GUI) doesn't need to do anything special.

rubenzorrilla:
> > I will get rid of the `nsteps` in the ajoints and will have usual `start_time` and `end_time` in `problem_data`. We don't have a problem with the `problem_data` section and the negative time step ryt? It is set at the `solver_settings` part.
> 
> As we discussed yesterday, I'd avoid a negative time step, as it is in my opinion a bit counterintuitive for the user (note that current CI error comes indeed from a check that ensures exactly this). Instead I'd modify the `_ComputeDeltaTime` in the `AdjointFluidSolver` to put the minus under the hood, so the user (or the GUI) doesn't need to do anything special.

@sunethwarna any advance in this regard. I don't mind taking care of the suggested changes.

rubenzorrilla:
Required changes in the adjoint were done in #10312.

philbucher:
@rubenzorrilla did you run all the tests in the CoSimApp?
I am afraid this can affect it too

FYI  @mpentek this might affect some results

rubenzorrilla:
> @rubenzorrilla did you run all the tests in the CoSimApp? I am afraid this can affect it too
> 
> FYI @mpentek this might affect some results

Aren't these run by the CI?

philbucher:
only the very small FSI tests are run in the CI
However since coupled sims usually run longer they are part of the validation suite

Can you hence please run all tests?
I thought for bigger changes such as this it was the responsibility of the dev to run the tests and not only rely on the CI?

rubenzorrilla:
> only the very small FSI tests are run in the CI However since coupled sims usually run longer they are part of the validation suite
> 
> Can you hence please run all tests? I thought for bigger changes such as this it was the responsibility of the dev to run the tests and not only rely on the CI?

OK. No problem, I'll run them again. Indeed, I think I ran all of them when amending the SDOF test  results.

rubenzorrilla:
> > only the very small FSI tests are run in the CI However since coupled sims usually run longer they are part of the validation suite
> > Can you hence please run all tests? I thought for bigger changes such as this it was the responsibility of the dev to run the tests and not only rely on the CI?
> 
> OK. No problem, I'll run them again. Indeed, I think I ran all of them when amending the SDOF test results.

Seems that all the tests are OK. :+1:

applications/FluidDynamicsApplication/python_scripts/fluid_solver.py
```diff
@@ -93,8 +94,10 @@ def PrepareModelPart(self):
...
             self._ReplaceElementsAndConditions()
-            ## Set buffer size
-            self.main_model_part.SetBufferSize(self.min_buffer_size)
+            ## Executes the check and prepare model process
+            self._ExecuteCheckAndPrepare()
+            ## Set and fill buffer
+            self._SetAndFillBuffer()
 
         ## Executes the check and prepare model process. Always executed as it also assigns neighbors which are not saved in a restart
         self._ExecuteCheckAndPrepare()
```
sunethwarna, line 101: We cab call `_ExecuteCheckAndPrepare` here and avoid calling it in line 98 ryt?

applications/FluidDynamicsApplication/python_scripts/fluid_solver.py
sunethwarna: This can be further extended for adjoint buffer filling as well. I'll take care of it. :)

sunethwarna (dismissed):
I am ok with the changes except for two converns.
1. Why is it required to call `ExecuteCheckAndPrepareMethod` twice?
2. I know there has to be reference value changes. I hope you have checked those ref value changes to make sure they only have a time shift :D :)

This is a cool addition I would extend to transient adjoints as well :) 

Thanks @rubenzorrilla  👍

applications/FluidDynamicsApplication/python_scripts/fluid_solver.py
rubenzorrilla: Yes. Indeed I think that this accidentally sneaked in. Nice catch.

sunethwarna (approved):
Nice :+1:

rubenzorrilla:
Merging. Just to be sure again, I ran all the fluid-related tests (conv-diff, potential flow, FSI, co-sim, rans, etc...).

PR #9593, opened 2022-02-02
[Fluid] Fixed errors in compressible N-S quadrilateral element

EduardGomezEscandell:
**📝 Description**
The quadrilateral element for compressible NS was straight-up wrong:
- The mass matrix was that of the triangle
- The integration assumed that the elements were are always perfect squares

**🆕 Changelog**
- Fixed mass matrix computation (in template)
- Fixed integration (in template and in generator)
- Fixed `CalculateMindpoint*` methods


applications/FluidDynamicsApplication/custom_elements/compressible_navier_stokes_explicit_2D3N.cpp
```diff
@@ -41,6 +41,12 @@
 
 namespace Kratos {
 
+template<>
+GeometryData::IntegrationMethod CompressibleNavierStokesExplicit<2,3>::GetIntegrationMethod() const
+{
+    return GeometryData::IntegrationMethod::GI_GAUSS_1;
```
rubenzorrilla, line 47: Correct me if I'm wrong, but isn't this element the one with "pre-computed" integration in the automatic differentiation? If so it is, I'd throw an error in here, as this method is expected to be never called.

rubenzorrilla: OK. Now I see, you use it for the auxiliary values. Fine then.

rubenzorrilla: Maybe we can put a comment saying that this is not used in the LHS and RHS evaluation.

applications/FluidDynamicsApplication/python_scripts/symbolic_generation/compressible_navier_stokes/templates/compressible_navier_stokes_explicit_template_2D4N.cpp
```diff
@@ -81,36 +87,62 @@ void CompressibleNavierStokesExplicit<2,4>::GetDofList(
...
+
+    Matrix DN_De;
+
+    Matrix J;
+    Matrix Jinv;
+
+    const auto center = rGeometry.Center();
+    rGeometry.ShapeFunctionsValues(rN, center);
+    rGeometry.Jacobian(J, center);
+    MathUtils<double>::InvertMatrix(J, Jinv, detJ);
```
rubenzorrilla: I think you can use `InverseJacobian` in here.

applications/FluidDynamicsApplication/python_scripts/symbolic_generation/compressible_navier_stokes/templates/compressible_navier_stokes_explicit_template_2D4N.cpp
```diff
@@ -184,27 +222,36 @@ void CompressibleNavierStokesExplicit<2,4>::CalculateMomentumProjection(const Pr
...
     {
+        const double w = gauss_point.Weight();
         r_geometry.ShapeFunctionsValues(N, gauss_point.Coordinates());
-        r_geometry.InverseOfJacobian(Jinv, gauss_point.Coordinates());
-        r_geometry.ShapeFunctionsLocalGradients(DN_DX_iso, gauss_point.Coordinates());
-        GeometryUtils::ShapeFunctionsGradients(DN_DX_iso, Jinv, DN_DX);
+
+        double detJ;
+        r_geometry.Jacobian(J, gauss_point.Coordinates());
+        MathUtils<double>::InvertMatrix(J, Jinv, detJ);
```
rubenzorrilla, line 239: Same `InverseJacobian` comment.

applications/FluidDynamicsApplication/custom_elements/compressible_navier_stokes_explicit_2D3N.cpp
EduardGomezEscandell: It is not used during integration but it is used in `CalculateMidPoint*` methods.

Before, we simply called `GI_GAUSS_1` directly. I moved it to this method to make it coherent with other elements.

I see the implementation in these midpoint methods assumes the shape function gradients to be uniform, so I guess there is some danger here since this mehtod gives the false impression of allowing change of integration method.

I can revert to simply using `GI_GAUSS_1` in place or remove the assumption for middle point computation (just like the quad element, so it would take no effort to change). Shouldn't really affect performance since we only need the shape function at the centerpoint.

rubenzorrilla (commented):
Fine about the simplex element changes. About the square, I think we can directly use the `InverseJacobian` function in many places.

applications/FluidDynamicsApplication/custom_elements/compressible_navier_stokes_explicit_2D3N.cpp
EduardGomezEscandell: Oh ok you noticed before I finished writing 😅

rubenzorrilla: No no, I'm fine with this. However, I'd put a comment in line 47 saying that this is only use for the postprocessing of auxiliary variables, no in the LHS and RHS calculation (just to avoid future misunderstandings).

applications/FluidDynamicsApplication/python_scripts/symbolic_generation/compressible_navier_stokes/templates/compressible_navier_stokes_explicit_template_2D4N.cpp
EduardGomezEscandell: In this place in particular I need both the inverse and determinant, and `InverseOfJacobian` only yields the former.

rubenzorrilla (approved):

Issue #9258, opened 2021-10-26
internal discussion about multistage


RiccardoRossi:
**Description**
Ongoing discussion about multistage.

this are just working notes of the internal discussion within the @KratosMultiphysics/technical-committee 

~~~json
{
    "stage1" : {
        "stage_initialize" : { 
            "modelers" : [{
                "modeler_name": "read_model_part_from_mdpa",
                "Parameters" : {...}
                },
                {
                    "modeler_name": "mapping_from_previous_stage",
                    "Parameters" : {...}
                },
            ],
            "stage_preprocess"            : [ { ... } ], //OPERATION(only implementing "Execute") to be called AFTER the modellers
        },
        "stage_finalize" : { 
            "stage_finalizer_list" : [
                {
                    "finalizer_name" : "...",
                    "Parameters" : {...}
                },
                {
                    "finalizer_name" : "my_custom_process",
                    "Parameters" : {...}
                }
            ],
            "stage_postprocess"           : [ { ... } ], //OPERATION(only implementing "Execute")
        },
        "problem_data"     : { ... },
        "solver_settings"  : { ... },
        "output_processes" : [ 
            "constraints_process_list"    : [ { ... } ],
            "loads_process_list"          : [ { ... } ],
            "list_other_processes"        : [ { ... } ],
        ]
    }
}
~~~

RiccardoRossi:
next iteration

~~~json
//things that live through an analysis:
Model
Project (yet to be added)
serialized_streams/files ????

if restart
  skip preprocess

{
    "stage1" : {
        "depends_on" : "nothing/previous/...", ??????????????? to be discussed in the next meeting (proposal by Riccardo if we want to have stages as tasks NOT DISCUSSED) 
        
        "stage_preprocess" : { 
            "model_restart" : {
                "read_restart" : false,
                "restart_options"                  : {
                    "SaveRestart"      : "False",
                    "RestartFrequency" : 0,
                    "LoadRestart"      : "False",
                    "Restart_Step"     : 0
                },
                //restart is mutually exclusive with the use of modelers
            },
            "modelers" : [
                {
                "modeler_name": "read_model_part_from_mdpa",
                "Parameters" : {...}
                },
                {
                    "modeler_name": "mapping_from_previous_stage",
                    "Parameters" : {...}
                },
            ],
            "stage_initializer_list"  : [ 
                {
                    "operation_name" : "...",
                    "Parameters" : {...}
                },
                {
[...]

PR #7108, opened 2020-06-18
[Core][Model/ModelPart] recursive SubModelParts

philbucher:
:warning:  **PLEASE READ THIS BEFORE BLOCKING** :warning: 
I am really hesitant to change both `Model` and `ModelPart`, but after implementing those functionalities for the 20th time in my own scripts I felt like it is time to do it properly

In a nutshell with this PR it becomes possible to use inputs such as `Inlet.SubModelPartName` for `Create(Sub)ModelPart`, `Get(Sub)ModelPart` and `Has(Sub)ModelPart`

This is very handy when dealing with more complex `ModelPart` hierarchies as it is necessary for larger models

PR #9598, opened 2022-02-03
[All][Core][Model] CreateModelPart new behavior

rubenzorrilla:
**📝 Description**
As it has been discussed in #9258, the creation of model parts must happen at the modelers stage in order to enable the multistage simulation (not at the solver instantiation as we're doing right now).

Changing the `CreateModelPart` method of the `Model` as suggested in here will ease the transition period. Long story short, right now we were throwing an error if the model part was already existing. With this change we will issue a warning and return the already existing model part.

Backwards compatibility is ensured as we were throwing an error so far.


philbucher:
does this interfere or interact in any other way with #7108?

rubenzorrilla:
> does this interfere or interact in any other way with #7108?

As far as I understand it shouldn't.

RiccardoRossi (approved):
approving on behalf of @KratosMultiphysics/technical-committee

PR #8849, opened 2021-06-03
[not to merge] Mdpa modeler

miguelmaso:
**Description**
Recently I've discovered the modelers. I think they are a nice feature and a good alternative to the `PythonSolver.ImportModelPart` because of its modularity.

If I had known the modelers existed, I would have implemented #8608 as a modeler in conjunction with this. Maybe it is interesting for special needs in the import stage, like #8806

This is to start a discussion about the import stage. Is it the intended usage of the modelers?

**Changelog**
- Add mdpa io modeler

PR #9599, opened 2022-02-03
[Core] Import model part modeler

rubenzorrilla:
**📝 Description**
Creating a simple modeler to do the input file parsing at the `AnalysisStage` level as suggested in #9258 in order to enable multistage simulations. It is important to mention that **the creation of root model parts occurs at the modeler instantiation**. This will make possible to immediately fill the nodal variable data by calling the corresponding solver `AddVariables` method in the `AnalysisStage` to then read the mesh (`SetupModelPart`) (see [here](https://github.com/KratosMultiphysics/Kratos/issues/9258#issuecomment-1028206989)).

Unit test has been added accordingly. I tried it with a real heat transfer problem and works with the current solvers (throwing the warning introduced in #9598 as expected).

Note that I've also changed the defaults of the base Python solver to ease the transition so our current `_ImportModelPart` method in `python_solver.py` and the modeler-based import can coexist for a while.

Having said this, I think we should start populating the use of this modeler in our GUIs / examples and tests ASAP.

Requires #9598 to be merged first.

EDIT: This is deliberately done in Python in order to extend it to the distributed import case in the future.


miguelmaso:
closes #8849

kratos/python_scripts/modelers/import_model_part_modeler.py
```diff
@@ -0,0 +1,50 @@
...
+        KratosMultiphysics.SingleImportModelPart.Import(
+            self.model_part,
+            self.settings,
+            self.settings["input_type"].GetString())
+
+    def __GetDefaultSettings(self):
+        default_settings = KratosMultiphysics.Parameters('''{
+            "echo_level" : 0,
+            "input_type" : "mdpa",
+            "input_filename" : "",
```
pooyan-dadvand: It would be great if we put the extension in the filename to avoid one of our historical problems.

kratos/python_scripts/modelers/import_model_part_modeler.py
```diff
@@ -0,0 +1,50 @@
+import KratosMultiphysics
+
+class ImportModelPartModeler(KratosMultiphysics.Modeler):
```
pooyan-dadvand: Please change it to ImportMDPAModeler to have it by the type.

kratos/python_scripts/modelers/import_model_part_modeler.py
```diff
@@ -0,0 +1,50 @@
...
+        # Note that at this point solvers must have added the variables to the nodal variable data
+        KratosMultiphysics.SingleImportModelPart.Import(
+            self.model_part,
+            self.settings,
+            self.settings["input_type"].GetString())
+
+    def __GetDefaultSettings(self):
+        default_settings = KratosMultiphysics.Parameters('''{
+            "echo_level" : 0,
+            "input_type" : "mdpa",
```
pooyan-dadvand: If this is the mdpa modeler then the type is always mdpa. so we can eliminate this field.

kratos/python_scripts/modelers/import_model_part_modeler.py
rubenzorrilla: Done.

kratos/python_scripts/modelers/import_model_part_modeler.py
rubenzorrilla: Resolved below.

kratos/python_scripts/modelers/import_model_part_modeler.py
rubenzorrilla: Done.

kratos/tests/test_import_model_part_modeler.py
```diff
@@ -0,0 +1,193 @@
...
+        KratosUtilities.DeleteFileIfExisting(GetFilePath("test_model_part_io_write.time"))
+        KratosUtilities.DeleteFileIfExisting(GetFilePath("test_model_part_io_write_mesh_only.out.mdpa"))
+        KratosUtilities.DeleteFileIfExisting(GetFilePath("test_model_part_io_write_mesh_only.out.time"))
+        KratosUtilities.DeleteFileIfExisting(GetFilePath("test_model_part_io_write_mesh_only.time"))
+
+    def testImportMDPAModeler(self):
+        # Set up the import model part modeler
+        model = KratosMultiphysics.Model()
+        settings = KratosMultiphysics.Parameters('''{
+            "input_type" : "mdpa",
```
roigcarlo: This should be part of the name or "use_input_model_part":

```Json
{
    "input_filename" : "",
    "model_part_name" : "Main.mdpa"
}
```

Like that no?

rubenzorrilla: Sorry, I forgot updating the test. ~~I think we agreed to use `Main` as `mdpa` is assumed.~~ `Main` is the model part name where the mdpa is to be stored, not the file to be parsed. Hence, only the `input_type` file has to be removed in these settings.

roigcarlo (commented):
Seems that CI is passing so :+1:

roigcarlo (dismissed):

roigcarlo (approved):

PR #9601, opened 2022-02-03
[Core][AMGCL] solving a memory bug

RiccardoRossi:
@ddemidov could u take a look if the fix makes sense to you

**📝 Description**
the vector where coordinates are stored is not allocated correctly

bugfix

**🆕 Changelog**


ddemidov:
I think the problem here is that B is defined inside the if block, its address is put into amgcl params, and the B is destroyed on the scope exit.

ddemidov (approved):

PR #9604, opened 2022-02-04
[MPI][GidOutput] use DataCommunicator from ModelPart

philbucher:
see title


rubenzorrilla (approved):
Makes more sense. Thanks!

PR #9608, opened 2022-02-06
[FastPR][Fluid] Computing model part creation

rubenzorrilla:
**📝 Description**
This makes optional the creation of the computing model part. This is required for the HROM postprocess stage.


philbucher:
Finally!!!
Can you please make that optional with a flag like we had in the structural?

rubenzorrilla:
> Finally!!! Can you please make that optional with a flag like we had in the structural?

Well, it is still strictly required. The point is that current implementation crashes if the computing model part is already in the mdpa (as it happens with the HROM mesh).

philbucher (approved):
actually I can directly use this in #9614 :D

PR #9610, opened 2022-02-06
[Core] adding some try catch

philbucher:
see title, nice to trace errors :)


roigcarlo (approved):
:+1: I personally find it easy to trace errors without this, but since we use it everywhere...

roigcarlo:
gdb

philbucher:
ah ok
sure for more detailed debugging I also use gdb, but still I like having already a clue where the error comes from
Doesn't harm I guess

PR #9531, opened 2022-01-11
[Core][MPI] fix Serializer flags in  distributed_restart_utility

philbucher:
This PR set correctly the flags for the `Serializer` in the `distributed_restart_utility` when used in MPI

closes #9468

@mpentek can you test this please?

PR #9614, opened 2022-02-06
[Fluid] always execute check process to assign NEIGHBOR_ELEMENTS

philbucher:
This PR proposes to always execute the check process, **independent if doing a restart or not**
~~(the computing modelpart is only created if it is not a restart!)~~ depends on #9608

The reasoning is the following:
1. #9468 discovered that the MPITwoFluidNS-Solver is not working in a restart.
    Reason is that it computes the `NEIGHBOR_NODES` which are `GlobalPointer`s. Since some nodes are remote, the serialization can only be done shallow
2. this motivated #9531 to only do a shallow serialization for restarts.
3. Unfortunately also the `NEIGHBOR_ELEMENTS` are `GlobalPointer`s which means that those are no longer saved in restart files. This however is used for most of the Slip boundary conditions
4. Hence I propose this change to recompute the `NEIGHBOR_ELEMENTS` all the time. Note that previously the restart was only working because the conditions are always in the same partition, hence the `GlobalPointer`s never had to serialize a remote object
https://github.com/KratosMultiphysics/Kratos/blob/7f594a0fc6667c39475c98375330143fcb7fd1ee/kratos/mpi/python_scripts/distributed_import_model_part_utility.py#L68-L69

~~(Most of the changes are from indenting the code one level bcs of the `IS_RESTARTED` check)~~

FYI @mpentek


philbucher:
this is ready to be merged

rubenzorrilla (approved):
Makes sense as well. Thanks!

PR #9616, opened 2022-02-07
[Compressible Potential] Updated GetNumThreads->ParallelUtilities

EduardGomezEscandell:
**📝 Description**
Small maintenance change to remove deprecation warning during compilation.

**🆕 Changelog**
- Call to `OpenMPUtils::GetNumThreads()` is now `ParallelUtils::GetNumThreads()`


marcnunezc (approved):
Thanks!

marcnunezc:
Btw, maybe the include of `openmp_utils.h` can now be removed, or at least replaced with `parallel_utilities.h`.

philbucher:
this loop would need to be refactored to use the ParallelUtils before the include can be removed

EduardGomezEscandell:
Yeah, I did the minimum to remove the warning, but the various pragmas should be replaced with `block_for_each`. Still, I think that can be done in a future PR.

PR #9618, opened 2022-02-07
[Core] cleanup ConDiff settings

philbucher:
- `default`ing the CTor (vars are properly initialized when defining them)
- adding missing `const`
- moving include to cpp


rubenzorrilla (approved):
Thanks for the clean up! :100:

PR #9619, opened 2022-02-07
[Core][ConDiffSettings] directly checking the variable

philbucher:
my plan is to get rid of the auxiliary booleans, they are not necessary (more PRs coming, doing it in small steps to ease the review and avoid mistakes)


rubenzorrilla (approved):
Makes sense. Thanks again.

PR #9626, opened 2022-02-08
Adding flag to optionally enable C++17 support

philbucher:
see title

C++17 support can now be enabled with the CMake flag `KRATOS_USE_CXX_17`

this also defines `KRATOS_USE_CXX_17` as macro in the code

**changelog**
- Adding flag `KRATOS_USE_CXX_17`
- Enabling C++17 mode in compilers
- Updating the CI (where already for a while one of the build uses C++17)


philbucher:
BTW I tested it with Gcc, clang and intel

In a followup PR I would add configure errors to specify the least version of a compiler that should be used
Same as what I did for the intel compiler already
https://github.com/KratosMultiphysics/Kratos/blob/7bba0aed48491df8b2267bcb73dd0f9fb3aebe2f/CMakeLists.txt#L200-L203

philbucher:
sorry for the noise, in my last commit I also fixed the indentation, please disable whitespace changes to see the real changes
sorry for the inconvenience

RiccardoRossi:
just to give my support to this solution. I'll add this to the next TC meeting

CMakeLists.txt
```diff
@@ -165,7 +170,12 @@ endif( ${CMAKE_BUILD_TYPE} MATCHES "Release" )
 
 # Set compiler flags
 if(${CMAKE_COMPILER_IS_GNUCXX})
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -funroll-loops -Wall -std=c++11 -Wl,--no-as-needed -ldl")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -funroll-loops -Wall -Wl,--no-as-needed -ldl")
+  if(KRATOS_USE_CXX_17 MATCHES ON)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")
```
loumalouomega:
```suggestion
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")
    
    # Check for standard to use
CHECK_CXX_COMPILER_FLAG(-std=c++17 HAVE_FLAG_STD_CXX17)
IF(HAVE_FLAG_STD_CXX17)
    # Have -std=c++17, use it
    MESSAGE(STATUS "HAVE_FLAG_STD_CXX17 available")
ELSE(HAVE_FLAG_STD_CXX17)
    CHECK_CXX_COMPILER_FLAG(-std=c++1z HAVE_FLAG_STD_CXX1Z)
    IF(HAVE_FLAG_STD_CXX1Z)
        # Have -std=c++1z, use it
        MESSAGE(STATUS "HAVE_FLAG_STD_CXX1Z available")
    ELSE(HAVE_FLAG_STD_CXX1Z)
        # And so on and on...
        MESSAGE(FATAL_ERROR "No modern C++17 found, compilation may fail")
    ENDIF(HAVE_FLAG_STD_CXX1Z)
ENDIF(HAVE_FLAG_STD_CXX17)
```

With this check you can verify the compiler actually supports the standard

philbucher: this is what I would like to add in followup PRs

this PR is simply about adding the flag, easing and improving the implementation should be done in later PRs to not clutter this one

but thx for the suggestion

philbucher:
I now found a better/cleaner/CMake-native way to enable C++17
I also added an example usage of the macro

roigcarlo:
I am pretty sure we can close this.

PR #9629, opened 2022-02-08
[Fluid] Improving compressible symbolic generation and testing

EduardGomezEscandell:
**📝 Description**
This PR is an improvement over https://github.com/KratosMultiphysics/Kratos/pull/9569.

The current way of testing (purely comparing text files generated) is sub-optimal because sometimes sympy rearranges the order of operations when minor changes are made to the generator (a+b may become b+a). These would cause the text comparison to fail, even if the behaviour of the file is the same once compiled. Sometimes the rearangments are a bit more complex than the ones of the example and are hard to interpret even by reading the file.

A new test is introduced and performs a two-step test:
- Calls the symbolic generator to create a test.py file.
- Import this file, and uses it to compare outputs

**🆕 Changelog**
- Created new test that runs the generated output from CompressibleNavierStokesSymbolicGenerator.
- Replaced text comparsion tests in nightly build
- Minor aesthetic improvements to generated elements
- Removed old hash-comparison-based tests since they are time-consuming to maintain.


philbucher:
you need to merge master to make the Win-Ci work
I kept canceling runs so that the CI fix could go through :)
Sorry for the inconvenience

applications/FluidDynamicsApplication/tests/compressible_navier_stokes_symbolic_generator_formulation_test.py
```diff
@@ -0,0 +1,148 @@
...
+            except Exception as e:
+                raise RuntimeError("Error in test " + testname) from e
+
+            if print_results:
+                print(result)
+
+            try:
+                self.assertVectorAlmostEqual(result, reference)
+            except AssertionError as e:
+                raise AssertionError("AssertionError in " + testname) from e
```
EduardGomezEscandell: This will become a one-liner if #9635 is accepted:
```
self.assertVectorAlmostEqual(result, reference, msg="Failure in sub-test "+testname)
```

The error message is also much cleaner.

philbucher: doesn't it automatically show which test failed?

philbucher:
Your tests seems pretty complicated, why don’t you follow the default test structure?

applications/FluidDynamicsApplication/tests/compressible_navier_stokes_symbolic_generator_formulation_test.py
EduardGomezEscandell: Yeah, but I want more detail. Without this, it would say that the failed test is, for instance `testQuadrilateral`. For each test multiple things are tested, so I wanted to know exactly which one fails.  Since it's a loop, just knowing the stack trace doesn't help.

I believe that the variable names here are confusing so I changed them (`testname` should be `subtest_name`).

EduardGomezEscandell:
> Your tests seems pretty complicated, why don’t you follow the default test structure?

The crux of the problem is that the tested code doesn't exist when the tests are run, and are generated during the test (since the whole point is to test code generation).

For each geometry there are various sinppets of code generated, so there are various things to test. We have concluded from experience that a simple text comparisson doesn't work, since sympy tends to generate diferent code with very minor changes (even while maintaining the same resulting behaviour). Hence the generated code needs to be run. All of this leads to having to do the following for each tested geometry:
- Generate the code `-> _Generate`
- Import the code `->_ImportSubTestSuite`
- Run the code snipets and compare each of the results against their references `-> _RunSubTestSuite`

I decided to call the generated code a sub-test suite and each snippet a sub-test, but better names will be accepted.
Well, better designs for these tests will also be accepted if anyone has any proposals  :sweat_smile:.

applications/FluidDynamicsApplication/tests/compressible_navier_stokes_symbolic_generator_formulation_test.py
```diff
@@ -0,0 +1,108 @@
...
+    def _RunSubTestSuite(self, sub_testsuite, print_results):
+        """Runs the generated code and compares it to the reference"""
+        tests = [subtest_name for subtest_name in dir(sub_testsuite)
+            if callable(getattr(sub_testsuite, subtest_name)) and subtest_name.startswith("test_")]
+
+        for subtest_name in tests:
+            try:
+                result, reference = getattr(sub_testsuite, subtest_name)()
+            except Exception as e:
+                raise RuntimeError("Error in sub-test " + subtest_name) from e
```
philbucher:
```suggestion
            self.subTest(subtest_name):
                result, reference = getattr(sub_testsuite, subtest_name)()
```

python already has this :)

=> then you would also not need the custom error message

check it out, I think this works for you

more examples: https://github.com/KratosMultiphysics/Kratos/blob/master/applications/TrilinosApplication/tests/test_trilinos_linear_solvers.py

other than that I don't have more suggestions

EduardGomezEscandell: Oh this is nice. Exactly what I wanted. Thanks.

applications/FluidDynamicsApplication/tests/compressible_navier_stokes_symbolic_generator_formulation_test.py
```diff
@@ -0,0 +1,115 @@
+import os
+
+import KratosMultiphysics
+import KratosMultiphysics.KratosUnittest as KratosUnitTest
+
+from KratosMultiphysics.FluidDynamicsApplication.symbolic_generation.compressible_navier_stokes \
```
philbucher, line 6: this uses `sympy` which is not part of the python standard library
Hence if someone (like me :D) that does not have sympy installed run the tests, then this test (and the entire test script) crashes 

Hence I recommend protecting this with a check in `test_FluidDynamicsApplication.py`
(see e.g. how `numpy` is handled in the [CoSim-tests](https://github.com/KratosMultiphysics/Kratos/blob/master/applications/CoSimulationApplication/tests/test_CoSimulationApplication.py))

applications/FluidDynamicsApplication/tests/compressible_navier_stokes_symbolic_generator_formulation_test.py
```diff
@@ -0,0 +1,115 @@
...
+            KratosMultiphysics.kratos_utilities.DeleteFileIfExisting(file_name)
+        self.files_to_remove = []
+
+    @classmethod
+    def _FormatVector(cls, vector, fmt = "{:>.6}"):
+        """Mimics KratosMultiphysics.vector.__str__ but with a customizable format string."""
+        output = "[{}]".format(vector.Size())
+        output = "("
+        output = output + ", ".join([fmt.format(x) for x in vector])
+        output = output + ")"
```
philbucher:
```suggestion
        output += "("
        output += output + ", ".join([fmt.format(x) for x in vector])
        output += output + ")"
```

I guess this is missing (or at least some of them)

applications/FluidDynamicsApplication/tests/compressible_navier_stokes_symbolic_generator_formulation_test.py
EduardGomezEscandell: Good point

applications/FluidDynamicsApplication/python_scripts/symbolic_generation/compressible_navier_stokes/compressible_navier_stokes_symbolic_generator.py
```diff
@@ -57,19 +59,45 @@ def __init__(self, settings):
...
+            "Inconsistent indentation in source file! Substitution {} found {} leading spaces, which is not a multiple of four.".format(
+            target_substring, leading_spaces
+        ))
+
     def _CollectAndReplace(self, target_substring, expression, name):
-        out = KratosSympy.OutputVector_CollectingFactors(expression, name, self.write_mode, replace_indices=False, assignment_op=" = ")
+        indentation_level  = self._FindAndRemoveIndentation(target_substring)
+
+        out = KratosSympy.OutputVector_CollectingFactors(expression, name, self.write_language, indentation_level=indentation_level, replace_indices=False, assignment_op=" = ")
         self.outstring = self.outstring.replace(target_substring, out)
```
rubenzorrilla, line 94: These might be useful for other elements. Maybe we can move them to the `sympy_fe_utilities.py` in the future.

rubenzorrilla (approved):

applications/FluidDynamicsApplication/python_scripts/symbolic_generation/compressible_navier_stokes/compressible_navier_stokes_symbolic_generator.py
EduardGomezEscandell: Agreed.

I left them here for now since it is quite inefficient (a very long string is copied in line 82, only to be copied again in line 94), so it should be used sparingly. If we want to put it in the core we might want to be smarter about it.

PR #9633, opened 2022-02-08
[IGA] Remove THICKNESS check from embedded truss element

tteschemacher:
this check isn't necessary but provides bad error messages.

furthermore deletes all trailing white spaces.


rickyaristio (approved):
@tteschemacher , thanks for pointing this out.

PR #9641, opened 2022-02-09
[BruteForcePointLocator] Add FindObject

miguelmaso:
**📝 Description**
With this changes it is possible to find a geometrical object (e.g. `FindIntersectedGeometricalObjects` returns a pointer vector of geometrical objects)

**🆕 Changelog**
- Add `FindObject`
- Style guide

### DIFF
--- a/kratos/utilities/brute_force_point_locator.cpp
+++ b/kratos/utilities/brute_force_point_locator.cpp
@@ -52,7 +52,7 @@ int BruteForcePointLocator::FindElement(const Point& rThePoint,
                                         const Globals::Configuration Configuration,
                                         const double LocalCoordTol) const
 {
-    int found_element_id; // if no element is found -1 will be returned
+    int found_element_id = -1; // if no element is found this will be returned
     const auto& r_elements = mrModelPart.GetCommunicator().LocalMesh().Elements();
     FindObject(r_elements, "Element",
                 rThePoint, found_element_id,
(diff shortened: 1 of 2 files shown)


philbucher:
Personally I am ok with merging this, but for me the question is more if we want to extend this class even though we have better options available (e.g. BinSearch) Why don't you use that one?

miguelmaso:
I want to project a line on a volume and do some operations at every node of the line (integrate along the direction of projection, search at a specified distance, find the bounds...)

g854

In a first stage, I use `FindIntersectedGeometricalObjects` (uses an octree, more or less like the bin based point locator) to find all the elements intersected by the projection of each point.

In a second stage, every node uses `BruteForcePointLocator` to perform the operations. Note that the candidates are very few elements.

miguelmaso:
> In a first stage, I use `FindIntersectedGeometricalObjects` (uses an octree, more or less like the bin based point locator) to find all the elements intersected by the projection of each point.
> 
> In a second stage, every node uses `BruteForcePointLocator` to perform the operations. Note that the candidates are very few elements.

Maybe the `BinBased` is faster than all of this?

philbucher (approved):
ok then :)
If you know what you are doing then fine :)

miguelmaso:
> the question is more if we want to extend this class even though we have better options available (e.g. BinSearch) Why don't you use that one?

thanks, this question helped me to find a better solution

PR #9644, opened 2022-02-10
[Trilinos][TrilinosExtensions] adding include dirs

philbucher:
This will significantly simplify the usage of Trilinos for anything that depends on it (i.e. the trilinos-extensions)
~~=> the magic will happen in the next PRs ;)~~

EDIT I updated the apps right away


applications/TrilinosApplication/CMakeLists.txt
```diff
@@ -22,11 +22,7 @@ endif(${TRILINOS_EXCLUDE_AMESOS_SOLVER} MATCHES ON)
 include_directories( ${KRATOS_SOURCE_DIR}/kratos )
 
 # Find trilinos solver package -- this is our own find package!!
-if (NOT TRILINOS_FOUND)
-    find_package(TRILINOS REQUIRED)
-endif()
-include_directories(SYSTEM ${TRILINOS_INCLUDE_DIR})
-link_directories(${TRILINOS_LIBRARIES})
+find_package(TRILINOS REQUIRED)
```
roigcarlo: Starting a discussion here, maybe not related but is a change I need to do and since this is opened....

Could we move this to the main CMakeLists.txt of Kratos? I am thinking on: 

```CMake
if(${USE_MPI} MATCHES ON )
  find_package(MPI REQUIRED)
  find_package(TRILINOS) # Notice I've removed the REQUIRED
  add_definitions( -DKRATOS_USING_MPI )
  set(CMAKE_CXX_COMPILE_FLAGS ${CMAKE_CXX_COMPILE_FLAGS} ${MPI_COMPILE_FLAGS})
  set(CMAKE_C_COMPILE_FLAGS ${CMAKE_C_COMPILE_FLAGS} ${MPI_COMPILE_FLAGS})
  set(CMAKE_CXX_LINK_FLAGS ${CMAKE_CXX_LINK_FLAGS} ${MPI_LINK_FLAGS})

  # Note: that this will be included in the compilation of the kratos core, so that it is not needed to include it somewhere else
  include_directories(SYSTEM ${MPI_INCLUDE_PATH})
endif(${USE_MPI} MATCHES ON )
```

I want to make this because I've found that, if someone wants to compile with MPI without actually having mpi capabilities in applications with trilinos extensions, is impossble to do it so without also compiling Trilinos ( as the extensions get added automagically when enabling MPI).

This change would enable to change the extensions to something like:

```CMake
## If trilinos is avaliable, switch on the Foo application
if(${USE_MPI} and ${TRILINOS_FOUND})
  message("Enabling trilinos extension for KratosFooApplication")
  add_subdirectory(trilinos_extension)
endif(${USE_MPI} and ${TRILINOS_FOUND})
```
and we can then issue a error in the trilinos app if this is not found.

roigcarlo: If you want to se it in action: [cd/mpi-container-linux](https://github.com/KratosMultiphysics/Kratos/tree/cd/mpi-container-linux)

philbucher: I know what you mean, I just never had this issue/usecase :)
As much as I would like to encapsulate Trilinos, I see that it is still necessary

For me it is ok for now
Since you implemented this already, could you please push it to this branch?

roigcarlo: Ok, I will merge my branch into this. It will have some more things but all is related to MPI so no problem.

roigcarlo:
Can we merge?

roigcarlo (approved):
Approving so if we can merge, we can merge.

PR #9647, opened 2022-02-10
[Core] Made Matrix and Vector versions consistent with assertAlmostEqual

EduardGomezEscandell:
**📝 Description**
Just noticed that the function signature of `assertVectorAlmostEqual` and `assertMatrixAlmostEqual` were:
```
assertVectorAlmostEqual(self, vector1, vector2, prec, msg)
assertMatrixAlmostEqual(self, vector1, vector2, prec, msg)
```
whereas python's unit test is:
```
assertAlmostEqual(self, first, second, places, msg, delta)
```
So I figured that by renaming `prec->places` and adding a delta, we could be consistent.

**🆕 Changelog**
To both `assertVectorAlmostEqual` and `assertMatrixAlmostEqual`:
- Renamed `prec` to `places`
- Changed function calls to use `places`.
- Added optional argument `delta`.


rubenzorrilla (commented):
Just saying that for the sake of consistency this makes sense to me.

roigcarlo (approved):
👍

PR #9649, opened 2022-02-11
[Mapping] improve searching for volume mapping

philbucher:
When running a large case I noticed that the search can be slow, hence this PR once again makes the search more efficient
Also I added some more informative prints (only with larger echo levels


mpentek (approved):
Useful development.

PR #9650, opened 2022-02-11
[CoSim] providing additional field

philbucher:
this can be used for anything that a solver might require


pooyan-dadvand (approved):
Thanks!

PR #9655, opened 2022-02-14
Thermal DEM Application

rlrangel:
First version of the Thermal DEM application, an extension of the standard DEM application to include thermal effects.


applications/DEMApplication/custom_elements/spheric_particle.cpp
```diff
@@ -914,6 +915,9 @@ void SphericParticle::ComputeBallToBallContactForce(SphericParticle::ParticleDat
                                   GlobalElasticContactForce, GlobalElasticExtraContactForce, TotalGlobalElasticContactForce, ViscoDampingLocalContactForce, cohesive_force, other_ball_to_ball_forces, r_elastic_force, r_contact_force, i, r_process_info);
             //TODO: make different AddUpForces for continuum and discontinuum (different arguments, different operations!)
 
+            // Store contact information needed for other processes
+            StoreBallToBallContactInfo(r_process_info, data_buffer, data_buffer.mpOtherParticle, GlobalContactForce, sliding);
```
maceligueta: Shouldn't this method be after the rotation forces have been computed? Otherwise we can only store information which does not contain "rotation forces"

applications/DEMApplication/custom_elements/spheric_particle.cpp
```diff
@@ -1103,6 +1107,9 @@ void SphericParticle::ComputeBallToRigidFaceContactForce(SphericParticle::Partic
                                      GlobalElasticContactForce, ViscoDampingLocalContactForce, cohesive_force, r_elastic_force,
                                      r_contact_force, mNeighbourRigidFacesElasticContactForce[i], mNeighbourRigidFacesTotalContactForce[i]);
 
+            // Store contact information needed for other processes
+            StoreBallToRigidFaceContactInfo(r_process_info, data_buffer, wall, GlobalContactForce, sliding);
```
maceligueta: Same comment. It seems like forces due to rotation can't be stored if we call the method at this location.

applications/DEMApplication/custom_conditions/RigidFace.cpp
```diff
@@ -343,7 +343,8 @@ void RigidFace3D::ComputeConditionRelativeData(int rigid_neighbour_index,
     if (points == 3 || points == 4)
     {
         unsigned int dummy_current_edge_index;
-        contact_exists = GeometryFunctions::FacetCheck(this->GetGeometry(), node_coordinates, radius, LocalCoordSystem, DistPToB, TempWeight, dummy_current_edge_index);
+        bool dummy_inside;
```
maceligueta: No need to call it _dummy_. It's ok if we don't use it after calling the function. It's something usual if the function requires the argument.

applications/DEMApplication/python_scripts/DEM_procedures.py
```diff
@@ -1463,6 +1463,8 @@ def AddGlobalVariables(self):
         if self.DEM_parameters["PostParticleMoment"].GetBool():
             self.PushPrintVar(self.PostParticleMoment, PARTICLE_MOMENT, self.global_variables)
         self.PushPrintVar(self.PostDeltaDisplacement, DELTA_DISPLACEMENT, self.global_variables)
+        self.PushPrintVar(self.PostTemperature, TEMPERATURE, self.global_variables)
```
maceligueta: Ideally, this should be in the new application. There are ways to do it. You can use your own Thermal_DEM_procedures.py which derives from this one and just adds this modification. If there is a Thermal DEM app, TEMPERATURE should appear nowhere in the DEM app.

maceligueta: Same for HEAT_FLUX

applications/ThermalDEMApplication/custom_elements/thermal_spheric_particle.cpp
```diff
@@ -0,0 +1,1443 @@
...
+    mpDirectConductionModel      = NULL;
+    mpIndirectConductionModel    = NULL;
+    mpConvectionModel            = NULL;
+    mpRadiationModel             = NULL;
+    mpFrictionModel              = NULL;
+    mpRealContactModel           = NULL;
+  }
+
+  Element::Pointer ThermalSphericParticle::Create(IndexType NewId, NodesArrayType const& ThisNodes, PropertiesType::Pointer pProperties) const {
+    return SphericParticle::Pointer(new ThermalSphericParticle(NewId, GetGeometry().Create(ThisNodes), pProperties));
```
maceligueta: Element::Pointer more consistent

applications/ThermalDEMApplication/custom_elements/thermal_spheric_particle.cpp
```diff
@@ -0,0 +1,1443 @@
...
+      mContactRadiusAdjusted      = mContactRadius;
+    }
+
+    KRATOS_CATCH("")
+  }
+
+  //------------------------------------------------------------------------------------------------------------
+  void ThermalSphericParticle::StoreBallToBallContactInfo(const ProcessInfo& r_process_info, SphericParticle::ParticleDataBuffer& data_buffer, SphericParticle* neighbor, double GlobalContactForce[3], bool sliding) {
+    KRATOS_TRY
+
```
maceligueta, line 355: I would here call the base class method. Now it is empty, but it may not be in the future.

applications/ThermalDEMApplication/custom_elements/thermal_spheric_particle.cpp
```diff
@@ -0,0 +1,1443 @@
...
+    // Add/Update parameters in map
+    mContactParamsParticle[neighbor] = params;
+
+    KRATOS_CATCH("")
+  }
+
+  //------------------------------------------------------------------------------------------------------------
+  void ThermalSphericParticle::StoreBallToRigidFaceContactInfo(const ProcessInfo& r_process_info, SphericParticle::ParticleDataBuffer& data_buffer, DEMWall* neighbor, double GlobalContactForce[3], bool sliding) {
+    KRATOS_TRY
+
```
maceligueta, line 405: Call to the base class method? Or you really want to discard the base class inf? I don't really know.

applications/ThermalDEMApplication/thermal_dem_application_variables.cpp
```diff
@@ -0,0 +1,80 @@
...
+  KRATOS_CREATE_VARIABLE(double,                               FLUID_LAYER_THICKNESS)
+  KRATOS_CREATE_VARIABLE(double,                               FLUID_DENSITY)
+  KRATOS_CREATE_VARIABLE(double,                               FLUID_VISCOSITY)
+  KRATOS_CREATE_VARIABLE(double,                               FLUID_THERMAL_CONDUCTIVITY)
+  KRATOS_CREATE_VARIABLE(double,                               FLUID_HEAT_CAPACITY)
+  KRATOS_CREATE_VARIABLE(double,                               FLUID_TEMPERATURE)
+  KRATOS_CREATE_VARIABLE(Vector,                               FLUID_VELOCITY)
+
+  // Flags (starting at 1000 to avoid conflicts with DEMFlags)
+  KRATOS_CREATE_LOCAL_FLAG(DEMThermalFlags, IS_ADIABATIC,                     1000);
```
maceligueta: 1000? I think there is only room for 64 bits. 32 are already used by the Kratos Core and the other 32 are partially used by the DEM.

maceligueta: It is weird that you did not have problems changing bit number 1000 of a variable which has 64 bits.

maceligueta (changes requested):

maceligueta:
Also, the DEM tests are not passing @rlrangel . Did you check which one is giving problems? We can help if you need. I suspect the changes in the way neighbours are searched/stored may have changed the behavior.

rlrangel:
Miguel Angel,

I will take a look at the suggestions later.

I forgot to ask you about the tests.

I noticed that it was not working, so I made an investigation some days ago.
I was adding my modifications one by one to the master to check which one
was causing the tests to fail.
I concluded that even my simplest modification was causing it, like
creating a new member function to the particle with empty implementation.
So I was wondering if the cause really was my modifications.

I will do this investigation again and inform you about if I get new
conclusions.

Best regards,
Rafael

applications/DEMApplication/custom_elements/spheric_particle.cpp
rlrangel: Yes, I was interested only on the normal forces, now I moved it to the end to be more generic.

applications/DEMApplication/custom_elements/spheric_particle.cpp
rlrangel: Same as above.

applications/DEMApplication/custom_conditions/RigidFace.cpp
rlrangel: Changed variable name.

applications/DEMApplication/python_scripts/DEM_procedures.py
rlrangel: Yes, in fact a have a subclass for doing exactly what you suggested, but it is not currently adding TEMPERATURE and HEAT_FLUX as these variables are still needed in the DEM Application because of the old thermal particles that were left there. When I remove these particles from the DEM App, I will also bring all the thermal variables.

applications/ThermalDEMApplication/custom_elements/thermal_spheric_particle.cpp
rlrangel: Ok

applications/ThermalDEMApplication/custom_elements/thermal_spheric_particle.cpp
rlrangel: Now calling empty base method.

applications/ThermalDEMApplication/custom_elements/thermal_spheric_particle.cpp
rlrangel: Now calling empty base method.

.gitignore
```diff
@@ -79,3 +79,5 @@ bin/*
 
 # External downloaded libs (TPL downloaded libs, etc...)
 external_libraries/tetgen/*
+
+applications/ThermalDEMApplication/DEM_HeatTransferModels.pdf
```
maceligueta: instead of modifying this file in the root folder, you can ignore this pdf file locally by editing `.git/info/exclude`.

applications/ThermalDEMApplication/thermal_dem_application_variables.cpp
rlrangel: I thought these flags were integers that we could give any value, but I noticed I get some problems if I repeat the same value from the DEM App, so I changed them for a high value that the DEM App would not reach. Really strange that it worked then.
For now, I am starting from 19, but soon I think will stop using these flags, it is less risky.

.gitignore
rlrangel: In the end, I decided not to ignore the pdf. I didn't know I made this modification in this file, now I removed.

applications/ThermalDEMApplication/thermal_dem_application_variables.cpp
rlrangel: I removed all the flags, but the IS_ADIABATIC because this one can be set also for the walls, and it would require a thermal wall sub-class just for adding a variable for this flag.

rlrangel:
> 

What is causing the tests to failure is adding "virtual" to functions **SetSearchRadiiOnAllParticles** and **SetSearchRadiiWithFemOnAllParticles** of the explicit strategy, so that I could override their implementations in my derived thermal strategy.
However, there are functions with the same name on the derived continuum strategy, and making them virtual in the base class is affecting the behavior.
Suggestions on how I can override them without adding the virtual specifier?

maceligueta:
I guess you just need to add ‘override’ in the declaration of the overriden method, in  the derived class, header file.

rlrangel:
If I write override in the declaration of the derived method without adding the virtual in the base method, I get an error when compiling saying that the derived method cannot override any method of the base class.

rlrangel:
I've just migrated all the old thermal implementations that was in the DEMApp to the new ThermalDEMApp.
Now, whoever wants to work with the continuum sintering particle, will have to do it from the thermal application.

maceligueta:
> If I write override in the declaration of the derived method without adding the virtual in the base method, I get an error when compiling saying that the derived method cannot override any method of the base class.

Of course you have to leave the virtual in the base class. I have no problem with that virtual.

rlrangel:
> > If I write override in the declaration of the derived method without adding the virtual in the base method, I get an error when compiling saying that the derived method cannot override any method of the base class.
> 
> Of course you have to leave the virtual in the base class. I have no problem with that virtual.

Ok. But then I should change something in the tests to make them work with the virtual?

maceligueta:
> > > If I write override in the declaration of the derived method without adding the virtual in the base method, I get an error when compiling saying that the derived method cannot override any method of the base class.
> > 
> > 
> > Of course you have to leave the virtual in the base class. I have no problem with that virtual.
> 
> Ok. But then I should change something in the tests to make them work with the virtual?

I was suggesting to add 'override' here (DEM app):
https://github.com/KratosMultiphysics/Kratos/blob/ecd7fee018e8a3237e4f4e21813235090f85ad56/applications/DEMApplication/custom_strategies/strategies/explicit_solver_continuum.h#L72
as this is overriding as well (now that the virtual is put in its base class)
With this, Kratos should at least compile, not sure about the tests. Let's go step by step.

rlrangel:
Oh ok, override in the continuum particle.
The result is the same, it fails on the same test.
I think that once the base method is virtual, the derived method is being called whether or not I set it to override.
What I don't understand is why this method was not virtual before. The derived implementation in the continuum particle is not supposed to be called, is it correct to be like this?

AlejandroCornejo:
> 

Once you add the `override `it compiles? In principle if you do not add the override it call the base method but you are never sure, any minor difference in the input parameters (a missing & is enough) can lead to missconceptions.

maceligueta:
> Oh ok, override in the continuum particle. The result is the same, it fails on the same test. I think that once the base method is virtual, the derived method is being called whether or not I set it to override. What I don't understand is why this method was not virtual before. The derived implementation in the continuum particle is not supposed to be called, is it correct to be like this?

Yes, it is weird. If it was not complaining before, why is it complaining now??? It seems they were being considered different methods, not an override. I do not know.

applications/DEMApplication/python_scripts/DEM_procedures.py
rlrangel: These variables were now moved to the thermal application.

rlrangel:
Ok.

maceligueta:
@rlrangel I think I fixed the problem in #9714 . You can merge master branch into this one.

rlrangel:
Thanks @maceligueta !
I merged master into this branch and now the tests of the DEM App are passing.

maceligueta (approved):

PR #9656, opened 2022-02-14
[FastPR][ROM] Fix nonlinear update case

rubenzorrilla:
**📝 Description**
Previous implementation is not valid for non-linear cases. Now we accumulate (and reset accordingly) the `ROM_SOLUTION_INCREMENT`.


Rbravo555 (approved):
LGTM!

PR #9657, opened 2022-02-14
[Core][Processes] Check domain size in vector based assign processes

rubenzorrilla:
**📝 Description**
So far, we were always fixing the third component in our Kratos core vector-based assign processes. Aside of the extra computational cost, this makes the addition of the `_Z` DOF component mandatory in order to use the `AssignVectorByDirectionProcess`.

I thought that the safest way to check was to search for the`DOMAIN_SIZE` variable in the `ProcessInfo` container of the model part of interest. If not found, the root model part is checked.

Creating the PR to trigger the CI.


rubenzorrilla:
To discuss: Right now I'm throwing an error if the domain size is neither 2 nor 3. Maybe we can default it to 3 to keep the current behavior.

loumalouomega (commented):
Clever, but CI is failing

rubenzorrilla:
> Clever, but CI is failing

Yes. The `IgaApplication` tests use model parts with `DOMAIN_SIZE` equal to 1 @tteschemacher. In the `ShapeOptimizationApplication` case `DOMAIN_SIZE` is not prescribed @armingeiser.

kratos/python_scripts/assign_vector_by_direction_process.py
```diff
@@ -69,13 +69,30 @@ def __init__(self, Model, settings ):
...
+        # Get domain size
+        # Note that we check both the current model part and the root one
+        if self.model_part.ProcessInfo.Has(KratosMultiphysics.DOMAIN_SIZE):
+            domain_size = self.model_part.ProcessInfo[KratosMultiphysics.DOMAIN_SIZE]
+            if domain_size not in [2,3]:
+                root_model_part = self.model_part.GetRootModelPart()
+                if root_model_part.ProcessInfo.Has(KratosMultiphysics.DOMAIN_SIZE):
+                    domain_size = root_model_part.ProcessInfo[KratosMultiphysics.DOMAIN_SIZE]
+                else:
+                    raise Exception(f"Domain size is found neither in '{self.model_part.Name}' nor in root model part '{self.model_part.GetRootModelPart().Name}' ProcessInfo containers.")
```
philbucher: why does it have to be an exception?
Why not optional (and defaulting to 3)?

I also noticed that there are a surprisingly large number of ModelParts that don't have DOAMIN_SIZE

philbucher: basically any ModelPart that was not created directly in the solver (i.e. any aux modelpart) - and those are probably like half of them

kratos/python_scripts/assign_vector_variable_process.py
```diff
@@ -67,11 +67,30 @@ def __init__(self, Model, settings ):
...
+        if self.model_part.ProcessInfo.Has(KratosMultiphysics.DOMAIN_SIZE):
+            domain_size = self.model_part.ProcessInfo[KratosMultiphysics.DOMAIN_SIZE]
+            if domain_size not in [2,3]:
+                root_model_part = self.model_part.GetRootModelPart()
+                if root_model_part.ProcessInfo.Has(KratosMultiphysics.DOMAIN_SIZE):
+                    domain_size = root_model_part.ProcessInfo[KratosMultiphysics.DOMAIN_SIZE]
+                else:
+                    raise Exception(f"Domain size is found neither in '{self.model_part.Name}' nor in root model part '{self.model_part.GetRootModelPart().Name}' ProcessInfo containers.")
+        else:
+            root_model_part = self.model_part.GetRootModelPart()
```
philbucher, line 83: not sure if this is necessary since the ProcessInfo is shared btw parent and SubModelParts

rubenzorrilla: Not necessarily. Depends on how you have created the model part (I'm thinking about the case of custom modelers, which is expected to be more common as we move towards the multistage).

kratos/python_scripts/assign_vector_by_direction_process.py
rubenzorrilla: My thought was to somehow enforce to have this field. However, I agree that maybe it's easier, at least for the moment, to default to three to keep current behavior (see my [comment](https://github.com/KratosMultiphysics/Kratos/pull/9657#issuecomment-1039361348)).

kratos/python_scripts/assign_vector_variable_process.py
philbucher: unless you deliberately change it, it is the same:
https://github.com/KratosMultiphysics/Kratos/blob/fbc8cc2c5d18cef13ca9c5e0357b86a7c7ed5dbb/kratos/sources/model_part.cpp#L2013

rubenzorrilla: Imagine this case. You have an mdpa and you parse it. At this point you have the model part and all its submodelparts with a pointer to the same empty `ProcessInfo`. Now you want to assign the `ProcessInfo` from another model part to this model part, so you call the `SetModelProcessInfo`. This sets the provided process info only in the current model part instance, not in its submodelparts. I faced this while developing a custom modeler, so I think it's a good idea to check the root if not found. At the end, it doesn't harm and it's on the safe side.

philbucher:
why is 1 not a valid input?
I think it is perfectly reasonable to use it for e.g 1D conduction 

To me it seems rather random to enforce a DOMAIN_SIZE of 2 or 3 for no good reason

philbucher:
to elaborate my answer: I don't think this process is the right place to constraint `DOMAIN_SIZE` to certain values. If this should be done, then somewhere else
From the implementational side I don't see any apparent reason to constrain it. You optional optimization makes sense to me

rubenzorrilla:
> your argumentation makes sense. I did not really consider that this is the **vector**... process No further comments then :)

Very good! Does this mean that as it is right now would be OK for you?

philbucher (commented):
some of the logic is duplicated and could probably be refactored to an external utility to be used by other processes too.E.g. in the `fix_vector_variable_process`

For me it is ok as is, approval I leave to @KratosMultiphysics/technical-committee

rubenzorrilla:
> some of the logic is duplicated and could probably be refactored to an external utility to be used by other processes too.E.g. in the `fix_vector_variable_process`
> 
> For me it is ok as is, approval I leave to @KratosMultiphysics/technical-committee

Cool. @tteschemacher changes in `IgaApplication` test settings make sense for you right?

tteschemacher:
> Cool. @tteschemacher changes in `IgaApplication` test settings make sense for you right?

sure. Thanks for editing it!

kratos/python_scripts/assign_vector_by_direction_process.py
```diff
@@ -69,13 +69,34 @@ def __init__(self, Model, settings ):
...
+            if root_model_part.ProcessInfo.Has(KratosMultiphysics.DOMAIN_SIZE):
+                domain_size = root_model_part.ProcessInfo[KratosMultiphysics.DOMAIN_SIZE]
+            else:
+                domain_size = 3
+                warn_msg = f"DOMAIN_SIZE is found neither in '{self.model_part.Name}' nor in root model part '{self.model_part.GetRootModelPart().Name}' ProcessInfo containers. Defaulting to 3."
+                KratosMultiphysics.Logger.PrintWarning("AssignVectorByDirectionProcess", warn_msg)
+
+        # Check the obtained domain size value
+        if domain_size not in [2,3]:
+            raise ValueError(f"Domain size must be either 2 or 3. Found value {domain_size} in model part '{self.model_part.FullName()}'.")
```
armingeiser: You could give a hint here that if domain size is 1 the scalar process should be used?

rubenzorrilla: Good point. Thanks for the feedback.

pooyan-dadvand:
@KratosMultiphysics/technical-committee approves it.

kratos/python_scripts/assign_vector_by_direction_process.py
rubenzorrilla: Done!

philbucher (dismissed):

rubenzorrilla:
CI is failing because of the fstrings in CentOS. What should we do in this regard? As far as I understand we support this (Python 3.6).

philbucher:
> CI is failing because of the fstrings in CentOS. What should we do in this regard? As far as I understand we support this (Python 3.6).

Officially we do not as they are 3.6

rubenzorrilla:
> > CI is failing because of the fstrings in CentOS. What should we do in this regard? As far as I understand we support this (Python 3.6).
> 
> Officially we do not as they are 3.6

And what about upgrading the CentOS CI?

philbucher:
> > > CI is failing because of the fstrings in CentOS. What should we do in this regard? As far as I understand we support this (Python 3.6).
> > 
> > 
> > Officially we do not as they are 3.6
> 
> And what about upgrading the CentOS CI?

I think this is what we have bern discussing for the last years wrt upgrading to C++17 :D

philbucher:
In any case, we cannot silently update the python requirements 
This should be discussed in a separate issue

rubenzorrilla:
> > In any case, we cannot silently update the python requirements This should be discussed in a separate issue
> 
> Agree. I'll remove the fstrings from this PR then.

Done.

rubenzorrilla:
Bad news about this...

I've just realized about a countercase for the `AssignVectorVariableProcess` current implementation. A 2D beam problem to which you want to fix the z-rotation.
Hence, my proposal would be to, instead of checking the `DOMAIN_SIZE`, check the provided `values` array. Then we can filter by the `null` keyword to avoid creating the useless components processes.

Concerning the implementation of `AssignVectorByDirectionProcess`, I think we can keep it with the `DOMAIN_SIZE` check as, according to my understanding, it is assumed an in-plane direction check (otherwise the  `AssignVectorVariableProcess` should be used.

miguelmaso:
> Hence, my proposal would be to, instead of checking the DOMAIN_SIZE, check the provided values array. Then we can filter by the null keyword to avoid creating the useless components processes.

I'm working with 2D simulations and this is my approach. It can be easily specified from the interface, see [this example](
https://github.com/KratosMultiphysics/GiDInterface/blob/317e7e4c6d8ae1860807f52911b6b520884e34c6/kratos.gid/apps/ShallowWater/xml/XmlController.tcl#L35)

rubenzorrilla:
> > Hence, my proposal would be to, instead of checking the DOMAIN_SIZE, check the provided values array. Then we can filter by the null keyword to avoid creating the useless components processes.
> 
> I'm working with 2D simulations and this is my approach. It can be easily specified from the interface, see [this example](https://github.com/KratosMultiphysics/GiDInterface/blob/317e7e4c6d8ae1860807f52911b6b520884e34c6/kratos.gid/apps/ShallowWater/xml/XmlController.tcl#L35)

The bad point of doing so is that we need to manually specify it for each BC in the GUI. However, after considering my last comment, I'm afraid that we cannot do something much better than this... Thanks.

rubenzorrilla:
I think that this is an end road.

PR #9659, opened 2022-02-14
[ROM] Sort variables alphabetically in ROM basis output process

rubenzorrilla:
**📝 Description**
It is required to follow the same sorting criteria that the DOFset uses in order to enable the HROM mesh visualization of mixed problems (e.g. CFD). This is achieved by sorting the ROM variables alphabetically.

Doing a standalone PR with this change to ease the current tests update in case it is necessary.


Rbravo555 (approved):
LGTM!

PR #9660, opened 2022-02-14
[FastPR][Core] Delete model part modeler

rubenzorrilla:
**📝 Description**
A simple modeler to enable deleting model parts in multi-stage simulations.


ddiezrod:
Maybe it would be nice to be able to remove more than one model part, having a list as an input. Not a strong opinion though.

rubenzorrilla:
> Maybe it would be nice to be able to remove more than one model part, having a list as an input. Not a strong opinion though.

The original idea is that, if you need to do this you create extra modelers in the modelers list. Nonetheless, I agree that this might be a bit annoying if you want to remove a collection of submodelparts.

Anyhow, and considering that the potential use of this is to remove a complete (main) model part, I think we can go on with the idea of one `DeleteModelPartModeler` per model part. If this becomes in a bottleneck when advancing in the multistage design we can either enhance this modeler or create a new one.

roigcarlo:
I think I go with @ddiezrod with this one. Since it is not much of a trouble (programmatically speaking) to delete multiple modelparts from a single model probably is a good idea that modeler are not per modelpart but per functionality.

If you have `A`, `B` modelparts associated to the calculation of whatever and you are not gonna use them in the following stage it makes sense to delete them together no?. In fact I must say I was implementing it this way on my side already :S

rubenzorrilla:
Actually, I think you're both right @ddiezrod and @roigcarlo.

If we enhance current design to accept a list of model parts you can either
- use `n` modelers featuring a unique item list to remove `n` model (or submodel) parts
- use `1` modeler featuring a list with `n` entries to remove `n` model (or submodel) parts
- a combination of both options to remove model parts according to any functionality criteria (e.g. one modeler deletes all the model parts used for the calculation of wathever and another modeler removes all the model parts used for the postprocessing of wathever)

I like the idea. Thanks for the feedback :love_you_gesture:

rubenzorrilla:
@ddiezrod @roigcarlo now a list of model parts can be provided :wink:

kratos/python_scripts/modelers/delete_model_parts_modeler.py
```diff
@@ -0,0 +1,71 @@
...
+        model_part_names.sort()
+
+        removed_sub_mp_list = []
+        for model_part_name in model_part_names:
+            if self.model.HasModelPart(model_part_name):
+                # Save the name of the model part and sub model parts that are to be removed
+                model_part = self.model.GetModelPart(model_part_name)
+                aux_to_remove_sub_mp_list = self.__GetSubModelPartsNames(model_part)
+
+                if self.settings["echo_level"].GetInt() > 1:
```
rubenzorrilla, line 36: Not sure if this is the verbosity level we want.

kratos/python_scripts/modelers/delete_model_parts_modeler.py
```diff
@@ -0,0 +1,71 @@
...
+                    aux_info_msg = " and submodelparts:" if len(aux_to_remove_sub_mp_list) else "."
+                    info_msg = f"About to delete model part '{model_part_name}'{aux_info_msg}"
+                    for name in aux_to_remove_sub_mp_list:
+                        info_msg += "\n - {}".format(name)
+                    KratosMultiphysics.Logger.PrintInfo("DeleteModelPartModeler", info_msg)
+
+                # Delete the model part through the model
+                self.model.DeleteModelPart(model_part_name)
+
+                if self.settings["echo_level"].GetInt() > 0:
```
rubenzorrilla, line 46: Not sure if this is the verbosity level we want.

roigcarlo (approved):
:+1:

PR #9661, opened 2022-02-15
[Mapping] additional checks

philbucher:
Basically adding check to check if historical variables are allocated
Before this was only checked in debug and causing memory errors in Release


mpentek (dismissed):
Thanks

mpentek (approved):
Approved

philbucher:
merged

PR #9662, opened 2022-02-15
[ROM] Corrected misleading error message

EduardGomezEscandell:
**📝 Description**
If yout try to use ROM on an application that is not suported, the error message was:
```
Python module 'KratosMultiphysics.CompressiblePotentialFlowApplication' is not available. Make sure 'CompressiblePotentialFlowApplication' is compiled.
```
when the issue is not compilation.

This has been changed to:
```
Exception: Python module 'KratosMultiphysics.CompressiblePotentialFlowApplication' is not available. Make sure 'CompressiblePotentialFlowApplication' is compiled and implemented.
Currently implemented applications are:
 - KratosMultiphysics.FluidDynamicsApplication
 - KratosMultiphysics.StructuralMechanicsApplication
 - KratosMultiphysics.ConvectionDiffusionApplication
To add a new implementation, do so in '_GetAvailableSolverWrapperModules' in /home/egomez/Work/Kratos/bin/Release/KratosMultiphysics/RomApplication/new_python_solvers_wrapper_rom.py
```

I'm new to this application so please go ahead and tell me if this doesn't make sense.

**🆕 Changelog**
- Changed error message so that it is more accurate.

### DIFF
--- a/applications/RomApplication/python_scripts/new_python_solvers_wrapper_rom.py
+++ b/applications/RomApplication/python_scripts/new_python_solvers_wrapper_rom.py
@@ -21,14 +21,17 @@ def CreateSolverByParameters(model, solver_settings, parallelism, analysis_stage
 
     # Filter and retrieve the Python solvers wrapper from the corresponding application
     #TODO: This filtering wouldn't be required if we were using a unified solvers wrapper module name
-    if application_module_name == "KratosMultiphysics.FluidDynamicsApplication":
-        solvers_wrapper_module_module_name = "python_solvers_wrapper_fluid"
-    elif application_module_name == "KratosMultiphysics.StructuralMechanicsApplication":
-        solvers_wrapper_module_module_name = "python_solvers_wrapper_structural"
-    elif application_module_name == "KratosMultiphysics.ConvectionDiffusionApplication":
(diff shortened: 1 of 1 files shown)


rubenzorrilla (commented):
Makes sense to me, but I wouldn't put the list as it is (otherwise we'll need to update the error each time an application is added). I'd say something like "compile it and make sure that it is added to the available_modules list in the `python_solvers_wrapper_rom.py`".

(note that I intentionally avoid the `new_` in the error message as the idea is to rename this file ASAP.

EduardGomezEscandell:
> Makes sense to me, but I wouldn't put the list as it is (otherwise we'll need to update the error each time an application is added). I'd say something like "compile it and make sure that it is added to the available_modules list in the `python_solvers_wrapper_rom.py`". 

The list is auto-generated from the dictionary, but I agree about directing the user to the location of the dictionary. See PR description for new message.

> (note that I intentionally avoid the `new_` in the error message as the idea is to rename this file ASAP.

I used `__file__` so it never goes outdated.

rubenzorrilla (dismissed):
:+1: Thanks!

rubenzorrilla (approved):

PR #9663, opened 2022-02-15
[Mapping] adding process for serial postprocessing

philbucher:
This PR adds the `SerialOutputProcess`. This process can be used in a distributed simulation to map results to a single rank and then do postprocessing on only this rank. This has two main advantages (see the documentation I added):
- Some output formats write one file per rank in distributed simulations, which leads to many files when running with many cores. This process collects the results on one rank and can hence reduce the number of files significantly
- Different meshes can be used to do the postprocessing. This is in particular useful when the computational mesh is very fine, but a coarser mesh would be sufficient for postprocessing.

Documentation is added incl an example usecase. I tested this already on the cluster with many cores
@mpentek please test if this also works for you


rubenzorrilla (commented):
Just mentioning that I went through it and makes sense for me.

applications/MappingApplication/python_scripts/serial_output_process.py
```diff
@@ -0,0 +1,126 @@
...
+        if settings["mapping_settings"].size() == 0:
+            raise Exception('no "mapping_settings" were specified!')
+
+        if len(settings["output_process_settings"].keys()) == 0:
+            raise Exception('no "output_process_settings" were specified!')
+
+        mdpa_file_name_destination = settings["mdpa_file_name_destination"].GetString()
+
+        model_part_origin = model.GetModelPart(settings["model_part_name_origin"].GetString())
+        self.model_part_destination = model.CreateModelPart(settings["model_part_name_destination"].GetString())
```
rubenzorrilla, line 48: Maybe you want to throw a warning in this case after #9598 .

mpentek (approved):
Seems to work.

PR #9621, opened 2022-02-07
Sparse matrices interface parallel

RiccardoRossi:
**📝 Description**
Adds basic python interface for distributed sparse matrices (and related objects)

Please mark the PR with appropriate tags: 
- Api Breaker, Mpi, etc...

**🆕 Changelog**
Adds basic python interface for distributed sparse matrices (and related objects). 
The goal is to enable basic manipulation from the python layer, so to eventually allow simple FEM assembly from the python layer as well as experimenting on preconditioners together with the scipy sparse interface

PR #9666, opened 2022-02-15
Sparse matrices interface 2

RiccardoRossi:
**📝 Description**
this PR completes the minimal interface needed to write a minimal, parallel FEM code in python using our data structures.

- Improvement

**🆕 Changelog**


RiccardoRossi:
for reference a minimal, admittedly not too clean, file to read a mdpa and ensamble a laplacian in parallel is as follows

~~~py
import KratosMultiphysics
import KratosMultiphysics.mpi
import KratosMultiphysics.ConvectionDiffusionApplication
import KratosMultiphysics.MetisApplication
import KratosMultiphysics.scipy_conversion_tools
from KratosMultiphysics.gid_output_process import GiDOutputProcess
from KratosMultiphysics.mpi.distributed_import_model_part_utility import DistributedImportModelPartUtility

import numpy
import scipy
import scipy.sparse
import scipy.sparse.linalg
import sys
import cg

def DivideInPartitions(NumTerms,NumThreads):
    Partitions = numpy.zeros(NumThreads + 1)
    PartitionSize = int(NumTerms / NumThreads)
    Partitions[0] = 0
    Partitions[NumThreads] = NumTerms
    for i in range(1,NumThreads):
        Partitions[i] = Partitions[i-1] + PartitionSize 
    return Partitions

def ComputeBounds(N,Ndivisions,current_rank ):                            
    partition = DivideInPartitions(N,Ndivisions)
    return numpy.array([partition[current_rank],partition[current_rank+1]], dtype=int)

comm  = KratosMultiphysics.DataCommunicator.GetDefault()

#identify variables to be used in sorvel the poisson-type problem
laplacian_settings = KratosMultiphysics.ConvectionDiffusionSettings()
laplacian_settings.SetUnknownVariable(KratosMultiphysics.TEMPERATURE)
laplacian_settings.SetDiffusionVariable(KratosMultiphysics.CONDUCTIVITY)
laplacian_settings.SetVolumeSourceVariable(KratosMultiphysics.HEAT_FLUX)
laplacian_settings.SetReactionVariable(KratosMultiphysics.REACTION_FLUX)

#read input file
current_model = KratosMultiphysics.Model()
model_part = current_model.CreateModelPart("Main")
model_part.AddNodalSolutionStepVariable(KratosMultiphysics.TEMPERATURE)
model_part.AddNodalSolutionStepVariable(KratosMultiphysics.REACTION_FLUX)
model_part.AddNodalSolutionStepVariable(KratosMultiphysics.CONDUCTIVITY)
model_part.AddNodalSolutionStepVariable(KratosMultiphysics.HEAT_FLUX)
model_part.AddNodalSolutionStepVariable(KratosMultiphysics.PARTITION_INDEX)

settings = KratosMultiphysics.Parameters("""{
            "model_import_settings" : {
                "input_type" : "mdpa",
                "input_filename" : "coarse_square",
                "partition_in_memory" : true
            },
            "echo_level" : 0
        }""")
import_util = DistributedImportModelPartUtility(model_part, settings)
import_util.ImportModelPart()
import_util.CreateCommunicators()



#configure the variables to be used in the solution
process_info = model_part.ProcessInfo
process_info[KratosMultiphysics.CONVECTION_DIFFUSION_SETTINGS] = laplacian_settings

#assign material properties (everythign to 1 to make it simple)
for node in model_part.Nodes:
    node.SetSolutionStepValue(KratosMultiphysics.CONDUCTIVITY,1.0)
    node.SetSolutionStepValue(KratosMultiphysics.HEAT_FLUX,1.0)

#generate dofs and assign an id
dof_list = []
for node in model_part.Nodes:
    node.AddDof(KratosMultiphysics.TEMPERATURE, KratosMultiphysics.REACTION_FLUX)
    dof = node.GetDof(KratosMultiphysics.TEMPERATURE)
    dof.EquationId = node.Id-1 #this is where the entry will go in the system of equations
    dof_list.append(dof)

list_of_nodes = []
for elem in model_part.Elements:
    for node in elem.GetGeometry():
        list_of_nodes.append(node)
list_of_nodes = set(list_of_nodes)

#apply dirichlet BCs
for node in model_part.GetSubModelPart("ImposedTemperature2D_Imposed_temperature_Auto2").Nodes:
    node.Fix(KratosMultiphysics.TEMPERATURE)

#compute local size
max_equation_id = 0
for dof in dof_list:
    max_equation_id = max(max_equation_id, dof.EquationId)
max_equation_id = comm.MaxAll(max_equation_id)
my_rank = comm.Rank()
dofs_bounds = ComputeBounds(max_equation_id+1,comm.Size(), my_rank)
local_size = dofs_bounds[1]-dofs_bounds[0]

#construct graph of system of equations 
rhs = KratosMultiphysics.Vector()
lhs = KratosMultiphysics.Matrix()

Agraph = KratosMultiphysics.mpi.DistributedSparseGraph(local_size, comm);
for elem in model_part.Elements:
    elem.Initialize(process_info)
    equation_ids = elem.EquationIdVector(process_info)
    Agraph.AddEntries(numpy.array(equation_ids))
Agraph.Finalize()

#Build FEM matrix
rhs = KratosMultiphysics.Vector()
lhs = KratosMultiphysics.Matrix()

A = KratosMultiphysics.mpi.DistributedCsrMatrix(Agraph)
A.SetValue(0.0)

b = KratosMultiphysics.mpi.DistributedSystemVector(Agraph)
b.SetValue(0.0)

A.BeginAssemble()
b.BeginAssemble()
for elem in model_part.Elements:
    equation_ids = elem.EquationIdVector(process_info)
    elem.CalculateLocalSystem(lhs,rhs,process_info)
    A.Assemble(lhs,equation_ids)
    b.Assemble(rhs,equation_ids)    
A.FinalizeAssemble()
b.FinalizeAssemble()

##apply dirichlet conditions
free_dofs_vector = KratosMultiphysics.mpi.DistributedSystemVector(b.GetNumbering())
free_dofs_vector.SetValue(1.0) #by default everything is free

#########################################################################
##find out fixity (quick and dirty, using a all-to-all communication)
fixed_dofs_equationids = []
for dof in dof_list:
    if(dof.IsFixed()):
        fixed_dofs_equationids.append(dof.EquationId)

gathered_equationids = comm.GathervInts(fixed_dofs_equationids,0)
complete_fixed_dofs_equationids = []
if(comm.Rank()==0):
    for msg in gathered_equationids:
        for eq_id in msg:
            complete_fixed_dofs_equationids.append(eq_id)
    print(complete_fixed_dofs_equationids)
complete_fixed_dofs_equationids = comm.BroadcastInts(complete_fixed_dofs_equationids,0)

for dof_gid in complete_fixed_dofs_equationids:
    if(A.GetRowNumbering().IsLocal(dof_gid)):
        dof_lid =  b.GetNumbering().LocalId(dof_gid)
        free_dofs_vector[dof_lid] = 0.0
#########################################################################

sys.stdout.flush()
A.ApplyHomogeneousDirichlet(free_dofs_vector,1.0,b)

#solve system matrix
dx = cg.cg(A,b) #note that in Kratos we always solve for the increment

#update Kratos Database
ids = []
for dof in dof_list:
    ids.append(dof.EquationId)

update_importer = KratosMultiphysics.mpi.DistributedVectorImporter(comm,ids,dx.GetNumbering())
values = update_importer.ImportData(dx)
for i in range(len(dof_list)):
    dof_list[i].SetSolutionStepValue(values[i])
#visualize results.
res_name = "results_" + str(comm.Rank())
gid_output = GiDOutputProcess(model_part,
                            res_name,
                            KratosMultiphysics.Parameters("""
                                {
                                    "result_file_configuration": {
                                        "gidpost_flags": {
                                            "GiDPostMode": "GiD_PostAscii",
                                            "WriteDeformedMeshFlag": "WriteUndeformed",
                                            "WriteConditionsFlag": "WriteConditions",
                                            "MultiFileFlag": "SingleFile"
                                        },
                                        "file_label": "time",
                                        "output_control_type": "step",
                                        "output_interval": 1.0,
                                        "body_output": true,
                                        "node_output": false,
                                        "skin_output": false,
                                        "plane_output": [],
                                        "nodal_results": ["TEMPERATURE"],
                                        "nodal_nonhistorical_results": [],
                                        "nodal_flags_results": [],
                                        "gauss_point_results": [],
                                        "additional_list_files": []
                                    }
                                }
                                """)
                            )

gid_output.ExecuteInitialize()
gid_output.ExecuteBeforeSolutionLoop()
gid_output.ExecuteInitializeSolutionStep()
gid_output.PrintOutput()
gid_output.ExecuteFinalizeSolutionStep()
gid_output.ExecuteFinalize()
~~~

RiccardoRossi:
and a working CG as follows

~~~py

import KratosMultiphysics
import KratosMultiphysics.mpi
import math

def cg(A, b, x0=None, tol=1e-5, maxiter=100, M=None):
    n = b.Size()
    rank = b.GetComm().Rank()

    y = KratosMultiphysics.mpi.DistributedSystemVector(b.GetNumbering())
    Ap = KratosMultiphysics.mpi.DistributedSystemVector(b.GetNumbering())
    x = KratosMultiphysics.mpi.DistributedSystemVector(b.GetNumbering())
    x.SetValue(0.0)
        
    #r = b - A@x
    r = KratosMultiphysics.mpi.DistributedSystemVector( b ) #copy constructor r = b
    y.SetValue(0.0)
    A.SpMV(x,y) # y = A@x
    r.Sub(y) #TODO: find out why r -= y does not work

    p = KratosMultiphysics.mpi.DistributedSystemVector( r ) # p=r
    rsold = r.Dot(r,0) #note that this is only available on rank0
    rsold = b.GetComm().Broadcast(rsold,0)

    for i in range(maxiter):
        #Ap=A@p
        Ap.SetValue(0.0)
        A.SpMV(p,Ap) 

        Ap_dot_p = Ap.Dot(p,0)
        Ap_dot_p = b.GetComm().Broadcast(Ap_dot_p,0)
        alpha = rsold/(Ap_dot_p)
        x.Add(alpha,p) #x = x + alpha * p;
        r.Add(-alpha,Ap) #r = r - alpha * Ap;
        rsnew = r.Dot(r,0)
        rsnew = b.GetComm().Broadcast(rsnew,0)
        
        if(rank == 0):
            print(i,math.sqrt(rsnew))
        
        if math.sqrt(rsnew) < tol :
            break
        else:
            p.Mult(rsnew/rsold)
            p.Add(r)
            rsold=rsnew

    if(rank == 0)      :
        if(i == maxiter):
            print("error: reached the max number of iterations")
        print("final iterations = ", iter)
        print("final norm = ",math.sqrt(rsnew))

    return x 
~~~

kratos/containers/csr_matrix.h
```diff
@@ -593,8 +592,45 @@ class CsrMatrix final
...
+    void ApplyHomogeneousDirichlet(const TVectorType1& free_dofs_vector, 
+                                    const TDataType DiagonalValue, 
+                                    TVectorType2& rRHS)
+    {
+        KRATOS_ERROR_IF(size1() != free_dofs_vector.size() ) << "ApplyDirichlet: mismatch between row sizes : " << size1()  
+            << " and free_dofs_vector size " << free_dofs_vector.size() << std::endl;
+        KRATOS_ERROR_IF(size2() != free_dofs_vector.size() ) << "ApplyDirichlet: mismatch between col sizes : " << size2()  
+            << " and free_dofs_vector size " << free_dofs_vector.size() << std::endl;
+
+        if(nnz() != 0)
```
loumalouomega: I just want to mention that our style is:

~~~
if () {
} else {
}
~~~

Idem for the others

kratos/containers/distributed_csr_matrix.h
```diff
@@ -456,7 +456,48 @@ class DistributedCsrMatrix final
...
+        auto off_diag_free = mpVectorImporter->ImportData(free_dofs_vector); //obtain the nonlocal components of the free_dofs_vector
+        if(GetOffDiagonalBlock().nnz() != 0)
+        {
+            KRATOS_ERROR_IF(GetOffDiagonalBlock().size1() != free_dofs_vector.LocalSize() ) << "ApplyHomogeneousDirichlet - OffDiagonalBlock: mismatch between row sizes : " << GetOffDiagonalBlock().size1()  
+                << " and free_dofs_vector size " << free_dofs_vector.LocalSize() << std::endl;
+            KRATOS_ERROR_IF(GetOffDiagonalBlock().size2() != off_diag_free.size() ) << "ApplyHomogeneousDirichlet - OffDiagonalBlock: mismatch between col sizes : " << GetOffDiagonalBlock().size2()  
+                << " and free_dofs_vector size " << off_diag_free.size() << std::endl;
+            IndexPartition<IndexType>(GetDiagonalBlock().size1()).for_each( [&](IndexType i){
+
+                IndexType row_begin = GetOffDiagonalBlock().index1_data()[i];
```
loumalouomega:
```suggestion
                const IndexType row_begin = GetOffDiagonalBlock().index1_data()[i];
```

kratos/containers/distributed_csr_matrix.h
```diff
@@ -456,7 +456,48 @@ class DistributedCsrMatrix final
...
+                IndexType row_end   = GetOffDiagonalBlock().index1_data()[i+1];
```
loumalouomega:
```suggestion
                const IndexType row_end   = GetOffDiagonalBlock().index1_data()[i+1];
```

kratos/containers/distributed_csr_matrix.h
```diff
@@ -456,7 +456,48 @@ class DistributedCsrMatrix final
...
+                if(std::abs(free_dofs_vector.GetLocalData()[i] - 1.0) < 1e-10) //row corresponding to free dofs NOTE that we check if it is approx zero
```
loumalouomega: Please, follow the style as previously stated

kratos/containers/distributed_csr_matrix.h
```diff
@@ -456,7 +456,48 @@ class DistributedCsrMatrix final
...
+                {
+                    for(IndexType k = row_begin; k < row_end; ++k){
+                        IndexType col = GetOffDiagonalBlock().index2_data()[k];
```
loumalouomega:
```suggestion
                        const IndexType col = GetOffDiagonalBlock().index2_data()[k];
```

kratos/mpi/python/add_distributed_sparse_matrices_to_python.cpp
```diff
@@ -23,8 +23,7 @@
 #include "containers/distributed_csr_matrix.h"
 #include "add_distributed_sparse_matrices_to_python.h"
 #include "mpi/utilities/amgcl_distributed_csr_spmm_utilities.h"
-
-
+#include <pybind11/operators.h>
```
loumalouomega: This should be in External includes

kratos/mpi/python/add_distributed_sparse_matrices_to_python.cpp
```diff
@@ -108,10 +107,19 @@ void AddDistributedSparseMatricesToPython(pybind11::module& m)
         })
         .def("Dot", &DistributedSystemVector<double,IndexType>::Dot)
         //inplace
-        .def("__iadd__", [](DistributedSystemVector<double,IndexType>& self, const DistributedSystemVector<double,IndexType>& other_vec){self += other_vec; }, py::is_operator())
-        .def("__isub__", [](DistributedSystemVector<double,IndexType>& self, const DistributedSystemVector<double,IndexType>& other_vec){self -= other_vec;  }, py::is_operator())
-        .def("__imul__", [](DistributedSystemVector<double,IndexType>& self, const double& value){ self*=value; }, py::is_operator())
-        .def("__itruediv__", [](DistributedSystemVector<double,IndexType>& self, const double& value){ self/=value; }, py::is_operator())
+        // .def(py::self += py::self)
```
loumalouomega: Remove if unused?

loumalouomega (changes requested):
Comments

kratos/containers/csr_matrix.h
```diff
@@ -593,8 +592,45 @@ class CsrMatrix final
     // RightScaling
     // SymmetricScaling
 
-    //TODO
-    //void ApplyDirichlet
+    template<class TVectorType1, class TVectorType2=TVectorType1>
+    void ApplyHomogeneousDirichlet(const TVectorType1& free_dofs_vector,
```
philbucher:
```suggestion
    void ApplyHomogeneousDirichlet(const TVectorType1& rFreeDofsVector, 
```

kratos/containers/csr_matrix.h
```diff
@@ -593,8 +592,45 @@ class CsrMatrix final
...
+        if(nnz() != 0)
+        {
+            IndexPartition<IndexType>(size1()).for_each( [&](IndexType i){
+
+                //set to zero the relevant row in the RHS
+                rRHS[i] *= free_dofs_vector[i];
+
+                IndexType row_begin = index1_data()[i];
+                IndexType row_end   = index1_data()[i+1];
+                if(std::abs(free_dofs_vector[i]-1.0) < 1e-10) //row corresponding to free dofs NOTE that we check if it is approx zero
```
philbucher: 1e-10 seems rather large no?
Did you choose this from experience?

kratos/containers/distributed_csr_matrix.h
```diff
@@ -456,7 +456,48 @@ class DistributedCsrMatrix final
         return tmp;
     }
 
+    void ApplyHomogeneousDirichlet(const DistributedSystemVector<TDataType,TIndexType>& free_dofs_vector,
```
philbucher:
```suggestion
    void ApplyHomogeneousDirichlet(const DistributedSystemVector<TDataType,TIndexType>& rFreeDofsVector, 
```

kratos/python/add_mesh_to_python.cpp
```diff
@@ -515,7 +515,12 @@ void  AddMeshToPython(pybind11::module& m)
 //     .def(SolutionStepVariableIndexingPython<Element, Variable<vector<double> > >())
 //     .def(SolutionStepVariableIndexingPython<Element, Variable<DenseMatrix<double> > >())
     .def("Initialize", &EntityInitialize<Element>)
-    //.def("CalculateLocalSystem", &Element::CalculateLocalSystem)
+    .def("EquationIdVector", [](const Element& self, const ProcessInfo& rProcessInfo){
```
philbucher, line 518: so `ublas::vector` is passed by copy but `ublas::matrix` is passed by reference?

kratos/python/add_sparse_matrices_to_python.cpp
```diff
@@ -84,10 +82,11 @@ void AddSparseMatricesToPython(pybind11::module& m)
     //************************************************************************************************************
     //************************************************************************************************************
     py::class_<CsrMatrix<double,IndexType>, CsrMatrix<double,IndexType>::Pointer >(m, "CsrMatrix")
+    .def(py::init<>())
```
philbucher, line 85:
```suggestion
```

I would not add this one as it can lead to compatibility issues with the distributed version

philbucher (commented):
overall looks good to me with minor comments

One large comment though regarding #9621: why did you add the distributed stuff in the MPICore? We discussed before that it does NOT depend on MPI as all the MPI-related functionalities are accessed though the `DataCommunicator`. Thats why we left it in the Core and didn't add it to the MPICore
Unfortunately I only realized this now

kratos/mpi/python/add_distributed_sparse_matrices_to_python.cpp
RiccardoRossi: i removed it all alltogether. It looks like it requires a more modern pybind than we have (or a higher c++ version)

kratos/mpi/python/add_distributed_sparse_matrices_to_python.cpp
RiccardoRossi: removed

kratos/containers/csr_matrix.h
RiccardoRossi: ok

kratos/containers/csr_matrix.h
RiccardoRossi: the values should be either 1 or 0

so an error of 1e-10 is 10 orders of magnitude smaller (and it needs only to take into account smaller communication-related glitches)

kratos/python/add_mesh_to_python.cpp
RiccardoRossi: i don't follow your comment...

EquationIdsVector is returned not passed

kratos/containers/csr_matrix.h
RiccardoRossi: i think it is clearer as it is now.

I passed astyle on the file for other style correcitons

kratos/containers/distributed_csr_matrix.h
RiccardoRossi: no, i prefer the style as it is now

kratos/python/add_sparse_matrices_to_python.cpp
RiccardoRossi: i would prefer to leave this as it is, as it allows doing some low level manipulation in order to construct the matrix from arrays (see e.g. what is being done in the amgcl_conversion_utils)

on the other side this cannot be done for the distributed matrix

RiccardoRossi:
per regards moving the matrix to the core instead of the distributed core, i think you are right, however i would prefer doing it as a new PR

kratos/containers/distributed_csr_matrix.h
RiccardoRossi: ok

kratos/python/add_sparse_matrices_to_python.cpp
RiccardoRossi: this is to allow some low level operations for the serial matrix (see e.g. what is done in the amgcl_conversion_utils)

kratos/containers/csr_matrix.h
RiccardoRossi: the values should be 1 or 0, so it is quite small. 

I'll make it 1e-14 but it cannot be smaller since it is relative to 1 (so the maximum i could use in double precision is 1e-15)

kratos/containers/distributed_csr_matrix.h
RiccardoRossi: ok

kratos/python/add_mesh_to_python.cpp
RiccardoRossi: i don't follow you. EquationIds are RETURNED...

kratos/containers/distributed_csr_matrix.h
RiccardoRossi: that was too harsh, sorry.

what i mean is that here we have two quite large blocks. i prefer to have them separated more clearly

kratos/containers/distributed_csr_matrix.h
RiccardoRossi: ok

kratos/python/add_mesh_to_python.cpp
philbucher: I mean why do you expose this one differently from `CalculateLocalSystem`
=> passing by ref would be more efficient and consistent

kratos/python/add_sparse_matrices_to_python.cpp
philbucher: ok up to you

philbucher:
> per regards moving the matrix to the core instead of the distributed core, i think you are right, however i would prefer doing it as a new PR

sure, this should be a separate PR

kratos/containers/csr_matrix.h
philbucher: ah ok I didn't understand that those are used for this purpose

kratos/python/add_mesh_to_python.cpp
RiccardoRossi: well, i thought it is easier to write

~~~py
ids = EquationIdVector(ProcessInfo)
~~~

but i could not do the same with the CalculateLocalSystem since it returns two parameters. (i guess it would be possible to do via pybind, but it is definitely not simple)

RiccardoRossi: also there appear to be some problem in passing DenseIntVector by reference. Don't ask me what is the problem exactly...

philbucher: > (i guess it would be possible to do via pybind, but it is definitely not simple)

this is actually quite straight forward (FYI): https://github.com/KratosMultiphysics/CoSimIO/blob/v1.0.1/co_sim_io/python/co_sim_io_python.cpp#L32-L55

> also there appear to be some problem in passing DenseIntVector by reference.

ok this is what I thought. I guess the DenseVector is not exposed like the matrix hence it doesn't work. Just wanted to confirm

philbucher (dismissed):
If you can move the interface to the Core it would be great, but this PR is ready for me

loumalouomega (dismissed):

kratos/containers/csr_matrix.h
```diff
@@ -514,21 +549,25 @@ class CsrMatrix final
             AtomicAdd(value_data()[k], rMatrixInput(i_local,0));
 
             //now find other entries. note that we assume that it is probably that next entries immediately follow in the ordering
-            for(unsigned int j_local=1; j_local<local_size; ++j_local){
+            for(unsigned int j_local=1; j_local<local_size; ++j_local)
+            {
                 J = EquationId[j_local];
 
-                if(k+1<row_end && index2_data()[k+1] == J){
+                if(k+1<row_end && index2_data()[k+1] == J)
```
loumalouomega: The style was Ok before...

RiccardoRossi:
thx @loumalouomega for the style corrections

@loumalouomega or @philbucher could u approve?

loumalouomega (approved):
Okay

RiccardoRossi:
@philbucher i looked into it and unfortunately this remains to be part of the Kratos mpi.

The point is that for SpMM we rely on AMGCL and the lib includes internally mpi.h

PR #9669, opened 2022-02-16
[Fluid][NighlyBuild] Set two fluid source tests as expected failures

marcnunezc:
**📝 Description**
These tests will be fixed after #9592 which may take some time. This PR sets these tests as expected failures to not hide other possible failures in the nightly build. 

**🆕 Changelog**
E.g.
- Set two fluid mass conservation source tests as expected failures

### DIFF
--- a/applications/FluidDynamicsApplication/tests/two_fluid_mass_conservation_source_test.py
+++ b/applications/FluidDynamicsApplication/tests/two_fluid_mass_conservation_source_test.py
@@ -55,16 +55,19 @@ class TwoFluidMassConservationTest(UnitTest.TestCase):
         self.print_reference_values = False
 
     # runs the two dimensinal test case
+    @UnitTest.expectedFailure #To be fixed in #9592
     def testTwoFluidMassConservationTest2D(self):
         self._has_inlet = False
         self._AuxiliaryRunTest("ProjectParameters2D.json")
 
     # runs the three dimensional test case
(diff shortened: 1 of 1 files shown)


ddiezrod:
Why not just update the current results of these tests? If I get it correctly, current results are not wrong, they are just slightly different from what they used to be. If #9592 is going to take some time, I would update the results now and then again when the other PR is merged.

rubenzorrilla:
> Why not just update the current results of these tests? If I get it correctly, current results are not wrong, they are just slightly different from what they used to be. If #9592 is going to take some time, I would update the results now and then again when the other PR is merged.

The point is that we're working on an alternative formulation to handle this. Hopefully we'll merge it soon, so we'll need to update results again.

ddiezrod:
> > Why not just update the current results of these tests? If I get it correctly, current results are not wrong, they are just slightly different from what they used to be. If #9592 is going to take some time, I would update the results now and then again when the other PR is merged.
> 
> The point is that we're working on an alternative formulation to handle this. Hopefully we'll merge it soon, so we'll need to update results again.

I understand it. Its just that I personally think its better to update the test results twice rather than deactivating and reactivating the test again in the future, as it takes more or less the same time to do the change. Im not blocking this though, if you prefer to do it this way.

rubenzorrilla (approved):

PR #9673, opened 2022-02-17
[FastPR][CMake] Update deprecations list

rubenzorrilla:
**📝 Description**
Simply updating the deprecation list in our `CMakeLists.txt`. So far we announced only the API-breaking changes. I think it is good to also post the the main behavior changes such as #9598.


roigcarlo (approved):
Nice idea! 👍

PR #9674, opened 2022-02-17
Sparse interface v3

RiccardoRossi:
**📝 Description**
Adding some missing functions to the sparse matrix interface, and improving the interoperability with numpy arays

enhancement

**🆕 Changelog**
adds interoperability with numpy arrays and bugfix in sparse interface
also bugfix in GetLocalData and GetData functions


RiccardoRossi:
this allows writing a distributed CG with scipy-based block preconditioning as follows:

~~~py
import KratosMultiphysics
import KratosMultiphysics.mpi
import math
import scipy
import scipy.sparse
import scipy.sparse.linalg
import numpy
import KratosMultiphysics.scipy_conversion_tools

def cg(A, b, x0=None, tol=1e-5, maxiter=100, M=None):
    n = b.Size()
    rank = b.GetComm().Rank()

    ##construct a local factorization of the diagonal block of A
    DiagBlockScipy = KratosMultiphysics.scipy_conversion_tools.to_csr(A.GetDiagonalBlock())
    Minv = scipy.sparse.linalg.factorized(DiagBlockScipy.tocsc())

    zk = KratosMultiphysics.mpi.DistributedSystemVector(b.GetNumbering())
    zk1 = KratosMultiphysics.mpi.DistributedSystemVector(b.GetNumbering())
    y = KratosMultiphysics.mpi.DistributedSystemVector(b.GetNumbering())
    Ap = KratosMultiphysics.mpi.DistributedSystemVector(b.GetNumbering())
    x = KratosMultiphysics.mpi.DistributedSystemVector(b.GetNumbering())
    x.SetValue(0.0)
        
    #r = b - A@x
    r = KratosMultiphysics.mpi.DistributedSystemVector( b ) #copy constructor r = b
    y.SetValue(0.0)
    A.SpMV(x,y) # y = A@x
    r -= y
    zk.GetLocalData()[:] = Minv(numpy.array(r.GetLocalData(), copy=False))
    p = KratosMultiphysics.mpi.DistributedSystemVector( zk ) # p=zk
    rk_dot_zk = r.Dot(zk,0) #note that this is only available on rank0
    rk_dot_zk = b.GetComm().Broadcast(rk_dot_zk,0)
    for i in range(maxiter):
        #Ap=A@p
        Ap.SetValue(0.0)
        A.SpMV(p,Ap) 

        Ap_dot_p = Ap.Dot(p,0)
        Ap_dot_p = b.GetComm().Broadcast(Ap_dot_p,0)

        alpha = rk_dot_zk/(Ap_dot_p)
        x.Add(alpha,p) #x = x + alpha * p;
        r.Add(-alpha,Ap) #r = r - alpha * Ap;
        
        zk1.GetLocalData()[:] = Minv(numpy.array(r.GetLocalData(), copy=False))
        
        r_dot_zk1 = r.Dot(zk1,0)
        r_dot_zk1 = b.GetComm().Broadcast(r_dot_zk1,0)
        
        if(rank == 0):
            print(i,math.sqrt(r_dot_zk1))
        
        if math.sqrt(r_dot_zk1) < tol :
            break
        else:
            p *= r_dot_zk1/rk_dot_zk  
            p += zk1  
            rk_dot_zk=r_dot_zk1

    if(rank == 0)      :
        if(i == maxiter):
            print("error: reached the max number of iterations")
        print("final iterations = ", i)
        print("final norm = ",math.sqrt(r_dot_zk1))

    return x 
~~~

next step (aside for adding distributed matrices to the core) will be to add capabilities needed for deflation solver

loumalouomega (dismissed):

kratos/python/add_vector_to_python.cpp
```diff
@@ -81,6 +81,28 @@ namespace Python
                 self[start] = value[i]; start += step;
             }
         });
+        binder.def(py::init( [](py::buffer b){
+          py::buffer_info info = b.request();
+          KRATOS_ERROR_IF( info.format != py::format_descriptor<typename TVectorType::value_type >::value ) << "Expected a double array\n";
```
philbucher: the error message might be confusing if used with a different type

loumalouomega (dismissed):

philbucher (dismissed):
Can you add your solver as a test?
I think this would be a good integration test

kratos/python/add_vector_to_python.cpp
```diff
@@ -35,7 +35,7 @@ namespace Python
 
     template< typename TVectorType > py::class_< TVectorType > CreateVectorInterface(pybind11::module& m, std::string Name )
     {
-        py::class_< TVectorType, std::shared_ptr<TVectorType> > binder(m,Name.c_str());
+        py::class_< TVectorType, std::shared_ptr<TVectorType> > binder(m,Name.c_str(), py::buffer_protocol());
```
philbucher: hm this is basically what I meant in the other PR with the `EquationIdVector` (where passing by reference did not work yet)
I think this is a larger change than we think

RiccardoRossi: do you see it as problematic?

RiccardoRossi: personally i would see as more problematic the possibility of automatically converting between vectors and matices (we have the same for matrices)

philbucher: I am positive abt this change, makes sense to me
My only concern is the change in behavior which seems to cause the CI failures.
Is this strongly linked to this PR or would it make sense to make a separate one, such that the CSR related changes can go in faster?

RiccardoRossi: the point is about vector assignement

i can do without this by the use of a Set function, howwever this one is needed to write things on the line of 

x[:] = other_data

which is something often needed in solvers

roigcarlo (approved):
:+1:

PR #9675, opened 2022-02-17
[Core] NurbsSurface degree elevation

rickyaristio:
**📝 Description**
This PR adds the degree elevation to the Nurbs surface refinement utility (adopted from Algorithm A5.10, Piegl 1995) and according tests.


rickyaristio:
> 

Thanks for the hint!

applications/IgaApplication/custom_modelers/refinement_modeler.cpp
```diff
@@ -172,6 +172,90 @@ namespace Kratos
                             << " by inserting knots in v, however insert_nb_per_span_v is set to 0." << std::endl;
                     }
                 }
+                if (rParameters["parameters"].Has("insert_degree_u")) {
```
tteschemacher: maybe the string should be "change_degree" if we set it to a specific value and "increase_degree" if increased by a factor? Second is a bit theoretical though...

rickyaristio: It's the second one since we increased it by a factor. I've changed it accordingly.

tteschemacher: ok great. thanks!

tteschemacher:
I think the CI is failing now because of one of those evilish changes from int to size type where the initial int was negative.

applications/IgaApplication/custom_modelers/refinement_modeler.cpp
```diff
@@ -172,6 +172,88 @@ namespace Kratos
...
+                            KnotsURefined, p_nurbs_surface->KnotsV(),
+                            WeightsRefined);
+                    } else {
+                        KRATOS_INFO_IF("::[RefinementModeler]::ApplyRefinement", mEchoLevel > 1)
+                            << "Trying to refine nurbs surface #" << p_nurbs_surface->Id()
+                            << " by elevating degree in u, however increase_degree_u is set to 0." << std::endl;
+                    }
+                }
+                if (rParameters["parameters"].Has("increase_degree_v")) {
+                    SizeType degree_v = rParameters["parameters"]["increase_degree_v"].GetInt();
```
tteschemacher: I think this variable name could be also confusing

tteschemacher: could be increase_degree_v

applications/IgaApplication/custom_modelers/refinement_modeler.cpp
```diff
@@ -172,6 +172,88 @@ namespace Kratos
                             << " by inserting knots in v, however insert_nb_per_span_v is set to 0." << std::endl;
                     }
                 }
+                if (rParameters["parameters"].Has("increase_degree_u")) {
```
tteschemacher: I think the whole increase polynomial degree section should be before the insert knots. I believe this way is more efficient.

applications/IgaApplication/custom_modelers/refinement_modeler.cpp
```diff
@@ -172,6 +172,88 @@ namespace Kratos
...
+                            PointsRefined, KnotsVRefined, WeightsRefined);
+
+                        // Recreate nodes in model part to ensure correct assignment of dofs
+                        IndexType node_id = (r_model_part.NodesEnd() - 1)->Id() + 1;
+                        for (IndexType i = 0; i < PointsRefined.size(); ++i) {
+                            if (PointsRefined(i)->Id() == 0) {
+                                PointsRefined(i) = r_model_part.CreateNewNode(node_id, PointsRefined[i][0], PointsRefined[i][1], PointsRefined[i][2]);
+                                node_id++;
+                            }
+                        }
```
tteschemacher: we need to setup an avenue to only add this in the end. In this way we increase the node id significantly without using the respective nodes...

tteschemacher: but that is for the future.

tteschemacher (commented):
generally, it looks good to me. Just some minor comments.

applications/IgaApplication/custom_modelers/refinement_modeler.cpp
rickyaristio: thanks for pointing this out

applications/IgaApplication/custom_modelers/refinement_modeler.cpp
rickyaristio: I have an idea about this which I'll implement in another PR.

applications/IgaApplication/custom_modelers/refinement_modeler.cpp
rickyaristio: Totally agree. We'll go with p-h-refinement (k-refinement) this way, which leads to smoother continuity.

rickyaristio:
@tteschemacher, do you still have more comments?

tteschemacher (approved):
It looks really good to me! Thanks a lot!

PR #9677, opened 2022-02-18
[ROM] Allowed for arbitrary Dof arrangement in elements

EduardGomezEscandell:
**📝 Description**
@SADPR and I found the source of a segmentation fault in the rom builder and solver testing ROM with the CompressiblePotentialFlowApplication.

As of now, it is assumed that the dofset from an element is oredered as such:
(exemplified with 1D fluid with velocity and pressure)
```
GetDofSet() -> [u1, v1, p1, u2, v2, p2]
                ^~~~~~~~~~~ ^~~~~~~~~~
                 Node 1      Node 2
```
If the ordering was different, a segmentation fault occurred in `GetPhiElemental`. Now any arrangement works. For example:
```
GetDofSet() -> [u1, u2, v1, v2, p1, p2]
```
This is necessary to support all elements, for instance those in the potential flow application.

**🆕 Changelog**
- Fixed `GetPhiElemental` to allow for arbitrary ordering of dofs.
- Made `GetPhiElemental` static and moved it to `custom_utilities/rom_auxiliary_utilities.cpp`

Regrading unit tests, I'm in a bit of a deadlock. To do a potential flow unit test, I need https://github.com/KratosMultiphysics/Kratos/pull/9678, but in order for it to work, I also need the current PR to be merged, so both PR depend on one another. The  unit test is created on the other PR, so go check there to see if you want to see it.

### DIFF
--- a/applications/RomApplication/custom_strategies/rom_builder_and_solver.h
+++ b/applications/RomApplication/custom_strategies/rom_builder_and_solver.h
@@ -27,6 +27,7 @@
 
 /* Application includes */
 #include "rom_application_variables.h"
+#include "custom_utilities/rom_auxiliary_utilities.h"
 
 namespace Kratos
 {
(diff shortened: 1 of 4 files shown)


SADPR:
This must also be changed on "applications/RomApplication/custom_utilities/rom_residuals_utility.h".

Rbravo555 (approved):
LGTM

PR #9678, opened 2022-02-18
[ROM & PotentialFlow] Created _GetBuilderAndSolver in potential flow analysis

EduardGomezEscandell:
**📝 Description**
The rom solver assumes that its superclass class uses _CreateBuilderAndSolver to override it. The potential flow solver was not calling this function, and instead used the default builder and solver. This meant that you could  not actually run a ROM and it was running a FOM under the hood.

**🆕 Changelog**
- Changed `potential_flow_solver.py` so that it calls `_CreateBuilderAndSolver`
- Created unit test in Rom App


applications/RomApplication/tests/test_compressible_potiential_rom.py
```diff
@@ -0,0 +1,60 @@
+import os
+import types
+
+try:
+    import numpy as np
```
marcnunezc, line 4: I see you stored the test results in .npy format. Do you need numpy only for this or does the application need numpy as well?

applications/RomApplication/tests/compressible_potential_test_files/ProjectParameters.json
```diff
@@ -0,0 +1,53 @@
...
+            "python_module" : "apply_far_field_process",
+            "kratos_module" : "KratosMultiphysics.CompressiblePotentialFlowApplication",
+            "process_name"  : "FarFieldProcess",
+            "Parameters"    : {
+                "model_part_name" : "FluidModelPart.PotentialWallCondition2D_FarField",
+                "angle_of_attack" : 0.0,
+                "mach_infinity"   : 0.03,
+                "speed_of_sound"  : 340.0,
+                "initialize_flow_field": false,
+                "perturbation_field": true
```
marcnunezc, line 44: You are not specifying the elemnt in "formulation", but here you are setting the perturbation field. 

You should either specify a "perturbation_*" element in the formulation or turn off the perturbation_field here.

applications/RomApplication/tests/compressible_potential_test_files/simple_airfoil.mdpa
```diff
@@ -0,0 +1,2810 @@
+Begin ModelPartData
```
marcnunezc, line 1: I see this mdpa is not so big but in the CompressibleApp we have this mdpa:

https://github.com/KratosMultiphysics/Kratos/blob/master/applications/CompressiblePotentialFlowApplication/tests/naca0012_small_mdpa/naca0012_small.mdpa

Which is much smaller, if it's not too much work and it fits your needs you can consider using it instead (I would copy-paste it in the romApp)

marcnunezc (commented):
Ok from my side, minor comments below

applications/RomApplication/tests/test_compressible_potiential_rom.py
EduardGomezEscandell: I mimicked the other tests in the application. I don't know if numpy is needed for the app in general. Maybe @Rbravo555 can tell us.

applications/RomApplication/tests/compressible_potential_test_files/ProjectParameters.json
EduardGomezEscandell: Yes, I forgot to put the formulation in.

applications/RomApplication/tests/compressible_potential_test_files/simple_airfoil.mdpa
EduardGomezEscandell: Yep, good point. Done.

applications/RomApplication/tests/test_compressible_potiential_rom.py
rubenzorrilla: It is required.

marcnunezc (approved):
Ok from my side

rubenzorrilla (approved):
Ready to be merged IMO.

PR #9679, opened 2022-02-18
[Core] Tinyexpr wrapper Depends on space bug

### LINKED ISSUE #9681: Wrong parsing of functions in .json
**Description**
This piece of code is in charge of detecting if a function depends on space.

https://github.com/KratosMultiphysics/Kratos/blob/39cc4fcff426d57bb6d19e859dd3c3be05e6837e/kratos/utilities/function_parser_utility.cpp#L39-L46

this fails for something as simple as `exp(t)`, because it contains an `x`.

**Scope**
When a function is passed e.g. for a boundary condition

**To Reproduce**
Check out https://github.com/KratosMultiphysics/Kratos/pull/9679
This branch contains two unit tests showing the problem
```
(1.0)*(50*(cos(t)-2)) -> DependsOnSpace() : false
(1.0)*(50*(exp(t)-2)) -> DependsOnSpace() : true
```

**Expected behavior**
The tests should pass

**Environment**
- OS: Ubuntu Ubuntu 21.10
- Branch: bug-assign-vector-direction-condition
- Python 3.9.7

EduardGomezEscandell:
**📝 Description**
This PR is to exemplify issue https://github.com/KratosMultiphysics/Kratos/issues/9681

**🆕 Changelog**
- Added tests so show problem


loumalouomega:
What's the problem?

loumalouomega:
I implemented this BTW

EduardGomezEscandell:
> What's the problem?

Sorry I forgot to link the issue
See here https://github.com/KratosMultiphysics/Kratos/issues/9681

loumalouomega (dismissed):

kratos/utilities/function_parser_utility.cpp
```diff
@@ -36,7 +60,7 @@ BasicGenericFunctionUtility::BasicGenericFunctionUtility(const std::string& rFun
     InitializeParser();
 
     // Check if it depends on space
-    if (mFunctionBody.find(std::string("x")) == std::string::npos &&
+    if (CheckThereIsNotx(mFunctionBody)                           &&
         mFunctionBody.find(std::string("y")) == std::string::npos &&
         mFunctionBody.find(std::string("z")) == std::string::npos &&
         mFunctionBody.find(std::string("X")) == std::string::npos &&
```
philbucher, line 71: wouldn't it be safer to check if the character are by themselves? I.e. not surrounded by other characters

loumalouomega: How do you check that is only x?, it is the same as saying not surrounded by other characters. If you find a character, you don't know if it is part of a keyword

philbucher: Depends if we allow e.g. `xy` or enforce `x*y`
not saying I am correct, just giving my thoughts

loumalouomega: IMO this is the easiest way with the least corner cases

kratos/utilities/function_parser_utility.cpp
```diff
@@ -23,6 +23,30 @@
 namespace Kratos
 {
 
+// A check only to check that the x is not part of exp
+inline bool CheckThereIsNotx(const std::string& rString)
```
philbucher, line 28: as it is, it seems to only check for the first x

what abt sth like `exp(2)*x`?

=> maybe you could improve the algorithm by removing all occurrences of `exp` and then checking again
You could even implement it in a way that it removes all fct names before checking (and the fct names to be removed could be extended if necessary)

philbucher: i.e. 
~~~c++
for (const auto& r_keyword : {"exp"}) {
    StringUtilities::ReplaceAllSubstrings(r_keyword, rString, "");
}

~~~

loumalouomega: My code checks all x, see again, the first line only check if thee is at least one x

loumalouomega: > i.e.
> 
> ```c++
> for (const auto& r_keyword : {"exp"}) {
>     StringUtilities::ReplaceAllSubstrings(r_keyword, rString, "");
> }
> ```

I don't want to do that, I have bad experience in the past

loumalouomega: > as it is, it seems to only check for the first x
> 
> what abt sth like `exp(2)*x`?

That will work, check yourself

loumalouomega:
> I could translate it to C++ if necessary. Thoughts?

The problem is from our side, I rather prefer the suggestion from @philbucher

loumalouomega:
> > I could translate it to C++ if necessary. Thoughts?
> 
> The problem is from our side, I rather prefer the suggestion from @philbucher

But I prefer my proposal

miguelmaso:
`sqrt` -> depends on time
I don't know if it's relevant

kratos/utilities/function_parser_utility.cpp
oberbichler: `std::regex_search(expr, std::regex("\\b[xyz]\\b"))`?

loumalouomega: Good idea but does not work, I am not experienced in regex enough to do the correct reguar expression

loumalouomega: image

loumalouomega: Fixed with  regex like a pro, thanks @oberbichler 

image

Today I learned regex

loumalouomega (dismissed):

kratos/utilities/function_parser_utility.cpp
oberbichler: `\b...\b` is similar to "Match Whole Word" in VS Code search function. `[xyz]` matches any of the chars. There is also an option for case-insensitivity. Just wondering why [`\b[xyzXYZ]\b` is not equivalent to "depends on space"](https://regex101.com/r/SzNK98/1) and [`\bt\b` to "depends on time"](https://regex101.com/r/T9TIny/1). The xy=x*y thing?

[...and keep in mind ;)](http://regex.info/blog/2006-09-15/247)

loumalouomega: With the regex looks like it fails, do not know why

loumalouomega: I reverted the regex, I am unable to translate to C++ a proper regex code to work:  https://regex101.com/r/YYDMKo/1 , this works, but once I translate to C++ it does not work

EduardGomezEscandell: This is quite strange because the same code seems to compile and run just fine in isolation:

https://godbolt.org/z/r993dbqxe

loumalouomega: The problem is not that, the problem is that "x" and "2*x" does not work, so you need an additional check, that in C++ fails, but in JS not

oberbichler: I thought that it might be enough to search for `xyz` or `t` (instead of checking for non-"exp"/non-"sqrt")

https://godbolt.org/z/hc41PYr7P

but I am probably missing some corner cases(?)

loumalouomega: I tried that the other day, and it didn't work, I will try again

loumalouomega: According to this should work now: https://regex101.com/r/upucEk/1

loumalouomega (dismissed):

oberbichler:
I'm not sure but Centos probably doesn't work because of gcc 4.8. regex support was introduced in 2014 with gcc 4.9. Too bad for the one-liner 🤷

Maybe revert to the working solution and add a comment for future development. The AST solution is also very nice but I guess more work.

loumalouomega:
> I'm not sure but Centos probably doesn't work because of gcc 4.8. regex support was introduced in 2014 with gcc 4.9. Too bad for the one-liner shrug
> 
> Maybe revert to the working solution and add a comment for future development. The AST solution is also very nice but I guess more work.

Effectively...

loumalouomega (dismissed):
Ready to merge

kratos/utilities/function_parser_utility.cpp
```diff
@@ -12,6 +12,9 @@
 //
 
 // System includes
+#ifdef KRATOS_MODERN_REGEX
+    #include <regex>
+#endif
```
philbucher: why is this needed?

loumalouomega: Because centos... (minimal version of gcc required for regex is 4.9, centos uses 4.8)

philbucher:
```suggestion
#if defined __has_include
    if __has_include(<regex>)
        #define KRATOS_MODERN_REGEX
    #endif
#endif

#ifdef KRATOS_MODERN_REGEX
    #include <regex>
#endif
```

this way would be more modern and automatic
or sth in that direction

philbucher: `__has_include` is officially C++17, but already supported by most compilers afaik
https://en.cppreference.com/w/cpp/preprocessor/include

loumalouomega: Is not a problem of compilation, is a problem of full support (apparently in 4.8, the regex support is partial)

philbucher: ah ok
hm maybe then the following:
```suggestion
#if !(defined(__GNUC__) && !defined(__clang__) && (__GNUC__ < 4 || (__GNUC__ == 4 && (__GNUC_MINOR__ < 9))))
    #include <regex>
#endif
```

from here:
https://github.com/KratosMultiphysics/Kratos/blob/9ea10775243e921f8caaa115eaca4a235ab9cb4d/applications/HDF5Application/custom_io/hdf5_file.cpp#L22

loumalouomega: So, you want to avoid the KRATOS_MODERN_REGEX macro?

philbucher: yes

loumalouomega: Do I also remove it?

kratos/utilities/function_parser_utility.cpp
```diff
@@ -36,14 +67,19 @@ BasicGenericFunctionUtility::BasicGenericFunctionUtility(const std::string& rFun
     InitializeParser();
 
     // Check if it depends on space
-    if (mFunctionBody.find(std::string("x")) == std::string::npos &&
+#ifdef KRATOS_MODERN_REGEX
```
loumalouomega:
```suggestion
#if !(defined(__GNUC__) && !defined(__clang__) && (__GNUC__ < 4 || (__GNUC__ == 4 && (__GNUC_MINOR__ < 9))))
```

loumalouomega (dismissed):

loumalouomega (approved):

philbucher (approved):

loumalouomega:
@EduardGomezEscandell feel free to merge

PR #9680, opened 2022-02-18
[Core][GiDIO] fix GP initialization

### LINKED ISSUE #3234: [ShapeOpt] Seg-fault when not initializing existing elements before GidIO is initialized
The problem is as follows:

Given an optimization model part which contains elements. Gid is specified as output format. If the elements of the optimization model part are not initialized (which may happen e.g. when no internal analyzer is defined), then the function GiDOutputProcess::ExecuteBeforeSolutionLoop will produce a seg-fault when it calls   

`__initialize_results --> gid_io.h line 643 inside the function InitializeResults. -> AddElement in gid_gauss_point_container.h line 72. `

The reason is that for a not initialized element, the function pElemIt->GetIntegrationMethod()  returns a random number causing a failure of 

`pElemIt->GetGeometry().IntegrationPoints(->GetIntegrationMethod() )`

That means as a consequence, all elements must be initialized before the gid output process is initialized. Right now, this is done by the analyzer through InitializeBeforeOptimizationLoop(), which internally calls the physical solver that does the job. However, this creates an unwanted link between the data logger and the analyzer. Also, it determined the call order. I.e. initializing the analyzer after the gid output process will again create a segfault. 

In either case the seg-fault can be avoided by writing only the design surface (whicho only has conditions without gauss points). As soon as the complete optimization model part shall be written, the error happens again.

As currently the call order in each algorithm is such that no seg-fault is produced, there is no immediate action required. But in the future, this unwanted behavior needs to be removed. This leads to two tasks:

- [ ] Better define output of the optimization model part so that a clean model with only the optimization results is written (e.g. no Gauss point results are specifiable)

- [ ] Make sure that there is no such implicit dependency of the data logger from the (internal) analyzer. Maybe initialize possibly given elements with default settings.

### LINKED ISSUE #9671: Segmentation Fault in Structural Mechanics Application + GiD output process
**Description**
This is very much an edge-case.

If you try to use a `BaseSolidElement` or derived class without calling `SetIntegrationMethod`, and later on try to use the `gid_output_process`, a segmentation fault occurs here:

https://github.com/KratosMultiphysics/Kratos/blob/46be0c8b2716663fbd495d95b3c5a71ac1170eaa/kratos/includes/gid_gauss_point_container.h#L70-L72

Particularly in the second condition
```
pElemIt->GetGeometry().IntegrationPoints(pElemIt->GetIntegrationMethod()).size() == mSize
                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                                    UNINITIALIZED VALUE
```
This is of course undefined behaviour and more often than not will cause a segmentation fault.

I understand that during normal operation of the application, this `SetIntegrationMethod` is always called at some point, but I was working in ROM where I only use this element because it's the one coming from the MDPA.

**Scope**
- KratosCore: Gid output
- StructuralMechanicsApplication @KratosMultiphysics/structural-mechanics 

**To Reproduce**
Switch to branch `rom/visualize-modes`.
Run unit test at `/applications/RomApplication/tests/test_basis_visualization_analysis.py`
Particularly the second one

**Expected behavior**
Probably set a default value for the integration method. I've never used this application so I'm not sure either.

**Environment**
- OS: Ubuntu 21.10
- Branch: rom/visualize-modes
- Python 3.9.7
Note that this branch doesn't change any of the afected files.

philbucher:
The GiDIO always initializes the GaussPoint containers, unconditionally if the user requested any GP-results.
In structural elements this causes problems with uninitialized elements as then the preparation of the GP-results can fail
Hence I am making the writing of GP-results optional, but keeping the default behavior

closes #3234
closes #9671

### DIFF
--- a/kratos/tests/auxiliar_files_for_python_unittest/reference_files/results_out_ref.ref
+++ b/kratos/tests/auxiliar_files_for_python_unittest/reference_files/results_out_ref.ref
@@ -1,13 +1,4 @@
 GiD Post Results File 1.0
-GaussPoints "tri1_element_gp" ElemType Triangle
-Number Of Gauss Points: 1
-Natural Coordinates: Internal
-End GaussPoints
-GaussPoints "lin1_element_gp" ElemType Linear
-Number Of Gauss Points: 1
-  Nodes not included
-Natural Coordinates: Internal
-End GaussPoints
 Result "DISPLACEMENT" "Kratos" 0 Vector OnNodes
 Values
 1 0.1 0 0
--- a/kratos/includes/gid_io.h
+++ b/kratos/includes/gid_io.h
@@ -147,7 +147,8 @@ public:
         const GiD_PostMode Mode,
         const MultiFileFlag UseMultipleFilesFlag,
         const WriteDeformedMeshFlag WriteDeformedFlag,
-        const WriteConditionsFlag WriteConditions
+        const WriteConditionsFlag WriteConditions,
+        const bool InitializeGaussPointContainers=true
          ) : mResultFileName(rDatafilename),
         mMeshFileName(rDatafilename),
         mWriteDeformed(WriteDeformedFlag),
@@ -160,7 +161,9 @@ public:
 
         InitializeResultFile(mResultFileName);
         SetUpMeshContainers();
-        SetUpGaussPointContainers();
+        if (InitializeGaussPointContainers) {
+            SetUpGaussPointContainers();
+        }
 
         GidIOBase& r_gid_io_base = GidIOBase::GetInstance();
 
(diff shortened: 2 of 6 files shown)


kratos/tests/auxiliar_files_for_python_unittest/reference_files/results_out_ref.ref
```diff
@@ -1,13 +1,4 @@
 GiD Post Results File 1.0
-GaussPoints "tri1_element_gp" ElemType Triangle
```
philbucher, line 2: I checked this in GiD, works in the same way as before

EduardGomezEscandell (approved):

PR #9686, opened 2022-02-18
[Shallow] hotfix - Avoid division by zero

miguelmaso:
**📝 Description**
a leftover in depth integration process

**🆕 Changelog**
- Avoid division by zero
- Aesthetic changes in output process

### DIFF
--- a/applications/ShallowWaterApplication/custom_processes/depth_integration_process.cpp
+++ b/applications/ShallowWaterApplication/custom_processes/depth_integration_process.cpp
@@ -159,6 +159,9 @@ void DepthIntegrationProcess<TDim>::Integrate(
             velocity = InterpolateVelocity(p_elem, rShapeFunctionsValues);
         }
     }
+    if (height < 1e-4) { // Sanity check to avoid Nan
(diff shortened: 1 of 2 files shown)


ipouplana (approved):
Good

applications/ShallowWaterApplication/custom_processes/depth_integration_process.cpp
```diff
@@ -159,6 +159,9 @@ void DepthIntegrationProcess<TDim>::Integrate(
             velocity = InterpolateVelocity(p_elem, rShapeFunctionsValues);
         }
     }
+    if (height < 1e-4) { // Sanity check to avoid Nan
```
rubenzorrilla, line 162: Not sure about the physics involved, but I'd use a smaller value for the zero check (of the order of 1.0e-12).

miguelmaso: in this case I prefer to be conservative. 1e-4 is 0.1mm, for a river, flooding event or wave runup this small enough.

PR #9559, opened 2022-01-19
[Fluid] Improvements in compressible NS symbolic generator

EduardGomezEscandell:
**📝 Description**
A refactor that for some reason yields better results. Probably some error was corrected inadvertently.

**🆕 Changelog**
- Refactored compressible NS symbolic generator.

**📈 Improvement**
The improvement is shown in the following set of images.

The figure is a 1D cross section the solution to the 2D Sod tube with a RK3-TVD time-scheme and autmatic stepping with a triangular mesh.
full

The second picture is a zoom of the center of the domain.
detail

It shows a minor improvement on the undershoot.

PR #9694, opened 2022-02-22
[Fluid][Compressible] Alternative way of computing derivatives with respect to dofs

EduardGomezEscandell:
**📝 Description**
Changing the way symbolic derivatives are taken.
```
d/dU[E(U)] = d/dV[E(V(U))] * d/dU[V(U)]
```
This is to solve the fact that linearizing `dV/dU` is invalid (as done in https://github.com/KratosMultiphysics/Kratos/pull/9559).

**🆕 Changelog**
- Many changes relating to this conversion in compressible NS symbolic generator

So far does not generate the same output as it should.


EduardGomezEscandell:
Closing due to lack of results

PR #9682, opened 2022-02-18
MPI changes fort trilinos and containers

roigcarlo:
**📝 Description**
Extending #9644:
- Trilinos become an optional component (unless compiling trilinos)
- Created a docker to generate wheels using ABI compatible mpi for linux (through mpich, no 100% guarantees it will work on all systems but should be pretty reliable)

Since the detection of mpi is done through the existance of the mpi folder in the module, I've just changes the way the wheelbuilder works and by default it tries to copy everything installed in the KratosMultiphysics folder in the install dir.

This technically overrides the "included_modules" section in the core configuration json. Once I find a better solution I will change it.

**🆕 Changelog**
- Making Trilinos a optional component in MPI compilations
- Created Docker builder for MPI-ready wheels for linux.

PR #9695, opened 2022-02-23
Cleaning up mpi wheelbuilder

roigcarlo:
**📝 Description**
A bunch of changes following up #9682

**🆕 Changelog**
- CoSim is temporally disabled from wheels until I resolve the problems with the includes (should take 1-2 days at most)
- Added option to enable the usage of LTO options from pybind11 in compilers that support it and added some initial support for Kratos. It is tested in GCC 11/Clang 11. I am aware of problems in Gcc 8 and below. Take it a a groundwork for now.
- Dropping the "DEV" tag thing in the versioning. From now on it will be always a number...
- Mpi is now included a a submodule in the core
- KratosKernel and KratosPythonInterface custom projects are created in a place so that allows nested subfolders (and not only first level subfolders) to add their targets. This forced the core to be compiled separately for every interface version.
- Re-enabling all python interfaces in the default script.
- Replaced the mpicc and mpicxx wrappers for the standard compiled because some wrappers come with flags like `-g` in production which causes large binaries to be generated.


scripts/wheels/linux/configure.sh
```diff
@@ -72,4 +72,5 @@ ${CMAKE} -H"${KRATOS_SOURCE}" -B"${KRATOS_BUILD}/${KRATOS_BUILD_TYPE}" \
 -DBLAS_LIBRARIES="/usr/lib64/libblas.so.3"                             \
 -DINCLUDE_MMG=ON                                                       \
 -DMMG_ROOT="/workspace/external_libraries/mmg/mmg_5_4_1"               \
+-DKRATOS_BUILD_TESTING=OFF                                             \
```
philbucher, line 75: this should probably be the default for release builds

philbucher (approved):
:+1:

scripts/wheels/linux/configure.sh
roigcarlo: yep. it is now :)

PR #9696, opened 2022-02-23
Disable C++ tests by default in Release

philbucher:
The C++ tests take a significant part of the compilation time. Oftentimes in the past they were compiled but never used, especially when compiling in Release (i.e. when one wants to run simulations and not develop)

Hence I propose to disable them by default in when compiling in release
If one is developing, then ofc they can still be enabled


CMakeLists.txt
```diff
@@ -113,6 +107,12 @@ if(NOT CMAKE_BUILD_TYPE)
 endif(NOT CMAKE_BUILD_TYPE)
 set (KratosMultiphysics_BUILD_TYPE ${CMAKE_BUILD_TYPE})
 
+# If no test policy enable by default (only in non-release builds)
+if(NOT DEFINED KRATOS_BUILD_TESTING AND NOT KratosMultiphysics_BUILD_TYPE STREQUAL "Release")
+message("-- No KRATOS_BUILD_TESTING is defined. Setting to ON")
+SET (KRATOS_BUILD_TESTING ON)
```
roigcarlo: @philbucher shouldn't this be set to OFF if you want to enable them only for the CI?

roigcarlo (commented):
I tend to agree with @philbucher with this. Having the CI makes enabling the tests by default no sense. 

I would enable them in the template scripts for users anyway so they at least know that exist.

CMakeLists.txt
philbucher: ah you would set it to `OFF` by default, also e.g. for `FullDebug` builds?
Fine for me, makes sense

philbucher:
> I would enable them in the template scripts for users anyway so they at least know that exist.

I would add it but set it to `OFF`, as most first time users have no idea abt the C++ tests (and they do take A LOT of time)

roigcarlo (dismissed):
On behalf of the @KratosMultiphysics/technical-committee  👍

CMakeLists.txt
philbucher: what about this comment @roigcarlo ?
Should I disable it always, i.e. independent of build type?

roigcarlo: Yep, let's keep it consistent

philbucher: done

roigcarlo (dismissed):

roigcarlo (changes requested):
I spoke with @RiccardoRossi and wants to discuss this in the @KratosMultiphysics/technical-committee

RiccardoRossi:
in essence, my point is that we should make easy to switch off the tests ... but i think it gives an incorrect message to have them off by default.

it is philosphical, but i think we should have a voting about this....

philbucher:
IMO this is a users vs devs issue
My reasoning is to disable them by default as they take almost as much time as the rest of the code, and it already has been remarked by external sources that Kratos takes way too long to compile

Since we have them in the CI we are anyway on the safe side regarding breaking stuff
Devs that need them also know how to enable them

loumalouomega:
> IMO this is a users vs devs issue My reasoning is to disable them by default as they take almost as much time as the rest of the code, and it already has been remarked by external sources that Kratos takes way too long to compile

Not only takes to long to compile, if we package them in the wheels these will be huge and at the end no one will care

roigcarlo:
Tests are not compiled during the generation of wheels. The configures for that are tweaked. 

As for the PR itself I tend to go more for the user experience, but as a developer I am also concerned people will not activate tests unless is done by default, so I am 60-40 biasesd towards leaving them disabled.

philbucher:
I think by now everyone relies on the CI anyway for testing, who compiles entire Kratos when modifying sth to check if tests are working? Be honest :laughing: 

if you need the C++ tests for developing, then you can enable them. If not then whats the point of compiling them?

loumalouomega:
> I think by now everyone relies on the CI anyway for testing, who compiles entire Kratos when modifying sth to check if tests are working? Be honest 😆
> 
> if you need the C++ tests for developing, then you can enable them. If not then whats the point of compiling them?

I am with @philbucher. Did you try to compile VTK with tests?, it takes forever. It is deactivated by default.

PETSc only compiles a very small list of C tests by default (but it is a C library, doesn't have a python interface)

PS: tests are super cool because you see a cow in 3D ([in 3D!!!!](https://www.youtube.com/watch?v=MOUjJbhdOZA))

RiccardoRossi:
well, if you are a developer you are setting your own cmake. Once we properly documented, a developer can switch off the tests with a single line. I don't think we need to make it even easier than that.

Having said this, i would seriously prefer that unexperienced users do compile everything, as compiling tests helps in detecting incompatibilities.

philbucher:

roigcarlo:
After long deliberation, @KratosMultiphysics/technical-committee decided not to disable tests by default.

In general developers should never turn off tests, or  in any case they already know how to do it.
If a user wants a plugn-and-play approach they have the pip packages. 

For every other cases:
- Novel users that never compiled Kratos
- Normal Developers
- CI
- Release compilations

Tests are encouraged.

PR #9700, opened 2022-02-23
[Core] MaxReduction is failing in empty container

ddiezrod:
**📝 Description**
As you can see in the added failing test, if the container is empty the max reduction is not working properly. I think the expected output of the test should be zero. This MR is just to show the problem, my idea is to add the fix in this MR also.

### DIFF
--- a/kratos/tests/cpp_tests/utilities/test_parallel_utilities.cpp
+++ b/kratos/tests/cpp_tests/utilities/test_parallel_utilities.cpp
@@ -540,6 +540,18 @@ KRATOS_TEST_CASE_IN_SUITE(OmpVsPureC11, KratosCoreFastSuite)
 
 }
 
+KRATOS_TEST_CASE_IN_SUITE(MaxReductionInEmptyVector, KratosCoreFastSuite)
+{
+    std::vector<double> data_vector;
+    double max_value = 0.0;
+
+    max_value = block_for_each<MaxReduction<double>>(data_vector,[&](double& value){
+        return value;
+    });
+
+    KRATOS_CHECK_NEAR(max_value, 0.0, 1e-10);
+}
+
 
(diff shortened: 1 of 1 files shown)


ddiezrod:
I see the problem is this line: https://github.com/KratosMultiphysics/Kratos/blob/3dd7267a47024529aedc70eedd80c463a9186cee/kratos/utilities/reduction_utilities.h#L165
I do not see any easy solution, maybe we could just live with this, although I guess its not really the expected outcome of the test. (Especially when the variable was initialized to zero)

rubenzorrilla:
Maybe it is possible to first check if the container size is zero in the `block_for_each` definition. If so it is, maybe we can call the default constructor of the `value_type` of the reducer to get the corresponding zero value (we'd need to check the behavior of the default constructors).

EduardGomezEscandell:
I believe the current implementation is best. An operation without arguments should, in my opinion, return the identity element of that operation. For scalars:
- Addition and subtraction should return 0.
- Multiplication, if it was implemented, should return 1.
- Max should return -∞, or `numeric_limits::lowest()` if we prefer avoiding infs.
- Min should return +∞, or `numeric_limits::max()` if we prefer avoiding infs.

This way one can safely chain operations. For instance, finding the max in with both MPI and OMP enabled.

RiccardoRossi:
well...what is the expected value 0f the maximum for an empty array?

ddiezrod:
I agree that its not as obvious as I initially thought, as 0.0 could be a quite arbitrary value to return. I am more worried from the practical point of view, I think many developers would expect that at least if there are no elements in the container, the original value is maintained (exactly what is tested in this MR), but I understand it might not worth it. In any case, I hope that this MR at least warns developers about using max/min in empty containers.

Issue #9689, opened 2022-02-21
Strange bug particular to my machine with tinyexpr function parsing


EduardGomezEscandell:
**Description**
For some reason tinyexpr fails to parse this expression when running RomApplication tests:
```
(0.0)*(50*(exp(t)-1))
```
I've taken some debugging steps that only leave me more confused:
- I've tried it in isolation (just writting a program in C to call tinyexpr) and there is no issue.
- The test passes in my collegues' machines
- The test passes in the CI run
- The test passes if I log onto my machine via ssh form another computer
- I wrote a unit test with this same example and it passes (https://github.com/KratosMultiphysics/Kratos/pull/9688)
- The error occurs both in Release and Debug compilations
- I re-compiled Kratos from scratch and the error persists.

The error is thrown in line 159:
https://github.com/KratosMultiphysics/Kratos/blob/0a14fc33b9587e87df9c9e0f5f28ec1115419bec/kratos/utilities/function_parser_utility.cpp#L157-L160

**Scope**
RomApplication

**To Reproduce**
So far it's only happened in my machine. See Environment for details.
```
cd applications/RomApplication/tests
python3 test_structural_rom.py 
```

**Expected behavior**
The test should pass

**Environment**
- OS: Ubuntu 21.10
- Branch: master
- Python 3.9.7
- Compiled attempted with both:
    -  g++ (Ubuntu 11.2.0-7ubuntu2) 11.2.0
    - Ubuntu clang version 13.0.0-2

**Console output**
```
egomez@PCCB199:~/Work/Kratos/applications/RomApplication/tests$ python3 test_structural_rom.py 
 |  /           |
 ' /   __| _` | __|  _ \   __|
 . \  |   (   | |   (   |\__ \
_|\_\_|  \__,_|\__|\___/ ____/
           Multi-Physics 9.0."Dev"-dc47ff578d-Debug
Compiled with threading support.
Maximum number of threads: 16.
Process Id: 87369 
Importing    KratosRomApplication 
Initializing KratosRomApplication...
Importing    KratosStructuralMechanicsApplication 
    KRATOS   ___|  |                   |                   |
           \___ \  __|  __| |   |  __| __| |   |  __| _` | |
                 | |   |    |   | (    |   |   | |   (   | |
           _____/ \__|_|   \__,_|\___|\__|\__,_|_|  \__,_|_| MECHANICS
Initializing KratosStructuralMechanicsApplication...
[...]

PR #9688, opened 2022-02-21
[Core] Example for issue #9689

EduardGomezEscandell:
**📝 Description**
Error exemplifying Issue https://github.com/KratosMultiphysics/Kratos/issues/9689

**🆕 Changelog**
- Created unit test

This PR is not meant to be merged, it's only to exemplify the linked issue.

Issue #9684, opened 2022-02-18
[NeuralNetworkApplication] Loading PredictFromModelProcess creates conflict with parsing utilities


danielandresarcones:
**Description**
The NeuralNetworkApplication implements a process that loads an external neural network model to be used in the solver. If this process is initialized after another that parses an expression from a string (e.g. assign a variable from an expression), the parsing fails from that moment. When the process that includes the parsing is called, a "Kratos::Exception" is called and terminate is called recursively. The exception is catched exactly in the parsing of the expression. This doesn't occur if the neural network model is loaded in the solver or the analysis, only if it is called from the process. The process is predict_from_model_process.py and the initialization should not affect the parsing:
https://github.com/KratosMultiphysics/Kratos/blob/e7ed680545cea3ea299c0538f728a32be313d69a/applications/NeuralNetworkApplication/python_scripts/predict_from_model_process.py#L16-L35
This issue only happens with this specific process, which has no relation with the error. Right now it can be avoided by evaluating the model in the solver, but this will be a problem in the future. Any help or comments are welcomed.

**Scope**
- NeuralNetworkApplication

**To Reproduce**
1. Switch to branch`test-neuralnetwork-application`.
2. Compile the required applications (StructuralMechanicsApplication, FluidDynamicsApplication, MeshMovingApplication, LinearSolversApplication, CoSimulationApplication and NeuralNetworkApplication)
3. Keras and Tensorflow are needed.
4. Download the example from the [Examples repository](https://github.com/KratosMultiphysics/Examples/tree/test-neuralnetwork-application/neural_network/cosim/mok_fsi/surrogate_structure_legacy).
5. Run `run.sh`.
6. For successive runs, `run_without_training.sh` can be run instead.

**Expected behavior**
The error should not occur. An example of a working code achieved by not initializing the process can be found in [surrogate_fluid](https://github.com/KratosMultiphysics/Examples/tree/test-neuralnetwork-application/neural_network/cosim/mok_fsi/surrogate_fluid).

**Environment**
- OS: Ubuntu 20.04
- Branch: test-neuralnetwork-application
- Python 3.8.10
- Tensorflow 2.7.0
- Keras 2.7.0

philbucher:
I am suspecting some weird behavior with the library we are using for a while
I am not sure what is the reason for those strange interactions, but I also encountered them :/
See #9689

EduardGomezEscandell:
[...]

PR #9702, opened 2022-02-23
[Core] improve error messages in function parser

philbucher:
Inspired by #9688 I added improved error messages which should help the user understand what is wrong in their code

Example:
Old error message:
~~~
Parsing error in function: (0)*(50*(expp(t)-1))
~~~

new error message:
~~~
Parsing error in function: (0)*(50*(expp(t)-1))
Error occurred near here :             ^ (char [12])
Check your locale (e.g. if "." or "," is used as decimal point)
~~~

related: #9689, #9684

FYI I also played around with the locale of C++, but without success, i.e. I could not add more useful information

### DIFF
--- a/kratos/tests/cpp_tests/utilities/test_function_parser_utility.cpp
+++ b/kratos/tests/cpp_tests/utilities/test_function_parser_utility.cpp
@@ -67,7 +67,7 @@ KRATOS_TEST_CASE_IN_SUITE(GenericFunctionUtility1, KratosCoreFastSuite)
     KRATOS_CHECK_STRING_EQUAL(function6.FunctionBody(), "(1.0)*(50*(exp(t)-2))");
     KRATOS_CHECK_DOUBLE_EQUAL(function6.CallFunction(0.0,0.0,0.0,0.0), -50);
 
-    KRATOS_CHECK_EXCEPTION_IS_THROWN(GenericFunctionUtility("A quien"), "Error: \nParsing error in function: A quien\nError occurred near here : ^ (char [0])\nCheck your locale (e.g. if \".\" or \",\" is used as decimal point)");
+    KRATOS_CHECK_EXCEPTION_IS_THROWN(GenericFunctionUtility("NotAValidExpression"), "Error: \nParsing error in function: NotAValidExpression\nError occurred near here : ^ (char [0])\nCheck your locale (e.g. if \".\" or \",\" is used as decimal point)");
 }
 
 KRATOS_TEST_CASE_IN_SUITE(GenericFunctionUtility2, KratosCoreFastSuite)
--- a/kratos/tests/cpp_tests/utilities/test_function_parser_utility.cpp
+++ b/kratos/tests/cpp_tests/utilities/test_function_parser_utility.cpp
@@ -67,7 +67,7 @@ KRATOS_TEST_CASE_IN_SUITE(GenericFunctionUtility1, KratosCoreFastSuite)
     KRATOS_CHECK_STRING_EQUAL(function6.FunctionBody(), "(1.0)*(50*(exp(t)-2))");
     KRATOS_CHECK_DOUBLE_EQUAL(function6.CallFunction(0.0,0.0,0.0,0.0), -50);
 
-    KRATOS_CHECK_EXCEPTION_IS_THROWN(GenericFunctionUtility("NotAValidExpression"), "Error: \nParsing error in function: NotAValidExpression\nError occurred near here : ^ (char [0])\nCheck your locale (e.g. if \".\" or \",\" is used as decimal point)");
+    KRATOS_CHECK_EXCEPTION_IS_THROWN(GenericFunctionUtility("NotAValidExpression"), "Error: \nParsing error in function: NotAValidExpression\nError occurred near here :                   ^ (char [18])\nCheck your locale (e.g. if \".\" or \",\" is used as decimal point)");
 }
 
 KRATOS_TEST_CASE_IN_SUITE(GenericFunctionUtility2, KratosCoreFastSuite)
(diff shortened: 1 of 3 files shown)


kratos/tests/cpp_tests/utilities/test_function_parser_utility.cpp
```diff
@@ -67,7 +67,7 @@ KRATOS_TEST_CASE_IN_SUITE(GenericFunctionUtility1, KratosCoreFastSuite)
     KRATOS_CHECK_STRING_EQUAL(function6.FunctionBody(), "(1.0)*(50*(exp(t)-2))");
     KRATOS_CHECK_DOUBLE_EQUAL(function6.CallFunction(0.0,0.0,0.0,0.0), -50);
 
-    KRATOS_CHECK_EXCEPTION_IS_THROWN(GenericFunctionUtility("A quien le importa lo que yo haga A quien le importa lo que yo diga  Yo soy asi, y asi seguire, nunca cambiare"), "Error: Parsing error in function: A quien le importa lo que yo haga A quien le importa lo que yo diga  Yo soy asi, y asi seguire, nunca cambiare");
+    KRATOS_CHECK_EXCEPTION_IS_THROWN(GenericFunctionUtility("A quien"), "Error: \nParsing error in function: A quien\nError occurred near here : ^ (char [0])\nCheck your locale (e.g. if \".\" or \",\" is used as decimal point)");
```
loumalouomega: My beautiful musical reference!!!

https://www.youtube.com/watch?v=XX_hWpPnd3I

loumalouomega:
```suggestion
    KRATOS_CHECK_EXCEPTION_IS_THROWN(GenericFunctionUtility("A quien le importa lo que yo haga A quien le importa lo que yo diga Yo soy asi, y asi seguire, nunca cambiare"), "Error: \nParsing error in function: A quien\nError occurred near here : ^ (char [0])\nCheck your locale (e.g. if \".\" or \",\" is used as decimal point)");
```

loumalouomega (changes requested):
Can we keep my musical reference?

kratos/tests/cpp_tests/utilities/test_function_parser_utility.cpp
philbucher: I changed it bcs it becomes quite nasty to debug :sweat_smile: (and doesnt matter for the check)
For me it is random letters anyway :D

EduardGomezEscandell:
```suggestion
    KRATOS_CHECK_EXCEPTION_IS_THROWN(GenericFunctionUtility("NotAValidExpression"), "Error: \nParsing error in function: NotAValidExpression\nError occurred near here : ^ (char [0])\nCheck your locale (e.g. if \".\" or \",\" is used as decimal point)");
```

Might as well make the intention of this string obvious, and get rid of non-english text.

kratos/tests/test_function_parser_utility.py
```diff
@@ -169,6 +169,11 @@ def test_ApplyFunctionToNodesUtilityTimeEvolutionCTernaryFail(self):
         with self.assertRaisesRegex(Exception, 'Parsing error in function: 1.5 if t<2.0 3.0 if defined, but not else'):
             KM.GenericFunctionUtility("1.5 if t<2.0 3.0", parameters)
 
+    def test_GenericFunctionUtilityError(self):
+        with self.assertRaisesRegex(Exception, 'Parsing error in function: \(0\)\*\(50\*\(expp\(t\)-1\)\)\nError occurred near here :             \^ \(char \[12\]\)\nCheck your locale \(e.g. if "." or "," is used as decimal point'):
```
EduardGomezEscandell, line 173: Note that with the wrong locale this test still passed in my computer, so this test might give wrong confidence about the issue. Tinyexpr only failed in the very specific circumstances I expose in my issue (the rom unit test).

philbucher: Yes I know, but checking the locale is quite tricky (e.g. needs to be installed on the system etc)
Hence I added your check abt the error location and a hint to check the locale (still better than nothing I think)

kratos/tests/cpp_tests/utilities/test_function_parser_utility.cpp
loumalouomega: But but, my cultural reference...

loumalouomega (approved):

kratos/utilities/function_parser_utility.cpp
```diff
@@ -182,11 +182,24 @@ void BasicGenericFunctionUtility::InitializeParser()
...
 
+        auto fct_checker = [](const bool IsValid, const std::string& rFunction, const int ErrPos){
+            if (IsValid) return;
+
+            std::stringstream ss;
+            ss << "\nParsing error in function: " << rFunction << '\n';
+            ss <<   "Error occurred near here : ";
+            for(int i=0; i<ErrPos-1; ++i) ss << ' ';
+            ss << "^ (char ["<< ErrPos-1 << "])\n";
+            ss << "Check your locale (e.g. if \".\" or \",\" is used as decimal point)";
```
EduardGomezEscandell, line 193:
```suggestion
            ss << "Ensure the expression is valid and check your locale (e.g. if \".\" or \",\" is used as decimal point)";
```

I know it should be obvious It's probably better for this to be here.

EduardGomezEscandell:
> FYI I also played around with the locale of C++, but without success, i.e. I could not add more useful information

I think @roigcarlo found the source of the issue in the MPI wrapper. Idk if there is any immediate fix.

philbucher:
> > FYI I also played around with the locale of C++, but without success, i.e. I could not add more useful information
> 
> I think @roigcarlo found the source of the issue in the MPI wrapper. Idk if there is any immediate fix.

I mean that I tried to print more info using https://en.cppreference.com/w/cpp/locale/locale but it was not working as I thought

PR #9703, opened 2022-02-23
[Mapping] making SerialOutputProc more flexible

philbucher:
@mpentek please check if it is now flexible enough for your application


mpentek (dismissed):
Tested and approved

mpentek (approved):
Works after bugfix

Issue #9253, opened 2021-10-22
[Parameters] pop method is missing


philbucher:
**Description**
`Kratos::Parameters` can do append items to array-type params but not pop them

i.e. 
~~~py

params = ...

params.Append(xxx)

params.Pop(0) # this functionality is missing
~~~

@KratosMultiphysics/technical-committee

PR #9704, opened 2022-02-23
[HDF] not writing mesh in restart

### LINKED ISSUE #9225: [HDF] SingleMesh processes write mesh again in restart
See title.

I think this can be easily changed, but I am not sure what would be the best way.
@sunethwarna do you have an idea how to best achieve this? Thx for the help!

philbucher:
See title
Now it is not writing the mesh in a restart

The impl is a bit awkward as there is no way to remove an array item (see #9253)

closes #9225


sunethwarna (changes requested):
Thanks @philbucher. I only have one minor concern with regard to `multiple_mesh_temporal_output_process`, Could you put the same if block there as well? Eventhough it is suppose to write mesh with a frequency, this if block will avoid writing it at "before_solution_loop" step when restarted.

PR #9706, opened 2022-02-23
[Core] improve restart utility

philbucher:
1. Improve the handling of the parameters in the restart utility (same as we do for many other classes)
2. Allow the usage of custom `DataCommunicator`s, same as what is done in the `DistributedImportModelPartUtility`


rubenzorrilla (approved):
Very good. Thanks!

PR #9707, opened 2022-02-23
[CoSim] some minor fixes

philbucher:
nothing very interesting :)


mpentek (dismissed):
Indeed nothing interesting, yet correct.

mpentek (approved):
Thanks

PR #9711, opened 2022-02-23
[MeshMoving] missing argument

philbucher:
Now it can be use with Communicators other than `MPI_COMM_WORLD`

Note that this should eventually be replaced with the `ConnectivityPreserveModeller`


mpentek (approved):
Good enhancement

PR #9712, opened 2022-02-23
[Core][MPI] improve DataComm handling in Communicator

philbucher:
Using the previously passed `DataCommunicator` in `Create`


roigcarlo (approved):
makes sense

PR #9713, opened 2022-02-23
[CoSim] Improve MPI handling

philbucher:
Now the definition of MPI settings is unified across all solvers


mpentek (approved):
Thanks for the better handling

PR #9714, opened 2022-02-24
[dem] Fixing override not overriding

maceligueta:
**📝 Description**
This PR fixes the problem detected by @rlrangel, where a method was supposed to be overriding but it was actually not overriding any method, as the base class method was not marked as virtual.

I also removed some useless and non-pythonic parentheses in other places.


salvalatorre (approved):
Everything looks fine!

PR #9719, opened 2022-02-25
[Docker] Upgrading Linix Wheelbuilder

roigcarlo:
**📝 Description**
Updating the base wheelbuilder image to Centos7

This allows us to reactivate CoSimApplication (we had to disable it because it depended on glib 2.16, ml2010 was glib 2.12)

Also I took the opportunity to 
- Upgrade mmg to 5.5.1 due to a bug in 5.4.3 with recent versions of gcc
- Changes to the tree structure are now done in the same layer 
- Using native cmake from the ml2014 image, which has the minimum version req for Kratos
- Numpy is now a core dependency

@philbucher this should solve the problems with the CoSimIO that I was having, just FYI.

**🆕 Changelog**
- Upgrading base image for linux wheelbuilder
- Upgrading mmg version for linux wheelbuilders
- Using native cmake in linux wheelbuilders

### DIFF
--- a/scripts/docker_files/docker_file_wheelbuilder_linux/Dockerfile
+++ b/scripts/docker_files/docker_file_wheelbuilder_linux/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/pypa/manylinux2010_x86_64
+FROM quay.io/pypa/manylinux2014_x86_64
 
 ENV MMG_ROOT=/external_libraries/mmg/mmg_5_4_1
 ENV LD_LIBRARY_PATH ${LD_LIBRARY_PATH}:{MMG_ROOT}/lib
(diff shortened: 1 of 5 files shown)


scripts/docker_files/docker_file_wheelbuilder_linux/Dockerfile
```diff
@@ -1,43 +1,39 @@
...
+# Prepare Shared Volumes
 RUN mkdir -p /data_swap_guest
 
+# Prepare File System
 RUN mkdir -p /workspace/scripts; \
-	mkdir -p /workspace/kratos
+	mkdir -p /workspace/kratos; \
+    mkdir -p /workspace/boost
+
+# Install Boost (from source)
```
roigcarlo, line 17: Before anyone comments. I know its commented. The intention is to have boost cached in the compilation server so we don't need to redownlod it every time we build the image (because is slow as hell), but I did not have time to finish implementing.

philbucher (approved):

PR #9627, opened 2022-02-08
[Core] Fixing sympy_fe_utilities

EduardGomezEscandell:
**📝 Description**
So far the `sympy_fe_utilities` worked well enough that they could be used for C++ code generation.

However, after trying to use them to generate Python code, some issues arose:
- Inconsistent use of indentation: Although there is a parameter to control it, it didn't work
- `const double` at the begining of variable declarations
- No calls to `sympy.pycode`, only using `__str__`.

Also some general issues:
- Duplicated code
- Inefficient code in some parts
- Out of style
  - Wildcard import
  - `if(a == True)`
  - `space( placement,not,like,style,guide )`
  - Using `%` instead of `.format` (Hopefully one day we can update to f-strings)

**🆕 Changelog**
- Fixed all the aforementioned issues.
- Changed a couple function calls due to deprecation of function parameter (see "Deprecation" section)
- Created basic unit tests

**💬 Deprecation**
I changed the code to replace matrix accessing. In simplified terms, it went from:
```Python
def _ReplaceIndices(string, max_index, etc):
    for i in range(max_index):
        string = string.replace("[i]", "_i")
   return string
```
to:
```Python
def _ReplaceIndices(string, etc):
    return re.sub(r"\[(\d+)\]", r"_\1", string)
```
The point is that the argument `max_index` is now unnecessary. This function is protected but this parameter is propagated up all the way to the API. Is this OK or would anyone prefer a deprecation warning?

I vote to get rid of them but please provide your thoughts.

**Effects**
@KratosMultiphysics/contact-mechanics I see you have your custom `sympy_fe_utilities`, be aware that some of the presented issues affect you as well and aren't fixed in this PR.

Other applications affected:
- @KratosMultiphysics/fluid-dynamics 
- @KratosMultiphysics/potential-flow-team 

In general you may want to re-run your symbolic generators just to check.
Nothing should break, and the only changes should be nicer whitespace.

The other potentially breaking change is chained wildcard imports. If you relied on `from sympy_fe_utilities import *` to also import all of `sympy`, then an error will occur. Otherwise we should be fine.

PR #9720, opened 2022-02-25
[Core] Fixing sympy_fe_utilities API breakage fallout

EduardGomezEscandell:
***📝 Description***
This PR deals with the fallout of breaking API in https://github.com/KratosMultiphysics/Kratos/pull/9627.

The following are the problems now solved.

***Matrix and vector access***
The `replace_indices` options converted:
`matrix_1_3 -> matrix[1,3]`

Accidentally, the aforementioned PR made it so the conversion became:
`matrix[1,3] -> matrix_1_3`

Of course this broke most symbolic generators. This has now been reverted.

***Wildcard imports***
As per the aforementioned PR:
> The other potentially breaking change is chained wildcard imports. If you relied on from sympy_fe_utilities import * to also import all of sympy, then an error will occur. Otherwise we should be fine.

Unfortunately many scripts did:
```Python
from sympy import *
from KratosMultiphysics import *
from KratosMultiphysics.sympy_fe_utilities import *
```
This caused a name collision between `Kratos.Matrix` and `sympy.Matrix`. This happened to work because the last of the imports imported the sympy matrix. This is no longer the case. 

***Whitespace***
The previous PR also modified the way in which whitespace is generated, so I recommend disabling whitespace diffs to see clearer into the changes.

**🆕 Changelog**
- Fixed `sympy_fe_utilities`.
- Fixed wildcard imports.
- Ran all symbolic generators to update the c++ versions.
- Updated some unit tests to validate the `replace_indices` behaviour.

**⚠️ Unsolved conflicts**
Some generated files were modified after being generated. For instance, the AUSAS solver had a few `const` added to methods in the generated file but not in the template.

Besides this, the following files had large changes when re-generated compared to the existing source files:
- `weakly_compressible_navier_stokes.cpp`
- `embedded_ausas_navier_stokes.cpp`


rubenzorrilla (approved):
Thanks for the quick fix.

PR #9728, opened 2022-02-28
[Core] Fixing uninitialized Dx in block builder and solver

pooyan-dadvand:
**📝 Description**
The Dx was uninitialized in `SystemSolveWithPhysics` and valgrind correctly detected it.

**🆕 Changelog**
Move the SetToZero call outside of the if to include all branches

Thanks to @pablobecker and @qiukailu for pin pointing the error!

### DIFF
--- a/kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver.h
+++ b/kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver.h
@@ -476,7 +476,9 @@ public:
         else
             norm_b = 0.00;
 
+        // Initialize the solution vector
         TSparseSpace::SetToZero(Dx);
+
 
         if (norm_b != 0.00) {
             //provide physical data as needed
--- a/kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver.h
+++ b/kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver.h
@@ -476,6 +476,8 @@ public:
         else
             norm_b = 0.00;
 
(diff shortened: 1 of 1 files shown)


kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver.h
```diff
@@ -476,6 +476,8 @@ class ResidualBasedBlockBuilderAndSolver
         else
             norm_b = 0.00;
 
+        TSparseSpace::SetToZero(Dx);
```
loumalouomega:
```suggestion
        // Initialize the solution vector
        TSparseSpace::SetToZero(Dx);

```

loumalouomega (dismissed):

RiccardoRossi (changes requested):
the change should not be done here, but rather on line 

447

https://github.com/KratosMultiphysics/Kratos/pull/9728/files#diff-668533ef54116b8368300eb04251244aa7d2f8e668bb97653e64f4582b95342dR447

otherwise you would be resetting it twice (and once after the application of Dirichlet) in the case in which you have no MPCs

pooyan-dadvand:
> the change should not be done here, but rather on line
> 
> 447
> 
> https://github.com/KratosMultiphysics/Kratos/pull/9728/files#diff-668533ef54116b8368300eb04251244aa7d2f8e668bb97653e64f4582b95342dR447
> 
> otherwise you would be resetting it twice (and once after the application of Dirichlet) in the case in which you have no MPCs

You are right! Changed in the last commit

RiccardoRossi (approved):

PR #9732, opened 2022-03-01
[ContactStructuralMechanicsApplication] Refactor ULM solver dofs

loumalouomega:
**📝 Description**
This updates the dofs in the ULM solver. With this now the AMGCL works (I didn't know that AMGCL also used the dofs for solving the problem)

**🆕 Changelog**
- Update dofs for only displacement
- Update test


AlejandroCornejo (approved):
let's see if the tests are passing! :)

PR #9736, opened 2022-03-02
[ContactStructuralMachApp] Reverting change regarding condensed solver

AlejandroCornejo:
In this PR we are trying to solve an issue regarding the condensed ulm solver

As you can see @loumalouomega, if I do not use AMGCL, the condensed ulm solver was not used :)


loumalouomega (approved):

loumalouomega (changes requested):
Checking something

PR #9740, opened 2022-03-03
sizes bugfix

RiccardoRossi:
**📝 Description**
bugfix for the case of rectangular matrices in the simplified interface

bugfix

**🆕 Changelog**
column size instead of row sizes was used in the simplified interface. fixing


rubenzorrilla (approved):

philbucher:
did you add tests for rectangular matrices?

PR #9743, opened 2022-03-03
[Core] Fixed compiler warning in reorder_and_optimize_modelpart_process.cpp

EduardGomezEscandell:
**📝 Description**
Removes unused block of code that clang was warning about.

Per clang:
```
In file included from /home/egomez/Work/Kratos/build/Release/kratos/CMakeFiles/KratosCore.dir/Unity/unity_2_cxx.cxx:31:
/home/egomez/Work/Kratos/kratos/processes/reorder_and_optimize_modelpart_process.cpp:194:26: warning: variable 'nnz' set but not used [-Wunused-but-set-variable]
            unsigned int nnz = 0; //number of nonzeros in the graph
```

**🆕 Changelog**
- Removed a block of code that computed a variable that was never used.


sunethwarna (approved):
Thanks @EduardGomezEscandell.

PR #9744, opened 2022-03-03
[Core] Fixed compiler warning in matrix_market_interface.h

EduardGomezEscandell:
**📝 Description**
Fixed warning arising from passing an unitialized value to a function as `const&`.

Per clang:
```
In file included from /home/egomez/Work/kratos_installs/kratos_ROM/kratos/python/add_matrix_market_interface_to_python.cpp:22:
/home/egomez/Work/kratos_installs/kratos_ROM/kratos/includes/matrix_market_interface.h:506:33: warning: variable 'T' is uninitialized when passed as a const reference argument here [-Wuninitialized-const-reference]
    if (!IsCorrectType(mm_code, T))
                                ^
/home/egomez/Work/kratos_installs/kratos_ROM/kratos/python/add_matrix_market_interface_to_python.cpp:42:37: note: in instantiation of function template specialization 'Kratos::ReadMatrixMarketVector<boost::numeric::ublas::vector<double>>' requested here
    m.def("ReadMatrixMarketVector", ReadMatrixMarketVector <Kratos::Vector>); 
```

This also prevents an admitedly unlikely segfault here:
```diff
ValueType *V = new ValueType[nnz];

// Check if matrix type matches MM file
- if (!IsCorrectType(mm_code, V[0]))
+ if (!IsCorrectType<ValueType>(mm_code))
```
in the rare case where `nnz` is 0.

**🆕 Changelog**
- Changed from implicit overload resolution to explicit template parameter.
- Removed trailing whitespace

**Potential improvement**
Since this check is performed at compile-time, the error could be changed to a static assert and disallow compilation of the ill-formed program.

### DIFF
--- a/kratos/includes/matrix_market_interface.h
+++ b/kratos/includes/matrix_market_interface.h
@@ -1,10 +1,10 @@
-//    |  /           | 
-//    ' /   __| _` | __|  _ \   __| 
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
 //    . \  |   (   | |   (   |\__ \.
-//   _|\_\_|  \__,_|\__|\___/ ____/ 
-//                   Multi-Physics  
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
 //
-//  License:		 BSD License 
+//  License:		 BSD License
 //					 Kratos default license: kratos/license.txt
 //
 //  Main authors:    Riccardo Rossi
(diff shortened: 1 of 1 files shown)


kratos/includes/matrix_market_interface.h
```diff
@@ -503,7 +508,7 @@ template <typename VectorType> inline bool ReadMatrixMarketVector(const char *Fi
     ValueType T;
 
     // Check if vector type matches MM file
-    if (!IsCorrectType(mm_code, T))
+    if (!IsCorrectType<ValueType>(mm_code))
```
RiccardoRossi, line 511: looks reasonable to me

RiccardoRossi (approved):

PR #9746, opened 2022-03-04
[Core] Added const GetDofSet to base builder and solver

EduardGomezEscandell:
**📝 Description**
Don't know if it didn't exist for some reason but I think it's useful.

**🆕 Changelog**
- Added `const` version of `BuilderAndSolver::GetDofSet`.
- Updated python bindings of B&S and Trillinos B&S


philbucher (dismissed):

kratos/solving_strategies/builder_and_solvers/builder_and_solver.h
```diff
@@ -514,6 +514,14 @@ class BuilderAndSolver
         return mDofSet;
     }
 
+    /**
+     * @brief It allows to get the list of Dofs from the element
+     */
+    virtual const DofsArrayType& GetDofSet() const
```
rubenzorrilla, line 520:
```suggestion
    const DofsArrayType& GetDofSet() const
```
I don't think we need this to be `virtual`.

RiccardoRossi: this is a large array, so the virtual does not really matter... i think we can leave it (it will be optimized away if you use directly the derived class)

EduardGomezEscandell: I made it virtual here to keep it the same as the non-const version. I figured if someone wants to override the non-const, they may want to override this one too.

rubenzorrilla: More than the efficiency, my point was to strictly forbid people deriving it (I don't see a potential use case). Anyway, not blocking because of this.

rubenzorrilla (approved):

PR #9748, opened 2022-03-05
[CLApp] Enhancing NeoHookean law to be used with small displacement elements

AlejandroCornejo:
I just adapted the NeoHookean cl to be used with smal displ element and together with rule of mixtures (they work with E and not F)

Also added a few noalias


AlejandroCornejo:
It's weird... In my windows pc the tests are passing both in `Release` and `FullDebug` :S

AlejandroCornejo:
@loumalouomega tests are passing :)

loumalouomega (approved):

PR #9752, opened 2022-03-07
V9.1.1

roigcarlo:
**📝 Description**
- Fixes a bug introduced while enabling mpi preventing the core whl build to be generated for windows.
- Some optimizations to the build process on win
- Fixes Missing CoSimApp dependency
- Fixes some apps not being able to generate wheels due to missing doc files.

### DIFF
--- a/applications/CoSimulationApplication/CoSimulationApplication.json
+++ b/applications/CoSimulationApplication/CoSimulationApplication.json
@@ -1,7 +1,7 @@
 {
 	"wheel_name": "KratosCoSimulationApplication",
 	"included_modules": ["CoSimulationApplication"],
-	"included_binaries": ["KratosCoSimulationApplication.*", "KratosCoSimulationCore.*", "libKratosCoSimulationCore.*"],
+	"included_binaries": ["KratosCoSimulationApplication.*", "KratosCoSimulationCore.*", "libKratosCoSimulationCore.*", "co_sim_io.*"],
     "excluded_binaries": [],
 	"dependencies": ["KratosMultiphysics==${KRATOS_VERSION}"],
 	"author": "Kratos Team",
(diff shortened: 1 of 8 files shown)


applications/CoSimulationApplication/CoSimulationApplication.json
```diff
@@ -1,7 +1,7 @@
 {
 	"wheel_name": "KratosCoSimulationApplication",
 	"included_modules": ["CoSimulationApplication"],
-	"included_binaries": ["KratosCoSimulationApplication.*", "KratosCoSimulationCore.*", "libKratosCoSimulationCore.*"],
+	"included_binaries": ["KratosCoSimulationApplication.*", "KratosCoSimulationCore.*", "libKratosCoSimulationCore.*", "co_sim_io.*"],
```
philbucher, line 4: is the MPI-binary included somewhere else?

roigcarlo: I am not sure to follow, what does this change has to do with mpi?

roigcarlo: Oh, I see you have a mpi extension. I can add it if you want but I would say that so far MPI support is still on transition. For example there is no trilinos and you can only work with MPICH/IntelMPI

roigcarlo:
can someone approve?

applications/CoSimulationApplication/CoSimulationApplication.json
philbucher: Yes, both the CoSimApp and the CoSimIo have an mpiextension
Also neither of the used Trilinos 

but sure this can be added in the future

philbucher (approved):

PR #9753, opened 2022-03-08
[Rom] Fixed test that was failing in MPI

EduardGomezEscandell:
**📝 Description**
Test `test_calculate_rom_basis_output_process.py` was failing in my machine due to a race condition. This is now fixed.

The test creates a file and then validates it. Some ranks were attempting to open the file before it was created.

**🆕 Changelog**
- Added a MPI barrier after output file is created.

### DIFF
--- a/applications/RomApplication/tests/test_calculate_rom_basis_output_process.py
+++ b/applications/RomApplication/tests/test_calculate_rom_basis_output_process.py
@@ -86,6 +86,8 @@ class TestCalculateRomBasisOutputProcess(KratosUnittest.TestCase):
 
             rom_basis_process.ExecuteFinalize()
 
+            model_part.GetCommunicator().GetDataCommunicator().Barrier()
+
     def __CheckResults(self, check_hrom_settings):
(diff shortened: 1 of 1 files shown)


rubenzorrilla (commented):
OK, that barrier doesn't hurt, but how do you end up running this test in MPI? The `calculate_rom_basis_output_process.py` is not MPI-compatible yet.

EduardGomezEscandell:
> OK, that barrier doesn't hurt, but how do you end up running this test in MPI? The calculate_rom_basis_output_process.py is not MPI-compatible yet.

I didn't really check the inner workings out, I ran the test and the only error I got was a `FileNotFoundError` so I fixed it.

Although looking at it now, I see that the modelpart is generated in each rank so it's not really running distributed, just running a diferent instance of the test in each rank.

Closing because the fix makes little practical sense until both the process and test are made MPI-compatible.

loumalouomega:
Do we still need this branch @EduardGomezEscandell ?, can we remove it?

PR #9495, opened 2021-12-27
[Core] Connectivity preser modeler - bugfix - memory error

ddiezrod:
**📝 Description**
This was found some time ago by @pooyan-dadvand and @pablobecker , but it was never pushed here. The problem is that we are pointing to the same PropertiesContainer after we call the ConnectivityPreserveModeler. If one of the properties is changed or replaced in one of the model parts it will also change in the other model part.

**🆕 Changelog**
- Bugfix when assigning the properties to the destination model part.

PR #9756, opened 2022-03-09
Copy properties modeler

miguelmaso:
**📝 Description**
In ConnectivityPreserveModeler and DuplicatePropertiesModeler the properties are shared. In some cases it's interesting to work with a copy of the properties. In my case, I want it for postprocessing.

**🆕 Changelog**
- Add CopyPropertiesModeler

### DIFF
--- a/kratos/modeler/copy_properties_modeler.cpp
+++ b/kratos/modeler/copy_properties_modeler.cpp
@@ -55,13 +55,7 @@ void CopyPropertiesModeler::SetupModelPart()
     auto& r_destination_model_part = mpModel->GetModelPart(destination_model_part_name);
 
     // clear the properties of the destination model part
-    std::vector<std::size_t> properties_ids;
-    for (const auto& r_prop : r_origin_model_part.rProperties()) {
-        properties_ids.push_back(r_prop.Id());
-    }
-    for (auto prop_id : properties_ids) {
-        r_destination_model_part.RemovePropertiesFromAllLevels(prop_id);
-    }
+    r_destination_model_part.SetProperties(Kratos::make_shared<ModelPart::PropertiesContainerType>());
 
     // make copies of the properties
     RecursivelyCopyProperties(r_origin_model_part, r_destination_model_part);
--- a/kratos/modeler/copy_properties_modeler.cpp
+++ b/kratos/modeler/copy_properties_modeler.cpp
@@ -31,10 +31,10 @@ CopyPropertiesModeler::CopyPropertiesModeler(
     mParameters.ValidateAndAssignDefaults(this->GetDefaultParameters());
 }
 
-
 CopyPropertiesModeler::CopyPropertiesModeler(
     ModelPart& rOriginModelPart,
     ModelPart& rDestinationModelPart)
+        : Modeler()
 {
     mpModel = &rOriginModelPart.GetModel();
     KRATOS_ERROR_IF_NOT(mpModel == &rDestinationModelPart.GetModel()) << "CopyPropertiesModeler. The model parts belong to different models." << std::endl;
--- a/kratos/modeler/copy_properties_modeler.cpp
+++ b/kratos/modeler/copy_properties_modeler.cpp
@@ -31,6 +31,17 @@ CopyPropertiesModeler::CopyPropertiesModeler(
     mParameters.ValidateAndAssignDefaults(this->GetDefaultParameters());
 }
 
+
+CopyPropertiesModeler::CopyPropertiesModeler(
+    ModelPart& rOriginModelPart,
+    ModelPart& rDestinationModelPart)
+{
+    mpModel = &rOriginModelPart.GetModel();
+    KRATOS_ERROR_IF_NOT(mpModel == &rDestinationModelPart.GetModel()) << "CopyPropertiesModeler. The model parts belong to different models." << std::endl;
+    mParameters.AddString("origin_model_part_name", rOriginModelPart.Name());
+    mParameters.AddString("destination_model_part_name", rDestinationModelPart.Name());
+}
+
 Modeler::Pointer CopyPropertiesModeler::Create(Model& rModel,
     const Parameters ModelParameters) const
 {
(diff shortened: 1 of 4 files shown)


kratos/modeler/copy_properties_modeler.cpp
```diff
@@ -0,0 +1,89 @@
...
+    })");
+    return default_parameters;
+}
+
+void CopyPropertiesModeler::SetupModelPart()
+{
+    const auto origin_model_part_name = mParameters["origin_model_part_name"].GetString();
+    const auto destination_model_part_name = mParameters["destination_model_part_name"].GetString();
+    ModelPart& r_origin_model_part = mpModel->GetModelPart(origin_model_part_name);
+    ModelPart& r_destination_model_part = mpModel->GetModelPart(destination_model_part_name);
```
rubenzorrilla:
```suggestion
    const std::string origin_model_part_name = mParameters["origin_model_part_name"].GetString();
    const std::string destination_model_part_name = mParameters["destination_model_part_name"].GetString();
    const auto& r_origin_model_part = mpModel->GetModelPart(origin_model_part_name);
    auto& r_destination_model_part = mpModel->GetModelPart(destination_model_part_name);
```
We tend to avoid using `auto` with fundamental types. On the contrary we tend to use it with custom (potentially more complex and prone to change) types. Besides I think that the origin model part reference can be `const`.

kratos/modeler/copy_properties_modeler.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+    // clear the properties of the destination model part
+    std::vector<std::size_t> properties_ids;
+    for (const auto& r_prop : r_origin_model_part.rProperties()) {
+        properties_ids.push_back(r_prop.Id());
+    }
+    for (auto prop_id : properties_ids) {
+        r_destination_model_part.RemovePropertiesFromAllLevels(prop_id);
+
```
rubenzorrilla:
```suggestion
```

kratos/modeler/copy_properties_modeler.cpp
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    // make copies of the properties
+    for (auto& r_prop : r_origin_model_part.rProperties()) {
+        r_destination_model_part.AddProperties(Kratos::make_shared<Properties>(r_prop));
```
rubenzorrilla: In the loop above you're removing the properties from all levels but in here you're only adding them for the "main" level. I think this should be done recursively so you end up with the same subproperties structure but different content.

kratos/modeler/copy_properties_modeler.cpp
```diff
@@ -0,0 +1,89 @@
...
+}
+
+template<class TContainerType>
+void CopyPropertiesModeler::ReplaceProperties(
+    TContainerType& rContainer,
+    ModelPart& rModelPart)
+{
+    block_for_each(rContainer, [&](typename TContainerType::value_type& rEntity){
+        auto properties_id = rEntity.GetProperties().Id();
+        auto p_properties = rModelPart.pGetProperties(properties_id);
```
rubenzorrilla, line 101: If `rEntity` features a property that was not present in the origin model part you will be creating an empty one. I'm afraid this is not the intended behavior of this modeler.

kratos/python/add_modeler_to_python.cpp
```diff
@@ -97,6 +98,12 @@ void  AddModelerToPython(pybind11::module& m)
     py::class_< DuplicateMeshModeler, DuplicateMeshModeler::Pointer, Modeler >(m,"DuplicateMeshModeler")
         .def(py::init<ModelPart&>())
     ;
+
+    py::class_< CopyPropertiesModeler, CopyPropertiesModeler::Pointer, Modeler >(m,"CopyPropertiesModeler")
+        .def(py::init< >())
```
rubenzorrilla: I think the default constructor exposure is useless in this case.

rubenzorrilla (changes requested):
There are some concerns that need to be addressed. Besides test is missing.

kratos/modeler/copy_properties_modeler.cpp
miguelmaso: > Besides I think that the origin model part reference can be `const`

Unfortunately, the method `rProperties` is not const

kratos/python/add_modeler_to_python.cpp
miguelmaso: done

kratos/modeler/copy_properties_modeler.cpp
miguelmaso: done

kratos/modeler/copy_properties_modeler.cpp
miguelmaso: Now I've avoided the creation of new properties declaring `rModelPart` const.

miguelmaso:
> test is missing

> I've avoided the creation of new properties declaring rModelPart const

I think I faced the same error reported in #9495

`DuplicateMeshModeler` is fine, problems come from `ConnectivityPreserveModeler`, the container is shared. IMO the properties container must be different but the properties the same, analogous to nodes, elements and conditions.

kratos/modeler/copy_properties_modeler.cpp
miguelmaso: if the property is not present in the model part, an error will be thrown in https://github.com/KratosMultiphysics/Kratos/blob/803321499be2d40005536fd238cca150fb9ae65e/kratos/sources/model_part.cpp#L701

rubenzorrilla: Perfect. :+1:

inigolcanalejo:
We have discussed this in the @KratosMultiphysics/implementation-committee and we agree with having this new feature. Thank you for the effort @miguelmaso! @rubenzorrilla was there something else you wanted to be addressed here or is it ready to merge?

miguelmaso:
> We have discussed this in the @KratosMultiphysics/implementation-committee and we agree with having this new feature. Thank you for the effort @miguelmaso! @rubenzorrilla / @KratosMultiphysics/technical-committee was there something else you wanted to be addressed here or is it ready to merge?

It's ready to merge. Thanks!

rubenzorrilla (approved):
All concerns addressed

miguelmaso:
Thanks for your suggestions @rubenzorrilla !

PR #9758, opened 2022-03-10
[Core] [ConnectivityPreserveModeler] copy of the properties container

miguelmaso:
**📝 Description**
Possible solution to #9495 
Related to #9756 

**🆕 Changelog**
- Copy of the properties container instead of sharing it
- `coupled_structural_thermal_solver`: calling the modeler after reading the properties (see the comment below)

### DIFF
--- a/kratos/modeler/connectivity_preserve_modeler.cpp
+++ b/kratos/modeler/connectivity_preserve_modeler.cpp
@@ -144,7 +144,7 @@ void ConnectivityPreserveModeler::CopyCommonData(
 
     // These should be safe for SubModelParts
     rDestinationModelPart.SetProcessInfo( rOriginModelPart.pGetProcessInfo() );
-    rDestinationModelPart.SetProperties( rOriginModelPart.pProperties() );
+    rDestinationModelPart.PropertiesArray() = rOriginModelPart.PropertiesArray();
     rDestinationModelPart.Tables() = rOriginModelPart.Tables();
 
     // Assign the nodes to the new model part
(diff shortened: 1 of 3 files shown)


miguelmaso:
After the updating the `ConnectivityPreserveModeler`, the coupled structural-thermal test was failing. This is because the modeler was called before reading the properties (prepare the model part). Changing the order of the calls in the `coupled_structural_thermal_solver` fixed the test (the same property is used by both solvers).
Maybe this change is more aligned with the modeler stages.

ddiezrod:
We reviewed this in @KratosMultiphysics/implementation-committee please take a look when you can @RiccardoRossi @KratosMultiphysics/technical-committee

ddiezrod:

kratos/modeler/connectivity_preserve_modeler.cpp
```diff
@@ -144,7 +144,7 @@ void ConnectivityPreserveModeler::CopyCommonData(
 
     // These should be safe for SubModelParts
     rDestinationModelPart.SetProcessInfo( rOriginModelPart.pGetProcessInfo() );
-    rDestinationModelPart.SetProperties( rOriginModelPart.pProperties() );
+    rDestinationModelPart.PropertiesArray() = rOriginModelPart.PropertiesArray();
```
RiccardoRossi, line 147: this mean that the DesitinationModelPart container is now different from the OriginModelPart container but populated with the same pointers, is that so?.

ddiezrod: thats it 👍

pooyan-dadvand:
I agree with this change. This also helps if we want to populate a sub modelpart without messing up with the properties

ddiezrod:
so... approval? @KratosMultiphysics/technical-committee

ddiezrod:

roigcarlo (approved):
Since @pooyan-dadvand agrees, and I can't see anything wrong :+1:

PR #9762, opened 2022-03-11
[Fluid] Compressible NS automatic_dt considers artificial diffusion

EduardGomezEscandell:
**📝 Description**
Small change in compressible NS solver to enforce the use of artificial diffusivities in computation of Fourier numbers when shock capturing and automatic time step are enabled.

**🆕 Changelog**
- consider_artificial_diffusion is now overidden to the same value as `SHOCK_CAPTURING_SWITCH`.

### DIFF
--- a/applications/FluidDynamicsApplication/python_scripts/navier_stokes_compressible_explicit_solver.py
+++ b/applications/FluidDynamicsApplication/python_scripts/navier_stokes_compressible_explicit_solver.py
@@ -148,18 +148,21 @@ class NavierStokesCompressibleExplicitSolver(FluidSolver):
             err_msg = err_msg + " - {}\n".format(key)
         raise RuntimeError(err_msg)
 
+    def _OverrideBoolParameterWithWarning(self, parent, child, value):
+        if parent.Has(child) and parent[child].GetBool() != value:
+            KratosMultiphysics.Logger.PrintWarning("", "User-specifed {} will be overriden with {}".format(parameter, value))
+        else:
+            parent.AddEmptyValue(child)
+        parent[child].SetBool(True)
+
     def _CreateEstimateDtUtility(self):
         """This method overloads FluidSolver in order to enforce:
         ```
         self.settings["time_stepping"]["consider_compressibility_in_CFL"] == True
         ```
         """
-        if self.settings["time_stepping"].Has("consider_compressibility_in_CFL"):
-            KratosMultiphysics.Logger.PrintWarning("", "User-specifed consider_compressibility_in_CFL will be overriden with TRUE")
-        else:
-            self.settings["time_stepping"].AddEmptyValue("consider_compressibility_in_CFL")
-
-        self.settings["time_stepping"]["consider_compressibility_in_CFL"].SetBool(True)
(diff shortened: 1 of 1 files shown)


applications/FluidDynamicsApplication/python_scripts/navier_stokes_compressible_explicit_solver.py
```diff
@@ -150,18 +150,24 @@ def _create_solution_strategy(self):
             err_msg = err_msg + " - {}\n".format(key)
         raise RuntimeError(err_msg)
 
+    @classmethod
+    def _OverrideBoolParameterWithWarning(cls, parent, child, value):
+        if parent.Has(child) and parent[child].GetBool() != value:
+            KratosMultiphysics.Logger.PrintWarning("", "User-specifed {} will be overriden with {}".format(child, value))
```
rubenzorrilla, line 156:
```suggestion
        if parent.Has(child):
            if parent[child].GetBool() != value:
                KratosMultiphysics.Logger.PrintWarning("", "User-specifed {} will be overriden with {}".format(child, value))
```

ddiezrod: Why? I think the original version looks better

rubenzorrilla: Wouldn't it crash if the `Has` is false when trying to access `parent[child]`? (here I'm assuming that all the loop clauses are evaluated, not 100% sure about how it works internally).

sunethwarna: AFAIK, `and` compound statements does not execute the trailing ones if one of the leading ones are false.

```
class A:
   def Is(self, t):
      print("test")
      return t
a = A()
b = A()
if a.Is(False) and b.Is(False):
   print("D")
```

ddiezrod: nope, "and" and "or" operators follow short-circuit evaluation. In the and, if first condition is false it does not access the second one. Similarly, in or, if first condition is true, second is not evaluated.

ddiezrod: https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not

EduardGomezEscandell: Python has short-circuiting just like C and C++.

See the [reference](https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not) regarding the `and` operator:
> This is a short-circuit operator, so it only evaluates the second argument if the first one is true.

In summary, this is a safe way of accessing parameters.

ddiezrod: pole

rubenzorrilla: Very good. Thanks. I wasn't aware.

rubenzorrilla (approved):

PR #9764, opened 2022-03-14
[StructuralMechanicsApplication] Adding pyramid solid element

loumalouomega:
**📝 Description**
Adding pyramid solid element

**🆕 Changelog**
- Adding pyramid solid element
- Added test
- Correcting missing methods in geometries
- Expose python


philbucher (approved):
thx!

AlejandroCornejo:
interesting! which is the main application of this element? :)

loumalouomega:
> interesting! which is the main application of this element? :)

To mix hexa and tetra meshes, or just a regular element with a very strange shape

loumalouomega:
> interesting! which is the main application of this element? :)

https://arxiv.org/pdf/1705.02451.pdf

PR #8478, opened 2021-03-12
[Fluid] Compressible NS slip wall porosity

rubenzorrilla:
**Description**
This PR adds the possibility of considering "porous" slip walls in the compressible NS explicit strategy slip BC imposition. This is quite helpful during the flow initialization.

So far I only add the changes in the strategy as well as the `SLIP_WALL_FACTOR` variable. Eventually, we can create a Python process that sets this value in the `ProcessInfo` in accordance to a user-defined table.

PR #9765, opened 2022-03-14
[Fluid] Created process for walls to gradually become solid at startup

EduardGomezEscandell:
**📝 Description**
This PR builds upon the idea presented in https://github.com/KratosMultiphysics/Kratos/pull/8478.

Both PR try to solve the issue of starting a compressible fluid simulation with uniform velocity field in a domain that contains a wall oposing this movement. This usually causes numerical issues, so one way to mitigate this is by slowly interpollating between an open outlet and a solid wall.

The previous approach subtracted a fixed proportion of speed every step. The issue with this is that smaller time-steps lead to faster deceleation, causing smaller time-steps to create more problematic flows.

The new approach is independent of time-step size, and simply subtracts a fixed proportion of momentum per unit time until close enough to solid, where it simply switches to a regular free slip boundary condition.

The law it follows is essentially an exponential decay:
```
After every time-step:

                { m * A^(-t/period)  if t < period
  m_corrected = {
                { m                  if t >= period
```
where:
- `m` is the momentum normal to the wall.
- `period` is a parameter to control, the smaller the more agressive the transition.
- `A` is just an arbitrary constant. It represents the proportion `A = m(t=0)/m(t=period)`. It's hardcoded to 1000, meaning that once the wall is 99.9% solid, it jumps to 100% next step.

**🆕 Changelog**
- Created `DeceleratingWallProcess`.
- Created unit test


applications/FluidDynamicsApplication/python_scripts/decelerating_wall_process.py
```diff
@@ -0,0 +1,99 @@
...
+
+        if time > self.period:
+            KratosMultiphysics.VariableUtils().SetFlag(
+                KratosMultiphysics.SLIP, True, self.model_part.Nodes
+            )
+            self.period_ended = True
+            return
+
+        if self.recompute_normals:
+            KratosMultiphysics.NormalCalculationUtils().CalculateNormals(self.model_part, True)
```
rubenzorrilla: Minor detail but, I think this must be done in the `ExecuteInitializeSolutionStep` so the `NORMAL` values you use for solving the problem are that ones you will use for the wall velocity manipulation.

applications/FluidDynamicsApplication/python_scripts/decelerating_wall_process.py
```diff
@@ -0,0 +1,99 @@
...
+            raise ValueError("Parameter period must be strictly positive.")
+
+        self.period_ended = False
+
+    def GetDefaultParameters(self):
+        return KratosMultiphysics.Parameters(
+            """
+            {
+                "model_part_name" : "SPECIFY_MODEL_PART_NAME",
+                "period"        : 0,
```
rubenzorrilla: Could we use an interval as the other processes do?

applications/FluidDynamicsApplication/python_scripts/decelerating_wall_process.py
```diff
@@ -0,0 +1,99 @@
...
+    def ExecuteFinalizeSolutionStep(self):
+        if self.period_ended:
+            return
+
+        time = self.model_part.ProcessInfo[KratosMultiphysics.TIME]
+
+        if time > self.period:
+            KratosMultiphysics.VariableUtils().SetFlag(
+                KratosMultiphysics.SLIP, True, self.model_part.Nodes
+            )
```
rubenzorrilla: Correct me if I'm wrong, but the idea is that once the "porous wall" period is over, we set the `SLIP` flag so the normal component of momentum is substracted in the compressible N-S strategy right? If so it is
1. I'd put a comment explaining it.
2. I'd make sure the flag is `false` by setting it in the `ExecuteInitialize`.

applications/FluidDynamicsApplication/python_scripts/decelerating_wall_process.py
```diff
@@ -0,0 +1,99 @@
...
+            KratosMultiphysics.NormalCalculationUtils().CalculateNormals(self.model_part, True)
+
+        dt = self.model_part.ProcessInfo[KratosMultiphysics.DELTA_TIME]
+        decay = (1000)**(-dt/self.period)
+
+        for node in self.model_part.Nodes:
+            mom = node.GetSolutionStepValue(KratosMultiphysics.MOMENTUM)
+
+            unit_normal = node.GetSolutionStepValue(KratosMultiphysics.NORMAL)
+            unit_normal /= unit_normal.norm_2()
```
rubenzorrilla: You need to check that the normal norm is not close to zero.

applications/FluidDynamicsApplication/python_scripts/decelerating_wall_process.py
```diff
@@ -0,0 +1,99 @@
...
+
+        self.period_ended = False
+
+    def GetDefaultParameters(self):
+        return KratosMultiphysics.Parameters(
+            """
+            {
+                "model_part_name" : "SPECIFY_MODEL_PART_NAME",
+                "period"        : 0,
+                "recompute_normals" : false
```
rubenzorrilla: Maybe we can put an extra field `wall_variable` (naming suggestions are welcome) that defaults to `MOMENTUM` so we can switch to velocity if we eventually move to the primitive form.

rubenzorrilla (commented):
Some minor comments. Besides, I think that `decelerating_wall_process` might be a bit misleading name (what comes to my mind is a moving ALE wall). I'd suggest something like `compressible_slip_wall_process`. Note that if the interval is [0.0,0.0] this sets the `SLIP` flag to true, which is what we need if no porosity is to be set. Long story short, we can adopt this for the compressible NS and stop using the `apply_slip_process`.

applications/FluidDynamicsApplication/python_scripts/decelerating_wall_process.py
EduardGomezEscandell: Agreed

applications/FluidDynamicsApplication/python_scripts/decelerating_wall_process.py
EduardGomezEscandell: Yes, no problem. Working on it.

applications/FluidDynamicsApplication/python_scripts/decelerating_wall_process.py
EduardGomezEscandell: Good catch.

EduardGomezEscandell:
> Some minor comments. Besides, I think that decelerating_wall_process might be a bit misleading name (what comes to my mind is a moving ALE wall). I'd suggest something like compressible_slip_wall_process.

Agreed. I changed it to your proposed name.

> Note that if the interval is [0.0,0.0] this sets the SLIP flag to true, which is what we need if no porosity is to be set.

I misread this at first but I think the end result may be better, let me know what you think. I made it so the time line is:
```
 Begin
   ├─ Free
interval[0]
   ├─ Porous wall
interval[0] + rampup_time
   ├─ Slip wall
interval[1]
   ├─ Free
  End
```
Anyway, the standard would be 
```json
{
    "interval": [0, "End"],
    "rampup_time": 0
}
```
to have a usual slip wall.

> Long story short, we can adopt this for the compressible NS and stop using the apply_slip_process.

Agreed.

applications/FluidDynamicsApplication/python_scripts/compressible_slip_wall_process.py
```diff
@@ -70,6 +70,7 @@ def __init__(self, model, settings):
         self.rampup_period = settings["rampup_period"].GetDouble()
         self.recompute_normals = settings["recompute_normals"].GetBool()
         self._stage = self.AWAITING
+        self.variable = KratosMultiphysics.KratosGlobals.GetVariable(settings["variable"].GetString())
```
rubenzorrilla:
```suggestion
        self.variable = KratosMultiphysics.KratosGlobals.GetVariable(settings["variable_name"].GetString())
```
To match the naming we normally use in Kratos.

rubenzorrilla: You should also check that the provided variable is of vector type.

EduardGomezEscandell: This is done during the `Check`.

applications/FluidDynamicsApplication/python_scripts/compressible_slip_wall_process.py
```diff
@@ -93,6 +94,13 @@ def Check(self):
             msg = "Parameter rampup_period must greater than the interval ({0} < {1}).".format(self.rampup_period, full_duration)
             raise ValueError(msg)
 
+        self.model_part.HasNodalSolutionStepVariable(self.variable)
```
rubenzorrilla: Shouldn't we do something with this?

EduardGomezEscandell: This is what happens when you take a coffee break mid-coding :laughing:.

Fixed.

applications/FluidDynamicsApplication/python_scripts/compressible_slip_wall_process.py
```diff
@@ -0,0 +1,171 @@
...
+
+        settings.ValidateAndAssignDefaults(self.GetDefaultParameters())
+        modelpart_name = settings["model_part_name"].GetString()
+
+        self.model_part = model[modelpart_name]
+        self.rampup_period = settings["rampup_period"].GetDouble()
+        self.rampup_enabled = True
+        self.recompute_normals = settings["recompute_normals"].GetBool()
+        self._stage = self.AWAITING
+        self.variable = KratosMultiphysics.KratosGlobals.GetVariable(settings["variable_name"].GetString())
```
rubenzorrilla, line 76: We should check that the variable is of vector type (I don't know what happened with my previous comment).

applications/FluidDynamicsApplication/python_scripts/compressible_slip_wall_process.py
```diff
@@ -0,0 +1,171 @@
...
+            """
+        )
+
+    def Check(self):
+        if self.rampup_period <= 0:
+            raise ValueError("Parameter rampup_period must be positive or zero.")
+
+        full_duration = self.interval.GetIntervalEnd() - self.interval.GetIntervalBegin()
+        if self.rampup_period > full_duration:
+            msg = "Parameter rampup_period must greater than the interval ({0} < {1})."\
```
rubenzorrilla:
```suggestion
            msg = "Parameter rampup_period must smaller than the interval ({0} < {1})."\
```

applications/FluidDynamicsApplication/python_scripts/compressible_slip_wall_process.py
```diff
@@ -0,0 +1,171 @@
...
+                .format(self.rampup_period, full_duration)
+            raise ValueError(msg)
+
+        self.model_part.HasNodalSolutionStepVariable(self.variable)
+
+        normal_type = type(KratosMultiphysics.NORMAL)
+        if not isinstance(self.variable, normal_type):
+            msg = "Variable must be a vector with the same size as NORMAL. Provided variable {} is not"\
+                .format(self.variable.Name())
+            raise TypeError(msg)
```
rubenzorrilla: You can easily check the type with `GetVariableType` of the `KratosGlobals`.

applications/FluidDynamicsApplication/python_scripts/compressible_slip_wall_process.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def ExecuteInitialize(self):
+        KratosMultiphysics.NormalCalculationUtils().CalculateNormals(self.model_part, True)
+        KratosMultiphysics.VariableUtils().SetFlag(KratosMultiphysics.SLIP, False, self.model_part.Nodes)
+
+        if self.rampup_period < 1e-18:
```
rubenzorrilla:
```suggestion
        if not self.rampup_period > 0.0:
```
I think it is cleaner.

applications/FluidDynamicsApplication/python_scripts/compressible_slip_wall_process.py
```diff
@@ -0,0 +1,171 @@
...
+        self._stage = self.AWAITING
+        self.variable = KratosMultiphysics.KratosGlobals.GetVariable(settings["variable_name"].GetString())
+
+    def GetDefaultParameters(self):
+        return KratosMultiphysics.Parameters(
+            """
+            {
+                "model_part_name"   : "SPECIFY_MODEL_PART_NAME",
+                "interval"   : [0, "End"],
+                "rampup_period"   : 0,
```
rubenzorrilla:
```suggestion
                "rampup_time"   : 0,
```
I wouldn't use period (inverse of the frequency). Not long ago we had the same discussion with the `GIDOutputProcess`.

applications/FluidDynamicsApplication/python_scripts/compressible_slip_wall_process.py
EduardGomezEscandell: I accidentally marked it as solved.

My response was that this check is already in the `Check` method.

rubenzorrilla: OK!

applications/FluidDynamicsApplication/python_scripts/compressible_slip_wall_process.py
rubenzorrilla: OK!

rubenzorrilla (approved):

Issue #7381, opened 2020-08-24
Interfaces of Element, Condition and Scheme updated


philbucher:
Dear @KratosMultiphysics/all

After many issues in the past (e.g. with race conditions) the @KratosMultiphysics/technical-committee decided to clean up and improve the interfaces of `Element`, `Condition` and `Scheme`. See #4938, #6643 and #5290.

The changes are done **FULLY BACKWARD COMPATIBLE** (please check the PRs for technical details), hence he updating can be done in multiple steps!!!
The Core and many applications have already been updated!

The main changes are adding `const` to the `ProcessInfo` that is passed to the functions. This is to avoid race conditions.

The following changes were done:
- `Element` & `Condition`:
    - All instances of `ProcessInfo` are being passed as `const`
    - The following functions are `const` functions now:
        - `GetValuesVector`
        - `GetFirstDerivativesVector`
        - `GetSecondDerivativesVector`
        - `Check`
    - `Initialize` gets an instance of `const ProcessInfo&` now
    - `CalculateLocalSystem` with `std::vector` is removed as unused
    - `GetValueOnIntegrationPoints` is deprecated in favor of `CalculateOnIntegrationPoints`. The functions have the same signature, hence search-replace is sufficient

- `Scheme`:
    - All instances of `ProcessInfo` are being passed as `const`
    - Element and Condition are no longer passed as `Pointer` but as reference
    - The function names starting with `Condition_` have the `Condition_` removed i.e. the names for Element and Condition are the same now

The following procedure is suggested for updating the code:
1. Update the c++ tests (if existing) to use the new interfaces, see e.g. #7366
2. Update the interfaces see e.g. #7371
In case some results change it is very likely that a race condition. If you cannot find the issue please let us know

Please check & update your applications. The timeline is (to be confirmed by @KratosMultiphysics/technical-committee):
- activating the deprecation warnings: 2 months from now, end of October
- removing the deprecated functions: 5 months from now, end of January

Please don't hesitate to ask questions if something is unclear!

The following applications have been updated already (if it was necessary)

- CableNetApplication
- ChimeraApplication
- CompressiblePotentialFlowApplication
- ConstitutiveModelsApplication
- ContactStructuralMechanicsApplication
- CoSimulationApplication
- CSharpWrapperApplication
- EigenSolversApplication
- ExaquteSandboxApplication
- ExternalSolversApplication
- FluidDynamicsApplication (partly)
- FreeSurfaceApplication
[...]

PR #9767, opened 2022-03-15
[Core] Added const to condition's GetIntegrationMethod

EduardGomezEscandell:
**📝 Description**
The base element has method `virtual IntegrationMethod GetIntegrationMethod() const`.
Base condition has method `virtual IntegrationMethod GetIntegrationMethod()`.

This discrepancy is fixed in favour of the const version. Note that this would break any derived condition with `IntegrationMethod GetIntegrationMethod() override`, so the non-const version is kept with a deprecation warning.

**🆕 Changelog**
- Made base condition's GetIntegrationMethod `const`.


philbucher (approved):
Thanks, I missed this in #7381

PR #9771, opened 2022-03-16
[Core] [Properties] Add const

miguelmaso:
**📝 Description**
a `const` is missing. Now, GetProperties is consistent with GetNode, GetElement and GetCondition

**🆕 Changelog**
- Add `const`

### DIFF
--- a/kratos/includes/model_part.h
+++ b/kratos/includes/model_part.h
@@ -862,7 +862,7 @@ public:
      * @param MeshIndex The Id of the mesh (0 by default)
      * @return The desired properties (pointer)
      */
-    PropertiesType::Pointer pGetProperties(IndexType PropertiesId, IndexType MeshIndex = 0) const;
+    const PropertiesType::Pointer pGetProperties(IndexType PropertiesId, IndexType MeshIndex = 0) const;
 
     /**
      * @brief Returns the Properties::Pointer  corresponding to it's identifier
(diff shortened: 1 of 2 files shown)


kratos/includes/model_part.h
```diff
@@ -862,7 +862,7 @@ class KRATOS_API(KRATOS_CORE) ModelPart : public DataValueContainer, public Flag
      * @param MeshIndex The Id of the mesh (0 by default)
      * @return The desired properties (pointer)
      */
-    PropertiesType::Pointer pGetProperties(IndexType PropertiesId, IndexType MeshIndex = 0) const;
+    const PropertiesType::Pointer pGetProperties(IndexType PropertiesId, IndexType MeshIndex = 0) const;
```
EduardGomezEscandell, line 865: This is not really const, the object it wraps around is still mutable.
```C++
const PropertiesType::Pointer // Not really const
```

See an example [here](https://godbolt.org/z/fTx97hand).


If you trully want to make it const, it should be
```c++
shared_ptr<const PropertiesType> pGetProperties(...) // Really const
```
unfortunately kratos does not have a `::ConstPointer` typedef.

EduardGomezEscandell, line 865: `const PropertiesType::Pointer` is not really const, the object it accesses is still mutable. See an example here: https://godbolt.org/z/fTx97hand.


If you trully want to make it const, it should be
```c++
shared_ptr<const PropertiesType> pGetProperties(...)
```
unfortunatelly kratos does not have a `::ConstPointer` typedef.

miguelmaso: You're right, your explanation in #9779 is very useful.
In this PR there is also the inconsistency with nodes, elements and conditions

EduardGomezEscandell: Well since I'm going to fix it in the near future I'd say let's allow it and I'll fix it in my PR.

EduardGomezEscandell (approved):

PR #9778, opened 2022-03-18
Radial Basis Functions Utility

SADPR:
This method calculates the RBF shape function values in one point using as
support the given cloud of points.

Works correctly, not yet optimized (might be slow).


kratos/tests/test_radial_basis_functions_utility.py
```diff
@@ -0,0 +1,105 @@
...
+                },
+                "nodal_results" : ["TEMPERATURE"]
+            }
+        }"""))
+    gid_output.ExecuteInitialize()
+    gid_output.ExecuteBeforeSolutionLoop()
+    gid_output.ExecuteInitializeSolutionStep()
+    gid_output.PrintOutput()
+    gid_output.ExecuteFinalizeSolutionStep()
+    gid_output.ExecuteFinalize()
```
EduardGomezEscandell: This function should be a method of `TestRadialBasisFunctionsUtility`.

kratos/tests/test_radial_basis_functions_utility.py
```diff
@@ -0,0 +1,105 @@
...
+                N_container)
+
+        # Save the obtained results in TEMPERATURE variable and visualize
+        output_results = False
+        if output_results:
+            i = 0
+            for node in self.main_model_part.Nodes:
+                node.SetSolutionStepValue(KratosMultiphysics.TEMPERATURE, 0, N_container[i])
+                i += 1
+            PostProcess(self.main_model_part, "test_radial_basis_functions_utility_1x1_square")
```
EduardGomezEscandell: Is `PostProcess` necessary for the test or just for debugging?

If the latter, I would comment it out or remove it altogether.

kratos/tests/test_radial_basis_functions_utility.py
```diff
@@ -0,0 +1,105 @@
...
+            i += 1
+
+        # Calculate the MLS shape functions in the square midpoint
+        midpoint = KratosMultiphysics.Vector(3)
+        midpoint[0] = 0.5
+        midpoint[1] = 0.5
+        midpoint[2] = 0.0
+        N_container = KratosMultiphysics.Vector()
+        h = 1 #Shape parameter for the kernel function (radial basis function, inverse multiquadratic)
+        InputShapeParameter = False
```
EduardGomezEscandell:
```suggestion
        input_shape_parameter = False
```
Local variables should be lower case

kratos/utilities/radial_basis_functions_utility.cpp
```diff
@@ -0,0 +1,208 @@
...
+// Project includes
+#include "includes/global_variables.h"
+#include "utilities/math_utils.h"
+#include "radial_basis_functions_utility.h"
+
+namespace Kratos
+{
+
+    double RadialBasisFunctionsUtility::EvaluateRBF(
+        double x,
```
EduardGomezEscandell:
```suggestion
        const double x,
```

kratos/utilities/radial_basis_functions_utility.cpp
```diff
@@ -0,0 +1,208 @@
...
+        // Initialize the RBF interpolation matrix and RBF interpolated vector
+        Matrix A = ZeroMatrix(n_points,n_points);
+        Vector Phi = ZeroVector(n_points);
+        double norm_xij;
+        double norm_X;
+
+        // Build the RBF interpolation matrix and RBF interpolated vector
+        for (std::size_t i_pt = 0; i_pt < n_points; ++i_pt) {
+            for (std::size_t j_pt = 0; j_pt < n_points; ++j_pt) {
+                norm_xij = norm_2(row(rPoints,i_pt)-row(rPoints,j_pt));
```
EduardGomezEscandell: Stack-allocated variables don't need to be preallocated. Declare `norm_xij` and `norm_X` only when you need them. Then you can also make them `const`.

kratos/utilities/radial_basis_functions_utility.cpp
```diff
@@ -0,0 +1,208 @@
...
+            Phi[i_pt] = EvaluateRBF(norm_X,h);
+        }
+
+        // Invert the RBF interpolation matrix
+        double A_det;
+        Matrix A_inv(n_points,n_points);
+        MathUtils<double>::InvertMatrix(A, A_inv, A_det);
+
+        // Obtain the RBF shape functions 
+        noalias(rN) = prod(A_inv,Phi);
```
EduardGomezEscandell: Inverting A can be quite expensive, in this case you can use `MathUtils::Solve` and solve the system `A*N = Φ`.

kratos/utilities/radial_basis_functions_utility.cpp
```diff
@@ -0,0 +1,208 @@
...
+        double norm_X;
+        
+
+        // Find shape parameter http://www.math.iit.edu/~fass/Dolomites.pdf [Hardy]
+        // TODO: Build a matrix with distance norm values to recycle for RBFs evaluation.
+        // Make a function for this 
+        double d = 0; // Total distance of nearest x_i neighbors 
+        double h; // Shape parameter
+        for (std::size_t i_pt = 0; i_pt < n_points; ++i_pt) {
+            double d_nearest_neighbor = 1e30; // Initialize distance to nearest x_i neighbor
```
EduardGomezEscandell:
```suggestion
            double d_nearest_neighbor = std::numeric_limits<double>::max; // Initialize distance to nearest x_i neighbor 
```

You will have to `#include<limits>`as well.

kratos/utilities/radial_basis_functions_utility.cpp
```diff
@@ -0,0 +1,208 @@
...
+        // Initialize the RBF interpolation matrix and RBF interpolated vector
+        Matrix A = ZeroMatrix(n_points,n_points);
+        Vector Phi = ZeroVector(n_points);
+        double norm_xij;
+        double norm_X;
+        
+
+        // Find shape parameter http://www.math.iit.edu/~fass/Dolomites.pdf [Hardy]
+        // TODO: Build a matrix with distance norm values to recycle for RBFs evaluation.
+        // Make a function for this
```
EduardGomezEscandell: Agree :smiley:

kratos/utilities/radial_basis_functions_utility.cpp
```diff
@@ -0,0 +1,208 @@
...
+        Vector Phi = ZeroVector(n_points);
+        double norm_xij;
+        double norm_X;
+        
+
+        // Find shape parameter http://www.math.iit.edu/~fass/Dolomites.pdf [Hardy]
+        // TODO: Build a matrix with distance norm values to recycle for RBFs evaluation.
+        // Make a function for this 
+        double d = 0; // Total distance of nearest x_i neighbors 
+        double h; // Shape parameter
```
EduardGomezEscandell: Wait until needed to declare the variable, then you can make it const.

kratos/utilities/radial_basis_functions_utility.cpp
```diff
@@ -0,0 +1,208 @@
...
+            }
+        }
+
+        // Invert the RBF interpolation matrix
+        double A_det;
+        Matrix A_inv(n_points,n_points);
+        MathUtils<double>::InvertMatrix(A, A_inv, A_det);
+
+        // Obtain the RBF shape functions 
+        noalias(rN) = prod(A_inv,Y);
```
EduardGomezEscandell: Same advice about solving rather than inverting

kratos/utilities/radial_basis_functions_utility.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+        // Interpolate solution
+        double Phi;
+        double norm_xi;
```
EduardGomezEscandell: No need for preallocation. Declare them when needed.

kratos/utilities/radial_basis_functions_utility.h
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * @brief Calculate the RBF value
+     * This function evaluates the Gaussian RBF for a norm
+     * @param x Norm of RBF argument (i.e. norm of radial vector)
+     * @param h Gaussian radius
+     * @return double The RBF value
+     */
+    static double EvaluateRBF(
+        double x,
```
EduardGomezEscandell:
```suggestion
        const double x,
```

kratos/utilities/radial_basis_functions_utility.cpp
```diff
@@ -0,0 +1,208 @@
...
+        // Invert the RBF interpolation matrix
+        double A_det;
+        Matrix A_inv(n_points,n_points);
+        MathUtils<double>::InvertMatrix(A, A_inv, A_det);
+
+        // Obtain the RBF shape functions 
+        noalias(rN) = prod(Phi,A_inv);
+
+        // Partition of unity
+        noalias(rN) =  rN/sum(rN);
```
EduardGomezEscandell: Could this block of code be replaced by the following line?

```C++
    CalculateShapeFunctions(rPoints, rX, h, rN);
```

This way we avoid duplicated code.

EduardGomezEscandell (changes requested):

kratos/utilities/radial_basis_functions_utility.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+        KRATOS_CATCH("");
+    }
+
+    double RadialBasisFunctionsUtility::CalculateShapeFunctionsAndInterpolation(
+        const Matrix& rPoints,
+        const array_1d<double,3>& rX,
+        const double h,
+        Vector& rN,
+        Vector Y)
```
EduardGomezEscandell:
```suggestion
        Vector& Y)
```
I think.

kratos/utilities/radial_basis_functions_utility.h
```diff
@@ -0,0 +1,80 @@
...
+        const Matrix& rPoints,
+        const array_1d<double,3>& rX,
+        Vector& rN);
+    
+    static double CalculateShapeFunctionsAndInterpolation(
+        const Matrix& rPoints,
+        const array_1d<double,3>& rX,
+        const double h,
+        Vector& rN,
+        Vector Y);
```
EduardGomezEscandell:
```suggestion
        Vector& Y);
```
Same as previous comment.

kratos/tests/test_radial_basis_functions_utility.py
```diff
@@ -0,0 +1,105 @@
+import os
+
```
EduardGomezEscandell:
```suggestion
```
Unused import

kratos/tests/test_radial_basis_functions_utility.py
SADPR: It is not necessary, we can comment it.

kratos/utilities/rbf_shape_functions_utility.h
```diff
@@ -0,0 +1,136 @@
...
+    /**
+     * @brief Calculate the RBF value
+     * This function evaluates the Gaussian RBF for a norm
+     * @param x Norm of RBF argument (i.e. norm of radial vector)
+     * @param h Gaussian radius
+     * @return double The RBF value
+     */
+    static double EvaluateRBF(
+        const double x,
+        const double h);
```
rubenzorrilla, line 73: I just noted that we can make this `private`. I'm assuming that you don't need/want to keep it `public` though.

rubenzorrilla:
It seems that the ill-conditioned test is failing.

rubenzorrilla:
@EduardGomezEscandell now it should be ready to go in.

kratos/utilities/rbf_shape_functions_utility.cpp
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        // Initialize the RBF interpolation matrix and RBF interpolated vector
+        Matrix A = ZeroMatrix(n_points,n_points);
+        Vector Phi = ZeroVector(n_points);
+
+        // Build the RBF interpolation matrix and RBF interpolated vector
+        for (std::size_t i_pt = 0; i_pt < n_points; ++i_pt) {
+            for (std::size_t j_pt = 0; j_pt < n_points; ++j_pt) {
+                const double norm_xij = norm_2(row(rPoints,i_pt)-row(rPoints,j_pt));
```
EduardGomezEscandell, line 62: This looks like it is symmetrical, could it be optimized?

kratos/utilities/rbf_shape_functions_utility.cpp
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        // Initialize the RBF interpolation matrix and RBF interpolated vector
+        Matrix A = ZeroMatrix(n_points,n_points);
+        double norm_xij;
+
+        // Build the RBF interpolation matrix and RBF interpolated vector
+        for (std::size_t i_pt = 0; i_pt < n_points; ++i_pt) {
+            for (std::size_t j_pt = 0; j_pt < n_points; ++j_pt) {
+                norm_xij = norm_2(row(rPoints,i_pt)-row(rPoints,j_pt));
```
EduardGomezEscandell, line 124: Same comment regarding symmetry

kratos/utilities/rbf_shape_functions_utility.cpp
```diff
@@ -0,0 +1,169 @@
...
+    void RBFShapeFunctionsUtility::CalculateShapeFunctions(
+        const Matrix& rPoints,
+        const array_1d<double,3>& rX,
+        const double h,
+        Vector& rN,
+        DenseQRPointerType pDenseQR)
+    {
+        KRATOS_TRY;
+
+        KRATOS_ERROR_IF(h < 1.0e-12) << "Reference distance close to zero." << std::endl;
```
EduardGomezEscandell, line 47: I'd either rename `h` to `reference_distance` or something similar; or change the error message to make it obvious `h` is the issue.

EduardGomezEscandell (approved):
Found some issues, small enough not to block

kratos/utilities/rbf_shape_functions_utility.cpp
RiccardoRossi: we should correct this, i think that Eduard's point is very valid!

PR #9782, opened 2022-03-22
[Fluid] Change in formulation of advective term in compressible Navier Stokes

EduardGomezEscandell:
**📝 Description**
Previously the formulation used the chain rule, so the advective fluxes `E(U)` were computed as:
```
∇·E(U) = A(U)∇U
```
where the Euler-Jacobian matrix is defined as `A(U):=∂E(U)/∂U`.

This transformation is no longer done. This requieres adding a condition since integrating results in a boundary term with shape functions `V`:
```
(V, ∇·E(U)) = - (∇V, E(U)) + ⟨V, E(U)·n⟩
                             ^~~~~~~~~~~
                             Boundary integral
```
Hopefully this will improve the behaviour of the solution near the boundaries.

E(U) contains the gradient of U, so a few extra methods have been added to the element so that condition can request its parent element to calculate the gradient of density (pre-existing), momentum (new) and total energy (new).

**🆕 Changelog**
- Changed formulation in symbolic element
- Created symbolic condition

**Dependencies**
Depends on #9820 for its unit tests.


rubenzorrilla:
I think that this is no longer required after @Marco1410 's findings.

PR #9783, opened 2022-03-22
[Core] Fixing const-correctness in kratos pointers

### LINKED ISSUE #9779: Incorrect const pointers in the kratos core
**Description**
There are many methods all over the core with signatures similar to:
```C++
const Type::Pointer T::pGetMember(...) const
```

The presence of a `const` makes it seem as though the pointed object is inmutable, but it is not. Indeed this code would be perfectly valid:
```c++
const T t();
const auto suposedly_const = t.pGetMember();
*suposedly_const += 1;
```
which clearly goes against the spirit of the const getter.

**Solution**
I argue that we should add a new typedef to the pointer definition:
```C++

#define KRATOS_CLASS_POINTER_DEFINITION(a)  \
/* Current */ \
typedef Kratos::shared_ptr<a > Pointer; \
typedef Kratos::shared_ptr<a > SharedPointer; \
typedef Kratos::weak_ptr  <a > WeakPointer; \
typedef Kratos::unique_ptr<a > UniquePointer; \
/* New  */ \
typedef Kratos::shared_ptr<const a > ConstPointer; \
typedef Kratos::shared_ptr<const a > ConstSharedPointer; \
typedef Kratos::weak_ptr  <const a > ConstWeakPointer; \
typedef Kratos::unique_ptr<const a > ConstUniquePointer // Maybe this one can be skipped
```
and something similar with `KRATOS_CLASS_INTRUSIVE_POINTER_DEFINITION`

And replace al const pGetters return types to the appropiate const pointer.

**Homework**
I don't mind implementing these changes myself, but I thought I would bring it up before putting in the work.

EduardGomezEscandell:
**📝 Description**
There are many `const pointer` that should be `pointer to const`. See https://github.com/KratosMultiphysics/Kratos/issues/9779 for a more detailed explanation.

**🆕 Changelog**
Defined:
  - `PointerToConst` (shared)
  - `PointerToConst` (intrusive)
  - `SharedPointerToConst`
  - `WeakPointerToConst`
  - `UniquePointerToConst`

Fixed return type of "const pGetters" from `const Type::Pointer` to `Type::PointerToConst`.

**Breaking changes**
The breaking changes were fixed in the applications. Users of Kratos outside the repository may need to remove `const` from some elements, conditions and modelparts to acomodate these changes. To avoid any chaos, a deprecation warning will be added for some time before merging this change. See #9790.

**Unfixed issue**
Method `Clone` of elements and conditions should not be `const`, since it includes a call to the non-const version of `pGetProperties`.

Fixing this would be too much (it would break practically every single element and condition). To avoid doing this work, I decided too keep `pGetProperties` with a the wrong return type `Properties::Pointer` rather than `Properties::PointerToConst`.

This is left outside the scope of this PR.


kratos/includes/shared_pointers.h
```diff
@@ -68,18 +68,27 @@ std::ostream& operator <<(std::ostream& rOStream, const Kratos::intrusive_ptr<T>
 } // namespace Kratos
 
 
-#define KRATOS_CLASS_POINTER_DEFINITION(a) typedef Kratos::shared_ptr<a > Pointer; \
-typedef Kratos::shared_ptr<a > SharedPointer; \
-typedef Kratos::weak_ptr<a > WeakPointer; \
-typedef Kratos::unique_ptr<a > UniquePointer
+#define KRATOS_CLASS_POINTER_DEFINITION(a)                \
+typedef Kratos::shared_ptr<a > Pointer;                   \
+typedef Kratos::shared_ptr<const a > ConstPointer;        \
```
roigcarlo: Please change the name to `PointerToConst` to prevent it to be confused with const shared<T>

applications/CompressiblePotentialFlowApplication/custom_processes/define_3d_wake_process.cpp
```diff
@@ -290,9 +290,9 @@ void Define3DWakeProcess::ComputeAndSaveLocalWakeNormal() const
 // of the wake surface in the wake direction. Note that the element size in span
 // direction is predetermined by the size of the conditions constituting the
 // trailing edge.
-void Define3DWakeProcess::ShedWakeSurfaceFromTheTrailingEdge() const
+void Define3DWakeProcess::ShedWakeSurfaceFromTheTrailingEdge()
 {
-    const Properties::Pointer pElemProp = mrStlWakeModelPart.pGetProperties(0);
```
RiccardoRossi, line 295: you could keep the const in the method and simply obtain a Properties::PointerToConst ? (or auto) no?

applications/ConvectionDiffusionApplication/custom_elements/embedded_laplacian_element.cpp
```diff
@@ -366,15 +366,15 @@ namespace EmbeddedLaplacianInternals {
 
 template <>
 ModifiedShapeFunctions::Pointer GetContinuousShapeFunctionCalculator<2, 3>(
-    const Element& rElement,
+    Element& rElement,
     const Vector& rNodalDistances)
 {
     return ModifiedShapeFunctions::Pointer(new Triangle2D3ModifiedShapeFunctions(rElement.pGetGeometry(), rNodalDistances));
```
RiccardoRossi, line 372: here it would be better to change the Triangle2D3ModifiedShapeFunctions instead

kratos/includes/shared_pointers.h
```diff
@@ -68,18 +68,27 @@ std::ostream& operator <<(std::ostream& rOStream, const Kratos::intrusive_ptr<T>
...
+typedef Kratos::weak_ptr<const a > WeakPointerToConst;      \
+typedef Kratos::unique_ptr<a > UniquePointer;               \
+typedef Kratos::unique_ptr<const a > UniquePointerToConst;
 
 namespace Kratos {
 template< class T > class GlobalPointer;
 }
 
-#define KRATOS_CLASS_INTRUSIVE_POINTER_DEFINITION(a) typedef typename Kratos::intrusive_ptr<a > Pointer; \
-typedef Kratos::GlobalPointer<a > WeakPointer; \
```
RiccardoRossi, line 81: note for myself. I think this definition is wrong. To be checked in a different PR/issue

applications/MeshingApplication/custom_utilities/mmg/mmg_utilities.cpp
```diff
@@ -4233,7 +4233,7 @@ void MmgUtilities<TMMGLibrary>::GenerateReferenceMaps(
     if (mDiscretization == DiscretizationOption::ISOSURFACE) {
         // Boundary conditions
         Condition const& r_clone_condition = KratosComponents<Condition>::Get("SurfaceCondition3D3N");
-        rRefCondition[10] = r_clone_condition.Create(0, r_clone_condition.pGetGeometry(), it_cond_begin->pGetProperties());
+        rRefCondition[10] = r_clone_condition.Create(0, const_cast<Condition&>(r_clone_condition).pGetGeometry(), it_cond_begin->pGetProperties());
```
RiccardoRossi, line 4236: mark this with a easily identfyable todo

applications/CompressiblePotentialFlowApplication/custom_processes/define_3d_wake_process.cpp
EduardGomezEscandell: Issue is when you follow the usage of pElemProp.
- It's used in `CreateWakeSurfaceNodesAndElements`
- Which uses it to call `CreateWakeSurfaceElements`
- Which uses it to call `ModelPart::CreateNewElement`, requiring a non-const version of the properties since I understand we want a reference to the same properties, not a copy.

EduardGomezEscandell: We do need a Properties::Pointer since, if you follow its usage, it ends up being passed to `ModelPart::CreateNewElement`.

However, since `pGetProperties` is not fixed by this PR, we can keep the `const` in the method.

kratos/modeler/copy_properties_modeler.cpp
```diff
@@ -94,7 +94,7 @@ void CopyPropertiesModeler::RecursivelyCopyProperties(
 template<class TContainerType>
 void CopyPropertiesModeler::ReplaceProperties(
     TContainerType& rContainer,
-    const ModelPart& rModelPart)
+    ModelPart& rModelPart)
```
ddiezrod, line 97: Just for me to understand it, why did you change this?

EduardGomezEscandell:
```c++
template<class TContainerType>
void CopyPropertiesModeler::ReplaceProperties(
    TContainerType& rContainer,
    ModelPart& rModelPart)
{
    block_for_each(rContainer, [&](typename TContainerType::value_type& rEntity){
        auto properties_id = rEntity.GetProperties().Id();
        auto p_properties = rModelPart.pGetProperties(properties_id);
//                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
//                                     This method is not const
        rEntity.SetProperties(p_properties);
    });
}
```

It wouldn't compile otherwise. `pGetProperties` is not const (for modelparts) as of this PR.

This is actually an inconsistency within this PR. Modelpart is:
- `Modelpart::pGetProperties -> Properties::PointerToConst` is const
- `Modelpart::pGetProperties -> Properties::Pointer` is not const (This is the one being used here)

Element:
- `Element::pGetProperties -> Properties::Pointer` is const but shouldn't, to avoid breaking `Element::Create`
- `Element::pGetProperties -> Properties::PointerToConst` cannot exist as it would ambiguosly overload the previous one.

Condition:
- `Condition::pGetProperties -> Properties::Pointer` is const but shouldn't, to avoid breaking `Condition::Create`
- `Condition::pGetProperties -> Properties::PointerToConst` cannot exist as it would ambiguosly overload the previous one.


The `Condition` and `Element` cases will be tackled in a future PR as they would break pretty much every single element and condition in Kratos.

The `Modelpart` one is done in this PR. Alternatively we could keep it "wrong" for consistency and fix it the future PR.

I don't mind either way.

ddiezrod: Ok I see, thanks for the explanation! I don't have a preference so from my side, do it as you prefer

rubenzorrilla:
@KratosMultiphysics/technical-committee has been discussing about this again.

We are in favour of the change as current behaviour might be misleading.

However, we're afraid of how this propagates through the code. In this regard, we would like to know the opinion of the @KratosMultiphysics/implementation-committee.

AlejandroCornejo:
Speaking on my own (to be discussed in the @KratosMultiphysics/implementation-committee ), I agree that the current notation is confusing (I was not sure of this behaviour until now!) so I would vote in favour of changing it. Especially after reading the very insightful https://github.com/KratosMultiphysics/Kratos/issues/9779
Thank you @EduardGomezEscandell

pooyan-dadvand:
Ping @KratosMultiphysics/implementation-committee

pooyan-dadvand:
@KratosMultiphysics/technical-committee agrees about this change, however we think that @KratosMultiphysics/implementation-committee should take the lead on finishing it.

EduardGomezEscandell:
Feel free to start it from zero, this branch is probably a bit too old now.

If you chose to continue on this branch, or use it as inspiration, don't hesitate to ask about any of the changes I made (although I can't promise I'll remember the reasoning 😅).

roigcarlo:
Hey, we still want to merge this. reopening.

EduardGomezEscandell:
Oh I thought you were starting from zero :sweat_smile: 

I was just trying to remove my old PRs.

Issue #9784, opened 2022-03-22
Nightly build is failing


miguelmaso:
**Description**
`RuntimeError: Error: Kratos components missing "AssociativePlasticDamageModel3DVonMisesVonMises"`

seems to be related to #9770 

**Scope**
- NightlyBuild
- ConstitutiveLawsApplication

AlejandroCornejo:
![imagen](https://user-images.githubusercontent.com/26402605/159648693-80f10c98-7744-4fd5-be22-5a9494e6d3f1.png)

loumalouomega:
> ![imagen](https://user-images.githubusercontent.com/26402605/159648693-80f10c98-7744-4fd5-be22-5a9494e6d3f1.png)

Deeply concerned...

AlejandroCornejo:
solved I think

ddiezrod:
I think there is another test failing from ConstitutiveLawsApp 
see https://github.com/KratosMultiphysics/Kratos/runs/5659661998?check_suite_focus=true

AlejandroCornejo:
On it...

AlejandroCornejo:
Should be fine now!

PR #9785, opened 2022-03-23
Solving nightly build

AlejandroCornejo:
**📝 Description**
Minor change to solve nightly build https://github.com/KratosMultiphysics/Kratos/issues/9784


miguelmaso (approved):

loumalouomega (approved):
Maybe we can use both names for the sake of consistency (it is possible, only is registering twice)

AlejandroCornejo:
> Maybe we can use both names for the sake of consistency (it is possible, only is registering twice)

It's another option, honestly eventually I will change the name of those CL that do not use any plastic potential (damage) because I've seen that it's confusing for the users.

PR #9690, opened 2022-02-21
[Fluid] Fixed Fulldebug error in compressible symbolic tests

EduardGomezEscandell:
**📝 Description**
Matrices were intialized with size `(block_size, num_nodes)` when they should have been `(num_nodes, block_size)`. This initially remained silent but the error was caught in FullDebug.

Noticed by @marcnunezc [here](https://github.com/KratosMultiphysics/Kratos/pull/9687#issuecomment-1046660890).

**🆕 Changelog**
Fixed matrix initialization in tests.

PR #9787, opened 2022-03-23
[CI] adding centos nightly tests

ddiezrod:
**📝 Description**
This MR adds a new job to run the nightly tests in centos. We've had some problems in the past in altair side with this, so I think it is safer to add this here.

FYI @KratosMultiphysics/altair

### DIFF
--- a/.github/workflows/nightly_build.yml
+++ b/.github/workflows/nightly_build.yml
@@ -136,3 +136,26 @@ jobs:
         set PYTHONPATH=%PYTHONPATH%;%GITHUB_WORKSPACE%/bin/%KRATOS_BUILD_TYPE%
         set PATH=%PATH%;%GITHUB_WORKSPACE%/bin/%KRATOS_BUILD_TYPE%/libs
         python kratos/python_scripts/run_tests.py -l nightly -c python
+
+  centos-nightly:
+    runs-on: ubuntu-latest
+    env:
+      KRATOS_BUILD_TYPE: Custom
+
+    container:
+      image: kratosmultiphysics/kratos-image-ci-centos7-python35:latest
+      options: --user 1001
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Build
+      run: |
+        cp .github/workflows/centos_configure.sh centos_configure.sh
+        bash centos_configure.sh
+
+    - name: Running nightly tests
+      run: |
+        export PYTHONPATH=${PYTHONPATH}:${GITHUB_WORKSPACE}/bin/Custom
+        export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:${GITHUB_WORKSPACE}/bin/Custom/libs
+        python3.5 kratos/python_scripts/run_tests.py -l nightly -c python3.5


roigcarlo (approved):
I suppose there is no problem. :+1:

roigcarlo:
@ddiezrod Nightlies are not need for merge, you an go ahead.

@KratosMultiphysics/technical-committee at some point we will have to discuss why nightly tests are run at every PR. I know it was said that this would prevent them from failing but this is not the case and are very expensive to maintain. And it completely invalidates the concept of having small and nightly tests to begin with...

EduardGomezEscandell:
It is quite strange, the test is failing in centos but not on ubuntu or windows. The CI claims to fail in function `test_substitute_rho_proj_2D`, line 244.

Yet when I generate the file, line 244 is deep inside function `test_substitute_mom_proj_2D`.

For some reason the generated file seems to be quite diferent in centos, maybe due to the Python version? I'd have to run it on centos but I honestly have no way how right now. 

The error thrown (bad index) reminds me of https://github.com/KratosMultiphysics/Kratos/pull/9690.

ddiezrod:
I will merge this and will try to fix the problems in FluidApp in another MR. About the problems in nightly ci, I agree with you @roigcarlo but I do not see an easy solution. Probably, the only way is to have a more robust small ci, with very fast tests that cover pretty much every function/method instead of adding bigger tests to the nightly suite. For example, I think for every bug we find, we should add a small test that covers it. And finally, of course, developers should be more careful when we merge things. Even if you can merge a PR, you should take a look to the nightly checks to make sure nothing got broken.

roigcarlo:
Haven't look in detail but for what I understand the problem is that you are extending some tests with template files (or at least loading them dynamically from a string)

I suspect that this causes the context of the test not be available during the unit-test report (I would bet because it has been unloaded). I'd said it crashes because the getattr(foo) is no longer able to return anything because the class is essentially empty.

But can't confirm 100%

ddiezrod:
@KratosMultiphysics/technical-committee I think I have found out why nightly tests are getting broken so often (you might already know this but I think it is quite explanatory)
Nightly checks are done only when a few apps are changed, see https://github.com/KratosMultiphysics/Kratos/blob/1c9589928fc6aa0446532d8ba9bb8f789b8e5a7c/.github/workflows/nightly_build.yml#L5

 - '.github/workflows/nightly_build.yml'
    - '.github/workflows/configure.sh'
    - '.github/workflows/configure.cmd'
    - 'kratos/**'
    - 'external_libraries/**'
    - 'cmake_modules/**'
    - 'CMakeLists.txt'
    - 'applications/LinearSolversApplication/**'
    - 'applications/MetisApplication/**'
    - 'applications/TrilinosApplication/**'

However, the nightly checks compile many more applications:
ConvectionDiffusionApplication
LinearSolversApplication
StructuralMechanicsApplication
FluidDynamicsApplication
MeshMovingApplication
DEMApplication
CSharpWrapperApplication
MetisApplication
TrilinosApplication
ShapeOptimizationApplication
CoSimulationApplication
CableNetApplication
RANSApplication
MappingApplication
FSIApplication
MeshingApplication
CompressiblePotentialFlowApplication
HDF5Application
ContactStructuralMechanicsApplication
IgaApplication
ParticleMechanicsApplication
ChimeraApplication
MultilevelMonteCarloApplication
StatisticsApplication
SwimmingDEMApplication
ConstitutiveLawsApplication
RomApplication
ShallowWaterApplication

So if for example, someone breaks something in ConstitutiveLawsApp (what just happened to @AlejandroCornejo ) they do not see at that moment because nightly checks are not run for their PR.

Maybe adding all these applications to the CI would be too much of a burden for the system, but I think I would at least add the "corest" ones: FluidDyn, Structural, Mapping, Meshing, ConvectionDIffusion

roigcarlo:
I am aware but in fact I think we should go in the opposite direction. See #9796 

At the end the nightly test should be in the nightly level because are either slow or not important. If we have them rung along with the normal ones we are essentially doing the same twice, hence, it makes no sense to have the distinction.

So either we merge them together or we remove the run.

Issue #9791, opened 2022-03-23
DEM App - Access Violation


VeronikaSinger:
Hi all,

I have troubles running examples with the DEM Application as I always get the error that a flag is not defined. 
The debugger reports an error ( Unhandled exception thrown: read access violation.
this was 0x10.) for this line:
       ` [const bool tag_condition = p_particle->IsNot(TO_ERASE)];`
[https://github.com/KratosMultiphysics/Kratos/blob/master/applications/DEMApplication/custom_utilities/create_and_destroy.cpp#:~:text=const%20bool%20tag_condition%20%3D%20p_particle%2D%3EIsNot(TO_ERASE)%3B](url)

This is an example file: [DEM_Files.zip](https://github.com/KratosMultiphysics/Kratos/files/8333979/DEM_Files.zip)
Are there any ideas how to resolve that issue? 

Best regards
Veronika

KlausBSautter:
Thanks, @VeronikaSinger I have the same problem. Maybe this is also interesting for @maceligueta.

maceligueta:
I will take a look today. Thanks for reporting.

maceligueta:
I found the source of error and I am fixing it.

PR #9794, opened 2022-03-24
Methods converted into function templates, so they can be applied to isolated particles or clusters

maceligueta:
**📝 Description**
As reported in #9791, the code was crashing when trying to acces information from a cluster. The dynamic cast was being done to SphericParticle, hence the problem when it is a Cluster3D

**🆕 Changelog**
Some methods have been 'templated' ('templatized'?) to accept single particles or clusters. This should fix the issue

If it works, this PR solves #9791 @VeronikaSinger @KlausBSautter


VeronikaSinger:
Dear @maceligueta 
unfortunately it does not compile on my windows machine. The error is reported here: https://github.com/KratosMultiphysics/Kratos/blob/master/applications/DEMApplication/custom_python/add_custom_utilities_to_python.cpp#:~:text=.def(%22DestroyParticlesOutsideBoundingBox%22%2C%20%26ParticleCreatorDestructor%3A%3ADestroyParticlesOutsideBoundingBox)

maceligueta:
The last trick did not work @roigcarlo . We can always move everything to the header file, no?

maceligueta:
It seems it is compiling now. I checked that it does not crash in linux. @VeronikaSinger can you check if it crashes in Windows? Also please check if the overall behaviour is as expected. I need to write a test for this. If merging is urgent I can add the test in a different PR.

KlausBSautter:
thanks @maceligueta I'll check my simulations

VeronikaSinger:
Thanks for fixing the issue. The example now works as expected.

roigcarlo (dismissed):

KlausBSautter:
Let me check my example, for a second. I am still compiling :/

KlausBSautter:
Thanks, it works :). I'll let you merge as soon as you are ready.

maceligueta:
The example provided by @VeronikaSinger showed a problem with conflicting ids when printing for GiD. I fixed it for that example, but now the thests are failing. I am on it.

maceligueta:
Finally, everything seems to work fine. The example provided by @VeronikaSinger was added as a test and some extra improvements were made to prevent (to some extent) the repetition of similar problems.
This can be merged in my opinion.

VeronikaSinger:
Thanks @maceligueta my examples are all working perfectly now.

VeronikaSinger (approved):

PR #9796, opened 2022-03-24
NightlyBuilds Reschedule

roigcarlo:
**📝 Description**
Gonna open the pandora box: This PR changes the criteria for the nightly runs to trigger. While it is true that running nightly builds is useful with changes in the core and some selected apps, there are some problems:

1 - ~100% of the PR right now contains a change that triggers the builds
2 - These build tests are extremely slow to run and consume resources that are starting to lack (see #9786 for example we already had to start increasing time for the regular builds)
3 - The nightly tests keeps failing on a regular basis. This gives me the impression that nobody is actually caring about what they report unless they see the badge failing in the main page. 

Remember that the CI is not your compiler of choice, is it there to validate changes. If you make a change that could break the tests they should be run in your machines first.

Changing back this aims to:
1 - Make PR that change nightly workflow require the nightly build to pass before merge. (which cannot be done now because it runs at virtually every PR)
2 - Lower the burden on the CI machines.
3 - Revert #9786

Regardless of the changes, the nightly will keep running... nightly, as it was intended in the first place.


roigcarlo:
After some discussion in the @KratosMultiphysics/technical-committee we have decided that it will be best to simply remove the concept of small and run everything at every PR.

Now limitations in general are the same as the old nightly ones. No special builds will exist for nightly after that as they are now the default. The expected time for compilation is about 1h.

loumalouomega:
> After some discussion in the @KratosMultiphysics/technical-committee we have decided that it will be best to simply remove the concept of small and run everything at every PR.
> 
> Now limitations in general are the same as the old nightly ones. No special builds will exist for nightly after that as they are now the default. The expected time for compilation is about 1h.

What is the difference in running both cases¿?

loumalouomega:
> If we leave it like it was before we were running and compiling the small suit twice. If we don't run enforce the nightly they are always broken.
> 
> This is the best compromise we can think of

Understood

roigcarlo:
@ddiezrod While updating this I've noticed that centos is using python 3.5. I did not saw it when It was introduced but in principle we do not support 3.5 anymore, and with that check being required we are limiting the options in the core. 

Is there any way we can mark tests or tools using 3.6 features so there are not run in Centos? I don't mind patching the core for a while but 80% of the apps are using fstrings for example.

roigcarlo:
I've taken the opportunity to include a very basic profiler in the test executer so we can have some insight about what tests are taking most of the time:

image

As you can se, once the nightly tests are included the amount of tests jumps up to about 2100. From those ~20 takes roughly 50% of the execution time.

Tentatively I've move the following tests to validation, and adapted the strcuture of the tests in the apps that did not have that suite:
```bash
test_execution (iga_test_factory.Shell5pHierarchicLinearScordelisTest) ... 197.736s ...ok
test_mlmc_Kratos (test_xmcAlgorithm.TestXMCAlgorithm) ... 162.740s ...ok
testBFECCConvection (bfecc_convection_test.BFECCConvectionTest) ... 161.793s ...ok
test_ALEFluidSolverOnSubdomains (test_ale_fluid_solver.ALEFluidSolverTest) ... 134.653s ...ok
test_execution (structural_mechanics_test_factory.Simple3D2NBeamCrTest) ... 129.208s ...ok
testSymbolicTriangle (compressible_navier_stokes_symbolic_generator_formulation_test.CompressibleNavierStokesSymbolicGeneratorFormulationTest) ... 126.594s ...ok
test_execution (structural_mechanics_test_factory.Simple3D2NBeamCrDynamicTest) ... 80.543s ...ok
test_execution (shape_optimization_test_factory.algorithm_penalized_projection_test) ... 62.608s ...ok
test_execution (shape_optimization_test_factory.algorithm_trust_region_test) ... 58.882s ...ok
testBFECCElementalLimiterConvection (bfecc_convection_test.BFECCConvectionTest) ... 58.357s ...ok
test_execution (iga_test_factory.Shell5pHierarchicNonLinearThickBeamTest) ... 57.580s ...ok
testSymbolicQuadrilateral (compressible_navier_stokes_symbolic_generator_formulation_test.CompressibleNavierStokesSymbolicGeneratorFormulationTest) ... 55.136s ...ok
test_execution (shape_optimization_test_factory.algorithm_steepest_descent_test) ... 51.534s ...ok
test_MPM_FEM_beam_penalty (co_simulation_test_factory.TestSmallCoSimulationCases) ... 48.791s ...ok
test_membrane_3d4n_dynamic_consistent_mm (test_patch_test_membrane.DynamicPatchTestMembrane) ... 40.484s ...ok
testSlipCylinder (adjoint_fluid_test.AdjointFluidTest) ... 35.767s ...ok
test_execution (structural_mechanics_test_factory.MembraneHemisphereTests) ... 35.575s ...ok
testCylinder (adjoint_fluid_test.AdjointFluidTest) ... 35.568s ...ok
test_membrane_3d4n_dynamic (test_patch_test_membrane.DynamicPatchTestMembrane) ... 35.507s ...ok
test_execution (structural_mechanics_test_factory.ShellT3AndQ4NonLinearStaticStructHingedCylRoofSnapthroughTests) ... 34.160s ...ok
test_execution (shape_optimization_test_factory.opt_process_step_adaption_test) ... 33.162s ...ok
```
As you can see the rule has been more than >30s in release (We can keep it under 60 if you want, but this will add 25% more time in the CI). If it is absolutely necessary to run any of these in a regular basis through the CI. If it is absolutely necessary to add one of these in the CI we can do it, but I would encourage to double check if the same checks can be done quicker. 

For example executing only a couple of iteration in the solver, even if the results do not make sense, test do not exists to check the correctness of the solution, only that the result is the expected (for the later you have the validation suit)

roigcarlo:
@philbucher Any ideas about the cosim fails? I cannot reproduce or identify the faulty test in my machine. Seems to be an incorrect assignment of a matrix or a vector but Don't know where.

philbucher:
> @philbucher Any ideas about the cosim fails? I cannot reproduce or identify the faulty test in my machine. Seems to be an incorrect assignment of a matrix or a vector but Don't know where.

Brief answer: don’t know, can take a look in the next days. This is probably the most important test in cosim, having it fail I would consider as blocker

roigcarlo:
@philbucher if you can use it as a clue, seems that it could be related with the FSI, but still unsure. For all I know both test fail at the same point (and it really really looks like an assigment of of a vector/matrix of incorrect size)

roigcarlo:
No no, I mean the FSI tests

rubenzorrilla (approved):

roigcarlo:
Essentially we gave green light to people to start using fstrings, and there has been a couple of PR were we needed to roll back some of those changes.

If you are planning on moving to 3.8 no problem :+1:

Issue #2165, opened 2018-05-23
Policy on using numpy, h5py and other useful Python modules


maceligueta:
I am opening this issue to discuss about the possibility of adding full support to the most common scientific Python modules. 
The advantages are quite clear to those who use them. But what about the inconvenients, can they fit in a next Release? Can they be added to Travis?

maceligueta:
Just referencing @KratosMultiphysics/team-maintainers .

pooyan-dadvand:
Do you mean keeping compatibility with them or also let dependency to them?

maceligueta:
I was just thinking of packing those modules with the Releases. Also adding them to the tests of some applications.

RiccardoRossi:
from my point of view this would be nice, however packing those is very difficult if we want to distribute our own python. For example numpy depends on blas and lapack and may not be easy to pack.

in my opinion one possible way forward would be to distribute docker-type packages, which contain "native" python as well as the packages it needs. that could include the gid in the docker package...

roigcarlo:
For the releases I would be very careful on adding things like numpy, as it require additional shared libs that are often compiled by pip.

But its also true that, for pure python modules, using pybind11 has made the things easier so maybe we can consider this. 

We should add this to the project and discuss it next week

pooyan-dadvand:
In @KratosMultiphysics/technical-committee we have not seen how we can get the numpy especially in windows. Another complexity is the GiD retrieve mechanism.

Compiling it with `run_kratos.exe` considered to be impossible to maintain.

maceligueta:
I just talked to @roigcarlo  about this and we realized that _nuitka_ or _cx_freeze_ could help. Both the Linux and the Windows release could be packed with these utilities, just like it is being done with SpreadDEM or Click2Cast... Actually, both have options for embedding all the necessary modules and libraries to make them standalone.

loumalouomega:
Did we solve this issue?

RiccardoRossi:
Nope,this is open I think

pooyan-dadvand:
In committee a the idea of having a freezed python version of "run_kratos" packed with numpy and hdf5 is proposed and seems to be viable. 

To be experimented by @roigcarlo and @maceligueta

loumalouomega:
FYI I found this:

https://github.com/lava/matplotlib-cpp

philbucher:
are there any advances here?
I would also be interested :)

roigcarlo:
[...]

PR #9797, opened 2022-03-24
[Core] Sparse matrices - fix error in centos and skip tests if numpy is not available

ddiezrod:
**📝 Description**
Two things related to latest changes in sparse matrices.
- We were getting an error in centos mpi build as we are trying to bing an lvalue with a &&. 
- I added a skip in two tests that use numpy if it is not available, this gives us some headached in altair CI.


RiccardoRossi (commented):
ok, even though we are now assuming numpy IS available in principle

ddiezrod:
> ok, even though we are now assuming numpy IS available in principle

I know, but the process of installing numpy is still manual in altair side so this will help us until we have dockers in that side too.

ddiezrod:
can you approve it if you are ok? @RiccardoRossi

kratos/mpi/tests/test_distributed_sparse_matrices.py
```diff
@@ -63,9 +67,6 @@ class TestDistributedSparseMatrices(KratosUnittest.TestCase):
         (17,10):1.0,(17,36):1.0,(17,7):1.0,(10,17):1.0,(10,36):1.0,(10,7):1.0,(36,17):1.0,(36,10):1.0,(7,17):1.0,(7,10):1.0,(25,30):1.0,(25,9):1.0,(14,30):1.0,(30,25):1.0,(30,14):1.0,(30,9):1.0,(9,25):1.0,(9,30):1.0
     }
 
-    bref_all = np.array([1,3,2,3,3,3,5,6,4,4,2,4,3,1,6,3,5,3,6,3,0,1,4,1,4,4,6,1,5,2,2,3,1,4,1,5,2,4,1,3])
```
RiccardoRossi, line 66: why did u take out this one?

ddiezrod: its not being used anywhere and it gives problems as it is not inside the skipped test and it uses numpy.

rubenzorrilla (changes requested):

ddiezrod:

roigcarlo (approved):

roigcarlo:
Just to comment, in principle from ours side is ok, but considering than now numpy is basically a dependency for Kratos, so this dependency in particular should be solved at CI side

rubenzorrilla (approved):
Approving. In any case still to be discussed in @KratosMultiphysics/technical-committee to establish a clear position about.

ddiezrod:
> Approving. In any case still to be discussed in @KratosMultiphysics/technical-committee to establish a clear position about.

Merging. Anyway, we will add numpy to centos image. It will take some time but it will avoid these problems in the future.

rubenzorrilla:
We discussed in about this in @KratosMultiphysics/technical-committee as well as in  #2165. The position of @KratosMultiphysics/technical-committee is that `numpy` and `h5py` are assumed to be there. Having said this, we accept this change as a temporary solution that must be reverted after you guys add `numpy` to your centos image.

PR #9800, opened 2022-03-25
[Mapping] Serial output process - error in python35

ddiezrod:
**📝 Description**
The f-strings are only available in python 3.6 and forward. We are using python3.5 in centos so this is throwing an error.


philbucher (approved):
I never thought anyone except @mpentek or me would use this :)
Thx for the fix

PR #9804, opened 2022-03-29
[ShapeOpt] fix zero constraint gradient

armingeiser:
**📝 Description**
Checks if constraint gradient is 0 before considering as active

### DIFF
--- a/applications/ShapeOptimizationApplication/python_scripts/algorithms/algorithm_gradient_projection.py
+++ b/applications/ShapeOptimizationApplication/python_scripts/algorithms/algorithm_gradient_projection.py
@@ -272,7 +272,7 @@ class AlgorithmGradientProjection(OptimizationAlgorithm):
         constraint_value = self.communicator.getStandardizedValue(identifier)
         if constraint["type"].GetString() == "=" or constraint_value >= 0:
             gradient_norm = self.optimization_utilities.ComputeMaxNormOfNodalVariable(
-                self.design_surface, self.constraint_gradient_variables[identifier]
+                self.design_surface, self.constraint_gradient_variables[identifier]["mapped_gradient"]
             )
             if gradient_norm == 0.0:
                 KM.Logger.PrintWarning("ShapeOpt", f"Gradient for constraint {identifier} is 0.0 - will not be considered!")
--- a/applications/ShapeOptimizationApplication/python_scripts/algorithms/algorithm_gradient_projection.py
+++ b/applications/ShapeOptimizationApplication/python_scripts/algorithms/algorithm_gradient_projection.py
@@ -274,7 +274,7 @@ class AlgorithmGradientProjection(OptimizationAlgorithm):
             gradient_norm = self.optimization_utilities.ComputeMaxNormOfNodalVariable(
                 self.design_surface, self.constraint_gradient_variables[identifier]["mapped_gradient"]
             )
-            if gradient_norm == 0.0:
+            if gradient_norm <= 1e-16:
                 KM.Logger.PrintWarning("ShapeOpt", f"Gradient for constraint {identifier} is 0.0 - will not be considered!")
                 return False
             return True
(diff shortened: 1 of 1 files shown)


applications/ShapeOptimizationApplication/python_scripts/algorithms/algorithm_gradient_projection.py
```diff
@@ -270,9 +270,13 @@ def __getActiveConstraints(self):
...
         constraint_value = self.communicator.getStandardizedValue(identifier)
-        if constraint["type"].GetString() == "=":
-            return True
-        elif constraint_value >= 0:
+        if constraint["type"].GetString() == "=" or constraint_value >= 0:
+            gradient_norm = self.optimization_utilities.ComputeMaxNormOfNodalVariable(
+                self.design_surface, self.constraint_gradient_variables[identifier]["mapped_gradient"]
+            )
+            if gradient_norm == 0.0:
+                KM.Logger.PrintWarning("ShapeOpt", f"Gradient for constraint {identifier} is 0.0 - will not be considered!")
```
sunethwarna: Is this gonna be a strict check, or should it be 
```suggestion
            if gradient_norm <= 1e-12:
                KM.Logger.PrintWarning("ShapeOpt", f"Gradient for constraint {identifier} is 0.0 - will not be considered!")
```

sunethwarna (commented):
I only have one minor comment

applications/ShapeOptimizationApplication/python_scripts/algorithms/algorithm_gradient_projection.py
armingeiser: Good point. From my experience 0.0 works in most cases since we really assign 0.0 in the sensitivity computation. But to be on the save side, we could also make it 1e-16 (i would prefer to make it very small)

armingeiser:
```suggestion
            if gradient_norm <= 1e-16:
                KM.Logger.PrintWarning("ShapeOpt", f"Gradient for constraint {identifier} is 0.0 - will not be considered!")
```

applications/ShapeOptimizationApplication/python_scripts/algorithms/algorithm_gradient_projection.py
```diff
@@ -270,9 +270,13 @@ def __getActiveConstraints(self):
...
         identifier = constraint["identifier"].GetString()
         constraint_value = self.communicator.getStandardizedValue(identifier)
-        if constraint["type"].GetString() == "=":
-            return True
-        elif constraint_value >= 0:
+        if constraint["type"].GetString() == "=" or constraint_value >= 0:
+            gradient_norm = self.optimization_utilities.ComputeMaxNormOfNodalVariable(
+                self.design_surface, self.constraint_gradient_variables[identifier]["mapped_gradient"]
+            )
+            if gradient_norm <= 1e-16:
```
philbucher:
```suggestion
            if math.isclose(gradient_norm, 0.0):
```

probably even better, I recently learned abt this

armingeiser: since the reference here is 0, it will do the same - comparing absolute value.

dbaumgaertner (dismissed):
I took the chance for a review here....  ;) Good from my side. Makes me just think that in gradient-based optimization using the projection algorithm one actually can't control a constraint if its gradient vanishes, right?

applications/ShapeOptimizationApplication/python_scripts/algorithms/algorithm_gradient_projection.py
Igarizza:
```suggestion
            if math.isclose(gradient_norm, 0.0, abs_tol=1e-16):
```
as a compromise =)

armingeiser: There seems to be much love for the isclose function =). I am convinced 🤣

armingeiser:
> one actually can't control a constraint if its gradient vanishes, right?

Exactly, if there is no constraint gradient, the opt. algorithm has no chance to do projection/correction w.r.t that constraint. 
This is for a very special case, where the gradient can be zero due to the aggregation of geometric constraints, normally this should not happen.

dbaumgaertner (approved):

PR #9805, opened 2022-03-30
[Core] Non-historical parallel distance calculation

rubenzorrilla:
**📝 Description**
Following #9780 this makes the parallel distance calculator compatible with non-historical variables. Note that this is way useful as it avoids adding `DISTANCE` to the nodal database, something that makes the usage of this kind of processes way intrusive as this needs to be done at the solver level.

As both historical and non-historical databases are used in the original algorithm (the origin input distance is saved in the non-historical one), I required to ~~create a new~~ migrate the `AUX_DISTANCE` variable in the `FluidDynamicsApplication` to the core so the variable storing the distance (potentially `DISTANCE`) and `AUX_DISTANCE` can coexist. 

There were no tests for this process so I added two, one for the current historical calculation and one for the new non-historical one.

Current behavior is kept by default.


EduardGomezEscandell (approved):
Looks good to me. Great to see the new unit test.

**Performance discussion**
I wonder what is the performance diference between an early chosen lambda and and a template. Something such as:
```C++
// In Execute()
switch(mNodalDatabase)
{
case Historical: ExtendDistancesByLayer<Historical>(...); break;
case NonHistorical: ExtendDistancesByLayer<NonHistorical>(...); break;
}
```
Where ExtendDistancesByLayer would call `GetDistance<TNodalDatabase>`.

Of course the point would be for the compiler to inline `GetDistance`. Lambdas cannot be inlined by the compiler (since they're not known at compile-time).

Either way, this effect is unlikely to be large, and not worth rewriting it all for. Just something to discuss for the future.

**CI**
Looks like only the CoSim test is failing, which is failing in master as well so I'm quite confident it's not this PR's fault.

rubenzorrilla:
> Looks good to me. Great to see the new unit test.
> 
> **Performance discussion** I wonder what is the performance diference between an early chosen lambda and and a template. Something such as:
> 
> ```c++
> // In Execute()
> switch(mNodalDatabase)
> {
> case Historical: ExtendDistancesByLayer<Historical>(...); break;
> case NonHistorical: ExtendDistancesByLayer<NonHistorical>(...); break;
> }
> ```
> 
> Where ExtendDistancesByLayer would call `GetDistance<TNodalDatabase>`.
> 
> Of course the point would be for the compiler to inline `GetDistance`. Lambdas cannot be inlined by the compiler (since they're not known at compile-time).
> 
> Either way, this effect is unlikely to be large, and not worth rewriting it all for. Just something to discuss for the future.
> 
> **CI** Looks like only the CoSim test is failing, which is failing in master as well so I'm quite confident it's not this PR's fault.

Maybe I'm missing something from your comment but for this I think we'd need partial template specialization (c++14...). Otherwise I don't see how to do it without adding an extra class template argument. In any case, as you point this has most probably a negligible impact.

PR #9812, opened 2022-03-31
[Core] Core QR decomposition (from AMGCL) API

rubenzorrilla:
**📝 Description**
So far the `qr_utilities` in the core were using their custom API. Recently we added a base virtual class for QR decompositions. In this PR I'm interfacing the QR in the Kratos core to use this.

Important thing to comment is that the QR decomposition we have in the Core is from AMGCL library. Years ago, when we included it AMGCL was not part of the Kratos core, so basically what we did in the `qr_utilities.h` was to copy & paste it from the library. Now the AMGCL is part of Kratos core, so what this PR does is to interface to the one in `external_libraries/amgcl/detail/qr.hpp `. With this we automatically make sure that the QR is updated each time we update the library (the old one we were using so far is way outdated because of this...).

In an upcoming PR I'll remove the old `qr_utility.h` as well as its usage.

EDIT: Tests have been added accordingly.

### DIFF
--- a/kratos/utilities/dense_householder_qr_decomposition.h
+++ b/kratos/utilities/dense_householder_qr_decomposition.h
@@ -92,7 +92,7 @@ public:
     {
         // Set input data
         // Note that we need a copy as QR decomposition is modifying the input
-        mpA = make_unique<MatrixType>(rInputMatrix);
+        mpA = std::make_unique<MatrixType>(rInputMatrix);
         DataType *p_0_0 = &((*mpA)(0,0));
         const std::size_t m = rInputMatrix.size1();
         const std::size_t n = rInputMatrix.size2();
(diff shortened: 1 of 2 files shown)


kratos/utilities/dense_householder_qr_decomposition.h
```diff
@@ -0,0 +1,355 @@
...
+     */
+    void MatrixR(MatrixType& rMatrixR) const override
+    {
+        // Check that QR has been already computed
+        KRATOS_ERROR_IF(!mpA) << "QR decomposition not computed yet. Please call 'Compute' before 'MatrixR'." << std::endl;
+
+        // Check input size
+        const std::size_t n = mpA->size2();
+        if (rMatrixR.size1() != n || rMatrixR.size2() != n) {
+            rMatrixR.resize(n,n);
```
RiccardoRossi: resize,false

rubenzorrilla:
@RiccardoRossi as we discussed, this avoids the copy of A. Hence, A is modified while doing the QR so if the original one is required afterwards a copy must be kept outside.

What I think is impossible is to avoid having the internal and our R and Q... The point is that our own instance cannot be passed from outside in the AMGCL QR case.

RiccardoRossi (approved):

kratos/utilities/dense_householder_qr_decomposition.h
rubenzorrilla: It is already done. See the outdated badge above.

PR #9819, opened 2022-04-02
added getElementalGradient function

e-dub:
**📝 Description**
Quick fix by adding GetElementalGradient function analogously to GetNodalGradient to allow usage like:
`stressNablaE = response_function.GetElementalGradient(sm.YOUNG_MODULUS_SENSITIVITY)`

**🆕 Changelog**
- Added function GetElementalGradient to structural_response.py

### DIFF
--- a/applications/StructuralMechanicsApplication/python_scripts/structural_response.py
+++ b/applications/StructuralMechanicsApplication/python_scripts/structural_response.py
@@ -92,6 +92,20 @@ class StrainEnergyResponseFunction(ResponseFunctionInterface):
             gradient[node.Id] = node.GetSolutionStepValue(variable)
         return gradient
 
+    def GetElementalGradient(self, variable):
+        if variable not in [
+                            StructuralMechanicsApplication.YOUNG_MODULUS_SENSITIVITY,
+                            StructuralMechanicsApplication.THICKNESS_SENSITIVITY,
+                            StructuralMechanicsApplication.I22_SENSITIVITY,
+                            StructuralMechanicsApplication.I33_SENSITIVITY,
+                            ]:
+            raise RuntimeError("GetElementalGradient: No gradient for {}!".format(variable.Name))
+        gradient = {}
+        for element in self.primal_model_part.Elements:
+            gradient[element.Id] = element.GetValue(variable)
+            #gradient[element.Id] = element.GetSolutionStepValue(variable)
+        return gradient
+
 # ==============================================================================
 class EigenFrequencyResponseFunction(StrainEnergyResponseFunction):
     """Eigenfrequency response function. The internal procedure is the same as
@@ -135,6 +149,8 @@ class EigenFrequencyResponseFunction(StrainEnergyResponseFunction):
 
         self.primal_analysis = StructuralMechanicsAnalysis(model, ProjectParametersPrimal)
         self.primal_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.SHAPE_SENSITIVITY)
+        self.primal_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.YOUNGS_MODULUS_SENSITIVITY)
+
 
         self.response_function_utility = StructuralMechanicsApplication.EigenfrequencyResponseFunctionUtility(self.primal_model_part, response_settings)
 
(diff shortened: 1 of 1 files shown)


applications/StructuralMechanicsApplication/python_scripts/structural_response.py
```diff
@@ -295,6 +323,21 @@ def GetNodalGradient(self, variable):
...
+                            StructuralMechanicsApplication.YOUNG_MODULUS_SENSITIVITY,
+                            StructuralMechanicsApplication.THICKNESS_SENSITIVITY,
+                            StructuralMechanicsApplication.I22_SENSITIVITY,
+                            StructuralMechanicsApplication.I33_SENSITIVITY,
+                            ]:
+            raise RuntimeError("GetElementalGradient: No gradient for {}!".format(variable.Name))
+        gradient = {}
+        for element in self.adjoint_model_part.Elements:
+            gradient[element.Id] = element.GetValue(variable)
+            #gradient[element.Id] = element.GetSolutionStepValue(variable)
```
armingeiser:
```suggestion
```

applications/StructuralMechanicsApplication/python_scripts/structural_response.py
```diff
@@ -295,6 +323,21 @@ def GetNodalGradient(self, variable):
             gradient[node.Id] = node.GetSolutionStepValue(variable)
         return gradient
 
+    def GetElementalGradient(self, variable):
```
armingeiser, line 305: I think this method should also be added to the `ResponseFunctionInterface` with a `NotImplementedError`

applications/StructuralMechanicsApplication/python_scripts/structural_response.py
```diff
@@ -135,6 +149,8 @@ def __init__(self, identifier, response_settings, model):
 
         self.primal_analysis = StructuralMechanicsAnalysis(model, ProjectParametersPrimal)
         self.primal_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.SHAPE_SENSITIVITY)
+        self.primal_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.YOUNGS_MODULUS_SENSITIVITY)
+
```
armingeiser: I think this EigenFrequencyResponseFunction,  StrainEnergyResponseFunction (and MassResponsefunction?) are only implemented for shape sensitivities at the moment.

Note: there is a difference in the implementation of these first response functions and the "AdjointResponseFunction" below, that has a more general implementation that also allows element properties to be used as variables.

applications/StructuralMechanicsApplication/python_scripts/structural_response.py
```diff
@@ -92,6 +92,20 @@ def GetNodalGradient(self, variable):
             gradient[node.Id] = node.GetSolutionStepValue(variable)
         return gradient
 
+    def GetElementalGradient(self, variable):
```
armingeiser, line 95: see comment below, i think this is not implemented.

applications/StructuralMechanicsApplication/python_scripts/structural_response.py
e-dub: I can confirm that these routines return dm/dE, dm/dt, dU/dE, dU/dt = 0. Removed and added error.

e-dub: well obviously a couple should be zero, but those that are obvioiusly neq 0 are zero.

armingeiser: Explanation: `EigenFrequencyResponseFunction, StrainEnergyResponseFunction, MassResponsefunction` have been the first implementations of response functions in Kratos, specifically with application in ShapeOptimization in mind. 
All other response functions with "adjoint" in there name, have been implemented by @MFusseder following a much more general structure that allows computing gradients w.r.t different variables than shape only.

applications/StructuralMechanicsApplication/python_scripts/structural_response.py
armingeiser: i just see its already there.

armingeiser (approved):

PR #9823, opened 2022-04-05
[LinearSolvers] Eigen inplace QR decompositions and optimization

rubenzorrilla:
**📝 Description**
This enhances our current interface to Eigen QR decompositions by:
- Doing the QR inplace (so far a copy of the input matrix was created)
- Removing most of the auxiliary arrays used to output de data
- Besides, I realized that, although the standard Householder QR is not rank revealing, we can always output the R. This has been added.
- Tests have been updated accordingly by keeping a copy of the original test matrix (note that now the original one is modified as the decomposition is done inplace) and by testing A = QR also in the QR without pivoting.

There shouldn't be anything affected by this change since the Eigen QR(s) were recently added and so far they're only used by the MVQN-RZ algorithm to calculate the randomized SVD, which is compatible with QR inplace decomposition.


applications/LinearSolversApplication/custom_decompositions/eigen_dense_householder_qr_decomposition.h
```diff
@@ -122,41 +126,65 @@ class EigenDenseHouseholderQRDecomposition : public DenseQRDecomposition<TDenseS
...
 
-        // Convert the matrix output data to vector type
-        IndexPartition<std::size_t>(n_rows).for_each([&](std::size_t i){rX(i) = aux_output(i,0);});
+        // Set the output matrix
+        std::size_t m = r_x.rows();
+        if (rX.size() != m) {
+            rX.resize(m,false);
+        }
+        Eigen::Map<EigenMatrix> eigen_x_map(rX.data().begin(), m, 1);
+        eigen_x_map = r_x;
```
RiccardoRossi: this is still a copy...

RiccardoRossi (dismissed):

applications/LinearSolversApplication/custom_decompositions/eigen_dense_householder_qr_decomposition.h
rubenzorrilla: Already fixed. See the outdated badge, it appears as it should in my "files changed" tab (I think GitHub got messed in here).

RiccardoRossi (approved):

PR #9825, opened 2022-04-05
[StructuralMechanics] Fix segfaults on empty model parts in ExplicitCentralDifferencesScheme

matekelemen:
**📝 Description**
```ExplicitCentralDifferencesScheme``` and ```MechanicalExplicitStrategy``` caused segfaults on model parts with no nodes. This PR adds a couple of checks to prevent that.

**🆕 Changelog**
None of the logic is changed, but lines which dereference the ```begin``` iterator of a node container are enclosed in an conditional branch that makes sure that the container is not empty. Some loops are skipped completely in these cases because they'd just loop through (an empty set of) nodes anyway.


loumalouomega (approved):
Good moment to parallel utils?

matekelemen:
@loumalouomega Should I do it in this PR?

loumalouomega:
> @loumalouomega Should I do it in this PR?

As you wish

PR #9828, opened 2022-04-06
[HDF5] Convert relevant processes to OutputProcesses

matekelemen:
**📝 Description**
HDF5 processes that write to files now inherit from ```OutputProcess``` and override ```OutputProcess::IsOutputStep```.

**🆕 Changelog**
Processes that use the core ```Factory``` must pass a tag that indicates what category (input or output) they belong in, which in turn determines whether they inherit from ```KratosMultiphysics.Process``` or ```KratosMultiphysics.OutputProcess```. It's not an elegant solution, but it doesn't require any code duplication and can keep the current design. Let me know if you have a better idea.


matekelemen:
Ditched the process tags; the process' type must be passed to the core ```Factory``` instead.
Added support for ```OutputProcess::PrintOutput``` that 'bypasses' checks by temporal controllers.

applications/HDF5Application/python_scripts/core/processes.py
```diff
@@ -6,95 +6,164 @@
...
-    def AddInitialize(self, func):
-        self.initialize_sequence.append(func)
+        def PrintOutput(self) -> None:
+            for f in self.output_sequence:
+                f()
 
-    def AddBeforeSolutionLoop(self, func):
-        self.before_solution_loop_sequence.append(func)
+        def IsOutputStep(self) -> bool:
+            """Always true, becuase this class aggregates controllers that decide whether to write anything separately."""
```
miguelmaso: is it possible to ask to the controller? It will be very useful

applications/HDF5Application/python_scripts/multiple_mesh_temporal_output_process.py
```diff
@@ -73,7 +73,7 @@ def Factory(settings, Model):
 
 
 def MultipleMeshTemporalOutputProcessFactory(core_settings, Model):
-    return core.Factory(core_settings, Model)
+    return core.Factory(core_settings, Model, KratosMultiphysics.OutputProcess)
```
miguelmaso, line 76: in the core settings, is it possible to do something like this?
```patch
-            "process_step": "finalize_solution_step",
+            "process_step": "print_output",
```
I don't know what consequences this may have for the existing code

miguelmaso (commented):
Thanks @matekelemen ! I have one comment. With this PR, the output processes are derived from `OuptutProcess` but behaving as a regular `Process`.

I don't know how painful would be changing some `process_step` to `output_step`. Maybe this issue can be addressed in a future PR

applications/HDF5Application/python_scripts/core/processes.py
miguelmaso: something like (I don't know if it works):
```py
for f in self.print_output_sequence:
    if f.controller._IsOutputStep()
        return True
return False
```

applications/HDF5Application/python_scripts/multiple_mesh_temporal_output_process.py
matekelemen: Yes it's possible and [implemented](https://github.com/KratosMultiphysics/Kratos/blob/9146edfbf36d87d85cc95ff1f7e3396f52a0ec5d/applications/HDF5Application/python_scripts/core/__init__.py#L60) (if you mean a process that performs an output on each call, without checks), but it's called ```"output"``` instead of ```"print_output"```.

applications/HDF5Application/python_scripts/core/processes.py
matekelemen: Good catch, I [corrected](https://github.com/KratosMultiphysics/Kratos/blob/9146edfbf36d87d85cc95ff1f7e3396f52a0ec5d/applications/HDF5Application/python_scripts/core/processes.py#L110) it. The only downside is that controller checks get executed twice on output steps, but that shouldn't be a big deal.

miguelmaso (dismissed):
wow, very nice!

miguelmaso (approved):

PR #9829, opened 2022-04-06
[Core] Apply table process

miguelmaso:
**📝 Description**
This process sets a variable from a table.
More or less, it works as AssignScalarVariableProcess, but the value is defined by a table.

**🆕 Changelog**
- Add ApplyTableToVariableProcess
- Add test


miguelmaso:
@AFranci does it solves your problem?

ipouplana (commented):
Very nice @miguelmaso !

AFranci:
> @AFranci does it solves your problem?

no pandas at the end??

miguelmaso:
> > @AFranci does it solves your problem?
> 
> no pandas at the end??

#9831 is even better

rubenzorrilla:
> > > @AFranci does it solves your problem?
> > 
> > 
> > no pandas at the end??
> 
> #9831 is even better

I was about to suggest this :+1:

PR #9831, opened 2022-04-07
[AssignVariableProcess] Read value from table

miguelmaso:
**📝 Description**
This PR extends `Assign[Scalar/Vector]VariableProcess`. There is a new option for `value`:

- `null`
- a constant scalar
- a function
- [New] a csv table

I think this implementation is better than #9829 

**🆕 Changelog**
- Add a read csv utility
- Extend the possibilities of `AssignScalarVariableProcess` and  `AssignVectorVariableProcess`
- Add a test


AFranci (dismissed):
Thanks @miguelmaso ! I have just tested in a PFEM example and works well!

AFranci:
@rubenzorrilla is it ok for you? if yes, can you approve it?

kratos/python_scripts/assign_scalar_variable_process.py
```diff
@@ -39,7 +50,7 @@ def __init__(self, Model, settings ):
             "variable_name"   : "SPECIFY_VARIABLE_NAME",
             "interval"        : [0.0, 1e30],
             "constrained"     : true,
-            "value"           : 0.0,
+            "value"           : {},
```
rubenzorrilla, line 51: Why changing the defaults? Note that this would break any case that is assuming the zero scalar default value (admittedly a weird usage).

rubenzorrilla: OK. Now I see below. Forget about my comment.

kratos/python_scripts/read_csv_table_utility.py
```diff
@@ -0,0 +1,79 @@
...
+    def __init__(self, settings):
+        """Constructor of the csv table reader: validate the parameters.
+
+        Keyword arguments:
+        self -- It signifies an instance of the class.
+        settings -- Kratos parameters containing solver settings.
+        """
+        default_settings =  KM.Parameters("""{
+            "name"       : "csv_table",
+            "file_name"  : "",
```
rubenzorrilla:
```suggestion
            "filename"  : "",
```
To be consistent with the rest of Kratos.

kratos/python_scripts/read_csv_table_utility.py
```diff
@@ -0,0 +1,79 @@
...
+        }""")
+        settings.ValidateAndAssignDefaults(default_settings)
+
+        self.file_name = settings["file_name"].GetString()
+        self.delimiter = settings["delimiter"].GetString()
+        self.skiprows = settings["skiprows"].GetInt()
+        self.table_id = settings["table_id"].GetInt()
+        self.na_replace = settings["na_replace"].GetDouble()
+
+    def Read(self, model_part):
```
rubenzorrilla:
```suggestion
    def Read(self, model_part = None):
```
So you don't need to pass the model part if the table is not to be stored.

kratos/python_scripts/assign_scalar_variable_process.py
miguelmaso: I think you're right, if a user does not set `"value" : whatever`, the default value would be `{}`, producing a runtime error.

Now I remember: I modified the default values because the method `void Parameters::SetValue(Parameters Value)` is missing. I can add an extra check in python to set `"value" : 0.0` if it's missing.

kratos/python_scripts/read_csv_table_utility.py
```diff
@@ -0,0 +1,79 @@
...
+                next(data)
+            for row in data:
+                if row:  # skip empty rows
+                    if len(row) < 2:
+                        raise Exception("Only 2-column tables are supported. However, a {}-column row is found.".format(len(row)))
+                    elif len(row) > 2:
+                        KM.Logger.PrintWarning("Only 2-column tables are supported. However, a {}-column row is found. Extra columns will be ignored.".format(len(row)))
+                    table.AddRow(self._Float(row[0]), self._Float(row[1]))
+        if self.table_id > -1:
+            model_part.AddTable(self.table_id, table)
```
rubenzorrilla:
```suggestion
        if self.table_id > -1:
            if model_part:
                model_part.AddTable(self.table_id, table)
            else:
                err_msg = "Asking to save table with id {} but no model part is provided.".format(self.table_id)
                raise Exception(err_msg)
```

kratos/python_scripts/read_csv_table_utility.py
```diff
@@ -0,0 +1,79 @@
...
+            data = csv.reader(table_file, delimiter=self.delimiter, skipinitialspace=True)
+            for _ in range(self.skiprows):
+                next(data)
+            for row in data:
+                if row:  # skip empty rows
+                    if len(row) < 2:
+                        raise Exception("Only 2-column tables are supported. However, a {}-column row is found.".format(len(row)))
+                    elif len(row) > 2:
+                        KM.Logger.PrintWarning("Only 2-column tables are supported. However, a {}-column row is found. Extra columns will be ignored.".format(len(row)))
+                    table.AddRow(self._Float(row[0]), self._Float(row[1]))
```
rubenzorrilla: Why not making optional which columns are considered and defaulting to 0 and 1? Consider the case of a multicolumn large csv, editing it to get your two columns of interest and copy them to a new file might be a bit annoying...

kratos/tests/auxiliar_files_for_python_unittest/test_processes/table_input.csv
```diff
@@ -0,0 +1,12 @@
...
+0,  0
+1,   0.1
+2,  0.3
+3,  0.6
+4,   8e-1
+5,   0.9
+9,  1
+10, N/A
+11, na
+12, inf
```
rubenzorrilla, line 11: I guess these three entries will be substituted by the `na_replace` value. Am I wrong?

kratos/python_scripts/read_csv_table_utility.py
```diff
@@ -0,0 +1,79 @@
...
+        if self.table_id > -1:
+            model_part.AddTable(self.table_id, table)
+        return table
+
+    def _Float(self, value):
+        try:
+            return float(value)
+        except ValueError:
+            return self.na_replace
```
rubenzorrilla, line 95: I think we should throw a warning in here. Consider again the case of a very large file.

kratos/tests/auxiliar_files_for_python_unittest/test_processes/table_input.csv
miguelmaso: exactly

kratos/python_scripts/assign_scalar_variable_process.py
rubenzorrilla: This would be the safest. In any case calling this process without providing `value` is a misuse in my opinion. Having said this we'd better throw an error if this is not provided.

kratos/python_scripts/read_csv_table_utility.py
miguelmaso: are you sure? It could throw a lot of warnings if there is missing data, e.g.: everything NA after a given time

kratos/python_scripts/read_csv_table_utility.py
miguelmaso: do you mean this?
```json
            "name"            : "csv_table",
            "filename"        : "",
            "delimiter"       : ",",
            "skiprows"        : 0,
            "time_column_id"  : 0,  <- new
            "value_column_id" : 1,  <-new
            "table_id"        : -1,
            "na_replace"      : 0.0
```

miguelmaso: done

kratos/python_scripts/read_csv_table_utility.py
rubenzorrilla: I'd prefer being too verbose rather than spending time figuring out why the solution is not the expected one. Imagine the case of a large dataset which, because any reason, has one entry that is n/a (e.g. any problem with the origin data). Besides, I'd expect that, in the majority of the cases, there are not so much N/A.

kratos/python_scripts/read_csv_table_utility.py
rubenzorrilla: Precisely. As a comment,  I'd call first/second (or left/right) column rather than time as this utility can be in principle use for any table-related purpose (e.g. temperature - Young modulus)

kratos/python_scripts/read_csv_table_utility.py
miguelmaso: ok

kratos/python_scripts/read_csv_table_utility.py
miguelmaso: done, first/second

kratos/python_scripts/read_csv_table_utility.py
miguelmaso: done

rubenzorrilla: Very good. Eventually we can remove it in the future if it becomes way too verbose.

rubenzorrilla (dismissed):

rubenzorrilla (approved):

PR #9833, opened 2022-04-07
Core/add fluid rve

bcsnvs:
Add fluid RVE analysis script, test and supplementary files


applications/FluidDynamicsApplication/python_scripts/fluid_rve_analysis.py
```diff
@@ -0,0 +1,203 @@
+import KratosMultiphysics
+import KratosMultiphysics.FluidDynamicsApplication
```
rubenzorrilla:
```suggestion
```
I think you can remove this import.

applications/FluidDynamicsApplication/python_scripts/fluid_rve_analysis.py
```diff
@@ -0,0 +1,203 @@
...
+from KratosMultiphysics.FluidDynamicsApplication.fluid_dynamics_analysis import FluidDynamicsAnalysis
+
+# Importing other libraries
+import math
+
+class FluidRVEAnalysis(FluidDynamicsAnalysis):
```
rubenzorrilla:
```suggestion
class FluidDynamicsAnalysisRVE(FluidDynamicsAnalysis):
```
I'd rename it like this to follow the naming convention of the other analysis stage of the app (`FluidDynamicsAnalysis`).

I also suggest renaming the file to `fluid_dynamics_analysis_rve.py` so they appear together when browsing the folder. Note that to do so without loosing the history of the file you need to open a terminal, navigate to the current folder and do `git mv fluid_rve_analysis.py fluid_dynamics_analysis_rve.py`.

(this is only a minor suggestion, up to you)

applications/FluidDynamicsApplication/python_scripts/fluid_rve_analysis.py
```diff
@@ -0,0 +1,203 @@
...
+    def __init__(self, model, project_parameters):
+
+        # input parameters of the analysis
+        self.boundary_mp_name = project_parameters["rve_settings"]["boundary_mp_name"].GetString()
+        self.averaging_mp_name = project_parameters["rve_settings"]["averaging_mp_name"].GetString()
+        self.print_rve_post = project_parameters["rve_settings"]["print_rve_post"].GetBool()
```
rubenzorrilla:
```suggestion
```
This variable is not used in the current file (see my suggestions about the output below).

applications/FluidDynamicsApplication/python_scripts/fluid_rve_analysis.py
```diff
@@ -0,0 +1,203 @@
...
+
+class FluidRVEAnalysis(FluidDynamicsAnalysis):
+    def __init__(self, model, project_parameters):
+
+        # input parameters of the analysis
+        self.boundary_mp_name = project_parameters["rve_settings"]["boundary_mp_name"].GetString()
+        self.averaging_mp_name = project_parameters["rve_settings"]["averaging_mp_name"].GetString()
+        self.print_rve_post = project_parameters["rve_settings"]["print_rve_post"].GetBool()   
+
+        self.averaging_volume = -1.0  # it will be computed in initialize
```
rubenzorrilla:
```suggestion
```
See my comment below.

applications/FluidDynamicsApplication/python_scripts/fluid_rve_analysis.py
```diff
@@ -0,0 +1,203 @@
...
+        self.averaging_mp = self.model[self.averaging_mp_name]
+
+        # Construct auxiliary modelparts
+        self.min_corner, self.max_corner = self._DetectBoundingBox(self.averaging_mp)
+        self._ConstructFaceModelParts(self.min_corner, self.max_corner, boundary_mp)
+
+        self.averaging_volume = 1.0
+        for i in range(self.domain_size):
+            self.averaging_volume *= self.max_corner[i]-self.min_corner[i]
+        KratosMultiphysics.Logger.PrintInfo(self._GetSimulationName(), "RVE undeformed averaging volume = ", self.averaging_volume)
```
rubenzorrilla:
```suggestion
```
Our periodicity conditions strictly imply no volume change so I consider this a leftover from the structural mechanics implementation that has not so much sense in our case.

applications/FluidDynamicsApplication/python_scripts/fluid_rve_analysis.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def ModifyAfterSolverInitialize(self):
```
rubenzorrilla:
```suggestion
    def ModifyAfterSolverInitialize(self):
        super().ModifyAfterSolverInitialize()
                        
```
Although it will always be potentially empty, I'd call the base class  `ModifyAfterSolverInitialize` (just in case we change the base class in the future).

applications/FluidDynamicsApplication/tests/FluidRVETest/fluid_rve_test_materials_2D.json
```diff
@@ -0,0 +1,16 @@
+{
```
rubenzorrilla, line 1: I think that we can have a single json file and edit the constitutive law name according to the problem DOMAIN_SIZE as you do with the other settings.

applications/FluidDynamicsApplication/tests/FluidRVETest/fluid_rve_test_parameters.json
```diff
@@ -0,0 +1,68 @@
+{
+    "problem_data": {
+        "problem_name": "fluid_rve_test",
+        "parallel_type": "OpenMP",
+        "echo_level": 0,
+        "start_time": 0.0,
+        "end_time": 1.0
```
rubenzorrilla:
```suggestion
        "end_time": 0.1
```
To reduce the test runtime.

applications/FluidDynamicsApplication/tests/FluidRVETest/fluid_rve_test_parameters.json
```diff
@@ -0,0 +1,68 @@
...
+    },
+    "rve_settings": {
+        "boundary_mp_name": "TO_BE_SET_IN_TEST",
+        "averaging_mp_name": "FluidModelPart",
+        "print_rve_post": true,
```
rubenzorrilla:
```suggestion
```
If required, the output should be added as a process. See my comment regarding the test output below.

applications/FluidDynamicsApplication/python_scripts/fluid_rve_analysis.py
```diff
@@ -0,0 +1,203 @@
+import KratosMultiphysics
+import KratosMultiphysics.FluidDynamicsApplication
+from KratosMultiphysics.FluidDynamicsApplication.fluid_dynamics_analysis import FluidDynamicsAnalysis
+
+# Importing other libraries
+import math
+
+class FluidRVEAnalysis(FluidDynamicsAnalysis):
+    def __init__(self, model, project_parameters):
+
```
rubenzorrilla: We should add a validation of the user `rve_settings` in here. See how we do so in the solvers. Long story short the idea would be to hardcode a `KratosMultiphysics.Parameters` object containing the expected settings with default values and then call the input `rve_settings` against these by doing
```py
input_rve_settings = project_parameters["rve_settings"] # take the rve_settings from the input
default_settings = self.GetDefaultParameters() # this method returns the rve_settings we expect
input_rve_settings.ValidateAndAssignDefaults(default_settings) # this method validates the input ones with it
```

applications/FluidDynamicsApplication/tests/FluidRVETest/fluid_rve_test_parameters.json
```diff
@@ -0,0 +1,68 @@
...
+            "automatic_time_step": false,
+            "time_step": 0.1
+        },
+        "formulation": {
+            "element_type": "qsvms",
+            "use_orthogonal_subscales": false,
+            "dynamic_tau": 1.0
+        },
+        "reform_dofs_at_each_step": false,
+        "analysis_type": "linear"
```
rubenzorrilla:
```suggestion
        "reform_dofs_at_each_step": false
```
The leftover from the structural mechanics we fixed the other day.

applications/FluidDynamicsApplication/tests/test_fluid_rve.py
```diff
@@ -0,0 +1,96 @@
...
+            with open("FluidRVETest/fluid_rve_test_parameters.json", 'r') as parameter_file:
+                parameters =  KratosMultiphysics.Parameters(parameter_file.read())
+                
+            parameters["solver_settings"]["domain_size"].SetInt(2)
+            parameters["solver_settings"]["model_import_settings"]["input_filename"].SetString("FluidRVETest/fluid_rve_test_2D")
+            parameters["solver_settings"]["material_import_settings"]["materials_filename"].SetString("FluidRVETest/fluid_rve_test_materials_2D.json")
+            parameters["rve_settings"]["boundary_mp_name"].SetString("FluidModelPart.Slip2D.Boundaries")
+            parameters["solver_settings"]["skin_parts"][0].SetString("Slip2D")
+            parameters["processes"]["initial_conditions_process_list"][0]["Parameters"]["value"][0].SetInt(0)
+            parameters["processes"]["initial_conditions_process_list"][0]["Parameters"]["value"][1].SetInt(0)
```
rubenzorrilla: Are we deliberately initializing the `VELOCITY` field to zero? This means doing nothing.

applications/FluidDynamicsApplication/tests/test_fluid_rve.py
```diff
@@ -0,0 +1,96 @@
...
+                parameters =  KratosMultiphysics.Parameters(parameter_file.read())
+                
+            parameters["solver_settings"]["domain_size"].SetInt(3)
+            parameters["solver_settings"]["model_import_settings"]["input_filename"].SetString("FluidRVETest/fluid_rve_test_3D")
+            parameters["solver_settings"]["material_import_settings"]["materials_filename"].SetString("FluidRVETest/fluid_rve_test_materials_3D.json")
+            parameters["rve_settings"]["boundary_mp_name"].SetString("FluidModelPart.Slip3D.Boundaries")
+            parameters["solver_settings"]["skin_parts"][0].SetString("Slip3D")
+            parameters["processes"]["initial_conditions_process_list"][0]["Parameters"]["value"][0].SetInt(0)
+            parameters["processes"]["initial_conditions_process_list"][0]["Parameters"]["value"][1].SetInt(0)
+            parameters["processes"]["initial_conditions_process_list"][0]["Parameters"]["value"][2].SetInt(0)
```
rubenzorrilla: Same comment.

applications/FluidDynamicsApplication/tests/test_fluid_rve.py
```diff
@@ -0,0 +1,96 @@
...
+            
+            self._aux_fluid_rve_computation(parameters)
+        
+    def _aux_fluid_rve_computation(self, parameters):
+        
+        domain_size = parameters["solver_settings"]["domain_size"].GetInt()
+        if parameters["rve_settings"]["print_rve_post"].GetBool() :
```
rubenzorrilla:
```suggestion
        if self.print_output:
```

applications/FluidDynamicsApplication/tests/test_fluid_rve.py
```diff
@@ -0,0 +1,96 @@
...
+import KratosMultiphysics.KratosUnittest as KratosUnittest
+
+import KratosMultiphysics.kratos_utilities as kratos_utilities # Might be needed for testing
+
+from KratosMultiphysics.FluidDynamicsApplication.fluid_rve_analysis import FluidRVEAnalysis
+
+# Previous line will import the Fluid RVE analysis class
+
+class TestFluidRVETest(KratosUnittest.TestCase):
+
```
rubenzorrilla:
```suggestion
    
    def setUp(self):
        self.print_output = False

```
By setting this to `True` we can print the test output for debugging. Then we remove the output value from the `rve_settings` so in a "standard" (out of the unit testing context) case we add the output as we do normally in the `output_process_list`.

rubenzorrilla (commented):
Congratulations. Very good contribution overall. However, there are some minor things that need to be amended before merging (see my review comments).

Besides, the tests are taking too much time (6 seconds in my desktop). I suspect this is because of the 3D test. To reduce this, I suggest:
- Making the mesh coarser, specially the 3D one
- Running a single time step (enough for the testing). As far as I see, right now we're running 10.

As a future TODO (I can take care of it once this is merged), we should create a sort of RVE utilities to unify things. There are some copy and paste from the structure RVE in here that can be unified in an independent class (mainly the operations related to the creation of the skin model part).

applications/FluidDynamicsApplication/python_scripts/fluid_rve_analysis.py
bcsnvs: Removed.

applications/FluidDynamicsApplication/python_scripts/fluid_rve_analysis.py
bcsnvs: Ok, both changes done.

applications/FluidDynamicsApplication/python_scripts/fluid_rve_analysis.py
bcsnvs: True, removed.

applications/FluidDynamicsApplication/python_scripts/fluid_rve_analysis.py
bcsnvs: Ok, added.

applications/FluidDynamicsApplication/tests/FluidRVETest/fluid_rve_test_materials_2D.json
bcsnvs: Done as well.

applications/FluidDynamicsApplication/tests/FluidRVETest/fluid_rve_test_parameters.json
bcsnvs: Removed!

applications/FluidDynamicsApplication/tests/test_fluid_rve.py
bcsnvs: Both this and  below were done like this because I did not know if some sort of field was needed. At some point it had a SetString() method with the corresponding "var * G" fields.

Should I delete it regardless?

applications/FluidDynamicsApplication/tests/FluidRVETest/fluid_rve_test_parameters.json
bcsnvs: Changed.

applications/FluidDynamicsApplication/tests/test_fluid_rve.py
```diff
@@ -0,0 +1,96 @@
...
+                }
+            }]''')
+            output_settings[0]["Parameters"]["output_name"].SetString("FluidRVETest/fluid_rve_test_{}D".format(domain_size))
+            parameters["output_processes"].AddValue("gid_output", output_settings)
+        
+        model = KratosMultiphysics.Model()
+        simulation = FluidRVEAnalysis(model, parameters)
+        simulation.Run()
+        
+        # Space for testing
```
rubenzorrilla: Reminder to do this.

applications/FluidDynamicsApplication/python_scripts/fluid_rve_analysis.py
bcsnvs: Removed!

applications/FluidDynamicsApplication/tests/FluidRVETest/fluid_rve_test_materials_2D.json
bcsnvs: In the end it was left as it was given that, when opening the materials json as a Kratos parameters object what's being edited (constitutive_law, name field) is the object, not the json.

applications/FluidDynamicsApplication/python_scripts/fluid_rve_analysis.py
bcsnvs: Fixed this way:
class FluidDynamicsAnalysisRVE(FluidDynamicsAnalysis):

    @classmethod
    def GetDefaultParameters(cls):

        ##settings string in json format
        default_settings = KratosMultiphysics.Parameters("""
        {
            "boundary_mp_name": "unknown_name",
            "averaging_mp_name": "FluidModelPart",
            "print_rve_post": false,
            "fix_pressure": true,
            "jump_XX": 0,
            "jump_XY": 1e-6,
            "jump_YY": 0,
            "jump_XZ": 1e-6,
            "jump_ZZ": 0,
            "jump_YZ": 1e-6
        }""")

        return default_settings
    
    def __init__(self, model, project_parameters):

        # Check validity of RVE settings
        input_rve_settings = project_parameters["rve_settings"]
        default_rve_settings = self.GetDefaultParameters()
        input_rve_settings.ValidateAndAssignDefaults(default_rve_settings)


The `print_rve_post` field will be deleted when I fix the output settings

applications/FluidDynamicsApplication/tests/test_fluid_rve.py
rubenzorrilla: OK. If we are to set an initial field  in order to trigger the problem we should set in here. If not we can remove it together with the corresponding entry in the `fluid_rve_test_parameters.json`. The point is that, by default, variables in the nodal historical database (such as `VELOCITY`) are initialized to zero by default, so if we want to start from a zero velocity field we just need to do nothing.

applications/FluidDynamicsApplication/tests/test_fluid_rve.py
rubenzorrilla: (see my comment above)

applications/FluidDynamicsApplication/tests/test_fluid_rve.py
rubenzorrilla: We still need to do this.

applications/FluidDynamicsApplication/python_scripts/fluid_rve_analysis.py
rubenzorrilla: OK. This was precisely what I had in mind. My unique comment would be to rename the `GetDefaultParameters` to `GetFluidRVEDefaultParameters` to make clear that these are the defaults of this portion of the input (not of the entire analysis stage).

applications/FluidDynamicsApplication/tests/test_fluid_rve.py
```diff
@@ -1,17 +1,18 @@
...
 
 # Previous line will import the Fluid RVE analysis class
 
 class TestFluidRVETest(KratosUnittest.TestCase):
     
     def setUp(self):
+        # Change to True to generate GiD Output 
         self.print_output = False
+        # Change to True if no refence value file exists
+        self.check_tolerance = 1e-4
```
rubenzorrilla: I'd use a more restrictive tolerance. Lets say 1.0e-8.

applications/FluidDynamicsApplication/tests/test_fluid_rve.py
```diff
@@ -82,14 +87,53 @@ def _aux_fluid_rve_computation(self, parameters):
...
+                    velocity = node.GetSolutionStepValue(KratosMultiphysics.VELOCITY)
+                    self.assertAlmostEqual(reference_vel_x, velocity[0], delta = self.check_tolerance)
+                    self.assertAlmostEqual(reference_vel_y, velocity[1], delta = self.check_tolerance)
+                    if self.domain_size == 3:
+                        self.assertAlmostEqual(reference_vel_z, velocity[2], delta = self.check_tolerance)
+                   
+
+                    line = reference_file.readline()
+                if line != '': 
+                    self.fail("The number of nodes in the mdpa is smaller than the number of nodes in the output file")
```
rubenzorrilla: Though it works, this is how we used to check results in the past when we didn't have the mechanism to properly do so. To avoid reimplementing this in each test, we created some processes. See how these can be used in here

https://github.com/KratosMultiphysics/Kratos/blob/9e351bc965ef0a8347ccd495eadadad1ce74053e/applications/FluidDynamicsApplication/tests/embedded_couette_flow_test.py#L253-L295

I think this will make things easier.

applications/FluidDynamicsApplication/tests/test_fluid_rve.py
rubenzorrilla: Done.

applications/FluidDynamicsApplication/tests/test_fluid_rve.py
rubenzorrilla: Done.

applications/FluidDynamicsApplication/tests/test_fluid_rve.py
rubenzorrilla: Done.

applications/FluidDynamicsApplication/python_scripts/fluid_rve_analysis.py
rubenzorrilla: Done.

rubenzorrilla:
@playpaolo @bcsnvs I think this is now ready to go in.

playpaolo (approved):
RVE ready

Issue #9602, opened 2022-02-04
Stress constriants and adjoint sensitivities for Topology Optimization Application


PhiHo-eng:
Hi,

I would like to consider stress in my topology optimization. This sets up a two part question:

First to the calculation:
I am using volume elements small_displacement_simp (based on small_displacement).
VON_MISES_STRESS should be calculated here in line 140 [small_dispalcement_simp_element.cpp](https://github.com/KratosMultiphysics/Kratos/pull/9480/files)
I am not seeming my error, but right now nothing is happening.
Do I need to implemente it? Or can I inherit the according functions from StracturalMechanicsApplication?

The second part is for the sensitivities:
Before I go to far here, I wanted to check what the capabilities in Kratos are now.
I see a couple tests with adjoint sensitivities for shells, beams and bar elements and in the response factory there is adjoint_max_stress and adjoint_local_stress. Should this work with volumes too?

Thank you in advance!

PhiHo-eng:
The first part of the question should be solved now, as I implemented the calculation of the VON_MISES_STRESS equivalent as it done in the Structural Mechanics Application.

loumalouomega:
I will mention @KratosMultiphysics/structural-mechanics

armingeiser:
@MFusseder @sunethwarna FYI.

e-dub:
We found what we were looking for (in part) in the [tests](https://github.com/KratosMultiphysics/Kratos/tree/master/applications/StructuralMechanicsApplication/tests/response_function_tests) and specifically [adjoint_max_stress](https://github.com/KratosMultiphysics/Kratos/blob/master/applications/StructuralMechanicsApplication/tests/response_function_tests/adjoint_max_stress_primal_parameters.json). Impressive work!  

To make things easier for us, I extended [structural_response.py](https://github.com/KratosMultiphysics/Kratos/blob/master/applications/StructuralMechanicsApplication/python_scripts/structural_response.py) to get to the elemental gradients like the nodal gradients with pull request #9819.

Through the grapevine, we have heard that work is being carried out on function aggregation and sensitivities for shape optimization.  May I inquire about the plans (and maybe the state) of this development?  Maybe we can take a look at this together to avoid duplicate activities.

PhiHo-eng:
Hi everyone, 
[...]

PR #5682, opened 2019-09-27
[ShapeOpt] [WIP] Adding new response with stress aggregation for optimization

dbaumgaertner:
This PR adds a new response function which allows for stress aggregation using the Kreisselmeier Steinhauser (KS) function.

By that, stress constraints may be considered over big parts of a structural domain whereas the given functionalities in Kratos only allowed for a single maximum value (whose location may heavily jump around). See figure below for an impression. There, a loaded hook structure is optimized for mass, while the stress in each element of the structure is not allowed to exceed a maximum value (dark red). The color plot shows the stresses on the surface for a constant value range. It can be seen, that the stresses significantly increase when reducing the mass (removing material). But they always stay within the specified limits.

ezgif com-crop (2)

This branch is **not yet ready to be merged** as it revealed an existing problem with thread safety in the current adjoint solution approach in case several elements are considered within the given stress response functions.  

Also, this PR first requires an approval and merge of #4364.

Roadmap:

- [ ] Identify and fix the bug that destroys thread safety (remove race conditions)
- [ ] Remove temporary workaround with a manual set of the number of threads to 1 
- [ ] Create a unit test for the stress aggregation response in the StructuralMechanicsApp
- [ ] Clean code and prepare for PR
- [x] Wait for the merge of #4364

PR #9835, opened 2022-04-08
StructMech: Adding max stress approximated via Kreisselmeier-Steinhauser function aggregation

e-dub:
**📝 Description**
This code change adds the max stress approximation via modified Kreisselmeier-Steinhauser (KS) function aggregation [1, 2].  Both the primal analysis and adjoint sensitivity analysis are included. KS is a continuous approximation of the inherently noncontinuous maximum operator, i.e. the maximum stress can "jump" between elements causing convergence issues when optimizing.  The modified version [2] avoids numerical overflow problems of the original function [1].  This is intended to be used in the topology optimization application (@PhiHo-eng) and should close issue #9602. 

[1] Kreisselmeier, G., Steinhauser, R.: Systematic control design by optimizing a vector performance vector. In: International Federation of Active Controls Symposium on Computer-Aided Design of Control Systems. (1979), DOI: 10.1016/S1474-6670(17)65584-8

[2] Martins, J.R.R.A., Poon, N.M.K.: On structural optimization using constraint aggregation. In: 6th World Congress on Structural and Multidisciplinary Optimization. (2005).

**🆕 Changelog**
- Added KS

### DIFF
--- a/applications/StructuralMechanicsApplication/custom_response_functions/response_utilities/adjoint_KS_max_stress_response_function.cpp
+++ b/applications/StructuralMechanicsApplication/custom_response_functions/response_utilities/adjoint_KS_max_stress_response_function.cpp
@@ -9,7 +9,6 @@
 //  Main authors:    E. Wehrle
 //                   based on max stress from Baumgaertner Daniel, https://github.com/dbaumgaertner
 //
-//
 
 // System includes
 
@@ -21,7 +20,6 @@
 
 namespace Kratos
 {
-    // --------------------------------------------------------------------------
     AdjointKSMaxStressResponseFunction::AdjointKSMaxStressResponseFunction(ModelPart& rAdjointModelPart, Parameters ResponseSettings)
     : AdjointStructuralResponseFunction(rAdjointModelPart, ResponseSettings),
       mrAdjointModelPart(rAdjointModelPart),
@@ -37,10 +35,9 @@ namespace Kratos
             << "AdjointKSMaxStressResponseFunction::AdjointKSMaxStressResponseFunction: Specified stress treatment not supported: " << ResponseSettings["stress_type"].GetString() << std::endl;
     }
 
-    // --------------------------------------------------------------------------
+    // primary analysis for max stress approximated via modified Kreisselmeier-Steinhauser aggregation
     AdjointKSMaxStressResponseFunction::~AdjointKSMaxStressResponseFunction(){}
 
-    // --------------------------------------------------------------------------
     double AdjointKSMaxStressResponseFunction::CalculateValue(ModelPart& rPrimalModelPart)
     {
         KRATOS_TRY;
@@ -48,10 +45,7 @@ namespace Kratos
         ModelPart& primal_agglomeration_part = rPrimalModelPart.GetSubModelPart(mCriticalPartName);
 
         // Find max mean value and corresponding element in primal model part
-        double max_mean_stress = 0.0;
-        double KS_max_mean_stress = 0.0;
-        double KS_sum = 0.0;
-        double pKS = 50.0;  //this should possibly be a variable to set
+        double pKS = 100.0;  //this should possibly be a variable to set
 
         IndexType elem_id_at_max = 0;
 
@@ -66,30 +60,21 @@ namespace Kratos
             for(IndexType i = 0; i < stress_vec_size; ++i)
                 mean_stress += element_stress[i];
             mean_stress /= stress_vec_size;
-
             if(mean_stress > max_mean_stress)
             {
                 max_mean_stress = mean_stress;
                 elem_id_at_max = elem.Id();
             }
+
+            mean_stress_vector[elem.Id()] = mean_stress;
         }
 
-        // I am looking twice to avoid save all the stresses, waste of time?
+        // reloop over elements, faster as vector operation?
         for(auto& elem : primal_agglomeration_part.Elements())
         {
-            Vector element_stress;
-            const ProcessInfo &r_process_info = rPrimalModelPart.GetProcessInfo();
-            StressCalculation::CalculateStressOnGP(elem, mTracedStressType, element_stress, r_process_info);
-
-            const SizeType stress_vec_size = element_stress.size();
-            double mean_stress = 0.0;
-            for(IndexType i = 0; i < stress_vec_size; ++i)
-                mean_stress += element_stress[i];
-            mean_stress /= stress_vec_size;
-            KS_sum += std::exp(pKS*(mean_stress-max_mean_stress)
+            KS_sum += std::exp(pKS*(mean_stress_vector[elem.Id()]-max_mean_stress));
         }
-        KS_max_mean_stress += max_mean_stress+1/pKS*std::log(KS_sum)
-
+        double KS_max_mean_stress = max_mean_stress+1/pKS*std::log(KS_sum);
         KRATOS_INFO_IF("AdjointKSMaxStressResponseFunction::CalculateValue", mEchoLevel > 0) << "Id of element with max stress value = " << elem_id_at_max << std::endl;
         KRATOS_INFO_IF("AdjointKSMaxStressResponseFunction::CalculateValue", mEchoLevel > 0) << "KS approx max mean stress value = " << KS_max_mean_stress << std::endl;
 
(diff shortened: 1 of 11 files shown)


applications/StructuralMechanicsApplication/custom_response_functions/response_utilities/adjoint_KS_max_stress_response_function.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+    double AdjointKSMaxStressResponseFunction::CalculateValue(ModelPart& rPrimalModelPart)
+    {
+        KRATOS_TRY;
+
+        ModelPart& primal_agglomeration_part = rPrimalModelPart.GetSubModelPart(mCriticalPartName);
+
+        IndexType elem_id_at_max = 0;
+
+        for(auto& elem : primal_agglomeration_part.Elements())
```
loumalouomega, line 62: I would do a paralel loop

applications/StructuralMechanicsApplication/custom_response_functions/response_utilities/adjoint_KS_max_stress_response_function.cpp
```diff
@@ -0,0 +1,251 @@
...
+            if(mean_stress > max_mean_stress)
+            {
+                max_mean_stress = mean_stress;
+                elem_id_at_max = elem.Id();
+            }
+            mean_stress_vector[elem.Id()] = mean_stress;
+        }
+
+        // reloop over elements, faster as vector operation?
+        for(auto& elem : primal_agglomeration_part.Elements())
```
loumalouomega, line 87: Idem

applications/StructuralMechanicsApplication/custom_response_functions/response_utilities/adjoint_KS_max_stress_response_function.h
```diff
@@ -0,0 +1,230 @@
...
+    ///@name Operators
+    ///@{
+
+    ///@}
+    ///@name Operations
+    ///@{
+
+    using AdjointStructuralResponseFunction::CalculateGradient;
+
+    void CalculateGradient(const Element& rAdjointElement,
```
loumalouomega, line 84: Doxygen doc (whole file)

loumalouomega (changes requested):
Suggestions

applications/StructuralMechanicsApplication/custom_response_functions/response_utilities/adjoint_KS_max_stress_response_function.h
```diff
@@ -0,0 +1,230 @@
...
+    ModelPart& mrAdjointModelPart;
+    std::string mCriticalPartName;
+    Element::Pointer mpTracedElementInAdjointPart = nullptr;
+    StressTreatment mStressTreatment;
+    TracedStressType mTracedStressType;
+    SizeType mEchoLevel = 0;
+    double max_mean_stress = 0.0;
+    double KS_exp_sum = 0.0;
+    std::map<int,double> mean_stress_vector;
+    double pKS = 100.0;  //this should be a variable to set
```
armingeiser: reminder to add this to the ResponseSettings that are passed to the constructor

applications/StructuralMechanicsApplication/custom_response_functions/response_utilities/adjoint_KS_max_stress_response_function.cpp
```diff
@@ -0,0 +1,251 @@
...
+                                                               const ProcessInfo& rProcessInfo)
+    {
+        KRATOS_TRY;
+
+        KRATOS_ERROR_IF(mpTracedElementInAdjointPart == nullptr)
+            << "AdjointKSMaxStressResponseFunction::CalculateGradient: Traced element not initialized. First call \"CalculateValue\"!" << std::endl;
+
+        Matrix stress_displacement_derivative;
+
+        mpTracedElementInAdjointPart->Calculate(STRESS_DISP_DERIV_ON_GP,
```
armingeiser: I think the gradient should be computed for all elements that take part in the KS aggregation function? Here only the element with the actual max. stress is considered, but several times.

armingeiser (changes requested):
There was an attempt in implementing this some time ago that has unfortunately never been merged: https://github.com/KratosMultiphysics/Kratos/pull/5682

Some things are handled a bit differently (subtracting max_mean_stress from element stress value here vs. scaling stress in the other MR)

Main difference is which elements contribute to gradient, partial sensitivity etc. I think it is correct in the old MR, since here as far as i understand only the maximum element contributes to the gradient.

@sunethwarna please have a look, we have just recently talked about this.

applications/StructuralMechanicsApplication/custom_response_functions/response_utilities/adjoint_KS_max_stress_response_function.cpp
armingeiser: https://github.com/KratosMultiphysics/Kratos/pull/5682/files#diff-8b6054994596597f5c757bec84c589176b473f5ac682fbbb72a91cacecbb2cb2R126

e-dub: Yes this was an issue throughout. Should be resolved with "inspiration" from previous MR

applications/StructuralMechanicsApplication/custom_response_functions/response_utilities/adjoint_KS_max_stress_response_function.cpp
e-dub: I would postpone this and do this all at once with the other responses here.

applications/StructuralMechanicsApplication/custom_response_functions/response_utilities/adjoint_KS_max_stress_response_function.h
e-dub: Sure. Can you provide a good example in the Kratos project...for inspiration?

armingeiser (dismissed):
my concerns have been adressed

dbaumgaertner:
Just wanted to say, it is really cool to see the KS function finally come in. Good job and thanks @e-dub! As @armingeiser already said, the old MR unfortunately never made it to the main repo due to a persisting bug that destroys thread saftey and causes a race condition, see #5682. Talking about it, @e-dub, @armingeiser do you see the same problem here, or do you think it is not a problem anymore? Since I started the implementation in #5682, a lot of things have changed so I cant really judge this.

e-dub:
@dbaumgaertner May I inquire what the consequence of this error? Fluctuating results as you mentioned in #5682 or also nan?

dbaumgaertner:
Can't remember exactly. But I think there were just fluctuating values. So, the values of the derivatives changed with each new run. But we did not dig deeper back then. All the findings are mentioned in #5682 Most probably it is not because of the local implementation of the stress aggregation but has something to do with the logic of how the adjoint state and subsequently the sensitivities are computed and combined. The thing is, it may cause silent errors if you not do something about it. So, I would be very cautious and verify the relevant sensitivities carefully.

armingeiser:
From what i remember, `os.environ["OMP_NUM_THREADS"] = "1"` was sufficient to prevent the race condition from appearing in the old MR #5682.

Since then, [parallel utilities](https://github.com/KratosMultiphysics/Kratos/wiki/Parallel-Utilities) have been used to replace quite some parallel loops that have previously been implemented with OpenMP, so maybe `os.environ["OMP_NUM_THREADS"] = "1"` is not the right way to switch parallelism off anymore? Not sure, maybe @philbucher has an idea.

EDIT: maybe this helps: [`os.environ["OMP_NUM_THREADS"] = "1"`](https://github.com/KratosMultiphysics/Kratos/wiki/Parallel-Utilities#switch-from-openmp-based-parallelization-to-c-based-parallelization) 

I have not experienced fluctuating sensitivities in ShapeOptimization before, the way how elemental/nodal derivatives at element level are computed is slightly different though.

@sunethwarna @MFusseder tagging you here since this is definitely relevant for you, especially if this concerns compliance as well.

sunethwarna:
Hi all,

I have not experienced any race condition if i run a simulation with `OMP_NUM_THREADS=1` I use it with `ShapeOptimizationApplication` very frequently. May be you can compile it with intel compiler and run it through intel-inspector.

I had a race condition when I used intel-mkl pardiso solver to solve a system matrix with high condition number. Then I ran intel compiled krtaos through intel-inspector and found out there is a race condition happening within the intel-mkl library and it is only when you have matrices with high condition numbers.

Here is the intel inspector output
image

Apparently, there are some parameters which are also inputs and outputs, and in the case of high condition numbers they provide some detailed information using these. But this race condition was not there when using `OMP_NUM_THREADS=1`.

Link for the table of inputs and outputs in intel-mkl pardiso solver.

http://cali2.unilim.fr/intel-xe/mkl/mklman/index.htm#GUID-264E311E-ACED-4D56-AC31-E9D3B11D1CBF.htm

PR #9838, opened 2022-04-08
[Core] Improving CSR interface and low overhead conversion to PETSc

RiccardoRossi:
**📝 Description**
Provide copy free access to indices and values_vector and a low overhead interface to PETSC matrix.

Interface is in python only so no compile dependency is added. Dependency at runtime is needed only if one explicitly imports the feature

**🆕 Changelog**
- Provide copy free access to indices and values_vector and a low overhead interface to PETSC matrix.


loumalouomega (changes requested):
No test?

RiccardoRossi:
no ... i do not want to add the dependency

loumalouomega:
> no ... i do not want to add the dependency

Okay, but you can add one and use the decorator to skip the test

loumalouomega (approved):

RiccardoRossi:
and btw, the methods were already present and are used in some of the tests, for example for the scipy conversion. the difference is that now they do not do a copy

PR #9842, opened 2022-04-13
[ReadMaterialsUtility] Unused mesh id

miguelmaso:
**📝 Description**
The mesh id is unused, so it is removed.

**🆕 Changelog**
- Removed the hard-coded `mesh_id` variable from `ReadMaterialsUtility`


loumalouomega (approved):
Okay if technical committee agressé

Issue #3935, opened 2019-01-23
Discussion about strain size for the case of PLANE STRAIN


RiccardoRossi:
A decision must be taken about the strain size to be used in the case of PLANE strain.
Mentioning explicitly @KratosMultiphysics/structural-mechanics  but please anyone feel free to contribute to the conversation

The point is that AS OF NOW kratos is assuming that the strain is provided in voigt format and contains

    eps_xx  eps_yy   2*eps_xy

This is done this way since by definition eps_zz=0 in the case of plane strain, and this allows the exact same 
kinematics for 2D structural elements in the case of plane stress and plane strain.

it exists however the option of using a size-4 strain vector of the type

    eps_xx  eps_yy eps_zz  2*eps_xy

the interest of the second approach is in the context of quasi-incompressible elasticity.
For example in a Q1P0 element, one could assume that eps_zz is not exactly zero on all of the gauss points 
and only zero in average.
In a sense this second approach is completely equivalent to using a 3D law, in which the input strain
is explicitly set as eps_yz=0 eps_xz=0 eps_zz=0 (approx).

My proposal is to keep the first definition and modify the B-bar element so that it always works with 
3D constitutive laws.
I would thus propose to make the first proposal the "official" standard.

Is there agreement on this? (i would appreciate thumbs up comments if you agree, and detailed answers if you don't)

loumalouomega:
Well, I think that defining properly an enum which identifies the different options as STRESS_3D, PLANE_STRAIN_2D3Comp, PLANE_STRAIN_2D4Comp or somthing like that we can later implement the proper cases

vryy:
👍 i would advocate to define size-4 strain vector as default in 2D

frastellini:
I agree with size-3 strain vector for plane-stress and plane-strain cases, but we should use size-4 strain vector for axilsymmetric cases.

marandra:
I agree too. I think this option is the less invasive. Also, we won't need to write and maintain a plane-strain version of the CL.

RiccardoRossi:
I strongly think that we need to take a decision.
plane strain should be either with size 3 or with size 4 **consistently**.
I personally vouch for size 3, however if someone defends convincingly the size 4 option (which would imply a modified kinematics at the element level) i am open to taking that decision. @vryy what is your detailed reasoning behind your proposal (backward compatibility is not an argument here, since something would have to be broken in both cases)

[...]

PR #9849, opened 2022-04-21
[Structural] using strain size 4 in plane strain

ddiezrod:
**📝 Description**
This is just a draft of the changes needed to use strain size = 4 in plain strain. I still have to review most of them myself.
See https://github.com/KratosMultiphysics/Kratos/issues/3935


applications/StructuralMechanicsApplication/custom_elements/total_lagrangian.cpp
```diff
@@ -261,19 +261,23 @@ void TotalLagrangian::CalculateKinematicVariables(
...
-        CalculateAxisymmetricB(
-            rThisKinematicVariables.B, rThisKinematicVariables.F,
-            rThisKinematicVariables.DN_DX, rThisKinematicVariables.N);
-    } else {
-        GeometryUtils::DeformationGradient(J, rThisKinematicVariables.InvJ0,
-                                           rThisKinematicVariables.F);
-        CalculateB(rThisKinematicVariables.B, rThisKinematicVariables.F,
-                   rThisKinematicVariables.DN_DX);
-    }
+    // if (IsAxissymmetric()) {
```
loumalouomega, line 264: We should fix this @AlejandroCornejo

loumalouomega (commented):
Okay

loumalouomega:
Of course, looks like the CI fails right now

applications/StructuralMechanicsApplication/custom_elements/small_displacement_mixed_volumetric_strain_element.cpp
```diff
@@ -943,14 +943,23 @@ void SmallDisplacementMixedVolumetricStrainElement::CalculateAnisotropyTensor(co
...
     // Such minimization is constrained such that the compressibility of the new C_iso is equivalent to the input anisotropic C one
     Vector u = ZeroVector(strain_size);
-    for (IndexType i = 0; i < dim; ++i) {
-        u(i) = 1.0 / std::sqrt(dim);
-    }
-    const Matrix J = outer_prod(u, u);
     Matrix I = ZeroMatrix(strain_size);
-    for (IndexType i = 0; i < strain_size; ++i){
-        I(i, i) = i < dim ? 1.0 : 0.5;
+    if (strain_size == 4) {
```
rubenzorrilla, line 947: I'm very concerned about this... As explained in the comment above these coefficients come from a delicate minimization operation.

I've checked the current cpp and Python tests and there is no direct coverage for these operations. Having said this, we must add an extra test before merging this changes. I will take care ASAP.

rubenzorrilla (changes requested):
Blocking to prevent accidental merge until the below described test is added.

rubenzorrilla:
Why closing? This is still a pending issue...

ddiezrod:
I restored the branch in case someone wants to continue with this... It does not make sense to have open PR for years imo

rubenzorrilla:
Fair enough. Let me reopen and put it in the @KratosMultiphysics/technical-committee next meeting TODOs to decide whether we want to open Pandora's box or not.

pooyan-dadvand:
@KratosMultiphysics/technical-committee has following comments:
1. In general it is a welcome change for structural mechanics application
2. It will affect in a non trivial way the mixed element formulation (so @rubenzorrilla and @RiccardoRossi should check it)
3. What about the use in the fluid? (We shall leave the fluid CLs without change because the shear rate is the only important factor there)

Passing it to the @KratosMultiphysics/implementation-committee

PR #9850, opened 2022-04-21
Updating standard C++ to 14

roigcarlo:
**📝 Description**
Seems that some of the dependencies that we use (boost) are deprecating C++11. Without this change boost::geometry will start to fail at compile time in newer editions. (since 1.75 [source](https://github.com/boostorg/geometry))

This is important because (for example, not to blame) the cosim app requires a fairly new version of boost (1.71 onward) so this leaves a very narrow marge of versions that are compatible out of the box.

@KratosMultiphysics/altair ping because centos.

**🆕 Changelog**
 - Setting C++14 as default

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 project (KratosMultiphysics)
 cmake_minimum_required (VERSION 2.8.6)
-set (CMAKE_CXX_STANDARD 11)
+set (CMAKE_CXX_STANDARD 14)
 
 # Setting some policies
 # No recursive dereferencing


loumalouomega:
At last!

loumalouomega:
BTW, there are ways to compile with newer standards in Centos, like using spack

loumalouomega (approved):

roigcarlo:
@loumalouomega Haha, thx but this is not consensual, (yet) and its only for 14. I need to know if it does not break anything.

pooyan-dadvand (commented):
@loumalouomega I don't think it is a good idea to approve such an important change in so short time

Blocking as it is not compatible with our centos yet until we update our build with devtools8

loumalouomega:
> @loumalouomega I don't think it is a good idea to approve such an important change in so short time
> 
> Blocking as it is not compatible with our centos yet until we update our build with devtools8

I was just approving, not merging

loumalouomega (changes requested):
Blocking for now

philbucher:
Really cosim requires a specific version?

roigcarlo:
@philbucher It is not specific, but it needs to be at least 1.71

roigcarlo:
It is due to the cosim_io module btw. I think there is one PR or issue about ti, if I find it i'll link it.

philbucher:
CoSimIO does not use boost
And neither does the CoSimApp I am fairly certain 

are you confusing it with the kernel thing we were discussing some weeks ago?

roigcarlo:
@philbucher Its included through the "asio" external lib, not sure if its part of the kernel or the app.

philbucher:
CoSimIO includes ASIO directly (as does boost as boost.ASIO)
We did this intentionally to avoid the boost dependency, esp since in the long run ASIO might be added to the standard as std::networking (there has been a proposal around for a while already)

ddiezrod:
I think they are still using c++11, centos is passing...

roigcarlo:
Well I will leave this on halt since it seems that C++17 could be close.

matekelemen:
I don't want to come off as too pushy, but I'm wondering whether we could go all the way to the current C++20 standard.
The new features are awesome (especially concepts are a godsend), but the main benefit would be to lay the foundations for supporting C++ modules early.

loumalouomega:
> I don't want to come off as too pushy, but I'm wondering whether we could go all the way to the current C++20 standard.
> The new features are awesome (especially concepts are a godsend), but the main benefit would be to lay the foundations for supporting C++ modules early.

I would stay in C++17 for now. The support for the C++20 standard is partial in many compilers (even in the last versions) : https://en.cppreference.com/w/cpp/compiler_support/20

roigcarlo:
Tend to agree with @loumalouomega in this one. I would very much like to support 20, but it doesn't seem mature enough.
Also,17 is integrated in the CI and Release workflows (but not activated) but I am concerned that adding 20 will imply significant work in that area.

RiccardoRossi:
On my side i would also love to go for c++20 as all the upcoming changes in the std lib are based on that and make extensive use of concepts. Nevertheless it looks like our industrail users still rely on a rather old centos, and in particular to gcc 8.*
so i am afraid we should not move beyond c+17 for now

roigcarlo:
Closing. See #9916

loumalouomega:
Do we still need this branch @roigcarlo ?, can we remove it?

roigcarlo:
yes

PR #9851, opened 2022-04-21
[ShapeOpt] Enable use of custom python scripts in evaluation folders

sunethwarna:
**📝 Description**
This PR adds evaluation path of an iteration to `sys.path` so it enables use of custom python scripts in that folder.

**🆕 Changelog**
- Adds current evaluation path of an interation to `sys.path`

### DIFF
--- a/applications/ShapeOptimizationApplication/python_scripts/analyzers/analyzer_internal.py
+++ b/applications/ShapeOptimizationApplication/python_scripts/analyzers/analyzer_internal.py
@@ -9,9 +9,6 @@
 #
 # ==============================================================================
 
-# Making KratosMultiphysics backward compatible with python 2.6 and 2.7
-from __future__ import print_function, absolute_import, division
-
 import os, pathlib, sys
 
 # Kratos Core and Apps
--- a/applications/ShapeOptimizationApplication/python_scripts/analyzers/analyzer_internal.py
+++ b/applications/ShapeOptimizationApplication/python_scripts/analyzers/analyzer_internal.py
@@ -12,7 +12,7 @@
 # Making KratosMultiphysics backward compatible with python 2.6 and 2.7
 from __future__ import print_function, absolute_import, division
 
-import os, pathlib
+import os, pathlib, sys
 
 # Kratos Core and Apps
 import KratosMultiphysics as KM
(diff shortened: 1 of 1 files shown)


philbucher:
I would not meddle with the path
I solved the same problem with this: https://github.com/KratosMultiphysics/Kratos/blob/master/applications/CoSimulationApplication/python_scripts/factories/base_factory.py

sunethwarna:
Thats really cool.

This PR enables running custom python scripts in the simulation folder via json. If I am to adopt your method then I have following problems.
1. User does not know the folder structure created by `ShapeOptimizationApplication` because it uses internal namings for different `ResponseFunctions`, etc. Therefore making it almost impossible to specify final import file absolute directory in json.
2. If absolute paths/relative paths are used, then the simulation will be broken when someone copies it from lets say a cluster to their own PC and try to rerun again. These evaluation folders for each iterations are made such that, if someone copies them to a different directory, they should be able to run them as a single primal simulation (useful for tweaking around optimized shapes).

I am not sure whether we can make it to work with the existing objectives with your suggestion.
-

philbucher:
Not sure I follow what you mean 😅

Basically with my solution you can import any py script as long as it is on PYTHONPATH
Works exactly in the same way as the process_factory

sunethwarna:
In this case, if there are 300 iterations in the optimization, I have to add 300 directories to PYTHONPATH. Because each iterations will have a different path given by the ShapeOptimization. Thats the big problem. :/

armingeiser:
I don't see any issue with this. `sys.path` is only valid for this current python process, so adding something should not be an issue, especially if the path is removed afterwards. However I am not sure if adding at the front is a good idea though? I would add at the end of the path (append) - like that you will have the same standard behaviour and only if the script is not found in any of the default places, you will actually call your custom script. 

>  => you can import from subfolders

also if you don't know the subfolder name?

philbucher:
> also if you don't know the subfolder name?

from the description it sounded like you know that

rest is up to you

armingeiser:
> from the description it sounded like you know that

You are right, we know the subfoldername, i did not express myself well. 

> However I am not sure if adding at the front is a good idea though? I would add at the end of the path (append) 

@sunethwarna forget about this comment, the working directory of the python process comes first, so if we switch working directory during evaluation of the response, the new working directory should also come first to be consistent.

armingeiser (approved):

sunethwarna:
Thanks @armingeiser

PR #9769, opened 2022-03-16
[Core] Made Condition::GetIntegrationMethod const

EduardGomezEscandell:
This PR changes the function signature from
```c++
virtual IntegrationMethod Condition::GetIntegrationMethod();
```
to
```c++
virtual IntegrationMethod Condition::GetIntegrationMethod() const;
```
**Deprecation warning**
This breaks the API so a deprecation warning has been added for some time (See https://github.com/KratosMultiphysics/Kratos/pull/9789).

@KratosMultiphysics/all :
- This PR will be merged on April 19th, 2022.
- Applications in the CI run have been adapted.
- Any other applications and forks will break.

**🆕 Changelog**
- Makes GetIntegrationMethod method of Condition const
- Changed overrides to const

**How to fix my condition?**
If this change breaks any of your `Condition`s, simply add the const in your `GetIntegrationMethod` declaration in the `.h` header file:
```diff
- IntegrationMethod GetIntegrationMethod() override
+ IntegrationMethod GetIntegrationMethod() const override
```
You will also need to modify the method definition in the `cpp` source file if it is separate from the declaration:
```diff
- IntegrationMethod YourCondition::GetIntegrationMethod() {
+ IntegrationMethod YourCondition::GetIntegrationMethod() const {
```

PR #9852, opened 2022-04-21
Fix Delaunay Meshing Application

miguelmaso:
**📝 Description**
related to #9769 

**🆕 Changelog**
- missing const


AFranci (approved):
thanks @miguelmaso

EduardGomezEscandell (approved):
The change looks good to me, but the whole method looks over-complicated.

This would do:
```C++
CompositeCondition::IntegrationMethod CompositeCondition::GetIntegrationMethod() const
{
  if(NumberOfChildren() != 0) {
    return mChildConditions[0].GetIntegrationMethod();
  }
  return Condition::GetIntegrationMethod();
}
```

PR #9855, opened 2022-04-21
[CoSim] update CoSimIO to 4.1

philbucher:
Now the codes can also run in admin mode


pooyan-dadvand (approved):
Thanks @philbucher! It worked for us!

pooyan-dadvand:
@roigcarlo Is it possible to have a new patch release of Kratos with a cherry pick of the commit in this branch?

It would simplify my work a lot!

roigcarlo:
Yep. Actually its easier if we just make a 9.1.3 with all the changes (and faster) but if you need something excluded I can merge just this.

PR #9856, opened 2022-04-22
[TopOpt] Adding material interpolation schemes SIMP, modified SIMP and RAMP

e-dub:
**📝 Description**
This pull request adds options to choose the material interpolation methods for topology optimization.   The methods available are now
- solid isotropic material with penalization (SIMP) [1]
- modified SIMP, e.g. [2]
- rational approximation of material properties (RAMP) [3]

[1] Bendsøe, M.P. Optimal shape design as a material distribution problem. Structural Optimization 1, 193–202 (1989). https://doi.org/10.1007/BF01650949
[2] Sigmund, O. Morphology-based black and white filters for topology optimization. Struct Multidisc Optim 33, 401–424 (2007). https://doi.org/10.1007/s00158-006-0087-x
[3] Stolpe, M., Svanberg, K. An alternative interpolation scheme for minimum compliance topology optimization. Struct Multidisc Optim 22, 116–124 (2001). https://doi.org/10.1007/s001580100129

**🆕 Changelog**
- material interpolation method as topology optimization option
- Modified SIMP was only method, now an option
- Added SIMP
- Added RAMP
- Reformatting of output logging printed
- Quick fix of OC (by PH)
- Added test for RAMP
- Updated tests to work again.


applications/TopologyOptimizationApplication/custom_elements/small_displacement_simp_element.cpp
```diff
@@ -127,14 +127,40 @@ void SmallDisplacementSIMPElement::Calculate(const Variable<double> &rVariable,
...
     KRATOS_TRY
 
+
+    // Get values
+    const double E_min     = this->GetProperties()[YOUNGS_MODULUS_MIN];
+    const double E_initial = this->GetProperties()[YOUNGS_MODULUS_0];
+    const double E_current = this->GetValue(YOUNG_MODULUS);
+    const double penalty   = this->GetValue(PENAL);
+    const double x_phys    = this->GetValue(X_PHYS);
+    const std::string mat_interp = this->GetValue(MAT_INTERP);
```
PhiHo-eng: We should try to make the MAT_INTERP a Property of the mdel, so that it does not have to be saved in every element. Like we did with YOUNGS_MODULUS_0 and YOUNGS_MODULUS_MIN, as this values do not change over the optimization and also the MAT_INTERP does not change when I am not mistaken?

applications/TopologyOptimizationApplication/custom_elements/small_displacement_simp_element.cpp
```diff
@@ -127,14 +127,40 @@ void SmallDisplacementSIMPElement::Calculate(const Variable<double> &rVariable,
...
+    {
+        interp_sens += penalty * (E_initial - E_min) * pow(x_phys, penalty - 1);
+    }
+    else if (mat_interp == "ramp")
+    {
+        interp_sens += (E_initial - E_min)*(penalty+1) / pow(1-penalty*(x_phys-1), 2);
+    }
+    else
+    {
+        // not defined...through error
```
PhiHo-eng: Maybe add: "Does not name a type of the element interpolation"

applications/TopologyOptimizationApplication/custom_elements/small_displacement_simp_element.cpp
e-dub: Agreed, but both of the example you named are currently set via Python (e.g. run_TopOpt.py) and not with the values in the JSON file.

This opens up the larger discussion to the setting up of topology optimization problems:

Should the following be in topology_optimizer_factory.py?
```
model_part.GetProperties()[1].SetValue(km.YOUNG_MODULUS, 1)
model_part.GetProperties()[1].SetValue(km.POISSON_RATIO, 0.3)
model_part.GetProperties()[1].SetValue(km.DENSITY, 1)
model_part.GetProperties()[1].SetValue(kto.YOUNGS_MODULUS_MIN, 1e-09)
model_part.GetProperties()[1].SetValue(kto.YOUNGS_MODULUS_0, 1)
cons_law = ksm.LinearElastic3DLaw()
model_part.GetProperties()[1].SetValue(km.CONSTITUTIVE_LAW, cons_law)
```

Also, penalty is currently saved in every element. Should this also be saved to the part?

PhiHo-eng: Yes you are right. They should only be defined in the .JSON file, but that caused some problems, and therefore we decided to keep it like this for the moment. The values are set in the StructuralMaterials.json too.

PhiHo-eng: Yes we talked about moving penalty to the StructuralMaterials.json, but we decided to not do that as penalty is no property of the model and this would confuse the user. 
I think you are right! We should move the defintion of the Properties in the factory. And there too we could set penalty and  MAT_INTERP as a property of the model!

PhiHo-eng (dismissed):

applications/TopologyOptimizationApplication/custom_elements/small_displacement_simp_element.cpp
e-dub: At this point, I will leave the MAT_INTERP in the element (along with PENAL).  We need to redesign a bit here in the future.

PhiHo-eng (dismissed):

PhiHo-eng (approved):

PR #9868, opened 2022-04-28
[Core][Fluid] Implemented BFECC explicit strategy and specialized for Compressible Navier Stokes

EduardGomezEscandell:
**📝 Description**
Implements a generic BFECC explicit startegy in the core and a subclass for the compressible Navier-Stokes equations in the FluidDynamics application.

BFECC stands for Back and Forth Error Compensation Correction.

Also creates a common wrapper class `CompressibleNavierStokesExplicitSolvingStrategy<TBaseExplicitStategy>` to act as a base for all compressible Navier Stokes explicit startegies. This avoids duplicated boilerplate code.

The class hierarchy becomes the one here ilustrated. `.ESS.` is an abbreviation for ExplicitSolvingStrategy, `.CNS.` is an abbreviation for CompressibleNavierStokes.
```
                          ExplicitSolvingStrategy                             ]
                                     |                                        ]
                +--------------------+-------------------+                    ]   CORE
                |                                        |                    ]
          ESS.RungeKutta                            ESS.BFECC                 ]
                |                                        |
     CNS.ESS<ESS.RungeKutta>                       CNS.ESS<ESS.BFECC>         ]
                |                                        |                    ]   FluidDynamicsApplication
        CNS.ESS.RungeKutta                          CNS.ESS.BFECC             ]
```

This also means that any new explicit solving strategies will be easily expanded to the Compressible Navier Stokes equations.

**🆕 Changelog**
Core:
- Created `ExplicitStartegyBFECC`
- Created unit and convergence test for `ExplicitStartegyBFECC`.

FluidDynamicsApplication
- Created `CompressibleNavierStokesExplicitStartegy<TBaseExplicitStartegy>`
- Changed `CompressibleNavierStokesExplicitStartegyRungeKutta` to derive from `CompressibleNavierStokesExplicitStartegy<ExplicitStartegyRungeKutta>`
- Created `CompressibleNavierStokesExplicitStartegyBFECC` as a child of `CompressibleNavierStokesExplicitStartegy<ExplicitStartegyBFECC>`.

- Moved all common methods between NavierStokesRK and NavierStokesBFECC to their common base `CompressibleNavierStokesExplicitStartegy`.
- Exposed `ExplicitStartegyBFECC` and `CompressibleNavierStokesExplicitStartegyBFECC` to Python
- Added `bfecc` to the solver options for `time_scheme`.


EduardGomezEscandell:
@rubenzorrilla

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_startegy.h
```diff
@@ -0,0 +1,576 @@
+//    |  /           |
```
rubenzorrilla: There is a typo in the filename.

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_startegy.h
```diff
@@ -0,0 +1,576 @@
...
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:		 BSD License
+//					 Kratos default license: kratos/license.txt
+//
+//  Main authors:    Ruben Zorrilla, Eduard Gómez
```
rubenzorrilla:
```suggestion
//  License:             BSD License
//                            Kratos default license: kratos/license.txt
//
//  Main authors:    Ruben Zorrilla, Eduard Gómez
```
Header formatting.

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_startegy.h
```diff
@@ -0,0 +1,576 @@
...
+
+///@}
+///@name Kratos Classes
+///@{
+
+/** @brief Explicit solving strategy base class
+ * @details This is the base class from which we will derive all the explicit strategies
+ * for the compressible Navier Stokes equations (BFECC, RK4, ...)
+ */
+template <typename TBaseExplicitStartegy>
```
rubenzorrilla:
```suggestion
template <typename TBaseExplicitStrategy>
```
Typo.

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_startegy.h
```diff
@@ -0,0 +1,576 @@
...
+    typedef typename BaseType::ExplicitBuilderType ExplicitBuilderType;
+
+    /// The local vector definition
+    typedef typename TBaseExplicitStartegy::LocalSystemVectorType LocalSystemVectorType;
+
+    /// Pointer definition of CompressibleNavierStokesExplicitSolvingStrategy
+    KRATOS_CLASS_POINTER_DEFINITION(CompressibleNavierStokesExplicitSolvingStrategy);
+
+    /// Local Flags
+    KRATOS_DEFINE_LOCAL_FLAG(SHOCK_CAPTURING);
```
rubenzorrilla, line 82: Which is the purpose of this local flag?

rubenzorrilla: Why is this flag required? As far as I see you're checking the shock capturing with the nullity of the shock capturing pointer.

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_startegy.h
```diff
@@ -0,0 +1,576 @@
...
+     * @brief This method provides the defaults parameters to avoid conflicts between the different constructors
+     * @return The default parameters
+     */
+    Parameters GetDefaultParameters() const override
+    {
+        KRATOS_TRY
+
+        Parameters default_parameters = Parameters(R"(
+        {
+            "name" : "compressible_navier_stokes_explicit_explicit_solving_strategy_runge_kutta_4",
```
rubenzorrilla: This is not current class name. In other words, this field must be implemented by each class by calling the add missing parameters. Otherwise the factory mechanism we're working on will not work,

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_strategy_bfecc.h
```diff
@@ -0,0 +1,385 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:		 BSD License
+//					 Kratos default license: kratos/license.txt
+//
+//  Main authors:    Eduard Gómez
```
rubenzorrilla: Header format.

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_strategy_bfecc.h
```diff
@@ -0,0 +1,385 @@
...
+
+
+    ///@}
+    ///@name Private Operations
+    ///@{
+
+    void StashDiffusiveConstants()
+    {
+        auto& model_part = BaseType::GetModelPart();
+        if(model_part.NumberOfElements() == 0) return;
```
rubenzorrilla: I think that an error should be thrown in this case.

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_strategy_bfecc.h
```diff
@@ -0,0 +1,385 @@
...
+
+        auto& properties = model_part.ElementsBegin()->GetProperties();
+
+        auto& r_conductivity = properties.GetValue(CONDUCTIVITY);
+        auto& r_dynamic_viscosity = properties.GetValue(DYNAMIC_VISCOSITY);
```
rubenzorrilla: I think that a `Has` check has be done for each one of these.

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_strategy_bfecc.h
```diff
@@ -0,0 +1,385 @@
...
+        mDiffusionStash.dynamic_viscosity = r_dynamic_viscosity;
+
+        r_conductivity = 0;
+        r_dynamic_viscosity = 0;
+    }
+
+    void PopDiffusiveConstants()
+    {
+        auto& model_part = BaseType::GetModelPart();
+        if(model_part.NumberOfElements() == 0) return;
```
rubenzorrilla: Same error comment.

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_strategy_bfecc.h
```diff
@@ -0,0 +1,385 @@
...
+
+        auto& properties = model_part.ElementsBegin()->GetProperties();
+
+        properties.SetValue(CONDUCTIVITY, mDiffusionStash.conductivity);
+        properties.SetValue(DYNAMIC_VISCOSITY, mDiffusionStash.dynamic_viscosity);
```
rubenzorrilla: Note that by doing this you will setting a zero value if `CONDUCTIVITY` (or `DYNAMIC_VISCOSITY`) were not present in the properties container before. I think this shouldn't be done. You only reset the previously existing values.

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_strategy_runge_kutta.h
```diff
@@ -7,31 +7,26 @@
...
 #include "solving_strategies/strategies/explicit_solving_strategy_runge_kutta.h"
-#include "utilities/atomic_utilities.h"
-#include "utilities/math_utils.h"
-#include "utilities/parallel_utilities.h"
 
 // Application includes
 #include "fluid_dynamics_application_variables.h"
-#include "custom_processes/shock_capturing_physics_based_process.h"
-#include "custom_processes/shock_capturing_entropy_viscosity_process.h"
+#include "compressible_navier_stokes_explicit_solving_startegy.h"
```
rubenzorrilla: Typo.

kratos/solving_strategies/strategies/explicit_solving_strategy.h
```diff
@@ -290,6 +290,26 @@ class ExplicitSolvingStrategy : public SolvingStrategy<TSparseSpace, TDenseSpace
...
+    ExplicitBuilderType& GetExplicitBuilder()
+    {
+        KRATOS_TRY
+
+        KRATOS_DEBUG_ERROR_IF(mpExplicitBuilder == nullptr) << "The explicit builder is not initialized" << std::endl;
+        return *mpExplicitBuilder;
+
+        KRATOS_CATCH("")
+    };
+
```
rubenzorrilla: Which changes in this PR make these necessary?

kratos/solving_strategies/strategies/explicit_solving_strategy_bfecc.h
```diff
@@ -0,0 +1,546 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:		 BSD License
+//					 Kratos default license: kratos/license.txt
+//
+//  Main authors:    Ruben Zorrilla, Edurd Gómez
```
rubenzorrilla: Header format.

kratos/solving_strategies/strategies/explicit_solving_strategy_bfecc.h
```diff
@@ -0,0 +1,546 @@
...
+//
+//
+
+#if !defined(KRATOS_EXPLICIT_SOLVING_STRATEGY_BFECC)
+#define KRATOS_EXPLICIT_SOLVING_STRATEGY_BFECC
+
+/* System includes */
+#include <numeric>
```
rubenzorrilla: Why is this required?

kratos/solving_strategies/strategies/explicit_solving_strategy_bfecc.h
```diff
@@ -0,0 +1,546 @@
...
+/** @brief Family of explicit Runge-Kutta schemes
+ *
+ * @details:
+ * Formulation:
+ *
+ * WIP
+ *
+ * @tparam TSparseSpace
+ * @tparam TDenseSpace
+ */
```
rubenzorrilla: Missing doc.

kratos/solving_strategies/strategies/explicit_solving_strategy_bfecc.h
```diff
@@ -0,0 +1,546 @@
...
+        }
+
+        KRATOS_CATCH("");
+    }
+
+    /**
+     * Ovrwrite the destination buffer position with data from the source buffer position.
+     * Additionally, we save in an auxiliary vector the value of the fixed DOFs in the destination buffer position.
+     */
+    LocalSystemVectorType CopySolutionStepData(const SizeType source, const SizeType destination)
```
rubenzorrilla:
```suggestion
    LocalSystemVectorType CopySolutionStepData(
        const SizeType source,
        const SizeType destination)
```
Minor style guide suggestion.

kratos/solving_strategies/strategies/explicit_solving_strategy_bfecc.h
```diff
@@ -0,0 +1,546 @@
...
+    {
+        // Get the required data from the explicit builder and solver
+        auto& explicit_bs = BaseType::GetExplicitBuilder();
```
rubenzorrilla:
```suggestion
        auto& r_explicit_bs = BaseType::GetExplicitBuilder();
```

kratos/solving_strategies/strategies/explicit_solving_strategy_bfecc.h
```diff
@@ -0,0 +1,546 @@
...
+     */
+    virtual void FinalizeBFECCFinalSubstep() {};
+
+    /**
+     * @brief Performs an intermediate RK4 step
+     * This functions performs all the operations required in an intermediate RK4 sub step
+     * @param rFixedDofsValues The vector containing the step n+1 values of the fixed DOFs
+     * @param Substep The type of substep it is
+     */
+    virtual void PerformSubstep(const LocalSystemVectorType& rFixedDofsValues, const Substep SubstepType)
```
rubenzorrilla:
```suggestion
    virtual void PerformSubstep(
        const LocalSystemVectorType& rFixedDofsValues,
        const Substep SubstepType)
```

kratos/solving_strategies/strategies/explicit_solving_strategy_bfecc.h
```diff
@@ -0,0 +1,546 @@
...
+    {
+        KRATOS_TRY
+
+        // Get the required data from the explicit builder and solver
+        auto& explicit_bs = BaseType::GetExplicitBuilder();
```
rubenzorrilla:
```suggestion
        auto& r_explicit_bs = BaseType::GetExplicitBuilder();
```

kratos/solving_strategies/strategies/explicit_solving_strategy_bfecc.h
```diff
@@ -0,0 +1,546 @@
...
+        // Get model part and information
+        const double dt = BaseType::GetDeltaTime();
+        KRATOS_ERROR_IF(dt < 1.0e-12) << "ProcessInfo DELTA_TIME is close to zero." << std::endl;
+        auto& r_model_part = BaseType::GetModelPart();
+        auto& r_process_info = r_model_part.GetProcessInfo();
+
+        // Perform the intermidate sub step update
+
+        double time_integration_theta = -1.0;
+        double time_direction = 0.0;
```
rubenzorrilla:
```suggestion
        double time_integration_theta;
        double time_direction;
```
To make clear that these are initialized below.

kratos/solving_strategies/strategies/explicit_solving_strategy_bfecc.h
```diff
@@ -0,0 +1,546 @@
...
+    }
+
+    /**
+     * Computes the error by comparing the buffer position 0 and the previous step
+     * solution. Then stores the corrected starting point in the buffer position 1.
+     * @param rPrevStepSolution
+     */
+    void CorrectErrorAfterForwardsAndBackwards(const LocalSystemVectorType& rPrevStepSolution)
+    {
+        auto& explicit_bs = BaseType::GetExplicitBuilder();
```
rubenzorrilla:
```suggestion
        auto& r_explicit_bs = BaseType::GetExplicitBuilder();
```

kratos/solving_strategies/strategies/explicit_solving_strategy_bfecc.h
```diff
@@ -0,0 +1,546 @@
...
+
+        IndexPartition<SizeType>(r_dof_set.size()).for_each(
+            [&](const SizeType dof_index)
+            {
+                const double prev_step_solution = rPrevStepSolution[dof_index];
+                auto& r_dof = *(r_dof_set.begin() + dof_index);
+
+                const double error = (r_dof.GetSolutionStepValue(0) - prev_step_solution)/2.0;
+
+                r_dof.GetSolutionStepValue(1) = prev_step_solution - error;
```
rubenzorrilla, line 491: Maybe I'm missing something but shouldn't the error correction be applied in the current step value?

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_startegy.h
EduardGomezEscandell: This was lifted from the old implementation of the Runge-Kutta for compressible Navier Stokes. Now that you mention it, it isn't used anywhere so it can be removed.

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_strategy_bfecc.h
EduardGomezEscandell: This may happen in MPI, for instance, without being erroneous.

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_startegy.h
EduardGomezEscandell: Okay. What do we do in case of templated classes? For instance, RungeKutta. For now I used the Name() method.

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_strategy_bfecc.h
EduardGomezEscandell: Agreed. I implemented a very reduced version of C++17's std::optional to do this. When the transition is done, we can remove my struct and everything should work.

kratos/solving_strategies/strategies/explicit_solving_strategy.h
EduardGomezEscandell: The ExtractSolutionStepData method in the BFECC startegy. It's not critical (I could simply use the pGetter) but in general I think it's better to avoid pointers when not needed.

I actually believe that the current pGetExplicitBuilder should return a Pointer, not a reference to it. I believe that returning references to shared pointers is asking for trouble (because the reference counter does not change).

kratos/solving_strategies/strategies/explicit_solving_strategy_bfecc.h
EduardGomezEscandell: This may lead to undefined behaviour if we're not careful in the future. I'd personally rather risk having a wrong value than an undeterministic one.

kratos/solving_strategies/strategies/explicit_solving_strategy_bfecc.h
EduardGomezEscandell: It must be applied to both. Here it is applied only to (1), because afterwards in line 286 it is copied into (0) during CopySolutionStepData (which must be called anyways to extract the fixed dofs).

kratos/solving_strategies/strategies/explicit_solving_strategy_bfecc.h
EduardGomezEscandell: Good catch. Done.

kratos/solving_strategies/strategies/explicit_solving_strategy_bfecc.h
EduardGomezEscandell: In the end I decided to put it all in a struct and have a function initialize the values.

kratos/solving_strategies/strategies/explicit_solving_strategy_bfecc.h
rubenzorrilla: OK. Then I'd put a comment explaining why it is not copied to the current position of the database.

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_startegy.h
rubenzorrilla: That's what I was wondering. The point is that either the `name` field or the `Name` method would be required to be defined in each one of the derived classes. Not sure how we can solve it...

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_strategy_bfecc.h
rubenzorrilla: But there is no MPI support yet. I think it's better to throw an error for the moment. In any case, up to you.

EduardGomezEscandell: Ok, I'll put an error message with a coment.

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_startegy.h
EduardGomezEscandell: I'd say we derive Name in each of the classes, and templated classes should procedurally generate their own names depending on template parameters when relevant.

For instance, Runge Kutta generates its name by concatenating:
```
static std::string Name()
{
    return "compressible_navier_stokes_explicit_solving_strategy_runge_kutta_" + TButcherTableau::Name();
}
```

Maybe we can make CompressibleNavierStokesExplicitStrategy<...>::Name() virtual to force it to be derived?

rubenzorrilla: Makes sense for me. I think this would do the job.

kratos/solving_strategies/strategies/explicit_solving_strategy.h
rubenzorrilla: Agree about the pointer. I'd create a `pGetExplicitBuilderAndSolver`. My point was more on the line that this is changing the public API of the base explicit strategy class, something that should be discussed in a standalone PR.

applications/FluidDynamicsApplication/custom_strategies/strategies/compressible_navier_stokes_explicit_solving_startegy.h
EduardGomezEscandell: Well I just noticed Name() is static, so we cannot make it virtual, thus polymorphism won't help us here.

I think I'll go back to the good old recursive call to GetDefaultParameters()

kratos/solving_strategies/strategies/explicit_solving_strategy.h
EduardGomezEscandell: You're right. I'll open a separate PR.

EduardGomezEscandell:
This PR now depends on https://github.com/KratosMultiphysics/Kratos/pull/9920, which implements GetExplicitBuilder.

rubenzorrilla (approved):

PR #9871, opened 2022-04-28
bug in calculation of total force

Vahid-Galavi:
**📝 Description**
A bug that leads to incorrect total force in the out-of-plane direction in the UP elements is solved.


loumalouomega (approved):
Ok for me

PR #9873, opened 2022-04-29
[Shallow] Add check for benchmarking utilities

miguelmaso:
**📝 Description**
This PR adds some checks to avoid segfaults related to wrong project user settings when some benchmark tools are used.

**🆕 Changelog**
- ConvergenceOutputProcess [HDF5]: Initialize non historical variable
- ConvergenceOutputProcess [HDF5]: Check the dtype from previous simulations
- BaseBenchmarkProcess: measuring time


roigcarlo (approved):
:+1:

PR #9874, opened 2022-04-29
[Core] Newton raphson strategy - not calling InitializeSolutionStep in Predict

### LINKED ISSUE #9870: [Core] Problem when running same time step with smaller dt
**Description**
We are including what we call "roll back" in our structural solver inside the altair. The idea is simply that if SolveSolutionStep does not converge we reduce the time step in half and try again.
The problem we have is that when we reduce the time step and call InitializeSolutionStep for the next try, it is actually not doing anything as the flag mSolutionStepIsInitialized is true at that moment:

see https://github.com/KratosMultiphysics/Kratos/blob/3acacff20d1d0c815d85310c2dc58d75dab1b463/kratos/solving_strategies/strategies/residualbased_newton_raphson_strategy.h#L797

I am not sure why this flag is there, it seems its being there since the very beginning, but I think it could be removed. The only use I see of this flag is that we avoid calling InitializeSolutionStep twice when we call the Predict later. But in principle, I think we can consider that these functions are completely separated, removing the call to InitializeSolutionStep that is inside the Predict, so the flag is no longer needed.

Maybe there are some dark reasons to keep as it is right now, but in my opinion removing the flag would make it cleaner (and also would solve my problem : D )

@KratosMultiphysics/technical-committee @KratosMultiphysics/implementation-committee

ddiezrod:
**📝 Description**
Fixes https://github.com/KratosMultiphysics/Kratos/issues/9870 , not calling InitializeSolutionStep in Predict.
For the moment this is just to check if nothing gets broken with this change.

### DIFF
--- a/kratos/solving_strategies/strategies/residualbased_newton_raphson_strategy.h
+++ b/kratos/solving_strategies/strategies/residualbased_newton_raphson_strategy.h
@@ -789,25 +789,24 @@ class ResidualBasedNewtonRaphsonStrategy
         typename TBuilderAndSolverType::Pointer p_builder_and_solver = GetBuilderAndSolver();
         ModelPart& r_model_part = BaseType::GetModelPart();
 
-        //set up the system, operation performed just once unless it is required
-        //to reform the dof set at each iteration
+        // Set up the system, operation performed just once unless it is required
+        // to reform the dof set at each iteration
         BuiltinTimer system_construction_time;
-        if (p_builder_and_solver->GetDofSetIsInitializedFlag() == false ||
-            mReformDofSetAtEachStep == true)
+        if (!p_builder_and_solver->GetDofSetIsInitializedFlag() || mReformDofSetAtEachStep)
         {
-            //setting up the list of the DOFs to be solved
+            // Setting up the list of the DOFs to be solved
             BuiltinTimer setup_dofs_time;
             p_builder_and_solver->SetUpDofSet(p_scheme, r_model_part);
             KRATOS_INFO_IF("ResidualBasedNewtonRaphsonStrategy", BaseType::GetEchoLevel() > 0) << "Setup Dofs Time: "
                 << setup_dofs_time.ElapsedSeconds() << std::endl;
 
-            //shaping correctly the system
+            // Shaping correctly the system
             BuiltinTimer setup_system_time;
             p_builder_and_solver->SetUpSystem(r_model_part);
             KRATOS_INFO_IF("ResidualBasedNewtonRaphsonStrategy", BaseType::GetEchoLevel() > 0) << "Setup System Time: "
                 << setup_system_time.ElapsedSeconds() << std::endl;
 
-            //setting up the Vectors involved to the correct size
+            // Setting up the Vectors involved to the correct size
             BuiltinTimer system_matrix_resize_time;
             p_builder_and_solver->ResizeAndInitializeVectors(p_scheme, mpA, mpDx, mpb,
                                                                 r_model_part);
@@ -829,7 +828,7 @@ class ResidualBasedNewtonRaphsonStrategy
         p_scheme->InitializeSolutionStep(r_model_part, rA, rDx, rb);
 
         // Initialisation of the convergence criteria
-        if (mpConvergenceCriteria->GetActualizeRHSflag() == true)
+        if (mpConvergenceCriteria->GetActualizeRHSflag())
         {
             TSparseSpace::SetToZero(rb);
             p_builder_and_solver->BuildRHS(p_scheme, r_model_part, rb);
@@ -837,8 +836,9 @@ class ResidualBasedNewtonRaphsonStrategy
 
         mpConvergenceCriteria->InitializeSolutionStep(r_model_part, p_builder_and_solver->GetDofSet(), rA, rDx, rb);
 
-        if (mpConvergenceCriteria->GetActualizeRHSflag() == true)
+        if (mpConvergenceCriteria->GetActualizeRHSflag()) {
             TSparseSpace::SetToZero(rb);
+        }
 
         KRATOS_CATCH("");
     }
--- a/kratos/solving_strategies/strategies/residualbased_newton_raphson_strategy.h
+++ b/kratos/solving_strategies/strategies/residualbased_newton_raphson_strategy.h
@@ -130,7 +130,6 @@ class ResidualBasedNewtonRaphsonStrategy
      */
     explicit ResidualBasedNewtonRaphsonStrategy(ModelPart& rModelPart, Parameters ThisParameters)
         : BaseType(rModelPart),
-          mSolutionStepIsInitialized(false),
           mInitializeWasPerformed(false),
           mKeepSystemConstantDuringIterations(false)
     {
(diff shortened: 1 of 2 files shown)


ddiezrod:
It seems the changes do not break anything. Leaving the approval for the @KratosMultiphysics/technical-committee

loumalouomega (changes requested):
I need to look at this with care, looks like a very relevant change

kratos/solving_strategies/strategies/residualbased_newton_raphson_strategy.h
```diff
@@ -794,65 +784,61 @@ class ResidualBasedNewtonRaphsonStrategy
...
+        typename TBuilderAndSolverType::Pointer p_builder_and_solver = GetBuilderAndSolver();
+        ModelPart& r_model_part = BaseType::GetModelPart();
 
-            KRATOS_INFO_IF("ResidualBasedNewtonRaphsonStrategy", BaseType::GetEchoLevel() > 0) << "System Construction Time: "
-                << system_construction_time.ElapsedSeconds() << std::endl;
+        //set up the system, operation performed just once unless it is required
+        //to reform the dof set at each iteration
+        BuiltinTimer system_construction_time;
+        if (p_builder_and_solver->GetDofSetIsInitializedFlag() == false ||
+            mReformDofSetAtEachStep == true)
```
loumalouomega:
```suggestion
            mReformDofSetAtEachStep)
```
Simpler

kratos/solving_strategies/strategies/residualbased_newton_raphson_strategy.h
```diff
@@ -794,65 +784,61 @@ class ResidualBasedNewtonRaphsonStrategy
...
-            }
+        // Pointers needed in the solution
+        typename TSchemeType::Pointer p_scheme = GetScheme();
+        typename TBuilderAndSolverType::Pointer p_builder_and_solver = GetBuilderAndSolver();
+        ModelPart& r_model_part = BaseType::GetModelPart();
 
-            KRATOS_INFO_IF("ResidualBasedNewtonRaphsonStrategy", BaseType::GetEchoLevel() > 0) << "System Construction Time: "
-                << system_construction_time.ElapsedSeconds() << std::endl;
+        //set up the system, operation performed just once unless it is required
+        //to reform the dof set at each iteration
```
loumalouomega:
```suggestion
        // To reform the dof set at each iteration
```

kratos/solving_strategies/strategies/residualbased_newton_raphson_strategy.h
```diff
@@ -794,65 +784,61 @@ class ResidualBasedNewtonRaphsonStrategy
...
-                    << system_matrix_resize_time.ElapsedSeconds() << std::endl;
-            }
+        // Pointers needed in the solution
+        typename TSchemeType::Pointer p_scheme = GetScheme();
+        typename TBuilderAndSolverType::Pointer p_builder_and_solver = GetBuilderAndSolver();
+        ModelPart& r_model_part = BaseType::GetModelPart();
 
-            KRATOS_INFO_IF("ResidualBasedNewtonRaphsonStrategy", BaseType::GetEchoLevel() > 0) << "System Construction Time: "
-                << system_construction_time.ElapsedSeconds() << std::endl;
+        //set up the system, operation performed just once unless it is required
```
loumalouomega:
```suggestion
        // Set up the system, operation performed just once unless it is required
```

kratos/solving_strategies/strategies/residualbased_newton_raphson_strategy.h
```diff
@@ -794,65 +784,61 @@ class ResidualBasedNewtonRaphsonStrategy
...
+        //to reform the dof set at each iteration
+        BuiltinTimer system_construction_time;
+        if (p_builder_and_solver->GetDofSetIsInitializedFlag() == false ||
+            mReformDofSetAtEachStep == true)
+        {
+            //setting up the list of the DOFs to be solved
```
loumalouomega:
```suggestion
            // Setting up the list of the DOFs to be solved
```

kratos/solving_strategies/strategies/residualbased_newton_raphson_strategy.h
```diff
@@ -794,65 +784,61 @@ class ResidualBasedNewtonRaphsonStrategy
...
+            BuiltinTimer setup_dofs_time;
+            p_builder_and_solver->SetUpDofSet(p_scheme, r_model_part);
+            KRATOS_INFO_IF("ResidualBasedNewtonRaphsonStrategy", BaseType::GetEchoLevel() > 0) << "Setup Dofs Time: "
+                << setup_dofs_time.ElapsedSeconds() << std::endl;
+
+            //shaping correctly the system
```
loumalouomega:
```suggestion
            // Shaping correctly the system
```

kratos/solving_strategies/strategies/residualbased_newton_raphson_strategy.h
```diff
@@ -794,65 +784,61 @@ class ResidualBasedNewtonRaphsonStrategy
...
+            BuiltinTimer setup_system_time;
+            p_builder_and_solver->SetUpSystem(r_model_part);
+            KRATOS_INFO_IF("ResidualBasedNewtonRaphsonStrategy", BaseType::GetEchoLevel() > 0) << "Setup System Time: "
+                << setup_system_time.ElapsedSeconds() << std::endl;
+
+            //setting up the Vectors involved to the correct size
```
loumalouomega:
```suggestion
            // Setting up the Vectors involved to the correct size
```

kratos/solving_strategies/strategies/residualbased_newton_raphson_strategy.h
```diff
@@ -794,65 +784,61 @@ class ResidualBasedNewtonRaphsonStrategy
...
+            p_builder_and_solver->BuildRHS(p_scheme, r_model_part, rb);
+        }
 
-            if (mpConvergenceCriteria->GetActualizeRHSflag() == true)
-                TSparseSpace::SetToZero(rb);
+        mpConvergenceCriteria->InitializeSolutionStep(r_model_part, p_builder_and_solver->GetDofSet(), rA, rDx, rb);
 
-            mSolutionStepIsInitialized = true;
-        }
+        if (mpConvergenceCriteria->GetActualizeRHSflag() == true)
```
loumalouomega:
```suggestion
        if (mpConvergenceCriteria->GetActualizeRHSflag())
```
Simpler

loumalouomega (commented):
Looks OK, I will remove my block as soon as @KratosMultiphysics/technical-committee agrees

kratos/solving_strategies/strategies/residualbased_newton_raphson_strategy.h
```diff
@@ -638,10 +633,6 @@ class ResidualBasedNewtonRaphsonStrategy
         if (mInitializeWasPerformed == false)
             Initialize();
```
miguelmaso, line 634: To be consistent, this should be removed as well. However, I'd wait to another PR

ddiezrod: good point, if we agree on this, I will continue with that later

ddiezrod:

RiccardoRossi (approved):
@KratosMultiphysics/technical-committee agrees about this (provided that tests pass)

loumalouomega (approved):
As long a test passes Okay

PR #9875, opened 2022-04-29
[Timer] Sub intervals and fix spelling

miguelmaso:
**📝 Description**
Again, more cleanup and enhancements for the timer. Now, it's possible to use sub-intervals, e.g:
```py
class MyPotentiallyExpensiveProcess:
    def ExecuteBeforeSolutionStep(self):
        KM.Timer.Start("MyExpensiveProcess/Task1")
        task1()
        KM.Timer.Stop("MyExpensiveProcess/Task1")
    def ExecuteFinalizeSolutionStep(self):
        KM.Timer.Start("MyExpensiveProcess/Task2")
        task2()
        KM.Timer.Stop("MyExpensiveProcess/Task2")
```
I don't know how to add a test. However, it's a very simple function.

**🆕 Changelog**
- Add sub-intervals
- Deprecate misspelled method


kratos/python_scripts/timer_process.py
```diff
@@ -70,5 +70,4 @@ def ExecuteFinalize(self):
         """
         self.timer.Stop(self.interval_name)
         self.timer.PrintTimingInformation()
-        if self.output_filename != "":
```
miguelmaso, line 73: The Timer performs this check internally, there is no need to check again

kratos/utilities/timer.cpp
```diff
@@ -155,6 +187,12 @@ int Timer::SetOuputFile(std::string const& rOutputFileName)
 }
 
 int Timer::CloseOuputFile()
```
loumalouomega, line 190: Can you use instead the deprecation macro?

kratos/utilities/timer.h
```diff
@@ -191,11 +191,13 @@ class KRATOS_API(KRATOS_CORE) Timer
      * @param rOutputFileName The name of the output file
      */
     static int SetOuputFile(std::string const& rOutputFileName);
+    static int SetOutputFile(std::string const& rOutputFileName);
```
loumalouomega, line 193: Use deprecation macro

loumalouomega (changes requested):
Use deprecation macro, and sorry for the typo

kratos/utilities/timer.cpp
miguelmaso: I've just tried, but it adds a warning at compile time. I wanted to add a warning at run time.
Should I add both?

loumalouomega: I understand, can you add both?, that way we will remove the c++ calls in addition to the python calls

miguelmaso: both are added

kratos/utilities/timer.h
miguelmaso: done

ddiezrod:
Looks ok to me ;)

miguelmaso:
Which is the status of this? Can someone approve or it should be closed?

loumalouomega (dismissed):
Remove block

loumalouomega:
@KratosMultiphysics/technical-committee and @KratosMultiphysics/implementation-committee  can you take a look?

miguelmaso:
Which is the status of this?

loumalouomega:
> Which is the status of this?

Now is a good moment to rediscuss

PR #9881, opened 2022-05-03
[FSI] Minor fixes, robustness and deprecations

rubenzorrilla:
**📝 Description**
As reported by @maxnezdyur there are some embedded FSI cases in the example repo that used to work but do not so right now.

After doing reverse engineering, I realized that the cases stopped working when we implemented the IBQN coupling. The point is that before we were always using a pressure load in the structure (meaning that the condition normal is used). In order to do the IBQN coupling, we need the interface residuals of force and displacement to match their sizes, so we switched to a nodal line load (meaning that the nodal normal is used). This slight difference, seems to affect robustness in some of the most complex cases such as the flexible embedded mixer.

Hence, I decided to continue using a pressure load in the IQN case (as we were always using) and use the line one (since we are obliged) in the IBQN one.

Last but not least, I took the chance to also clean up things a bit.


applications/FSIApplication/python_scripts/partitioned_embedded_fsi_base_solver.py
```diff
@@ -177,7 +177,9 @@ def _AdvanceInTimeCouplingInterfaces(self, new_time):
         # Even though these are auxiliary model parts, this is mandatory to be done to properly set up the database
         # Note that if this operations are removed, some auxiliary utils (e.g. FM-ALE algorithm in embedded) will perform wrong
         self._GetFSICouplingInterfaceFluid().GetInterfaceModelPart().GetRootModelPart().CloneTimeStep(new_time)
+        self._GetFSICouplingInterfaceFluid().GetInterfaceModelPart().ProcessInfo[KratosMultiphysics.STEP] = self._GetFSICouplingInterfaceFluid().GetFatherModelPart().ProcessInfo[KratosMultiphysics.STEP]
```
sunethwarna, line 180: Curious, is it only the `STEP` in here?

sunethwarna (approved):

applications/FSIApplication/python_scripts/partitioned_embedded_fsi_base_solver.py
rubenzorrilla: What you mean? If the only variable in that `ProcessInfo` is `STEP`? If so it is, I'd say yes. Indeed, even `STEP` is not strictly required by the coupling. I added it to make possible outputting the interface model parts.

PR #9886, opened 2022-05-05
[FluidBio] Addition of FluidDynamicsBiomedicalApplication

rubenzorrilla:
**📝 Description**
This PR adds a new `FluidDynamicsBiomedicalApplication` application to Kratos. As you can see it is conceived as an extension of the current `FluidDynamicsApplication` containing the extra functionalities required for the analysis of blood vessels. Such functionalities were implemented in an old version of Kratos by @soudah. This PR makes these more efficient while adding some new capabilities.

Application has been added to CI. About the requirements for this:

- [x] Testing: Right now it has full coverage.
- [x] Maintainers: The two maintainers are @rubenzorrilla (myself) and @soudah.
- [x] Testing total time: The impact of these is negligible compared to the current time.

**🆕 Changelog**
- New `FluidDynamicsBiomedicalApplication`


applications/FluidDynamicsBiomedicalApplication/FluidDynamicsBiomedicalApplication.py
```diff
@@ -0,0 +1,7 @@
+from KratosFluidDynamicsBiomedicalApplication import *
```
sunethwarna, line 2: Woudn't it be better to import `FluidDynamicsApplication` here since there is a compile time dependency?

applications/FluidDynamicsBiomedicalApplication/custom_utilities/parabolic_profile_utilities.cpp
```diff
@@ -0,0 +1,241 @@
...
+        auto& r_normal = std::get<0>(rTLS);
+        auto& r_wall_vect = std::get<1>(rTLS);
+
+        // Get condition geometry and calculate normal
+        auto& r_geom = rCondition.GetGeometry();
+        r_normal = r_geom.UnitNormal(0,GeometryData::IntegrationMethod::GI_GAUSS_1);
+
+        // Get parent geometry to search for the first layer of nodes
+        const auto& r_neighbours = rCondition.GetValue(NEIGHBOUR_ELEMENTS);
+        KRATOS_ERROR_IF(r_neighbours.size() == 0) << "No parent found in condition with id. " << rCondition.Id() << "." << std::endl;
```
sunethwarna:
```suggestion
        KRATOS_ERROR_IF(r_neighbours.size() == 0) << "No parent found in condition with id " << rCondition.Id() << " in " <<  rWallModelPart.FullName() << "." << std::endl;
```

I would put the model part name also in the error.

applications/FluidDynamicsBiomedicalApplication/custom_utilities/parabolic_profile_utilities.cpp
```diff
@@ -0,0 +1,241 @@
...
+    parallel_distance_settings["max_distance"].SetDouble(char_length);
+    const std::size_t domain_size = rFluidModelPart.GetProcessInfo()[DOMAIN_SIZE];
+
+    // Calculate parallel distance
+    // Note that we check that the initial maximum levels are sufficient
+    double current_max_dist = char_length;
+    std::size_t current_max_levels = WallDistanceLevels / 2;
+    while (std::abs(current_max_dist - char_length) < 1.0e-12) {
+        // Calculate the parallel distance with the new number of levels
+        parallel_distance_settings["max_levels"].SetInt(2 * current_max_levels);
```
sunethwarna: Why do you set `"max_levels"` within the while loop? I don't see how `current_max_levels` are changed within the while loop.

sunethwarna: Isn't this expensive? Wouldn't it be possible to do trial run before the simulation (with only this process, and no analysis stage) to check the `WALL_DISTANCE`?

applications/FluidDynamicsBiomedicalApplication/custom_utilities/wss_statistics_utilities.cpp
```diff
@@ -0,0 +1,151 @@
...
+            // Calculate the FACE_LOAD (distributed REACTION) projections
+            const auto& r_face_load = rNode.GetValue(FACE_LOAD);
+            const double projection = inner_prod(r_face_load, normal);
+            normal_load = projection * normal;
+            tangential_load = r_face_load - normal_load;
+
+            // Save computed magnitudes
+            rNode.GetValue(WSS) = norm_2(tangential_load);
+            rNode.GetValue(WSS_NORMAL_STRESS) = normal_load;
+            rNode.GetValue(WSS_TANGENTIAL_STRESS) = tangential_load;
```
sunethwarna, line 109: These `static` methods are standalone. I would use `SetValue` before this `block_for_each` to ensure these variables do exists in the node.

applications/FluidDynamicsBiomedicalApplication/custom_utilities/wss_statistics_utilities.cpp
```diff
@@ -0,0 +1,151 @@
...
+            double& r_osi = rNode.GetValue(OSI);
+            r_osi = r_tawss / r_twss > 1.0 ? 0.0 : 0.5 * (1.0 - r_tawss/r_twss);
+            if (r_twss > 1.0e-12) {
+                if (std::abs(r_osi - 0.5) < 1.0e-12) {
+                    rNode.GetValue(RRT) = 0.0;
+                } else {
+                    rNode.GetValue(RRT) = 1.0 / ((1.0 - 2.0*r_osi) * r_twss);
+                }
+            }
+            rNode.GetValue(ECAP) = r_osi / r_twss;
```
sunethwarna, line 146: These `static` methods are standalone. I would use `SetValue` before this `block_for_each` to ensure these variables do exists in the node.

applications/FluidDynamicsBiomedicalApplication/custom_utilities/wss_statistics_utilities.cpp
```diff
@@ -0,0 +1,151 @@
...
+            r_tawss = norm_2(r_temporal_osi / step);
+
+            // Calculate OSI and OSI-dependent magnitudes
+            double& r_osi = rNode.GetValue(OSI);
+            r_osi = r_tawss / r_twss > 1.0 ? 0.0 : 0.5 * (1.0 - r_tawss/r_twss);
+            if (r_twss > 1.0e-12) {
+                if (std::abs(r_osi - 0.5) < 1.0e-12) {
+                    rNode.GetValue(RRT) = 0.0;
+                } else {
+                    rNode.GetValue(RRT) = 1.0 / ((1.0 - 2.0*r_osi) * r_twss);
```
sunethwarna, line 143: Is there a possibility to `r_twss` to be zero?

applications/FluidDynamicsBiomedicalApplication/custom_utilities/wss_statistics_utilities.cpp
```diff
@@ -0,0 +1,151 @@
...
+        rNode.SetValue(RRT, 0.0);
+        rNode.SetValue(OSI, 0.0);
+        rNode.SetValue(ECAP, 0.0);
+        rNode.SetValue(TWSS, 0.0);
+        rNode.SetValue(TAWSS, 0.0);
+        rNode.SetValue(TEMPORAL_OSI, aux_zero);
+        rNode.SetValue(WSS_NORMAL_STRESS, aux_zero);
+        rNode.SetValue(WSS_TANGENTIAL_STRESS, aux_zero);
+    });
+}
```
sunethwarna, line 46: Why a separate method? Can't we integrate these `SetValue` methods in corresponding methods where they are set? That way there is less room for errors.

applications/FluidDynamicsBiomedicalApplication/fluid_dynamics_biomedical_application.h
```diff
@@ -0,0 +1,245 @@
...
+
+///@}
+///@name  Enum's
+///@{
+
+///@}
+///@name  Functions
+///@{
+
+///@}
```
sunethwarna, line 57: Are these empty groups necessary?

applications/FluidDynamicsBiomedicalApplication/custom_utilities/parabolic_profile_utilities.cpp
```diff
@@ -0,0 +1,241 @@
...
+    // Get the maximum distance to the wall
+    const double max_dist = block_for_each<MaxReduction<double>>(rModelPart.Nodes(), [](NodeType& rNode){
+        return rNode.GetValue(WALL_DISTANCE);
+    });
+    KRATOS_ERROR_IF(std::abs(max_dist) < 1.0e-12) << "WALL_DISTANCE is close to zero." << std::endl;
+
+    // Impose the parabolic profile values
+    block_for_each(rModelPart.Nodes(), array_1d<double,3>(), [time, max_dist, &rMaxParabolaValue](NodeType& rNode, array_1d<double,3>& rTLSValue){
+        //Calculate distance for each node
+        const double wall_dist = rNode.GetValue(WALL_DISTANCE) < 0.0 ? 0.0 : rNode.GetValue(WALL_DISTANCE);
```
sunethwarna, line 163: Do negative `WALL_DISTANCE` make sense in this app? If not, i would check and set this in the `CalculateWallParallelDistance`.

applications/FluidDynamicsBiomedicalApplication/custom_utilities/parabolic_profile_utilities.cpp
```diff
@@ -0,0 +1,241 @@
...
+        // Calculate the inlet value module
+        const double max_value = GetMaxParabolaValue(time, rNode, rMaxParabolaValue);
+        const double value_in = max_value * (1.0-(std::pow(max_dist-wall_dist,2)/std::pow(max_dist,2)));
+        rTLSValue *= value_in;
+
+        // Set and fix the VELOCITY DOFs
+        rNode.FastGetSolutionStepValue(VELOCITY) = rTLSValue;
+        rNode.Fix(VELOCITY_X);
+        rNode.Fix(VELOCITY_Y);
+        rNode.Fix(VELOCITY_Z);
```
sunethwarna, line 184: Is it necessary to fix always when this transient value is computed? Wouldn't it be better if fixing of variables are moved out and only called once?

applications/FluidDynamicsBiomedicalApplication/python_scripts/apply_parabolic_inlet_process.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def ExecuteFinalizeSolutionStep(self):
+        # Here we free all of the nodes in the inlet
+        if self.step_is_active:
+            inlet_model_part = self.model.GetModelPart(self.settings["inlet_model_part_name"].GetString())
+            KratosBio.ParabolicProfileUtilities.FreeParabolicInlet(inlet_model_part)
+        self.step_is_active = False
+
+    @classmethod
+    def _ReturnDistanceProcessPrototype(cls, domain_size):
```
sunethwarna, line 113: Why `classmethod` and not `staticmethod`?

applications/FluidDynamicsBiomedicalApplication/python_scripts/apply_parabolic_inlet_process.py
```diff
@@ -0,0 +1,117 @@
+# Importing the Kratos Library
+import KratosMultiphysics
+
+# Import applications
+import KratosMultiphysics.FluidDynamicsBiomedicalApplication as KratosBio
+
+def Factory(settings, model):
+    if(type(settings) != KratosMultiphysics.Parameters):
```
sunethwarna: The new way would be to use `isinstance`

applications/FluidDynamicsBiomedicalApplication/python_scripts/compute_wss_statistics_process.py
```diff
@@ -0,0 +1,67 @@
+# Importing the Kratos Library
+import KratosMultiphysics
+
+# Import applications
+import KratosMultiphysics.FluidDynamicsBiomedicalApplication as KratosBio
+
+def Factory(settings, model):
+    if(type(settings) != KratosMultiphysics.Parameters):
```
sunethwarna: The new way would be to use `isinstance`

applications/FluidDynamicsBiomedicalApplication/FluidDynamicsBiomedicalApplication.py
rubenzorrilla: OK. Makes sense.

applications/FluidDynamicsBiomedicalApplication/custom_utilities/parabolic_profile_utilities.cpp
rubenzorrilla: Cool.

applications/FluidDynamicsBiomedicalApplication/custom_utilities/wss_statistics_utilities.cpp
rubenzorrilla: They are static but expected to be called at each output process so it's better to keep the get for the sake of performance. Note that these methods will be always called from the `compute_wss_statistics_process.py` that calls the initialization in the `ExecuteInitialize`.

applications/FluidDynamicsBiomedicalApplication/custom_utilities/wss_statistics_utilities.cpp
rubenzorrilla: Same comment.

applications/FluidDynamicsBiomedicalApplication/custom_utilities/wss_statistics_utilities.cpp
rubenzorrilla: See my comment above.

applications/FluidDynamicsBiomedicalApplication/fluid_dynamics_biomedical_application.h
rubenzorrilla: Nope. I just followed our standard doxygen format.

applications/FluidDynamicsBiomedicalApplication/python_scripts/apply_parabolic_inlet_process.py
rubenzorrilla: Changing it.

applications/FluidDynamicsBiomedicalApplication/python_scripts/compute_wss_statistics_process.py
rubenzorrilla: Done. Copy and pasting leftover.

applications/FluidDynamicsBiomedicalApplication/python_scripts/apply_parabolic_inlet_process.py
rubenzorrilla: Done. Copy and pasting leftover.

applications/FluidDynamicsBiomedicalApplication/custom_utilities/parabolic_profile_utilities.cpp
rubenzorrilla: No it needs to be fixed and unfixed at each time step as we normally do in all processes. The point of doing it in here is that I avoid looping the nodes twice, one to assign the value and one to fix it.

applications/FluidDynamicsBiomedicalApplication/custom_utilities/parabolic_profile_utilities.cpp
rubenzorrilla: Indeed the boundary values are reset to zero when calculating the distance. In any case I'd keep it just in case as the performance impact is negligible.

applications/FluidDynamicsBiomedicalApplication/custom_utilities/parabolic_profile_utilities.cpp
rubenzorrilla: Now `current_max_levels` is updated within the loop. Thanks for spotting it. This ensures that the layers are enough to cover the entire blood vessel thickness. It is done once at the beginning of the simulation so performance isn't an issue.

applications/FluidDynamicsBiomedicalApplication/custom_utilities/wss_statistics_utilities.cpp
sunethwarna: I was thinking of may be having `SetValue` methods outside the `block_for_each` and use `GetValue` as done in this PR. So, there won't be any performance hit, and also don't have to use `ExecuteInitialize` as well.

applications/FluidDynamicsBiomedicalApplication/custom_utilities/wss_statistics_utilities.cpp
rubenzorrilla: Nope. Note that we're checking it a few lines above.

applications/FluidDynamicsBiomedicalApplication/custom_utilities/wss_statistics_utilities.cpp
rubenzorrilla: But then where would you do the `SetValue`? You cannot do it before the `block_for_each` as some of these variables depend on their previous value. Besides you'd be looping the nodes twice. I've the feeling I'm not getting your point at all...

sunethwarna: Now i see, I was thinking since they are static methods they should be standalone. I get your point. I am just curious on why did you make `CalculateOSI` method static? it could have been a class member method so there can be an `Initialize` method indicating this mehod has a life cycle beyond `CalculateOSI` method.

rubenzorrilla: No reason at all. Most probably to keep it as simple as possible while making clear that nothing is stored and that the WSS needs to be independently called.

sunethwarna (approved):
I'm ok with the changes. Thanks @rubenzorrilla . Btw there are CentOS test failures in lots of PRs.

PR #9887, opened 2022-05-06
[Fluid] Created process to compute pressure coefficient

EduardGomezEscandell:
**📝 Description**
A similar process existed for the potential flow formulation, so this one fills the same niche in the context of Navier-Stokes equations (both compressible and incompressible, or really any formulation with nodal historical pressure).

Uses the formula:
```
cp = (p - p∞) / 0.5*ρ∞*V∞²
```
for pressure coefficient.

**🆕 Changelog**
- Created process: `ComputeAerodynamicCoefficientsProcess`
- Created unit test


mpentek:
Why do not leave it as a simple postprocessing step?

For the drag and lift coefficient you neglect the viscous components.

Also, it is not generic, whereas the FluidDynamicsApplication is generic. With that I mean, one could formulate the total force Fx, Fy, Fz (we actually have it), also the respective moments. From here respective coefficients can be calculated.

The naming drag and lift is often used, but it is only "correct" as naming for 2d-type geometries (airfoils, wind turbine blades, perhaps vehicles). In the generic context (here also looking at wind and generic flow) "drag" is well defined, lift not anymore.

@rubenzorrilla @sunethwarna there was a discussion to have a generic way of calculating the total force (independent of the flow model and the boundary condition). That would be the way to go I think. 

@EduardGomezEscandell My may concern is with your draft that it tries to fill a niche for the generic Navier-Stokes, remaining still too specific. For this case omitting viscous forces and deriving respective force coefficient for ONLY "drag" and "lift". You correctly say we deal with the generic Navier-Stokes, yet the implementation is still specific for a certain application case (or a class of it).

EduardGomezEscandell:
@rubenzorrilla

applications/FluidDynamicsApplication/custom_processes/compute_pressure_coefficient_process.h
```diff
@@ -0,0 +1,199 @@
...
+///@{
+
+///@name Kratos Globals
+///@{
+
+///@}
+///@name Type Definitions
+///@{
+
+using NodeType = ModelPart::NodeType;
```
rubenzorrilla: Here you're defining the `NodeType` for the entire Kratos namespace. The definition should go into the class.

applications/FluidDynamicsApplication/custom_processes/compute_pressure_coefficient_process.h
```diff
@@ -0,0 +1,199 @@
...
+///@}
+///@name  Functions
+///@{
+
+///@}
+///@name Kratos Classes
+///@{
+
+/// Utility to modify the distances of an embedded object in order to avoid bad intersections
+/// Besides, it also deactivate the full negative distance elements
```
rubenzorrilla: Please update documentation.

applications/FluidDynamicsApplication/custom_processes/compute_pressure_coefficient_process.h
```diff
@@ -0,0 +1,199 @@
...
+    KRATOS_CLASS_POINTER_DEFINITION(ComputePressureCoefficientProcess);
+
+    ///@}
+    ///@name Life Cycle
+    ///@{
+
+    /// Constructor with Kratos parameters.
+    ComputePressureCoefficientProcess(
+        ModelPart& rModelPart,
+        Parameters Params);
```
rubenzorrilla: If not strictly required, I'd remove the old model part based constructor.

applications/FluidDynamicsApplication/custom_processes/compute_pressure_coefficient_process.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    Params.ValidateAndAssignDefaults(default_parameters);
+
+    mComputeAsPostProcess = Params["compute_only_as_postprocess"].GetBool();
+
+    mFreestreamStaticPressure = Params["freestream_pressure"].GetDouble();
+
+    const double freestream_density = Params["freestream_density"].GetDouble();
+    const double free_stream_velocity = Params["freestream_velocity"].GetDouble();
+    mFreestreamDynamicPressure = 0.5 * freestream_density * free_stream_velocity * free_stream_velocity;
```
rubenzorrilla: An error must be thrown if this ends up being zero.

applications/FluidDynamicsApplication/custom_processes/compute_pressure_coefficient_process.cpp
```diff
@@ -0,0 +1,115 @@
...
+        "model_part_name"     : "PLEASE_PROVIDE_A_MODELPART_NAME",
+        "freestream_density" : 0,
+        "freestream_velocity" : 0,
+        "freestream_pressure" : 0,
+        "compute_only_as_postprocess" : true
+    })" );
+
+    Params.ValidateAndAssignDefaults(default_parameters);
+
+    mComputeAsPostProcess = Params["compute_only_as_postprocess"].GetBool();
```
rubenzorrilla: This IMO is a bit misleading for the user. I'd put something like `"execution_step" : "ExecuteBeforeOutputStep"` by default. Then you check that the provided value is either `ExecuteBeforeOutputStep` or `ExecuteFinalizeSolutionStep`. Otherwise an error is thrown. Based on this you set the bool member var.

applications/FluidDynamicsApplication/CMakeLists.txt
```diff
@@ -93,6 +93,7 @@ set( KRATOS_FLUID_DYNAMICS_APPLICATION_CORE_SOURCES
   ${CMAKE_CURRENT_SOURCE_DIR}/custom_processes/shock_capturing_physics_based_process.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/custom_processes/shock_capturing_entropy_viscosity_process.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/custom_processes/apply_compressible_navier_stokes_boundary_conditions_process.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/custom_processes/compute_pressure_coefficient_process.cpp
```
rubenzorrilla: Ideally, these should be alphabetically sorted (ideally).

applications/FluidDynamicsApplication/custom_python/add_custom_processes_to_python.cpp
```diff
@@ -42,6 +42,8 @@
 #include "custom_processes/calulate_levelset_consistent_nodal_gradient_process.h"
 #include "custom_processes/apply_compressible_navier_stokes_boundary_conditions_process.h"
 #include "custom_processes/shock_capturing_entropy_viscosity_process.h"
+#include "custom_processes/compute_pressure_coefficient_process.h"
```
rubenzorrilla: Same alphabetically sort comment.

applications/FluidDynamicsApplication/custom_processes/compute_pressure_coefficient_process.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+
+void ComputePressureCoefficientProcess::Execute()
+{
+    KRATOS_TRY
+
+    block_for_each(mrModelPart.Nodes(),
+        [&](NodeType& r_node)
+        {
+            const auto pressure = r_node.GetSolutionStepValue(PRESSURE);
```
rubenzorrilla: In here you're assuming the `PRESSURE` to be in the historical database. I'd make this optional.

applications/FluidDynamicsApplication/custom_python/fluid_dynamics_python_application.cpp
```diff
@@ -129,7 +129,7 @@ PYBIND11_MODULE(KratosFluidDynamicsApplication,m)
     KRATOS_REGISTER_IN_PYTHON_VARIABLE(m, VELOCITY_DIVERGENCE);
     KRATOS_REGISTER_IN_PYTHON_3D_VARIABLE_WITH_COMPONENTS(m, VELOCITY_ROTATIONAL);
     KRATOS_REGISTER_IN_PYTHON_VARIABLE(m,SMOOTHING_COEFFICIENT);
-    KRATOS_REGISTER_IN_PYTHON_VARIABLE(m,SPECTRAL_RADIUS_LIMIT)
+    KRATOS_REGISTER_IN_PYTHON_VARIABLE(m,SPECTRAL_RADIUS_LIMIT);
```
rubenzorrilla: I think this belong to another PR.

rubenzorrilla (commented):
Some minor comments. Otherwise OK.

applications/FluidDynamicsApplication/custom_processes/compute_pressure_coefficient_process.h
EduardGomezEscandell: Agreed

applications/FluidDynamicsApplication/custom_python/fluid_dynamics_python_application.cpp
EduardGomezEscandell: Agreed

applications/FluidDynamicsApplication/CMakeLists.txt
EduardGomezEscandell: They weren't really sorted before but they are now.

applications/FluidDynamicsApplication/custom_python/add_custom_processes_to_python.cpp
EduardGomezEscandell: Sorted

applications/FluidDynamicsApplication/custom_processes/compute_pressure_coefficient_process.h
EduardGomezEscandell: Good catch. Fixed.

applications/FluidDynamicsApplication/custom_processes/compute_pressure_coefficient_process.cpp
EduardGomezEscandell: Done.

rubenzorrilla (approved):

PR #9891, opened 2022-05-09
[Core] Remove use of FLAG_VARIABLE in variational distance process

marcnunezc:
**📝 Description**
This PR removes the use of FLAG_VARIABLE and replaces it with a flag. This will avoid defining it in the solver. 

Is there a reason to use the historical container instead of the non-historical or a flag?

**🆕 Changelog**
E.g.
- Use a flag to track the fixity in the variational distance process instead of FLAG_VARIABLE

### DIFF
--- a/kratos/processes/variational_distance_calculation_process.h
+++ b/kratos/processes/variational_distance_calculation_process.h
@@ -234,14 +234,14 @@ public:
             // Free the DISTANCE values
             rNode.Free(DISTANCE);
             // Set the fix flag to 0
-            node.Set(BLOCKED, false);
+            rNode.Set(BLOCKED, false);
 
             // Save the distances
             rNode.SetValue(DISTANCE, d);
 
             if(d == 0){
                 d = 1.0e-15;
-                node.Set(BLOCKED, true);
+                rNode.Set(BLOCKED, true);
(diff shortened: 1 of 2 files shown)


rubenzorrilla (commented):
Indeed, rather than using a nodal variable, we should use a nodal flag for the sake of performance.

In any case, I'm positive about this change.

ddiezrod:
I agree with @rubenzorrilla, if I remember correctly it was not possible to use a flag some years ago, I guess thats why we had to use this at that moment. I did a quick search and it seems this can be changed without any problem.

rubenzorrilla:
> I agree with @rubenzorrilla, if I remember correctly it was not possible to use a flag some years ago, I guess thats why we had to use this at that moment. I did a quick search and it seems this can be changed without any problem.

Yes. The problem was the synchronization of the MPI flags, which was not present when we implemented this. This is no longer a problem.

marcnunezc:
I changed the use of FLAG_VARIABLE to the BLOCKED flag since it was a fast change.

If you prefer to do this in two PRs we can revert to the first commit.

rubenzorrilla (approved):
Very good. Thanks for the upgrade.

PR #9892, opened 2022-05-09
[PfemFluid] Improvement in two-fluid analysis

AFranci:
The number of particles of the smaller material is conserved as much as possible. 3D remeshing parameters at the material interface are modified.


AlejandroCornejo:
One general comment, since the fine tuning of these parameters is always tricky and case-dependent, why not introducing the values of thesefactors  in the `ProjectParameters.json`? We could define a default value of it and easily modify it without recompiling by setting its value in the json.

AFranci:
> One general comment, since the fine tuning of these parameters is always tricky and case-dependent, why not introducing the values of thesefactors in the `ProjectParameters.json`? We could define a default value of it and easily modify it without recompiling by setting its value in the json.

I see your point, but in the remeshing there are a bunch of parameters and, furthermore, I think it's better to find a hard-coded setting that is as universal as possible than leave the user to modify them from the projectparameter.

AlejandroCornejo (approved):

PR #9894, opened 2022-05-09
[core] generic neighbour elements of elements

pablobecker:
**📝 Description**

New process that finds neighbour elements of elements. 
Note that unlike the original neigh elem of elem, this one returns an nullptr when there's no neigh element in the selected face.

**🆕 Changelog**
- Added generic_find_elements_neighbours_process
-  Added serial test


pablobecker:
missing MPI test.
I tried reading the mdpa from the cpp test to create an mpi version of the test, but i had issues with the path to the mdpa, so i discarded it

Perhaps I could create a helper function (exposed to python) only for the test, so that I can create a python test and then read correctly an mdpa in mpi

Do you have a better idea?

pablobecker:
ping @KratosMultiphysics/implementation-committee

kratos/processes/generic_find_elements_neighbours_process.h
```diff
@@ -0,0 +1,341 @@
...
+
+        //main loop
+        block_for_each(mr_model_part.Elements(), [&](Element & rElement) {
+            //creating temp face conditions
+            Geometry<Node<3> >& geom = rElement.GetGeometry();
+            const PointerVector<GeometryType> ElemFaces = geom.GenerateFaces();
+            const unsigned int nFaces = ElemFaces.size();
+            ElementPointerVector& rElemNeighbours = rElement.GetValue(NEIGHBOUR_ELEMENTS);
+            //resizing vector to correct dimension
+            rElemNeighbours.resize(nFaces);
```
ddiezrod: Isnt this done already in InitializeElementNeighbours?

ddiezrod:
> missing MPI test. I tried reading the mdpa from the cpp test to create an mpi version of the test, but i had issues with the path to the mdpa, so i discarded it
> 
> Perhaps I could create a helper function (exposed to python) only for the test, so that I can create a python test and then read correctly an mdpa in mpi
> 
> Do you have a better idea?

But does this work in MPI? From what I see in the code, I dont think so.

ddiezrod:
> > missing MPI test. I tried reading the mdpa from the cpp test to create an mpi version of the test, but i had issues with the path to the mdpa, so i discarded it
> > Perhaps I could create a helper function (exposed to python) only for the test, so that I can create a python test and then read correctly an mdpa in mpi
> > Do you have a better idea?
> 
> But does this work in MPI? From what I see in the code, I dont think so.

I've discussed this with @pablobecker , I had not seen that in the comparison you are only checking the pointer itself, not the elements, so everything should work fine.

kratos/processes/generic_find_elements_neighbours_process.h
```diff
@@ -0,0 +1,341 @@
...
+        return nullptr;
+    }
+
+    //TODO: CheckForNeighbourElems should be done with unordered_map instead
+    //unfortunately the operator == is not yet supported by the GlobalPointer, so it does not work
+    //when this feature is added, replace the previous code with this one:
+    /*
+    {
+        const int main_elem_id = rElement.Id();
+        const ElementPointerVector& rNeighCandidates_0 = rFaceGeom[0].GetValue(NEIGHBOUR_ELEMENTS);
```
AlejandroCornejo: Do we need this comments?

pablobecker: @pooyan-dadvand suggested to keep it for the future. I have no preference here

AlejandroCornejo:
besides we should separate the implementastion in .h and .cpp right?

kratos/processes/generic_find_elements_neighbours_process.h
```diff
@@ -0,0 +1,341 @@
...
+        //the element itsef will also be present in the n nodes, so must be avoided
+        const unsigned int main_elem_id = rElement.Id();
+        //we will take the pointers in the first node as candidates
+        const unsigned int nCandidates =  rFaceGeom[0].GetValue(NEIGHBOUR_ELEMENTS).size();
+        const unsigned int nCompleteList = PointersOfAllFaceNodes.size();
+        for( unsigned int j = 0 ; j < nCandidates; j++){ 
+            unsigned int repetitions = 1 ; //the current one must be added
+            if(PointersOfAllFaceNodes[j].Id()!=main_elem_id){ //ignoring main node
+                for( unsigned int k = nCandidates ; k < nCompleteList; k++){ //starting from pointers belonging to the next node
+                    if(&(*PointersOfAllFaceNodes(j))==&(*PointersOfAllFaceNodes(k))){
```
roigcarlo: This comparison in particular can be problematic in MPI. If I understand correctly, you are comparing the physical address of the Element to remove duplicates. If you access by `&*` I assume because you want the address of the element to avoid having duplicated elements from different processes, you are calling the overloaded `*` from the global_pointer, which will segfault. use:

```C++
if( PointersOfAllFaceNodes(j).get() == PointersOfAllFaceNodes(k).get() ){
```

This will return the address without accessing the data.

Another solution will be to get the vector with the id's of the pointers via a communicator (like its done in the nodal_neighbor process)

miguelmaso:
> Note that unlike the original neigh elem of elem, this one returns an nullptr when there's no neigh element in the selected face.

IMG_20220512_172349

neighbor elements of 3 with the old process: `[2, 4]`
neighbor elements of 3 with the new process: `[2, nullptr, 4]`
being the first face shared by elems 3-2 and the third face, shared by 3-4

Maybe it is interesting to pass a flag or a bool to the old process in order to select the desired behavior (e.g. `sort_neighbours_by_faces`)

kratos/processes/generic_find_elements_neighbours_process.h
pablobecker: didnt know, thank you!

pablobecker:
> neighbor elements of 3 with the old process: `[2, 4]`

no, the previous one returned a pointer to the element itself in case of no neigh in the face:
https://github.com/KratosMultiphysics/Kratos/blob/master/kratos/processes/find_elements_neighbours_process.h#L362
so the new and old processes return the neigh elems in the same positions 
 in this case it'd be [2,3,4]

RiccardoRossi:
why returing a ppointer to the element itself is a problem??

pablobecker:
> why returing a ppointer to the element itself is a problem??

IMO it's misleading, the n position of the vector should have the pointer to the element in the n face. returning a pointer to the elem itself forces you to access the element to check if it was a real neighbour or itself

RiccardoRossi:
but then we shall we remove it for good (making the corresponding changes through the code) rather than creating a new object which provides the same capabilities except this detail...

miguelmaso:
> 

> > why returing a ppointer to the element itself is a problem??
> 
> IMO it's misleading, the n position of the vector should have the pointer to the element in the n face. returning a pointer to the elem itself forces you to access the element to check if it was a real neighbour or itself

I agree, it's very misleading. If I had known this, I'd have modified some implementations

> but then we shall we remove it for good (making the corresponding changes through the code) rather than creating a new object which provides the same capabilities except this detail...

and add some doxygen explaining it

pablobecker:
> but then we shall we remove it for good (making the corresponding changes through the code) rather than creating a new object which provides the same capabilities except this detail...

the current one only works with tets, while this one works for all geometries (as long as they have implemented geom.GenerateFaces() )

i can delete the old one, but to keep this PR smaller, i'd like to first merge this one and then in a second PR remove the old one and replace it with this new one, with the necessary changes in all the apps that are using it

RiccardoRossi:
will the numbering remain consistent in the cases in which tets are used?

i am not too positive of the two step approach since it will open the way to no doing the change at all...

pablobecker:
> 


> will the numbering remain consistent in the cases in which tets are used?

yes, they are the same. both the current process and the GenerateFaces() use the same face order

rubenzorrilla:
> will the numbering remain consistent in the cases in which tets are used?
> 
> i am not too positive of the two step approach since it will open the way to no doing the change at all...

I agree about the two steps comment. Note that the addition in this PR has no interaction with the rest of code so there is no point at all in doing it in two steps. Besides this, during some time two utilities aiming at doing the same operation with different behaviour will coexist, something that is in my opinion misleading.

kratos/processes/generic_find_elements_neighbours_process.h
pablobecker: merged everythin into a single elemental loop

pablobecker:
all fixed! could you please review it now?

pablobecker:
sorry @KratosMultiphysics/dem i mistakenly added you to the reviewrs

applications/ShallowWaterApplication/custom_elements/conservative_element.cpp
```diff
@@ -145,11 +145,17 @@ void ConservativeElement<TNumNodes>::CalculateArtificialViscosity(
 {
     double jump = .0;
     array_1d<double,2> inner_grad_h = prod(rData.nodal_h + rData.nodal_z, rDN_DX);
-    for (auto& r_elem : this->GetValue(NEIGHBOUR_ELEMENTS)) {
```
miguelmaso, line 148: Please, wait a little before merging this branch, let me check if the behavior is unmodified. I will run some validation tests

pablobecker: ping @miguelmaso

miguelmaso: sorry for the delay. I've just checked the results of the tests and it is OK

miguelmaso (approved):
from my side, it is OK

kratos/processes/generic_find_elements_neighbours_process.h
```diff
@@ -0,0 +1,228 @@
...
+
+///@name Kratos Globals
+///@{
+
+///@}
+///@name Type Definitions
+///@{
+typedef  ModelPart::NodesContainerType NodesContainerType;
+typedef  ModelPart::ElementsContainerType ElementsContainerType;
+typedef Geometry<Node < 3 > > GeometryType;
```
rubenzorrilla, line 36: These types must be defined inside the class. Note that you're defining them for the entire Kratos in here!

kratos/processes/generic_find_elements_neighbours_process.h
```diff
@@ -0,0 +1,228 @@
...
+///@name  Functions
+///@{
+
+///@}
+///@name Kratos Classes
+///@{
+
+/// Short class definition.
+/** Detail class definition.
+*/
```
rubenzorrilla, line 53: Documentation is missing. I think it is mandatory for such kind of processes.

kratos/processes/generic_find_elements_neighbours_process.h
```diff
@@ -0,0 +1,228 @@
...
+protected:
+    ///@name Protected static Member Variables
+    ///@{
+
+
+    ///@}
+    ///@name Protected member Variables
+    ///@{
+
+    ModelPart& mr_model_part;
```
rubenzorrilla, line 139: `mrModelPart` according to style guide.

kratos/processes/generic_find_elements_neighbours_process.h
```diff
@@ -0,0 +1,228 @@
...
+    ///@name Protected Operators
+    ///@{
+
+
+    ///@}
+    ///@name Protected Operations
+    ///@{
+
+
+    GlobalPointer<Element> CheckForNeighbourElems (const Geometry<Node<3> >& rBoundaryGeom,
```
rubenzorrilla, line 151: Why protected if there is no derived class?

kratos/processes/generic_find_elements_neighbours_process.h
```diff
@@ -0,0 +1,228 @@
...
+    ///@name Private Inquiry
+    ///@{
+
+
+    ///@}
+    ///@name Un accessible methods
+    ///@{
+
+    /// Assignment operator.
+    GenericFindElementalNeighboursProcess& operator=(GenericFindElementalNeighboursProcess const& rOther);
```
rubenzorrilla, line 192: This must be in a `private` section. Note that it is now protected as the `private` keyword is missing.

kratos/processes/generic_find_elements_neighbours_process.h
pablobecker: ok, fixing it

kratos/processes/generic_find_elements_neighbours_process.h
pablobecker: ok, adding it

kratos/processes/generic_find_elements_neighbours_process.h
pablobecker: ok

kratos/processes/generic_find_elements_neighbours_process.h
pablobecker: ok

kratos/processes/generic_find_elements_neighbours_process.h
pablobecker: ok

kratos/processes/generic_find_elements_neighbours_process.cpp
```diff
@@ -0,0 +1,151 @@
...
+    KRATOS_TRY
+
+    //finding elemental neighbours of nodes
+    FindGlobalNodalElementalNeighboursProcess nodal_neigh_proc(mr_model_part);
+    nodal_neigh_proc.Execute();
+
+    //main loop
+    block_for_each(mr_model_part.Elements(), [&](Element & rElement) {
+        //creating temp face conditions
+        Geometry<Node<3> >& geom = rElement.GetGeometry();
```
rubenzorrilla, line 34: `auto` type is advisable to be used in here. Besides, variable should be named as `r_geom` according to our style guide.

kratos/processes/generic_find_elements_neighbours_process.cpp
```diff
@@ -0,0 +1,151 @@
...
+        const PointerVector<GeometryType> ElemBoundaries = geom.LocalSpaceDimension()==3 ? geom.GenerateFaces() : geom.GenerateEdges() ;
```
rubenzorrilla, line 35: `const auto elem_boundaries = ....`

pablobecker: ok

kratos/processes/generic_find_elements_neighbours_process.cpp
pablobecker: ok

kratos/processes/generic_find_elements_neighbours_process.cpp
```diff
@@ -0,0 +1,151 @@
...
+        const unsigned int nBoundaries = ElemBoundaries.size();
+        
+        //initializing elem neighbours
+        if (rElement.Has(NEIGHBOUR_ELEMENTS)) {
+            ElementPointerVector& r_neighbour_elements = rElement.GetValue(NEIGHBOUR_ELEMENTS);
+            r_neighbour_elements.reserve(nBoundaries); 
+            r_neighbour_elements.erase(r_neighbour_elements.begin(),r_neighbour_elements.end() );
```
rubenzorrilla, line 42: Considering that you're calling the `resize` below, I think that this can be substituted by a call to `clear`.

kratos/processes/generic_find_elements_neighbours_process.cpp
```diff
@@ -0,0 +1,151 @@
...
+        } else {
+            ElementPointerVector empty_vector;
+            empty_vector.reserve(nBoundaries); 
+            rElement.SetValue(NEIGHBOUR_ELEMENTS, empty_vector);
+        }
+        ElementPointerVector& rElemNeighbours = rElement.GetValue(NEIGHBOUR_ELEMENTS);
+        //resizing vector to correct dimension
+        rElemNeighbours.resize(nBoundaries);
```
rubenzorrilla, line 50: Wouldn't be sufficient to do the `reserve` in here?

kratos/processes/generic_find_elements_neighbours_process.cpp
```diff
@@ -0,0 +1,151 @@
...
+        for(unsigned int i_boundary=0; i_boundary<nBoundaries; i_boundary++){
+            rElemNeighbours(i_boundary) = CheckForNeighbourElems(ElemBoundaries[i_boundary], rElement);
+        }
+    });
+
+    KRATOS_CATCH("Error finding the elemental neighbours")
+}
+
+
+GlobalPointer<Element> GenericFindElementalNeighboursProcess::CheckForNeighbourElems (const Geometry<Node<3> >& rBoundaryGeom,
```
rubenzorrilla, line 62: Formatting issues.

kratos/processes/generic_find_elements_neighbours_process.cpp
```diff
@@ -0,0 +1,151 @@
...
+                                                 Element & rElement)
+{
+    //creating a a vector of all the elem pointers
+    //therefore the elem of interest will be repeated as many times as nodes in the faces are
+    ElementPointerVector PointersOfAllBoundaryNodes;
+    const unsigned int nNodes = rBoundaryGeom.size();
+    for( unsigned int node_i = 0 ; node_i < nNodes; node_i++){ //starting from node 1
+        const ElementPointerVector& rNeighCandidates_i = rBoundaryGeom[node_i].GetValue(NEIGHBOUR_ELEMENTS);
```
rubenzorrilla, line 70: `auto` type is preferred in here.

kratos/processes/generic_find_elements_neighbours_process.cpp
```diff
@@ -0,0 +1,151 @@
...
+        //for(const auto iter = CandidatesRepetitions.begin(); iter != CandidatesRepetitions.end(); ++iter){
+        for(const auto iter : CandidatesRepetitions) {
+            if(iter.second==rFaceGeom.size()){
+                return iter.first;
+            }
+        }
+
+        //if not found, return pointer to element itself
+        return nullptr;
+    } */
```
rubenzorrilla, line 134: Please remove commented code.

kratos/processes/generic_find_elements_neighbours_process.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+
+std::vector<bool> GenericFindElementalNeighboursProcess::HasNeighboursInFaces(const Element& rElement)
+{
+    std::vector<bool> has_neigh_in_faces;    
+    if (rElement.Has(NEIGHBOUR_ELEMENTS)) {
+        const ElementPointerVector& r_neighbour_elements = rElement.GetValue(NEIGHBOUR_ELEMENTS);
```
rubenzorrilla, line 141: `auto`

rubenzorrilla (commented):
There are still some minors. Sorry, but IMO we should be a bit picky with key utilities like this one. 
Besides, there are several style guide and formatting issues.

Thanks for the patience.

kratos/processes/generic_find_elements_neighbours_process.cpp
pablobecker: ok

kratos/processes/generic_find_elements_neighbours_process.cpp
pablobecker: no, i am not doing push_back later and size must always be nBoundaries

kratos/processes/generic_find_elements_neighbours_process.cpp
maceligueta: @rubenzorrilla  what is the rule of thumb on using `auto` or standard declaration?

pablobecker:
thanks for the review. Unfortunately i didnt expect the TC to jump in and merged after i got the approval. Apologies, i was not aware of the protocol for PRs like these.
I created anew  branch with the fixed and will push it once it's compiled locally and tests pass

kratos/processes/generic_find_elements_neighbours_process.cpp
pablobecker: ok

kratos/processes/generic_find_elements_neighbours_process.cpp
pablobecker: ok

kratos/processes/generic_find_elements_neighbours_process.cpp
rubenzorrilla: At the end there is no rule at all, but it is advisable for complex types, like the one in here, in which the `auto` becomes very handy. Imagine the situation in which we finally get rid of the template argument of the `Node`, as we did in the  past with the `Point`, we'll need to  go one by one changing the declarations. If `auto` is used, this becomes automatic.

On the contrary, we tend to avoid it when using simple types like `double`, `int` and similars. Besides, I've experienced some weird missbehaviors when doing algebraic operations with variables, so I also tend to avoid it with `array_1d<double,3>`-like types.

As a rule of thumb, I'd say this.

rubenzorrilla:
> thanks for the review. Unfortunately i didnt expect the TC to jump in and merged after i got the approval. Apologies, i was not aware of the protocol for PRs like these. I created anew branch following your review and will push it once it's compiled locally and tests pass

OK. Also note that `test_generic_neighbour_elements.cpp` are failing in the nightly build. I think this is potentially related to this.

rubenzorrilla:
I'm also getting a linking error when compiling in Windows...

rubenzorrilla:
> I'm also getting a linking error when compiling in Windows...

Strange, it didn't happen anymore...

PR #9897, opened 2022-05-10
[Core] ProcessInfoOutputProcess and FinalizeCouplingStep

frawahl:
**📝 Description**
An output process for writing variables of `ProcessInfo` to a file.
For example, this can be used to output the `NL_ITERATION_NUMBER` when using the `ResidualBasedNewtonRaphsonStrategy`.

Additionally,  the methods `FinalizeCouplingStep` and `ExecuteFinalizeCouplingStep` were added to `analysis_stage.py` and `process.h` respectively. This allows to output variables of `ProcessInfo`, like the non-linear iteration number of a solver, which is part of a CoSimulation, not only for each time step, but for each coupling iteration and therefore each `SolveSolutionStep` of the solver.

**🆕 Changelog**
- Added `ExecuteFinalizeCouplingStep` to `process.h`  and `add_process_to_python.cpp`
- Added `FinalizeCouplingStep` to `analysis_stage.py`, which calls `ExecuteFinalizeCouplingStep` of all processes
- Added call to `FinalizeCouplingStep` of `analysis_stage.py` to `kratos_base_wrapper.py `of `CoSimulationApplication`
- Added `ProcessInfoOutputProcess` derived from `OutputProcess` to kratos/python_scripts/
- Added tests for `ProcessInfoOutputProcess` and corresponding reference files to `test_processes.py`
- Adapted/ extended `SolutionLoopPointOutputProcesses` of `test_processes.py` for tests of an actual `OutputProcess` and to using `ExecuteFinalizeCouplingStep`


rubenzorrilla (changes requested):
This is changing the public API of the `Process` base class. Must be discussed by the @KratosMultiphysics/technical-committee.

PR #9898, opened 2022-05-11
[Core] Coupling step methods for Process

frawahl:
**📝 Description**
This PR extends `process.h` by the methods  `ExecuteInitializeCouplingStep` and `ExecuteFinalizeCouplingStep`. Correspondingly, the methods `InitializeCouplingStep` and `FinalizeCouplingStep` were added to `analysis_stage.py`.

This way,  `InitializeCouplingStep` and `FinalizeCouplingStep` can be called by a coupled simulation for a kratos solver, which would allow operations of a `Process` of that solver to happen not only for each time step, but for each coupling iteration and therefore each `SolveSolutionStep` of a solver.
- for example for debugging and post-processing: output the `NL_ITERATION_NUMBER` or displacement of a point in the `ModelPart` of that solver over the number of coupling iterations

**🆕 Changelog**
- Added `ExecuteInitializeCouplingStep` to `process.h`  and `add_process_to_python.cpp`
- Added `ExecuteFinalizeCouplingStep` to `process.h`  and `add_process_to_python.cpp`
- Added `InitializeCouplingStep` to `analysis_stage.py`, which calls `ExecuteInitializeCouplingStep` of all processes
- Added `FinalizeCouplingStep` to `analysis_stage.py`, which calls `ExecuteFinalizeCouplingStep` of all processes


roigcarlo:
@KratosMultiphysics/technical-committee agrees with ruben suggestion of making a `coupling_process.h` derived from `process.h` with the new methods

frawahl:
> @KratosMultiphysics/technical-committee agrees with ruben suggestion of making a `coupling_process.h` derived from `process.h` with the new methods

Alright, I have two questions/ confirmations:

1. There will be a new list_of_coupling_processes then in analysis_stage and the parameters json?
2. A process which needs coupling step methods as well as output process methods, like #9900 , will it get two parent classes? Is it sorted into all three lists in analysis_stage?

rubenzorrilla:
> > @KratosMultiphysics/technical-committee agrees with ruben suggestion of making a `coupling_process.h` derived from `process.h` with the new methods
> 
> Alright, I have two questions/ confirmations:
> 
>     1. There will be a new list_of_coupling_processes then in analysis_stage and the parameters json?
> 

About this I note that you can have as lists as you want in the parameters json. The thing is that these new methods will not be called by the standard analysis stage. At least, this is not considered in current discussion.

>     2. A process which needs coupling step methods as well as output process methods, like [[Core] Process info output process #9900](https://github.com/KratosMultiphysics/Kratos/pull/9900) , will it get two parent classes? Is it sorted into all three lists in analysis_stage?

We didn't consider this... Not sure if there are many cases that might fall into such casuistry... Can you develop the use case in #9900?

frawahl:
> > > @KratosMultiphysics/technical-committee agrees with ruben suggestion of making a `coupling_process.h` derived from `process.h` with the new methods
> > 
> > 
> > Alright, I have two questions/ confirmations:
> > ```
> > 1. There will be a new list_of_coupling_processes then in analysis_stage and the parameters json?
> > ```
> 
> About this I note that you can have as lists as you want in the parameters json. The thing is that these new methods will not be called by the standard analysis stage. At least, this is not considered in current discussion.
> 
> > ```
> > 2. A process which needs coupling step methods as well as output process methods, like [[Core] Process info output process #9900](https://github.com/KratosMultiphysics/Kratos/pull/9900) , will it get two parent classes? Is it sorted into all three lists in analysis_stage?
> > ```
> 
> We didn't consider this... Not sure if there are many cases that might fall into such casuistry... Can you develop the use case in #9900?

We used it to look at the non-linear iteration number of the fluid and the structural solver and at the displacement of the tip of the structure over the course of the coupling iterations .. to get an impression of the convergence behavior for the coupling. You could print out any ProcessInfo at any point of a simulation, including coupling steps, no need to extract information from the terminal output. Interesting for debugging.

pooyan-dadvand:
@KratosMultiphysics/technical-committee we think it would be better to do a study in the present coupled solvers and see the similarities and if these methods fit them.

Please note that usually we do the coupling in the solver side where we don't have processes. (In the cosim the "solver" is an strategy so you have processes but not in the rest of the Kratos)

rubenzorrilla:
@sunethwarna @frawahl any advance in this regard?

RiccardoRossi:
current state of discussion within the  @KratosMultiphysics/technical-committee 

this is not a decision,  it is just a brainstorming, pending discussion also with @KratosMultiphysics/implementation-committee 

~~~py

CouplingStage : {
    stage_settings : {
        "InitializeCouplingStep" : [...list of operations ...],
        FluidStage : {
        }
        StructuralStage : {
        }
    }
}



RunSolutionLoop
    for t in times:
        InitializeSolutionStep
        while not converged
            InitializeCouplingStep
            FluidStage.RunSolutionStep()
            AfterFirstStage
            StructuralStage.RunSolutionStep()
            AfterSecondStage
        FluidStage.FinalizeSolutionStep()
        StructuralStage.FinalizeSolutionStep()


def InitializeCouplingStep(self):
    for ops in initialize_coupling_step_list:
        ops.Execute()
~~~

AlejandroCornejo:
sounds great indeed. My only piece of advise, coming from my previous thermomechanical solver experience, is that we must be really careful about the ordering in the predict of both solvers.

What happened to me was, If I compute the initial residual of the solid part accordingto the old temperature field, you cannot converge since the residual was badly computed.

This is why I had to to the predict and solve of the thermal before predicting and solving the mechanical one.

matekelemen:
I'm also interested in the topic because I ran into a fair amount of trouble with deciphering the order of solution steps. Here's what I think:

The number of solution step types is becoming unmanageable. If you think about a coupled analysis, you may run into:
- coupling solution steps that handle the solution of the entire coupled system, which involves calling `SolveSolutionStep` of the partitions and performing coupling-related tasks
- domain-specific solution steps that solve a partition within the coupled system (the original purpose of `SolveSolutionStep`)
- nonlinear solution steps
- ... (others we don't know about yet)

these all have a weird relationship with advancing in time as well: some require a new time step, while others rely on `TIME` persisting across solution steps.

In my opinion, we should only have one set of `InitializeSolutionStep`-`SolveSolutionStep`-`FinalizeSolutionStep` and the derived analysis/solver should decide what that means in its own context.

In the most complex case, an analysis might have to decide whether to advance in time and what kind of initialization/solution/finalization to do:
```py
def SolveSolutionStep(self):
    while not self.ShouldTerminate():
        if self.ShouldAdvanceInTime(): # <== decide whether to advance in time
            self.AdvanceInTime() 
        self.InitializeSolutionStep() # <== decides internally what initialization to do: standard, coupling, nonlinear, etc.
        self.SolveSolutionStep() # <== decides internally what kind of solution to do
        self.FinalizeSolutionStep() # <== decides internally what kind of finalization to do 
```

In my opinion, adding a dedicated new function for a specific type of solution step and hard-coding its position in the solution sequence limits modularity and will cause more headaches in the future.

pooyan-dadvand:
@matekelemen, I am not sure if I understood correctly your comment. Here are some comments:
- The idea here is to avoid having many coupled stages for coupling other stages when all their differences is calling one process in the middle of the two solver.SolveSolutionStep() calls. (Usually in transferring data and resetting things)
- On the other hand for coupled solvers we have the same problem as we cannot "inject" any process to them and we are having more and more coupled solvers every day
- I agree that maybe the template pattern is not well suited here but I don't think having 1 coupled stage for each coupled problem is the best option
- I don't see how your suggestion of overriding in the `InitializeSolutionStep`-`SolveSolutionStep`-`FinalizeSolutionStep` can handle correctly the coupled use cases like transferring data or resetting a coupling iteration. Would you please elaborate a bit more?
- Finally, I think we should find a way to unify the Cosim stages into the ones of the native Kratos. Something that with the current stages of Kratos I am not sure if is possible

frawahl:
Again, I am not sure that I understood all of you correctly, but I would like to clarify some things:

- Yes, this PR would add two methods to `analysis_stage.py `which probably only people profit from who perform a coupled simulation using the `CoSimulationApplication` . 

- This PR (together with the change in `kratos_base_wrapper.py `of PR #9900) would allow ALL kratos solvers which are being coupled by `CoSimulationApplication`  to have the processes which are "attached" to them do something before and after their respective `SolveSolutionStep` WITHOUT any further change in the structure of the code or inside stages or inside any coupled solver. 

Explanation: All of the kratos solvers which are coupled by `CoSimulation` have their `SolveSolutionStep` being called by their wrapper, which is called by a "coupled_solver" in `CoSimulationApplication`. That wrapper inherits from `kratos_base_wrapper` and would tell the analysis stage of the kratos solver to call the `ExecuteInitializeCouplingStep` and `ExecuteFinalizeCouplingStep` of the solver's processes.

- As far as I can see it is quite a small change, which allows every kratos solver (fluid, structure, dem, convDiff, particel, ..) to have output or a process do something else in between their coupling iterations, so whenever their `SolveSolutionStep` is called, no matter whether time advances or what else is happening outside the respective solver. Maybe @sunethwarna can elaborate more about the advantages of being able to do that.

PR #9901, opened 2022-05-13
[BDF Scheme] Reducing verbosity: Print warning if it's relevant

miguelmaso:
**📝 Description**
For orders higher than 2, the warning was thrown at every step, even if the time step is constant.
Now, the warning is thrown if the order is higher than 2 and the time step is variable.

**🆕 Changelog**
- Reduce the verbosity


rubenzorrilla (commented):
Change makes sense to me, but shouldn't it be an error?

miguelmaso:
> shouldn't it be an error?

makes sense since the scheme will be unstable

rubenzorrilla (approved):

PR #9906, opened 2022-05-13
cpp suite

miguelmaso:
**📝 Description**
wip, creating cpp suite


matekelemen:
Any updates here? I think it's a good idea to have a unified interface for cpp and python tests.

pooyan-dadvand:
@KratosMultiphysics/technical-committee would put this one on standby until the test environment to be used is changed. Most probably we can find such a feature in the new library to be used.

PR #9908, opened 2022-05-16
[Core] Convert point output process

### LINKED ISSUE #7877: [Core] PointOutputProcess is not an output process (no IsOutputStep nor PrintOutput)

frawahl:
**📝 Description**
This PR converts `PointOutputProcess` to be an actual `KratosMultiphysics.OutputProcess` and adapts affiliated output processes as well as their tests.

**🆕 Changelog**
- Converted `PointOutputProcess` and `MultiplePointsOutputProcess` to be derived from `OutputProcess`
- Adapted `LineOutputProcess`, `GiDOutputProcess`, `GiDDamOutputProcess`
- Adapted `test_processes.py` and its function `SolutionLoopPointOutputProcesses` to accommodate `OutputProcess`es
- Fixed #7877, 
made possible by #9529

### DIFF
--- a/kratos/python_scripts/gid_output_process.py
+++ b/kratos/python_scripts/gid_output_process.py
@@ -275,6 +275,8 @@ class GiDOutputProcess(KM.OutputProcess):
     def PrintOutput(self):
         if self.point_output_process is not None:
             self.point_output_process.ExecuteBeforeOutputStep()
+            if self.point_output_process.IsOutputStep():
+                self.point_output_process.PrintOutput()
 
         # Print the output
         time = self.__get_pretty_time(self.model_part.ProcessInfo[KM.TIME])
--- a/kratos/python_scripts/line_output_process.py
+++ b/kratos/python_scripts/line_output_process.py
@@ -5,22 +5,13 @@ import KratosMultiphysics
 from KratosMultiphysics.multiple_points_output_process import MultiplePointsOutputProcess
 
 def Factory(settings, Model):
-    if(type(settings) != KratosMultiphysics.Parameters):
+    if not isinstance(settings, KratosMultiphysics.Parameters):
         raise Exception("expected input shall be a Parameters object, encapsulating a json string")
-    return LineOutputProcess(Model, settings["Parameters"])
 
-class LineOutputProcess(KratosMultiphysics.Process):
-    """This process writes output for several points along a line to a file
-    Internally it holds an object of type "MultiplePointsOutputProcess"
-    Usage:
-        - specifying start and end point defining the line
-        - specifying the number of sampling points along the line (start and end points will be included)
-    """
-    def __init__(self, model, params):
-        KratosMultiphysics.Process.__init__(self)
+    params = settings["Parameters"]
 
-        default_settings = KratosMultiphysics.Parameters('''{
-            "help"              : "This process writes output for several points along a line to a file. Internally it holds an object of type MultiplePointsOutputProcess",
+    default_settings = KratosMultiphysics.Parameters('''{
+        "help"              : "This process writes output for several points along a line to a file. Internally it holds an object of type MultiplePointsOutputProcess",
             "model_part_name"   : "",
             "entity_type"       : "element",
             "interval"          : [0.0, 1e30],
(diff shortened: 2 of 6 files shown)


kratos/python_scripts/multiple_points_output_process.py
```diff
@@ -83,6 +83,13 @@ def ExecuteBeforeOutputStep(self):
         for proc in self.point_output_processes:
             proc.ExecuteBeforeOutputStep()
 
+    def IsOutputStep(self):
+        return self.point_output_processes[0].IsOutputStep()
```
rubenzorrilla, line 88:
```suggestion
        # We check only the first process as the interval is the same in all of them
        return self.point_output_processes[0].IsOutputStep()
```
Just an annotation for the future.

kratos/python_scripts/line_output_process.py
```diff
@@ -5,22 +5,13 @@
 from KratosMultiphysics.multiple_points_output_process import MultiplePointsOutputProcess
 
 def Factory(settings, Model):
-    if(type(settings) != KratosMultiphysics.Parameters):
+    if not isinstance(settings, KratosMultiphysics.Parameters):
         raise Exception("expected input shall be a Parameters object, encapsulating a json string")
-    return LineOutputProcess(Model, settings["Parameters"])
-
-class LineOutputProcess(KratosMultiphysics.Process):
```
rubenzorrilla, line 12: Why is the class removed?

rubenzorrilla: My proposal would be to derive this `LineOutputProcess` from the  `MultiplePointsOutputProcess` so you can get rid of all the calls below (as you already did).

Then we keep the `Factory` function as it was before and we only implement the constructor as well as the `GetDefaultParameters`. Once the input parameters have been validated, we tweak them so the call of the base process constructor doesn't crash.

rubenzorrilla (commented):
I'm very positive about these changes. However, there are a few comments with regard to the `LineOutputProcess`.

rubenzorrilla:
Just mentioning #9910 as it might be relevant in here.

kratos/python_scripts/line_output_process.py
frawahl: Sure, I can derive it from `MultiplePointsOutputProcess`.
I'm not sure about the advantages/ disadvantages ...
I changed the `LineOutputProcess` to what @philbucher suggested for the `CSVPointsOutputProcess`, which actually is the reason I didn't have to change anything now for the `CSVPointsOutputProcess`.

rubenzorrilla: IMO it is more object oriented solution as Kratos is, thus making possible to create an instance of the object by directly calling the constructor of the class. On top of this, setting `LineOutputProcess` in the json settings but getting a `MultiplePointsOutputProcess` instance in your process list might be misleading...

rubenzorrilla: OFC this is a personal opinion. It is clear that current solution works.

frawahl: Should I change CSVPointsOutputProcess then as well? Doesn't it make sense to be consistent?

kratos/python_scripts/gid_output_process.py
```diff
@@ -275,6 +275,8 @@ def IsOutputStep(self):
     def PrintOutput(self):
         if self.point_output_process is not None:
             self.point_output_process.ExecuteBeforeOutputStep()
+            if self.point_output_process.IsOutputStep():
```
matekelemen, line 278: As far as I know, the idea behind```OutputProcess::PrintOutput``` is to write the output without checking ```OutputProcess::IsOutputStep```.

```PythonSolver``` already [performs the check](https://github.com/KratosMultiphysics/Kratos/blob/c1fb8e578ba79317b88f3c519cb5e055780f559d/kratos/python_scripts/analysis_stage.py#L157) directly in the solution loop once, and then calls ```PrintOutput``` if successful. I don't think it's good practice, but @sunethwarna tells me that some people rely on this specific behaviour.

frawahl: Yes, you are right, that was a mistake.

Same goes for gid_dam_output_process of course.

kratos/python_scripts/line_output_process.py
rubenzorrilla: I'd leave it as it is, at least in this PR.

mpentek:
Great to see progress on this!

kratos/python_scripts/gid_output_process.py
frawahl: I just noticed, there was reasoning behind this .. 
you are right that the `PrintOutput` of `GiDOutputProcess` will only be called if `IsOutputStep` returns true. But analysis_stage won't know about `PointOutputProcess`es inside `GiDOutputProcess` and their output interval will either be checked inside the `PrintOutput` of `GiDOutputProcess` or not at all and the interval could be different, right? (I don't know what these point outputs inside the GiD output are used for)

rubenzorrilla (approved):

PR #9909, opened 2022-05-17
[ShapeOpt] Iteration restart

sunethwarna:
**📝 Description**
This adds restart capability for each iteration primal and adjoint solutions seperately. It does not change the existing behaviour. These restart files are not written by default.

The cost of reading and writing files is negligible in the fluid optimization problems compared to structural optimization problems. Therefore this is made optional.

**🆕 Changelog**
- Add iteration restart capability to shape opt.

### DIFF
--- a/applications/ShapeOptimizationApplication/python_scripts/model_part_controller_factory.py
+++ b/applications/ShapeOptimizationApplication/python_scripts/model_part_controller_factory.py
@@ -47,7 +47,8 @@ class ModelPartController:
             },
             "mesh_motion" : {
                 "apply_mesh_solver" : false
-            }
+            },
+            "write_iteration_restart_files": false            
         }""")
 
         self.model_settings.ValidateAndAssignDefaults(default_settings)
@@ -76,6 +77,7 @@ class ModelPartController:
         self.design_surface = None
         self.damping_utility = None
         self.direction_dampings = []
+        self.is_iteration_restart_files_written = self.model_settings["write_iteration_restart_files"].GetBool()
 
     # --------------------------------------------------------------------------
     def Initialize(self):
@@ -149,6 +151,10 @@ class ModelPartController:
     # --------------------------------------------------------------------------
     def GetDesignSurface(self):
         return self.design_surface
+
+    # --------------------------------------------------------------------------
+    def IsIterationRestartFilesWritten(self):
+        return self.is_iteration_restart_files_written        
 
     # --------------------------------------------------------------------------
     def DampNodalSensitivityVariableIfSpecified(self, variable):
--- a/applications/ShapeOptimizationApplication/python_scripts/response_data_io.py
+++ b/applications/ShapeOptimizationApplication/python_scripts/response_data_io.py
@@ -1,0 +1,77 @@
+import json
+from pathlib import Path
+
+import KratosMultiphysics as Kratos
+from KratosMultiphysics.response_functions.response_function_interface import ResponseFunctionInterface
+
+def GetObjectiveValue(response: ResponseFunctionInterface, identifier: str, optimization_iteration: int, optimization_results_absolute_path: str, is_iteration_restart_files_written: bool) -> float:
+    if is_iteration_restart_files_written:
+        iteration_data_file_path = Path(optimization_results_absolute_path) / Path("Raw_Data/iteration_{:05d}.json".format(optimization_iteration))
+
+        if not iteration_data_file_path.is_file():
+            iteration_data_file_path.parent.mkdir(parents=True, exist_ok=True)
+            # write dummy dict to create the file
+            with open(str(iteration_data_file_path), "w") as file_output:
+                file_output.write(json.dumps({}, indent=4, sort_keys=True))
+
+        with open(str(iteration_data_file_path), "r") as file_input:
+            data_dict = json.loads(file_input.read())
+
+        if identifier not in data_dict.keys():
+            data_dict[identifier] = {}
+
+        if "value" not in data_dict[identifier].keys():
+            response.CalculateValue()
+            response_value = response.GetValue()
+            data_dict[identifier]["value"] = response_value
+
+            # now write the objective value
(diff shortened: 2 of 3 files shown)


applications/ShapeOptimizationApplication/python_scripts/model_part_controller_factory.py
```diff
@@ -76,6 +77,7 @@ def __init__(self, model_settings, model):
         self.design_surface = None
         self.damping_utility = None
         self.direction_dampings = []
+        self.is_iteration_restart_files_written = self.model_settings["write_iteration_restart_files"].GetBool()
```
armingeiser, line 80:
```suggestion
        self.write_iteration_restart_files = self.model_settings["write_iteration_restart_files"].GetBool()
```

why not call it the same as the setting?

applications/ShapeOptimizationApplication/python_scripts/model_part_controller_factory.py
```diff
@@ -150,6 +152,10 @@ def GetModel(self):
     def GetDesignSurface(self):
         return self.design_surface
 
+    # --------------------------------------------------------------------------
+    def IsIterationRestartFilesWritten(self):
```
armingeiser, line 156: This sounds like you would ask if the files have already been written, not if you should write them.

applications/ShapeOptimizationApplication/python_scripts/response_data_io.py
```diff
@@ -0,0 +1,77 @@
...
+        iteration_data_file_path = Path(optimization_results_absolute_path) / Path("Raw_Data/iteration_{:05d}.json".format(optimization_iteration))
+
+        if not iteration_data_file_path.is_file():
+            iteration_data_file_path.parent.mkdir(parents=True, exist_ok=True)
+            # write dummy dict to create the file
+            with open(str(iteration_data_file_path), "w") as file_output:
+                file_output.write(json.dumps({}, indent=4, sort_keys=True))
+
+        with open(str(iteration_data_file_path), "r") as file_input:
+            data_dict = json.loads(file_input.read())
```
armingeiser, line 18:
```suggestion
        if iteration_data_file_path.is_file():
            with open(str(iteration_data_file_path), "r") as file_input:
                data_dict = json.loads(file_input.read())
        else:
            data_dict = {}
```

applications/ShapeOptimizationApplication/python_scripts/response_data_io.py
```diff
@@ -0,0 +1,77 @@
+import json
+from pathlib import Path
+
+import KratosMultiphysics as Kratos
+from KratosMultiphysics.response_functions.response_function_interface import ResponseFunctionInterface
+
+def GetObjectiveValue(response: ResponseFunctionInterface, identifier: str, optimization_iteration: int, optimization_results_absolute_path: str, is_iteration_restart_files_written: bool) -> float:
```
armingeiser, line 7:
```suggestion
def GetResponseValue(response: ResponseFunctionInterface, identifier: str, optimization_iteration: int, optimization_results_absolute_path: str, is_iteration_restart_files_written: bool) -> float:
```

applications/ShapeOptimizationApplication/python_scripts/response_data_io.py
```diff
@@ -0,0 +1,77 @@
...
+
+        if identifier not in data_dict.keys():
+            data_dict[identifier] = {}
+
+        if "value" not in data_dict[identifier].keys():
+            response.CalculateValue()
+            response_value = response.GetValue()
+            data_dict[identifier]["value"] = response_value
+
+            # now write the objective value
```
armingeiser, line 28:
```suggestion
            # now write the response value
```

applications/ShapeOptimizationApplication/python_scripts/response_data_io.py
```diff
@@ -0,0 +1,77 @@
...
+            with open(str(iteration_data_file_path), "w") as file_output:
```
armingeiser, line 29:
```suggestion

            if not iteration_data_file_path.is_file():
                iteration_data_file_path.parent.mkdir(parents=True, exist_ok=True)
            with open(str(iteration_data_file_path), "w") as file_output:
```

applications/ShapeOptimizationApplication/python_scripts/response_data_io.py
```diff
@@ -0,0 +1,77 @@
...
+        else:
+            response_value = data_dict[identifier]["value"]
+            Kratos.Logger.PrintInfo("ShapeOpt", "Found recorded objective value for iteration {:d} {:s} objective. Remove \"{:s}/value\" entry from {:s} to re-calculate objective value for this iteration.".format(optimization_iteration, identifier, identifier, str(iteration_data_file_path.absolute())))
+    else:
+        response.CalculateValue()
+        response_value = response.GetValue()
+
+    return response_value
+
+def GetObjectiveSensitivityValues(response: ResponseFunctionInterface, identifier: str, optimization_iteration: int, optimization_results_absolute_path: str, is_iteration_restart_files_written: bool) -> dict:
```
armingeiser, line 40:
```suggestion
def GetResponseGradient(response: ResponseFunctionInterface, identifier: str, optimization_iteration: int, optimization_results_absolute_path: str, is_iteration_restart_files_written: bool) -> dict:
```

armingeiser (changes requested):
Do i understand this correctly, that the optimization still runs from scratch, except for all iterations that are already there, you use existing values/gradients instead of recomputing them?

Discussion point: 
In my opinion overwriting existing results should be the default behaviour - it is for all other analyses. I am not sure if the name `write_iteration_restart_files` is making clear that restart files are not only written, but also used if found. 

Thats why i would propose to have two flags:
- write_restart_files -> just triggers writing of the restart files
- use_restart_files -> triggers that you actually also read them 
Otherwise you have to manually delete the iteration directories if you want to "overwrite" the already existing results.

PR #9915, opened 2022-05-19
[Several] Fix -Werror=maybe-uninitialized

loumalouomega:
**📝 Description**
Fix -Werror=maybe-uninitialized with gcc 12

**🆕 Changelog**
- Fix -Werror=maybe-uninitialized with gcc 12


rubenzorrilla (approved):
Very good. Thanks @loumalouomega

PR #9920, opened 2022-05-24
[Core] Fixed explicit startegy minor issues

EduardGomezEscandell:
**📝 Description**
This PR fixes a couple of minor issues with Explicit strategies. It breaks API in a minor way, but Explicit startegies are relatively recent so the harm should be very limited.

**🆕 Changelog**
- pGetExplicitBuilder now returns a shared pointer rather than a reference to a shared pointer. This is for memory safety, since using a reference does not increase the refcount. This is the API breakage.
- Created GetExplicitBuilder, which returns a reference to the object directy. Also a const version.
- Replaced instances of pGetExplicitBuilder with GetExplicitBuilder where the former is not necessary.

### DIFF
--- a/kratos/solving_strategies/strategies/explicit_solving_strategy.h
+++ b/kratos/solving_strategies/strategies/explicit_solving_strategy.h
@@ -296,7 +296,7 @@ public:
      */
     ExplicitBuilderType& GetExplicitBuilder()
     {
-        KRATOS_DEBUG_ERROR_IF(mpExplicitBuilder == nullptr) << "Asking for builder and solver when it is empty" << std::endl;
+        KRATOS_ERROR_IF(mpExplicitBuilder == nullptr) << "Asking for builder and solver when it is empty" << std::endl;
         return *mpExplicitBuilder;
     };
 
@@ -306,7 +306,7 @@ public:
      */
     const ExplicitBuilderType& GetExplicitBuilder() const
     {
-        KRATOS_DEBUG_ERROR_IF(mpExplicitBuilder == nullptr) << "Asking for builder and solver when it is empty" << std::endl;
+        KRATOS_ERROR_IF(mpExplicitBuilder == nullptr) << "Asking for builder and solver when it is empty" << std::endl;
         return *mpExplicitBuilder;
     };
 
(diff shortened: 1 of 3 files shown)


kratos/solving_strategies/strategies/explicit_solving_strategy.h
```diff
@@ -283,25 +283,45 @@ class ExplicitSolvingStrategy : public SolvingStrategy<TSparseSpace, TDenseSpace
...
+        return *mpExplicitBuilder;
+    };
+
+    /**
+     * @brief Operations to get the explicit builder and solver
+     * @return The explicit builder and solver
+     */
+    const ExplicitBuilderType& GetExplicitBuilder() const
+    {
+        KRATOS_DEBUG_ERROR_IF(mpExplicitBuilder == nullptr) << "Asking for builder and solver when it is empty" << std::endl;
```
rubenzorrilla:
```suggestion
        KRATOS_ERROR_IF(mpExplicitBuilder == nullptr) << "Asking for builder and solver when it is empty" << std::endl;
```

rubenzorrilla: I'd always throw the error.

kratos/solving_strategies/strategies/explicit_solving_strategy_runge_kutta.h
```diff
@@ -260,10 +260,10 @@ class ExplicitSolvingStrategyRungeKutta : public ExplicitSolvingStrategy<TSparse
         KRATOS_TRY
 
         // Get the required data from the explicit builder and solver
-        const auto p_explicit_bs = BaseType::pGetExplicitBuilder();
-        auto& r_dof_set = p_explicit_bs->GetDofSet();
-        const unsigned int dof_size = p_explicit_bs->GetEquationSystemSize();
-        const auto& r_lumped_mass_vector = p_explicit_bs->GetLumpedMassMatrixVector();
+auto& r_explicit_bs = BaseType::GetExplicitBuilder();
```
rubenzorrilla: Minor indentation issue.

rubenzorrilla (approved):
I think this makes sense. Approving on behalf of the @KratosMultiphysics/technical-committee .

PR #9922, opened 2022-05-24
[Everywhere] Removed const from return type of GetDefaultParameters

EduardGomezEscandell:
Many function signatures look like:
```c++
const Parameters GetDefaultParameters() const override
```
The first `const` is superflous at best, and misleading at worse. Unfortunatelly this breaks API so you may decide wether it's worth it or not.

**🆕 Changelog**
I performed the following three find and replace:

84 changes:
```
Find:    const Parameters GetDefaultParameters() const override
Replace: Parameters GetDefaultParameters() const override
```

4 changes:
```
Find:    const Parameters GetDefaultParameters() const
Replace: Parameters GetDefaultParameters() const
```

75 changes (regex):
```
Find:    const Parameters (.*)::GetDefaultParameters\(\) const
Replace: Parameters \1::GetDefaultParameters() const
```

**Sidenote**
If we wanted to return actual inmutable parameters we have a ccouple of options.

_Alternative 1_
We can re-write the Parameters class to allow for internally holding a pointer to const json object (not worth it in my opinion).

_Alternative 2_
Returning a const reference to a static:
```c++
const Parameters& GetDefaultParameters() const override
{
    static Parameters default_parameters = { /* Whatever */}
    default_parameters.RecursivelyAddMissingParameters(BaseType::GetDefaultParameters());
    return default_parameters;
}
```
Once again, not worth the effort in my opinion. Besides, one could side-step the const because Parameters' implementation is not const-correct (e.g GetValue is marked const while allowing modification of the underlying object).

### DIFF
--- a/applications/ContactStructuralMechanicsApplication/custom_processes/alm_variables_calculation_process.h
+++ b/applications/ContactStructuralMechanicsApplication/custom_processes/alm_variables_calculation_process.h
@@ -137,7 +137,7 @@ public:
     /**
      * @brief This method provides the defaults parameters to avoid conflicts between the different constructors
      */
-    const Parameters GetDefaultParameters() const override
+    Parameters GetDefaultParameters() const override
     {
         const Parameters default_parameters = Parameters(R"(
         {
--- a/applications/ContactStructuralMechanicsApplication/custom_processes/base_contact_search_process.cpp
+++ b/applications/ContactStructuralMechanicsApplication/custom_processes/base_contact_search_process.cpp
@@ -1872,7 +1872,7 @@ bool BaseContactSearchProcess<TDim, TNumNodes, TNumNodesMaster>::IsNotInvertedSe
 /***********************************************************************************/
 
 template<SizeType TDim, SizeType TNumNodes, SizeType TNumNodesMaster>
-const Parameters BaseContactSearchProcess<TDim, TNumNodes, TNumNodesMaster>::GetDefaultParameters() const
+Parameters BaseContactSearchProcess<TDim, TNumNodes, TNumNodesMaster>::GetDefaultParameters() const
 {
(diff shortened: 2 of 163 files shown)


matekelemen:
I feel like if this issue were to be addressed, it should be done so at its root and make ```Parameters``` const-correct.

As things stand right now, we at least know the authors' intents and have a clue about when people may rely on the constness of ```Parameters```. However, this PR would completely eliminate that and make it much harder in the future to adapt to a const-correct ```Parameters```.

EduardGomezEscandell:
> I feel like if this issue were to be addressed, it should be done so at its root and make Parameters const-correct.

Making Parameters const-correct would not solve this issue, since returning `const T` is the same as returning a `T` (regardless of the type of T, except for references). So much so that clang and gcc raise a waring when compiling in -Wextra:
```
warning: 'const' type qualifier on return type has no effect [-Wignored-qualifiers]
```

However, fixing const-correctness would allow for _Alternative 2_ explained in this PR's description. 


Ultimately, I don't know how useful this alternative would be. As it is, modifying the object returned by GetDefaultParameters does not change what GetDefaultParameters returns, so one should be free to change it in my opinion.

matekelemen:
>  returning const T is the same as returning a T (regardless of the type of T, except for references)

That's not *entirely* true; functions overloaded on the constness of temporaries (not too useful) and template shenanigans that use functions' return types (can be very useful) come to mind, but I wouldn't expect Kratos developers to have such esoteric use cases.

More importantly, I meant that developers who wrote functions returning ```const Parameters``` (hopefully) had a reason why the returned object should be immutable, and may have relied on that property elsewhere in their code. We'll have a hard time tracking related problems if we remove every trace of that right now.

PR #9925, opened 2022-05-25
[Fluid] ComputeDragProcess now flushes its output every step

EduardGomezEscandell:
**📝 Description**
This PR solves the problem pointed out in the Fluid Dynamics project: 

https://github.com/KratosMultiphysics/Kratos/projects/10#card-37224441.
> Drag process only prints at the end of the simulation. This is a problem if the simulation crashes or if it is intentionally stopped by the user.
Open and close the file to write the drag each time it is computed.

Instead of opening and closing every time, I added a flush.

**🆕 Changelog**
- Added flush after write-to-file 
- Added unit test to validate it. I tested commenting the flushes out and it didn't pass, however I'm not certain on how deterministic this test is (as in: it will always pass when there is no problem, but may sometimes pass even if there is a problem).

### DIFF
--- a/applications/FluidDynamicsApplication/python_scripts/compute_drag_process.py
+++ b/applications/FluidDynamicsApplication/python_scripts/compute_drag_process.py
@@ -82,6 +82,7 @@ class ComputeDragProcess(KratosMultiphysics.Process):
                 file_header = self._GetFileHeader()
                 self.output_file = TimeBasedAsciiFileWriterUtility(self.model_part,
                     file_handler_params, file_header).file
+                self.output_file.flush()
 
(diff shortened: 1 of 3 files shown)


rubenzorrilla (approved):

philbucher:
I recommend against forced flush as it can have a performance impact on clusters, aka the initial design was done deliberately 

you could have already set the buffer size to achieve your desired behavior

rubenzorrilla:
> I recommend against forced flush as it can have a performance impact on clusters, aka the initial design was done deliberately
> 
> you could have already set the buffer size to achieve your desired behavior

Thanks for the hint. I'll try to figure it out in that way.

Issue #9916, opened 2022-05-20
Adoption of C++17


roigcarlo:
Dear @KratosMultiphysics/all 

As some of you may have know some time ago we starting discussion the possibility of moving to C++17. After some preparation, we believe we finally are in a comfortable position to make the change and we tentatively want to make it official by **July 1**

We will create a branch with the proper changes in the configuration files, etc so we can se beforehand if there is any incompatibility. Note that, even if we did not support it officially, we have been running with C++17 standard in selected configurations of the CI builds for quite some time now, so we expect (hopefully) no major issues. 

Note that C++17 should be compatible with the already supported versions of Visual Studio (2017+), GCC 8+, Clang 7+ and 19.0.1+

If you use older versions of the mentioned compilers we kindly ask you to update before July 1st so everything keeps working.

RiccardoRossi:
note also that at least gcc 8 will be needed

loumalouomega:
https://en.cppreference.com/w/cpp/compiler_support/17

Clang is not fully supported until 7. (for example, filesystem is supported in LLVM 7)

roigcarlo:
ok updated, I missed the library part :P

matekelemen:
A recent update got ```GCC 12.1.0``` on my system, and I'm getting a lot of warnings about ```std::unary_function``` and ```std::binary_function``` that were deprecated in ```C++11```. Here's where they're used:
- ```kratos/utilities/body_distance_calculation_utils.h```
- ```kratos/containers/vector_map.h```
- ```kratos/containers/geometry_container.h```
- ```kratos/containers/set_identity_function.h```
- ```kratos/containers/pointer_vector_map.h```
- ```kratos/containers/pointer_vector_set.h```
- ```kratos/spaces/ublas_space.h```
- ```kratos/includes/model_part.h```

No errors just yet, but they'll have to be replaced sooner or later since they were [officially removed](https://en.cppreference.com/w/cpp/utility/functional#Old_binders_and_adaptors) in ```C++17```.

loumalouomega:
Idem (I have the same issue with gcc 12)

roigcarlo:
@KratosMultiphysics/all As you may have noticed, all changes are already in place and our official version is C++17. Thx all!

PR #9937, opened 2022-05-30
Core/enable cxx17

roigcarlo:
**DO NOT MERGE**

Draft of changes for C++17. Wip. 
This PR is to address some problems observed in #9916 mostly regarding deprecation of features towards C++20.

List of current changes:
- All `unitary_function` and `binary_function` instances have been removed from the code.
- The `return_type` from the functions above has been replaced with `decltype()` when necessary
- Current instances of code that already used `decltype()` using an object have been replaced with the a `std::declval` of that object to avoid creating an instance:
    ```C++
    decltype(TDataType)
    ```

    Changes to:

    ```C++
    decltype(std::declval<TDataType>())
    ```

    This is specially important because of Dofs in `model_part.h`:

    ```C++
    typedef PointerVectorSet<DofType,
            SetIdentityFunction<DofType>,
            std::less<SetIdentityFunction<DofType>>::result_type,
            std::equal_to<SetIdentityFunction<DofType>>::result_type,
            DofType* > DofsArrayType;
    ```

    Changes to:

    ```C++
    typedef PointerVectorSet<DofType,
            SetIdentityFunction<DofType>,
            std::less<decltype(std::declval<IndexedObject>()(std::declval<DofType>()))>,
            std::equal_to<decltype(std::declval<IndexedObject>()(std::declval<DofType>()))>,
            DofType* > DofsArrayType;
    ```

    which has undermined some inconsistencies with `DofType` in several core classes (For instance `scheme.h`):

    ```C++
    /// DoF iterator type definition
    typedef typename PointerVectorSet<TDofType, IndexedObject>::iterator DofIterator;
    /// DoF constant iterator type definition
    typedef typename PointerVectorSet<TDofType, IndexedObject>::const_iterator DofConstantIterator;
    ```

    Should be defined as:

    ```C++
    /// DoF iterator type definition
    typedef typename PointerVectorSet<TDofType, 
            std::less<decltype(std::declval<IndexedObject>()(std::declval<DofType>()))>,
            std::equal_to<decltype(std::declval<IndexedObject>()(std::declval<DofType>()))>,
            DofType* >::iterator DofIterator;
    /// DoF constant iterator type definition
    typedef typename PointerVectorSet<TDofType, 
            std::less<decltype(std::declval<IndexedObject>()(std::declval<DofType>()))>,
            std::equal_to<decltype(std::declval<IndexedObject>()(std::declval<DofType>()))>,
            DofType* >::const_iterator DofConstantIterator;
    ```

    Otherwise it forces the `PointerVectorSet` of some classes to work with `DofType::Pointer` instead of `DofType*` Which could potentially have an impact in performance.
    Also there seems to be an inconsistency on whether to use `IndexedObject` or `SetIdentityFunction` for the Key/Hash of the `PointerVectorSet<DofType>`. I asume `IndexedObject` is the correct approach since otherwise the sorting function fails ( seems that only RomApplication was sorting dofs...)


matekelemen:
> ```
> typedef PointerVectorSet<DofType,
>      SetIdentityFunction<DofType>,
>      std::less<decltype(std::declval<IndexedObject>()(std::declval<DofType>()))>,
>      std::equal_to<decltype(std::declval<IndexedObject>()(std::declval<DofType>()))>,
>      DofType* > DofsArrayType;
> ```

```IndexObject::operator()``` is a template and (as far as I can tell) has no specializations, so why not use the member alias it uses for its return type?
```
typedef PointerVectorSet<DofType,
        SetIdentityFunction<DofType>,
        std::less<IndexObject::IndexType>,
        std::equal_to<IndexObject::IndexType>,
        DofType* > DofsArrayType;
```

**Comment by roigcarlo** [comment]
@loumalouomega or @pooyan-dadvand 

Any idea about this error:
```
/home/roigcarlo/Kratos/kratos/processes/from_json_check_result_process.cpp:584:81: error: no matching function for call to ‘distance(Kratos::Parameters::const_iterator, Kratos::Parameters::const_iterator)’
  584 |     auto size_parameters =[](const Parameters& rParameters){return std::distance(rParameters.begin(), rParameters.end());};
      |                                                                    ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/c++/12.1.0/string:46,
                 from /home/roigcarlo/Kratos/kratos/containers/model.h:18,
                 from /home/roigcarlo/Kratos/kratos/processes/from_json_check_result_process.cpp:18:
/usr/include/c++/12.1.0/bits/stl_iterator_base_funcs.h:146:5: note: candidate: ‘template<class _InputIterator> constexpr typename std::iterator_traits< <template-parameter-1-1> >::difference_type std::distance(_InputIterator, _InputIterator)’
  146 |     distance(_InputIterator __first, _InputIterator __last)
      |     ^~~~~~~~
/usr/include/c++/12.1.0/bits/stl_iterator_base_funcs.h:146:5: note:   template argument deduction/substitution failed:
/usr/include/c++/12.1.0/bits/stl_iterator_base_funcs.h: In substitution of ‘template<class _InputIterator> constexpr typename std::iterator_traits< <template-parameter-1-1> >::difference_type std::distance(_InputIterator, _InputIterator) [with _InputIterator = Kratos::Parameters::const_iterator_adaptor]’:
/home/roigcarlo/Kratos/kratos/processes/from_json_check_result_process.cpp:584:81:   required from here
/usr/include/c++/12.1.0/bits/stl_iterator_base_funcs.h:146:5: error: no type named ‘difference_type’ in ‘struct std::iterator_traits<Kratos::Parameters::const_iterator_adaptor>’
```

I added those to the definition of both const_iterator and iterator in the Parameters class:

```C++
using iterator_category = std::forward_iterator_tag;
using difference_type   = std::ptrdiff_t;
using value_type        = Parameters;
using pointer           = Parameters*;
using reference         = Parameters&;
```

matekelemen:
Just a guess, but the ```pointer``` and ```reference``` member aliases of ```const_iterator_adaptor``` probably need to inherit the constness:
```cpp
using pointer = const Parameters*;
using reference = const Parameters&;
```

roigcarlo:
@matekelemen mm unfortunately no :S, but thx for the suggestion

pooyan-dadvand:
Are you sure that the definition of the value_iterator is not in conflict with others?

Meanwhile, to me, the error seems to come from some cross includes

loumalouomega:
> Meanwhile, to me, the error seems to come from some cross includes

I also thought that, I tried different configurations and I didn't get it work...

matekelemen:
@roigcarlo ah found it, your aliases are ```private```, so ```iterator_traits``` can't find them. Make them ```public```.

applications/CoSimulationApplication/custom_external_libraries/CoSimIO/CMakeLists.txt
```diff
@@ -85,7 +85,7 @@ if(MSVC)
     endif()
 
 elseif(${CMAKE_COMPILER_IS_GNUCXX})
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17")
```
philbucher: this is a small standalone lib, this should not be changed I would say

roigcarlo:
I would preffer to have everything compiled in the same standard (as long as its possible). Is there any reason to leave it in C++11?

philbucher:
Yes, compatibility with external solvers that use C++11
I wouldn't want to debug incompatibilities because of different standards 😅

In principle I would love to use 17 there as well but @pooyan-dadvand and I decided to leave it 11

roigcarlo:
@philbucher Done

philbucher:
Thanks!

applications/CoSimulationApplication/custom_external_libraries/CoSimIO/CMakeLists.txt
matekelemen: does that mean that ```CoSimIO``` will use a different ```libstdc++``` than the rest of Kratos? Is that safe?

roigcarlo: Yes it is safe, its the same scenario as linking with a fotran module

philbucher: Same as for any other system lib no?
Isnt that sorta garantied through ABI stability?

roigcarlo: Well C++11 broke the ABI compatibility for the strings with old standards (remember the infamous `_GLIBCXX_USE_CXX11_ABI`). It is expectable, but not 100% guaranteed

matekelemen: To be honest, I don't know much about compatibility and maintenance. I just noticed that all my binaries are updated whenever I get a new GCC version. And new C++ standards sometimes remove features / modify some signatures.

roigcarlo:
@pooyan-dadvand  @philbucher seems that the only problem is the LockObject move constructor in intel and we are ready to go.

Do we have any alternative to this? Sorry to ask but I did not follow the development of neither the Lock object and the chunk closely so I am a little bit blind on what is used for.

roigcarlo:
@philbucher Please take a look at the last changes (the ones related with the default MPIDatacomms). I am never 100% that I am not forcing a communicator that cannot be used with the cosim app.

philbucher:
> @philbucher Please take a look at the last changes (the ones related with the default MPIDatacomms). I am never 100% that I am not forcing a communicator that cannot be used with the cosim app.

Coincidentally I fixed the utility properly in #10110

In a nutshell, don't hardcode communicators to `world`. EIther take it from a ModelPart or, if it is a utility, then make it an input argument

roigcarlo:
@philbucher I just move your commits here and adapted the rest (nothing should be using world now, its either the default or the one from the mpda in the process)

If there is no further issues, this should be ready to merge.

philbucher (dismissed):

roigcarlo:
Mmm seems that after the last changes a new deprecation appeared, any idea `CheckSerializationForSimpleType`. It looks like an important check to me, why is used in core comms if its deprecated Oo?

philbucher:
Iirc this is a sort of security check added by Jordi
Basically the data exchange is done via serialization even though it can be done directly (i.e. for simple types)

However deprecation warnings should be disabled in the CI, maybe the compiler has changed

=> you can disable the deprecation warnings, I am pretty sure they were disabled when I added the Intel compiler 
And also they are disabled for the other compilers

RiccardoRossi:
well if it follows a suboptimal path we should discover why...

philbucher (approved):
I pushed some minor CI improvement, from my side it is ready :tada:

loumalouomega:
Are we C++17 already?

philbucher:
Yes 🎉

PR #9938, opened 2022-05-30
[CLApp] avoid to call InitializeSolutionStep in CL

AlejandroCornejo:
I'm adding the RequiresInitializeSolutionStep to false since it is not necessary. Safe computational cost.


AlejandroCornejo:
Theoretically this should not change anything. I've changed those CL in which the initializeSolStep is empty (all but the HCF law).

SergioJimenezReyes (approved):
It makes sense to me, but I guess you refer to the InitializeMaterialResponse

PR #9940, opened 2022-06-03
[CLApp] Adding function accessor to the core (Thermomechanical PR related)

AlejandroCornejo:
**📝 Description**

In this PR I am adding a proposal to a property accessor prototype discussed in https://github.com/KratosMultiphysics/Kratos/pull/9053.

**This is a work in progress**, currently is defined only for `doubles` but it will be templated for any kind of variable.
@RiccardoRossi @philbucher @loumalouomega @rubenzorrilla 

Let the hunger games begin!


kratos/includes/property_accessor.h
```diff
@@ -0,0 +1,244 @@
...
+
+///@}
+///@name Kratos Classes
+///@{
+
+/**
+ * PropertyAccessor sets a proper way of returning a value from the properties
+ * It is used to properly interpolate material properties according to temperature, tables, etc...
+*/
+class KRATOS_API(KRATOS_CORE) PropertyAccessor
```
loumalouomega: Move the code to  a cpp. There is no template, so it's fine

kratos/includes/property_accessor.h
```diff
@@ -0,0 +1,244 @@
...
+    ///@name Type Definitions
+    ///@{
+    typedef Geometry<Node<3>> GeometryType;
+
+    typedef std::function<double(const std::string&, const Properties&, const GeometryType&) > AccessorType;
+
+    /// Pointer definition of ProcessInfo
+    KRATOS_CLASS_POINTER_DEFINITION(PropertyAccessor);
+
+
```
loumalouomega: To many spaces

kratos/includes/property_accessor.h
```diff
@@ -0,0 +1,244 @@
...
+    ///@}
+    ///@name Life Cycle
+    ///@{
+
+
```
loumalouomega: To many spaces

loumalouomega (changes requested):
Huuum, looks empty and incomplete.

kratos/includes/property_accessor.h
AlejandroCornejo: It will be a template for the variable type I think

loumalouomega: Huuum, remove KRATOS_API then

AlejandroCornejo:
Hi @loumalouomega ! it's true that it looks "empty" but this is the idea. This class is totally auxiliar. The properties will define new accessors according to your needs. In my case I need the standard return of the prop, and according to a table. 

If another person needs an analytical evolution of the prop, he/she will add the accessor to the property.

I think this was the original idea of @RiccardoRossi :)

loumalouomega:
> I dont completely get the idea of how this works, At which moment and who adds the accessors? Are these ownned by the element..?

A test would clarify...

AlejandroCornejo:
> I dont completely get the idea of how this works, At which moment and who adds the accessors? Are these ownned by the element..?

Hi! the discussion is here: https://github.com/KratosMultiphysics/Kratos/pull/9053

The accessor will be inside the `Properties` :)

AlejandroCornejo:
> > I dont completely get the idea of how this works, At which moment and who adds the accessors? Are these ownned by the element..?
> 
> A test would clarify...

I'll do a test when a proper implementation is decided 👯

ddiezrod:
> > I dont completely get the idea of how this works, At which moment and who adds the accessors? Are these ownned by the element..?
> 
> Hi! the discussion is here: #9053
> 
> The accessor will be inside the `Properties` :)

So when you add a Property to the container, you also add an accessor to it? I think that would work, although you will not be able to have two properties with same name with different types (probably its better that this is not allowed but it could break code I guess)

kratos/includes/property_accessor.h
```diff
@@ -0,0 +1,242 @@
...
+private:
+    ///@name Static Member Variables
+    ///@{
+
+    ///@}
+    ///@name Member Variables
+    ///@{
+
+    double mValue;
+    std::map<std::string, AccessorType*> mListOfAccessors;
```
loumalouomega: Unordered map will be more efficient

loumalouomega:
```suggestion
    std::unordered_map<std::string, AccessorType*> mListOfAccessors;
```

kratos/includes/property_accessor.h
```diff
@@ -0,0 +1,242 @@
...
+
+
+    ///@}
+    ///@name Operators
+    ///@{
+
+    /**
+     * @brief default constructor
+     */
+    PropertyAccessor(double Value)
```
loumalouomega:
```suggestion
    PropertyAccessor(const double Value)
```

kratos/includes/property_accessor.h
```diff
@@ -0,0 +1,242 @@
...
+    {
+        mValue = Value;
+    }
+
+    /**
+     * @brief This method add a new acessor to a certain variable name
+     */
+    void AddAccessor(const std::string& rVariableName, AccessorType* pAccessor)
```
loumalouomega: I would consider the id of the variable, it is faster than the string

kratos/includes/property_accessor.h
```diff
@@ -0,0 +1,242 @@
...
+    {
+        mListOfAccessors[rVariableName] = pAccessor;
+    }
+
+    /**
+     * @brief This method return the value of the required variable
+     */
+    double GetProperty(const std::string& rVariableName, const Properties& rProperties, const GeometryType& rGeometry)
```
loumalouomega: Idem

kratos/includes/property_accessor.h
```diff
@@ -0,0 +1,242 @@
...
+    {
+        // If it is in the list, give back the corresponding accessor, otherwise give back value
+        auto it = mListOfAccessors.find(rVariableName);
+        if (it != mListOfAccessors.end()) {
+            auto& r_function = *(*it).second;
```
loumalouomega: const?

loumalouomega (commented):
I added more comments

kratos/includes/property_accessor.h
```diff
@@ -0,0 +1,242 @@
...
+
+///@}
+///@name Kratos Classes
+///@{
+
+/**
+ * PropertyAccessor sets a proper way of returning a value from the properties
+ * It is used to properly interpolate material properties according to temperature, tables, etc...
+*/
+class KRATOS_API(KRATOS_CORE) PropertyAccessor
```
loumalouomega:
```suggestion
class PropertyAccessor
```

kratos/includes/property_accessor.h
```diff
@@ -0,0 +1,242 @@
...
+//                   Multi-Physics
+//
+//  License:         BSD License
+//                     Kratos default license: kratos/license.txt
+//
+//  Main authors:    Alejandro Cornejo
+//                   Riccardo Rossi
+//
+//
+
```
loumalouomega:
```suggestion
```

kratos/includes/property_accessor.h
```diff
@@ -0,0 +1,242 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:         BSD License
+//                     Kratos default license: kratos/license.txt
```
loumalouomega: @rubenzorrilla saying that the indentation is not right in 1,2,3...

loumalouomega:
@jrubiogonzalez this could be of your interest

AlejandroCornejo:
> @jrubiogonzalez this could be of your interest

Actually I'd like to really ove this forward... For a project we want this to be available in the master. I offer myself to implement this but I need your opinion and support!! :) @RiccardoRossi @loumalouomega

loumalouomega:
> > @jrubiogonzalez this could be of your interest
> 
> Actually I'd like to really move this forward... For a project we want this to be available in the master. I offer myself to implement this but I need your opinion and support!! :) @RiccardoRossi @loumalouomega

I was pinging because I think in @KratosMultiphysics/altair this feature will be valuable. Maybe we can discuss soon all together soon.

AlejandroCornejo:
> > > @jrubiogonzalez this could be of your interest
> > 
> > 
> > Actually I'd like to really move this forward... For a project we want this to be available in the master. I offer myself to implement this but I need your opinion and support!! :) @RiccardoRossi @loumalouomega
> 
> I was pinging because I think in @KratosMultiphysics/altair this feature will be valuable. Maybe we can discuss soon all together soon.

That could be great!! let me know whenever you start with this

PR #9941, opened 2022-06-03
[MeshingApplication][MMG] Add option to set detection angle value

marcnunezc:
**📝 Description**
This PR adds the option to set the detection angle for sharp edges which was not included in the interface.

https://www.mmgtools.org/mmg-remesher-try-mmg/mmg-remesher-options/mmg-remesher-option-ar

**🆕 Changelog**
E.g.
- Added parameter to force the detection angle and another parameter to set the value.


loumalouomega (approved):
Consistent with current implemenetation

PR #9949, opened 2022-06-09
Core/flux based redistance

pablobecker:
**📝 Description**
adds a new redistance process,
The method has two distinct steps:
Step 1 computes a pseudo-potential to determine the direction in which the DISTANCE must grow. 
Step 2 takes the direction of this potential and solves an equation such that the gradient in this direction == 1. Unlike the variational redistance, this new method does not allow the direction to be modified. i.e. the direction obtained from step 1 is kept.  THis is done by solving a pure convection equation with an RHS to increase the DIST along the strealine , with the desired gradient. Step 2 is iterative, until |grad(dist)|=1.
The element has some tricks to ensure robustness, such as checking converging/diverging flow or checking if upwind element is present.

**🆕 Changelog**
- Added flux based redistance
- Added flux redistance element
- Added test for the new process.
- Adds needed variables.


maceligueta:
No problem from the DEM side.

RiccardoRossi:
@rubenzorrilla @uxuech this looks very promising for your applications.

pablobecker:
I have to add the interface for the hexas, but for the moment you can check for tetras.

Note that for step1 , we could use the "potential" that @RiccardoRossi implemented for the variational redistance, step1. It could be better than the transient diffusion approach i used here, especially near the borders of the domain ( the solution to the diffusion problem in the boundaries has zero gradients, so the direction will not be ok in this case. I assume getting the distance correct in these regions, far away from the interface, is not important. But if it is for you, then ricc's approach might be better)

pablobecker:
ping @rubenzorrilla @uxuech could you take a look?

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+    Ncontainer(1,0) = one_sixt; Ncontainer(1,1) = two_third; Ncontainer(1,2) = one_sixt;
+    Ncontainer(2,0) = two_third; Ncontainer(2,1) = one_sixt; Ncontainer(2,2) = one_sixt;
+}
+
+//unused, but needed for compilation
+template <unsigned int TDim , unsigned int TNumNodes >
+void DistanceCalculationFluxBasedElement<TDim,TNumNodes>::GetSimplexShapeFunctionsOnGauss(BoundedMatrix<double,TNumNodes,TNumNodes>& Ncontainer)
+{
+    KRATOS_ERROR << "CALLING TEMPLATED GetSimplexShapeFunctionsOnGauss" << std::endl;
+}
```
rubenzorrilla, line 52:
```suggestion
```
This is required because in the bottom you're doing the specialization for 3 and 8 template arguments. Indeed, as it is right now it doesn't work as calling the 3 and 8 class specialization would get here and throw an error. You either remove that specialization or implement the `GetSimplexShapeFunctionsOnGauss` for the 3 and 8 template arguments case.

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+    double volume;
+    GeometryUtils::CalculateGeometryData(rGeometry, DN_DX, N, volume);
+    GetSimplexShapeFunctionsOnGauss(rNContainer);
+    for (unsigned int i = 0; i < 4; ++i) {
+        rGaussWeights[i] = volume/4.0;
+        rDN_DXContainer[i] = DN_DX;
+    }
+
+    
+}
```
rubenzorrilla, line 91: These specializations are not required. You can do the same by using the template arguments.

kratos/elements/distance_calculation_flux_based_element.cpp
rubenzorrilla: OK. Now I see... For the hexas you're using the generic low performant function below...

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+template <unsigned int TDim , unsigned int TNumNodes >
+void DistanceCalculationFluxBasedElement<TDim,TNumNodes>::CalculateGaussPointsData(
+    const GeometryType& rGeometry,
+    BoundedVector<double,TNumNodes> &rGaussWeights,
+    BoundedMatrix<double,TNumNodes,TNumNodes> &rNContainer,
+    array_1d<BoundedMatrix<double, TNumNodes, TDim>, TNumNodes>& rDN_DXContainer 
+    )
+{
+    Vector det_j_vector(TNumNodes);
+    // const auto& r_geom = this->GetGeometry();
```
rubenzorrilla:
```suggestion
```

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+    const auto integration_method = GeometryData::IntegrationMethod::GI_EXTENDED_GAUSS_2;
+    rNContainer = rGeometry.ShapeFunctionsValues(integration_method);
+    ShapeFunctionsGradientsType DN_DXContainer_unbounded;
+    //getting information from geometry in DenseVector<Matrix> format
+    rGeometry.ShapeFunctionsIntegrationPointsGradients(DN_DXContainer_unbounded, det_j_vector, integration_method);
+
+
+    const unsigned int number_of_gauss_points = rGeometry.IntegrationPointsNumber(integration_method);
+    const GeometryType::IntegrationPointsArrayType& integration_points = rGeometry.IntegrationPoints(integration_method);
```
rubenzorrilla:
```suggestion
    const auto& r_integration_points = rGeometry.IntegrationPoints(integration_method);
```

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+    for (unsigned int g = 0; g < number_of_gauss_points; ++g){
+        rGaussWeights[g] = det_j_vector[g] * integration_points[g].Weight();
```
rubenzorrilla:
```suggestion
        rGaussWeights[g] = det_j_vector[g] * r_integration_points[g].Weight();
```

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+void DistanceCalculationFluxBasedElement<TDim, TNumNodes >::CalculateLocalSystem(
+    MatrixType &rLeftHandSideMatrix,
+    VectorType &rRightHandSideVector,
+    const ProcessInfo &rCurrentProcessInfo)
+{
+    KRATOS_TRY
+
+    // Resize and intialize output
+    if (rLeftHandSideMatrix.size1() != LocalSize)
+        rLeftHandSideMatrix.resize(LocalSize, LocalSize, false);
```
rubenzorrilla:
```suggestion
    if (rLeftHandSideMatrix.size1() != LocalSize || rLeftHandSideMatrix.size2() != LocalSize)
        rLeftHandSideMatrix.resize(LocalSize, LocalSize, false);
```

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+    if (rRightHandSideVector.size() != LocalSize)
+        rRightHandSideVector.resize(LocalSize, false);
+
+    noalias(rLeftHandSideMatrix) = ZeroMatrix(LocalSize, LocalSize);
+    noalias(rRightHandSideVector) = ZeroVector(LocalSize);
+
+    const unsigned int step = rCurrentProcessInfo[FRACTIONAL_STEP];
```
rubenzorrilla:
```suggestion
    const unsigned int step = rCurrentProcessInfo[FRACTIONAL_STEP];
```
I think a more self-descriptive name variable must be used for this.

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+    
+    if(step == 1){ //solve a transcient diffusion problem
+        CalculatePotentialFlowSystem(rLeftHandSideMatrix, 
+                                     rRightHandSideVector,
+                                     rCurrentProcessInfo);
```
rubenzorrilla:
```suggestion
        CalculatePotentialFlowSystem(
            rLeftHandSideMatrix, 
            rRightHandSideVector,
            rCurrentProcessInfo);
```
Style guide suggestion. Also happens in other places.

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+    array_1d<double, NumNodes > nodal_values;
+    for(unsigned int i=0; i<NumNodes; i++){
+        nodal_values[i] = GetGeometry()[i].FastGetSolutionStepValue(DISTANCE);
+    }
+
+    //computing density so that Fourier number of the complete domain = 1
+    //this ensures that all regions notice the effect of the fixed nodes.
+    const double domain_length = rCurrentProcessInfo[CHARACTERISTIC_LENGTH];
+    const double density  = 1.0/ (domain_length*domain_length);
```
rubenzorrilla, line 187: A check for non-zero `CHARACTERISTIC_LENGTH` must be added. As far as I understand, this is not expected to change during the simulation so I'd add it in the `Check` function to do it once. Note that this requires setting this value at the time that the solver `Initialize` happens.

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+    
+    //Looping Gauss Points
+    for (unsigned int igauss = 0; igauss<NumNodes; igauss++) {
```
rubenzorrilla:
```suggestion
    const std::size_t n_gauss = gauss_weights.size();
    for (unsigned int igauss = 0; igauss<n_gauss; igauss++) {
```
Otherwise it would be no longer valid if changing the quadrature.

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+        array_1d<double, TNumNodes > N = row(N_container, igauss);
+        noalias(rLeftHandSideMatrix) += gauss_weights[igauss]*prod(DN_DX_container[igauss],trans(DN_DX_container[igauss]));
+        const double mass_factor = density*gauss_weights[igauss];
+        const double d_gauss = inner_prod(N,nodal_values);
+        const double initial_value = d_gauss > 0.0 ? domain_length*10.0 : -domain_length*10.0 ;
```
rubenzorrilla: Why multiplying by 10?

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+        nodal_values[i] = GetGeometry()[i].FastGetSolutionStepValue(DISTANCE);
+        const auto& rVel_i = GetGeometry()[i].GetValue(POTENTIAL_GRADIENT);
+        for(unsigned int j=0; j<TDim; j++){
+            v[i][j] = rVel_i[j];
+        }
+        v_unit[i] =  v[i]/norm_2(v[i]); // |v_unit| = 1
+        avg_v_unit += v_unit[i];
+        has_fixed_node = has_fixed_node || GetGeometry()[i].IsFixed(DISTANCE);
+    }
+    avg_v_unit/=static_cast<double>(NumNodes);
```
rubenzorrilla: Isn't the cast automatically done?

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+    BoundedMatrix<double, TNumNodes, TDim> avg_DN_DX = ZeroMatrix(TNumNodes, TDim);
+    for(unsigned int i=0; i<TNumNodes; i++){
+        avg_DN_DX += DN_DX_container[i] /static_cast<double>(TNumNodes);
```
rubenzorrilla: Isn't the cast automatically done?

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+    }
+    
+    //adjusting RHS to get |grad|=1
+    if(rCurrentProcessInfo[NL_ITERATION_NUMBER] > 1) {
```
rubenzorrilla: As far as I understand, this `NL_ITERATION_NUMBER` refers to the non-linear iterations of the NR strategy solving this problem. If this is the case, I'd better document it in here.

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+        const array_1d<double, TDim> avg_grad = prod(trans(avg_DN_DX), nodal_values);
+        const double grad_modulus = norm_2(avg_grad);
```
rubenzorrilla: Shouldn't you check that this norm is not zero in here?

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+        const double new_correction_coeff = 1.0/grad_modulus;
+        const double relaxation_factor = 0.5;
```
rubenzorrilla: Do you think it is a good idea to make this user-definable? Besides, we can explore the usage of some formulas to accelerate this.

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+        mCorrectionCoefficient = mCorrectionCoefficient *(1.0-relaxation_factor) + new_correction_coeff*relaxation_factor;
+    }
+    else {
+        mCorrectionCoefficient = 1.0;
+    }
+
+    //computing element size(for Tau)
+    const double h = ComputeH(avg_DN_DX);
```
rubenzorrilla: Please use the utilities at `ElementSizeCalculator`.

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+    //DECIDE IF CONVECTION IS ACTIVE/INACTIVE
+   //we decide by checking if flow converges to /diverges from this element
+   //if all velocity vectors are oriented in the same direction, then convection is valid
+   //but if the element acts as a "sink" or "source", convection is turned off to avoid instabiliities.
+    bool add_convection = true;
+    double average_unit_vel_misaligment = 0.0;
+    for(unsigned int i=0; i<NumNodes; i++) {
+        average_unit_vel_misaligment += norm_2( v_unit[i]  - avg_v_unit );
+    }
+    average_unit_vel_misaligment/=static_cast<double>(NumNodes);
```
rubenzorrilla: Isn't the cast done automatically?

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+    if(average_unit_vel_misaligment>0.33) add_convection=false;
```
rubenzorrilla, line 253: Why 0.33?

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+    //checking if the flow comes from face that has neighbour element (only if it is not inlet)
+    if( !has_fixed_node && this->Has(NEIGHBOUR_ELEMENTS) && this->GetValue(NEIGHBOUR_ELEMENTS).size()==GetGeometry().FacesNumber())
```
rubenzorrilla:
```suggestion
    if( !has_fixed_node && this->GetValue(NEIGHBOUR_ELEMENTS).size()==GetGeometry().FacesNumber())
```
As far as I understand this requires the `NEIGHBOUR_ELEMENTS` to be computed. Then I think we can assume that these are already there by either calling the neighbors calculation or checking it in the `FluxBasedRedistanceProcess`.

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+    {
+        double value_lowest_prod = 0.0;
+        unsigned int face_lowest_prod = 0;
+        const auto elem_boundaries = GetGeometry().LocalSpaceDimension()==3 ? GetGeometry().GenerateFaces() : GetGeometry().GenerateEdges() ;
+
+        for(unsigned int i=0; i<elem_boundaries.size(); i++) {
+            Point cond_center = elem_boundaries[i].Center();
```
rubenzorrilla:
```suggestion
            const auto cond_center = elem_boundaries[i].Center();
```
`UnitNormal` accepts a const input parameter.

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+    BoundedMatrix<double,NumNodes, NumNodes> aux2 = ZeroMatrix(NumNodes, NumNodes); //terms multiplying phi
+    
+    //source term
+    array_1d<double, NumNodes> rhs_volumetric_heat = ZeroVector(NumNodes);
+
+    //Gauss point container (using nodal integration)
+    const bounded_matrix<double, NumNodes, NumNodes> Ncontainer=IdentityMatrix(NumNodes); 
+
+    //Looping Gauss Points
+    for (unsigned int igauss = 0; igauss<NumNodes; igauss++) {
```
rubenzorrilla:
```suggestion
    for (unsigned int igauss = 0; igauss<n_gauss; igauss++) {
```
Same comment, I wouldn't assume this just in case we change the quadrature.

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+{
+    KRATOS_TRY;
+    // Generic geometry check
+    int out = Element::Check(rCurrentProcessInfo);
+    if (out != 0) {
+        return out;
+    }
+
+    const GeometryType& r_geometry = this->GetGeometry();
+
```
rubenzorrilla:
```suggestion
    const auto& r_geometry = this->GetGeometry();
```

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+    for(unsigned int i=0; i<NumNodes; ++i){
+        const Node<3>& rNode = r_geometry[i];
+        KRATOS_CHECK_VARIABLE_IN_NODAL_DATA(DISTANCE,rNode);
+        // KRATOS_CHECK_VARIABLE_IN_NODAL_DATA(ADJOINT_VECTOR_1,rNode);
```
rubenzorrilla:
```suggestion
```

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+    int out = Element::Check(rCurrentProcessInfo);
+    if (out != 0) {
+        return out;
+    }
+
+    const GeometryType& r_geometry = this->GetGeometry();
+
+    for(unsigned int i=0; i<NumNodes; ++i){
+        const Node<3>& rNode = r_geometry[i];
+        KRATOS_CHECK_VARIABLE_IN_NODAL_DATA(DISTANCE,rNode);
```
rubenzorrilla, line 365: You also use `VELOCITY` right?

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+        KRATOS_CHECK_DOF_IN_NODE(DISTANCE,rNode);
+    }
+
+    // If this is a 2D problem, check that nodes are in XY plane
+    if ( TDim == 2){
+        for (unsigned int i=0; i<NumNodes; ++i) {
+            if (std::abs(r_geometry[i].Z())>1e-9)
+                KRATOS_ERROR << "Node " << r_geometry[i].Id() << "has non-zero Z coordinate." << std::endl;
+        }
+    }
```
rubenzorrilla: Wouldn't it be better to check this in the process?

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+void DistanceCalculationFluxBasedElement<TDim,TNumNodes>::PrintInfo(
+    std::ostream &rOStream) const
+{
+    rOStream << this->Info() << std::endl;
+}
+
+template <unsigned int TDim , unsigned int TNumNodes >
+void DistanceCalculationFluxBasedElement< TDim, TNumNodes  >::EquationIdVector(EquationIdVectorType &rResult, const ProcessInfo &rCurrentProcessInfo) const
+{
+    const GeometryType& r_geometry = this->GetGeometry();
```
rubenzorrilla:
```suggestion
    const auto& r_geometry = this->GetGeometry();
```

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+    for (unsigned int i = 0; i < NumNodes; ++i){
+        rResult[LocalIndex++] = r_geometry[i].GetDof(DISTANCE).EquationId();
+    }
+}
+
+
+template <unsigned int TDim , unsigned int TNumNodes >
+void DistanceCalculationFluxBasedElement< TDim , TNumNodes >::GetDofList(DofsVectorType &rElementalDofList, const ProcessInfo &rCurrentProcessInfo) const
+{
+    const GeometryType& r_geometry = this->GetGeometry();
```
rubenzorrilla:
```suggestion
    const auto& r_geometry = this->GetGeometry();
```

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+            double h_inv = 0.0;
+            for(unsigned int k=0; k<TDim; k++)
+            {
+                h_inv += DN_DX(i,k)*DN_DX(i,k);
+            }
+            h = std::max(h, 1.0 / h_inv);
+        }
+        h = std::sqrt(h);
+        return h;
+}
```
rubenzorrilla: As commented above, I'd remove this in favor of the `ElementSizeCalculator`.

Aside of this, I think that the gradient-based element size calculator cannot be used for the hexas. I'd suggest using the average element size in there instead.

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+template <unsigned int TDim , unsigned int TNumNodes >
+void DistanceCalculationFluxBasedElement< TDim, TNumNodes >::AddExplicitContribution(const ProcessInfo& rCurrentProcessInfo)
+{
+   	GeometryType& rGeometry = this->GetGeometry();
```
rubenzorrilla:
```suggestion
   	auto& rGeometry = this->GetGeometry();
```

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ \.
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:		 BSD License
+//			 Kratos default license: kratos/license.txt
+//
+//  Main authors:    Daniel Diez, Pablo Becker
```
rubenzorrilla: Header tabs formatting.

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+        return h;
+}
+
+template <unsigned int TDim , unsigned int TNumNodes >
+void DistanceCalculationFluxBasedElement< TDim, TNumNodes >::AddExplicitContribution(const ProcessInfo& rCurrentProcessInfo)
+{
+   	GeometryType& rGeometry = this->GetGeometry();
+
+	BoundedVector<double,TNumNodes> gauss_weights;
+    BoundedMatrix<double,TNumNodes,TNumNodes> N_container;
```
rubenzorrilla:
```suggestion
    BoundedVector<double,TNumNodes> gauss_weights;
    BoundedMatrix<double,TNumNodes,TNumNodes> N_container;
```
Our tabbing is 4 spaces.

kratos/elements/distance_calculation_flux_based_element.cpp
```diff
@@ -0,0 +1,502 @@
...
+                h_inv += DN_DX(i,k)*DN_DX(i,k);
+            }
+            h = std::max(h, 1.0 / h_inv);
+        }
+        h = std::sqrt(h);
+        return h;
+}
+
+template <unsigned int TDim , unsigned int TNumNodes >
+void DistanceCalculationFluxBasedElement< TDim, TNumNodes >::AddExplicitContribution(const ProcessInfo& rCurrentProcessInfo)
```
rubenzorrilla, line 433: I don't understand this. Isn't this method conceived to work with implicit strategies?

kratos/elements/distance_calculation_flux_based_element.h
```diff
@@ -0,0 +1,359 @@
...
+//                   Multi-Physics
+//
+//  License:		 BSD License
+//			 Kratos default license: kratos/license.txt
+//
+//  Main authors:    Daniel Diez, Pablo Becker
+//
+
+#if !defined(KRATOS_DISTANCE_CALCULATION_FLUX_BASED_ELEMENT)
+#define  KRATOS_DISTANCE_CALCULATION_FLUX_BASED_ELEMENT
```
rubenzorrilla:
```suggestion
#pragma once
```

kratos/elements/distance_calculation_flux_based_element.h
```diff
@@ -0,0 +1,359 @@
...
+
+}; // Class DistanceCalculationFluxBasedElement
+///@name Type Definitions
+///@{
+
+
+
+} // namespace Kratos.
+
+#endif
```
rubenzorrilla:
```suggestion
```
As we're using the `#pragma once`.

kratos/elements/distance_calculation_flux_based_element.h
```diff
@@ -0,0 +1,359 @@
...
+
+
+    ///@}
+private:
+
+    friend class Serializer;
+
+    void save(Serializer& rSerializer) const override
+    {
+        KRATOS_SERIALIZE_SAVE_BASE_CLASS(rSerializer, Element );
```
rubenzorrilla, line 320: I think you also need to save the `mCorrectionCoefficient`.

kratos/elements/distance_calculation_flux_based_element.h
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    void load(Serializer& rSerializer) override
+    {
+        KRATOS_SERIALIZE_LOAD_BASE_CLASS(rSerializer, Element);
```
rubenzorrilla, line 325: I think you also need to load the `mCorrectionCoefficient`.

kratos/elements/distance_calculation_flux_based_element.h
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    /// Assignment operator.
+    DistanceCalculationFluxBasedElement& operator=(DistanceCalculationFluxBasedElement const& rOther);
+
+    /// Copy constructor.
+    DistanceCalculationFluxBasedElement(DistanceCalculationFluxBasedElement const& rOther);
```
rubenzorrilla:
```suggestion
    /// Copy constructor.
    DistanceCalculationFluxBasedElement(DistanceCalculationFluxBasedElement const& rOther) = delete;
```
And move it to the life cycle section where it actually belongs.

kratos/elements/distance_calculation_flux_based_element.h
```diff
@@ -0,0 +1,359 @@
...
+        KRATOS_SERIALIZE_SAVE_BASE_CLASS(rSerializer, Element );
+    }
+
+    void load(Serializer& rSerializer) override
+    {
+        KRATOS_SERIALIZE_LOAD_BASE_CLASS(rSerializer, Element);
+    }
+
+    /// Assignment operator.
+    DistanceCalculationFluxBasedElement& operator=(DistanceCalculationFluxBasedElement const& rOther);
```
rubenzorrilla:
```suggestion
```
We can remove it if we delete the copy constructor.

kratos/elements/distance_calculation_flux_based_element.h
```diff
@@ -0,0 +1,359 @@
...
+        const ProcessInfo &rCurrentProcessInfo);
+        
+    //computes characteristic length
+    double ComputeH(BoundedMatrix<double,NumNodes, TDim>& DN_DX);
+
+    void CalculateGaussPointsData(  const GeometryType& rGeometry,
+                                    BoundedVector<double,TNumNodes> &rGaussWeights,
+                                    BoundedMatrix<double,TNumNodes,TNumNodes> &rNContainer,
+                                    array_1d<BoundedMatrix<double, TNumNodes, TDim>, TNumNodes>& rDN_DXContainer 
+                                 );
```
rubenzorrilla:
```suggestion
    void CalculateGaussPointsData(
        const GeometryType& rGeometry,
        BoundedVector<double,TNumNodes> &rGaussWeights,
        BoundedMatrix<double,TNumNodes,TNumNodes> &rNContainer,
        array_1d<BoundedMatrix<double, TNumNodes, TDim>, TNumNodes>& rDN_DXContainer);
```
Style guide suggestion. Please do so in the other places.

kratos/elements/distance_calculation_flux_based_element.h
```diff
@@ -0,0 +1,359 @@
...
+     * For the FillTime, the assembled source term is 1.0
+     * therefore the gradient is 1/|vel| in the directiion of the velocity
+     */
+    void CalculateDistanceSystem(
+        MatrixType &rLeftHandSideMatrix,
+        VectorType &rRightHandSideVector,
+        const ProcessInfo &rCurrentProcessInfo);
+        
+    //computes characteristic length
+    double ComputeH(BoundedMatrix<double,NumNodes, TDim>& DN_DX);
```
rubenzorrilla:
```suggestion
```
This is the one I suggest removing.

kratos/elements/distance_calculation_flux_based_element.h
rubenzorrilla: Besides doxygen documentation is missing.

kratos/elements/distance_calculation_flux_based_element.h
```diff
@@ -0,0 +1,359 @@
...
+    ///@{
+
+    /**
+     * @brief Assembles the local system for a laplacian, no coeffs
+     * @return A laplacian systes
+     */
+    // void CalculateLaplacianSystem(
+    //     MatrixType &rLeftHandSideMatrix,
+    //     VectorType &rRightHandSideVector,
+    //     const ProcessInfo &rCurrentProcessInfo);
```
rubenzorrilla:
```suggestion
```

kratos/elements/distance_calculation_flux_based_element.h
```diff
@@ -0,0 +1,359 @@
...
+    * @param NewId the ID of the new element
+    * @param pGeom a pointer to the geomerty to be used to create the element
+    * @param pProperties the properties assigned to the new element
+    * @return a Pointer to the new element
+    */
+    Element::Pointer Create(IndexType NewId,
+        GeometryType::Pointer pGeom,
+        Properties::Pointer pProperties) const override 
+    {
+        return Element::Pointer(new DistanceCalculationFluxBasedElement(NewId, pGeom, pProperties));
```
rubenzorrilla: `make_intrusive` function defined in `smart_pointers.h` should be used instead.

kratos/elements/distance_calculation_flux_based_element.h
```diff
@@ -0,0 +1,359 @@
...
+    * @param NewId the ID of the new element
+    * @param ThisNodes the nodes of the new element
+    * @param pProperties the properties assigned to the new element
+    * @return a Pointer to the new element
+    */
+    Element::Pointer Create(IndexType NewId,
+        NodesArrayType const& ThisNodes,
+        Properties::Pointer pProperties)const override
+    {
+        return Element::Pointer(new DistanceCalculationFluxBasedElement(NewId, GetGeometry().Create(ThisNodes), pProperties));
```
rubenzorrilla: `make_intrusive` function defined in `smart_pointers.h` should be used instead.

kratos/elements/distance_calculation_flux_based_element.h
```diff
@@ -0,0 +1,359 @@
...
+    * @param pGeometry Pointer to a geometry object
+    * @param pProperties Pointer to the element's properties
+    */
+    DistanceCalculationFluxBasedElement(IndexType NewId, GeometryType::Pointer pGeometry, Properties::Pointer pProperties) :
+        Element(NewId, pGeometry, pProperties)
+    {}
+
+    /// Destructor.
+    virtual ~DistanceCalculationFluxBasedElement() override
+    {}
```
rubenzorrilla:
```suggestion
    virtual ~DistanceCalculationFluxBasedElement() override = default;
```

kratos/elements/distance_calculation_flux_based_element.h
```diff
@@ -0,0 +1,359 @@
...
+
+    /// Constuctor using geometry and properties.
+    /**
+    * @param NewId Index of the new element
+    * @param pGeometry Pointer to a geometry object
+    * @param pProperties Pointer to the element's properties
+    */
+    DistanceCalculationFluxBasedElement(IndexType NewId, GeometryType::Pointer pGeometry, Properties::Pointer pProperties) :
+        Element(NewId, pGeometry, pProperties)
```
rubenzorrilla:
```suggestion
    DistanceCalculationFluxBasedElement(
        IndexType NewId,
        GeometryType::Pointer pGeometry,
        Properties::Pointer pProperties)
        : Element(NewId, pGeometry, pProperties)
```
Style guide suggestion. Please do so in the other constructors.

kratos/elements/distance_calculation_flux_based_element.h
```diff
@@ -0,0 +1,359 @@
...
+    typedef std::size_t IndexType;
+    typedef std::size_t SizeType;
+    typedef std::vector<std::size_t> EquationIdVectorType;
+    typedef std::vector< Dof<double>::Pointer > DofsVectorType;
+    typedef PointerVectorSet<Dof<double>, IndexedObject> DofsArrayType;
+    typedef GeometryType::ShapeFunctionsGradientsType ShapeFunctionsGradientsType;
+
+    static constexpr unsigned int Dim = TDim;
+
+    static constexpr unsigned int NumNodes = TNumNodes;
```
rubenzorrilla:
```suggestion
```
I don't think it is a good idea using these aliases for the template arguments, specially considering that there is no benefit as they are as verbose (only the `T` is removed). My point is that using the `TDim` and `TNumNodes` makes clear that these are template arguments when browsing the code. Besides you are mixing the usage of `NumNodes` and `TNumNodes` in the source implementation.

kratos/elements/distance_calculation_flux_based_element.h
```diff
@@ -0,0 +1,359 @@
...
+    typedef Geometry<NodeType> GeometryType;
+    typedef Geometry<NodeType>::PointsArrayType NodesArrayType;
+    typedef Vector VectorType;
+    typedef Matrix MatrixType;
+    typedef std::size_t IndexType;
+    typedef std::size_t SizeType;
+    typedef std::vector<std::size_t> EquationIdVectorType;
+    typedef std::vector< Dof<double>::Pointer > DofsVectorType;
+    typedef PointerVectorSet<Dof<double>, IndexedObject> DofsArrayType;
+    typedef GeometryType::ShapeFunctionsGradientsType ShapeFunctionsGradientsType;
```
rubenzorrilla: Instead of creating new type definitions the base class one must be reused as much as possible. This eases the maintenance of the code in the case of an eventual API change of the base `Element` class.

Besides I think that we should promote the `using` syntax instead of the `typedef` one considering that this is a new file addition. Thought in this case it makes no difference, it is a good practice for the future IMO (see a discussion [here](https://stackoverflow.com/questions/10747810/what-is-the-difference-between-typedef-and-using-in-c11)).

kratos/elements/distance_calculation_flux_based_element.h
```diff
@@ -0,0 +1,359 @@
...
+#define  KRATOS_DISTANCE_CALCULATION_FLUX_BASED_ELEMENT
+
+// System includes
+
+// Project includes
+#include "includes/define.h"
+#include "includes/element.h"
+#include "includes/variables.h"
+#include "includes/serializer.h"
+#include "utilities/geometry_utilities.h"
```
rubenzorrilla: I think this one can be moved to the source file.
```suggestion
```

kratos/tests/test_flux_based_redistance.py
```diff
@@ -0,0 +1,200 @@
...
+    #     distance_calculator.CalculateDistances(
+    #         model_part,
+    #         KratosMultiphysics.DISTANCE,
+    #         KratosMultiphysics.NODAL_VOLUME,
+    #         2,
+    #         2.0,
+    #         KratosMultiphysics.ParallelDistanceCalculator3D.CALCULATE_EXACT_DISTANCES_TO_PLANE)
+
+    #     for node in model_part.Nodes:
+    #         self.assertAlmostEqual(node.GetSolutionStepValue(KratosMultiphysics.DISTANCE), node.Y - free_surface_level, 10 )
```
rubenzorrilla:
```suggestion
```
Please remove commented code.

kratos/tests/test_flux_based_redistance.py
```diff
@@ -0,0 +1,200 @@
...
+            }
+        }""")
+
+        vtk_output_parameters["Parameters"]["model_part_name"].SetString("Main")
+        vtk_output_parameters["Parameters"]["file_format"].SetString("ascii")
+        process = vtk_output_process.Factory(vtk_output_parameters, current_model)
+        process.ExecuteInitialize()
+        process.ExecuteInitializeSolutionStep()
+        process.ExecuteFinalizeSolutionStep()
+        process.PrintOutput()
```
rubenzorrilla: This is always printing the output. This must be either removed or defaulted to false to be only switched on in case of need.

kratos/tests/test_flux_based_redistance.py
```diff
@@ -0,0 +1,200 @@
...
+import KratosMultiphysics
+from KratosMultiphysics import python_linear_solver_factory as linear_solver_factory
+import KratosMultiphysics.vtk_output_process as vtk_output_process
+import math
+import os
+
+def GetFilePath(fileName):
+    return os.path.join(os.path.dirname(os.path.realpath(__file__)), fileName)
+
+class TestRedistance(KratosUnittest.TestCase):
```
rubenzorrilla: The test is not added to any Kratos suite so it will be never executed. Besides the class is named as the variational redistance test one.

kratos/tests/test_flux_based_redistance.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def _ExpectedDistance(self,x,y,z):
+        d = x
+        if( d > 0.2):
+            d = 0.2
+        if( d < -0.2):
+            d = -0.2
+        return x
+        #return -(math.sqrt(x**2+y**2+z**2) - 0.4)
```
rubenzorrilla: Is this correct?

kratos/tests/test_flux_based_redistance.py
```diff
@@ -0,0 +1,200 @@
+import KratosMultiphysics.KratosUnittest as KratosUnittest
+import KratosMultiphysics
+from KratosMultiphysics import python_linear_solver_factory as linear_solver_factory
+import KratosMultiphysics.vtk_output_process as vtk_output_process
+import math
+import os
+
+def GetFilePath(fileName):
+    return os.path.join(os.path.dirname(os.path.realpath(__file__)), fileName)
```
rubenzorrilla, line 9: I remember there was some util that does exactly this, but I didn't manage to find it. @philbucher do you remember where it was placed?

kratos/processes/flux_based_redistance_process.h
```diff
@@ -0,0 +1,649 @@
...
+//                   Multi-Physics
+//
+//  License:		 BSD License
+//			 Kratos default license: kratos/license.txt
+//
+//  Main authors:    Daniel Diez, Pablo Becker
+//
+
+#if !defined(KRATOS_POTENTIAL_FILLING_UTILITY_INCLUDED)
+#define KRATOS_POTENTIAL_FILLING_UTILITY_INCLUDED
```
rubenzorrilla:
```suggestion
#pragma once
```

kratos/processes/flux_based_redistance_process.h
```diff
@@ -0,0 +1,649 @@
...
+///@name Input and output
+///@{
+
+/// Input stream function
+
+///@}
+
+} // namespace Kratos.
+
+#endif //
```
rubenzorrilla:
```suggestion
```

kratos/processes/flux_based_redistance_process.h
```diff
@@ -0,0 +1,649 @@
...
+#include <iostream>
+#include <algorithm>
+
+// External includes
+
+// Project includes
+#include "includes/define.h"
+#include "includes/kratos_flags.h"
+#include "elements/distance_calculation_flux_based_element.h"
+#include "geometries/geometry_data.h"
```
rubenzorrilla:
```suggestion
```
I think this shouldn't be required as it is already included by the geometry.

kratos/processes/flux_based_redistance_process.h
```diff
@@ -0,0 +1,649 @@
...
+		} else if (TDim == 3) {
+			KRATOS_ERROR_IF(rBaseModelPart.ElementsBegin()->GetGeometry().GetGeometryFamily() != GeometryData::KratosGeometryFamily::Kratos_Tetrahedra) << "In 3D the element type is expected to be a tetrahedra" << std::endl;
+		}
+		*/
+
+	}
+
+	void CreateSolvingStrategy(typename TLinearSolver::Pointer pLinearSolver, Parameters &Settings)
+	{
+		//typename SchemeType::Pointer pscheme = Kratos::make_shared<ResidualBasedIncrementalAitkenStaticScheme<TSparseSpace, TDenseSpace>>(1.0);
```
rubenzorrilla:
```suggestion
```

kratos/processes/flux_based_redistance_process.h
```diff
@@ -0,0 +1,649 @@
...
+
+// External includes
+
+// Project includes
+#include "includes/define.h"
+#include "includes/kratos_flags.h"
+#include "elements/distance_calculation_flux_based_element.h"
+#include "geometries/geometry_data.h"
+#include "solving_strategies/schemes/residualbased_incrementalupdate_static_scheme.h"
+#include "solving_strategies/schemes/residualbased_incremental_aitken_static_scheme.h"
```
rubenzorrilla:
```suggestion
```
This is not used.

kratos/processes/flux_based_redistance_process.h
```diff
@@ -0,0 +1,649 @@
+//    |  /           |
```
rubenzorrilla, line 1: There is quite a lot of commented code and formatting issues in this file. I'd acknowledge a bit of clean up for the review. Thanks.

rubenzorrilla (changes requested):
I think this is a very promising contribution. However, there are still some things to be polished or clarified.

Besides, unit testing of the new element is missing.

kratos/processes/flux_based_redistance_process.h
```diff
@@ -0,0 +1,649 @@
...
+///@name Kratos Classes
+///@{
+
+/// Short class definition.
+/**takes a model part full of SIMPLICIAL ELEMENTS (triangles and tetras) and computes a field which resembles the filltime
+	 * This field is to be used to compute the optimization problem of gate location
+	 * 
+	*/
+template <unsigned int TDim, unsigned int TNumNodes, class TSparseSpace, class TDenseSpace, class TLinearSolver>
+class FluxBasedRedistanceProcess : public Process
```
rubenzorrilla: As far as I see, there is some code that could be reused from previous redistance (e.g. the generation of the modelpart or the strategy). Maybe we should consider creating a base class for these.

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: do you want me to create a specific variable for it?

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok, added in the Check

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: you're right, it does not affect the final shape of the solution

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: tbh i don't know, so i always add the casting whenever i have to mix doubles and ints

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: tbh i don't know, so i always add the casting whenever i have to mix doubles and ints

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: i've just removed the correction coefficient.
basically, it is not possible to guarantee grad=1 everywhere (especially when using the diffusion equation to define the direction of the gradient of the distance.) For geometries with holes, it means that from the source (dist=0) there will be diverging and converging "streamlines" . Having grad=1 everywhere will be impossible, since a converging streamline will have an inconsistent value on what the distance should be at that converging point. The best approach i've found is to use the gradient of the diffusion eq as a weight. the higher it is, the shorter the path and hence the closest to the real distance.
so to sum up, this distance method is not exact (neither was the old one), but at least it's a more robust approximation

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: removed

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: removed

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: same as above

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: it seemed to work best.

pablobecker:
wip

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok, moved it to the check

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: no, it's the POTENTIAL_GRADIENT (non historical)

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: in case someone uses this element without the process (i've been actually doing this)

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: yes, i've switched to that one, thanks

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: yes, but in order to compute the potential gradient we must do an explicit assembly, just like the FS elem.
(the FS elem calls Calculate, but i don't like it since it expects an rOutput that is never defined. )

kratos/elements/distance_calculation_flux_based_element.h
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.h
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.h
pablobecker: removed the member var

kratos/elements/distance_calculation_flux_based_element.h
pablobecker: removed

kratos/elements/distance_calculation_flux_based_element.h
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.h
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.h
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.h
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.h
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.h
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.h
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.h
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.h
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.h
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.h
pablobecker: it's used for the simplex geometries

kratos/tests/test_flux_based_redistance.py
pablobecker: oops, forgot that one

kratos/tests/test_flux_based_redistance.py
pablobecker: ok

kratos/tests/test_flux_based_redistance.py
pablobecker: removed commented code, it's just a plane

kratos/processes/flux_based_redistance_process.h
pablobecker: ok

kratos/tests/test_flux_based_redistance.py
pablobecker: ok

kratos/processes/flux_based_redistance_process.h
pablobecker: ok

kratos/processes/flux_based_redistance_process.h
pablobecker: ok

kratos/processes/flux_based_redistance_process.h
pablobecker: for the moment i'll derive the new redistance from the variational one

kratos/elements/distance_calculation_flux_based_element.h
rubenzorrilla: I know. I meant moving this include to the source (cpp) file as it is not used in the header.

kratos/elements/distance_calculation_flux_based_element.cpp
rubenzorrilla: Yes. Maybe `REDISTANCE_STEP` or something of this sort. In my opinion, this is intended to be used in the FS schemes context. Chances are that at some point we end messing up the value if we eventually combine these two. Avoiding introducing more variables to the core is a fair argument too so up to you.

kratos/elements/distance_calculation_flux_based_element.cpp
rubenzorrilla: I'm pretty sure in this case is not required.

kratos/elements/distance_calculation_flux_based_element.cpp
rubenzorrilla: See above.

kratos/elements/distance_calculation_flux_based_element.cpp
rubenzorrilla: It is not required.

kratos/elements/distance_calculation_flux_based_element.cpp
rubenzorrilla: If so I'd put a comment briefly explaining what is this value and that 0.33 seemed to perform better in the experiments.

kratos/elements/distance_calculation_flux_based_element.cpp
rubenzorrilla: Fair enough.

kratos/elements/distance_calculation_flux_based_element.cpp
rubenzorrilla: The point is that this method is somehow the explicit counterpart of the implicit `CalculateAll`. Hence, if in the future someone is to use this formulation with an explicit strategy the RHS contribution would be implemented in here (this is how our explicit strategies work).

Though the useless input argument is fair to me, I think this is a misuse of the `AddExplicitContribution`. Note that this is the same for the OSS projections, which we always compute in the `Calculate`. Same in the FS as you pointed, therefore I'd try to be consistent.

@KratosMultiphysics/technical-committee  Maybe we should consider adding a `Calculate` without output argument for these purposes.

kratos/processes/flux_based_redistance_process.h
rubenzorrilla: OK. In the future we can refactor both to derive from the same base class.

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.h
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: good point, i'll add the new var

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

kratos/processes/flux_based_redistance_process.h
pablobecker: ok

kratos/elements/distance_calculation_flux_based_element.cpp
pablobecker: ok

PR #9953, opened 2022-06-14
for several inlets each number of valid elements must be zeroed indep…

maceligueta:
**📝 Description**
Fixing bug which caused SegFault at inlet.cpp when using more than one inlet. One counter had to be initialized to 0 before using it, and independently for every inlet, not shared for all inlets.


maceligueta:
@jginternational this should fix the issue you reported in our last meetings. Thanks!

salvalatorre (approved):
Looks pretty good...

PR #9959, opened 2022-06-15
[Core] Variational distance - Unset blocked flag exiting process

marcnunezc:
**📝 Description**
After #9891, I introduced a bug when using the variational distance and the MMG process.

The MMG process uses the BLOCKED flag to fix the entities that will not be remeshed. Since the variational distance process now uses this flag, this was causing an unexpected behaviour when using MMG in level set cases.

For this reason, now I add a call to unset the blocked flag when exiting the variational distance process. 

**🆕 Changelog**
E.g.
- Unset BLOCKED flag when exiting Variational distance process.

### DIFF
--- a/kratos/processes/variational_distance_calculation_process.h
+++ b/kratos/processes/variational_distance_calculation_process.h
@@ -344,8 +344,8 @@ public:
 
         // Unfix the distances
         VariableUtils().ApplyFixity(DISTANCE, false, r_distance_model_part.Nodes());
+        VariableUtils().SetFlag(BOUNDARY, false, r_distance_model_part.Nodes());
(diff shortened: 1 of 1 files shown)


rubenzorrilla (dismissed):
Good catch. I'd take the chance to check the initialization of the rest of flags.

marcnunezc:
> Good catch. I'd take the chance to check the initialization of the rest of flags.

True, I also added the unsettling of the BOUNDARY flag that is used in the process.

Now the modelpart does not have any BOUNDARY nor BLOCKED flag after exiting the process.

rubenzorrilla (approved):

PR #9964, opened 2022-06-16
[FastPR][CLaw]Fix Cauchy stress and det(F) in NeoHookean

rubenzorrilla:
**📝 Description**
Two minor fixes in the NeoHookean constitutive model. First one is the calculation of the Cauchy stress, which was done based on the Almansi strain, while the implementation assumes the Green-Lagrange strain to be always passed (check the `GetLawFeatures` method). Besides the Jacobian of the deformation was always computed from the Right Cauchy-Green deformation tensor, although it is already passed in the constitutive law features. Now it is checked if it is already provided and recomputed only if required.

EDIT: We shouldn't be worried about this as the Cauchy is only used in post-processing. Concerning the det(F), it's just a matter of avoiding doing twice the same operation.

### DIFF
--- a/applications/ConstitutiveLawsApplication/custom_constitutive/hyper_elastic_isotropic_neo_hookean_3d.cpp
+++ b/applications/ConstitutiveLawsApplication/custom_constitutive/hyper_elastic_isotropic_neo_hookean_3d.cpp
@@ -90,9 +90,8 @@ void  HyperElasticIsotropicNeoHookean3D::CalculateMaterialResponsePK2(Constituti
     const double poisson_coefficient = material_properties[POISSON_RATIO];
 
     // The deformation gradient
-    const Matrix& deformation_gradient_f = rValues.GetDeformationGradientF();
     double determinant_f = rValues.GetDeterminantF();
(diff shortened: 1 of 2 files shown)


AlejandroCornejo (changes requested):
let me check it pls, I changed some parts to be compatible with rule of mixtures. Which was the main issue? :)

AlejandroCornejo:
as far as I've seen there's no behaviour change in the calculateMatResp right? is only rearranging the Kirchhoff stress computation to be able to use it in the printing of the Cauchy

loumalouomega (approved):

rubenzorrilla:
> as far as I've seen there's no behaviour change in the calculateMatResp right? is only rearranging the Kirchhoff stress computation to be able to use it in the printing of the Cauchy

Indeed it's not only rearranging. Previous implementation assumes the Almansi strain to be passed while the CL features specifies that the Green-Lagrange is passed (see the detailed description above). I checked the result by hand and now it postprocess the correct result.

AlejandroCornejo (approved):
Thank you @rubenzorrilla !!

Issue #9645, opened 2022-02-10
About Initialize, FillBuffer and InitialTime


miguelmaso:
**Description**
Related to #9592 . In the past I experienced some issues with filling the buffer and the initial time, I think it can be interesting for more people.
Briefly, a possible order of the initialize can be:

1. Import the model part
2. (or 3) Set the buffer size
3. (or 2) Set/Get the initial time
4. Apply the initial condition
5. Fill the buffer
6. *Run the solution loop*

**Explanation**

- *Import model part*: obviously, it goes first. If restarted, get initial time is done instead of setting it.
- *Set the buffer size*: this must be done before setting data. Otherwise, the data or initial condition will be lost.
- *Set the initial time*: this must be done before setting the data. Some processes may be time dependent.
- *Set the initial time*: this must be done before `CloneSolutionStep` is called (i.e. the solution loop and filling the buffer) in order to compute the time step (see #9267)
- *Apply the initial condition*: this must be done before filling the buffer, otherwise the initial condition at the previous time steps will be zero
- *Fill the buffer*: following #9592 this is done before running the solution loop

what do you think @KratosMultiphysics/technical-committee @KratosMultiphysics/implementation-committee ?

If this is OK, there will be another discussion about which class does each operation

PR #9966, opened 2022-06-16
[GidOutput] Add GP results and list file for initial state

miguelmaso:
**📝 Description**

IMO this class has some problems related to a bad design, related to #9592 and #9645. This is a quick fix and should be properly addressed in the future.

**🆕 Changelog**
- Print Gauss point results at the initial state
- Add the initial state to the list file


rubenzorrilla:
Makes sense for me. I'd approve it as it shouldn't break something right?

miguelmaso:
> it shouldn't break something right?

It doesn't break anything and it works for me

rubenzorrilla (approved):
Go ahead!

PR #9976, opened 2022-06-20
[Core] Missing float in VTK output

### LINKED ISSUE #9969: [core][VTK] numbers <e-40 cause issues in ascii format
**Description**

Scalar variables in kratos are double. This means that the  smallest number is E-307. However, when the vtk files are created , the scalar ( and vectorial) variables are labeled as type float (see https://github.com/KratosMultiphysics/Kratos/blob/master/kratos/input_output/vtk_output.cpp#L937 ) . This means that the smallest number supported is E-38  . When a vectorial var that contains values lower than E-38 is written in ascii format, Paraview is unable to read these numbers outside the float range and throws an error.

I did not directly create a PR to fix this because I assume there were reasons to declare these variables as float ( RAM maybe?). So my proposal would be to cast the numbers to float before printing them. (note that this happens also for scalar vars, but for some reason paraview reads them without issues. Nodal positions are also affected.

**Scope**

Kratos Core , VTK output with ASCII mode in VECTOR variables (e.g. VELOCITY)

**To Reproduce**

In the VELOCITY variable, assign values smaller than 1e-38 and try to read them in Paraview. An error will be returned.

**Expected behavior**

Paraview should open the file correctly, results are good.

**Environment**
- OS: Windows

loumalouomega:
**📝 Description**
Fixes #9969

**🆕 Changelog**
- Adding missing float in ASCII VTK output


loumalouomega:
> Probably all tests that were using that format need to be updated :S

Probably...but I assumed that the values were correct, because otherwise should not be possible to open the files

roigcarlo:
I think the problem is that now they print more decimals, results are ok but reference files have less numbers

loumalouomega:
> I think the problem is that now they print more decimals, results are ok but reference files have less numbers

It would be less, we are casting a doubles as a float

roigcarlo (approved):

PR #9984, opened 2022-06-24
[RANS] GCC 12 Bugfix

sunethwarna:
**📝 Description**
Make `RANSApplication` compatible with gcc 12 compiler.

**🆕 Changelog**
- bugfix for elements


matekelemen (commented):
Fine by me. GCC bug or is it intentional?

matekelemen (approved):
Fine by me. GCC bug or is it intentional?

matekelemen:
It's a bug, and it's fixed on GCC trunk => will probably be released with GCC 12.2:
- [stackoverflow question](https://stackoverflow.com/questions/72280185/is-this-gcc-12-1-const-problem-a-bug-or-feature-attempts-to-call-non-const-fun)
- [Bugzilla ticket](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105637)

Issue #7928, opened 2020-12-04
usage of c++11 thread_local and openmp


RiccardoRossi:
**Description**
In principle there is no guarantee that thread_local is compatible with c++11, however the following code seems to work reliably with recent gcc and clang

can anyone try this with msvc or intel?

~~~cpp
#include <vector>
#include <iostream>
#include <chrono>
#include <cassert>

#include "omp.h"

class tls{
    public:
    std::vector<double> vec;
};

class A
{
    public:

    static tls& GetTLS()
    {
        thread_local static tls my_tls; //----> note that here it is defined as thread_local!
        return my_tls;
    }

    static std::size_t GetTlsSize()
    {
        return GetTLS().vec.size();
    }
};

int main() {
    
    std::cout << " n threads = " << omp_get_max_threads() << std::endl;    

    A a;
    unsigned int tot=0;
    #pragma omp parallel
    {
        a.GetTLS().vec.resize(omp_get_thread_num());
    }

    std::cout << a.GetTlsSize() << std::endl;

    #pragma omp parallel
    {    
        assert(a.GetTlsSize()==omp_get_thread_num());
        
        #pragma omp atomic 
        tot += a.GetTLS().vec.size();
    }

    int reference_tot = 0;
    for(unsigned int i=0;i<omp_get_max_threads(); ++i)
        reference_tot+=i;
    assert(tot==reference_tot);
    std::cout << tot << std::endl;

    return 0;
}
~~~

philbucher:
with Intel compiler it seems to work, I tested 1-20 threads
(used to compile: `icpc threadlocal_test.cpp -fopenmp`)

philbucher:
@pooyan-dadvand can you also post the configuration that you tested this with?

pooyan-dadvand:
I have tested it with MSVC 2017 x64: cl test_thread_local.cpp /openmp

and running several times with 8 threads works fine:

```
>test_thread_local.exe
 n threads = 8
0
28
``

PR #9985, opened 2022-06-24
[Core] Thread Local Storage for external methods

sunethwarna:
**📝 Description**
This adds the templated class `SMPStorage`. All of the methods in `parallel_utilities.h` has the ability to use thread local storage. But in the case where this parallelized block (ex: `block_for_each`) is initiated in one place and it calls to some method in a different place which is external to the parallelization block (ex: as in `Scheme::CalculateLocalSystem`), then it is not possible to have a thread local storage for the calling method. This class acts as a wrapper thread local storage for such cases. An example is also added to illustrate the use case. 

`SMPStorage` class should work with both `OpenMP` parallelism and `C++` SMP parallelism.

The snippet of the use case is as follows:
```c++
    class Scheme
    {
    public:
        void CalculateLocalSystem(int& LHS, int ProcessInfo) {
            auto& tls = mSMPStorage.GetThreadLocalStorage();
            tls.b = ProcessInfo * 3;
            LHS = tls.b;
        }
    private:
        struct TLS
        {
            int b;
        };

        SMPStorage<TLS> mSMPStorage;
    };

    Scheme scheme;

    int n = 1e+4;
    const int result = IndexPartition<int>(n).for_each<SumReduction<int>>([&](const int ProcessInfo) -> int {
        int lhs;
        scheme.CalculateLocalSystem(lhs, ProcessInfo);
        return lhs;
    });
```

**🆕 Changelog**
- Adds `SMPStorage` class
- Adds tests

### DIFF
--- a/kratos/tests/cpp_tests/utilities/test_smp_storage.cpp
+++ b/kratos/tests/cpp_tests/utilities/test_smp_storage.cpp
@@ -1,0 +1,57 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
+//
+//  Main authors:    Suneth Warnakulasuriya
+//
+
+// System includes
+#include <vector>
+
+// Project includes
+#include "utilities/reduction_utilities.h"
+#include "utilities/parallel_utilities.h"
+#include "utilities/smp_storage.h"
+#include "testing/testing.h"
+
+namespace Kratos {
+namespace Testing {
+
+KRATOS_TEST_CASE_IN_SUITE(SMPStorage, KratosCoreFastSuite)
+{
+    class Scheme
+    {
+    public:
+        void CalculateLocalSystem(int& LHS, int ProcessInfo) {
+            auto& tls = mSMPStorage.GetThreadLocalStorage();
+            tls.b = ProcessInfo * 3;
+            LHS = tls.b;
+        }
+    private:
+        struct TLS
+        {
+            int b;
+        };
+
+        SMPStorage<TLS> mSMPStorage;
+    };
+
+    Scheme scheme;
+
+    int n = 1e+4;
+    const int result = IndexPartition<int>(n).for_each<SumReduction<int>>([&](const int ProcessInfo) -> int {
+        int lhs;
+        scheme.CalculateLocalSystem(lhs, ProcessInfo);
+        return lhs;
+    });
+
+    KRATOS_CHECK_EQUAL(result, n * (n-1) * 3 / 2);
+}
+
+}
+}
--- a/kratos/utilities/smp_storage.h
+++ b/kratos/utilities/smp_storage.h
@@ -1,0 +1,100 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
+//
+//  Main authors:    Suneth Warnakulasuriya
+//
+
+#if !defined(KRATOS_SMP_STORAGE_H_INCLUDED )
+#define  KRATOS_SMP_STORAGE_H_INCLUDED
+
+// System includes
+#ifdef KRATOS_SMP_OPENMP
+#include <vector>
+#include "utilities/openmp_utils.h"
+#endif
+
+// Project includes
+#include "includes/define.h"
+#include "utilities/parallel_utilities.h"
+
+namespace Kratos
+{
+///@addtogroup KratosCore
+///@{
+
+///@name Kratos Classes
+///@{
+
+template<class TStorageType>
+class SMPStorage
+{
+public:
+    ///@name Type Definitions
+    ///@{
+
+    ///@}
+    ///@name Life Cycle
+    ///@{
+
+    /// Default constructor.
+    SMPStorage(){
+        #ifdef KRATOS_SMP_OPENMP
+            mThreadLocalStorageContainer.resize(ParallelUtilities::GetNumThreads());
+        #endif
+    }
+
+    /// Destructor.
+    ~SMPStorage() = default;
+
+    /// Deleted copy constructor
+    SMPStorage(SMPStorage const& rOther) = delete;
+
+    /// Deleted assignment operator
+    SMPStorage& operator=(SMPStorage const& rOther) = delete;
+
+    ///@}
+    ///@name Operations
+    ///@{
+
+    TStorageType& GetThreadLocalStorage()
+    {
+        #ifdef KRATOS_SMP_OPENMP
+            return mThreadLocalStorageContainer[OpenMPUtils::ThisThread()];
+        #elif defined(KRATOS_SMP_CXX11)
+            thread_local TStorageType tls;
+            return tls;
+        #else
+            return mThreadLocalStorageContainer;
+        #endif
+    }
+
+    ///@}
+private:
+    ///@}
+    ///@name Private Members
+    ///@{
+
+#ifdef KRATOS_SMP_OPENMP
+    std::vector<TStorageType> mThreadLocalStorageContainer;
+#elif defined(KRATOS_SMP_CXX11)
+#else
+    TStorageType mThreadLocalStorageContainer;
+#endif
+
+    ///@}
+
+}; // Class SMPStorage
+
+///@}
+
+///@} addtogroup block
+
+}  // namespace Kratos.
+
+#endif // KRATOS_SMP_STORAGE_H_INCLUDED  defined
--- a/kratos/utilities/smp_storage.h
+++ b/kratos/utilities/smp_storage.h
@@ -20,7 +20,6 @@
 #endif
 
 // Project includes
-#include "includes/define.h"
 #include "utilities/parallel_utilities.h"
 
 namespace Kratos
@@ -35,10 +34,6 @@ template<class TStorageType>
 class SMPStorage
 {
 public:
-    ///@name Type Definitions
-    ///@{
-
-    ///@}
     ///@name Life Cycle
     ///@{
 
@@ -93,7 +88,7 @@ private:
 
 ///@}
 
-///@} addtogroup block
+///@} addtogroup KratosCore
 
 }  // namespace Kratos.
 
--- a/kratos/utilities/smp_storage.h
+++ b/kratos/utilities/smp_storage.h
@@ -62,7 +62,7 @@ public:
     ///@name Operations
     ///@{
 
-    TStorageType& GetThreadLocalStorage()
+    inline TStorageType& GetThreadLocalStorage()
     {
         #ifdef KRATOS_SMP_OPENMP
             return mThreadLocalStorageContainer[OpenMPUtils::ThisThread()];


kratos/utilities/smp_storage.h
```diff
@@ -0,0 +1,95 @@
...
+    ///@}
+    ///@name Operations
+    ///@{
+
+    inline TStorageType& GetThreadLocalStorage()
+    {
+        #ifdef KRATOS_SMP_OPENMP
+            return mThreadLocalStorageContainer[OpenMPUtils::ThisThread()];
+        #elif defined(KRATOS_SMP_CXX11)
+            thread_local TStorageType tls;
```
matekelemen, line 65: This is really weird. I know ```thread_local``` modifies the lifetime of ```tls``` but it still looks like you're returning a dangling reference. How about making ```tls``` a proper member variable?
```cpp
#ifdef KRATOS_SMP_OPENMP
    std::vector<TStorageType> mThreadLocalStorageContainer;
#elif defined(KRATOS_SMP_CXX11)
    thread_local static TStorageType tls;
#else
    TStorageType mThreadLocalStorageContainer;
#endif
```
(+ obviously changing the name of ```tls``` to conform to Kratos' style guide)

philbucher: I think thread_local implies static but not sure

Also I wanted to do the same a while ago, but I was quite sure that thread_local does not work for that, but unfortunately I dont remember the reason, so I might be wrong.

@RiccardoRossi also did some tests with it, maybe he remembers?

sunethwarna: As far as i understand, `thread_local` creates static variables when it is called from a thread. So where this is declared matters because, as in the given example, `mSMPStorage` is created when the `Scheme` is created in a serial execution (no threading is allowed for creation of `Scheme`. In the case of @matekelemen suggestion, it will create one `thread_local` variable in the master thread, and it will live untill the termination of the master thread (i.e. termination of the application). Then when this is requested from a parallelized block, all the threads will retireve master thread's thread local storage creating a race condition.

In the case of my suggestion, the `thread_local` variable will be created everytime a parallellized block is executed and also called in parallel creating thread specific local storage for each thread which is the intended behaviour.

I will try to run this through custom made C11 threading for Kratos for testing to check whether the CXX11 threading will work with this. May be if @RiccardoRossi has done something in this regard, that would be very helpfull to know. :)

sunethwarna: Ok I have done the testing. Followings are the observations:

1. defining a member variable as `thread_local` for a class is not allowed. (https://en.cppreference.com/w/cpp/language/storage_duration)
2. The proposed methodology works with C++17. Following is the code block which i used to test.

```c++
#include <vector>
#include <algorithm>
#include <numeric>
#include <execution>
#include <iostream>
#include <mutex>
#include <iostream>

int main()
{
    class Scheme
    {
    public:
        void CalculateLocalSystem(int& LHS, int ProcessInfo) {
            // following line gives race condition as expected.
            // auto& tls = mTLS;

            // following works as a thread_local storage
            thread_local TLS tls;

            tls.b = ProcessInfo * 3;
            LHS = tls.b;
        }
    private:
        struct TLS
        {
            TLS() {
                std::cout << "TLS is created.\n";
            }
            int b;
        };

        TLS mTLS;
    };

    Scheme scheme;

    int v = 0;
    std::mutex m;

    int n = 1e+4;
    std::vector<int> data(n, 0);
    std::iota(data.begin(), data.end(), 0);

    std::for_each(std::execution::par, data.begin(), data.end(), [&](int& ProcessInfo) {
        int lhs;
        scheme.CalculateLocalSystem(lhs, ProcessInfo);
        m.lock();
        v += lhs;
        m.unlock();
    });

    std::cout << "v = " << v << std::endl;
}
```

compiled with `g++  --std=c++17 cxx_11_par.cpp -ltbb`

Any comments would be nice :)

pooyan-dadvand: hmmmm, but this would even work with `TLS` created without `thread_local`. I mean, this is working because `TLS` is now a local variable in the `CalculateLocalSystem`. Isn't it?

sunethwarna: @pooyan-dadvand not exactly, `thread_local` will make a thread local static variable for the context. So, if you run the code snippet, it will create TLS objects for each thread, not for each call. You will see it with the cout clearly.

sunethwarna: `thread_local` implies `static` as well

RiccardoRossi: just to quickly sneak in into this discussion,

in my opinion the way we use thread local stuff in scheme should be avoided. I wonder if we should modify the scheme so that instead of doing in builder and solver

~~~
loop on elements:
      scheme(elem.LeftHandSide)
      Assemble
~~~

we call the builder and solver FROM the scheme so that the loop on elements is in the scheme instead in the builder and solver.

that way each scheme could implement more optimally the tls storage it needs by just rewriting the element loop

RiccardoRossi:
btw, if we were to use the new KratosCSRMAtrix (of its distributed counterpart) than it essentially contains the builder and solver, so this becomes quite easy...

sunethwarna:
Ok, I have checked and followings are the results:

1. I checked for the memory cap for `thread_local`, and it is only bound by the physical memory so the existing method (with `std::vector` and `thread_local` will have the same memory cap.
2. I checked for the performance impact.
    1. `thread_local` can construct the objects in parallel whereas the existing method with `std::vector` we only construct them in serial (which can be done in parallel as well, but at the moment we don't do it). Therefore, it is faster to have `thread_local`. I am intending to change the class to create objects in parallel for `omp` case as well.
    2. I ran a profiling, and the results of it was, `thread_local` was faster. (It took only 23% of the time which was taken by the `omp`, that means `thread_local` calls saved 77% time).
    
I am attaching herewith the script which i used for `omp` which is compiled by using ` g++ --std=c++17 -o omp.out -fopenmp thread_local_storage_omp.cpp`
```c++
#include <vector>
#include <numeric>
#include <iostream>
#include <array>
#include <omp.h>
#include <chrono>

const std::size_t STACK_MAX_SIZE=1;
const std::size_t HEAP_MAX_SIZE=1;

int main()
{
    class Scheme
    {
    public:
        Scheme()
        {
            auto start_time = std::chrono::high_resolution_clock::now();

            const std::size_t max_thread = omp_get_max_threads();
            mTLS.resize(omp_get_max_threads());

            auto tls_resize_time = std::chrono::high_resolution_clock::now();
            std::cout << "TLS resize time:  " << std::chrono::duration_cast<std::chrono::microseconds>(tls_resize_time - start_time).count() << " micro seconds" << std::endl;

            #pragma omp parallel for
            for (std::size_t i = 0; i < max_thread; ++i) {
                mTLS[i].Initialize();
            }

            auto end_time = std::chrono::high_resolution_clock::now();
            std::cout << "Scheme construction time:  " << std::chrono::duration_cast<std::chrono::microseconds>(end_time - start_time).count() << " micro seconds" << std::endl;
        }

        void CalculateLocalSystem(std::size_t& LHS, std::size_t ProcessInfo) {
            // following works as a thread_local storage
            auto& tls = mTLS[omp_get_thread_num()];

            tls.b = ProcessInfo * 3;
            LHS = tls.heap_data.size() + tls.stack_data.size();
            if (tls.heap_data.size() != HEAP_MAX_SIZE)
                std::cout<< tls.heap_data.size() << std::endl;
        }
    private:
        struct TLS
        {
            TLS() {
                std::cout << "TLS is created.\n";
                heap_data.resize(HEAP_MAX_SIZE);

            }

            void Initialize()
            {
                for (std::size_t i = 0; i < HEAP_MAX_SIZE; ++i) {
                    heap_data[i] = i;
                }

                for (std::size_t i = 0; i < STACK_MAX_SIZE; ++i) {
                    stack_data[i] = i;
                }
            }

            int b;
            std::array<double, STACK_MAX_SIZE> stack_data;
            std::vector<double> heap_data;
        };

        std::vector<TLS> mTLS;
    };

    Scheme scheme;

    std::size_t v = 0;

    std::size_t n = 1e+9;
    std::vector<std::size_t> data(n, 0);
    std::iota(data.begin(), data.end(), 0);

    auto start_time = std::chrono::high_resolution_clock::now();

    const std::size_t max_loops = 10;
    std::size_t total_time = 0;

    std::cout << "Starting loops:" << std::endl;
    for (std::size_t loop_counter = 0; loop_counter < max_loops; ++loop_counter) {
        auto local_start_time = std::chrono::high_resolution_clock::now();

        #pragma omp parallel for
        for (std::size_t i = 0; i < data.size(); ++i) {
            std::size_t lhs;
            scheme.CalculateLocalSystem(lhs, data[i]);
            // #pragma omp critical
            // {
            //     v += lhs;
            // }
        }

        auto end_time = std::chrono::high_resolution_clock::now();
        total_time += std::chrono::duration_cast<std::chrono::microseconds>(end_time - local_start_time).count();
        std::cout << loop_counter << " ) Loop time:  " << std::chrono::duration_cast<std::chrono::microseconds>(end_time - local_start_time).count() << " micro seconds" << std::endl;
    }

    std::cout << "Average time for loop: " << total_time / max_loops << " micro seconds" << std::endl;

    std::cout << "         v = " << v << std::endl;
    std::cout << "expected v = " << n * (STACK_MAX_SIZE + HEAP_MAX_SIZE) << std::endl;
}
```

here is the cpp based smp file which is compiled with `g++ --std=c++17 -o cpp.out thread_local_storage_cpp.cpp -ltbb`
```c++
#include <vector>
#include <algorithm>
#include <numeric>
#include <execution>
#include <iostream>
#include <mutex>
#include <iostream>
#include <array>
#include <chrono>

const std::size_t STACK_MAX_SIZE=1;
const std::size_t HEAP_MAX_SIZE=1;

int main()
{
    class Scheme
    {
    public:
        void CalculateLocalSystem(std::size_t& LHS, std::size_t ProcessInfo) {
            // following works as a thread_local storage
            thread_local TLS tls;

            tls.b = ProcessInfo * 3;
            LHS = tls.heap_data.size() + tls.stack_data.size();
            if (tls.heap_data.size() != HEAP_MAX_SIZE)
                std::cout<< tls.heap_data.size() << std::endl;
        }
    private:
        struct TLS
        {
            TLS() {
                std::cout << "TLS is created.\n";
                heap_data.resize(HEAP_MAX_SIZE);
                for (std::size_t i = 0; i < HEAP_MAX_SIZE; ++i) {
                    heap_data[i] = i;
                }

                for (std::size_t i = 0; i < STACK_MAX_SIZE; ++i) {
                    stack_data[i] = i;
                }
            }
            int b;
            std::array<double, STACK_MAX_SIZE> stack_data;
            std::vector<double> heap_data;
        };

        TLS mTLS;
    };

    Scheme scheme;

    std::size_t v = 0;
    std::mutex m;

    std::size_t n = 1e+9;
    std::vector<std::size_t> data(n, 0);
    std::iota(data.begin(), data.end(), 0);

    auto start_time = std::chrono::high_resolution_clock::now();

    const std::size_t max_loops = 10;
    std::size_t total_time = 0;

    std::cout << "Starting loops:" << std::endl;
    for (std::size_t loop_counter = 0; loop_counter < max_loops; ++loop_counter) {
        auto local_start_time = std::chrono::high_resolution_clock::now();

        std::for_each(std::execution::par, data.begin(), data.end(), [&](std::size_t& ProcessInfo) {
            std::size_t lhs;
            scheme.CalculateLocalSystem(lhs, ProcessInfo);
            // m.lock();
            // v += lhs;
            // m.unlock();
        });

        auto end_time = std::chrono::high_resolution_clock::now();
        total_time += std::chrono::duration_cast<std::chrono::microseconds>(end_time - local_start_time).count();
        std::cout << loop_counter << " ) Loop time:  " << std::chrono::duration_cast<std::chrono::microseconds>(end_time - local_start_time).count() << " micro seconds" << std::endl;

    }

    std::cout << "Average time for loop: " << total_time / max_loops << " micro seconds" << std::endl;

    std::cout << "         v = " << v << std::endl;
    std::cout << "expected v = " << n * (STACK_MAX_SIZE + HEAP_MAX_SIZE) << std::endl;
}
```

@RiccardoRossi @pooyan-dadvand @KratosMultiphysics/technical-committee

kratos/utilities/smp_storage.h
```diff
@@ -0,0 +1,95 @@
...
+    ///@}
+private:
+    ///@}
+    ///@name Private Members
+    ///@{
+
+#ifdef KRATOS_SMP_OPENMP
+    std::vector<TStorageType> mThreadLocalStorageContainer;
+#elif defined(KRATOS_SMP_CXX11)
+#else
```
philbucher, line 81:
```suggestion
#elif defined(KRATOS_SMP_NONE)
```

kratos/utilities/smp_storage.h
```diff
@@ -0,0 +1,95 @@
...
+namespace Kratos
+{
+///@addtogroup KratosCore
+///@{
+
+///@name Kratos Classes
+///@{
+
+template<class TStorageType>
+class SMPStorage
```
philbucher, line 34:
```suggestion
class ThreadLocalStorage
```

lets call this what it actually is :)

kratos/utilities/smp_storage.h
```diff
@@ -0,0 +1,95 @@
...
+//                   Multi-Physics
+//
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
+//
+//  Main authors:    Suneth Warnakulasuriya
+//
+
+#if !defined(KRATOS_SMP_STORAGE_H_INCLUDED )
+#define  KRATOS_SMP_STORAGE_H_INCLUDED
```
philbucher, line 14:
```suggestion
#pragma once
```

kratos/tests/cpp_tests/utilities/test_smp_storage.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <vector>
+
+// Project includes
+#include "utilities/reduction_utilities.h"
+#include "utilities/parallel_utilities.h"
+#include "utilities/smp_storage.h"
+#include "testing/testing.h"
+
+namespace Kratos {
+namespace Testing {
```
philbucher, line 23:
```suggestion
namespace Kratos::Testing {
```

welcome to the future :)

kratos/utilities/smp_storage.h
```diff
@@ -0,0 +1,95 @@
...
+
+    /// Default constructor.
+    SMPStorage(){
+        #ifdef KRATOS_SMP_OPENMP
+            mThreadLocalStorageContainer.resize(ParallelUtilities::GetNumThreads());
+        #endif
+    }
+
+    /// Destructor.
+    ~SMPStorage() = default;
```
philbucher, line 48:
```suggestion
```

now necessary, let the compiler do its job :)

kratos/utilities/smp_storage.h
```diff
@@ -0,0 +1,95 @@
...
+class SMPStorage
+{
+public:
+    ///@name Life Cycle
+    ///@{
+
+    /// Default constructor.
+    SMPStorage(){
+        #ifdef KRATOS_SMP_OPENMP
+            mThreadLocalStorageContainer.resize(ParallelUtilities::GetNumThreads());
```
philbucher, line 43: please move this to the source file. This way the parallel-utils include can be moved there too

kratos/utilities/smp_storage.h
```diff
@@ -0,0 +1,95 @@
...
+//
+
+#if !defined(KRATOS_SMP_STORAGE_H_INCLUDED )
+#define  KRATOS_SMP_STORAGE_H_INCLUDED
+
+// System includes
+#ifdef KRATOS_SMP_OPENMP
+#include <vector>
+#include "utilities/openmp_utils.h"
+#endif
```
philbucher, line 20:
```suggestion
// System includes
#include <vector>

// External includes
#ifdef KRATOS_SMP_OPENMP
#include "omp.h"
#endif
```

kratos/utilities/smp_storage.h
```diff
@@ -0,0 +1,95 @@
...
+    SMPStorage& operator=(SMPStorage const& rOther) = delete;
+
+    ///@}
+    ///@name Operations
+    ///@{
+
+    inline TStorageType& GetThreadLocalStorage()
+    {
+        #ifdef KRATOS_SMP_OPENMP
+            return mThreadLocalStorageContainer[OpenMPUtils::ThisThread()];
```
philbucher, line 63: after this PR is merged we can finally deprecate the omp-utils, hence can you please use the omp-call directly?

kratos/utilities/smp_storage.h
```diff
@@ -0,0 +1,95 @@
+//    |  /           |
```
philbucher, line 1: general comment, according to #7928 the ``thread_local` solution would work for even with OMP
Maybe it is not necessary to have the special case for OMP?

And while we are at it, since `thread_local` also works without any shared-memory parallelization, how about only having one case and getting rid of all the `ifdef`s?

philbucher (changes requested):
Overall this is great, I only have some minor comments regarding the implementation

kratos/utilities/smp_storage.h
pooyan-dadvand: I agree!

philbucher:
> * There is a nother solution using the new `MatrixCSR` from the code that would requiere to change the scheme but doesn't require changes to `thread_local`

but this requires to refactor all schemes no?

> The only reason to use this adaptor is the lack of support from openmp of thread_local for non primitives variables

thats why this PR does introduce an OMP-specific solution

TBH I think this was a good PR

roigcarlo:
We decided to refactor the schemes so they use the new CSR matrices as they can already have parallel build and solve. There was nothing wrong with the PR.

If we find another situation with the same problem before we move to C++20 we can reuse this same approach.

philbucher:
Ok thx for the explanation 

What feature is part of c++20 that would help here?

roigcarlo:
None, I express myself superbad.

The problem  is that `thread_local` dos not work with OpenMP, And we will not get rid of that until we move to 20. Also, in this particular case probably is not the best idea to use.

So we would ended up having a feature for patching lack of support for something we will eventually drop and that does not solve the problem in the best possible way.

PR #9994, opened 2022-06-30
[Deprecation message] Remove MeshIndex from GetProperties

miguelmaso:
**📝 Description**
Added message for #9774 

**🆕 Changelog**
- Added message to the *List of upcoming API-breaking and behavior changes* in CMakeLists


rubenzorrilla (approved):

loumalouomega:
When do we plan to remove all MeshesIndex?, in fact, if we remove all we can store in the ModelPart a single mesh instead of a vector of Meshes, which will speed up a bit the access to the mesh. (and we do this all the time, so maybe at the end the speed up could be relevant)

loumalouomega:
> When do we plan to remove all MeshesIndex?, in fact, if we remove all we can store in the ModelPart a single mesh instead of a vector of Meshes, which will speed up a bit the access to the mesh. (and we do this all the time, so maybe at the end the speed up could be relevant)

@KratosMultiphysics/technical-committee I can open an issue to discuss this in deep

rubenzorrilla:
Removing the `MeshIndex` at once is in my opinion a bit risky. Instead, I'd start promoting actions like the one in #9774, so we start removing it at a constant but slow pace. For doing so, it would be nice to have a sort of strategy so maybe it's worth to create such issue to discuss which are the steps.

PR #10003, opened 2022-07-07
updating Centos to use C++14 ready compiler

roigcarlo:
**📝 Description**
Updating centos docker

**🆕 Changelog**
- Centos docker now uses C++14 compiler and python 3.8.10


roigcarlo:
@pooyan-dadvand This should be ready, the new images will only be pushed when we merge this. So I had to disable the centos CI temporaly (deadlock).

Once we have it working I will re enable again

scripts/docker_files/docker_file_ci_centos_7/DockerFile
```diff
@@ -4,28 +4,43 @@ USER root
...
-RUN tar xzf Python-3.5.4.tgz -C /usr/src
-RUN ls /usr/src/Python-3.5.4
-RUN sh /usr/src/Python-3.5.4/configure --enable-optimizations --with-lto --enable-shared LDFLAGS="-Wl,-rpath /usr/local/lib"
+RUN wget https://www.python.org/ftp/python/3.8.10/Python-3.8.10.tgz
+RUN tar xzf Python-3.8.10.tgz -C /usr/src
+RUN ls /usr/src/Python-3.8.10
+RUN sh /usr/src/Python-3.8.10/configure --enable-optimizations --with-lto --enable-shared LDFLAGS="-Wl,-rpath /usr/local/lib"
 RUN make altinstall
-RUN rm Python-3.5.4.tgz
-RUN python3.5 -m pip install 'cx_Freeze==6.1' numpy sympy
```
ddiezrod: you are not installing numpy and sympy? I think we use them in some places in the code

roigcarlo: Oh good catch, I forgot to re-add them for 3.8

pooyan-dadvand (approved):
Thank you! 

Let's merge it so we can make the centos test required again

rubenzorrilla:
Does this mean that we can start using c++14 features?

roigcarlo:
It means Centos Image should support them now (actually it should support C++17)

Issue #6815, opened 2020-04-24
Include another file in json files


pooyan-dadvand:
There are several cases that we need (or it would be better) to include another json file in our project parameters and materials.

pooyan-dadvand:
@KratosMultiphysics/technical-committee we arrived to the conclusion to use define keywords of inclusion:

- @include_json
- @include_csv
- @include_grf

or similar ones for other formats. 

The implementation should be done in a new Jsonfstream which substitutes the included files and give back the complete string to the actual Parameters class.

AriadnaCortesDanes:
The implementation could be done in different ways
* As a list of elements
    "@include_json" : ["file_name1.json", file_name2.json"]
* Unique element
    "@include_json" : "file_name.json"
This case is more simple but does not allow to include more than one json file (due to "duplicate key")

pooyan-dadvand:
I would go for the unique one until we find a use case for the list. @KratosMultiphysics/technical-committee and @KratosMultiphysics/implementation-committee what you think?

rubenzorrilla:
> I would go for the unique one until we find a use case for the list. @KratosMultiphysics/technical-committee and @KratosMultiphysics/implementation-committee what you think?

Agree. Besides, taking into account how we write information in our GUIs, I've the feeling that we'll never need concatenating a list of json files.

roigcarlo:
Closing, this is implemented in #10004

PR #10004, opened 2022-07-07
[Core] Included files in json

AriadnaCortesDanes:
**📝 Description**
Allows json files to include other json files using the keyword "@include_json" (issue #6815). The code adds a post-process to the original json data, and substitutes "@include_json" entries by the contents of the new json file especified. 

Just to illustrate. If we have a first json: 
`{
"data1" : "hello",
"@include_json" : "foo.json",
"data2" : 2.0
}`

and the file foo.json contains:
`{
"data3" :  3.0,
"data4" : 4.0
}`

the final processed data would be equivalent as if we had:  
`{
"data1" : "hello",
"data2" : 2.0,
"data3" :  3.0,
"data4" : 4.0
}`

Note that the function is recursive in the number of includes.

**🆕 Changelog**
- Added a function (and header) in order to read a json file
- Added a function (and header) in order to post-process the json data
- Added corresponding tests


kratos/sources/kratos_parameters.cpp
```diff
@@ -15,6 +15,7 @@
 
 // External includes
 #include "json/json.hpp" // Import nlohmann json library
+#include <stack>
```
loumalouomega: This is not external dependency, it goes in the upper one

kratos/sources/kratos_parameters.cpp
```diff
@@ -209,12 +210,70 @@ Parameters::Parameters()
 /***********************************************************************************/
 /***********************************************************************************/
 
+nlohmann::json Parameters::ReadFile(std::string file_name) {
```
loumalouomega: Input is always CammelCase
```suggestion
nlohmann::json Parameters::ReadFile(std::string FileName) {
```

kratos/sources/kratos_parameters.cpp
```diff
@@ -209,12 +210,70 @@ Parameters::Parameters()
...
 {
     mpRoot = Kratos::make_shared<nlohmann::json>(nlohmann::json::parse( rJsonString, nullptr, true, true));
     mpValue = mpRoot.get();
+    SolveIncludes(*mpValue); 
 }
 
+/***********************************************************************************/
+/***********************************************************************************/
+
+void Parameters::SolveIncludes(nlohmann::json& rJson) {
```
loumalouomega:
```suggestion
void Parameters::SolveIncludes(nlohmann::json& rJson) 
{
```
Style

kratos/includes/kratos_parameters.h
```diff
@@ -273,6 +273,7 @@ class KRATOS_API(KRATOS_CORE) Parameters
 
         ///@}
     };
+    //aqui?
```
loumalouomega:
```suggestion
```
I assume it is a leftover

loumalouomega (changes requested):
Some comments. Implementation simple and elegant

kratos/includes/kratos_parameters.h
```diff
@@ -1023,6 +1023,20 @@ class KRATOS_API(KRATOS_CORE) Parameters
...
+     * @return This method leaves in rJson the final json object with no include dependencies
+     */
+    void SolveIncludes(nlohmann::json& rJson);
+
+    /**
+     * @brief This method read a json file
+     * @param rJson The json file name 
+     * @return The json object obtained from reading the file 
+     */
+    nlohmann::json ReadFile(std::string file_name);
```
loumalouomega:
```suggestion
    nlohmann::json ReadFile(std::string FileName);
```

loumalouomega (dismissed):

loumalouomega (dismissed):

pooyan-dadvand:
I think this needs the approval of the @KratosMultiphysics/technical-committee as a generic core feature

loumalouomega:
New test are failing

loumalouomega (changes requested):
Blocking until @KratosMultiphysics/technical-committee agrees

matekelemen:
Should circular includes be checked or is this something we don't worry about?

roigcarlo:
> Should circular includes be checked or is this something we don't worry about?

We should totally check that if we do not

matekelemen:
> We should totally check that if we do not

I can do that but we should probably get the tests working first. I don't want to commit to this PR without approval so I'll use [another branch](https://github.com/KratosMultiphysics/Kratos/compare/core/included-files-in-json...core/acyclic-json-links?w=1) in the meantime.

> New test are failing

The problem is with the [path to the included json](https://github.com/KratosMultiphysics/Kratos/blob/77f365ce8184ef92b9fc50c0126cbffb9ca82ef2/kratos/tests/cpp_tests/sources/test_kratos_parameters.cpp#L274). Kratos implements tools for changing the working directory during testing for python scripts, so the filesystem can be safely navigated, but there's no such infrastructure for cpp tests.

I vaguely remember someone working on a unified test interface for cpp and python tests, which would be a good place to implement working directory manipulation. Does anyone know anything about the status of that project?

roigcarlo:
> I can do that but we should probably get the tests working first. I don't want to commit to this PR without approval so I'll use [another branch](https://github.com/KratosMultiphysics/Kratos/compare/core/included-files-in-json...core/acyclic-json-links?w=1) in the meantime.

👍 

>  Kratos implements tools for changing the working directory during testing for python scripts, so the filesystem can be safely navigated, but there's no such infrastructure for cpp tests

You should be able to use `Kratos::CurrentWorkingDirectory()` it should be a wrapper for `boost/std::filesystem::current_path` (https://github.com/KratosMultiphysics/Kratos/blob/master/kratos/includes/kratos_filesystem.h) and move from there. Its an external lib now but it will be part of c++17 when we move and we have an abstract interface for it so you will not notice any diference

matekelemen:
Ready for review.

(No idea what's happening on CentOS but I doubt it's related to this PR)

roigcarlo:
@KratosMultiphysics/technical-committee Agrees with the changes, but @pooyan-dadvand will make a final review before approving. Gj!

ddiezrod:
any advance on this?

loumalouomega:
> any advance on this?

I think we are waiting for @pooyan-dadvand final review

pooyan-dadvand:
@mpentek Why you store the `adjacency_list`? Why you are not only checking the stack to see if you have the same entry somewhere there?

matekelemen:
> Why you store the adjacency_list? Why you are not only checking the stack to see if you have the same entry somewhere there?

The stack's purpose is to help navigating the JSON structure, not to store ```@include_json``` entries. It stores nested json objects so it can't be used to check for included file names.

Moreover, having duplicate includes doesn't necessarily mean that there's a cycle in the include graph, which is part of the reason why we need a full graph representation of the include structure. That's what ```adjacency_list``` (or ```rAdjacencyMap```) is for.

pooyan-dadvand:
I understand that we can include the same file in different places but that should not appear in the stack twice. So still checking the filename in the stack should be valid. 

In other word, the stack of the calls would be the same as the adjacency map from the root to the given leaf. So I think we can use it to travers the map without really creating map.

matekelemen:
@pooyan-dadvand ok so you don't mean [this stack](https://github.com/KratosMultiphysics/Kratos/blob/c272bd3acd0e46aaf7ed9ce01ab62ce0831abc64/kratos/sources/kratos_parameters.cpp#L345), but rather your suggestion is to replace [the adjacency map](https://github.com/KratosMultiphysics/Kratos/blob/c272bd3acd0e46aaf7ed9ce01ab62ce0831abc64/kratos/sources/kratos_parameters.cpp#L236) with **a** stack that only contains the current sequence of included files? I can see that working.

kratos/includes/kratos_filesystem.h
```diff
@@ -98,6 +98,16 @@ std::vector<std::string> KRATOS_API(KRATOS_CORE) ListDirectory(const std::string
...
 
+/** 
+ * @brief Resolve symlinks recursively.
+ *  @param rPath: path to a symbolic link.
+ *  @return The result of the recursive dereferencing.
+ *  @throws If the input path does not exist or the symlink is cyclic.
+ *  @note The existence of the final result is not checked and is up to the user to handle.
+ *  @note The input is returned if it is not a symlink.
+ */
+std::string ResolveSymlinks(const std::string& rPath);
```
philbucher: Can you please make a separate PR for this?
I have some comments

pooyan-dadvand:
@matekelemen thanks for the confirmation. I think stack would be a better choice. And also we can keep track of the include file paths in the same stack of the call. But if you prefer a separate stack I would not oppose it.

@AriadnaCortesDanes would you please proceed with this change?

matekelemen:
> I think stack would be a better choice. And also we can keep track of the include file paths in the same stack of the call.

I'm afraid I don't know how to do that :(

> But if you prefer a separate stack I would not oppose it.

@pooyan-dadvand it's already implemented via cbc87dba43ecc1e185641d6288178fa25fb89d1a.

kratos/includes/kratos_parameters.h
```diff
@@ -16,6 +16,8 @@
 #define  KRATOS_KRATOS_PARAMETERS_H_INCLUDED
 
 // System includes
+#include <map>
```
pooyan-dadvand: Do we need this one?

pooyan-dadvand (dismissed):
I think the current stack implementation for the cyclic dependency is ok. So let's go with it!

kratos/includes/kratos_parameters.h
matekelemen: Nope, good catch!

kratos/includes/kratos_parameters.h
```diff
@@ -1036,6 +1034,22 @@ class KRATOS_API(KRATOS_CORE) Parameters
...
     void InternalSetValue(const Parameters& rOtherValue);
 
+    /**
+     * @brief This method solves all the include dependencies in a json file
+     * @param rJson The json object
+     * @param rFileName name of the current json file ("root" if called from the constructor)
+     * @param rIncludeSequence a stack containing the current sequence of included JSON files
+     * @return This method leaves in rJson the final json object with no include dependencies
+     */
+    void SolveIncludes(nlohmann::json& rJson, const std::string& rFileName, std::vector<std::string>& rIncludeSequence);
```
philbucher:
```suggestion
    void SolveIncludes(nlohmann::json& rJson, const std::filesystem::path& rFileName, std::vector<std::string>& rIncludeSequence);
```

kratos/includes/kratos_parameters.h
```diff
@@ -1036,6 +1034,22 @@ class KRATOS_API(KRATOS_CORE) Parameters
...
+
+    /**
+     * @brief This method read a json file
+     * @param rJson The json file name
+     * @return The json object obtained from reading the file
+     */
+    nlohmann::json ReadFile(std::string FileName);
```
philbucher:
```suggestion
    nlohmann::json ReadFile(const std::filesystem::path& rFileName);
```

philbucher (commented):
lets make use of C++17 since we now finally have it :tada:

pooyan-dadvand:
Seems that windows didn't liked the file system. Maybe a merge from master is needed.

loumalouomega:
> Seems that windows didn't liked the file system. Maybe a merge from master is needed.

C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.33.31629\include\iterator(151,27): error C2039: 'insert': is not a member of 'std::filesystem::path' [D:\a\Kratos\Kratos\build\Custom\kratos\KratosCore.vcxproj] 

Looks like this is the problem

matekelemen:
I'm not quite sure what the problem is but I'll only be able to take a look at it during the weekend. I'll let you know when I fixed it.

matekelemen:
@pooyan-dadvand ok looks like MSVC's satisfied too.

### Details

For anyone who later might fall into the same trap with ```string```s and ```path```s, here's what was wrong:

Unlike any sane toolchain, MSVC's implementation of the standard library stores the data of an ```std::filesystem::path``` in a ```std::wstring``` internally instead of a normal ```std::string```. You might not have noticed this, because implicit conversions between STL types are well implemented and have no ADL-related issues since they're all within the same namespace.

However, if a class outside the ```std``` namespace implements ```operator std::string``` but not ```operator std::wstring``` (such as the ```nlohmann::basic_json``` class), the conversion on MSVC will fail because it'd have to cast through the following chain: ```your_class``` => ```std::string``` => ```std::wstring``` => ```std::filesystem::path```. Solution: explicitly cast your type to something in the ```std``` namespace, and then construct an ```std::filesystem::path``` from that. Thank you Microsoft!

philbucher:
:window: :man_facepalming:

pooyan-dadvand (approved):
Thank you all!

pooyan-dadvand:
@loumalouomega would you please check if you can unblock the PR?

loumalouomega (approved):
Approving

loumalouomega:
> @loumalouomega would you please check if you can unblock the PR?

Done

PR #10005, opened 2022-07-07
[ICX] warnings about unsupported and unused options in CMakeLists.txt

marandra:
**📝 Description**
Compilation with new Intel compiler (ICX) generates the following warning for each file built:

```
[1/268] Building CXX object kratos/CMakeFiles/KratosVersion.dir/sources/kratos_version.cpp.o
icx: command line warning #10430: Unsupported command line options encountered
These options as listed are not supported.
For more information, use '-qnextgen-diag'.
option list: 
	-wd654
	-wd10010
clang: warning: -lpthread: 'linker' input unused [-Wunused-command-line-argument]
```

**🆕 Changelog**
- Removed option `lpthread` from CMakeLists.txt when using with Intel compilers
- ~~Removed options -wd654 and -wd1001 from CMakeLists.txt when using with Intel compilers~~
- Disabling of warnings 654 and 10010 adapted to the new ICX compiler
- Removed commented code (with those same options, btw)


marandra:
Seems that at least one check is not passing.
@roigcarlo do you have experience with this issue?

marandra:
The problem seems to be that the newer versions of the Intel compiler (ICX) expect compilation options in a different format than the "Classic" Intel compiler ICC.

roigcarlo (approved):

PR #10008, opened 2022-07-10
[Core] Check JSON include cycles

matekelemen:
### Cycles
Circular includes are checked at each encountered ```"@include_json"``` during parsing in ```Parameters::SolveIncludes```. Include relationships define a directed graph that should be acyclic, so a depth-first search is performed before parsing each included JSON file to detect cycles. The search is naive and thus has terrible complexity, but I'm guessing users won't have lengthy and complicated JSON includes. If we do expect heavy use of this feature, we could use basic memoization during the search (it's not difficult to implement but why add unnecessary complexity).

### Parsing
The inner parsing loop in ```Parameters::SolveIncludes``` is changed to avoid using invalidated iterators.

### Tests
I found it easier to write included JSON files on the fly in C++ tests instead of finding existing test files, but this does mean that these shouldn't be executed during MPI runs (as far as I know, ```KratosCoreFastSuite``` is skipped in MPI so we're safe).

### DIFF
--- a/kratos/includes/kratos_filesystem.h
+++ b/kratos/includes/kratos_filesystem.h
@@ -93,7 +93,7 @@ void KRATOS_API(KRATOS_CORE) MPISafeCreateDirectories(const std::string& rPath);
  *  @param rPath: path to a symbolic link.
  *  @return The result of the recursive dereferencing.
  *  @throws If the input path does not exist or the symlink is cyclic.
- *  @note The existence of the final result is not checked and is up to the user.
+ *  @note The existence of the final result is not checked and is up to the user to handle.
  *  @note The input is returned if it is not a symlink.
  */
 std::string ResolveSymlinks(const std::string& rPath);
--- a/kratos/includes/kratos_filesystem.h
+++ b/kratos/includes/kratos_filesystem.h
@@ -90,7 +90,6 @@ void KRATOS_API(KRATOS_CORE) MPISafeCreateDirectories(const std::string& rPath);
 
 /** 
  * @brief Resolve symlinks recursively.
- *
  *  @param rPath: path to a symbolic link.
  *  @return The result of the recursive dereferencing.
  *  @throws If the input path does not exist or the symlink is cyclic.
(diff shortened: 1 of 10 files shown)


loumalouomega (commented):
Okay from my side

loumalouomega:
> Okay from my side

You don't need approval for merging as it is not the master

kratos/includes/kratos_filesystem.h
```diff
@@ -88,6 +88,16 @@ std::vector<std::string> KRATOS_API(KRATOS_CORE) ListDirectory(const std::string
  */
 void KRATOS_API(KRATOS_CORE) MPISafeCreateDirectories(const std::string& rPath);
 
+/** @brief Resolve symlinks recursively.
```
loumalouomega:
```suggestion
/** 
 * @brief Resolve symlinks recursively.
```

kratos/includes/kratos_filesystem.h
```diff
@@ -88,6 +88,16 @@ std::vector<std::string> KRATOS_API(KRATOS_CORE) ListDirectory(const std::string
...
+ *
```
loumalouomega:
```suggestion
```

AriadnaCortesDanes (commented):
Looks good to me. Also thanks for solving the problem with iterators.

kratos/sources/kratos_parameters.cpp
```diff
@@ -228,52 +231,154 @@ Parameters::Parameters(const std::string& rJsonString)
 {
     mpRoot = Kratos::make_shared<nlohmann::json>(nlohmann::json::parse( rJsonString, nullptr, true, true));
     mpValue = mpRoot.get();
-    SolveIncludes(*mpValue); 
+
+    // Initialize the include graph
+    std::map<std::string,std::set<std::string>> adjacency_list;
```
pooyan-dadvand, line 236: Do we really need this map? The stack itself is not enough?

matekelemen: the ```stack``` is used for parsing the ```JSON```, the ```map``` is needed for keeping track of which file includes which other files, to detect infinite circular includes.

PR #10011, opened 2022-07-11
[PfemFluid] Additional wave monitoring function

AFranci:
This PR adds a new function for monitoring the wave height. It considers the closest node to the gauge and does not need the relative radius as an input variable. It gives a less smooth solution than the previous implementation but it should capture better the wave peaks. It also works in the case of using mesh refinement. Calculate_wave utility has also been modified to avoid issues when mesh refinement is used. In this case, the user can provide the mesh sizes of the zones where the gauges are placed. If the mesh sizes are not provided, the solver computes the mean mesh size of the whole mesh and uses it for computing the wave height (as it was done before).


applications/PfemFluidDynamicsApplication/custom_utilities/calculate_wave_height_utility.cpp
```diff
@@ -46,18 +46,26 @@ CalculateWaveHeightUtility::CalculateWaveHeightUtility(
...
-double CalculateWaveHeightUtility::Calculate(const array_1d<double,3>& rCoordinates) const
+double CalculateWaveHeightUtility::Calculate(const array_1d<double,3>& rCoordinates, const double& rMeshSize) const
 {
     KRATOS_TRY
 
+    double search_radius = 0.0;
+    if (rMeshSize < std::numeric_limits<double>::epsilon()) {
+        search_radius = mRelativeSearchRadius * mMeanElementSize + mSearchRadiusTolerance;
+    } else {
+        search_radius = mRelativeSearchRadius * rMeshSize + mSearchRadiusTolerance;
```
miguelmaso, line 66: Instead of adding a new parameter, you can use the existing parameters. I.e., settings the `search_radius_tolerance` to the desired element size, will produce the same results.

applications/PfemFluidDynamicsApplication/python_scripts/wave_height_output_process.py
```diff
@@ -30,6 +31,7 @@ def GetDefaultParameters(self):
         default_parameters = KM.Parameters("""{
             "model_part_name"           : "",
             "coordinates"               : [[0.0, 0.0, 0.0]],
+            "mesh_size"                 : [0.0],
```
miguelmaso, line 34: please, use a more descriptive name

miguelmaso, line 34: This is breaking the api

applications/PfemFluidDynamicsApplication/custom_utilities/calculate_closest_wave_height_utility.h
```diff
@@ -0,0 +1,169 @@
...
+//    . \  |   (   | |   (   |\__ \.
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics PfemFluidDynamics Application
+//
+//  License:         BSD License
+//    Kratos default license:
+//  kratos/license.txt
+//
+//  Main authors:    Alejandro Cornejo Velazquez
+//                   Miguel Maso Sotomayor
```
miguelmaso, line 12: copy paste error

applications/PfemFluidDynamicsApplication/custom_utilities/calculate_closest_wave_height_utility.cpp
```diff
@@ -0,0 +1,78 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
+//
+//  Main authors:    Miguel Maso Sotomayor
```
miguelmaso, line 10: copy paste error

applications/PfemFluidDynamicsApplication/custom_python/add_custom_utilities_to_python.cpp
```diff
@@ -24,6 +24,7 @@
 #include "custom_utilities/pfem_fluid_gid_io.h"
 #include "custom_utilities/move_mesh_utility.h"
 #include "custom_utilities/calculate_wave_height_utility.h"
+#include "custom_utilities/calculate_closest_wave_height_utility.h"
```
miguelmaso, line 27: which is the difference between the two processes? Probably it is possible to use the previous one without adding more code

applications/PfemFluidDynamicsApplication/custom_utilities/calculate_closest_wave_height_utility.h
```diff
@@ -0,0 +1,169 @@
...
+///@name Kratos Classes
+///@{
+
+/** 
+* @ingroup PfemFluidDynamicsApplication
+* @author Miguel Maso Sotomayor
+* @brief This function computes the wave height at a given point
+* @details The direction is taken from the gravity variable in the ProcessInfo
+*/
+class KRATOS_API(PFEM_FLUID_DYNAMICS_APPLICATION) CalculateClosestWaveHeightUtility
```
miguelmaso, line 38: inherit from CalculateWaveHeightUtility

miguelmaso (changes requested):
I think that the new features can be accomplished with the previous process

AFranci:
I close this PR and I will reopen it when the following main changes have been implemented:
1- unify the two utilities of wave monitoring in only one using a user-provided boolean variable that selects the utility 
2- implement inside the utility a _fast_ way to compute the local mesh size at the gauge zone. This could be done only once.

PR #10014, opened 2022-07-11
Testing shape functions second derivatives

jgonzalezusua:
**📝 Description**
This PR is aimed to test the shape functions second derivatives. The Hessian matrix of the shape functions is computed and compared with the provided from `ShapeFunctionsSecondDerivatives`.

Tags: 
- Testing

**🆕 Changelog**

- Adding `VerifyShapeFunctionsSecondDerivativesValues` function in `geometry_tester.cpp`

### DIFF
--- a/kratos/utilities/geometry_tester.cpp
+++ b/kratos/utilities/geometry_tester.cpp
@@ -215,7 +215,7 @@ bool GeometryTesterUtility::StreamTestTriangle2D3N(
     VerifyStrainExactness( geometry, GeometryData::IntegrationMethod::GI_GAUSS_4, rErrorMessage);
 //         VerifyStrainExactness( geometry, GeometryData::IntegrationMethod::GI_GAUSS_5, rErrorMessage);
 
-    array_1d<double,3> point_in(geometry.Dimension(),1.0/3.0);
+    array_1d<double,2> point_in(2,1.0/3.0);
     if( !VerifyShapeFunctionsSecondDerivativesValues(geometry,point_in,rErrorMessage) ) successful = false;
 
     rErrorMessage << std::endl;
@@ -257,7 +257,7 @@ bool GeometryTesterUtility::StreamTestTriangle2D6N(
     VerifyStrainExactness( geometry, GeometryData::IntegrationMethod::GI_GAUSS_4, rErrorMessage);
 //         VerifyStrainExactness( geometry, GeometryData::IntegrationMethod::GI_GAUSS_5, rErrorMessage);
 
-    array_1d<double,3> point_in(3,1.0/3.0);
+    array_1d<double,2> point_in(2,1.0/3.0);
     if( !VerifyShapeFunctionsSecondDerivativesValues(geometry,point_in,rErrorMessage) ) successful = false;
 
     rErrorMessage << std::endl;
@@ -298,7 +298,7 @@ bool GeometryTesterUtility::StreamTestQuadrilateral2D4N(
     VerifyStrainExactness( geometry, GeometryData::IntegrationMethod::GI_GAUSS_4, rErrorMessage);
 //         VerifyStrainExactness( geometry, GeometryData::IntegrationMethod::GI_GAUSS_5, rErrorMessage);
 
-    array_1d<double,3> point_in(3,1.0/3.0);
+    array_1d<double,2> point_in(2,1.0/3.0);
     if( !VerifyShapeFunctionsSecondDerivativesValues(geometry,point_in,rErrorMessage) ) successful = false;
 
     rErrorMessage << std::endl;
@@ -341,7 +341,7 @@ bool GeometryTesterUtility::StreamTestQuadrilateral2D9N(
     VerifyStrainExactness( geometry, GeometryData::IntegrationMethod::GI_GAUSS_4, rErrorMessage);
 //         VerifyStrainExactness( geometry, GeometryData::IntegrationMethod::GI_GAUSS_5, rErrorMessage);
 
-    array_1d<double,3> point_in(3,1.0/3.0);
+    array_1d<double,2> point_in(2,1.0/3.0);
     if( !VerifyShapeFunctionsSecondDerivativesValues(geometry,point_in,rErrorMessage) ) successful = false;
 
     rErrorMessage << std::endl;
@@ -372,7 +372,7 @@ bool GeometryTesterUtility::StreamTestQuadrilateralInterface2D4N(
         successful=false;
     }
 
-    array_1d<double,3> point_in(3,1.0/3.0);
+    array_1d<double,2> point_in(2,1.0/3.0);
     if( !VerifyShapeFunctionsSecondDerivativesValues(geometry,point_in,rErrorMessage) ) successful = false;
 
     return successful;
@@ -982,7 +982,7 @@ bool GeometryTesterUtility::VerifyShapeFunctionsSecondDerivativesValues(
     Vector ei,ej,f;
 
     double delta = 1e-1;
-    unsigned int dim = rGeometry.WorkingSpaceDimension();
+    unsigned int dim = rGeometry.Dimension();
     rGeometry.ShapeFunctionsSecondDerivatives(DDN_DX,local_coordinates);
 
     if ( H.size() != rGeometry.size() )
--- a/kratos/utilities/geometry_tester.cpp
+++ b/kratos/utilities/geometry_tester.cpp
@@ -217,6 +217,9 @@ bool GeometryTesterUtility::StreamTestTriangle2D3N(
     VerifyStrainExactness( geometry, GeometryData::IntegrationMethod::GI_GAUSS_4, rErrorMessage);
 //         VerifyStrainExactness( geometry, GeometryData::IntegrationMethod::GI_GAUSS_5, rErrorMessage);
 
+    array_1d<double,3> point_in(geometry.Dimension(),1.0/3.0);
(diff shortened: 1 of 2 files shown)


loumalouomega (dismissed):
Noice

jgonzalezusua:
If the error persists, I will close the PR and solving it locally. It is strange since in my computer everything works...

roigcarlo:
Mm seems that its a problem with the size of a vector assignement: `size1 == size2`
let me do a quick look at the code to see if I find something odd.

kratos/utilities/geometry_tester.cpp
```diff
@@ -944,8 +964,74 @@ bool GeometryTesterUtility::VerfiyShapeFunctionsValues(
...
+            ej = ZeroVector(dim);
+            ej[j] = 1.0;
+            std::transform(ei.begin(), ei.end(), ei.begin(), [delta](double &c){ return c*delta; });
+            std::transform(ej.begin(), ej.end(), ej.begin(), [delta](double &c){ return c*delta; });
+            rGeometry.ShapeFunctionsValues(f_1,local_coordinates + ei + ej);
+            rGeometry.ShapeFunctionsValues(f_2,local_coordinates + ei - ej);
+            rGeometry.ShapeFunctionsValues(f_3,local_coordinates - ei + ej);
+            rGeometry.ShapeFunctionsValues(f_4,local_coordinates - ei - ej);
+            f = f_1-f_2-f_3+f_4;
+            transform(f.begin(), f.end(), f.begin(), [delta](double &c){ return c/(4.0*std::pow(delta,2)); });
```
roigcarlo:
```suggestion
           std::transform(f.begin(), f.end(), f.begin(), [delta](double &c){ return c/(4.0*std::pow(delta,2)); });
```

kratos/utilities/geometry_tester.cpp
```diff
@@ -567,6 +584,9 @@ bool GeometryTesterUtility::StreamTestPrism3D6N(
     VerifyStrainExactness( geometry, GeometryData::IntegrationMethod::GI_GAUSS_4, rErrorMessage);
     VerifyStrainExactness( geometry, GeometryData::IntegrationMethod::GI_GAUSS_5, rErrorMessage);
 
+    array_1d<double,3> point_in(3,1.0/3.0);
+    if( !VerifyShapeFunctionsSecondDerivativesValues(geometry,point_in,rErrorMessage) ) successful = false;
```
roigcarlo: This function is causing me to trigger (it works ok in the other calls, is just for this geometry): 

```Bash
    TestGeometryTester Failed with message: 
        Error: Calling base class ShapeFunctionsSecondDerivatives method instead of derived class one. Please check the definition of derived class. 3 dimensional prism with six nodes in 3D space
    Dimension               : 3
    working space dimension : 3
    Local space dimension   : 3

	Point 1	 :  (0, 0, 0)
	Point 2	 :  (0.333333, 0, 0)
	Point 3	 :  (0, 0.333333, 0)
	Point 4	 :  (0, 0, 0.333333)
	Point 5	 :  (0.333333, 0, 0.333333)
	Point 6	 :  (0, 0.333333, 0.333333)
	Center	 :  (0.111111, 0.111111, 0.166667)

    Jacobian in the origin	 : [3,3]((0.333333,0,0),(0,0.333333,0),(0,0,0.333333))

```

roigcarlo (commented):
I've run some tests but I've been unable to reproduce the assign problem, but I've found a couple issues. Not sure if they have anything to do with the results.

kratos/utilities/geometry_tester.cpp
jgonzalezusua: I am not sure if this will work since this modification was included in some commits before. However, I will apply it.

kratos/utilities/geometry_tester.cpp
jgonzalezusua: That is because the prism_3d_6.h does not have the second derivatives implemented, I do not know why this call is in this function. This call has been removed now.

kratos/utilities/geometry_tester.cpp
roigcarlo: To be honest no idea, but it looks to me like if it is getting the `transform` definition from somewhere outside this class (in the CI the unity built is enabled so it may have access to definitions it normally should not)

In any case is worth taking a look just to confirm that is doing what it was suppose to.

jgonzalezusua:
It is still not working in the CI... It is a strange behavior. Might it be due to the boost version?

roigcarlo:
MM doubt it, that check should be version indendent

kratos/utilities/geometry_tester.cpp
```diff
@@ -944,8 +961,74 @@ bool GeometryTesterUtility::VerfiyShapeFunctionsValues(
...
+        DenseVector<Matrix> temp( rGeometry.size() );
+        H.swap( temp );
+    }
+
+
+    for ( unsigned int i = 0; i < rGeometry.size(); ++i ) H[i].resize(dim, dim, false);
+
+    for (unsigned int i = 0; i<dim;i++){
+        for (unsigned int j = 0; j<dim;j++){
+            ei = ZeroVector(dim);
```
matekelemen: Your size mismatches probably originate from here. ```dim``` is a loop variable going from 0 to 2. So you're constructing zero vectors of differing sizes in each iteration.

Just construct the unit vectors here during the iteration instead of always resetting them. This is just a test so you don't need to worry about performance that much.

kratos/utilities/geometry_tester.cpp
```diff
@@ -944,8 +961,74 @@ bool GeometryTesterUtility::VerfiyShapeFunctionsValues(
...
+            ei[i] = 1.0;
+            ej = ZeroVector(dim);
+            ej[j] = 1.0;
+            std::transform(ei.begin(), ei.end(), ei.begin(), [delta](double &c){ return c*delta; });
+            std::transform(ej.begin(), ej.end(), ej.begin(), [delta](double &c){ return c*delta; });
+            rGeometry.ShapeFunctionsValues(f_1,local_coordinates + ei + ej);
```
matekelemen, line 1004: ... and then you're trying to add the 0..2 size vectors to ```local_coordinates```, which is probably of size 3.

kratos/utilities/geometry_tester.cpp
```diff
@@ -944,8 +961,74 @@ bool GeometryTesterUtility::VerfiyShapeFunctionsValues(
...
+            rGeometry.ShapeFunctionsValues(f_2,local_coordinates + ei - ej);
+            rGeometry.ShapeFunctionsValues(f_3,local_coordinates - ei + ej);
+            rGeometry.ShapeFunctionsValues(f_4,local_coordinates - ei - ej);
+            f = f_1-f_2-f_3+f_4;
+            std::transform(f.begin(), f.end(), f.begin(), [delta](double &c){ return c/(4.0*std::pow(delta,2)); });
```
matekelemen: Same, no refs necessary.

kratos/utilities/geometry_tester.cpp
```diff
@@ -944,8 +961,74 @@ bool GeometryTesterUtility::VerfiyShapeFunctionsValues(
 /***********************************************************************************/
 /***********************************************************************************/
 
+bool GeometryTesterUtility::VerifyShapeFunctionsSecondDerivativesValues(
+    GeometryType& rGeometry,
+    GeometryType::CoordinatesArrayType& rGlobalCoordinates,
```
matekelemen, line 966: I know all other test have the same signature, but you don't need mutable references here, right? It'd put my mind at ease if references were immutable unless you wanted to manipulate the objects they point to.

kratos/utilities/geometry_tester.cpp
matekelemen: you don't really need the references here, they can just be plain ```double```s.

matekelemen (commented):
I don't have access to a computer this week so I can't check whether this really is the issue. Please get back to me after handling the suggestions.

kratos/utilities/geometry_tester.cpp
jgonzalezusua: dim is not a loop variable but i and j. dim is always the same

kratos/utilities/geometry_tester.cpp
jgonzalezusua: That may be the point, I will try setting `ej `and `ei` as `ZeroVector` of size 3 instead of size `dim` and see whats happens.

kratos/utilities/geometry_tester.cpp
matekelemen: ah my bad, nevermind then

jgonzalezusua:
Seems like everything works now except from centos, which gives an error related to `KratosSympy`. I think it is not correlated to this PR.

roigcarlo:
You can ignore centos until is fixed, its not required until we merge #10003

kratos/utilities/geometry_tester.cpp
jgonzalezusua: I prefer to maintain the same signature for all tests. There is no point in changing this one if all are implemented in the same way.

jgonzalezusua:
I addressed @matekelemen's comments. If the checks are passed, please approve the PR.

jgonzalezusua:
Ok thanks!

roigcarlo:
It works, now, if there is no other problem it can be approved

matekelemen (approved):

PR #10015, opened 2022-07-11
[FastPR][ConsLaws] Minor fix in NeoHookean postprocess

rubenzorrilla:
**📝 Description**
This adds a minor fix to the Neo-Hookean model (only affects to postprocess as material response is computed in terms of the PK2 stress). Besides I corrected two misleading typos in the documentation.


AlejandroCornejo (approved):
Looks better now hehe thx

PR #10022, opened 2022-07-12
[FastPR][Core] Minor improvement in automatic differentiation utilities

rubenzorrilla:
**📝 Description**
I've experienced some weird things when using the `DoubleContraction` implemented in the `sympy_fe_utilities.py` with symbolic arrays. These are caused by the `rank` function, which is not returning the expected value. Besides, it is extremely slow. Instead, the length of the matrix shape can be used.

### DIFF
--- a/kratos/python_scripts/sympy_fe_utilities.py
+++ b/kratos/python_scripts/sympy_fe_utilities.py
@@ -193,9 +193,10 @@ def DoubleContraction(A,B):
     - A -- Left tensor
     - B -- Right tensor
     """
+    rank_A = len(A.shape)
     AB = sympy.tensorproduct(A,B)
-    tmp = sympy.tensorcontraction(AB, (A.rank()-1,A.rank()+1))
-    output = sympy.tensorcontraction(tmp, (A.rank()-2,A.rank()-1))
+    tmp = sympy.tensorcontraction(AB, (rank_A-1,rank_A+1))
+    output = sympy.tensorcontraction(tmp, (rank_A-2,rank_A-1))
     return output
 
 def MatrixB(DN):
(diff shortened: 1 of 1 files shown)


matekelemen:
um, I'm guessing ```A.rank()``` refers to the [rank of the matrix in a linear algebraic sense](https://en.m.wikipedia.org/wiki/Rank_(linear_algebra)), which can be an intensive task to compute on a large symbolic matrix.

rubenzorrilla:
> um, I'm guessing `A.rank()` refers to the [rank of the matrix in a linear algebraic sense](https://en.m.wikipedia.org/wiki/Rank_(linear_algebra)), which can be an intensive task to compute on a large symbolic matrix.

Yes. Clearly is due to this. Besides in this context it makes no sense to use the rank of the matrix as we only need the sizes to do the contractions from the products of the outer tensors.

kratos/python_scripts/sympy_fe_utilities.py
```diff
@@ -391,6 +392,7 @@ def Compute_RHS(functional, testfunc, do_simplifications=False):
     """
     rhs = sympy.Matrix(sympy.zeros(testfunc.shape[0],1))
     for i in range(testfunc.shape[0]):
+        print(i)
```
sunethwarna:
```suggestion
```

This can be removed ryt?

sunethwarna (changes requested):
This is fine for me except for the print(i)

kratos/python_scripts/sympy_fe_utilities.py
rubenzorrilla: Ouch. Thanks for spotting.

RiccardoRossi (approved):
this is ok to me, with the only comment (also for myself) that this will only work for square tensors

not sure if in the future it makes sense to extend to the case of nonsquare tensors

rubenzorrilla:
@sunethwarna we need you to unblock this.

sunethwarna (approved):
👍

PR #10024, opened 2022-07-13
[Core][CMake] Adding error if Boost is not found

loumalouomega:
**📝 Description**
I just realized that there is not final check to assert if Boost has been found, and this makes the compilation to continue (and fail due to our Boost actual dependency). I added a FATAL_ERROR, but maybe we should move the whole Boost check to a cmake_module in order to avoid the pollution of the main CMake file

**🆕 Changelog**
- Missing error added


CMakeLists.txt
```diff
@@ -381,8 +381,11 @@ find_package(Boost)
 
 if(NOT Boost_FOUND)
   # Trying to find in in external libraries
-	set(BOOST_ROOT "${KRATOS_SOURCE_DIR}/external_libraries/boost")
+  set(BOOST_ROOT "${KRATOS_SOURCE_DIR}/external_libraries/boost")
```
loumalouomega, line 391: I also replaced the tabulation with spaces

roigcarlo (approved):
:heart:

roigcarlo:
Centos is under control (#10003). Boost seems to be related with sourceforge. I will prepare a docker with boost already installed to minimize this sort of problems, but until its ready we will have to keep trying :(

loumalouomega:
@roigcarlo looks like boost is still problematic. If instead of using sourceforge we use a more reliable mirror?

PR #10026, opened 2022-07-13
[Core] fix generic find elements neighbors in mpi

ddiezrod:
**📝 Description**
This process was failing in mpi as some point, (line 83), it was asking the id of the element to make sure it was not the same as the main one.
Of course, this cannot be done in mpi as you cannot access the element if its in another region. However, if the element is in a different region we do not need to check the id, as for sure, it will not be the same element if they are in different cores.

Adding @pablobecker as the change comes mainly from him.

**🆕 Changelog**
- Fix segmentation fault in mpi in the generic process to find elements neighbours


rubenzorrilla (approved):
For me the fix makes perfect sense. Thanks @ddiezrod. In any case, @pablobecker feel free to revoke my approval if you have any further consideration.

RiccardoRossi (commented):
looks ok to me

ddiezrod:
merging, I already talked to @pablobecker about this and he's ok with the changes

PR #10027, opened 2022-07-13
[Shallow] Python scripts cleanup

miguelmaso:
**📝 Description**
The same functionality with less code.

**🆕 Changelog**
- Extended the dam break benchmark
- Removed the dry dam break benchmark, since it is can be defined with a generic dam break
- Fixed default parameters


ipouplana (approved):
Good!

PR #10032, opened 2022-07-14
bug in truss element

Vahid-Galavi:
**📝 Description**
Solved a problem that leads to an internal error in the Intel compiler.

### DIFF
--- a/applications/GeoMechanicsApplication/custom_elements/geo_truss_element_base.cpp
+++ b/applications/GeoMechanicsApplication/custom_elements/geo_truss_element_base.cpp
@@ -213,8 +213,6 @@ template< unsigned int TDim, unsigned int TNumNodes >
 {
     KRATOS_TRY
 
-    static constexpr unsigned int NDof = TDim * TNumNodes;
-
     const double A = GetProperties()[CROSS_AREA];
     const double L = GeoStructuralMechanicsElementUtilities::CalculateReferenceLength3D2N(*this);
     const double rho = GetProperties()[DENSITY];
@@ -706,9 +704,6 @@ void GeoTrussElementBase<3,2>::
 {
     KRATOS_TRY
 
-    static constexpr unsigned int DIM = 3;
-    static constexpr unsigned int NUM_NODES = 2;
-
     Vector current_displacement = ZeroVector(DIM*NUM_NODES);
     GetValuesVector(current_displacement, 0);
     rReferenceCoordinates[0] = GetGeometry()[0].X0() + current_displacement[0];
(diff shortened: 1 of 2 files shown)


applications/GeoMechanicsApplication/custom_elements/geo_truss_element_base.cpp
```diff
@@ -206,34 +207,36 @@ void GeoTrussElementBase<TDim,TNumNodes>::
...
 template< unsigned int TDim, unsigned int TNumNodes >
-void GeoTrussElementBase<TDim,TNumNodes>::
+ void GeoTrussElementBase<TDim,TNumNodes>::
     CalculateConsistentMassMatrix(MatrixType& rMassMatrix,
                                   const ProcessInfo& rCurrentProcessInfo) const
 {
-    KRATOS_TRY;
+    KRATOS_TRY
+
+    static constexpr unsigned int NDof = TDim * TNumNodes;
```
rubenzorrilla: Isn't this unused?

applications/GeoMechanicsApplication/custom_elements/geo_truss_element_base.cpp
```diff
@@ -206,34 +207,36 @@ void GeoTrussElementBase<TDim,TNumNodes>::
...
+
     const double A = GetProperties()[CROSS_AREA];
     const double L = GeoStructuralMechanicsElementUtilities::CalculateReferenceLength3D2N(*this);
     const double rho = GetProperties()[DENSITY];
 
-    const unsigned int factor = double(TDim * TNumNodes);
+    const double factor = double(TDim * TNumNodes);
```
rubenzorrilla:
```suggestion
    const double factor = TDim * TNumNodes;
```
I think the cast is automatically done in this case.

applications/GeoMechanicsApplication/custom_elements/geo_truss_element_base.cpp
Vahid-Galavi: It is used in the same function.

rubenzorrilla: For the future we can define these in the header, so we can use them all throughout the class.

Vahid-Galavi: This variable is only used in this function. That's why I defined it in the function but I can also move it to the header file.

rubenzorrilla: My point is that there is plenty of places in which the `TDim*TNumNodes` is used. In any case I'd leave this for the future (out of the scope of this PR) :wink:

Vahid-Galavi: Anyway, I moved all static constexpr to the header file.

rubenzorrilla (approved):

Vahid-Galavi:
Thanks, @rubenzorrilla for reviewing!

PR #10037, opened 2022-07-15
New windows base image

roigcarlo:
**📝 Description**
A new base image that include boost + pip dependencies so we don't have to download them every time. That should prevent most of the latest error about "boost cannot be downloaded" because Sourceforge...

Also I split the current docker build workflows and created a new one so only the proper image get re-built when changes happen.

**🆕 Changelog**
- Added new default windows building env.
- Split the build workflows for ubuntu and centos
- Added new workflow to build win

### DIFF
--- a/.github/workflows/build_docker_images_for_ci_centos.yml
+++ b/.github/workflows/build_docker_images_for_ci_centos.yml
@@ -7,23 +7,23 @@ on:
     branches:
       - 'master'
     paths:
-      - 'scripts/docker_files/docker_file_ci_windows/*'
-      - '.github/workflows/build_docker_images_for_ci_windows.yml'
+      - 'scripts/docker_files/docker_file_ci_centos_7/*'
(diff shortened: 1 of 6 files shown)


.github/workflows/build_docker_images_for_ci_centos.yml
```diff
@@ -0,0 +1,35 @@
+name: Build Docker Images for CI
```
loumalouomega, line 1: This is Centos, is the title wrong?

loumalouomega (approved):
Okay (title looks confusing, I see things not Windows related)

.github/workflows/build_docker_images_for_ci_centos.yml
roigcarlo: Nop, as I said a also updated the other workflows

roigcarlo:
Nice but we cannot merge because of https://github.com/actions/runner/issues/904. :cry:

PR #10038, opened 2022-07-15
[Core] Adding missing Access methods to PointerVectorSet 

loumalouomega:
**📝 Description**
Adding missing Access methods to PointerVectorSet, making the inteface inconsistent. this is part of afuture and more complex PR.

**🆕 Changelog**
- Added GetMaxBufferSize
- Added GetSortedPartSize/SetSortedPartSize

### DIFF
--- a/kratos/containers/pointer_vector_map.h
+++ b/kratos/containers/pointer_vector_map.h
@@ -486,14 +486,6 @@ public:
     /**
      * @brief Get the maximum size of buffer used in the container.
      */
-    size_type GetMaxBufferSize()
-    {
-        return mMaxBufferSize;
-    }
-
-    /**
-     * @brief Get the maximum size of buffer used in the container (const version)
-     */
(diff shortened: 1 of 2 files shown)


matekelemen:
Why do we need the non-const getters? (I know all other getters have non-const couterparts here, just wondering what the reason behind them is)

rubenzorrilla:
> Why do we need the non-const getters? (I know all other getters have non-const counterparts here, just wondering what the reason behind them is)

Agree. We're returning copies in this case so I think there is no point on adding the non-const getters.

loumalouomega:
> > Why do we need the non-const getters? (I know all other getters have non-const counterparts here, just wondering what the reason behind them is)
> 
> Agree. We're returning copies in this case so I think there is no point on adding the non-const getters.

Agree. Done

loumalouomega:
Ready

rubenzorrilla (approved):

PR #10045, opened 2022-07-19
[MeshingApp] Tetrahedra3D10 refine utility

AriadnaCortesDanes:
**📝 Description**
Utility deriving form "Local_Refine_Tetrahedra_Mesh". Instead of dividing each tetrahedra3D4 of the mesh into other tetrahedra3D4, this utility adds new nodes between all the edges of every tetrahedra3D4 and then replaces it by a Tetrahedra3D10.

**🆕 Changelog**
-New derived class "Tetrahedra10_mesh_converter_utility"
-Corresponding tests


rubenzorrilla (changes requested):
I don't know which is the need for a new class. I think that the same can be achieved by templating these methods in the base class and specializing them for each one of the geometries. Besides this is in my opinion a much clean design, it allows to automatically filter the function to be used. With this design, the user needs to specify which class to call.

pooyan-dadvand:
I think that the confusion comes from the "Refinement" in the name. This utility converts linear tetrahedra to quadratic ones and do not refine! So I would suggest @AriadnaCortesDanes to change the name of the utility.

@rubenzorrilla I am not sure how you want to achieve this by making it template. I think refactoring the base class and separate the part which adds the nodes from the one which creates the elements would be useful but I am not sure if @AriadnaCortesDanes  can do this efficiently.

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
```diff
@@ -0,0 +1,499 @@
...
+///@}
+///@name  Enum's
+///@{
+///@}
+///@name  Functions
+///@{
+///@}
+///@name Kratos Classes
+///@{
+class Tet10RefinementUtility : public LocalRefineTetrahedraMesh
```
pooyan-dadvand: This utility don't refine the mesh. It converts it to quadratic tetrahedra. Also, I would use Tetrahedra instead of Tet

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
```diff
@@ -0,0 +1,499 @@
...
+    ///@}
+    ///@name Operators
+    ///@{
+
+    ///@}
+    ///@name Operations
+    ///@{
+        
+    /**
+    * Refine the mesh locally, call all the commands necessaries to compute the remeshing
```
pooyan-dadvand: Same comment about the naming and what the class does.

rubenzorrilla:
> I think that the confusion comes from the "Refinement" in the name. This utility converts linear tetrahedra to quadratic ones and do not refine! So I would suggest @AriadnaCortesDanes to change the name of the utility.
> 
> @rubenzorrilla I am not sure how you want to achieve this by making it template. I think refactoring the base class and separate the part which adds the nodes from the one which creates the elements would be useful but I am not sure if @AriadnaCortesDanes can do this efficiently.

We can refactor current class so we have a private method with a template that takes the geometry type as argument. I think this would do the job. At the end, current design is only overriding two methods so it should be possible IMO.

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
```diff
@@ -0,0 +1,499 @@
...
+	{
+	    i = -1;
+	}
+
+	for (ElementsArrayType::iterator it = it_begin; it != it_end; ++it)
+	{
+	  Element::GeometryType& geom = it->GetGeometry();
+          CalculateEdges(geom, Coord, edge_ids, aux);
+
+	  create_element = TetrahedraSplit::Split_Tetrahedra(edge_ids, t, &number_elem, &splitted_edges, &internal_node);
```
pooyan-dadvand: Why are you calling the Split_Tetrahedra method?

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
```diff
@@ -0,0 +1,499 @@
...
+
+	  if (create_element == true)
+	  {
+	      to_be_deleted++;
+
+          GlobalPointersVector< Element >& rChildElements = it->GetValue(NEIGHBOUR_ELEMENTS);
```
pooyan-dadvand: Do you need it?

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
```diff
@@ -0,0 +1,499 @@
...
+	}
+
+	for (ElementsArrayType::iterator it = it_begin; it != it_end; ++it)
+	{
+	  Element::GeometryType& geom = it->GetGeometry();
+          CalculateEdges(geom, Coord, edge_ids, aux);
+
+	  create_element = TetrahedraSplit::Split_Tetrahedra(edge_ids, t, &number_elem, &splitted_edges, &internal_node);
+
+	  if (create_element == true)
```
pooyan-dadvand: In your case should be always true. In fact I would send an error if all the edges for this element don't have their corresponding nodes created

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
```diff
@@ -0,0 +1,499 @@
...
+          unsigned int i4 = aux[4];
+		  unsigned int i6 = aux[5];
+		  unsigned int i7 = aux[6];
+		  unsigned int i5 = aux[7];
+          unsigned int i8 = aux[8];
+		  unsigned int i9 = aux[9];
+          
+
+		  Tetrahedra3D10<Node < 3 > > geom(
+		      this_model_part.Nodes()(i0),
```
pooyan-dadvand: you may use GetNode(id) method

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
```diff
@@ -0,0 +1,499 @@
...
+		      this_model_part.Nodes()(i3),
+              this_model_part.Nodes()(i4),
+		      this_model_part.Nodes()(i5),
+		      this_model_part.Nodes()(i6),
+		      this_model_part.Nodes()(i7),
+              this_model_part.Nodes()(i8),
+		      this_model_part.Nodes()(i9)
+		  );
+
+		  // Generate new element by cloning the base one
```
pooyan-dadvand: This part would be better to be done a separate method.

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
```diff
@@ -0,0 +1,499 @@
...
+		  {
+		      InterpolateInteralVariables(number_elem, *it.base(), p_element, rCurrentProcessInfo);
+		  }
+
+		  // Transfer elemental variables
+		  p_element->Data() = it->Data();
+		  p_element->GetValue(SPLIT_ELEMENT) = false;
+		  NewElements.push_back(p_element);
+
+          rChildElements.push_back( Element::WeakPointer(p_element) );
```
pooyan-dadvand: We don't want to keep trace of the old tethrahedra

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
```diff
@@ -0,0 +1,499 @@
...
+
+		  current_id++;
+	      it->SetId(large_id);
```
pooyan-dadvand: This mechanism for deleting can have side effect in other modelparts sharing the elements. I would suggest to use the flag (TO_REMOVE) and standard modelpart erase mechanism

@rubenzorrilla and @RiccardoRossi this is used in the base class also. IMO, to be replaced there also.

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
```diff
@@ -0,0 +1,499 @@
...
+      rElements.Sort();
+
+      /* Now remove all of the "old" elements */
+      rElements.erase(this_model_part.Elements().end() - to_be_deleted, this_model_part.Elements().end());
+
+      std::cout << "NEW NUMBER ELEMENTS: " << rElements.size() << std::endl;
+
+
+      // Now update the elements in SubModelParts
+      if (NewElements.size() > 0)
```
pooyan-dadvand: Would be better to put this part as a separate method.

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
```diff
@@ -0,0 +1,499 @@
...
+	int number_elem = 0;
+	int splitted_edges = 0;
+	int internal_node = 0;
+
+	const ProcessInfo& rCurrentProcessInfo = this_model_part.GetProcessInfo();
+	int edge_ids[6];
+	int t[56];
+	std::vector<int> aux;
+
+        std::cout << "****************** REFINING MESH ******************" << std::endl;
```
pooyan-dadvand: You may use KRATOS_INFO macro and please avoid add such a title like output

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
```diff
@@ -0,0 +1,499 @@
...
+              }
+
+              // Delete old elements
+              iSubModelPart->Elements().Sort();
+              iSubModelPart->Elements().erase(iSubModelPart->Elements().end() - to_be_deleted, iSubModelPart->Elements().end());
+
+              KRATOS_WATCH(iSubModelPart->Info());
+              KRATOS_WATCH(to_be_deleted);
+              KRATOS_WATCH(iSubModelPart->Elements().size());
+              KRATOS_WATCH(this_model_part.Elements().size());
```
pooyan-dadvand: Debugging leftover?

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
```diff
@@ -0,0 +1,499 @@
...
+    * @param Coord: The coordinates of the nodes of the geometry
+    * @return this_model_part: The model part of the model (it is the input too)
+    */
+
+    void EraseOldConditionsAndCreateNew(
+            ModelPart& this_model_part,
+            const compressed_matrix<int>& Coord
+            ) override
+    {
+        KRATOS_TRY;
```
pooyan-dadvand: Most of the comments about the Element can be applied here

Shall we make it template passing the array of Elements or Conditions as input?

applications/MeshingApplication/tests/cpp_tests/test_tet10_refinement_utility.cpp
```diff
@@ -0,0 +1,98 @@
...
+        for(auto cond : modelpart.Conditions()) {
+            GeometryPtrType geom = cond.pGetGeometry();
+            const auto geometryType = geom->GetGeometryType();
+            KRATOS_CHECK_EQUAL(geometryType, GeometryData::KratosGeometryType::Kratos_Triangle3D6);
+
+            auto points = geom->Points();
+            KRATOS_CHECK_EQUAL(Distance(points[0],points[1]), Distance(points[0],points[3]) + Distance(points[3],points[1]) );
+            KRATOS_CHECK_EQUAL(Distance(points[1],points[2]), Distance(points[1],points[4]) + Distance(points[4],points[2]) );
+            KRATOS_CHECK_EQUAL(Distance(points[2],points[0]), Distance(points[2],points[5]) + Distance(points[5],points[0]) );
+        }
```
pooyan-dadvand: I would definitely check the volume of the new element to see if is positive and equal to the original one.
I would also check the volume of the condition and its normal that should be equal to the original one.

pooyan-dadvand (commented):
Nice work @AriadnaCortesDanes but there some concerns to be addressed

pooyan-dadvand:
> > I think that the confusion comes from the "Refinement" in the name. This utility converts linear tetrahedra to quadratic ones and do not refine! So I would suggest @AriadnaCortesDanes to change the name of the utility.
> > @rubenzorrilla I am not sure how you want to achieve this by making it template. I think refactoring the base class and separate the part which adds the nodes from the one which creates the elements would be useful but I am not sure if @AriadnaCortesDanes can do this efficiently.
> 
> We can refactor current class so we have a private method with a template that takes the geometry type as argument. I think this would do the job. At the end, current design is only overriding two methods so it should be possible IMO.

I am not sure if I get your idea. This is not refining the tetrahedra. It just reuses the created nodes in the edges to covert it to the quadratic ones. The only overwritten method is `EraseOldElementAndCreateNewElement`. I think it is very confusing if a template refinement utility refines the mesh for one type and converting it for another type.

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
AriadnaCortesDanes: Ups, just a big leftover!

rubenzorrilla:
> > > I think that the confusion comes from the "Refinement" in the name. This utility converts linear tetrahedra to quadratic ones and do not refine! So I would suggest @AriadnaCortesDanes to change the name of the utility.
> > > @rubenzorrilla I am not sure how you want to achieve this by making it template. I think refactoring the base class and separate the part which adds the nodes from the one which creates the elements would be useful but I am not sure if @AriadnaCortesDanes can do this efficiently.
> > 
> > 
> > We can refactor current class so we have a private method with a template that takes the geometry type as argument. I think this would do the job. At the end, current design is only overriding two methods so it should be possible IMO.
> 
> I am not sure if I get your idea. This is not refining the tetrahedra. It just reuses the created nodes in the edges to covert it to the quadratic ones. The only overwritten method is `EraseOldElementAndCreateNewElement`. I think it is very confusing if a template refinement utility refines the mesh for one type and converting it for another type.

OK. Now I see your point. If this is the case, I think this new feature should be added as an option to current utility, rather than creating a new class.

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
AriadnaCortesDanes: If we template this methods (that are called from the base class for any geometry), the implementation in the base class of the mesh refinement (and all its derived classes) should also be changed according to the new implentation. Furthermore, since we only treat Elements and Conditions, in my opinion  templating this two classes wouldn't make the difference worth the re-implementation.

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
AriadnaCortesDanes: Right!

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
AriadnaCortesDanes: Actually not since the method creates the same number of elements than the original ones

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
AriadnaCortesDanes: Totally agree. Thanks for the suggestion, already updated it in this code.
(PD: i think you meant the TO_ERASE flag 😄)

applications/MeshingApplication/tests/cpp_tests/test_tet10_refinement_utility.cpp
AriadnaCortesDanes: 👍

pooyan-dadvand (changes requested):
Good work! I would suggest to move the implementation to the correponding cpp file to have less compile time dependency between files and minimize the compilation time.

AriadnaCortesDanes:
> Good work! I would suggest to move the implementation to the correponding cpp file to have less compile time dependency between files and minimize the compilation time.

Totally agree 👍

AriadnaCortesDanes:
Ping @RiccardoRossi @rubenzorrilla

AriadnaCortesDanes:
> Good work! I would suggest to move the implementation to the correponding cpp file to have less compile time dependency between files and minimize the compilation time.

As shown here  https://stackoverflow.com/questions/495021/why-can-templates-only-be-implemented-in-the-header-file, the only possible way to implement a template class/method in a .cpp file is to use "crossed" includes, which doesn't correspond to the expected style, and IMO does not look very good.

AriadnaCortesDanes:
> Reference

I tried to move the implementation but, as the methods are template, they needed explicit instantiation, which doesn't convince me.  Personally, I would leave it this way.

AriadnaCortesDanes:
> Good work! I would suggest to move the implementation to the correponding cpp file to have less compile time dependency between files and minimize the compilation time.

I tried to move the implementation but, as the methods are template, they needed explicit instantiation, which doesn't convince me. Personally, I would leave it this way.

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+    ///@name Operators
+    ///@{
+
+    ///@}
+    ///@name Operations
+    ///@{
+        
+    /**
+    * Replaces the geometry of the mesh of Tetrahedra3D4 elements. Resulting is a mesh of Tetrahedra3D10 created
+    * by adding intermediate nodes to each Tetrahedra3D4, with same ids as the originals. Does the same for conditions,
```
pooyan-dadvand: The new nodes don't have originals to have the same id

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+    * replacing the Triangle3D3 by Triangle3D6
+    * @param refine_on_reference: Boolean that defines if refine or not the mesh according to the reference
+    * @param interpolate_internal_variables: Boolean that defines if to interpolate or not the internal variables
+    */
+    void LocalConvertTetrahedra10Mesh(bool refine_on_reference, bool interpolate_internal_variables) {
+            for (auto element : mModelPart.Elements()) element.SetValue(SPLIT_ELEMENT,true);
```
pooyan-dadvand: Please replace the `for `loop with `for_each` from parallel utils. So the use the shared memory parallelization.

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+            for (auto condition : mModelPart.Conditions()) condition.SetValue(SPLIT_ELEMENT,true);
+            LocalRefineMesh(refine_on_reference, interpolate_internal_variables);
+        } 
+  
+    
+protected:
```
pooyan-dadvand: You may delete the protected section as it is not needed here

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+    ///@}
+    ///@name Private Operations
+    ///@{
+    
+    /**
+    * Creates a new tetrahedra3D10
+    * @return the reference to the new tetrahedra
+    */
+    Tetrahedra3D10<Node<3>> GenerateTetrahedra(ModelPart& this_model_part, std::vector<int>& node_ids) {
+        unsigned int i0 = node_ids[0];
```
pooyan-dadvand: here optimizer will take out the local variable, but, good practice is to use `const` for this temporaries.

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+    ///@name Private Operators
+    ///@{
+
+    ///@}
+    ///@name Private Operations
+    ///@{
+    
+    /**
+    * Creates a new tetrahedra3D10
+    * @return the reference to the new tetrahedra
```
pooyan-dadvand: you are returning a copy. (Though, to be precise, a movable copy)

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+            this_model_part.Nodes()(i7),
+            this_model_part.Nodes()(i8),
+            this_model_part.Nodes()(i9)
+        );
+        return geom;
+    }
+
+    /**
+    * Creates a new triangle3D6
+    * @return the reference to the new triangle
```
pooyan-dadvand: Same error here

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+	for (ElementsArrayType::iterator it = it_begin; it != it_end; ++it)
+	{
+        GlobalPointersVector< Element >& rChildElements = it->GetValue(NEIGHBOUR_ELEMENTS);
+        rChildElements.resize(0);
+
+        CalculateEdges(it->GetGeometry(), Coord, edge_ids, node_ids);
+
+        // Generate the new Tetrahedra3D10 element
+        Tetrahedra3D10<Node<3>> geom = GenerateTetrahedra(this_model_part, node_ids);
+        const Element& rElem = KratosComponents<Element>::Get("Element3D10N");
```
pooyan-dadvand: You may ask this component once out of the loop. Please note that `Get` method does a binary search.

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+
+    typedef Node<3> NodeType;
+    typedef Geometry<NodeType> GeometryType;
+    typedef GeometryType::Pointer GeometryPtrType;
+
+    ///@name Life Cycle
+    ///@{
+
+    /// Default constructors
+    Tetrahedra10MeshConverter(ModelPart& model_part) : LocalRefineTetrahedraMesh(model_part)
```
pooyan-dadvand: Styling, method arguments should be CamelCase. It is repeated in many places.

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+    * Replaces the geometry of the mesh of Tetrahedra3D4 elements. Resulting is a mesh of Tetrahedra3D10 created
+    * by adding intermediate nodes to each Tetrahedra3D4, with same ids as the originals. Does the same for conditions,
+    * replacing the Triangle3D3 by Triangle3D6
+    * @param refine_on_reference: Boolean that defines if refine or not the mesh according to the reference
+    * @param interpolate_internal_variables: Boolean that defines if to interpolate or not the internal variables
+    */
+    void LocalConvertTetrahedra10Mesh(bool refine_on_reference, bool interpolate_internal_variables) {
+            for (auto element : mModelPart.Elements()) element.SetValue(SPLIT_ELEMENT,true);
+            for (auto condition : mModelPart.Conditions()) condition.SetValue(SPLIT_ELEMENT,true);
+            LocalRefineMesh(refine_on_reference, interpolate_internal_variables);
```
pooyan-dadvand: You may store the `interpolate_internal_variables` as a member variable to avoid passing it to all methods.

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+        // Transfer elemental variables to new element
+        p_element->Data() = it->Data();
+        p_element->GetValue(SPLIT_ELEMENT) = false;
+        NewElements.push_back(p_element);
+
+        rChildElements.push_back( Element::WeakPointer(p_element) );
+    }
+
+    // Now replace the elements in SubModelParts
+    if ( NewElements.size() > 0 ) ReplaceElementsInSubModelPart(this_model_part);
```
pooyan-dadvand: Here is ok because you have a vector, but in general the `.empty()` is more efficient, especially for lists and trees.

pooyan-dadvand: Styling, one line if and loops are not encouraged.

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+    * @param Coord: The coordinates of the nodes of the geometry
+    * @return this_model_part: The model part of the model (it is the input too)
+    */
+
+    void EraseOldConditionsAndCreateNew(
+            ModelPart& this_model_part,
+            const compressed_matrix<int>& Coord
+            ) override
+    {
+        KRATOS_TRY;
```
pooyan-dadvand: Same comments from the element version applies here

applications/MeshingApplication/tests/cpp_tests/test_tetrahedra10_mesh_converter_utility.cpp
```diff
@@ -0,0 +1,129 @@
...
+#include "testing/testing.h"
+#include "containers/model.h"
+#include "custom_utilities/tetrahedra10_mesh_converter_utility.h"
+
+
+namespace Kratos {
+namespace Testing {
+
+    //Type definitions 
+    typedef Node<3> NodeType;
```
pooyan-dadvand: I would not put global typedefs (even in a cpp) you may use unnamed namespaces

applications/MeshingApplication/tests/cpp_tests/test_tetrahedra10_mesh_converter_utility.cpp
```diff
@@ -0,0 +1,129 @@
...
+    typedef GeometryType::PointsArrayType PointsArrayType;
+
+    //Auxiliar functions
+    /**
+     * @brief Returns the distance between two 3D points.
+     * @param rPoint0 reference to the first point
+     * @param rPoint1 reference an array of 3 coordinates representing the second point
+     * @return Distance 
+     */  
+    static double Distance(const NodeType& Point0, const array_1d<double,3>& Point1) {
```
pooyan-dadvand: Same for global functions. Again unamed namespace would be better

pooyan-dadvand (changes requested):
Nice work! But some minor comments to address.

applications/MeshingApplication/custom_python/add_custom_utilities_to_python.cpp
```diff
@@ -26,6 +26,8 @@
 #include "custom_utilities/local_refine_triangle_mesh.hpp"
 #include "custom_utilities/local_refine_prism_mesh.hpp"
 #include "custom_utilities/local_refine_tetrahedra_mesh.hpp"
+#include "custom_utilities/tetrahedra10_mesh_converter_utility.h"
```
rubenzorrilla:
```suggestion
#include "custom_utilities/simplex_to_quadratic_tetrahedra_mesh_converter_utility.h"
```
I think this is a more descriptive name (same for the class). Note that the other one might be misunderstood with other tetrahedral mesh conversion operations (e.g. octree meshing).

applications/MeshingApplication/custom_utilities/local_refine_geometry_mesh.hpp
```diff
@@ -239,7 +239,7 @@ class KRATOS_API(MESHING_APPLICATION) LocalRefineGeometryMesh
 protected:
     ///@name Protected static Member Variables
     ///@{
-
+        ModelPart& mModelPart;
```
rubenzorrilla: Very minor, but this should be placed in the `Protected member Variables` section below as it is not static. Meaningless right now but at some point we'll fix our Doxygen compilation.

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
+// KRATOS  __  __ _____ ____  _   _ ___ _   _  ____
+//        |  \/  | ____/ ___|| | | |_ _| \ | |/ ___|
+//        | |\/| |  _| \___ \| |_| || ||  \| | |  _
+//        | |  | | |___ ___) |  _  || || |\  | |_| |
+//        |_|  |_|_____|____/|_| |_|___|_| \_|\____| APPLICATION
+//
+//  License:		 BSD License
+//                                       Kratos default license: kratos/license.txt
+//
+//  Main authors:    Ariadna Cortés
```
rubenzorrilla: Header formatting (very minor).

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
rubenzorrilla: [Here](https://github.com/KratosMultiphysics/Kratos/wiki/Style-Guide) you can find the style guide.

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+
+    ///@}
+    ///@name Private Operations
+    ///@{
+    
+    /**
+    * Creates a new tetrahedra3D10
+    * @return the reference to the new tetrahedra
+    */
+    Tetrahedra3D10<Node<3>> GenerateTetrahedra(ModelPart& this_model_part, std::vector<int>& node_ids) {
```
rubenzorrilla:
```suggestion
    Tetrahedra3D10<Node<3>> GenerateTetrahedra(
        ModelPart& rModelPart,
        std::vector<int>& rNodeIds) {
```
(minor) Multiline for more than one argument methods/functions and camel case arguments. Also note the `r` referring to reference in front.

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+            this_model_part.Nodes()(i9)
+        );
+        return geom;
+    }
+
+    /**
+    * Creates a new triangle3D6
+    * @return the reference to the new triangle
+    */
+    Triangle3D6<Node<3>> GenerateTriangle3D6(ModelPart& this_model_part, array_1d<int, 6>& node_ids) {
```
rubenzorrilla: (same style comment).

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+    * @param New_Elements: The new elements created
+    * @param interpolate_internal_variables: A boolean that defines if it is necessary to interpolate the internal variables
+    * @return this_model_part: The model part of the model (it is the input too)
+    */
+
+    void EraseOldElementAndCreateNewElement(
+            ModelPart& this_model_part,
+            const compressed_matrix<int>& Coord,
+            PointerVector< Element >& NewElements,
+            bool interpolate_internal_variables
```
rubenzorrilla: Same style comments. Besides we use 4 spaces for indentation.

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+    ) override
+    {
+	ElementsArrayType& rElements = this_model_part.Elements();
```
rubenzorrilla:
```suggestion
	auto& r_elements = this_model_part.Elements();
```
(also note the style guide suggestion, snake_case for local variables).

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+	ElementsArrayType::iterator it_begin = rElements.ptr_begin();
+	ElementsArrayType::iterator it_end = rElements.ptr_end();
```
rubenzorrilla:
```suggestion
	auto it_begin = rElements.ptr_begin();
	auto it_end = rElements.ptr_end();
```

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+
+	const ProcessInfo& rCurrentProcessInfo = this_model_part.GetProcessInfo();
```
rubenzorrilla:
```suggestion
	const auto& r_current_process_info = this_model_part.GetProcessInfo();
```

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+	int edge_ids[6];
+	std::vector<int> node_ids;
+
+	for (ElementsArrayType::iterator it = it_begin; it != it_end; ++it)
+	{
```
rubenzorrilla: For the sake of code readability, we always indent the loop content.

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+        CalculateEdges(it->GetGeometry(), Coord, edge_ids, node_ids);
+
+        // Generate the new Tetrahedra3D10 element
+        Tetrahedra3D10<Node<3>> geom = GenerateTetrahedra(this_model_part, node_ids);
+        const Element& rElem = KratosComponents<Element>::Get("Element3D10N");
+        Element::Pointer p_element;
+        p_element = rElem.Create(it->Id(), geom, it->pGetProperties());
+        p_element->Initialize(rCurrentProcessInfo);
+        p_element->InitializeSolutionStep(rCurrentProcessInfo);
+        p_element->FinalizeSolutionStep(rCurrentProcessInfo);
```
rubenzorrilla: I assume that you require this in order to do the internal variables interpolation below. However, I'm afraid that in here you might be messing the internal values in constitutive models that are historical.

Shouldn't this be a pure geometrical operation to be done at the beginning of the simulation?

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+    // Now replace the elements in SubModelParts
+    if ( NewElements.size() > 0 ) ReplaceElementsInSubModelPart(this_model_part);
+}
+
+    /**
+    * It replaces the old Tetrahedra3D4 elements in its corresponding submodelpart by its new Tetrahedra3D10
+    * @param this_model_part: The modelpart or submodelpart to replace the elements
+    */
+    void ReplaceElementsInSubModelPart(ModelPart& this_model_part) {
+        for(Element::Pointer& p_element : this_model_part.ElementsArray()){
```
rubenzorrilla:
```suggestion
        for(auto& p_element : this_model_part.ElementsArray()){
```

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+                if( p_element->GetValue(SPLIT_ELEMENT) )
+                {
+                    GlobalPointersVector< Element >& children = p_element->GetValue(NEIGHBOUR_ELEMENTS);
```
rubenzorrilla:
```suggestion
                    auto& r_children = p_element->GetValue(NEIGHBOUR_ELEMENTS);
```

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+
+    /**
+    * Remove the old Trangle3D3 conditions and replace them by the new Trangle3D6 ones
+    * @param Coord: The coordinates of the nodes of the geometry
+    * @return this_model_part: The model part of the model (it is the input too)
+    */
+
+    void EraseOldConditionsAndCreateNew(
+            ModelPart& this_model_part,
+            const compressed_matrix<int>& Coord
```
rubenzorrilla: Can't we use an std::vector instead?

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+    {
+        KRATOS_TRY;
+
+        PointerVector< Condition > NewConditions;
+        ConditionsArrayType& rConditions = this_model_part.Conditions();
+
+        if(rConditions.size() > 0)
+        {
+            ConditionsArrayType::iterator it_begin = rConditions.ptr_begin();
+            ConditionsArrayType::iterator it_end = rConditions.ptr_end();
```
rubenzorrilla:
```suggestion
            auto it_begin = rConditions.ptr_begin();
            auto it_end = rConditions.ptr_end();
```

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+
+    void EraseOldConditionsAndCreateNew(
+            ModelPart& this_model_part,
+            const compressed_matrix<int>& Coord
+            ) override
+    {
+        KRATOS_TRY;
+
+        PointerVector< Condition > NewConditions;
+        ConditionsArrayType& rConditions = this_model_part.Conditions();
```
rubenzorrilla:
```suggestion
        auto& r_conditions = this_model_part.Conditions();
```

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+
+        if(rConditions.size() > 0)
+        {
+            ConditionsArrayType::iterator it_begin = rConditions.ptr_begin();
+            ConditionsArrayType::iterator it_end = rConditions.ptr_end();
+            int  edge_ids[3];
+            array_1d<int, 6> node_ids;
+
+            const ProcessInfo& rCurrentProcessInfo = this_model_part.GetProcessInfo();
```
rubenzorrilla:
```suggestion
            const auto& r_current_processInfo = this_model_part.GetProcessInfo();
```

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+            {
+                CalculateEdgesFaces(it->GetGeometry(), Coord, edge_ids, node_ids);
+
+                GlobalPointersVector< Condition >& rChildConditions = it->GetValue(NEIGHBOUR_CONDITIONS);
+                rChildConditions.resize(0);
+
+                //Generate the new condition
+                Triangle3D6<Node<3> > geom = GenerateTriangle3D6 (this_model_part, node_ids);
+                Condition::Pointer pcond;
+                const Condition& rCond = KratosComponents<Condition>::Get("SurfaceCondition3D6N");
```
rubenzorrilla: This is a costly operation that can be done once outside the loop.

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+
+    /**
+    * It replaces the old Triangle3D3 elements in its corresponding submodelpart by its new Triangle3D6
+    * @param this_model_part: The modelpart or submodelpart to replace the elements
+    */
+    void ReplaceConditionsInSubModelPart(ModelPart& this_model_part) {
+        for(Condition::Pointer& p_cond : this_model_part.ConditionsArray()){
+                if( p_cond->GetValue(SPLIT_ELEMENT) )
```
rubenzorrilla: Please use 4 spaces as indentation.

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
```diff
@@ -0,0 +1,356 @@
...
+        KRATOS_CATCH("");
+    }
+
+
+    /**
+    * It replaces the old Triangle3D3 elements in its corresponding submodelpart by its new Triangle3D6
+    * @param this_model_part: The modelpart or submodelpart to replace the elements
+    */
+    void ReplaceConditionsInSubModelPart(ModelPart& this_model_part) {
+        for(Condition::Pointer& p_cond : this_model_part.ConditionsArray()){
```
rubenzorrilla:
```suggestion
        for(auto& p_cond : this_model_part.ConditionsArray()){
```

applications/MeshingApplication/tests/cpp_tests/test_tetrahedra10_mesh_converter_utility.cpp
```diff
@@ -0,0 +1,129 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:     BSD License
+//           Kratos default license: kratos/license.txt
+//
+//  Main authors:    Ariadna Cortés
```
rubenzorrilla: Header formatting.

applications/MeshingApplication/tests/cpp_tests/test_tetrahedra10_mesh_converter_utility.cpp
```diff
@@ -0,0 +1,129 @@
...
+            const double vol = geom->Volume();
+            KRATOS_CHECK_NEAR(volumes[elem.Id()], vol ,0.001);
+
+            auto points = geom->Points();
+            KRATOS_CHECK_EQUAL(Distance(points[0],points[1]), Distance(points[0],points[4]) + Distance(points[4],points[1]) );
+            KRATOS_CHECK_EQUAL(Distance(points[1],points[2]), Distance(points[1],points[5]) + Distance(points[5],points[2]) );
+            KRATOS_CHECK_EQUAL(Distance(points[2],points[3]), Distance(points[2],points[9]) + Distance(points[9],points[3]) );
+            KRATOS_CHECK_EQUAL(Distance(points[3],points[0]), Distance(points[3],points[7]) + Distance(points[7],points[0]) );
+            KRATOS_CHECK_EQUAL(Distance(points[0],points[2]), Distance(points[0],points[6]) + Distance(points[6],points[0]) );
+            KRATOS_CHECK_EQUAL(Distance(points[1],points[3]), Distance(points[1],points[8]) + Distance(points[8],points[1]) );
```
rubenzorrilla: Considering that `Distance` function does floating point operations, I'd use `KRATOS_CHECK_NEAR` with a very strict tolerance instead.

applications/MeshingApplication/tests/cpp_tests/test_tetrahedra10_mesh_converter_utility.cpp
```diff
@@ -0,0 +1,129 @@
...
+            GeometryPtrType geom = cond.pGetGeometry();
+            const auto geometryType = geom->GetGeometryType();
+            KRATOS_CHECK_EQUAL(geometryType, GeometryData::KratosGeometryType::Kratos_Triangle3D6);
+            const double vol = geom->Area();
+            KRATOS_CHECK_NEAR(volumes[3], vol ,0.001);
+
+            auto points = geom->Points();
+            KRATOS_CHECK_EQUAL(Distance(points[0],points[1]), Distance(points[0],points[3]) + Distance(points[3],points[1]) );
+            KRATOS_CHECK_EQUAL(Distance(points[1],points[2]), Distance(points[1],points[4]) + Distance(points[4],points[2]) );
+            KRATOS_CHECK_EQUAL(Distance(points[2],points[0]), Distance(points[2],points[5]) + Distance(points[5],points[0]) );
```
rubenzorrilla: Same comment.

rubenzorrilla:
> 

I tend to agree @pooyan-dadvand about this. Also note that this is the general rule of the project (we use header-only solution when there is no technical alternative). Admittedly, explicit template instantiation is not nice, but it is what it is...

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
AriadnaCortesDanes: Right thanks!

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
AriadnaCortesDanes: This function is being called from the base class, where a compressed_matrix is used. We could use std::vector but we would have to re-implement the base class and all the subclasses (not very worthy I think).

applications/MeshingApplication/tests/cpp_tests/test_tetrahedra10_mesh_converter_utility.cpp
AriadnaCortesDanes: Totally right

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
AriadnaCortesDanes: I would leave it this way since it think helps to visualize what we are dealing with. Personally, when reading some similar code for the first time, I was really thankful that this code was explicit, so I would prefer to leave it this way. 😄

rubenzorrilla: Totally agree. My point goes in the direction of code maintenance. Imagine that in the future (because reasons) we change the `NEIGHBOUR_ELEMENTS` type, then we'll need to go one by one changing the declarations. Having said so, leave it as you want 😉 .

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
rubenzorrilla: OK. Yes, we can eventually change it in a different PR. The point is to, step by step, get rid of the Boost dependency.

applications/MeshingApplication/custom_utilities/simplex_to_quadratic_tetrahedra_mesh_converter_utility.h
```diff
@@ -156,77 +156,77 @@ class Tetrahedra10MeshConverter : public LocalRefineTetrahedraMesh
...
-        Tetrahedra3D10<Node<3>> geom = GenerateTetrahedra(ThisModelPart, node_ids);
-        Element::Pointer p_element;
-        p_element = rElem.Create(it->Id(), geom, it->pGetProperties());
-        p_element->Initialize(rCurrentProcessInfo);
-        p_element->InitializeSolutionStep(rCurrentProcessInfo);
-        p_element->FinalizeSolutionStep(rCurrentProcessInfo);
-
-        // Setting the internal variables in the "child" elem (the element replacing the old one)
-        if (interpolate_internal_variables == true)
+        auto& rElements = rThisModelPart.Elements();
```
pooyan-dadvand: local variables are snake_case (i.e. r_elements) I would suggest you to take a look at [Kratos Style Guide](https://github.com/KratosMultiphysics/Kratos/wiki/Style-Guide)

applications/MeshingApplication/custom_utilities/simple_to_quadratic_tetrahedra_mesh_converter_utility.h
```diff
@@ -0,0 +1,347 @@
...
+///@}
+///@name  Enum's
+///@{
+///@}
+///@name  Functions
+///@{
+///@}
+///@name Kratos Classes
+///@{
+class Tetrahedra10MeshConverter : public LocalRefineTetrahedraMesh
```
rubenzorrilla:
```suggestion
class SimpleToQuadraticTetrahedraMeshConverter : public LocalRefineTetrahedraMesh
```
Classes must be renamed accordingly. Also the test file.

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
pooyan-dadvand: I can only imagine that this should be run at the beginning of the stage. (as a modeler) but then elements may have constitutive laws and historical value from previous stages. Something to consider in our multi stage discussion.

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
AriadnaCortesDanes: Perfect

applications/MeshingApplication/custom_utilities/tet10_refinement_utility.h
AriadnaCortesDanes: Done :)

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
AriadnaCortesDanes: This would not be very practical, since the class is overriding the base class methods (which pass this variable). If we store it as a member variable and declare new methods that dont need this passing, this methods will no longer be overriding the old ones (but overloading?), so the code could not work in the same way it does right now.

applications/MeshingApplication/custom_utilities/tetrahedra10_mesh_converter_utility.h
rubenzorrilla: I also agree with @pooyan-dadvand 's observation. The unique way that comes to my mind to solve this is to do an L2 to projection and then a mortar mapping between the two meshes (note that we'll require to have both meshes at same time).  

This mechanism is definitively out of the scope of this PR so lets leave this as an observation for the future.

rubenzorrilla:
> > 
> 
> I tend to agree @pooyan-dadvand about this. Also note that this is the general rule of the project (we use header-only solution when there is no technical alternative). Admittedly, explicit template instantiation is not nice, but it is what it is...

There is still no cpp. Aside of this I think this is ready to go in.

AriadnaCortesDanes:
> > > 
> > 
> > 
> > I tend to agree @pooyan-dadvand about this. Also note that this is the general rule of the project (we use header-only solution when there is no technical alternative). Admittedly, explicit template instantiation is not nice, but it is what it is...
> 
> There is still no cpp. Aside of this I think this is ready to go in.

Yes sorry, the method are no longer template (I had no idea why they were in the first place) but I continue to get a strange linking error when moving the implementation to a cpp. Probably I forget some kind of definition or something I don't know about and it crashes.

rubenzorrilla:
> > > > 
> > > 
> > > 
> > > I tend to agree @pooyan-dadvand about this. Also note that this is the general rule of the project (we use header-only solution when there is no technical alternative). Admittedly, explicit template instantiation is not nice, but it is what it is...
> > 
> > 
> > There is still no cpp. Aside of this I think this is ready to go in.
> 
> Yes sorry, the method are no longer template (I had no idea why they were in the first place) but I continue to get a strange linking error when moving the implementation to a cpp. Probably I forget some kind of definition or something I don't know about and it crashes.

From what I see  in the CI it doesn't seem to be a linking error. It seems to me that you have a leftover from when you renamed the class.

AriadnaCortesDanes:
> > > > > 
> > > > 
> > > > 
> > > > I tend to agree @pooyan-dadvand about this. Also note that this is the general rule of the project (we use header-only solution when there is no technical alternative). Admittedly, explicit template instantiation is not nice, but it is what it is...
> > > 
> > > 
> > > There is still no cpp. Aside of this I think this is ready to go in.
> > 
> > 
> > Yes sorry, the method are no longer template (I had no idea why they were in the first place) but I continue to get a strange linking error when moving the implementation to a cpp. Probably I forget some kind of definition or something I don't know about and it crashes.
> 
> From what I see in the CI it doesn't seem to be a linking error. It seems to me that you have a leftover from when you renamed the class.

Besides the rename problem, the linking error was due to a forgotten KRATOS_API. Now it works correcty in the cpp.

rubenzorrilla (approved):
No more issues from my side. I let @pooyan-dadvand to give the final approval.

pooyan-dadvand (approved):
Thanks! :+1:

PR #10053, opened 2022-07-20
[LinearSolversApplication] Update last version of MKL

loumalouomega:
**📝 Description**
Update last version of MKL in application README

**🆕 Changelog**
- Update last version of MKL in application README


roigcarlo (approved):

roigcarlo:
Unrelated, but should we add *.md files as exceptions for the CI and forward a positive results? this will ease the process of editing doc now that is going to be centralized etc...

loumalouomega:
> Unrelated, but should we add *.md files as exceptions for the CI and forward a positive results? this will ease the process of editing doc now that is going to be centralized etc...

Agree

PR #10054, opened 2022-07-20
[CableNetApp] No compression

KlausBSautter:
Hi, with this PR we neglect the element contributions to forces and stiffness if the element is compressed.

Klaus


applications/CableNetApplication/custom_elements/sliding_cable_element_3D.cpp
```diff
@@ -318,6 +318,7 @@ Vector SlidingCableElement3D::GetDirectionVectorNt() const
 
 Vector SlidingCableElement3D::GetInternalForces()
 {
+  const double numerical_limit = std::numeric_limits<double>::epsilon();
```
loumalouomega: This can defined as as constexpr

applications/CableNetApplication/custom_elements/sliding_cable_element_3D.cpp
```diff
@@ -338,9 +339,15 @@ Vector SlidingCableElement3D::GetInternalForces()
   Values.SetStressVector(temp_stress);
   mpConstitutiveLaw->CalculateMaterialResponse(Values,ConstitutiveLaw::StressMeasure_PK2);
 
-
+ 
   double total_internal_force = (temp_stress[0]+prestress) * area * current_length / ref_length;
 
+  mIsCompressed = false;
+  if ((total_internal_force < 0.00)&&(std::abs(current_length-ref_length)>numerical_limit)) {
```
loumalouomega, line 346: Do you knwo what is more probable to be false?, because the one more probable to be false should go first

applications/CableNetApplication/custom_elements/sliding_cable_element_3D.hpp
```diff
@@ -194,6 +194,9 @@ namespace Kratos
 
     private:
 
+        // boolean for the cable --> does not resist to compression
+        bool mIsCompressed;
```
loumalouomega, line 198: Huum, I would use a flag

loumalouomega (commented):
Some comments

applications/CableNetApplication/custom_elements/sliding_cable_element_3D.cpp
KlausBSautter: OK, good to know. It is already the right way around

applications/CableNetApplication/custom_elements/sliding_cable_element_3D.hpp
KlausBSautter: this is consistent with the cable element. We can change this at some point altogether

loumalouomega (dismissed):
Okay

applications/CableNetApplication/custom_elements/sliding_cable_element_3D.cpp
oberbichler: Why `std::numeric_limits<double>::epsilon`?

KlausBSautter: also, consistency with the other elements.

KlausBSautter:
I replaced the test result data. It varied a little bit but you can see that it is basically still the same: 
grafik

I could think of another test in the future, this on is a bit weird....

KlausBSautter:
@loumalouomega

loumalouomega (approved):

PR #10057, opened 2022-07-21
[core] templated divide geometry utility

pablobecker:
**📝 Description**
This PR templates the divide geometry utility so that the input geometry can be of type node<3> or indexednode type.
This allows to divide recursively the geometries returned by the utilty

**🆕 Changelog**
-Modifies divide geometry utility, derived classes and rest of the code that uses them


pablobecker:
ping @rubenzorrilla

rubenzorrilla (approved):
Very nice, thanks for the effort. Indeed, I think this is how I've should have done it when I implemented it.

PR #10060, opened 2022-07-22
Extending the element size calculator to quadratic elements

jgonzalezusua:
**📝 Description**
This PR is aimed to include quadratic elements support to `element_size_calculator.cpp`.

Tags: 
- Core, Features

**🆕 Changelog**
- Triangle2D6, Quadrilateral2D9, Tetrahedra3D10, Hexahedra3D27 in `element_size_calculator.cpp`.
- Added tests for these elements


kratos/tests/cpp_tests/utilities/test_element_size_calculator.cpp
```diff
@@ -138,6 +172,42 @@ namespace Kratos
                 ElementSizeCalculator<2, 4>::AverageElementSizeDerivative, 1e-8, 1e-7);
         }
 
+        KRATOS_TEST_CASE_IN_SUITE(Quadrilateral2D9MinimumElementSizeDerivatives, KratosCoreFastSuite)
+        {
+            Geometry<NodeType>::PointsArrayType nodes;
+            nodes.push_back(NodeType::Pointer(new NodeType(1, 0.0, 0.0, 0.0)));
```
loumalouomega, line 178: Avoid push_back, define the vector with the final size and assign later. Push_back is inefficient. The same for the others

kratos/utilities/element_size_calculator.cpp
```diff
@@ -116,6 +116,105 @@ double ElementSizeCalculator<2,3>::MinimumElementSizeDerivative(
...
+{
+
+    /* Calculate node-edge distances */
+    const double x10 = rGeometry[1].X() - rGeometry[0].X();
+    const double y10 = rGeometry[1].Y() - rGeometry[0].Y();
+    const double x20 = rGeometry[2].X() - rGeometry[0].X();
+    const double y20 = rGeometry[2].Y() - rGeometry[0].Y();
+
+    // node 0, edge 12
+    double nx = -(y20-y10);
```
loumalouomega: const when possible

kratos/utilities/element_size_calculator.cpp
```diff
@@ -116,6 +116,105 @@ double ElementSizeCalculator<2,3>::MinimumElementSizeDerivative(
...
+    const double y10_derivative = EdgeLengthDerivative(DerivativeNodeIndex, DerivativeDirectionIndex, 1, 0, 1);
+
+    const double x20 = rGeometry[2].X() - rGeometry[0].X();
+    const double x20_derivative = EdgeLengthDerivative(DerivativeNodeIndex, DerivativeDirectionIndex, 2, 0, 0);
+
+    const double y20 = rGeometry[2].Y() - rGeometry[0].Y();
+    const double y20_derivative = EdgeLengthDerivative(DerivativeNodeIndex, DerivativeDirectionIndex, 2, 0, 1);
+
+    // node 0, edge 12
+    double nx = -(y20-y10);
```
loumalouomega: const when possible

kratos/utilities/element_size_calculator.cpp
```diff
@@ -383,6 +605,159 @@ double ElementSizeCalculator<3,4>::MinimumElementSizeDerivative(
...
+    const double y20 = rGeometry[2].Y() - rGeometry[0].Y();
+    const double z20 = rGeometry[2].Z() - rGeometry[0].Z();
+
+    const double x30 = rGeometry[3].X() - rGeometry[0].X();
+    const double y30 = rGeometry[3].Y() - rGeometry[0].Y();
+    const double z30 = rGeometry[3].Z() - rGeometry[0].Z();
+
+    // face 123
+    double nx = (y30-y10)*(z20-z10) - (z30-z10)*(y20-y10);
+    double ny = (z30-z10)*(x20-x10) - (x30-x10)*(z20-z10);
```
loumalouomega: Const when possible

loumalouomega (changes requested):
Some comments

kratos/tests/cpp_tests/utilities/test_element_size_calculator.cpp
jgonzalezusua: I just followed the format of the script. The nodes vector is created using this way in all tests.

jgonzalezusua:
I defined the `ProjectedElementSize` functions for quadratic elements but I have not implemented nothing in them. This is something especific from FIC elements, so the extension to quadratic elements should be implemented for FIC developers

kratos/utilities/element_size_calculator.cpp
jgonzalezusua: These variables cannot be `const` since are modified below

kratos/utilities/element_size_calculator.cpp
jgonzalezusua: These variables cannot be `const` since are modified below

kratos/utilities/element_size_calculator.cpp
jgonzalezusua: These variables cannot be `const` since are modified below

loumalouomega: Okay, my main complain was about the push_back

jgonzalezusua: Yeah, I know, but it is a test, so it is run only once. The rest of the element size tests are created in a similar way. Are you sure we will appreciate the enhancement?

kratos/utilities/element_size_calculator.cpp
```diff
@@ -116,6 +116,105 @@ double ElementSizeCalculator<2,3>::MinimumElementSizeDerivative(
     KRATOS_CATCH("");
 }
 
+// Triangle2D6 version.
+template<>
+double ElementSizeCalculator<2,6>::MinimumElementSize(const Geometry<Node<3> >& rGeometry)
```
rubenzorrilla, line 121: As far as I see the implementation of the quadratic triangle is exactly the same as that of the linear one. If so it is, I'd call the other method in here.

kratos/utilities/element_size_calculator.cpp
```diff
@@ -116,6 +116,105 @@ double ElementSizeCalculator<2,3>::MinimumElementSizeDerivative(
...
+}
+// Triangle2D6 version.
+template<>
+double ElementSizeCalculator<2,6>::MinimumElementSizeDerivative(
+    const unsigned int DerivativeNodeIndex,
+    const unsigned int DerivativeDirectionIndex,
+    const Geometry<Node<3> >& rGeometry)
+{
+    KRATOS_TRY
+
```
rubenzorrilla, line 137: I'm not sure if the derivative is the same for quadratic geometries. @sunethwarna can you confirm please.

Besides, if the implementation is the same we must call the linear method.

kratos/utilities/element_size_calculator.cpp
```diff
@@ -190,49 +289,317 @@ double ElementSizeCalculator<2,4>::MinimumElementSizeDerivative(
...
+    return 0.5 * h2_derivative / std::sqrt(h2);
+
+    KRATOS_CATCH("");
+}
+
+// Quadrilateral2D9 version.
+template<>
+double ElementSizeCalculator<2,9>::MinimumElementSize(const Geometry<Node<3> >& rGeometry)
+{
+
```
rubenzorrilla, line 265: Same comment. This implementation is the same than that of the linear quadrilateral. Call the other method in here.

kratos/utilities/element_size_calculator.cpp
```diff
@@ -190,49 +289,317 @@ double ElementSizeCalculator<2,4>::MinimumElementSizeDerivative(
...
+}
+
+template<>
+double ElementSizeCalculator<2,9>::MinimumElementSizeDerivative(
+    const unsigned int DerivativeNodeIndex,
+    const unsigned int DerivativeDirectionIndex,
+    const Geometry<Node<3> >& rGeometry)
+{
+    KRATOS_TRY
+
```
rubenzorrilla: @sunethwarna can you confirm the quadratic derivative follows the same implementation.

kratos/utilities/element_size_calculator.cpp
```diff
@@ -190,49 +289,317 @@ double ElementSizeCalculator<2,4>::MinimumElementSizeDerivative(
...
+    // face 012
+    nx = y10*z20 - z10*y20;
+    ny = z10*x20 - x10*z20;
+    nz = x10*y20 - y10*x20;
+    hsq = x30*nx + y30*ny + z30*nz;
+    hsq *= hsq / (nx*nx + ny*ny + nz*nz);
+    Hsq = (hsq < Hsq) ? hsq : Hsq;
+    return std::sqrt(Hsq);
+}
```
rubenzorrilla: Why all these lines appear as modified? I can't see the changes with old implementation.

rubenzorrilla: OK. Now I see. GitHub got crazy because the implementation is copied and pasted. Please avoid copying and pasting code by calling the linear tetrahedron function.

kratos/utilities/element_size_calculator.cpp
```diff
@@ -282,16 +649,16 @@ double ElementSizeCalculator<3,4>::MinimumElementSize(const Geometry<Node<3> >&
 }
 
 template<>
-double ElementSizeCalculator<3,4>::MinimumElementSizeDerivative(
+double ElementSizeCalculator<3,10>::MinimumElementSizeDerivative(
     const unsigned int DerivativeNodeIndex,
     const unsigned int DerivativeDirectionIndex,
     const Geometry<Node<3> >& rGeometry)
 {
     KRATOS_TRY
```
rubenzorrilla, line 455: Same comments. @sunethwarna can you confirm and avoid copying and pasting.

kratos/utilities/element_size_calculator.cpp
```diff
@@ -636,14 +1003,179 @@ double ElementSizeCalculator<3,8>::MinimumElementSizeDerivative(
...
+    return 0.5 * h2_derivative / std::sqrt(h2);
+
+    KRATOS_CATCH("");
+}
+
+// Hexahedra3D27 version. We use the distance between face centers to compute lengths.
+template<>
+double ElementSizeCalculator<3,27>::MinimumElementSize(const Geometry<Node<3> >& rGeometry)
+{
+
```
rubenzorrilla: Same copying and pasting comment.

kratos/utilities/element_size_calculator.cpp
```diff
@@ -636,14 +1003,179 @@ double ElementSizeCalculator<3,8>::MinimumElementSizeDerivative(
...
+}
+
+template<>
+double ElementSizeCalculator<3,27>::MinimumElementSizeDerivative(
+    const unsigned int DerivativeNodeIndex,
+    const unsigned int DerivativeDirectionIndex,
+    const Geometry<Node<3> >& rGeometry)
+{
+    KRATOS_TRY
+
```
rubenzorrilla: @sunethwarna this as well.

rubenzorrilla (changes requested):
All quadratic geometries implementations are copied and pasted. This can be avoided by calling the corresponding linear geometries methods.

kratos/utilities/element_size_calculator.cpp
```diff
@@ -827,6 +1406,98 @@ double ElementSizeCalculator<3,4>::AverageElementSizeDerivative(
...
+    const double y10 = rGeometry[1].Y() - rGeometry[0].Y();
+    const double z10 = rGeometry[1].Z() - rGeometry[0].Z();
+
+    const double x20 = rGeometry[2].X() - rGeometry[0].X();
+    const double y20 = rGeometry[2].Y() - rGeometry[0].Y();
+    const double z20 = rGeometry[2].Z() - rGeometry[0].Z();
+
+    const double x30 = rGeometry[3].X() - rGeometry[0].X();
+    const double y30 = rGeometry[3].Y() - rGeometry[0].Y();
+    const double z30 = rGeometry[3].Z() - rGeometry[0].Z();
```
matekelemen: [MathUtils](https://github.com/KratosMultiphysics/Kratos/blob/a602d4c431df9a1ead0a881af9bfc4fc95cdbf0a/kratos/utilities/math_utils.h)? We have cross products, dot products, norms, determinants, etc in there too.

rubenzorrilla: These functions were implemented in this closed form manner to avoid allocating any auxiliary matrix/array. Note that these are likely called for each element at each non-linear iteration.

jgonzalezusua:
> All quadratic geometries implementations are copied and pasted. This can be avoided by calling the corresponding linear geometries methods.

Yes, it is true. I think that I can define only the some nodes (in the vertices) of the quadratic elements to run the cpp tests

rubenzorrilla:
> > All quadratic geometries implementations are copied and pasted. This can be avoided by calling the corresponding linear geometries methods.
> 
> Yes, it is true. I think that I can define only the some nodes (in the vertices) of the quadratic elements to run the cpp tests

Exactly. What I'd is to specialize the templates for the quadratic geometries but these only interface the "old" linear geometries methods.

jgonzalezusua:
> > > All quadratic geometries implementations are copied and pasted. This can be avoided by calling the corresponding linear geometries methods.
> > 
> > 
> > Yes, it is true. I think that I can define only the some nodes (in the vertices) of the quadratic elements to run the cpp tests
> 
> Exactly. What I'd do is to specialize the templates for the quadratic geometries but these only interface the "old" linear geometries methods.

I think this will work with the calculation of the element size, but not with `ElementSizeDerivatives` since it checks the number of nodes of the element.

rubenzorrilla:
> > > > All quadratic geometries implementations are copied and pasted. This can be avoided by calling the corresponding linear geometries methods.
> > > 
> > > 
> > > Yes, it is true. I think that I can define only the some nodes (in the vertices) of the quadratic elements to run the cpp tests
> > 
> > 
> > Exactly. What I'd do is to specialize the templates for the quadratic geometries but these only interface the "old" linear geometries methods.
> 
> I think this will work with the calculation of the element size, but not with `ElementSizeDerivatives` since it checks the number of nodes of the element.

About the derivatives I'd like to hear @sunethwarna 's opinion before merging. He is the one that really knows what's going on in there. In any case, you can move the implementation to an auxiliary private method which is called after the check.

kratos/utilities/element_size_calculator.cpp
sunethwarna: Hi, you are correct @rubenzorrilla. Derivatives are not the same. If you look at the following code snippet
https://github.com/KratosMultiphysics/Kratos/blob/f0b86d60271e58b4681800bb2b771596c7567d7c/kratos/utilities/element_size_calculator.cpp#L62-L64

The quadratice element has six nodes, and it only uses first three nodes for the `MinimumElementSize` calculation. It will give correct answers for the first three `DerivativeNodeIndex`, and will crash if the derivative w.r.t. forth or higher node index node is requested. Therefore I would have another templated method calling the `ElementSizeCalculator<2,3>::MinimumElementSizeDerivative` for first three nodes, and return zero for the rest of the nodes and their directions.

I think, if you run the test case in debug, it will crash due to the fact that error is produced only in debug mode.

kratos/utilities/element_size_calculator.cpp
sunethwarna: It is same as the previous comment. There needs to be another templated function which gives the same derivatives until the forth node, and rest should be made zero.

sunethwarna:
The existing derivative implementation will work for existing number of nodes only. For the higher order elements even though they don't use the additional nodes for element size calculations, the derivatives w.r.t. those nodes should be defined. All the tests created in this PR will fail in debug mode. Therefore I would create new templated methods for all the derivatives and call the corresponding existing derivative method for lower node indices and return zero for higher node indices.

jgonzalezusua:
It passes the tests in my local machine (even in debug mode). @rubenzorrilla, @sunethwarna. I hope everything it's OK.

kratos/utilities/element_size_calculator.cpp
```diff
@@ -116,6 +116,33 @@ double ElementSizeCalculator<2,3>::MinimumElementSizeDerivative(
...
+double ElementSizeCalculator<2,6>::MinimumElementSize(const Geometry<Node<3> >& rGeometry)
+{
+
+    const double minimum_element_size = ElementSizeCalculator<2,3>::MinimumElementSize(rGeometry);
+
+
+    return minimum_element_size;
+}
+// Triangle2D6 version.
+template<>
```
rubenzorrilla:
```suggestion
}

// Triangle2D6 version.
template<>
```
Very minor to ease going through the code.

kratos/utilities/element_size_calculator.cpp
```diff
@@ -230,6 +257,31 @@ double ElementSizeCalculator<2,4>::MinimumElementSizeDerivative(
...
+{
+    double element_size_derivative = 0.0;
+
+    if (DerivativeNodeIndex < 4)
+        element_size_derivative = ElementSizeCalculator<2,4>::MinimumElementSizeDerivative(DerivativeNodeIndex,DerivativeDirectionIndex,rGeometry);
+
+    return element_size_derivative;
+}
+
+
```
rubenzorrilla:
```suggestion
```
(again very minor to avoid useless whitespaces)

rubenzorrilla (dismissed):
Changes make sense for me. Removing my blocker. Thanks @jgonzalezusua .

jgonzalezusua:
@loumalouomega, do you agree with the changes?

loumalouomega (commented):
Okay from my side

jgonzalezusua:
Thanks, @loumalouomega. Could you approve?

loumalouomega (approved):
I didn't remember to block this in first place :P

PR #10065, opened 2022-07-26
bugfix minimum element size calculator for quadrilaterals

jgonzalezusua:
**📝 Description**
This PR is aimed to fix a small error in function `ElementSizeCalculator<2,4>::MinimumElementSize` of the `element_size_calculator.ccp`. Instead of comparing the distance between the centers of the faces, it is comparing the distance between the center of the faces and the center of the element.

Tags: 
- Bugfix

**🆕 Changelog**
- Changed x10 and y10 variables. Now they represent the x and y coordinates of the the center of the 10 edge instead of 30 which is a diagonal.


sunethwarna (commented):
I think this makes sense. But you may have to change the reference values of those failing tests.

jgonzalezusua:
Yes, I will change them in a few a days

jgonzalezusua:
@sunethwarna, can you approve the PR when it passes the checks? The new `if` statement in the `test_element_calculator.cpp` is done to check the derivatives when regular quadrilaterals are used.

jgonzalezusua:
@rubenzorrilla, the changes in the fluid tests come from fixing the bug in the minimum lement size calculator

kratos/utilities/element_size_calculator.cpp
```diff
@@ -223,7 +223,9 @@ double ElementSizeCalculator<2,4>::MinimumElementSizeDerivative(
     const double h2_eta_derivative = 2 * deta_x * deta_x_derivative + 2 * deta_y * deta_y_derivative;
 
     const double h2 = h2_xi < h2_eta ? h2_xi : h2_eta;
-    const double h2_derivative = h2_xi < h2_eta ? h2_xi_derivative : h2_eta_derivative;
+    double h2_derivative = h2_xi < h2_eta ? h2_xi_derivative : h2_eta_derivative;
+    if (h2_xi == h2_eta)
+        h2_derivative = std::min(h2_xi_derivative,h2_eta_derivative);
```
sunethwarna: I'm curious about the reasoning for this? derivative at this point is undefined.

sunethwarna (dismissed):
I am ok with the changes. May be you can wait till @rubenzorrilla also approves.

kratos/utilities/element_size_calculator.cpp
jgonzalezusua: This is to avoid an error in the test when the quadrilateral is a square.

jgonzalezusua:
@rubenzorrilla, do you agree?

jgonzalezusua:
@rubenzorrilla, is this PR ok for you?

kratos/utilities/element_size_calculator.cpp
```diff
@@ -223,7 +223,9 @@ double ElementSizeCalculator<2,4>::MinimumElementSizeDerivative(
     const double h2_eta_derivative = 2 * deta_x * deta_x_derivative + 2 * deta_y * deta_y_derivative;
 
     const double h2 = h2_xi < h2_eta ? h2_xi : h2_eta;
-    const double h2_derivative = h2_xi < h2_eta ? h2_xi_derivative : h2_eta_derivative;
+    double h2_derivative = h2_xi < h2_eta ? h2_xi_derivative : h2_eta_derivative;
+    if (h2_xi == h2_eta)
```
rubenzorrilla:
```suggestion
    if (std::abs(h2_xi - h2_eta) < 1.0e-12)
```
Since you're comparing floats.

rubenzorrilla (commented):
Aside of the floating point comparison issue, I'm fine with the changes.

rubenzorrilla (approved):

rubenzorrilla:
Fine for me. Thanks for the fix.

PR #10070, opened 2022-07-27
use the "advanced" version of the bin search

RiccardoRossi:
**📝 Description**
this is a performance fix to avoid the creation of temporary for every node. 

also a large number of maximum results is used which help prevents segfaults in cases in which the maximum number of search results is exceeded. (marked as bugfix because of this reason)

- Performance, Bugfix

**🆕 Changelog**
usingthe "advanced" version of the bin search instead of the simplified one so that TLS storage can be used for auxilary search array

### DIFF
--- a/applications/MeshMovingApplication/custom_utilities/fixed_mesh_ale_utilities.cpp
+++ b/applications/MeshMovingApplication/custom_utilities/fixed_mesh_ale_utilities.cpp
@@ -151,6 +151,7 @@ namespace Kratos
 
     void FixedMeshALEUtilities::ComputeMeshMovement(const double DeltaTime)
     {
+
         // Initialize the PRESSURE and VELOCITY virtual mesh values
         this->InitializeVirtualMeshValues();
 
@@ -165,6 +166,7 @@ namespace Kratos
 
         // Set the mesh moving strategy
         this->SolveMeshMovementStrategy(DeltaTime);
+
     }
 
     void FixedMeshALEUtilities::UndoMeshMovement()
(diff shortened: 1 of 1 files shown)


applications/MeshMovingApplication/custom_utilities/fixed_mesh_ale_utilities.cpp
```diff
@@ -192,13 +194,24 @@ namespace Kratos
 
         // Search the origin model part nodes in the virtual mesh elements and
         // interpolate the values in the virtual element to the origin model part node
-        block_for_each( rOriginModelPart.Nodes(),
-            [&]( Node<3>& rNode )
+
+        unsigned int max_results=10000;
```
RiccardoRossi, line 198: note that i am defining a larger number of results than it was used before. with 1000 it was crashing (which in turn suggests that we should tune the cell size for performance reasons)

applications/MeshMovingApplication/custom_utilities/fixed_mesh_ale_utilities.cpp
```diff
@@ -192,13 +194,24 @@ namespace Kratos
...
+            [&]( Node<3>& rNode,
+                 typename BinBasedFastPointLocator<TDim>::ResultContainerType& search_results
+             )
             {
                 // Find the origin model part node in the virtual mesh
-                Vector aux_N;
+                Vector aux_N(4);
                 Element::Pointer p_elem = nullptr;
-                const bool is_found = bin_based_point_locator.FindPointOnMeshSimplified(rNode.Coordinates(), aux_N, p_elem);
+                const bool is_found = bin_based_point_locator.FindPointOnMesh(
```
loumalouomega, line 208: This is faster than FindPointOnMeshSimplified?

RiccardoRossi: yes, because it avoids creating the temporary array "search results"

loumalouomega: Good to know, because there are two processes in the MeshingApplication using this

applications/MeshMovingApplication/custom_utilities/fixed_mesh_ale_utilities.cpp
```diff
@@ -192,13 +194,24 @@ namespace Kratos
...
+        unsigned int max_results=10000;
+        double search_tol = 1e-5;
+        block_for_each( rOriginModelPart.Nodes(), typename BinBasedFastPointLocator<TDim>::ResultContainerType(max_results),
+            [&]( Node<3>& rNode,
+                 typename BinBasedFastPointLocator<TDim>::ResultContainerType& search_results
+             )
             {
                 // Find the origin model part node in the virtual mesh
-                Vector aux_N;
+                Vector aux_N(4);
```
rubenzorrilla:
```suggestion
                Vector aux_N(TDim+1);
```

rubenzorrilla (approved):

rubenzorrilla:
Merging as I need to do some modifications which will conflict this.

PR #10074, opened 2022-07-27
[BDF] Constant DT and high order schemes

miguelmaso:
**📝 Description**
In #9901 the verbosity was reduced and the warning was turned into an error.
Now, the comparison of small floats is throwing an error because of the roundoff. This is a problem because the simulation is killed.

**🆕 Changelog**
- Better comparison of small floats


loumalouomega (approved):
Okay

miguelmaso:
thanks

Issue #10077, opened 2022-07-27
Reading out of bounds in function_parser_utility


EduardGomezEscandell:
**Description**
Accessing element -1 of a string

**Scope**
If the expression string starts with 'x', then we reach line 38 with `i=0`.
https://github.com/KratosMultiphysics/Kratos/blob/fc2e6d1b32d98dbbedb06b244eb81deedd1ad30b/kratos/utilities/function_parser_utility.cpp#L28-L46
Then we access element `MAX_LONG -1`.
Testing it, I've seen that:
- With `libstdc++` (GNU's standard library implementation), this is understood as reading the `\0` sentinel, so no issue. [This reduced example](https://godbolt.org/z/scW1qTPW7) shows it reading `\0`.
- With `libc++` (LLVM's standard library), this is understood as a good old read outside limits, hence undefined behaviour. [This reduced example](https://godbolt.org/z/Ybvz58oaT) shows the sanitizer complaining and reading `\20`, some random ASCII character.

I haven't tested with Microsoft's implementation (Compiler explorer doesn't allow for running the minimal example either).

**To Reproduce**
This piece of code runs on Centos only, so it can be quite niche. Furthermore, it seems that with gcc it won't break. But it is still undefined behaviour.

**Expected behavior**
No illegal memory accessing.

matekelemen:
I'm curious: shouldn't this have already caused segfaults? For example, [this test](https://github.com/KratosMultiphysics/Kratos/blob/55d962ddeff800edcc6ae4c9f3dd3cda5fb2be28/kratos/tests/test_function_parser_utility.py#L28) begins with an ```x``` but passed the CI on every compiler for years now.

EduardGomezEscandell:
That's what I thought, but aparently on gcc's implementation (the one in CI) accessing `string::npos` returns `\0`. Note that `(std::size_t)(-1)` undeflows to `string::npos`.

With clang's implementation this breaks, but the Centos CI run doesn't test it.

matekelemen:
> With clang's implementation this breaks

We also have a clang run in the CI, don't we? Does this only become an issue in release builds?

EduardGomezEscandell:
We have a ubuntu clang CI, but this piece of code is only run on Centos (it is the fallback when regex is not available).

It is indeed possible that we're building with libstdc++, I don't know which one is the default and which one is used.

Issue #10041, opened 2022-07-18
Niche error in function_parser_utility


EduardGomezEscandell:
**Description**
This is not really problematic, I'm reporting mostly so its documented. I don't believe anyone is afected by this.

In function `CheckThereIsNotx` in here:
https://github.com/KratosMultiphysics/Kratos/blob/fc2e6d1b32d98dbbedb06b244eb81deedd1ad30b/kratos/utilities/function_parser_utility.cpp#L28-L46

It ensures that functions such as `exp` are not confused with an `x` as a variable name. Tinyexpr also supports hexadecimal number so `f(t) = 0x13 + t` would still be misparsed as a function of x.

**Scope**
Parsing of functions in json

**To Reproduce**
With some process that allows functions, write something like `0x100 + t` and see how this function is labeled as depending on `x` by `CheckThereIsNotx`.

**Expected behavior**
Should not be parsed as depending on `x`.

**Environment**
I haven't reproduced this issue, I've reached this conclusion with static analysis. Note that I have confirmed that tinyexpr supports hex (With Ubuntu 22.04 and gcc, if that's relevant).

matekelemen:
Oh god; good chance to replace this with a regex?

EduardGomezEscandell:
Solved in https://github.com/KratosMultiphysics/Kratos/pull/10251

PR #10078, opened 2022-07-27
Fixing bugs in function_parser_utility

EduardGomezEscandell:
## 📝 Description
This PR fixes two bugs:
 - #10041
 - #10077

Ideally we'd use a regex (even more ideally [CTRE](https://github.com/hanickadot/compile-time-regular-expressions)) and be done, but aparently that is not possible on Centos.
This PR fixes the bugs in the fallback case, and unifies behaviour into a single path.

Splitting it into two paths only allowed for bugs to lay dormant.

## Implementation details
I created a fancy version of adjacent_find, but with 3 instead of 2 adjacent elements passed to the predicate.
```
    Hello, world
      ~~~
     Pred('l', 'l', 'o')       window analyses three entries at a time
```
```
    Hello, world
   ~~~        ~~~
              Pred('l', 'd', '\0')    first and last iterations are padded
   Pred('\0', 'H', 'e')
```
Then the logic is implemented for an arbitrary set of three contiguous characters. Basically manually appplying the following regex:
```regex
/\b[xyzXYZ]\b/
```
based on the documentation of tinyexpr.

## Alternative solution
Keeping the regex when not in centos, and hardcoding DependsOnSpace=true when in centos. This check is only an optimization to avoid recomputing the function at every node, so we could do without it if we prefered so.

### DIFF
--- a/kratos/utilities/function_parser_utility.cpp
+++ b/kratos/utilities/function_parser_utility.cpp
@@ -12,7 +12,10 @@
 //
 
 // System includes
-#include <regex>
+#include <algorithm>
+#include <array>
+#include <cctype>
+#include <iterator>
 
 // External includes
 #include "tinyexpr/tinyexpr/tinyexpr.h"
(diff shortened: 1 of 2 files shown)


loumalouomega (commented):
From my side OK

loumalouomega:
Just one thing, in Centos the compiler is going to be updated to accept C++17, so this change can be avoided

kratos/utilities/function_parser_utility.cpp
```diff
@@ -24,25 +27,64 @@
...
-                    return false;
-                }
-            }
-        }
+    // Edge case: length 0
+    if(Begin == End) {
+        return End;
+    }
+    
+    auto wmiddle = std::next(Begin); // Middle of the window
```
matekelemen, line 51: > The window is padded such that every element of the sequence is the middle element of the window once.

Minor thing: this is a bit confusing because ```wmiddle``` will never point to ```Begin```, but I see the necessary check is performed in line 58 so I think the behaviour is correct.

EduardGomezEscandell:
How son is that comming?

kratos/utilities/function_parser_utility.cpp
```diff
@@ -61,19 +103,7 @@ BasicGenericFunctionUtility::BasicGenericFunctionUtility(const std::string& rFun
     InitializeParser();
 
     // Check if it depends on space
-#if !(defined(__GNUC__) && !defined(__clang__) && (__GNUC__ < 4 || (__GNUC__ == 4 && (__GNUC_MINOR__ < 9))))
```
loumalouomega, line 64: My question is if it is more efficient keep this check and do the new method only in older compilers

loumalouomega:
> How soon is that comming?

Ping @roigcarlo

kratos/utilities/function_parser_utility.cpp
EduardGomezEscandell: Well the issue is that splitting the code is what led to these bugs being uncaught.

Besides, std::regex is notoriously slow, so I wouldn't be surprised if it was slower than the hand-rolled implementation. But no way of knowing this without testing.

However, since this piece of code is only ran a few times before starting the simulation, performance is not critical.

kratos/utilities/function_parser_utility.cpp
```diff
@@ -24,25 +27,64 @@
...
+ *  - Function containing x:          exp
+ */
+bool ExpressionDependsOnSpace(std::string const& rExpression) {
+    return PaddedWindowFind(rExpression.cbegin(), rExpression.cend(), '\0', 
+        [] (char lead, char middle, char last) {
+            constexpr auto const& variables = "xyzXYZ";
+            if(std::find(std::begin(variables), std::end(variables), middle) == std::end(variables)) {
+                return false; // Middle character is not xyzXYZ
+            }
+            return !std::isalnum(lead) && !std::isalnum(last);
```
matekelemen, line 85: What about cases with leading digits other than 0? This would reject ```1x``` ... ```9x```, right?

matekelemen (commented):
Awesome!

kratos/utilities/function_parser_utility.cpp
EduardGomezEscandell: Those can be two cases:
- `longer_variable_name1x...`: Should, and is rejected
- `1x...`: Invalid variable name (must start with a letter): Will crash earlier, would be rejected anyways

kratos/utilities/function_parser_utility.cpp
EduardGomezEscandell: Yeah, the behaviour is correct (as you pointed out about line 58). I could make it clearer but that would require an extra iterator copy and a couple extra `std::advance`s.

kratos/utilities/function_parser_utility.cpp
matekelemen: Right, I wasn't sure whether tinyexpr supported shorthand factors (eg.: ```2x``` for ```2*x```) but I see it doesn't (it would be ambiguous with hex formats anyway).

kratos/utilities/function_parser_utility.cpp
matekelemen: Yeah I'd say don't worry about it, it just takes a minute to figure out what's going on.

rubenzorrilla:
Do these changes have any performance impact?

EduardGomezEscandell:
Algorithmic complexity is O(n) with length of the string, same or better than before, but strings are usually short so the constants are more relevant here. Whether it is faster or slower I don't know, but it's going to be similar.

The impact is going to be negligible anyways. Only processes with function fields (like typical boundary conditions with `"value": "2*x"`) use this, and is used only once at construction.

rubenzorrilla:
> Algorithmic complexity is O(n) with length of the string, same or better than before, but strings are usually short so the constants are more relevant here. Whether it is faster or slower I don't know, but it's going to be similar.
> 
> The impact is going to be negligible anyways. Only processes with function fields (like typical boundary conditions with `"value": "2*x"`) use this, and is used only once at construction.

Perfect then. If changes only affect the parsing performance is not an issue.

EduardGomezEscandell:
> Just one thing, in Centos the compiler is going to be updated to accept C++17, so this change can be avoided

If I understand correctly, this PR can be closed now, right? We can use std::regex on centos now.

loumalouomega:
Yes, we can

EduardGomezEscandell:
Closing. Note that the bug is still out there, to fix it we need to apply the proper regex: `/\b[xyzXYZ]\b/`

loumalouomega:
@EduardGomezEscandell can we close the branch?=

EduardGomezEscandell:
Yes, no need for it

PR #10079, opened 2022-07-27
New Heat Generation Mechanisms in Thermal DEM Application

rlrangel:
In the DEMApp, a new input argument was added to two functions in order to store the damping components of the contact force between elements.

The rest of the modifications are in the ThermalDEMApp to add new heat generation mechanisms (by viscous forces and rolling friction), and write graphs of these data.


applications/DEMApplication/custom_elements/spheric_particle.cpp
```diff
@@ -1931,9 +1931,9 @@ std::unique_ptr<DEMDiscontinuumConstitutiveLaw> SphericParticle::pCloneDiscontin
 
 void SphericParticle::ComputeOtherBallToBallForces(array_1d<double, 3>& other_ball_to_ball_forces) {}
 
-void SphericParticle::StoreBallToBallContactInfo(const ProcessInfo& r_process_info, SphericParticle::ParticleDataBuffer& data_buffer, SphericParticle* other_element, double GlobalContactForce[3], bool sliding) {}
+void SphericParticle::StoreBallToBallContactInfo(const ProcessInfo& r_process_info, SphericParticle::ParticleDataBuffer& data_buffer, double GlobalContactForceTotal[3], double LocalContactForceDamping[3], bool sliding) {}
```
maceligueta, line 1938: Why not keep `SphericParticle* other_element `as one of the arguments? I think it can be useful in the future. `StoreBallToRigidFaceContactInfo` is still getting `DEMWall* other_element`, so here we could do the same.

maceligueta: If the neighbour is in the DataBuffer then we can remove the neighbour from both places.

rlrangel: I removed the particle neighbor from the input arguments of StoreBallToBallContactInfo because it is in the DataBuffer. However, the wall neighbor is not, so I left it in the argumentos of StoreBallToRigidFaceContactInfo.

maceligueta: Ok. Maybe the DataBuffer should include the wall neighbour as well for symmetry and completeness (to reduce number of arguments everywhere and create similar functions for `Store*ContactInfo`). Do you think it's a good idea, @rlrangel ? It's been a long time since I don't touch this part of the code.

rlrangel: Yes, this last commit is doing it.

maceligueta (approved):

PR #10081, opened 2022-07-28
Adding features for release of eFlows4HPC European Project

Rbravo555:
**📝 Description**
Adds features for parallel computation of SVD and Simulations 

- *eFlows4HPC* , *MPI* , *RomApplication*

**🆕 Changelog**
- Added features to the `fluid_dynamic_analysis_rom` to ease its launching in clusters (no extra tests added, since it depends on external library COMPSs)
- Added a parallel fixed-rank randomized svd script  (no extra tests added, since it depends on external library COMPSs)
- Added a parallel fixed-presicion randomized svd. Test added


applications/RomApplication/python_scripts/fluid_dynamics_analysis_rom.py
```diff
@@ -2,21 +2,36 @@
 import KratosMultiphysics.FluidDynamicsApplication
```
rubenzorrilla, line 2: These changes go against the latest modifications. Implementations must be done in the `rom_analysis`. Note that this class is deprecated and is only keep for retrocompatibility.

Is there any technical limitation that avoids implementing these new features in the `rom_analysis`?

applications/RomApplication/tests/test_randomized_singular_value_decomposition_mpi.py
```diff
@@ -0,0 +1,81 @@
...
+    TestMatrix = np.zeros((rows,degree))
+    x = np.linspace(0,1,rows)
+    for i in range(degree):
+        TestMatrix[:,i] = np.power(x,i)
+    return np.matlib.repmat(TestMatrix, 1, repetitions)
+
+class TestRandomizedSVDMPI(KratosUnittest.TestCase):
+
+    @KratosUnittest.skipUnless(numpy_and_mpi4py_available, "numpy and mpi4py are required for TSQR MPI")
+    def test_radomized_svd(self):
```
EduardGomezEscandell: Unit tests should be deterministic. The random algorithm should be seeded with some arbitrary (but consistent) value at the begining of every test.

SADPR: It is already being seeded ("0") since the mpi algorithm requires it for all cases. In addition, note that it's being compared by assertLess to a given tolerance, this must be satisfied for random theory in all cases. Referring to Brunton's book "There is an extensive literature on random matrix theory, where the above stereotypes (this case) are almost certainly true, meaning that they are true with high probability".

EduardGomezEscandell: Oh I couldn't find the seeding with a quick scan. Good then.

applications/RomApplication/python_scripts/fluid_dynamics_analysis_rom.py
Rbravo555: The `rom_analysis` that we have now works well for testing small cases, since it allows to train a ROM quite automatically. However, this automatic workflow was not inmmediatly usable in a larger parallel training campain.

That is, the `rom_analysis` builds the ROMs using a single time-dependent, or a single quasi-static simulation(with multiple load-steps). If more parameter variations are to be studied, one needs to compute the SVD of the snapshots for the whole set of simulations. The scripts for such a larger parallel training are not included in this PR, since they involve the external library COMPSs. They are being place in the Kratos/Examples repository.

Incorporating an expanded workflow for the training of a ROM in the `rom_analysis` will be done in another PR.

Notice that the modifications on the `fluid_dynamic_analysis_rom` add but a single extra functionality in comparison to the `rom_analysis`. The PATH is accepted as an extra parameter, in order to allow to run in a cluster. All the real handling is done externally.

applications/RomApplication/tests/tests_RomApplication_mpi.py
```diff
@@ -0,0 +1,42 @@
+# Importing the Kratos Library
```
roigcarlo: I think this file is duplicated

applications/RomApplication/python_scripts/fluid_dynamics_analysis_rom.py
rubenzorrilla: > 
> Incorporating an expanded workflow for the training of a ROM in the `rom_analysis` will be done in another PR.
> 

If there is no technical limitation, can't we just simply do this?. From the sake of code sustainability it makes no sense to invest effort on developing and maintaining deprecated code.

> Notice that the modifications on the `fluid_dynamic_analysis_rom` add but a single extra functionality in comparison to the `rom_analysis`. The PATH is accepted as an extra parameter, in order to allow to run in a cluster. All the real handling is done externally.

This can be easily added to current `rom_analysis`.

rubenzorrilla: We had a brief discussion with @Rbravo555 about this. As the due date of the eFlows4HPC release is close, we decided to do the release with the old stage version to ensure everything is working fine. This comes with the compromise all these developments to the current `rom_analysis` after the deliverable submission.

roigcarlo (approved):
:+1: If no other comments I will approve, since we need a release soon

PR #10084, opened 2022-07-28
Fix surface load distribution process

armingeiser:
**📝 Description**
The surface load distribution process assumed that the `SURFACE_LOAD` is a force resultant, instead it is a distributed load (force over area). The Load distribution process is fixed accordingly. 

**🆕 Changelog**
- Fix distribute_load_on_surface_process

### DIFF
--- a/applications/StructuralMechanicsApplication/custom_processes/distribute_load_on_surface_process.cpp
+++ b/applications/StructuralMechanicsApplication/custom_processes/distribute_load_on_surface_process.cpp
@@ -29,7 +29,7 @@ DistributeLoadOnSurfaceProcess::DistributeLoadOnSurfaceProcess(ModelPart& rModel
 {
     Parameters default_parameters(R"(
         {
-            "help"            : "This process distributes a load on surface load conditions belonging to a modelpart. The load is distributed according to the surface area.",
+            "help"            : "This process distributes a force on a all surface load conditions belonging to a modelpart. The surface load is computed by equally distributing the force over the surface area",
             "model_part_name" : "please_specify_model_part_name",
             "interval"        : [0.0, 1e30],
             "load"           : [1.0, 0.0, 0.0]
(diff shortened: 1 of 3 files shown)


loumalouomega (changes requested):
Technically true, but this will crash many things. Blocking

armingeiser:
> Technically true, but this will crash many things. Blocking

By breaking things you mean it will change current behaviour? Yes thats true. 

How would you prefer to make such a change/fix? Introducing a new process with a different name and throw a deprecation warning/error for this old one?

KlausBSautter:
Very nice, we need that change!

KlausBSautter:
We will need this for a Project soon. I am Happy you die or @armingeiser , thanks :)
As you already said, Maybe renaming it would be a good solution If @loumalouomega wants to keep the old process. But at some Point replacing it would be the best solution I think.

loumalouomega:
> We will need this for a Project soon. I am Happy you did it @armingeiser , thanks :) As you already said, Maybe renaming it would be a good solution If @loumalouomega wants to keep the old process. But at some Point replacing it would be the best solution I think.

As I see that the CI is not broken, I will approve if someone else from @KratosMultiphysics/structural-mechanics , specially @RiccardoRossi agrees

loumalouomega:
> One more question: @armingeiser did you check if the area load is finally multiplied with the current area or the reference area? @loumalouomega I remember we had a similar problem with the mass matrices.

Yes, but my concern is not to break current things

KlausBSautter:
> > One more question: @armingeiser did you check if the area load is finally multiplied with the current area or the reference area? @loumalouomega I remember we had a similar problem with the mass matrices.
> 
> Yes, but my concern is not to break current things

Sure! But the area configuration also needs to be clarified before merging.

KlausBSautter:
> > One more question: @armingeiser did you check if the area load is finally multiplied with the current area or the reference area? @loumalouomega I remember we had a similar problem with the mass matrices.
> 
> No i did not verify that. ( i am only using it for linear statics...)

Ok, anyway I think this works from my side

RiccardoRossi:
what about this one? should be retaken?

PR #10086, opened 2022-07-29
[Core] Fix split triangle class

AriadnaCortesDanes:
**📝 Description**
Split triangle was defined as a set of functions without any class (nor namespace). Just fixed that and the derived changes. 😄


pooyan-dadvand (commented):
For me is ok! @rubenzorrilla and @RiccardoRossi if you don't have any inconvenience I will approve it.

pooyan-dadvand (approved):
Thanks!

rubenzorrilla (approved):
Thanks for the fix. Please go ahead.

PR #10087, opened 2022-07-29
[MeshingApplication] Add support to specify local entity parameters 

marcnunezc:
**📝 Description**
This adds an interface to specify local parameters (hmin, hmax, hausd) to entities with the same reference with MMG.

The need for this started with a fluid dynamics case from @uxuech where the bounding box is not a classical parallelepiped. The issue was that corners of the bounding box were refined. The issue was reported in:

https://forum.mmgtools.org/t/unwanted-refinement-at-domain-corners/633

Using parameters (https://www.mmgtools.org/parameter-file) we were able to fix the issue by providing high local hausdorff values at all faces except the ground (the geometry of interest is touching the ground in this example.

image

image

This case was run with the following mmg_parameters:

```
KratosMultiphysics.Parameters("""
    {
        "filename"                         : "out",
        "discretization_type"              : "STANDARD",
        "advanced_parameters"                  :
        {
            "local_entity_parameters_list"        : [
                {
                    "model_part_name_list" : ["Top", "Wall2", "Inlet", "Outlet", "Wall1"],
                    "hmin"            : 50.0,
                    "hmax"            : 100.0,
                    "hausdorff_value" : 1000.0
                }]
        }
    }
    """)
```

**🆕 Changelog**
E.g.
- Added interface to specify local entities mmg parameters

### DIFF
--- a/applications/MeshingApplication/custom_processes/mmg/mmg_process.cpp
+++ b/applications/MeshingApplication/custom_processes/mmg/mmg_process.cpp
@@ -747,7 +747,7 @@ void MmgProcess<TMMGLibrary>::ApplyLocalParameters() {
     }
 
     // Count number of parameters given by the user
-    auto parameter_array = mThisParameters["advanced_parameters"]["local_entity_parameters_list"];
+    const auto parameter_array = mThisParameters["advanced_parameters"]["local_entity_parameters_list"];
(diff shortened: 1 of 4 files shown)


loumalouomega (approved):

loumalouomega:
Thanks

PR #10088, opened 2022-07-29
Bug fix in TetGen for the Thermal DEM Application

rlrangel:
Attempt to fix a problem in TetGen:

In the Windows machines that I've tried, when running in Debug mode, both ThermalDEM and PFEMforFluidDynamics applications crash in line 557 of TetGen file predicates.cxx (error shown in the image attached).

When in Release, the PFEM application works well, but the ThermalDEM still has problems with that line.

Similar behavior was observed in Linux systems.

My dangerous, but apparently effective solution, was to comment the problematic line.
It started working well after that.

However, in order to make this edition to the TetGen code, I added the files to the ThermalDEM application, instead of using the downloaded files from the DelaunayMeshing application.
I didn't know we were not allowed to add the files to the repo.

error


AlejandroCornejo:
Since it only affects your App, ok!

just pinging @AFranci , @maceligueta and @KratosMultiphysics/technical-committee to inform them that this issue can happen in other tetgen dependent apps

applications/ThermalDEMApplication/CMakeLists.txt
```diff
@@ -3,14 +3,14 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
 message("**** configuring KratosThermalDEMApplication ****")
 
 kratos_add_dependency(${KRATOS_SOURCE_DIR}/applications/DEMApplication)
-kratos_add_dependency(${KRATOS_SOURCE_DIR}/applications/DelaunayMeshingApplication)
+## kratos_add_dependency(${KRATOS_SOURCE_DIR}/applications/DelaunayMeshingApplication) ## UNCOMMENT TO GET TETGEN FROM DELAUNAY_MESHING APPLICATION
```
AlejandroCornejo, line 6: maybe is it better to solve directly the DelaunayApp? @KratosMultiphysics/technical-committee

applications/ThermalDEMApplication/external_libraries/tetgen/example.poly
```diff
@@ -0,0 +1,84 @@
+28  3  0  1
+1  0  0  0  1
+2  2  0  0  1
+3  2  2  0  1
```
AlejandroCornejo, line 4: what is the purpose of this?

AlejandroCornejo (changes requested):

rubenzorrilla:
> We cannot add tetgen to the repository. Those files should be removed. Tetgen options must be activated so that it gets downloaded and compiled automatically. I thought that was already solved in the Delaunay App. See https://github.com/KratosMultiphysics/Kratos/blob/master/INSTALL.md#tpl-libraries

That's the main point. @roigcarlo would you mind keeping track of this on behalf of the @KratosMultiphysics/technical-committee ?

roigcarlo (changes requested):
Blocking as well. Under no circumstance we can allow tetgen code in the repo.

If tetgen is needed it needs to be included via the usage of any of the flavours of `USE_TETGEN_NONFREE_TPL` as explained [here](https://github.com/KratosMultiphysics/Kratos/blob/master/INSTALL.md#tetgen)

Also please provide a description of the changes in the PR description as otherwise is hard to keep track of them (we have some automated systems that depen on this).

Finally, if you remove tetgen directly in this PR, when eventually merging, please use the `merge and squash` as this will prevent tetgen files appearing in the history as well.

rlrangel:
Sorry for not adding a proper title and description to the PR.
I have just uploaded them.

I was not aware that the TetGen files cannot be added to the repo (is it related to license?)

I will later check these links to see if there is a solution to my issue.
If not, I currently have no clue on how to solve.

roigcarlo:
@rlrangel Yep its because tetgen licence invalidates BSD... for us is not strictly a problem but as long as someone needs to integrate Kratos in their code it becomes an ordeal.

We created the macro for that, so that you can include tetgen without having it into the repo, I think there is a PR currently open with a similar scenario (#10127) You can take a look at how they are including tetgen. The CMakes for the meshing application should give yo an idea as well.

P.s. Thanks for updatin the description!

rlrangel:
Thanks for the clarifications @roigcarlo.

I think this PR can be closed.

rubenzorrilla:
Closing as suggested by @rlrangel.

PR #10091, opened 2022-08-02
[Core] GenericFindElementsNeighbours using Execute

rubenzorrilla:
**📝 Description**
The `GenericFindElementsNeighbours` process uses the `ExecuteInitialize` method instead of the `Execute` one, which is the one the other neighbor calculation processes are using. On top of being inconsistent, this is quite misleading for developers, as calling the `Execute` turns into do nothing as it is calling the base class empty implementation. This turns into a segfault when asking for the neighbors. Figuring this out might not be straightforward for newbies.

Besides, I also realized that the current implementation is not following the base `Process` public API as there is a `public` `HasNeighboursInFaces`. This is only used in the tests so we may remove it (asking @pablobecker since, as far as I understand, he implemented it in #9894). If it can't be removed, I think it should be moved to a separate class and be implemented as a `static` function. Note that this doesn't require anything from the process, but makes it mandatory to keep it alive to be called.

### DIFF
--- a/applications/ConvectionDiffusionApplication/python_scripts/move_particle_utility_process.py
+++ b/applications/ConvectionDiffusionApplication/python_scripts/move_particle_utility_process.py
@@ -42,12 +42,10 @@ class MoveParticleUtilityProcess(KM.Process):
 
     def ExecuteBeforeSolutionLoop(self):
         dimension = self.model_part.ProcessInfo[KM.DOMAIN_SIZE]
-        num_of_avg_elems = 10
-        num_of_avg_nodes = 10
         neighbor_search = KM.FindNodalNeighboursProcess(self.model_part)
         neighbor_search.Execute()
         neighbor_elements_search = KM.GenericFindElementalNeighboursProcess(self.model_part)
-        neighbor_elements_search.ExecuteInitialize()
+        neighbor_elements_search.Execute()
 
         settings = self.model_part.ProcessInfo[KM.CONVECTION_DIFFUSION_SETTINGS]
         self.unknown_var = settings.GetUnknownVariable()
--- a/applications/FemToDemApplication/custom_processes/extend_pressure_condition_process.cpp
+++ b/applications/FemToDemApplication/custom_processes/extend_pressure_condition_process.cpp
@@ -311,7 +311,7 @@ void ExtendPressureConditionProcess<TDim>::Execute()
 {
     // We search the neighbours for the generation of line loads
     auto find_neigh = GenericFindElementalNeighboursProcess(mrModelPart);
-    find_neigh.ExecuteInitialize();
+    find_neigh.Execute();
     auto& r_process_info = mrModelPart.GetProcessInfo();
 
     // Remove previous line loads-> Only the 1st iteration
--- a/applications/FemToDemApplication/custom_processes/regenerate_pfem_pressure_conditions_process.cpp
+++ b/applications/FemToDemApplication/custom_processes/regenerate_pfem_pressure_conditions_process.cpp
@@ -205,7 +205,7 @@ void RegeneratePfemPressureConditionsProcess<TDim>::Execute()
 {
     // We search the neighbours for the generation of line loads
     auto find_neigh = GenericFindElementalNeighboursProcess(mrModelPart);
-    find_neigh.ExecuteInitialize();
+    find_neigh.Execute();
 
     if (!mrModelPart.HasSubModelPart("PFEMPressureConditions")) {
         mrModelPart.CreateSubModelPart("PFEMPressureConditions");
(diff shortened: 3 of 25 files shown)


loumalouomega (approved):
Okay from my side, I see that ExecuteInitialize still works and throws a warning

rubenzorrilla:
> Okay from my side, I see that ExecuteInitialize still works and throws a warning

Yes. We'd better keep it for a while. In any case I'll wait for @pablobecker 's answer about the `HasNeighboursInFaces`. If possible, I'd remove it in this PR.

pablobecker:
Since the goal of the original PR was to directly replace the old process and that this is meant to be used a a tool by other processes, I thought it made more sense to compute the neighb at Initialize. But you're right that other neigh processes do the computation at Execute, so good for me to keep things consistent. Regarding the HasNeighboursInFaces function, why not make it an utility? it could be useful (for example I am checking faces with nullptr exactly for this reason in https://github.com/KratosMultiphysics/Kratos/pull/9949 )

ddiezrod:
This goes further than the scope of this MR, but I am quite hesitant about the separation in processes that are called only in Execute and those that are called in ExecuteInitialize, InitializeSolStep etc. This one is a good example. It only does one thing so it would make sense to only implement Execute function, but also, it would make sense to be able to add it to the list of processes in case your solver/analysis does not call it ( I was in fact already doing this)

I think this is something that was discussed some time ago, but probably adding a list of "operations" in your ProjectParameters would fix this problem. This could be a list of processes which are called at the same moment as the ExecuteInitialize but they would only be called through Execute function,  I think that would be more clear.

rubenzorrilla:
Here my two cents.

> Regarding the HasNeighboursInFaces function, why not make it an utility? it could be useful (for example I am checking faces with nullptr exactly for this reason in #9949 )

OK. Now I see why there are no usages yet (#9949 is still pending). My concern is that placing it inside a process is not the proper place. I think this is very suited to be either inside any of the current utilities in the Kratos core (not sure in which one)  or in a new one. Having said so, maybe we can do this in a standalone PR as we're changing many files in here.

> This goes further than the scope of this PR, but I am quite hesitant about the separation in processes that are called only in Execute and those that are called in ExecuteInitialize, InitializeSolStep etc. This one is a good example. It only does one thing so it would make sense to only implement Execute function, but also, it would make sense to be able to add it to the list of processes in case your solver/analysis does not call it ( I was in fact already doing this)

Absolutely agree. Indeed these kind of routines are a process because historical reasons. However, they are not a utility as you want to enforce an API. Not long ago we created the concept of `Operations` for the sake of multistage coupling, which only feature such `Execute` method. So these, are IMO an operation. The point is that refactoring everything at this point is most probably a worthless effort.

About the possibility of calling these kind of routines from the json process list, I tend to disagree. Such processes lists were conceived (mainly) for the initial and BCs impositions. Note that adding them in the json lists translates the logic to the GUI that writes those json files. Besides, processes like the one in this PR are normally crucial operations in your resolution procedure that are to be called once, so you want to keep them controlled and hidden inside the solver.

A different thing would be other kind of processes with many customizable settings (e.g. level set convection). I agree that your comment in this context makes much more sense. However, these are not the majority of the cases (I mean processes with many customizable settings that are not BC related or postprocessing).

> 
> I think this is something that was discussed some time ago, but probably adding a list of "operations" in your ProjectParameters would fix this problem. This could be a list of processes which are called at the same moment as the ExecuteInitialize but they would only be called through Execute function, I think that would be more clear.

While it is technically possible, this would add extra complexity to our execution, which is a bit complex right now (I'm posting the complaints and feedback from some newcomers). Hence, I'd only add this if there is a technical limitation that impedes achieving the same with current capabilities.

ddiezrod:
Ok  @rubenzorrilla I see you point and also makes sense. Anyway, I do not want to move the internal logics of the solver to the GUI. I was thinking more in very specific cases like benchmarks or validations where you need to do an operation that you would not do in a "standard" run. You might end up at the end adding more things to the solver to be able to run the problem, when they are not needed usually and might be quite heavy (as neighbors elements are), or even worse, you just cannot run your validation because the solver/analysis cannot be changed at that point.

rubenzorrilla:
> Ok @rubenzorrilla I see you point and also makes sense. Anyway, I do not want to move the internal logics of the solver to the GUI. I was thinking more in very specific cases like benchmarks or validations where you need to do an operation that you would not do in a "standard" run. You might end up at the end adding more things to the solver to be able to run the problem, when they are not needed usually and might be quite heavy (as neighbors elements are), or even worse, you just cannot run your validation because the solver/analysis cannot be changed at that point.

Yes... sometimes it would be nice to have a bit more of flexibility while developing or testing, but I think that we'll need to live with that... :disappointed:

rubenzorrilla:
I think we discussed all the concerns so I'll merge this to start the PR with the modifications related to the `HasNeighboursInFaces`. Besides, this is blocking #10096.

PR #10092, opened 2022-08-02
[HDF5] Fix Integration Point Operation Names

matekelemen:
Fix inconsistent naming in HDF5 output operations related to gauss points (now consistent with VTK output). This fixes mesh output processes not finding integration point output operations.


sunethwarna:
Again the naming problem. You may have to check the actual name.

matekelemen:
We've worked on a different branch this morning, which is why we had the naming issues, so this fix is not actually relevant in the current master.

The last two commits I pushed are cherry-picked from [another branch](https://github.com/KratosMultiphysics/Kratos/tree/hdf5/checkpoint) and are the ones that justify the fix. @sunethwarna please have a look and decide whether these changes should be considered now or later when I'll open a PR for the orher branch.

PR #10095, opened 2022-08-02
[HDF5] PointSetOutput Updates

matekelemen:
- fix the ```HDF5File``` context portability issues and make it create the parent directory structure if it doesn't exist yet.
- make the parameters used in python tests explicit so they can be used as examples

### DIFF
--- a/applications/HDF5Application/custom_io/hdf5_file.cpp
+++ b/applications/HDF5Application/custom_io/hdf5_file.cpp
@@ -191,7 +191,9 @@ File& File::operator=(File&& rOther)
 
 File::~File()
 {
-    H5Fclose(m_file_id);
+    if (0 <= m_file_id) {
+        H5Fclose(m_file_id);
+    }
 }
 
 bool File::HasPath(const std::string& rPath) const
@@ -627,6 +629,17 @@ void File::Flush()
 {
     KRATOS_ERROR_IF(H5Fflush(m_file_id, H5F_SCOPE_GLOBAL) < 0)
         << "H5Fflush failed." << std::endl;
+}
+
+void File::Close()
+{
+    if (0 <= m_file_id) {
+        const auto close_result = H5Fclose(m_file_id);
+        KRATOS_ERROR_IF(close_result < 0) << "Failed to close " << m_file_name << " with error code " << close_result;
+        m_file_id = -1;
+    } else {
+        KRATOS_WARNING("Invalid file handle") << "Attempt to close an invalid file" << std::endl;
+    }
 }
 
 unsigned File::GetFileSize() const
--- a/applications/HDF5Application/python_scripts/core/file_io.py
+++ b/applications/HDF5Application/python_scripts/core/file_io.py
@@ -150,3 +150,22 @@ def Create(settings, data_comm):
     io.file_driver = settings['file_driver']
     io.echo_level = settings['echo_level']
     return io
+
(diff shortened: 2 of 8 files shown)


matekelemen:
@sunethwarna ready for review

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -191,7 +191,9 @@ File& File::operator=(File&& rOther)
 
 File::~File()
 {
-    H5Fclose(m_file_id);
+    if (0 <= m_file_id) {
+        H5Fclose(m_file_id);
+    }
```
sunethwarna, line 196: I suppose you have already checked this with the documentation ryt? There is no anyother way given by hdf5 to check whether the handle is valid?

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -629,6 +631,17 @@ void File::Flush()
         << "H5Fflush failed." << std::endl;
 }
 
+void File::Close()
+{
+    if (0 <= m_file_id) {
+        const auto close_result = H5Fclose(m_file_id);
+        KRATOS_ERROR_IF(close_result < 0) << "Failed to close " << m_file_name << " with error code " << close_result;
+        m_file_id = -1;
```
sunethwarna, line 639: I suppose again hdf5 is not using the negative values.

applications/HDF5Application/python_scripts/core/file_io.py
```diff
@@ -150,3 +150,24 @@ def Create(settings, data_comm):
...
+        if self.__parameters.Has("io_type"):
+            self.__parameters["io_type"].SetString(distributed)
+        else:
+            self.__parameters.AddString("io_type", distributed)
+        self.__file = Create(self.__parameters, self.__model_part.GetCommunicator().GetDataCommunicator()).Get(self.__model_part)
+        return self.__file
+
+    def __exit__(self, exit_type, exit_value, exit_traceback) -> None:
+        self.__file.Close()
+        self.__file = None
```
sunethwarna, line 173: We do not need `self.__file = None` anymore ryt? the objects will automatically be destroyed when exiting the context.

sunethwarna (commented):
This is fine by me.. I only have minor comments.

applications/HDF5Application/custom_io/hdf5_file.cpp
matekelemen: I checked the doc and haven't found any exposed function that'd check whether the file is open, but it fails if it's closed twice.

applications/HDF5Application/custom_io/hdf5_file.cpp
matekelemen: negative values are reserved for error codes, and ```m_file_id``` defaults to ```-1``` so it should be safe.

applications/HDF5Application/python_scripts/core/file_io.py
matekelemen: True, it's not really necessary; I'm just setting it so it's impossible to abuse the context. There's no way to check whether ```HDF5::File``` is in a valid state right now, so I'm setting ```__file``` to ```None``` whenever it's in an invalid state.

sunethwarna (approved):

PR #10096, opened 2022-08-02
[Core] Adding Templated FindGlobalNodalEntityProcess

sunethwarna:
**📝 Description**
This PR templates `FindGlobalNodalElementalNeighboursProcess` to `FindGlobalNodalEntityNeighboursProcess` and uses it to create `FindGlobalNodalElementalNeighboursProcess` and `FindGlobalNodalConditionNeighboursProcess`.

Still the file `find_global_nodal_elemental_neighbours_process.h` file is kept because some of the applications are using this directly in their processes. Removal of these will be done in a seperate PR. For now I added the deprecated message.

Core processes where `FindGlobalNodalElementalNeighboursProcess` is replaced with the templated version.

**🆕 Changelog**
- Added `FindGlobalNodalEntityNeighboursProcess`
- Python exposure of `FindGlobalNodalElementalNeighboursProcess` and `FindGlobalNodalConditionNeighboursProcess` using the templated class
- Added tests for both `Element` and `Condition` type classes
- Updated core processes to use templated class
- Added a deprecated message to existing c++ `FindGlobalNodalElementalNeighboursProcess`


rubenzorrilla (changes requested):
Preventing merge until #10091 is inside. Otherwise it might become quite messy.

kratos/processes/find_global_nodal_entity_neighbours_process.h
```diff
@@ -0,0 +1,177 @@
...
+    ///@}
+    ///@name Operators
+    ///@{
+
+    FindGlobalNodalEntityNeighboursProcess<EntityType>& operator=(FindGlobalNodalEntityNeighboursProcess<EntityType> const& rOther) = delete;
+
+    void operator()()
+    {
+        Execute();
+    }
```
rubenzorrilla: I'd take the chance to remove the operator call to make explicit that we're calling the `Execute` in the places we use the process.

rubenzorrilla: (I'm assuming there are very few usages of it).

kratos/python/add_processes_to_python.cpp
```diff
@@ -175,18 +175,23 @@ void  AddProcessesToPython(pybind11::module& m)
...
-    ;
+    using FindGlobalNodalElementalNeighboursProcessType = FindGlobalNodalEntityNeighboursProcess<ModelPart::ElementsContainerType>;
+    py::class_<FindGlobalNodalElementalNeighboursProcessType, typename FindGlobalNodalElementalNeighboursProcessType::Pointer, Process>(m,"FindGlobalNodalElementalNeighboursProcess")
+        .def(py::init([](const DataCommunicator& rDataComm, ModelPart& rModelPart) {
+            KRATOS_WARNING("FindGlobalNodalElementalNeighboursProcess") << "Using deprecated constructor. Please use constructor without DataCommunicator.";
+            return Kratos::make_shared<FindGlobalNodalElementalNeighboursProcessType>(rModelPart);
+        }))
+        .def(py::init<ModelPart&>())
+        .def("ClearNeighbours",&FindGlobalNodalElementalNeighboursProcessType::ClearNeighbours)
+        .def("GetNeighbourIds",&FindGlobalNodalElementalNeighboursProcessType::GetNeighbourIds)
```
rubenzorrilla, line 189: I'm not sure if we really want to expose this... I don't see a clear use case in Python and it goes against the public `Process` API.

kratos/processes/find_global_nodal_entity_neighbours_process.h
```diff
@@ -0,0 +1,177 @@
...
+
+
+    ///@}
+    ///@name Operations
+    ///@{
+
+    void Execute() override;
+
+    void ClearNeighbours();
```
rubenzorrilla:
```suggestion
    void Clear() override;
```
This is the one in the base `Process` API. Then we move the implementation inside the `ClearNeighbours` to `Clear`. I'd keep the `ClearNeighbours` for a while to issue a deprecation warning.

kratos/python/add_processes_to_python.cpp
```diff
@@ -175,18 +175,23 @@ void  AddProcessesToPython(pybind11::module& m)
...
-    .def("GetNeighbourIds",&FindGlobalNodalElementalNeighboursProcess::GetNeighbourIds)
-    ;
+    using FindGlobalNodalElementalNeighboursProcessType = FindGlobalNodalEntityNeighboursProcess<ModelPart::ElementsContainerType>;
+    py::class_<FindGlobalNodalElementalNeighboursProcessType, typename FindGlobalNodalElementalNeighboursProcessType::Pointer, Process>(m,"FindGlobalNodalElementalNeighboursProcess")
+        .def(py::init([](const DataCommunicator& rDataComm, ModelPart& rModelPart) {
+            KRATOS_WARNING("FindGlobalNodalElementalNeighboursProcess") << "Using deprecated constructor. Please use constructor without DataCommunicator.";
+            return Kratos::make_shared<FindGlobalNodalElementalNeighboursProcessType>(rModelPart);
+        }))
+        .def(py::init<ModelPart&>())
+        .def("ClearNeighbours",&FindGlobalNodalElementalNeighboursProcessType::ClearNeighbours)
```
rubenzorrilla:
```suggestion
        .def("Clear",&FindGlobalNodalElementalNeighboursProcessType::Clear)
        .def("ClearNeighbours",&FindGlobalNodalElementalNeighboursProcessType::ClearNeighbours)
```

kratos/python/add_processes_to_python.cpp
```diff
@@ -175,18 +175,23 @@ void  AddProcessesToPython(pybind11::module& m)
...
+        .def("GetNeighbourIds",&FindGlobalNodalElementalNeighboursProcessType::GetNeighbourIds)
+        ;
+
+    using FindGlobalNodalConditionNeighboursProcessType = FindGlobalNodalEntityNeighboursProcess<ModelPart::ConditionsContainerType>;
+    py::class_<FindGlobalNodalConditionNeighboursProcessType, typename FindGlobalNodalConditionNeighboursProcessType::Pointer, Process>(m,"FindGlobalNodalConditionNeighboursProcess")
+        .def(py::init<ModelPart&>())
+        .def("ClearNeighbours",&FindGlobalNodalConditionNeighboursProcessType::ClearNeighbours)
+        .def("GetNeighbourIds",&FindGlobalNodalConditionNeighboursProcessType::GetNeighbourIds)
```
rubenzorrilla: Same comments.

rubenzorrilla (commented):
Minor comments about the API. The implementations are correct IMO.

mpentek:
Oh, it turns out we actually need this as well :-)

kratos/python/add_processes_to_python.cpp
philbucher: I use/need this :)

Actually it is used in several places too

rubenzorrilla: OK. I don't really like but if it is already populated we need to live with that.

kratos/processes/find_global_nodal_entity_neighbours_process.h
```diff
@@ -0,0 +1,173 @@
...
+    ///@{
+
+    /// Constructor.
+    FindGlobalNodalEntityNeighboursProcess(
+        ModelPart& rModelPart);
+
+    FindGlobalNodalEntityNeighboursProcess(
+        ModelPart& rModelPart,
+        const Variable<GlobalEntityPointersVectorType>& rOutputVariable);
+
```
rubenzorrilla, line 83: Maybe it's a good idea to have a Model and Parameters constructor.

kratos/processes/find_global_nodal_entity_neighbours_process.cpp
```diff
@@ -0,0 +1,219 @@
...
+    });
+
+    KRATOS_CATCH("");
+}
+
+template<class TContainerType>
+void FindGlobalNodalEntityNeighboursProcess<TContainerType>::Clear()
+{
+    auto& rNodes = mrModelPart.Nodes();
+    VariableUtils().SetVariable(mrOutputVariable, GlobalEntityPointersVectorType(), rNodes);
```
rubenzorrilla:
```suggestion
    VariableUtils().SetNonHistoricalVariable(mrOutputVariable, GlobalEntityPointersVectorType(), rNodes)

```
According to line 80, shouldn't this be in the non-historical database?

sunethwarna: good catch... may be no one was using `Clear()` until now...

kratos/processes/find_global_nodal_entity_neighbours_process.h
sunethwarna: done :)

kratos/processes/find_global_nodal_entity_neighbours_process.h
sunethwarna: done :)

kratos/processes/find_global_nodal_entity_neighbours_process.h
sunethwarna: done :)

kratos/python/add_processes_to_python.cpp
sunethwarna: done :)

kratos/processes/find_global_nodal_entity_neighbours_process.cpp
rubenzorrilla: Most probably yes...

sunethwarna:
@rubenzorrilla I applied all of your suggestions and fixed the unity build and test problems. Could you have a look?

rubenzorrilla (approved):
Fine for me. Thanks @sunethwarna .

PR #10097, opened 2022-08-03
[Fluid] Compute Y+ Process

sunethwarna:
**📝 Description**
This PR Adds `ComputeYPlusProcess` and its python process output process named "YPlusOutputProcess" which can be used for the fly or post Y+ calculation. Appropriate tests are added to demonstrate the use case.

**🆕 Changelog**
- Added `Y_PLUS` variable to `FluidDynamicsApplication`
- Added `ComputeYPlusProcess` `c++` process
- Added `YPlusOutputProcess` python process
- Added tests


applications/FluidDynamicsApplication/tests/test_compute_y_plus_process.py
```diff
@@ -0,0 +1,135 @@
...
+        cls.model_part.AddNodalSolutionStepVariable(Kratos.REACTION)
+        ReadModelPart("AdjointVMSSensitivity2DTest/cylinder_test", cls.model_part)
+
+        cls.density = 2.0
+        cls.kinematic_viscosity = 3.0
+
+        for element in cls.model_part.Elements:
+            element.Properties[Kratos.DENSITY] = cls.density
+            element.Properties[Kratos.DYNAMIC_VISCOSITY] = cls.kinematic_viscosity * cls.density
+            break
```
matekelemen, line 25: We really need to add a proper container interface for ```PointerVectorSet```.

applications/FluidDynamicsApplication/tests/test_compute_y_plus_process.py
```diff
@@ -0,0 +1,135 @@
...
+        # now populate the reactions
+        for node in cls.model_part.Nodes:
+            node.SetSolutionStepValue(Kratos.REACTION, Kratos.Array3([node.Id, node.Id * 2, node.Id * 3]))
+
+    def testComputeYPlusProcess(self):
+        parameters = Kratos.Parameters("""{
+            "model_part_name"                  : "test",
+            "output_variable_name"             : "Y_PLUS",
+            "output_to_elements"               : true,
+            "calculate_normals_every_time_step": false,
```
matekelemen, line 40: how about something shorter, like ```"update_normals"```?

applications/FluidDynamicsApplication/tests/test_compute_y_plus_process.py
```diff
@@ -0,0 +1,135 @@
...
+            5:0.000277998,
+            6:0.000255058,
+            7:0.000300881,
+            8:0.000300881,
+            9:0.000331303,
+            10:0.000331303,
+        }
+
+        # now check the shear stresses
+        for condition_id, y in ref_y_values.items():
```
matekelemen, line 62: minor thingy, but you can use [```enumerate``` with an offset](https://docs.python.org/3/library/functions.html#enumerate) in this case:
```py
        ref_y_values = {
            0.000255058,
            0.000277998,
            0.000262711,
            0.000262711,
            0.000277998,
            0.000255058,
            0.000300881,
            0.000300881,
            0.000331303,
            0.000331303,
        }

    for condition_id, y in enumerate(ref_y_values, 1):
```

applications/FluidDynamicsApplication/tests/test_compute_y_plus_process.py
```diff
@@ -0,0 +1,135 @@
...
+
+        factory = KratosProcessFactory(self.model)
+        process_list = factory.ConstructListOfProcesses(settings)
+
+        time_steps = [1.0, 1.5, 2.0, 2.5]
+        for process in process_list:
+            process.Check()
+
+        step = 1
+        for time_step in time_steps:
```
matekelemen: again, ```enumerate```

applications/FluidDynamicsApplication/tests/test_compute_y_plus_process.py
```diff
@@ -0,0 +1,135 @@
...
+            self.model_part.ProcessInfo[Kratos.TIME] = time_step
+            self.model_part.ProcessInfo[Kratos.STEP] = step
+            step += 1
+
+            process.ExecuteInitializeSolutionStep()
```
matekelemen: ```process``` is going to be the last item from ```process_list``` from the iteration in line 111, but then it gets updated at the end of the iteration in line 122. I'm guessing ```process_list``` has only 1 item, but this is still a bit nasty.

applications/FluidDynamicsApplication/tests/test_compute_y_plus_process.py
```diff
@@ -0,0 +1,135 @@
...
+            process.ExecuteFinalizeSolutionStep()
+            if process.IsOutputStep():
+                process.PrintOutput()
+
+        for process in process_list:
+            process.ExecuteFinalize()
+
+        kratos_utilities.DeleteFileIfExisting("test_y_plus.dat")
```
matekelemen, line 129: Minor again: I'd put deleting this file in ```setUp``` and ```tearDown``` to avoid complications when tests get interrupted (Ctrl+C, segfaults, etc.)

applications/FluidDynamicsApplication/python_scripts/y_plus_output_process.py
```diff
@@ -0,0 +1,184 @@
...
+
+        default_settings = Kratos.Parameters("""
+            {
+                "model_part_name"                  : "",
+                "interval"                         : [0.0, 1e30],
+                "y_plus_output_limit"              : 1000.0,
+                "print_to_screen"                  : false,
+                "print_format"                     : ".8f",
+                "write_output_file"                : true,
+                "output_step"                      : 8,
```
matekelemen, line 40: ```"output_step"``` is used as a frequency, so the name should indicate that

applications/FluidDynamicsApplication/python_scripts/y_plus_output_process.py
```diff
@@ -0,0 +1,184 @@
...
+    A class responsible for the Y+ output, which is an condition value in Kratos.
+    """
+
+    def __init__(self, model, params):
+        Kratos.OutputProcess.__init__(self)
+
+        default_settings = Kratos.Parameters("""
+            {
+                "model_part_name"                  : "",
+                "interval"                         : [0.0, 1e30],
```
matekelemen: ```IntervalUtility``` can handle ```"End"``` as well
```
"interval" : [0.0, "End"]
```

applications/FluidDynamicsApplication/python_scripts/y_plus_output_process.py
```diff
@@ -0,0 +1,184 @@
...
+
+            if (self.print_to_screen):
+                result_msg = 'y+ evaluation for model part ' + \
+                    self.model_part_name + '\n'
+                result_msg += ', '.join([a+b for a,
+                                            b in zip(res_labels, output_vals)])
+                self._PrintToScreen(result_msg + "%")
+
+            if (self.write_output_file):
+                self.output_file.write(' '.join(output_vals) + "\n")
```
matekelemen, line 141: We're gonna have to change ```TimeBasedAsciiFileWriterUtility``` to support appending files instead of keeping them open for the entire duration of the simulation at some point.

matekelemen (commented):
Fine by me; minor comments only.

applications/FluidDynamicsApplication/tests/test_compute_y_plus_process.py
sunethwarna: nice catch. I did not notice it yet. I'll fix it.

applications/FluidDynamicsApplication/tests/test_compute_y_plus_process.py
sunethwarna: makes sense. will do. 👍

applications/FluidDynamicsApplication/tests/test_compute_y_plus_process.py
sunethwarna: I am ok with both. Historically we have been using `calculate_normals_every_time_step`. Lets see what others have to say.

applications/FluidDynamicsApplication/tests/test_compute_y_plus_process.py
sunethwarna: I kept it like that to indicate that these are each conditions' y values. If we put enumerate, then it kind of gives the impression these will work only for consecutive condition ids.

applications/FluidDynamicsApplication/tests/test_compute_y_plus_process.py
sunethwarna: done :)

applications/FluidDynamicsApplication/tests/test_compute_y_plus_process.py
sunethwarna: done :)

applications/FluidDynamicsApplication/python_scripts/y_plus_output_process.py
sunethwarna: good idea... may be for future 👍

applications/FluidDynamicsApplication/tests/test_compute_y_plus_process.py
matekelemen: ah, allright

applications/FluidDynamicsApplication/tests/test_compute_y_plus_process.py
matekelemen: I guess it shouldn't be changed if this name is used elsewhere too.

applications/FluidDynamicsApplication/python_scripts/y_plus_output_process.py
mpentek: > We're gonna have to change `TimeBasedAsciiFileWriterUtility` to support appending files instead of keeping them open for the entire duration of the simulation at some point.

For this actually I think of it as following: ASCII, VTK and GID output is good for fast postprocessing, even though perhaps not so efficient (either in size or having it open - general file handling). For everything else, there should be for example H5 output.

mtnzguillermo:
I've just tested it and everything looks OK to me. :-)

applications/FluidDynamicsApplication/tests/test_compute_y_plus_process.py
mtnzguillermo: `"calculate_normals_every_time_step"` seems long but it is absolutely clear what it means. I'm not sure if I would understand it with `"update_normals"`. What about a slight improvement of the first one? Maybe `"update_normals_each_step"`.

sunethwarna:
Can someone approve if everything is ok @rubenzorrilla @mpentek @matekelemen @mtnzguillermo ?

matekelemen (approved):
👍

rubenzorrilla:
I've several concerns with this PR.

About the C++ compute Y-plus process, nothing to say. I think that it is a valuable contribution to the application. However, there are no unit tests of it. Note that it only tested through the outputting process, so if it eventually fails debugging would be more complicated. I suggest adding a simple cpp test for the process.

Concerning the outputting process, as far as I see, what it does is to compute the Y-plus to then output some statistics of it. If this is correct, the `FluidDynamicsApplication` is not the proper place for it, specially considering that you require the `StatisticsApplication` to use it (most of CFD users/developers use the `FluidDynamicsApplication` without the statistics one). 

Take into account that the `FluidDynamicsApplication` is a large (so hard to maintain) and key application, so the project requires to keep it as lightweight and controlled as possible.

As a comment aside I'd like to add that, we the @KratosMultiphysics/technical-committee decided to keep the codeowners for notification to make things more agile. However, it is customary to wait for the review of the application responsible(s) before merging, specially in the core applications. Note that we're in the summer vacations period, so reviews might take longer.

PR #10098, opened 2022-08-03
[Core] Adding architecture to the Kratos version

loumalouomega:
**📝 Description**
Adding architecture to the Kratos version. As ARM64 is becoming more popular thanks to Apple, Ampere and Pine64 among others, I think it is relevant to add an architecture in the Kratos version.

**🆕 Changelog**
- Adding architecture to the Kratos version


roigcarlo (approved):
Noice :)

PR #10100, opened 2022-08-03
[StructMech-Eigen] Creating output folder subdirectories

mpentek:
As in the title. Was only working for a single directory, but not a folder structure with subdirectories.

### DIFF
--- a/applications/StructuralMechanicsApplication/custom_processes/postprocess_eigenvalues_process.cpp
+++ b/applications/StructuralMechanicsApplication/custom_processes/postprocess_eigenvalues_process.cpp
@@ -17,6 +17,7 @@
 // External includes
 
 // Project includes
+#include "includes/kratos_filesystem.h"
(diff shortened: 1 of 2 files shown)


loumalouomega (dismissed):

philbucher (commented):
I would do this on the C++ side for better compatibility 

I can do it in the next days

mpentek:
> I would do this on the C++ side for better compatibility
> 
> I can do it in the next days

For output we used to have it on the Python side quite often. I do not mind any of the solutions. You could remove this if you have time to do the other version. But I would keep it now and preferably merge it.

philbucher:
Yes it used to be in python, but we started moving it to C++ since then it can safely also be used from there

Up to you

Suggestion:
You can merge it on Monday if I don't manage until then?

KlausBSautter (approved):

PR #10105, opened 2022-08-04
[Core] Zero element contribution bug in elimination b&s fix

loumalouomega:
**📝 Description**
This bug was found by @pablobecker. When using the elimination builder and solver, if the contribution of the element is zero there is no check that there is no zero in the diagonal as it is done in the block builder and solver, making the system of equations not solvable.

**🆕 Changelog**
- Check the diagonal of the elimination B&S LHS consistently with the block B&S
- Adding B&S test with element of zero contribution
- Clean up test, using KRATOS_CHECK_RELARIVELY_NEAR
- Separate diagonal values calculation in an independent utility, so it can be reused by different B&S
- Adding test for the diagonal value utility
- Refactor diagonal values utility with the parallel and reduction utilities


kratos/utilities/strategies_utilities.h
```diff
@@ -0,0 +1,198 @@
...
+     * @brief This method returns the diagonal norm considering for scaling the diagonal
+     * @param rA The LHS matrix
+     * @return The diagonal norm
+     */
+    template<class TSparseSpace>
+    static double GetDiagonalNorm(typename TSparseSpace::MatrixType& rA)
+    {
+        double diagonal_norm = 0.0;
+        diagonal_norm = IndexPartition<std::size_t>(TSparseSpace::Size1(rA)).for_each<SumReduction<double>>([&](std::size_t Index){
+            return std::pow(rA(Index,Index), 2);
```
philbucher: accessing the `compressed_matrix` with `()` is slow and I am not sure what happens if the entry does not exist (should hopefully not be the case but better safe than sorry)

This is how it should be accessed: https://github.com/KratosMultiphysics/Kratos/blob/c7f5682de5b3c06819005e332d563923f70b27c3/kratos/spaces/ublas_space.h#L331-L334

=> check also the corresponding test I wrote

There are some resources available on google, I think it is worth doing it properly here

@RiccardoRossi maybe you have more suggestions, I think you know this well

loumalouomega: This code is just ported from the BlockBuilderAndSolver: https://github.com/KratosMultiphysics/Kratos/blob/c7f5682de5b3c06819005e332d563923f70b27c3/kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver.h#L1748 and followings

I can take a look to make it faster, but then it will lose generality (it will work only for a specific sparse space)

loumalouomega: BTW, UblasSpace has not being refactored with ParallelUtilities yet

loumalouomega: Optimized!

kratos/utilities/strategies_utilities.h
```diff
@@ -0,0 +1,230 @@
...
+     * @return The diagonal norm
+     */
+    template<class TSparseSpace>
+    static double GetDiagonalNorm(typename TSparseSpace::MatrixType& rA)
+    {
+        double* Avalues = rA.value_data().begin();
+        std::size_t* Arow_indices = rA.index1_data().begin();
+        std::size_t* Acol_indices = rA.index2_data().begin();
+
+        const double diagonal_norm = IndexPartition<std::size_t>(TSparseSpace::Size1(rA)).for_each<SumReduction<double>>([&](std::size_t Index){
```
RiccardoRossi: i cannot do a full review now, however this sprcific function can be made faster by taking the norm of the values vector without looking into cols and rows.

we might even want to use std parallel functions to do it in parallel since we are now using c++17

loumalouomega: Right now I am moving to an independent utility the code to be reused, I don't want to do a huge refactor (only the parts I fell trivial), I would refactor that functions properly in a future PR.

loumalouomega: BTW, currently still using C++11, until no official migration I will not use C++17

loumalouomega: We are now using C++17, what was your idea @RiccardoRossi ?

kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_lagrange_multiplier.h
```diff
@@ -501,7 +501,7 @@ class ResidualBasedBlockBuilderAndSolverWithLagrangeMultiplier
                 TSystemMatrixType A(BaseType::mEquationSystemSize, BaseType::mEquationSystemSize);
                 BaseType::ConstructMatrixStructure(pScheme, A, rModelPart);
                 this->BuildLHS(pScheme, rModelPart, A);
-                const double constraint_scale_factor = mConstraintFactorConsidered == CONSTRAINT_FACTOR::CONSIDER_NORM_DIAGONAL_CONSTRAINT_FACTOR ? this->GetDiagonalNorm(A) : this->GetDiagonalNorm(A);
+                const double constraint_scale_factor = mConstraintFactorConsidered == CONSTRAINT_FACTOR::CONSIDER_NORM_DIAGONAL_CONSTRAINT_FACTOR ? StrategiesUtilities::GetMaxDiagonal<TSparseSpace>(A) : StrategiesUtilities::GetDiagonalNorm<TSparseSpace>(A);
```
RiccardoRossi: this is a behaviour change no?

kratos/utilities/strategies_utilities.h
```diff
@@ -0,0 +1,228 @@
...
+     */
+    template<class TSparseSpace>
+    static double GetMaxDiagonal(typename TSparseSpace::MatrixType& rA)
+    {
+        double* Avalues = rA.value_data().begin();
+        std::size_t* Arow_indices = rA.index1_data().begin();
+        std::size_t* Acol_indices = rA.index2_data().begin();
+
+        return IndexPartition<std::size_t>(TSparseSpace::Size1(rA)).for_each<MaxReduction<double>>([&](std::size_t Index){
+            const std::size_t col_begin = Arow_indices[Index];
```
RiccardoRossi: just wondering (i don't know the answer) if for this it would not be faster to do the following

~~~cpp
return IndexPartition<std::size_t>(TSparseSpace::Size1(rA)).for_each<MaxReduction<double>>([&](std::size_t Index){
            const std::size_t col_size = Arow_indices[Index+1] - Arow_indices[Index];
            if(col_size>0) {
                    return std::abs(Avalues(i,i));
                }
            else
                   return std::numeric_limits<double>::max();
            }
        });
~~~

kratos/utilities/strategies_utilities.h
RiccardoRossi: i thought we were computing the norm of them matrix, smthg for which we do not look into the js

@RiccardoRossi I cannot answer you if you post it as this. I don't understand what you say. In any case this preserves current behaviour


yes, i am saying that my comment was wrong

kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_lagrange_multiplier.h
loumalouomega: No, it is the same, but calling an external utility instead of a private method. This way can be reused among different B&S

kratos/utilities/strategies_utilities.h
loumalouomega: Huuum, I can try

kratos/utilities/strategies_utilities.h
loumalouomega: What is Js?

kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_lagrange_multiplier.h
RiccardoRossi: look at what is called in the if

loumalouomega: You are right, but if you see the previous one was wrong: `this->GetDiagonalNorm(A) : this->GetDiagonalNorm(A);`

loumalouomega: This is correcting it

kratos/utilities/strategies_utilities.h
loumalouomega: Nope, doesn't work

RiccardoRossi: ? the (i,j) methos is there ...

loumalouomega: Maybe is because of the max, should be 0?

loumalouomega: Let me see

loumalouomega: Now it works

RiccardoRossi: the point is which one is faster (this is what i don't know)

kratos/utilities/strategies_utilities.h
RiccardoRossi: the columns. But anyhow, i am telling that my comment was wrong

kratos/utilities/strategies_utilities.h
loumalouomega: I think depends on the implementation of `A(i,j)`, worst case scenario should be the same, as we were just iterating one by one

kratos/utilities/strategies_utilities.h
loumalouomega: Ah Ok, i->row, j->column.

kratos/utilities/strategies_utilities.h
loumalouomega: Nope, https://www.boost.org/doc/libs/1_67_0/boost/numeric/ublas/matrix_sparse.hpp, previous should be faster

kratos/utilities/strategies_utilities.cpp
```diff
@@ -0,0 +1,163 @@
...
+/***********************************************************************************/
+
+template<>
+double StrategiesUtilities::GetMinDiagonal<UblasSparseSpaceType>(UblasSparseSpaceType::MatrixType& rA)
+{
+    double* Avalues = rA.value_data().begin();
+    std::size_t* Arow_indices = rA.index1_data().begin();
+    std::size_t* Acol_indices = rA.index2_data().begin();
+
+    return IndexPartition<std::size_t>(UblasSparseSpaceType::Size1(rA)).for_each<MinReduction<double>>([&](std::size_t Index){
```
RiccardoRossi: given that the implementation is essentially the same alwsys, wouldn't it make sense to template  in "functor_type" to be max,min,abs, etc?

kratos/utilities/strategies_utilities.h
RiccardoRossi: no, i would not use iterators. access by row_arrays is the fastest and the most portable...and it will still work once we pass to the new matrix

kratos/utilities/strategies_utilities.h
```diff
@@ -0,0 +1,228 @@
...
+            const std::size_t col_end = Arow_indices[Index+1];
+            for (std::size_t j = col_begin; j < col_end; ++j) {
+                if (Acol_indices[j] == Index ) {
+                    return std::abs(Avalues[j]);
+                }
+            }
+            return 0.0;
```
philbucher:
```suggestion
            return std::numeric_limits<double>::lowest();
```

otherwise it is wrong if you only have negative vals on the diagonal!

kratos/utilities/strategies_utilities.h
```diff
@@ -0,0 +1,228 @@
...
+
+        // Detect if there is a line of all zeros and set the diagonal to a 1 if this happens
+        IndexPartition<std::size_t>(system_size).for_each([&](std::size_t Index){
+            bool empty = true;
+
+            const std::size_t col_begin = Arow_indices[Index];
+            const std::size_t col_end = Arow_indices[Index + 1];
+
+            for (std::size_t j = col_begin; j < col_end; ++j) {
+                if(Avalues[j] != 0.0) {
```
philbucher: please check against epsilon

kratos/solving_strategies/builder_and_solvers/residualbased_elimination_builder_and_solver.h
```diff
@@ -1034,9 +1041,13 @@ class ResidualBasedEliminationBuilderAndSolver
     ///@{
 
 #ifdef USE_LOCKS_IN_ASSEMBLY
-   std::vector<omp_lock_t> mLockArray;
+   std::vector<omp_lock_t> mLockArray; /// TODO: DISCUSS TO REMOVE THIS AS IN THE BlockBuilderAndSolver
```
philbucher: whether or not it will be removed, this should be `LockObject`

kratos/utilities/strategies_utilities.h
```diff
@@ -0,0 +1,228 @@
...
+    /**
+     * @brief This method returns the scale norm considering for scaling the diagonal
+     * @param rModelPart The problem model part
+     * @param rA The LHS matrix
+     * @param ScalingDiagonal The type of caling diagonal considered
+     * @return The scale norm
+     */
+    template<class TSparseSpace>
+    static double GetScaleNorm(
+        ModelPart& rModelPart,
```
philbucher: the modelpart is not used, only the process info, hence the signature can be reduced (and make it const)

kratos/utilities/strategies_utilities.h
```diff
@@ -0,0 +1,228 @@
...
+     * @brief This method returns the diagonal max value
+     * @param rA The LHS matrix
+     * @return The diagonal  max value
+     */
+    template<class TSparseSpace>
+    static double GetMaxDiagonal(typename TSparseSpace::MatrixType& rA)
+    {
+        double* Avalues = rA.value_data().begin();
+        std::size_t* Arow_indices = rA.index1_data().begin();
+        std::size_t* Acol_indices = rA.index2_data().begin();
```
philbucher:
```suggestion
        const double* Avalues = rA.value_data().begin();
        const std::size_t* Arow_indices = rA.index1_data().begin();
        const std::size_t* Acol_indices = rA.index2_data().begin();
```

kratos/utilities/strategies_utilities.h
```diff
@@ -0,0 +1,228 @@
...
+     * @brief This method returns the diagonal min value
+     * @param rA The LHS matrix
+     * @return The diagonal min value
+     */
+    template<class TSparseSpace>
+    static double GetMinDiagonal(typename TSparseSpace::MatrixType& rA)
+    {
+        double* Avalues = rA.value_data().begin();
+        std::size_t* Arow_indices = rA.index1_data().begin();
+        std::size_t* Acol_indices = rA.index2_data().begin();
```
philbucher:
```suggestion
        const double* Avalues = rA.value_data().begin();
        const std::size_t* Arow_indices = rA.index1_data().begin();
        const std::size_t* Acol_indices = rA.index2_data().begin();
```

kratos/utilities/strategies_utilities.h
```diff
@@ -0,0 +1,228 @@
...
+        });
+    }
+
+    /**
+     * @brief This method returns the diagonal min value
+     * @param rA The LHS matrix
+     * @return The diagonal min value
+     */
+    template<class TSparseSpace>
+    static double GetMinDiagonal(typename TSparseSpace::MatrixType& rA)
```
philbucher:
```suggestion
    static double GetMinDiagonal(const typename TSparseSpace::MatrixType& rA)
```

kratos/utilities/strategies_utilities.h
```diff
@@ -0,0 +1,228 @@
...
+        return 0.5 * (GetMaxDiagonal<TSparseSpace>(rA) + GetMinDiagonal<TSparseSpace>(rA));
+    }
+
+    /**
+     * @brief This method returns the diagonal max value
+     * @param rA The LHS matrix
+     * @return The diagonal  max value
+     */
+    template<class TSparseSpace>
+    static double GetMaxDiagonal(typename TSparseSpace::MatrixType& rA)
```
philbucher:
```suggestion
    static double GetMaxDiagonal(const typename TSparseSpace::MatrixType& rA)
```

same for the other functions

kratos/solving_strategies/builder_and_solvers/residualbased_elimination_builder_and_solver.h
```diff
@@ -142,6 +144,7 @@ class ResidualBasedEliminationBuilderAndSolver
         typename TLinearSolver::Pointer pNewLinearSystemSolver)
         : BaseType(pNewLinearSystemSolver)
     {
+        mScalingDiagonal = SCALING_DIAGONAL::NO_SCALING;
```
philbucher:
```suggestion

```

please initialize this when declaring the member

kratos/solving_strategies/builder_and_solvers/residualbased_elimination_builder_and_solver.h
```diff
@@ -1034,9 +1041,13 @@ class ResidualBasedEliminationBuilderAndSolver
     ///@{
 
 #ifdef USE_LOCKS_IN_ASSEMBLY
-   std::vector<omp_lock_t> mLockArray;
+   std::vector<omp_lock_t> mLockArray; /// TODO: DISCUSS TO REMOVE THIS AS IN THE BlockBuilderAndSolver
 #endif
 
+    double mScaleFactor = 1.0;         /// The manually set scale factor
+
+    SCALING_DIAGONAL mScalingDiagonal; /// We identify the scaling considered for the dirichlet dofs
```
philbucher:
```suggestion
    SCALING_DIAGONAL mScalingDiagonal = SCALING_DIAGONAL::NO_SCALING;; /// We identify the scaling considered for the dirichlet dofs

```

kratos/utilities/strategies_utilities.h
```diff
@@ -0,0 +1,228 @@
...
+        ModelPart& rModelPart,
+        typename TSparseSpace::MatrixType& rA,
+        typename TSparseSpace::VectorType& rb,
+        const SCALING_DIAGONAL ScalingDiagonal = SCALING_DIAGONAL::NO_SCALING
+        )
+    {
+        const std::size_t system_size = rA.size1();
+
+        double* Avalues = rA.value_data().begin();
+        std::size_t* Arow_indices = rA.index1_data().begin();
```
philbucher:
```suggestion
        const double* Avalues = rA.value_data().begin();
        const std::size_t* Arow_indices = rA.index1_data().begin();
```

kratos/utilities/strategies_utilities.h
```diff
@@ -0,0 +1,228 @@
...
+            case SCALING_DIAGONAL::CONSIDER_PRESCRIBED_DIAGONAL: {
+                const ProcessInfo& r_current_process_info = rModelPart.GetProcessInfo();
+                KRATOS_ERROR_IF_NOT(r_current_process_info.Has(BUILD_SCALE_FACTOR)) << "Scale factor not defined at process info" << std::endl;
+                return r_current_process_info.GetValue(BUILD_SCALE_FACTOR);
+            }
+            case SCALING_DIAGONAL::CONSIDER_NORM_DIAGONAL:
+                return GetDiagonalNorm<TSparseSpace>(rA)/static_cast<double>(rA.size1());
+            case SCALING_DIAGONAL::CONSIDER_MAX_DIAGONAL:
+                return GetMaxDiagonal<TSparseSpace>(rA);
+//                 return TSparseSpace::TwoNorm(rA)/static_cast<double>(rA.size1());
```
philbucher: please remove
```suggestion
```

kratos/utilities/strategies_utilities.h
```diff
@@ -0,0 +1,228 @@
...
+     * @brief This method returns the diagonal norm considering for scaling the diagonal
+     * @param rA The LHS matrix
+     * @return The diagonal norm
+     */
+    template<class TSparseSpace>
+    static double GetDiagonalNorm(typename TSparseSpace::MatrixType& rA)
+    {
+        double* Avalues = rA.value_data().begin();
+        std::size_t* Arow_indices = rA.index1_data().begin();
+        std::size_t* Acol_indices = rA.index2_data().begin();
```
philbucher:
```suggestion
        const double* Avalues = rA.value_data().begin();
        const std::size_t* Arow_indices = rA.index1_data().begin();
        const std::size_t* Acol_indices = rA.index2_data().begin();
```

philbucher (changes requested):
Looks pretty good to me in terms of implementation

I added some comments on the code structure

Also one big comment: Since this is neither compatible with the trilinos matrix nor the Kratos matrix, can you please move the new functions to the `ublas_space`? This I think is a much better place that already contains similar functionalities

kratos/utilities/strategies_utilities.h
RiccardoRossi: i think there is an abs on line 194

RiccardoRossi (commented):
overall i would be ok with this refactor, even though admittedly i am afraid of this kind of changes in a foundational class as the builder and solver.

take a look at my comments, but do not feel obliged to incorporate them if you do not consider.

i let @philbucher  approve

kratos/utilities/strategies_utilities.h
loumalouomega: There are abs, yes. These functions already exist in the Block B&S

kratos/utilities/strategies_utilities.h
loumalouomega: Okay, but this is behaviour change

loumalouomega:
@philbucher please, for code changes use standard comments instead of reviews, so I can do batch commit

loumalouomega:
Done @philbucher

kratos/spaces/ublas_space.h
```diff
@@ -689,25 +679,174 @@ class UblasSpace
...
+
+        return IndexPartition<std::size_t>(Size1(rA)).for_each<MaxReduction<double>>([&](std::size_t Index){
+            const std::size_t col_begin = Arow_indices[Index];
+            const std::size_t col_end = Arow_indices[Index+1];
+            for (std::size_t j = col_begin; j < col_end; ++j) {
+                if (Acol_indices[j] == Index ) {
+                    return std::abs(Avalues[j]);
+                }
+            }
+            return 0.0;
```
philbucher:
```suggestion
            return std::numeric_limits<double>::lowest();
```

kratos/spaces/ublas_space.h
```diff
@@ -689,25 +679,174 @@ class UblasSpace
...
+        const std::size_t* Arow_indices = rA.index1_data().begin();
+
+        // Define  zero value tolerance
+        const double zero_tolerance = std::numeric_limits<double>::epsilon();
+
+        // The diagonal considered
+        const double scale_factor = GetScaleNorm(rProcessInfo, rA, ScalingDiagonal);
+
+        // Detect if there is a line of all zeros and set the diagonal to a 1 if this happens
+        IndexPartition<std::size_t>(system_size).for_each([&](std::size_t Index){
```
philbucher:
```suggestion
        IndexPartition(system_size).for_each([&](std::size_t Index){
```

FYI this is possible with C++17
Class template argument deduction (CTAD)

philbucher (approved):

RiccardoRossi:
but...you may have a perfectly valid system with zeros in the diagonal. That is the case for example of a lagrange multiplier.

what in that case?

loumalouomega:
> but...you may have a perfectly valid system with zeros in the diagonal. That is the case for example of a lagrange multiplier.
> 
> what in that case?

This PR is already merged, in the serial block builder and solver we have been doing this check during long time.

loumalouomega:
> > but...you may have a perfectly valid system with zeros in the diagonal. That is the case for example of a lagrange multiplier.
> > what in that case?
> 
> This PR is already merged, in the serial block builder and solver we have been doing this check during long time.

In respect to the LM, see the comment in the code, it is only added if all the row is all zeros, in the case of LM the row is not all zeros.

~~~c++
            // Detect if there is a line of all zeros and set the diagonal to a 1 if this happens
            #pragma omp parallel for
            for(int k = 0; k < static_cast<int>(mDoFToSolveSystemSize); ++k) {
                const IndexType col_begin = Arow_indices[k];
                const IndexType col_end = Arow_indices[k+1];
                bool empty = true;
                for (IndexType j = col_begin; j < col_end; ++j) {
                    if(std::abs(Avalues[j]) > zero_tolerance) {
                        empty = false;
                        break;
                    }
              }
~~~

RiccardoRossi:
ok

loumalouomega:
> ok

The point is that this checks lacks in the Block builder and solver in Trilinos, and I was asking you the permission to implement it

philbucher:
> The point is that this checks lacks in the Block builder and solver in Trilinos, and I was asking you the permission to implement it

sounds good, but remember that the trilinos version does not yet support MPCs yet (Aditya implemented it but never pushed it afaik in his branch) not sure if it would also work outside of the Chimera stuffs

loumalouomega:
> > The point is that this checks lacks in the Block builder and solver in Trilinos, and I was asking you the permission to implement it
> 
> sounds good, but remember that the trilinos version does not yet support MPCs yet (Aditya implemented it but never pushed it afaik in his branch) not sure if it would also work outside of the Chimera stuffs

That's what I am working right now, and this change would be a transition change. If @RiccardoRossi gives me green light, I will implement it

RiccardoRossi:
well, my only comment is that implementing this on the topof the "new" DistributedCSR would be easier...

loumalouomega:
> well, my only comment is that implementing this on the topof the "new" DistributedCSR would be easier...

Do we have any estimated delivery, we would like to have it ASAP

RiccardoRossi:
the DistributedCSR is there and working, what we miss is a builder and solver.

the tests have a lot of examples of usage

loumalouomega:
Do you suggest implementing the Builder and Solver directly in DistributedCSR ?, the problem is that for MPC we need sparse matrix multiplication and ExpetraExt provides already that functionality.

RiccardoRossi:
also the DistributedCSR provides that capability. if you check the tests there are examples of SpMM

loumalouomega:
> also the DistributedCSR provides that capability. if you check the tests there are examples of SpMM

I will discuss with my team, "we are in a hurry"

Issue #10007, opened 2022-07-10
CoSim important FSI tests are disabled


philbucher:
@KratosMultiphysics/cosimulation the Mok-FSI tests are currently disabled (see [here](https://github.com/KratosMultiphysics/Kratos/commit/42b62c1ff7dd709777de2052ed221338190f46f7) and [here](https://github.com/KratosMultiphysics/Kratos/commit/0b853675c8e5e360f629c7511a4e923de19e6dca))
https://github.com/KratosMultiphysics/Kratos/blob/b9769b6578ba5cd9472e4834ad6cf4c193bc9501/applications/CoSimulationApplication/tests/test_CoSimulationApplication.py#L81-L83

@roigcarlo according to the history you made these changes? Those tests are probably the most important tests in the app, why did you disable them?

roigcarlo:
Those were disabled when we set nightly as the default suit because they were faulty (they were failing 50% of times they were executed) , they can be re-enable as long as they are fixed.

philbucher:
Ah I thought that was not merged because of failing tests 😅

In any case, out of all tests in Kratos those two were the ones I trusted most
If they are failing then I am a bit scared, as they test the entire workflow

@KratosMultiphysics/cosimulation would someone habe time to run them to see if they fail?

mpentek:
@roigcarlo it is pretty concerning it these were turned off
@philbucher I will try to take a look asap @mtnzguillermo this might effect you as well

mpentek:
Hopefully fixed with #10108.

PR #10108, opened 2022-08-05
[CoSim] Fix python deprecation and re-enable test

mpentek:
It fixes some python deprecation warnings, which were perhaps triggering the failure of tests with GCC. Re-enabling tests.

These were the only fixes I could see necessary doing. I could not trigger an actual failure, but there were some obvious warnings needing these remedies.

As a result of the discussion in #10007.

### DIFF
--- a/applications/CoSimulationApplication/tests/test_CoSimulationApplication.py
+++ b/applications/CoSimulationApplication/tests/test_CoSimulationApplication.py
@@ -79,8 +79,8 @@ def AssembleTestSuites():
     nightSuite.addTests(KratosUnittest.TestLoader().loadTestsFromTestCases([TestSmallCoSimulationCases]))
     
     # This one has errors in GCC
-    # nightSuite.addTest(TestMokFSI('test_mok_fsi_mvqn'))
-    # nightSuite.addTest(TestMokFSI('test_mok_fsi_aitken'))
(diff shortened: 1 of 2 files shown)


applications/CoSimulationApplication/tests/test_CoSimulationApplication.py
```diff
@@ -79,8 +79,8 @@ def AssembleTestSuites():
     nightSuite.addTests(KratosUnittest.TestLoader().loadTestsFromTestCases([TestSmallCoSimulationCases]))
     
     # This one has errors in GCC
```
philbucher, line 81:
```suggestion
    # NO MATTER WHAT, DO NOT DISABLE THESE TESTS, THEY ARE VITAL FOR THE COSIMAPP!!!!!!!!!!
```

philbucher (approved):
awesome thx!

mpentek:
@roigcarlo Do you agree with this? I cannot trigger any errors on my local machine, when compiling with GCC. If these were to occur again during nightly, I would address it ASAP.

roigcarlo:
Hi, as long as they don't crash or are marked as expected failure OK to me. I am on vacations until Monday so I had no change to take a deep look, but seems OK to me if you want to merge.

PR #10114, opened 2022-08-08
[Core] Registered Component Names

matekelemen:
## Motivation
Currently, there's no good way of getting a list of flags' names from ```KratosComponents```, which would be a nice feature to have for the default behaviour of ```OutputProcess```es.

## Changes
- Add a light ```Range``` class, representing a view of an existing container.
- Add ```KratosComponents::GetComponentNames``` that creates a ```Range``` of the registered components' names.
- Add python interface for getting flags' names.

### DIFF
--- a/kratos/includes/kratos_components.h
+++ b/kratos/includes/kratos_components.h
@@ -28,6 +28,7 @@
 #include "containers/flags.h"
 #include "containers/variable.h"
 #include "utilities/quaternion.h"
+#include "utilities/ranges.h"
 
 namespace Kratos
 {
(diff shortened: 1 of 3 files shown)


loumalouomega (commented):
Nice contribution

loumalouomega:
> Nice contribution

@roigcarlo and @pooyan-dadvand

pooyan-dadvand:
We are migrating the Kratos components to a new registry system. Let's add this one to the new registry.

matekelemen:
Sounds good to me. I'll do it next week and update you when I'm done.

matekelemen:
@pooyan-dadvand I hear ```Registry``` is getting worked on and is not in its final form. Can we add this in the meantime and convert it later, or will ```Registry``` be completed soon and we should wait for that?

pooyan-dadvand:
We are working on it but it is not very far from being functional. I would wait (may be some weeks)

PR #10121, opened 2022-08-09
[Core] Some improvement to allow vectorized operations in VariableUtils

RiccardoRossi:
**📝 Description**

Adding functions to get coordinates in initial and final positions for all nodes and to get and set vector values

**🆕 Changelog**

- Adding functions to get coordinates in initial and final positions for all nodes and to get and set vector values in VariableUtils
- Export functions to python


kratos/python/add_variable_utils_to_python.cpp
```diff
@@ -452,6 +452,63 @@ void AddVariableUtilsToPython(pybind11::module &m)
         .def("ApplyFixity", ApplyFixity<Variable<double>>)
         .def("ApplyFixity", ApplyFlaggedFixity<Variable<double>>)
         .def("ApplyVector", &VariableUtils::ApplyVector<Variable<double>>)
+        .def("ApplyVector", []( VariableUtils& rVariableUtils,
```
loumalouomega: I would define it independently, otherwise a big chuck of code discontinuous with the rest of the code

kratos/python/add_variable_utils_to_python.cpp
```diff
@@ -452,6 +452,63 @@ void AddVariableUtilsToPython(pybind11::module &m)
...
+                                )
+                {
+                    if(rNodes.size() != 0 && rNodes.size()*Dim == rData.size()) {
+                        // First we do a check
+                        rVariableUtils.CheckVariableExists(rVar, rNodes);
+
+                        IndexPartition<std::size_t>(rNodes.size()).for_each([&](std::size_t index){
+                            auto it_node = rNodes.begin() + index;
+                            auto& v = it_node->FastGetSolutionStepValue(rVar);
+                            for(unsigned int k=0; k<Dim; k++)
```
loumalouomega: Why do you need dim, when you can deduce the dime from the vector?

loumalouomega: Ah, Ok, because the data contains the values for all the nodes.... But in that case you can deduce the size from the fact that you are not considering Variable<Vector>, but  array_1d<double, 3>, and therefor the size must be 3. Except if you want to customize for 2D vectors...

kratos/python/add_variable_utils_to_python.cpp
```diff
@@ -452,6 +452,63 @@ void AddVariableUtilsToPython(pybind11::module &m)
...
+        .def("SetSolutionStepValuesVector", []( VariableUtils& rVariableUtils, 
+                                const Variable<array_1d<double,3>>& rVar,
+                                unsigned int Step,
+                                const Vector& rData,
+                                ModelPart::NodesContainerType& rNodes
+                                )
+                {
+                    KRATOS_ERROR_IF(rData.size()%rNodes.size()!=0) << "incompatible number of nodes and position data" << std::endl;
+
+                    unsigned int Dimension = rData.size()/rNodes.size();
```
loumalouomega: const when possible
```suggestion
                    const unsigned int Dimension = rData.size()/rNodes.size();
```

kratos/utilities/variable_utils.cpp
```diff
@@ -371,6 +371,79 @@ void VariableUtils::WeightedAccumulateVariableOnNodes(
     KRATOS_CATCH("");
 }
 
+    ///This function returns the CURRENT coordinates of all the nodes in a consecutive vector.
+    ///in case Dimension == 1 the vector is in the form (X X X ....)
+    ///in case Dimension == 2 the vector is in the form (X Y X Y X Y ....)
+    ///in case Dimension == 3 the vector is in the form (X Y Z X Y Z ....)
+    KRATOS_API(KRATOS_CORE) Vector VariableUtils::GetCurrentPositionsVector(
```
loumalouomega: A priori the API is defined in the header. @roigcarlo ?

kratos/utilities/variable_utils.cpp
```diff
@@ -371,6 +371,79 @@ void VariableUtils::WeightedAccumulateVariableOnNodes(
     KRATOS_CATCH("");
 }
 
+    ///This function returns the CURRENT coordinates of all the nodes in a consecutive vector.
```
loumalouomega: Description is defined in header

kratos/utilities/variable_utils.cpp
```diff
@@ -371,6 +371,79 @@ void VariableUtils::WeightedAccumulateVariableOnNodes(
...
+    ///in case Dimension == 1 the vector is in the form (X X X ....)
+    ///in case Dimension == 2 the vector is in the form (X Y X Y X Y ....)
```
loumalouomega: Okay, now I see the dimension thing

kratos/utilities/variable_utils.cpp
```diff
@@ -371,6 +371,79 @@ void VariableUtils::WeightedAccumulateVariableOnNodes(
...
+        return pos;
+    }
+
+    KRATOS_API(KRATOS_CORE) void VariableUtils::SetCurrentPositionsVector(
+        const ModelPart::NodesContainerType& rNodes, 
+        const Vector& rPositions)
+    {
+        KRATOS_ERROR_IF(rPositions.size()%rNodes.size()!=0) << "incompatible number of nodes and position data" << std::endl;
+
+        unsigned int Dimension = rPositions.size()/rNodes.size();
```
loumalouomega:
```suggestion
        const unsigned int Dimension = rPositions.size()/rNodes.size();
```

kratos/utilities/variable_utils.cpp
```diff
@@ -371,6 +371,79 @@ void VariableUtils::WeightedAccumulateVariableOnNodes(
...
+            }
+        );
+        return pos;
+    }
+
+    KRATOS_API(KRATOS_CORE) void VariableUtils::SetCurrentPositionsVector(
+        const ModelPart::NodesContainerType& rNodes, 
+        const Vector& rPositions)
+    {
+        KRATOS_ERROR_IF(rPositions.size()%rNodes.size()!=0) << "incompatible number of nodes and position data" << std::endl;
```
loumalouomega:
```suggestion
        KRATOS_ERROR_IF(rPositions.size()%rNodes.size()!=0) << "Incompatible number of nodes and position data" << std::endl;
```

kratos/utilities/variable_utils.cpp
```diff
@@ -371,6 +371,79 @@ void VariableUtils::WeightedAccumulateVariableOnNodes(
...
+            }
+        );
+        return pos;
+    }
+
+    KRATOS_API(KRATOS_CORE) Vector VariableUtils::GetInitialPositionsVector(
+        const ModelPart::NodesContainerType& rNodes, 
+        unsigned int Dimension)
+    {
+        KRATOS_ERROR_IF((Dimension>3)) << " only Dimension<=3 is admitted by the function" << std::endl;
```
loumalouomega:
```suggestion
        KRATOS_ERROR_IF((Dimension>3)) << "Only Dimension<=3 is admitted by the function" << std::endl;
```

kratos/utilities/variable_utils.cpp
```diff
@@ -371,6 +371,79 @@ void VariableUtils::WeightedAccumulateVariableOnNodes(
...
 
+    ///This function returns the CURRENT coordinates of all the nodes in a consecutive vector.
+    ///in case Dimension == 1 the vector is in the form (X X X ....)
+    ///in case Dimension == 2 the vector is in the form (X Y X Y X Y ....)
+    ///in case Dimension == 3 the vector is in the form (X Y Z X Y Z ....)
+    KRATOS_API(KRATOS_CORE) Vector VariableUtils::GetCurrentPositionsVector(
+        const ModelPart::NodesContainerType& rNodes, 
+        unsigned int Dimension)
+    {
+        KRATOS_ERROR_IF((Dimension>3)) << " only Dimension<=3 is admitted by the function" << std::endl;
```
loumalouomega:
```suggestion
        KRATOS_ERROR_IF((Dimension>3)) << "Only Dimension<=3 is admitted by the function" << std::endl;
```

kratos/utilities/variable_utils.cpp
```diff
@@ -371,6 +371,79 @@ void VariableUtils::WeightedAccumulateVariableOnNodes(
...
+                coords[k] = rPositions[i*Dimension+k];
+            }
+        );
+    }
+
+    KRATOS_API(KRATOS_CORE) void VariableUtils::SetInitialPositionsVector(
+        const ModelPart::NodesContainerType& rNodes, 
+        const Vector& rPositions)
+    {
+        KRATOS_ERROR_IF(rPositions.size()%rNodes.size()!=0) << "incompatible number of nodes and position data" << std::endl;
```
loumalouomega:
```suggestion
        KRATOS_ERROR_IF(rPositions.size()%rNodes.size()!=0) << "Incompatible number of nodes and position data" << std::endl;
```

kratos/utilities/variable_utils.cpp
```diff
@@ -371,6 +371,79 @@ void VariableUtils::WeightedAccumulateVariableOnNodes(
...
+
+        unsigned int Dimension = rPositions.size()/rNodes.size();
```
loumalouomega:
```suggestion
        const unsigned int Dimension = rPositions.size()/rNodes.size();
```

kratos/utilities/variable_utils.h
```diff
@@ -1389,6 +1389,33 @@ class KRATOS_API(KRATOS_CORE) VariableUtils
         const IndexType BufferPosition = 0
         );
 
+    ///This function returns the CURRENT coordinates of all the nodes in a consecutive vector.
```
loumalouomega: Can you use the style of the rest of Doxygen doc?

kratos/utilities/variable_utils.h
```diff
@@ -1389,6 +1389,33 @@ class KRATOS_API(KRATOS_CORE) VariableUtils
...
+    ///in case Dimension == 1 the vector is in the form (X X X ....)
+    ///in case Dimension == 2 the vector is in the form (X Y X Y X Y ....)
+    ///in case Dimension == 3 the vector is in the form (X Y Z X Y Z ....)
+    Vector GetCurrentPositionsVector(
+        const ModelPart::NodesContainerType& rNodes, 
+        unsigned int Dimension);
```
loumalouomega: Make it const (and the others)

~~~
const unsigned int Dimension);
~~~

loumalouomega (changes requested):
Some comments

kratos/python/add_variable_utils_to_python.cpp
philbucher: those functions already exist in `AuxiliarModelPartUtilities`

kratos/python/add_variable_utils_to_python.cpp
miguelmaso: I'd extract the Dimension from `rData.size() / rNodes.size()`. Having `Dim` as an argument seems superfluous.

kratos/python/add_variable_utils_to_python.cpp
```diff
@@ -452,6 +452,63 @@ void AddVariableUtilsToPython(pybind11::module &m)
...
+                            auto& v = it_node->FastGetSolutionStepValue(rVar);
+                            for(unsigned int k=0; k<Dim; k++)
+                                v[k] = rData[index*Dim+k];
+                        });
+                    } else
+                        KRATOS_ERROR  << "There is a mismatch between the size of data array and the number of nodes ";
+                })
+        .def("GetSolutionStepValuesVector", []( VariableUtils& rVariableUtils, 
+                                const Variable<array_1d<double,3>>& rVar,
+                                unsigned int Dimension,
```
miguelmaso: For the sake of consistency, I'd remove the Dimension and always use 3. See `Node<3>`, `SetVariableToZero`, `array_1d<double,3>`, etc.

At least, I'd move `Dimension` to the last position and set `py::arg("Dimension") = 3`.

philbucher (changes requested):

kratos/python/add_variable_utils_to_python.cpp
RiccardoRossi: no, i am adding the dim on purpose. the point is that when i am in 2D i want to obtain only the xy component. this is very useful for example to write a rotation in 2D

kratos/utilities/variable_utils.h
RiccardoRossi: i don't really follow ... you mean /** **/ ?

kratos/python/add_variable_utils_to_python.cpp
RiccardoRossi: no, i explicitly do not want to do this. the point is that i want to be able to extract only the x y components if i am in 2D.

this is really useful because  this way one can write vectorized operations without having to operate on the third component. 

for example this allows me to write a tool with multiple relative rotations around the axis Z (i need it for a project but it is quite specific as)

~~~py
# Importing the Kratos Library
import KratosMultiphysics
import math
import numpy as np

def Factory(settings, Model):
    if not isinstance(settings, KratosMultiphysics.Parameters):
    return RotationHelperProcess(Model, settings["Parameters"])

class RotationHelperProcess(KratosMultiphysics.Process):
    """This process applies a rotation around the axis Z to all of the subparts involved in the project.
    Then each of the subparts is also allowed to have a rotation around its own axis (and possibly at different roational speed)
    """

    def ComputeRotationMatrix(self,theta):
        s = np.sin(theta)
        c = np.cos(theta)
        R = np.array([[c, -s],[s,c]])
        return R

    def ApplyRotation(self,xc0, R, nodes, disp_var):

        var_utils = KratosMultiphysics.VariableUtils();
        #[n1.x0 n1.y0  n2.x0 n2.y0  n3.x0 n3.y0 ....]
        dx = np.array(var_utils.GetInitialPositionsVector(nodes, 2)) #, copy=False)

        #[[n1.x0 n1.y0  
        #  n2.x0 n2.y0
        #  n3.x0 n3.y0 
        #  ....]
        dx = dx.reshape(len(nodes),2,order="c")
        
        #[[n1.x0-xc[0] n1.y0-xc[1]  
        #  n2.x0-xc[0] n2.y0-xc[1]
        #  n3.x0-xc[0] n3.y0 -xc[1]
        #  ....]
        dx -= xc0

        #x-xc0 = R@(x0-xc0) -> x = xc0 + R@(x0-xc0) -> u = x - x0 = -(x0-xc0) + R@(x0-xc0)
        #(R*positions0.T).T = positions0@R.T
        u = dx@R.T 
        u -= dx
        u = KratosMultiphysics.Vector(u.reshape(len(nodes)*2))

        var_utils.ApplyVector(disp_var, 2, u, nodes)
        # for node in nodes:
        #     dx = np.array([node.X0-xc0[0], node.Y0-xc0[1]])

        #     #x-xc0 = R@(x0-xc0) -> x = xc0 + R@(x0-xc0) -> u = x - x0 = -(x0-xc0) + R@(x0-xc0)
        #     u = -dx 
        #     u += R@dx
        #     aux = np.array([u[0],u[1],0.0])
        #     node.SetSolutionStepValue(disp_var,0,aux)

    def GetDefaultSettings(self):
        return KratosMultiphysics.Parameters("""
                {
                    "displacement_variable"            : "MESH_DISPLACEMENT",
                    "velocity_variable"                : "MESH_VELOCITY",
                    "initial_xy_coords_global_rotation_center" : [0.0, 0.0],
                    "base_model_part_name"             : "AAA",
                    "rpms" : 10.0, //roations/minute
                    "parts"       : [
                        { "part_name" : "name_of_part",
                        "initial_xy_coords_part_center_of_rotation" : [0.0,0.0],
                        "rpms" : 10.0
                        }
                    ]
                }""")




    def __init__(self, Model, settings ):
        """ The default constructor of the class

        Keyword arguments:
        self -- It signifies an instance of a class.
        Model -- the container of the different model parts.
        settings -- Kratos parameters containing solver settings.
        """
        KratosMultiphysics.Process.__init__(self)

        self.var_utils = KratosMultiphysics.VariableUtils()

        #The value can be a double or a string (function)
        default_settings = self.GetDefaultSettings()

        settings.ValidateAndAssignDefaults(default_settings)

        #global data
        self.model_part = Model[settings["base_model_part_name"].GetString()]
        self.disp_var = KratosMultiphysics.KratosGlobals.GetVariable(settings["displacement_variable"].GetString())
        self.vel_var = KratosMultiphysics.KratosGlobals.GetVariable(settings["velocity_variable"].GetString())
        self.global_angular_velocity = settings["rpms"].GetDouble()*2*math.pi/60
        self.xc0 = settings["initial_xy_coords_global_rotation_center"].GetVector()

        #data for each blade
        self.part_data = []
        for part in settings["parts"]:
            mp = Model[part["part_name"].GetString()]
            xc0 = np.array( part["initial_xy_coords_part_center_of_rotation"].GetVector() )
            w = part["rpms"].GetDouble()*2*math.pi/60
            self.part_data.append([mp,xc0,w])

        #obtain the list of all nodes in parts involved in the rotation
        #and save the initial coordinates list
        #self.initial_coords_list = []
        self.all_nodes = KratosMultiphysics.NodesArray() 
        for part in self.part_data:
            for node in part[0].Nodes:
                self.all_nodes.append(node)
                #self.initial_coords_list.append(np.array([node.X0,node.Y0,node.Z0]))
        self.initial_coords_list = self.var_utils.GetInitialPositionsVector(self.all_nodes,2)

        #measure the min and max
        self.xmin=node.X0
        self.ymin=node.Y0
        self.zmin=node.Z0
        self.xmax=self.xmin
        self.ymax=self.ymin
        self.zmax=self.zmin
        for node in self.all_nodes:
            self.xmin=min(self.xmin,node.X0)
            self.ymin=min(self.ymin,node.Y0)
            self.zmin=min(self.zmin,node.Z0)
            self.xmax=max(self.xmax,node.X0)
            self.ymax=max(self.ymax,node.Y0)
            self.zmax=max(self.zmax,node.Z0)
        sumx=abs(self.xmin)+abs(self.xmax)
        sumy=abs(self.ymin)+abs(self.ymax)
        sumz=abs(self.zmin)+abs(self.zmax)
        print("x limits",self.xmin,self.xmax, " sum", sumx, " xc = ", self.xmin+sumx/2, " diam=",sumx/2.0)
        print("y limits",self.ymin,self.ymax, " sum", abs(self.ymin)+abs(self.ymax))
        print("z limits",self.zmin,self.zmax, " sum", abs(self.zmin)+abs(self.zmax))

    def ResetInitialPositions(self):
        self.var_utils.SetInitialPositionsVector(self.all_nodes,self.initial_coords_list)
        # for node,initial_pos in zip(self.all_nodes, self.initial_coords_list):
        #     node.X0 = initial_pos[0]
        #     node.Y0 = initial_pos[1]
        #     node.Z0 = initial_pos[2]

    def ComputeTotalDisplacementAndResetInitialPositions(self):
        d = self.var_utils.GetCurrentPositionsVector(self.all_nodes,2) - self.initial_coords_list
        self.var_utils.SetSolutionStepValuesVector(self.disp_var,0,d,self.all_nodes)

        self.var_utils.SetInitialPositionsVector(self.all_nodes,self.initial_coords_list)

        # d = KratosMultiphysics.Array3()
        # for node,initial_pos in zip(self.all_nodes, self.initial_coords_list):
        #     node.X0 = initial_pos[0]
        #     node.Y0 = initial_pos[1]
        #     node.Z0 = initial_pos[2]
        #     d[0] = node.X - node.X0
        #     d[1] = node.Y - node.Y0
        #     d[2] = node.Z - node.Z0

        #     node.SetSolutionStepValue(self.disp_var, 0, d)


    def ExecuteInitializeSolutionStep(self):
        
        """This method is executed in order to initialize the current step

        Keyword arguments:
        self -- It signifies an instance of a class.
        """
        current_time = self.model_part.ProcessInfo[KratosMultiphysics.TIME]
        dt = self.model_part.ProcessInfo[KratosMultiphysics.DELTA_TIME]
        #self.ResetInitialPositions()
        print("current_time",current_time)
        #apply rotation to all of the blades which compose the system
        #note that we TEMPORARILY actualize the InitialCoordinates
        for part in self.part_data:
            theta_part = current_time*part[2]
            xc0_part = part[1]
            nodes = part[0].Nodes
            Rpart = self.ComputeRotationMatrix(theta_part)
            self.ApplyRotation(xc0_part, Rpart, nodes, self.disp_var)
            x0vec = self.var_utils.GetInitialPositionsVector(nodes,2)
            dvec = self.var_utils.GetSolutionStepValuesVector(self.disp_var,2,0,nodes)
            x0vec += dvec
            self.var_utils.SetInitialPositionsVector(nodes, x0vec)
            # for node in nodes: #actualize the initial position to later apply the global rotation
            #     d = node.GetSolutionStepValue(self.disp_var)
            #     node.X0 += d[0]
            #     node.Y0 += d[1]
            #     node.Z0 += d[2]
            
        theta = current_time*self.global_angular_velocity
        R = self.ComputeRotationMatrix(theta) #global rotation matrix
        self.ApplyRotation(self.xc0, R, self.all_nodes, self.disp_var)
        x0vec = self.var_utils.GetInitialPositionsVector(self.all_nodes,2)
        dvec = self.var_utils.GetSolutionStepValuesVector(self.disp_var,2,0,self.all_nodes)
        x0vec += dvec
        self.var_utils.SetCurrentPositionsVector(self.all_nodes, x0vec)
        # for node in self.all_nodes: #actualize the initial position to later apply the global rotation
        #     d = node.GetSolutionStepValue(self.disp_var,0)
        #     node.X = node.X0 + d[0]
        #     node.Y = node.Y0 + d[1]
        #     node.Z = node.Z0 + d[2]

        self.ComputeTotalDisplacementAndResetInitialPositions()

        v = self.var_utils.GetSolutionStepValuesVector(self.disp_var,2,0,self.all_nodes)
        v -= self.var_utils.GetSolutionStepValuesVector(self.disp_var,2,1,self.all_nodes)
        v /= dt
        self.var_utils.SetSolutionStepValuesVector(self.vel_var,0,v,self.all_nodes)
        # for node in self.all_nodes: 
        #     #compute velocity
        #     v = d - node.GetSolutionStepValue(self.disp_var,1)
        #     v/=dt 
        #     node.SetSolutionStepValue(self.vel_var,0,v)

        self.ResetInitialPositions()

~~~

which is hundrueds of time faster than doing it with the loops in python

kratos/python/add_variable_utils_to_python.cpp
RiccardoRossi: ok, i was unaware. Still i would like a version with provided dimension to get only the x y component (or xyz in 3D)

RiccardoRossi: mmm, as i see there is no function to get the original or current coordinates, is that correct?

also the function  GetVectorData and GetScalarData are not exported to python, and even if they were they do not allow specifying the step, so in any case i would need to have a version of those. Am i correct?

One option would be to modify the current header to 

~~~cpp
template<class TDataType>
void GetVectorData(
        const Variable<TDataType>& rVariable,
        const DataLocation DataLoc,
        std::vector<double>& data,
        int TSize=-1, //impossible number, if we later detect it is -1, we would use automatic detection
        int Step=0 //only used for historical data...
) const
{
   ...
   if(TSidze==-1) automatic detection of Tsize
}
~~~

as an aside, why returing a std::vector? it would be better to either template the return type or to default to Vector which is faster...

philbucher: - yes you are right, there are no functions for the coordinates, hence adding them makes sense to me
- Ok I didn't realize they are that different. I leave it up to you if you want to either extend the existing fcts or add new ones. If you add new ones, can you please add them properly, not in `add_to_python_...`, including a test?
- I choose `std::vector` since it also supports non-`double` types (and bcs the interface of CoSimIO uses it :wink:)

philbucher (dismissed):
unblocking, but please take my comments into consideration, I think this needs still some cleanup work and some tests :)

kratos/python/add_variable_utils_to_python.cpp
RiccardoRossi: to add to my previous comment, the function is model_oart_utils onnly works on modelpart, while it is very useful to be able to pass a list of nodes disconnected from a modelpart.

summing up all of these reasons i think i 'll add the function to variable_utils as i did (incorporating the suggestions of @loumalouomega and @miguelmaso )

as a final comment i wanted to report that, as discussed privetely with @philbucher it is generally not a good idea using std::vector in performance sensitive parts of the code, since the std vector is default initialized to zero (when the default allocator is used). A possible solution for this is to use boost's noinit_adaptor

https://www.boost.org/doc/libs/master/libs/core/doc/html/core/noinit_adaptor.html#:~:text=The%20header%20%3Cboost%2Fcore%2Fnoinit_adaptor.hpp%3E%20provides%20the%20class%20template%20boost%3A%3Anoinit_adaptor,new%2C%20and%20whose%20destroy%28ptr%29%20invokes%20value_type%20destructor%20directly.

in any case this is a separate discussion and i am reporting it here just to leave trace of what i consider to be an important  aspect

RiccardoRossi (commented):
@loumalouomega @miguelmaso (and in CC, @philbucher ) i think i addressed the concerns.

concretely
- i changed the order of arguments to clean it up
- i added a text
- i changed the doxygen
- i added all the implementationso in the same file

kratos/python/add_variable_utils_to_python.cpp
```diff
@@ -452,6 +452,25 @@ void AddVariableUtilsToPython(pybind11::module &m)
...
+        .def("SetSolutionStepValuesVector", py::overload_cast<
+                            ModelPart::NodesContainerType&,
+                            const Variable<array_1d<double,3>>&,
+                            const Vector&,
+                            const unsigned int>(&VariableUtils::SetSolutionStepValuesVector))
+        .def("SetSolutionStepValuesVector", py::overload_cast<
+                            ModelPart::NodesContainerType&,
+                            const Variable<double>&,
+                            const Vector&,
+                            const unsigned int>(&VariableUtils::SetSolutionStepValuesVector))
```
loumalouomega, line 473: I know it is the same, can you put unsigned int instead of just unsigned?

kratos/tests/test_variable_utils.py
```diff
@@ -4,6 +4,7 @@
 
 import math
 import os
+import numpy as np
```
loumalouomega: If you use numpy, please add the decorator to skip the tests. It is used in other tests depending on numpy

kratos/utilities/variable_utils.h
```diff
@@ -1389,6 +1389,136 @@ class KRATOS_API(KRATOS_CORE) VariableUtils
...
+     * current position as the initial position (X0, Y0, Z0) plus such variable value
+     * in case Dimension == 1 the expected input vector is in the form (X0 X0 X0 ....)
+     * in case Dimension == 2 the expected input vector is in the form (X0 Y0 X0 Y0 X0 Y0 ....)
+     * in case Dimension == 3 the expected input vector is in the form (X0 Y0 Z0 X0 Y0 Z0 ....)
+     * @param rNodes array of nodes from which coordinates will be extracted
+     * @param rPositions vector containing the INITIAL positions 
+     */
+    void SetInitialPositionsVector(
+        ModelPart::NodesContainerType& rNodes, 
+        const Vector& rPositions);
```
loumalouomega:
```suggestion
        const Vector& rPositions
        );
```

kratos/utilities/variable_utils.h
```diff
@@ -1389,6 +1389,136 @@ class KRATOS_API(KRATOS_CORE) VariableUtils
...
+     * current position as the initial position (X, Y, Z) plus such variable value
+     * in case Dimension == 1 the expected input vector is in the form (X X X ....)
+     * in case Dimension == 2 the expected input vector is in the form (X Y X Y X Y ....)
+     * in case Dimension == 3 the expected input vector is in the form (X Y Z X Y Z ....)
+     * @param rNodes array of nodes from which coordinates will be extracted
+     * @param rPositions vector containing the CURRENT positions 
+     */
+     void SetCurrentPositionsVector(
+        ModelPart::NodesContainerType& rNodes, 
+        const Vector& rPositions);
```
loumalouomega:
```suggestion
        const Vector& rPositions
        );
```

kratos/utilities/variable_utils.h
```diff
@@ -1389,6 +1389,136 @@ class KRATOS_API(KRATOS_CORE) VariableUtils
...
+     * current position as the initial position (X0, Y0, Z0) plus such variable value
+     * in case Dimension == 1 the vector is in the form (X0 X0 X0 ....)
+     * in case Dimension == 2 the vector is in the form (X0 Y0 X0 Y0 X0 Y0 ....)
+     * in case Dimension == 3 the vector is in the form (X0 Y0 Z0 X0 Y0 Z0 ....)
+     * @param rNodes array of nodes from which coordinates will be extracted
+     * @param Dimension number of desired components
+     */
+    Vector GetInitialPositionsVector(
+        const ModelPart::NodesContainerType& rNodes, 
+        const unsigned int Dimension);
```
loumalouomega:
```suggestion
        const unsigned int Dimension
        );
```

kratos/utilities/variable_utils.h
```diff
@@ -1389,6 +1389,136 @@ class KRATOS_API(KRATOS_CORE) VariableUtils
...
+     * current position as the initial position (X0, Y0, Z0) plus such variable value
+     * in case Dimension == 1 the vector is in the form (X X X ....)
+     * in case Dimension == 2 the vector is in the form (X Y X Y X Y ....)
+     * in case Dimension == 3 the vector is in the form (X Y Z X Y Z ....)
+     * @param rNodes array of nodes from which coordinates will be extracted
+     * @param Dimension number of desired components
+     */
+    Vector GetCurrentPositionsVector(
+        const ModelPart::NodesContainerType& rNodes, 
+        const unsigned int Dimension);
```
loumalouomega:
```suggestion
        const unsigned int Dimension
        );
```

loumalouomega (commented):
Last comments

kratos/tests/test_variable_utils.py
RiccardoRossi: no, numpy is now accepted

kratos/python/add_variable_utils_to_python.cpp
RiccardoRossi: o think it is "unsigned int" everywhere...can u point me where it is not?

kratos/tests/test_variable_utils.py
RiccardoRossi: i mean, it is accepted since when we package using wheels

loumalouomega (dismissed):

kratos/tests/test_variable_utils.py
```diff
@@ -693,6 +693,75 @@ def test_apply_vector(self):
...
+        model_part = current_model.CreateModelPart("Main")
+        model_part.AddNodalSolutionStepVariable(KratosMultiphysics.VISCOSITY)
+        model_part.AddNodalSolutionStepVariable(KratosMultiphysics.DISPLACEMENT)
+        model_part.AddNodalSolutionStepVariable(KratosMultiphysics.VELOCITY)
+        model_part.CreateNewNode(1,1.0,2.0,3.0)
+        model_part.CreateNewNode(2,11.0,25.0,31.0)
+        model_part.CreateNewNode(10,12.0,26.0,32.0)
+
+        #reference_results
+        ref_initial_coords2d = [1.0,2.0,11.0,25.0,12.0,26.0]
```
RiccardoRossi, line 709: @loumalouomega i removed the numpy dependency since this test could be done without introducing numpy (i only need lists).
still numpy Is allowed, and should be also present in all of the compile bots.

I saw that you approved, could u please approve again?

loumalouomega (approved):

PR #10123, opened 2022-08-10
[Core] Volume inside voxel utility

AriadnaCortesDanes:
**📝 Description**
Set of utilities to approximate the real portion of a voxel that is filled with volume, using the variable distance of the nodes and the set of triangles intersecting the voxel's edges.

**🆕 Changelog**
- Added the class VolumeInsideVoxelUtility, which contains different static methods using different approximations (depending on their cost and reliability). Added the header and .cpp of this class
- Added the corresponding tests.


kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,393 @@
...
+                volume+=(1.0/nodes.size()); 
+            } 
+        }
+        return volume;
+    }
+
+    /***********************************************************************************
+     **********************************************************************************/
+
+    double VolumeInsideVoxelUtility::EdgesApproximation(
```
loumalouomega: If there is only one argument, you can put it i one line

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,393 @@
...
+    ) {
+        double volume = 0;
+        GeometryArrayType Edges = rVoxel.GenerateEdges();
+        std::vector<double> Distances;
+
+        for (int i = 0; i < Edges.size(); i++) {
+            Distances.push_back(0);
+            PointsArrayType ends = Edges[i].Points();
+
+            for (auto triangle : rTriangles) {
```
loumalouomega:
```suggestion
            for (auto& r_triangle : rTriangles) {
```

references avoid copies (memory efficient and reduces computation time due to allocation)

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,393 @@
...
+        int NodesInside = 0;
+        for (int i = 0; i < Nodes.size(); i++) {
+            if (Nodes[i].GetSolutionStepValue(DISTANCE) > 0) NodesInside++;
+        }
+
+        if(NodesInside == 3) {
+            for (int i = 0; i < Nodes.size(); i++)  {
+                Nodes[i].GetSolutionStepValue(DISTANCE) = (-1)*Nodes[i].GetSolutionStepValue(DISTANCE);
+            }
+            double Area = HexaVolume2D(rFace,rTriangles);
```
loumalouomega:
```suggestion
            const double area = HexaVolume2D(rFace,rTriangles);
```
snake_case when it is not an input. And use const if possible

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,393 @@
...
+            for (int i = 0; i < Nodes.size(); i++) {
+                Nodes[i].GetSolutionStepValue(DISTANCE) = (-1)*Nodes[i].GetSolutionStepValue(DISTANCE);
+            }
+            return 1-Area;
+        }
+
+        std::vector<double> Length(Edges.size()); 
+        for(int i = 0; i < Edges.size(); i++) {
+            PointsArrayType ends = Edges[i].Points();
+            double l = Distance(ends[0], ends[1]);
```
loumalouomega:
```suggestion
            const double l = Distance(ends[0], ends[1]);
```

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,393 @@
...
+        for (int i = 0; i < rTriangles.size(); i++) {
+            int Result = 0; 
+            array_1d<double,3> Intersection;
+            int j = 0;
+            while(!Result && j < Edges.size()) { 
+                PointsArrayType ends = Edges[j].Points();
+                Result = IntersectionUtilities::ComputeTriangleLineIntersection(rTriangles[i],ends[0],ends[1],Intersection);
+
+                if (Result) {
+                    double Dist = Distance(ends[0], Intersection);
```
loumalouomega:
```suggestion
                    const double Dist = Distance(ends[0], Intersection);
```

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,393 @@
...
+                    if ( Dist < (MinDistanceToNode[j].first*Length[j])) {
+                        MinDistanceToNode[j].first = Dist/Length[j];
+                    } 
+
+                    double Dist2 = Distance(ends[1], Intersection);
```
loumalouomega:
```suggestion
                   const double Dist2 = Distance(ends[1], Intersection);
```

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,393 @@
...
+            double PartialArea;
+            double factor = 1;
+            double left = 0;
+            double right = 0;
+            if (Nodes[i].GetSolutionStepValue(DISTANCE) > 0) {
+                if (Case == 0) {
+                    factor = 0.5;
+                    left = MinDistanceToNode[(i+3)%4].second;
+                    right = MinDistanceToNode[i].first;
+                }
```
loumalouomega:
```suggestion
                } else if ()....
```
This would be the syle

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,393 @@
...
+                else if (Case == 1)  {
+                    left = MinDistanceToNode[(i+3)%4].second;
+                    right = std::min(0.5,MinDistanceToNode[i].first);
+                }
```
loumalouomega: The same for the rest

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,393 @@
...
+                                                    Nodes[i].Y() + 0.5*v_left[1] + 0.5*v_right[1], 
+                                                    Nodes[i].Z() + 0.5*v_left[2] + 0.5*v_right[2]));
+                PointsArrayType points;
+                points.push_back(Max_left);
+                points.push_back(&Nodes[i]);
+                points.push_back(Max_right);
+                points.push_back(Max_c);
+                double max_volume = TetraVolume(points)/FaceArea;
+
+                if (Case == 1) {
```
loumalouomega: Idem

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,393 @@
...
+                points.push_back(Int_right);
+                points.push_back(c);
+                PartialArea = factor*TetraVolume(points)/FaceArea;
+            } else  {
+                NodePtrType Max_left(new Node<3>(1, Nodes[i].X() + 0.5*v_left[0], Nodes[i].Y() + 0.5*v_left[1], Nodes[i].Z() + 0.5*v_left[2]));
+                NodePtrType Max_right(new Node<3>(2, Nodes[i].X() + 0.5*v_right[0], Nodes[i].Y() + 0.5*v_right[1], Nodes[i].Z() + 0.5*v_right[2]));
+                NodePtrType Max_c(new Node<3>(2, Nodes[i].X() + 0.5*v_left[0] + 0.5*v_right[0], 
+                                                    Nodes[i].Y() + 0.5*v_left[1] + 0.5*v_right[1], 
+                                                    Nodes[i].Z() + 0.5*v_left[2] + 0.5*v_right[2]));
+                PointsArrayType points;
```
loumalouomega: Avoid push_bask if you know the size a priori

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,393 @@
...
+                points.push_back(Max_left);
+                points.push_back(&Nodes[i]);
+                points.push_back(Max_right);
+                points.push_back(Max_c);
+                double max_volume = TetraVolume(points)/FaceArea;
```
loumalouomega: const

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,393 @@
...
+                    left = std::min(0.5,MinDistanceToNode[(i+3)%4].second);
+                    right = std::min(0.5,MinDistanceToNode[i].first);
+                }
+                NodePtrType Int_left(new Node<3>(1, Nodes[i].X() + left*v_left[0], Nodes[i].Y() + left*v_left[1], Nodes[i].Z() + left*v_left[2]));
+                NodePtrType Int_right(new Node<3>(2, Nodes[i].X() + right*v_right[0], Nodes[i].Y() + right*v_right[1], Nodes[i].Z() + right*v_right[2]));
+                NodePtrType c(new Node<3>(2, Nodes[i].X() + left*v_left[0] + right*v_right[0], 
+                                                    Nodes[i].Y() + left*v_left[1] + right*v_right[1], 
+                                                    Nodes[i].Z() + left*v_left[2] + right*v_right[2]));
+                points.clear();
+                points.push_back(Int_left);
```
loumalouomega: Avoid push_back

loumalouomega (changes requested):
Very nice, some comments

kratos/utilities/volume_inside_voxel_utility.cpp
AriadnaCortesDanes: True thanks!

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,382 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:		 BSD License
+//					 Kratos default license: kratos/license.txt
+//
+//  Main authors:    Ariadna Cortés
```
rubenzorrilla: Avoid the accent. We had weird issues with strange characters in the past.

kratos/utilities/volume_inside_voxel_utility.h
```diff
@@ -0,0 +1,194 @@
...
+     * @return Approximated volume 
+     * @note This approximation assigns a fraction of volume (1/8) to each node of the
+     * voxel, and counts it as volume if the node is inside the object (NodeDistance > 0)
+     * This operation is VERY cheap
+     */  
+    static double NodesApproximation(const GeometryType& rVoxel);
+
+    /*This method is completly useless since it does the same calculation as the previous 
+    one but in a different way. Helps to illustrate use of Edges */
+    static double EdgesApproximation(const GeometryType& rVoxel);
```
rubenzorrilla: If it is useless remove it.

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,382 @@
...
+    /***********************************************************************************
+     **********************************************************************************/
+
+    double VolumeInsideVoxelUtility::EdgesPortionApproximation(
+        const GeometryType& rVoxel,  
+        const GeometryArrayType& rTriangles     
+    ) {
+        double volume = 0;
+        GeometryArrayType Edges = rVoxel.GenerateEdges();
+        std::vector<double> Distances;
```
rubenzorrilla:
```suggestion
        std::vector<double> distances;
```
Please stick to the style guide.

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+    double VolumeInsideVoxelUtility::Distance(const NodeType& Point0, const array_1d<double,3>& Point1) {
+        const double lx = Point0.X() - Point1[0];
+        const double ly = Point0.Y() - Point1[1];
+        const double lz = Point0.Z() - Point1[2];
+
+        const double length = lx * lx + ly * ly + lz * lz;
+
+        return std::sqrt( length );
+    }
```
rubenzorrilla: This is a straight `norm_2()` call. Use it instead and remove this function.

kratos/utilities/volume_inside_voxel_utility.cpp
rubenzorrilla: Besides, you know the size a priori so you can allocate the space required and avoid the `push_back` below.

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,382 @@
...
+                array_1d<double,3> intersection;
+                int result = IntersectionUtilities::ComputeTriangleLineIntersection(r_triangle,ends[0],ends[1],intersection);
+                
+                if(result == 1) {
+                    double Dist = Distance(ends[0], intersection);
+                    Distances.push_back(Dist);
+                }  
+            } 
+            Distances.push_back(Distance(ends[0],ends[1]));
+            std::sort(Distances.begin(),Distances.end());       //WOULD A SET BE MORE EFFICIENT?
```
rubenzorrilla: Why you need this sort at each loop cycle?

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,382 @@
...
+     **********************************************************************************/
+
+    double VolumeInsideVoxelUtility::VoxelVolume2D(
+        const GeometryType& rFace,  
+        const GeometryArrayType& rTriangles     
+    ) {
+        double Area = 0;
+        GeometryArrayType Edges = rFace.GenerateEdges();
+        PointsArrayType Nodes = rFace.Points(); 
+        std::vector<std::pair<double,double>> MinDistanceToNode(Edges.size(),{0.5,0.5});
```
rubenzorrilla: Please stick to the style guide. Use `snake_case` for local variables.

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,382 @@
...
+    /***********************************************************************************
+     **********************************************************************************/
+
+    double VolumeInsideVoxelUtility::VoxelVolume2D(
+        const GeometryType& rFace,  
+        const GeometryArrayType& rTriangles     
+    ) {
+        double Area = 0;
+        GeometryArrayType Edges = rFace.GenerateEdges();
+        PointsArrayType Nodes = rFace.Points();
```
rubenzorrilla:
```suggestion
        const auto& r_nodes = rFace.Points(); 
```
Note the `&` to avoid the copy.

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,382 @@
...
+                    double Dist2 = Distance(ends[1], Intersection);
+                    if (Dist2 < (MinDistanceToNode[j].second*Length[j])) {
+                        MinDistanceToNode[j].second = Dist2/Length[j];
+                    } 
+                }
+                j++;
+            }
+        }
+
+        std::vector<std::vector<double>> Neighbours{{3,1},{0,2},{1,3},{0,2}};
```
rubenzorrilla: I'd declare it as a constexpr in the header. Otherwise you're allocating it at each iteration of the loop.

rubenzorrilla: Besides, I think you can use an `std::vector<array_1d<std::size_t, 2>>` (or a pair)

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,382 @@
...
+        std::vector<std::pair<double,double>> MinDistanceToNode(Edges.size(),{1,1}); 
+        
+        int NodesInside = 0;
+        for (int i = 0; i < Nodes.size(); i++) {
+            if (Nodes[i].GetSolutionStepValue(DISTANCE) > 0) NodesInside++;
+        }
+
+        if(NodesInside == 3) {
+            for (int i = 0; i < Nodes.size(); i++)  {
+                Nodes[i].GetSolutionStepValue(DISTANCE) = (-1)*Nodes[i].GetSolutionStepValue(DISTANCE);
```
rubenzorrilla:
```suggestion
                Nodes[i].GetSolutionStepValue(DISTANCE) = -Nodes[i].GetSolutionStepValue(DISTANCE);
```
Same below.

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,382 @@
...
+                    const double Dist2 = Distance(ends[1], Intersection);
+                    if (Dist2 < (MinDistanceToNode[j].second*Length[j])) {
+                        MinDistanceToNode[j].second = Dist2/Length[j];
+                    } 
+                }
+                j++;
+            }
+        }
+
+        std::vector<std::vector<double>> Neighbours{{3,1},{0,2},{1,3},{2,0}};
```
rubenzorrilla: Same comment.

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,382 @@
...
+            nodes[neighbours[node][0]].GetSolutionStepValue(DISTANCE) > 0 && nodes[neighbours[node][1]].GetSolutionStepValue(DISTANCE) > 0) {
+                return 0.5;
+            }
+        return 1;
+    }
+
+    /***********************************************************************************
+     **********************************************************************************/
+
+    int VolumeInsideVoxelUtility::GetCase(const PointsArrayType& nodes, const std::vector<std::vector<double>>& neighbours,const int node) {
```
rubenzorrilla:
```suggestion
    int VolumeInsideVoxelUtility::GetCase(
        const PointsArrayType& rNodes,
        const std::vector<std::vector<double>>& rNeighbours,
        const int NodeIndex)
    {
```
Please stick to the style guide. Also note that by Node we usually refer to the object, you intend to pass the node index.

kratos/utilities/volume_inside_voxel_utility.h
```diff
@@ -0,0 +1,194 @@
...
+     * @brief Aproximates the actual area inside a quadrilateral with 90º angles
+     * @param rFace references to the quadrilateral3D4 whose actual area will be approximated
+     * @param rTriangles references to the triangles which may intersect the quadrilateral at some edge.
+     * @return Approximated area 
+     * @note this method is cheaper than HexaVolume2D, but won't work correctly with != 90º angles
+     */  
+    static double VoxelVolume2D(const GeometryType& rFace, const GeometryArrayType& rTriangles);
+
+    /**
+     * @brief Aproximates the actual area inside a quadrilateral
```
rubenzorrilla:
```suggestion
     * @brief Aproximates the actual area inside an hexahedra
```
No?

kratos/utilities/volume_inside_voxel_utility.h
```diff
@@ -0,0 +1,194 @@
...
+    static double VoxelVolume2D(const GeometryType& rFace, const GeometryArrayType& rTriangles);
+
+    /**
+     * @brief Aproximates the actual area inside a quadrilateral 
+     * @param rFace references to the quadrilateral3D4 whose actual area will be approximated
+     * @param rTriangles references to the triangles which may intersect the quadrilateral at some edge.
+     * @return Approximated area 
+     * @note Attention: This method is quite expensive 
+     */  
+    static double HexaVolume2D(const GeometryType& rFace,const GeometryArrayType& rTriangles);
```
rubenzorrilla: The name is way misleading. There is no 2D hexa. I suggest finding a more self-descriptive naming.

kratos/utilities/volume_inside_voxel_utility.h
```diff
@@ -0,0 +1,194 @@
...
+    static double EdgesPortionApproximation(const GeometryType& rVoxel, const GeometryArrayType& rTriangles);
+
+    /**
+     * @brief Aproximates the actual area inside a quadrilateral with 90º angles
+     * @param rFace references to the quadrilateral3D4 whose actual area will be approximated
+     * @param rTriangles references to the triangles which may intersect the quadrilateral at some edge.
+     * @return Approximated area 
+     * @note this method is cheaper than HexaVolume2D, but won't work correctly with != 90º angles
+     */  
+    static double VoxelVolume2D(const GeometryType& rFace, const GeometryArrayType& rTriangles);
```
rubenzorrilla: This naming is way misleading... Also the documentation. It is not clear for future users what this method does.

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,382 @@
...
+        return std::sqrt( length );
+    }
+
+    /***********************************************************************************
+     **********************************************************************************/
+
+    double VolumeInsideVoxelUtility::TetraVolume(const PointsArrayType& rPoints) {
+        GeometryPtrType pGeom =Kratos::make_shared<Quadrilateral3D4<NodeType>>(rPoints);
+        return pGeom->Volume();
+    }
```
rubenzorrilla: This is extremely expensive. Use the closed formula from the point coordinates.

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,382 @@
...
+            inside = false;
+        } else {    //rEnds[0].GetSolutionStepValue(DISTANCE) < 0 && rEnds[1].GetSolutionStepValue(DISTANCE) < 0
+            if (Distances.size() % 2 == 1) Distances.pop_back();
+            if (Distances.size() == 2) return 0;
+            inside = false;
+        }
+
+        for(int i = 1; i < Distances.size(); i++) {
+            if (inside) {
+                portion += abs(Distances[i]-Distances[i-1])/Length;
```
rubenzorrilla:
```suggestion
                portion += std::abs(Distances[i]-Distances[i-1])/Length;
```

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,382 @@
...
+     **********************************************************************************/
+
+    double VolumeInsideVoxelUtility::GetFactor(const PointsArrayType& nodes, const std::vector<std::vector<double>>& neighbours,const int node) {
+
+        if( nodes[node].GetSolutionStepValue(DISTANCE) > 0 && nodes[neighbours[node][0]].GetSolutionStepValue(DISTANCE) < 0 &&
+            nodes[neighbours[node][1]].GetSolutionStepValue(DISTANCE) < 0 || nodes[node].GetSolutionStepValue(DISTANCE) < 0 && 
+            nodes[neighbours[node][0]].GetSolutionStepValue(DISTANCE) > 0 && nodes[neighbours[node][1]].GetSolutionStepValue(DISTANCE) > 0) {
+                return 0.5;
+            }
+        return 1;
```
rubenzorrilla:
```suggestion
        return 1.0;
```

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,382 @@
...
+                j++;
+            }
+        }
+
+        std::vector<std::vector<double>> Neighbours{{3,1},{0,2},{1,3},{2,0}};  
+        for(int i = 0; i < Nodes.size(); i++ ) {
+            array_1d<double,3> v_left{Nodes[i].X() -Nodes[Neighbours[i][0]].X(), Nodes[i].Y() -Nodes[Neighbours[i][0]].Y(), Nodes[i].Z() -Nodes[Neighbours[i][0]].Z()};
+            array_1d<double,3> v_right{Nodes[i].X() -Nodes[Neighbours[i][1]].X(), Nodes[i].Y() -Nodes[Neighbours[i][1]].Y(), Nodes[i].Z() -Nodes[Neighbours[i][1]].Z()};
+
+            double Case = GetCase(Nodes, Neighbours,i);
```
rubenzorrilla:
```suggestion
            const int Case = GetCase(Nodes, Neighbours,i);
```

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,382 @@
...
+                return 0;
+            } else if (nodes[neighbours[node][0]].GetSolutionStepValue(DISTANCE) < 0 &&
+                nodes[neighbours[node][1]].GetSolutionStepValue(DISTANCE) > 0) {
+                return 1;
+            } else if (nodes[neighbours[node][0]].GetSolutionStepValue(DISTANCE) > 0 &&
+                nodes[neighbours[node][1]].GetSolutionStepValue(DISTANCE) < 0) {
+                return 2;
+            } else  {
+                return 3;
+            }
```
rubenzorrilla: This can be a switch.

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,382 @@
...
+            if (nodes[neighbours[node][0]].GetSolutionStepValue(DISTANCE) > 0 && 
+                nodes[neighbours[node][1]].GetSolutionStepValue(DISTANCE) > 0) {
+                return 0;
+            } else if (nodes[neighbours[node][0]].GetSolutionStepValue(DISTANCE) > 0 && 
+                nodes[neighbours[node][1]].GetSolutionStepValue(DISTANCE) < 0) {
+                return 1;
+            } else if (nodes[neighbours[node][0]].GetSolutionStepValue(DISTANCE) < 0 && 
+                nodes[neighbours[node][1]].GetSolutionStepValue(DISTANCE) > 0) {
+                return 2;
+            } else return 3;
```
rubenzorrilla: This can be a switch.

kratos/python_scripts/run_cpp_tests.py
```diff
@@ -1,3 +1,9 @@
-from KratosMultiphysics import *
-Tester.SetVerbosity(Tester.Verbosity.TESTS_OUTPUTS)
-Tester.RunAllTestCases()
```
rubenzorrilla: This change must be done in a standalone PR. Checkout this file in a new branch, create a new PR and once it is merge, merge the master in here.

rubenzorrilla (changes requested):
Several comments. Please, alto take into account the [style  guide](https://github.com/KratosMultiphysics/Kratos/wiki/Style-Guide).

rubenzorrilla:
Besides, I think it would be a good idea to rename the class to `VoxelUtilities` so we can wrap up all the future implementations in here (not only the volume ones).

kratos/utilities/volume_inside_voxel_utility.h
AriadnaCortesDanes: No, this function is meant to work for the quadrilateral faces of an hexahedra. Sorry, I think the name of the two functions ("VoxelVolume2D" and  "HexaVolume2D") may be confusing since they intend to refer to the quadrilateral faces of a Voxel (90º angles) and an hexa (general quadrilater with any angle), respectively. I will try to rename them to a more intuitive name but I totally accept suggestions.

kratos/utilities/volume_inside_voxel_utility.h
AriadnaCortesDanes: Maybe something like HexahedraFaceArea would sound more descriptive? And VoxelFaceArea to the other function?

kratos/utilities/volume_inside_voxel_utility.cpp
AriadnaCortesDanes: Each cycle i , the vector is re-filled with the distances to the intersecting triangles at the edge[i] , which may always be unordered. The function called immediately after, finds the portion we want but needs the vector to be sorted.

kratos/utilities/volume_inside_voxel_utility.cpp
AriadnaCortesDanes: Maybe I am wrong, but I think there is no way to declare a std::vector as a constexpr (due to the way the constructor allocates memory?)

kratos/utilities/volume_inside_voxel_utility.cpp
AriadnaCortesDanes: > This can be a switch.

I re-implemented it in a simpler way, no need for a switch or the else-if. Please re-review it 😄

kratos/tests/cpp_tests/utilities/test_volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,748 @@
...
+    typedef Node<3>::Pointer NodePtrType;
+    typedef Geometry<NodeType> GeometryType;
+    typedef GeometryType::Pointer GeometryPtrType;
+    typedef GeometryType::GeometriesArrayType GeometryArrayType;
+    typedef GeometryType::PointsArrayType PointsArrayType;
+
+    /*this functions are creating a model for each element used in the test in order to avoid trouble with repeated Ids.
+    This is not optimal but since it is just a test (in real cases, funcions will be used with well-defined models)...
+    */
+    GeometryPtrType GenerateHexahedra3D8(const std::vector<double>& rDistances)
```
pooyan-dadvand: I would separate the `GenerateHex `and `SetDistance`. The reason is that you create the voxel once and you only change the distances afterward. In your implementation you are recreating the same nodes several times which is returning you the previously created ones and changing their distances would result in changing the original one distances.

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+//Project includes
+#include "volume_inside_voxel_utility.h"
+
+namespace Kratos {
+
+    double VolumeInsideVoxelUtility::NodesApproximation(const GeometryType& rVoxel) {
+        double volume = 0;
+        const PointsArrayType& nodes = rVoxel.Points();
+        for (std::size_t i = 0; i < nodes.size(); i++) {
```
pooyan-dadvand: I would use a local const variable for nodes.size(). In principle optimizer should do it well because` nodes` is constant but helping it wouldn't be bad.

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,336 @@
...
+            if (nodes[i].GetSolutionStepValue(DISTANCE) > 0) {
+                volume+=(1.0/nodes.size());
```
pooyan-dadvand: and use it here too

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,336 @@
...
+                array_1d<double,3> intersection;
+                int result = IntersectionUtilities::ComputeTriangleLineIntersection(r_triangle,ends[0],ends[1],intersection);
+                
+                if(result == 1) {
+                    double dist = norm_2(ends[0].Coordinates() - intersection);
+                    distances.push_back(dist);
+                }  
+            } 
+            distances.push_back(norm_2(ends[0].Coordinates() - ends[1].Coordinates()));
+            std::sort(distances.begin(),distances.end());       //WOULD A SET BE MORE EFFICIENT?
```
pooyan-dadvand: No. Set would pay off for large number of entities. For small amount the vector and brute force works better

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,336 @@
...
+                    const double dist2 = norm_2(ends[1].Coordinates() - intersection);
+                    if (dist2 < (min_distance_to_node[j].second*length[j])) {
+                        min_distance_to_node[j].second = dist2/length[j];
+                    } 
+                }
+                j++;
+            }
+        }
+
+        //std::vector<std::vector<double>> mNeighbours{{3,1},{0,2},{1,3},{0,2}};
```
pooyan-dadvand: Commented code

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+    /***********************************************************************************
+     **********************************************************************************/
+
+    double VolumeInsideVoxelUtility::EdgesPortionApproximation(
+        const GeometryType& rVoxel,  
+        const GeometryArrayType& rTriangles     
+    ) {
+        double volume = 0;
+        GeometryArrayType edges = rVoxel.GenerateEdges();
```
pooyan-dadvand: I would write this function without creating explicitly the edges because this operation is expensive. Here we can use the connectivity rather than array of the real line geometries to get the points and work with them directly.

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+    /***********************************************************************************
+     **********************************************************************************/
+
+    double VolumeInsideVoxelUtility::VoxelFaceArea(
+        const GeometryType& rFace,  
+        const GeometryArrayType& rTriangles     
+    ) {
+        double area = 0;
+        GeometryArrayType edges = rFace.GenerateEdges();
```
pooyan-dadvand: Same here

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    /***********************************************************************************
+     **********************************************************************************/
+
+    double VolumeInsideVoxelUtility::HexahedraFaceArea(
+        const GeometryType& rFace,  
+        const GeometryArrayType& rTriangles     
+    ) {
+        double area = 0;
```
pooyan-dadvand: Large amount of duplicated code

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    /***********************************************************************************
+     **********************************************************************************/
+
+    double VolumeInsideVoxelUtility::GetFactor(
+        const PointsArrayType& rNodes, 
+        const int NodeIndex) 
+    {
+
```
pooyan-dadvand: I would store the distances to call the `GetValue `once per each node.

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    /***********************************************************************************
+     **********************************************************************************/
+
+    double VolumeInsideVoxelUtility::EdgeFilledPortion(std::vector<double>& Distances, const PointsArrayType& rEnds) {
+        const double length = Distances[Distances.size() - 1];
+        double portion = 0; 
+        bool inside = true;       
+
```
pooyan-dadvand: I would create local variables for distances avoiding several calls.

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,336 @@
...
+                if(result == 1) {
+                    double dist = norm_2(ends[0].Coordinates() - intersection);
+                    distances.push_back(dist);
+                }  
+            } 
+            distances.push_back(norm_2(ends[0].Coordinates() - ends[1].Coordinates()));
+            std::sort(distances.begin(),distances.end());       //WOULD A SET BE MORE EFFICIENT?     
+            double edgePortion = VolumeInsideVoxelUtility::EdgeFilledPortion(distances, ends);
+            volume += edgePortion/edges.size();  
+            distances.clear();
```
pooyan-dadvand: Why are you calculate the distances to throw them away?

kratos/utilities/volume_inside_voxel_utility.h
```diff
@@ -0,0 +1,184 @@
...
+///@name Kratos Classes
+///@{
+
+/**
+ * @class VolumeInsideVoxel
+ * @ingroup KratosCore
+ * @brief Utilities to compute the real volume inside a voxel
+ * @details This class provides static methods to compute (using different approximations) the portion of a 
+ * voxel that is actually filled with volume, according to the known triangle elements that intersect the voxel
+ * @author Ariadna Cortés
```
loumalouomega:
```suggestion
 * @author Ariadna Cortes
```
Evil tilde!!!

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,336 @@
...
+//
+//  Main authors:    Ariadna Cortes
+//
+
+//Project includes
+#include "volume_inside_voxel_utility.h"
+
+namespace Kratos {
+
+    double VolumeInsideVoxelUtility::NodesApproximation(const GeometryType& rVoxel) {
```
loumalouomega: I would remove the tabulation in the cpp. Makes the code more compact

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,336 @@
...
+        for (std::size_t i = 0; i < edges.size(); i++) {
+            distances.push_back(0);
+            PointsArrayType& ends = edges[i].Points();
+
+            for (auto& r_triangle : rTriangles) {
+                array_1d<double,3> intersection;
+                int result = IntersectionUtilities::ComputeTriangleLineIntersection(r_triangle,ends[0],ends[1],intersection);
+                
+                if(result == 1) {
+                    double dist = norm_2(ends[0].Coordinates() - intersection);
```
loumalouomega:
```suggestion
                    const double dist = norm_2(ends[0].Coordinates() - intersection);
```

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    /***********************************************************************************
+     **********************************************************************************/
+
+    int VolumeInsideVoxelUtility::GetCase(
+        const PointsArrayType& rNodes,
+        const int NodeIndex) 
+    {
+        int me_inside = (int) rNodes[NodeIndex].GetSolutionStepValue(DISTANCE) > 0;
```
loumalouomega:
```suggestion
        const int me_inside = (int) rNodes[NodeIndex].GetSolutionStepValue(DISTANCE) > 0;
```

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,336 @@
...
+        int left_inside = (int) (rNodes[mNeighbours[NodeIndex][0]].GetSolutionStepValue(DISTANCE) > 0);
```
loumalouomega:
```suggestion
        const int left_inside = (int) (rNodes[mNeighbours[NodeIndex][0]].GetSolutionStepValue(DISTANCE) > 0);
```

kratos/utilities/volume_inside_voxel_utility.cpp
```diff
@@ -0,0 +1,336 @@
...
+        int right_inside = (int) (rNodes[mNeighbours[NodeIndex][1]].GetSolutionStepValue(DISTANCE) > 0);
```
loumalouomega:
```suggestion
        const int right_inside = (int) (rNodes[mNeighbours[NodeIndex][1]].GetSolutionStepValue(DISTANCE) > 0);
```

loumalouomega (changes requested):
More comments

pooyan-dadvand (commented):
Some comments but seems that it needs a refactory.

kratos/utilities/volume_inside_voxel_utility.h
```diff
@@ -34,6 +34,17 @@ namespace Kratos
...
+/**
+ * @namespace VolumeInsideVoxedDefinition
+ * @ingroup KratosCore
+ * @brief Neighbours definition
+ * @author Ariadna Cortes
+ */
+namespace VolumeInsideVoxedDefinition
+{
+    static constexpr std::size_t Neighbours[4][2] = {{3,1},{0,2},{1,3},{2,0}};
+}
```
rubenzorrilla: I meant a `constexpr` inside the class. You don't want to make this available for the entire Kratos.

kratos/utilities/volume_inside_voxel_utility.cpp
rubenzorrilla: Adding to @loumalouomega 's suggestion that some time ago we agreed to don't have indentations within namespaces.

kratos/utilities/volume_inside_voxel_utility.h
loumalouomega: That was my proposal to solve a compilation issue with Unity. I tried to do the definition in the CPP, but it didn't work. I will try again. Sorry for the noise.

loumalouomega: Okay reverted, apparently last time when I tried I didn't save or something, but now it works

kratos/utilities/volume_inside_voxel_utility.cpp
AriadnaCortesDanes: i'll take it into account thanks!

kratos/utilities/volume_inside_voxel_utility.cpp
AriadnaCortesDanes: > This is extremely expensive. Use the closed formula from the point coordinates.

I'm not using a perfect formula but an approximation (the 4 points are approximated as two triangles). It is cheaper and will work in a perfect way for most of the use cases of the utility.

AriadnaCortesDanes:
> 



> Besides, I think it would be a good idea to rename the class to `VoxelUtilities` so we can wrap up all the future implementations in here (not only the volume ones).

Done 👍

kratos/utilities/volume_inside_voxel_utility.cpp
AriadnaCortesDanes: Doneee

kratos/utilities/volume_inside_voxel_utility.cpp
AriadnaCortesDanes: 👍

kratos/utilities/volume_inside_voxel_utility.cpp
AriadnaCortesDanes: perfect thanks!

kratos/utilities/volume_inside_voxel_utility.cpp
AriadnaCortesDanes: Yes, already removed it!

kratos/utilities/volume_inside_voxel_utility.cpp
AriadnaCortesDanes: Done

kratos/utilities/volume_inside_voxel_utility.cpp
AriadnaCortesDanes: 👍

kratos/utilities/volume_inside_voxel_utility.cpp
AriadnaCortesDanes: I need them to call  VoxelUtilities::EdgeFilledPortion(distances, ends) on each edge 😢

kratos/utilities/volume_inside_voxel_utility.cpp
AriadnaCortesDanes: Solved :)

kratos/utilities/volume_inside_voxel_utility.cpp
AriadnaCortesDanes: 👍

kratos/utilities/volume_inside_voxel_utility.cpp
AriadnaCortesDanes: already did a refactory 😄

kratos/tests/cpp_tests/utilities/test_volume_inside_voxel_utility.cpp
AriadnaCortesDanes: Okey done!

kratos/utilities/voxel_utilities.cpp
```diff
@@ -0,0 +1,307 @@
...
+/**********************************************************************************/
+/**********************************************************************************/
+
+double VoxelUtilities::NodesApproximation(const GeometryType& rVoxel) {
+    double volume = 0;
+    const auto& nodes = rVoxel.Points(); //PointsArrayType
+    const std::size_t n = nodes.size();
+    for (std::size_t i = 0; i < n; i++) {
+        if (nodes[i].GetSolutionStepValue(DISTANCE) > 0) {
+            volume+=(1.0/nodes.size());
```
pooyan-dadvand, line 33: I would use at least n instead of nodes.size() and for better performance, I would have a const variable for 1./n outside the loop:
```c++
const double nodal_volume_fraction = 1.0 / n;
for(...)
    volume_fraction += nodal_volume_fraction;

```

kratos/utilities/voxel_utilities.cpp
```diff
@@ -0,0 +1,307 @@
...
+#include "voxel_utilities.h"
+
+namespace Kratos 
+{
+constexpr std::size_t VoxelUtilities::mNeighbours[4][2];
+
+/**********************************************************************************/
+/**********************************************************************************/
+
+double VoxelUtilities::NodesApproximation(const GeometryType& rVoxel) {
```
pooyan-dadvand, line 27: The name is not clear. For example: NodalVolumeFractionApproximation

kratos/utilities/voxel_utilities.cpp
```diff
@@ -0,0 +1,307 @@
...
+    double volume = 0;
```
pooyan-dadvand, line 28: volume_fraction, no?

kratos/utilities/voxel_utilities.cpp
```diff
@@ -0,0 +1,307 @@
...
+            volume+=(1.0/nodes.size()); 
+        } 
+    }
+    return volume;
+}
+
+/**********************************************************************************/
+/**********************************************************************************/
+
+double VoxelUtilities::EdgesPortionApproximation(
```
pooyan-dadvand, line 42: Same here, the name is not mentioning the volume fraction calculation

kratos/utilities/voxel_utilities.h
```diff
@@ -0,0 +1,173 @@
...
+
+    /// Destructor
+    virtual ~VoxelUtilities(){}
+
+    ///@}
+    ///@name Operations
+    ///@{
+
+    /**
+     * @brief Aproximates the actual volume inside the voxel
```
pooyan-dadvand, line 95: Looking into your code, you are calculating the volume fraction rather than the volume. And you are not mentioning that the volume fraction is for the nodes with negative distance

kratos/utilities/voxel_utilities.h
```diff
@@ -0,0 +1,173 @@
...
+     * @param rVoxel references to the voxel whose actual volume will be approximated
+     * @return Approximated volume 
+     * @note This approximation assigns a fraction of volume (1/8) to each node of the
+     * voxel, and counts it as volume if the node is inside the object (NodeDistance > 0)
+     * This operation is VERY cheap
+     */  
+    static double NodesApproximation(const GeometryType& rVoxel);
+    
+    /**
+     * @brief Aproximates the actual volume inside the voxel
```
pooyan-dadvand, line 105: same here

kratos/utilities/voxel_utilities.h
```diff
@@ -0,0 +1,173 @@
...
+     * @param rVoxel references to the voxel whose actual volume will be approximated
+     * @param rTriangles references to the triangles which intersect the voxel at some edge.
+     * @return Approximated volume 
+     * @note This approximation finds the portion of each edge that is part of the volume (using
+     * intersection points with triangles of the mesh).
+     */  
+    static double EdgesPortionApproximation(const GeometryType& rVoxel, const GeometryArrayType& rTriangles);
+
+    /**
+     * @brief Aproximates the actual area inside a quadrilateral
```
pooyan-dadvand, line 115: You are not approximating the area of quadrilateral. You are finding the part which has inside a given boundary having nodes with negative distances

kratos/utilities/voxel_utilities.h
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * @brief Aproximates the actual area inside a quadrilateral 
+     * @param rFace references to the quadrilateral3D4 whose actual area will be approximated
+     * @param rTriangles references to the triangles which may intersect the quadrilateral at some edge.
+     * @return Approximated area 
+     */  
+    static double FaceArea(const GeometryType& rFace,const GeometryArrayType& rTriangles);
+
+    /**
+     * @brief Aproximates the portion of the edge (Line3D2) that represents volume
```
pooyan-dadvand, line 123: Again the comment is missleading

kratos/utilities/voxel_utilities.cpp
```diff
@@ -0,0 +1,307 @@
...
+/**********************************************************************************/
+
+double VoxelUtilities::EdgeFilledPortion(std::vector<double>& Distances, const PointsArrayType& rEnds) {
+    const double length = Distances[Distances.size() - 1];
+    double portion = 0; 
+    bool inside = true;       
+
+    const bool left_inside = rEnds[0].GetSolutionStepValue(DISTANCE) > 0;
+    const bool right_inside = rEnds[1].GetSolutionStepValue(DISTANCE) > 0; 
+
```
pooyan-dadvand, line 246: I would create a `const` variable for `Distances.size()` as you are using it several times.

kratos/utilities/voxel_utilities.cpp
```diff
@@ -0,0 +1,307 @@
...
+double VoxelUtilities::FaceArea(
+    const GeometryType& rFace,  
+    const GeometryArrayType& rTriangles     
+) {
+    double area = 0;
+    GeometryArrayType edges = rFace.GenerateEdges();
+    auto nodes = rFace.Points(); 
+    const double face_area = PointsArea(nodes);
+    std::vector<std::pair<double,double>> min_distance_to_node(edges.size(),{1,1});
```
pooyan-dadvand, line 83: if your algorithm only works for quadrilateral, I would suggest to use an array of size 4. (And check if the face is quadrilateral)

pooyan-dadvand (changes requested):
Thanks for the implementation! There are some minor comments.

PR #9042, opened 2021-08-10
Revert "[Core] level set process - move functions from constructor to…

ddiezrod:
This reverts commit 50d7ab514bfed192e65b63872e1d3740ac47199e.

we found that the solution we used in https://github.com/KratosMultiphysics/Kratos/pull/9027 is not really working well.

I found two issues:
- If we do not call Execute (e.g. in a test) we have an error in the destructor because mAuxModelPartName is not initialized.
- We cannot fix the dofs of the LevelSetVar because they are added in Execute.

We tried to find a solution for those two points, but it does not look good in general so we decided that we will revert these last changes and we will try to use this process directly in altair, without deriving it in our app.

@rubenzorrilla @mrhashemi

PR #10124, opened 2022-08-10
[Core] Level set convection - allowing using a custom element

ddiezrod:
**📝 Description**
We are very interested in altair in being able to use the level set convection process from kratos core. However, the main problem is that we want to be able to use our own elements, as their formulation cannot be shared openly.

With these minimum changes we will be able to do so. My proposal here is to check if the "element_type" matches the available element types in the process. If it is not there, we throw a messa and try to use this string directly and get the element from Kratos Components.

I think it is not big harm for the rest of the users and will help us a lot.

**🆕 Changelog**
- Adding option to use custom element in level set convection process.

### DIFF
--- a/kratos/processes/levelset_convection_process.h
+++ b/kratos/processes/levelset_convection_process.h
@@ -799,17 +799,8 @@ private:
     {
         mLevelSetConvectionSettings = ThisParameters;
 
-        // Convection element formulation settings
-        std::string element_type = ThisParameters["element_type"].GetString();
-        const auto element_list = GetConvectionElementsList();
-        if (std::find(element_list.begin(), element_list.end(), element_type) == element_list.end()) {
-            KRATOS_INFO("") << "Specified \'" << element_type << "\' is not in the available elements list. " <<
-            "Attempting to use custom specified element." << std::endl;
-            mConvectionElementType = element_type;
-        } else {
-            mConvectionElementType = GetConvectionElementName(element_type);
-        }
-        std::string element_register_name = mConvectionElementType + std::to_string(TDim) + "D" + std::to_string(TDim + 1) + "N";
+        std::string element_register_name = GetConvectionElementRegisteredName(ThisParameters);
+
         mpConvectionFactoryElement = &KratosComponents<Element>::Get(element_register_name);
         mElementRequiresLimiter =  ThisParameters["element_settings"].Has("include_anti_diffusivity_terms") ? ThisParameters["element_settings"]["include_anti_diffusivity_terms"].GetBool() : false;
 
@@ -843,6 +834,26 @@ private:
         // Limiter related settings
         mpLevelSetGradientVar = (mIsBfecc || mElementRequiresLevelSetGradient) ? &(KratosComponents<Variable<array_1d<double, 3>>>::Get(ThisParameters["levelset_gradient_variable_name"].GetString())) : nullptr;
         mEvaluateLimiter = (mIsBfecc || mElementRequiresLimiter) ? true : false;
+    }
+
+    std::string GetConvectionElementRegisteredName(Parameters ThisParameters)
+    {
+        // Convection element formulation settings
+        std::string element_type = ThisParameters["element_type"].GetString();
+        const auto element_list = GetConvectionElementsList();
+        if (std::find(element_list.begin(), element_list.end(), element_type) == element_list.end()) {
+            KRATOS_INFO("") << "Specified \'" << element_type << "\' is not in the available elements list. " <<
+            "Attempting to use custom specified element." << std::endl;
+            mConvectionElementType = element_type;
+        } else {
+            mConvectionElementType = GetConvectionElementName(element_type);
+        }
+        std::string element_register_name = mConvectionElementType + std::to_string(TDim) + "D" + std::to_string(TDim + 1) + "N";
+        if (!KratosComponents<Element>::Has(element_register_name)) {
+            KRATOS_ERROR << "Specified \'" << element_type << "\' is not in the available elements list: " << element_list 
+            << " and it is nor registered as a kratos element either. Please check your settings\n";
+        }
+        return element_register_name;
     }
 
     /**
@@ -897,8 +908,6 @@ private:
                 "requires_distance_gradient" : false,
                 "time_integration_theta" : 0.5
             })");
-        } else {
-            default_parameters = Parameters();
         }
 
         return default_parameters;
--- a/kratos/processes/levelset_convection_process.h
+++ b/kratos/processes/levelset_convection_process.h
@@ -802,8 +802,13 @@ private:
         // Convection element formulation settings
         std::string element_type = ThisParameters["element_type"].GetString();
         const auto element_list = GetConvectionElementsList();
-        KRATOS_ERROR_IF(std::find(element_list.begin(), element_list.end(), element_type) == element_list.end()) << "Specified \'" << element_type << "\' is not in the available elements list." << std::endl;
-        mConvectionElementType = GetConvectionElementName(element_type);
+        if (std::find(element_list.begin(), element_list.end(), element_type) == element_list.end()) {
+            KRATOS_INFO("") << "Specified \'" << element_type << "\' is not in the available elements list. " <<
+            "Attempting to use custom specified element." << std::endl;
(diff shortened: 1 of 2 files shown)


kratos/processes/levelset_convection_process.h
```diff
@@ -893,7 +898,7 @@ class LevelSetConvectionProcess
                 "time_integration_theta" : 0.5
             })");
         } else {
-            KRATOS_ERROR << "Default parameters are not implemented for the specified \'" << ElementType << "\' element. Available options are \n\t- \'levelset_convection_supg\'\n\t- \'levelset_convection_algebraic_stabilization\'" << std::endl;
+            default_parameters = Parameters();
```
rubenzorrilla: I'd throw a warning in here and then return the empty `default_parameters` declared above.

rubenzorrilla (changes requested):
I'm a bit hesitant about this... The point is that common users don't know which are the available `element_names`. Current error thrown and defaults assignment was conceived with this in mind. These changes go against this idea since, as far as I understand, the error thrown will be that of the `KratosComponents` which doesn't provide any extra information.

To be discussed with @RiccardoRossi.

rubenzorrilla:
Indeed, now I remember that we designed this in such a way that you can internally override these methods to add your custom elements.

ddiezrod:
> Indeed, now I remember that we designed this in such a way that you can internally override these methods to add your custom elements.

I tried to do that https://github.com/KratosMultiphysics/Kratos/pull/9042
I think this is not such a big change to make everything work. Anyway, I can always copy-paste this into our side and forget about kratos core, although that is not the way I prefer to do things...

rubenzorrilla:
> > Indeed, now I remember that we designed this in such a way that you can internally override these methods to add your custom elements.
> 
> I tried to do that #9042 I think this is not such a big change to make everything work. Anyway, I can always copy-paste this into our side and forget about kratos core, although that is not the way I prefer to do things...

Not very nice but maybe we can add the `custom_` keyword in front of your custom element names. Then, if the element is not found and there is no `custom_` keyword in front, we keep the current user-friendly error thrown. If the element is not found but there is the `custom_` keyword in front we go to the `KratosComponents` as you suggest. What you think?

ddiezrod:
> 
> Not very nice but maybe we can add the `custom_` keyword in front of your custom element names. Then, if the element is not found and there is no `custom_` keyword in front, we keep the current user-friendly error thrown. If the element is not found but there is the `custom_` keyword in front we go to the `KratosComponents` as you suggest. What you think?

If your concern is that user does not get a correct error message, I can check if the element exists with 

`&KratosComponents<Element>::Has(element_register_name)
`

and if it is not there throw a error message that explains clearly to the user what happened.

rubenzorrilla:
> > Not very nice but maybe we can add the `custom_` keyword in front of your custom element names. Then, if the element is not found and there is no `custom_` keyword in front, we keep the current user-friendly error thrown. If the element is not found but there is the `custom_` keyword in front we go to the `KratosComponents` as you suggest. What you think?
> 
> If your concern is that user does not get a correct error message, I can check if the element exists with
> 
> `&KratosComponents<Element>::Has(element_register_name) `
> 
> and if it is not there throw a error message that explains clearly to the user what happened.

My concern is that your proposal doesn't return the list of available elements as the error thrown is managed by the `KratosComponents`. This is IMO quite important for non-experienced users, specially for those that use the packaged version (with no code access).

ddiezrod:
@rubenzorrilla I added a test which shows the error as a user will see it, is that ok to you?

rubenzorrilla:
> @rubenzorrilla I added a test which shows the error as a user will see it, is that ok to you?

Yes. This is great, we keep previous usability while adding the possibility to use custom elements. Thanks!

rubenzorrilla (approved):

rubenzorrilla (changes requested):
Wait. Can you add the available KratosCore elements to the error thrown as we had before?

kratos/processes/levelset_convection_process.h
```diff
@@ -840,6 +836,26 @@ class LevelSetConvectionProcess
...
     }
 
+    std::string GetConvectionElementRegisteredName(Parameters ThisParameters)
+    {
+        // Convection element formulation settings
+        std::string element_type = ThisParameters["element_type"].GetString();
+        const auto element_list = GetConvectionElementsList();
+        if (std::find(element_list.begin(), element_list.end(), element_type) == element_list.end()) {
+            KRATOS_INFO("") << "Specified \'" << element_type << "\' is not in the available elements list. " <<
+            "Attempting to use custom specified element." << std::endl;
```
rubenzorrilla, line 846:
```suggestion
            KRATOS_INFO("") << "Specified \'" << element_type << "\' is not in the available elements list. Available elements are: \n\t- \'levelset_convection_supg\'\n\t- \'levelset_convection_algebraic_stabilization\'" <<     "\nAttempting to use custom specified element." << std::endl;
```
I mean this, so newbies (and users of the binary version) know which are the available elements without requiring to check the code.

kratos/processes/levelset_convection_process.h
```diff
@@ -840,6 +836,26 @@ class LevelSetConvectionProcess
...
+            mConvectionElementType = element_type;
+        } else {
+            mConvectionElementType = GetConvectionElementName(element_type);
+        }
+        std::string element_register_name = mConvectionElementType + std::to_string(TDim) + "D" + std::to_string(TDim + 1) + "N";
+        if (!KratosComponents<Element>::Has(element_register_name)) {
+            KRATOS_ERROR << "Specified \'" << element_type << "\' is not in the available elements list: " << element_list
```
rubenzorrilla, line 853: But here

```suggestion
            KRATOS_INFO("") << "Specified \'" << element_type << "\' is not in the available elements list. Available elements are: \n\t- \'levelset_convection_supg\'\n\t- \'levelset_convection_algebraic_stabilization\'" <<     "\nAttempting to use custom specified element." << std::endl;
```
I mean this, so newbies (and users of the binary version) know which are the available elements without requiring to check the code.

kratos/processes/levelset_convection_process.h
rubenzorrilla: Not here...

kratos/processes/levelset_convection_process.h
mrhashemi: As pointed out by @rubenzorrilla it is essential to somehow inform the user about using the "custom" (not default) element.

ddiezrod: Hmm I am not sure if I am following you @rubenzorrilla . The list of available elements was being printed and I think this should be an error, as the element name is not registered in Kratos so it will throw a segmentation error later.

rubenzorrilla: I did explain my self super bad... Sorry.

Forget about the previous comments and let me try again
1. We check if the element is in the elements list. If no we throw an info message saying that we will check for a custom element in the `KratosComponents`.
2. Then we do the `Has` of the `KratosComponents` as you did. If the `Has` returns `false` meaning that the custom element doesn't exist we throw the error message listing the elements i.e.
`"Specified \'" << element_type << "\' is neither in the available elements list nor registered in the KratosComponents as a custom elements. Available elements in the elements list are: \n\t- \'levelset_convection_supg\'\n\t- \'levelset_convection_algebraic_stabilization\'" << std::endl;`

(it is true that this error message is not saying which are the custom elements, but this is the best we can do I think).

I hope this clarifies.

ddiezrod: But this is exactly what I am currently doing :D . (See the error message in the test:)

ddiezrod: Only difference is that I am using the vector of element_list directly instead of explicitly writing the names. I thought this was better so if we add an element we do not have to remember to change this in the future.

rubenzorrilla: Shit. I just realized that you created an `element_list` variable. Sorry for the noise. Approving. 🤦 🤦 🤦

rubenzorrilla (approved):
Approving again as it does what it should.

ddiezrod:
Thanks @rubenzorrilla , merging!

PR #10125, opened 2022-08-11
[Core][MPI] Add Python String Broadcast

matekelemen:
Add python bindings for string broadcasting.


loumalouomega:
One comment, this will conflict with #10126 I just created, as I move the cpp tests for the sake of consistency

sunethwarna (commented):
I am ok with the changes. Lets wait untill #10126 goes through.

loumalouomega:
> One comment, this will conflict with #10126 I just created, as I move the cpp tests for the sake of consistency

In fact no conflict, I thought you also added a C++ test. Go ahead

loumalouomega (approved):

PR #10128, opened 2022-08-11
[CI] Extent errors compilation in CI

loumalouomega:
**📝 Description**

This adds errors that used to be (according to my memory) part of the ancient CI with Travis. I don't know if I don't remember properly, or it is just my imagination.

**🆕 Changelog**
- Extent errors compilation in CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,11 +43,11 @@ jobs:
           export CC=/usr/bin/gcc
           export CXX=/usr/bin/g++
           export KRATOS_CMAKE_OPTIONS_FLAGS="-DUSE_EIGEN_MKL=ON -DUSE_EIGEN_FEAST=ON -DTRILINOS_EXCLUDE_AMESOS2_SOLVER=OFF -DMMG_ROOT=/external_libraries/mmg/mmg_5_5_1/ -DPMMG_ROOT=/external_libraries/ParMmg_5ffc6ad -DINCLUDE_PMMG=ON"
-          export KRATOS_CMAKE_CXX_FLAGS="-std=c++11 -Werror -Wno-deprecated-declarations -Wignored-qualifiers"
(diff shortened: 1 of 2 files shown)


philbucher:
Dont we have Wall and Wpendantic somewhere?

Probably better to add those if they are not already there

Do you get warnings?
Can also be compiler dependent

loumalouomega:
> Dont we have Wall and Wpendantic somewhere?

We used to in Travis..., but when e migrated to Github CI we remove them....

> Probably better to add those if they are not already there

Okay

> Do you get warnings? Can also be compiler dependent

It is compiler dependent (the flags are different in clang and gcc)

loumalouomega:
> Do you get warnings? 

In my local machine if I compile with these flags it fails, I do not understand why here it works...

loumalouomega:
> > Do you get warnings?
> 
> In my local machine if I compile with these flags it fails, I do not understand why here it works...

Okay, it is because Unity and in the Fluid, we just do not compile with Unity the Core

roigcarlo:
@loumalouomega as @philbucher we already have `-Wall`. We can add this but it won't add any other flag.

loumalouomega:
> @loumalouomega as @philbucher we already have `-Wall`. We can add this but it won't add any other flag.

So if we want to add flags?

roigcarlo:
Not sure what you mean, you can add all the flags you want, but they are already included in the -Wall

roigcarlo:
Here: https://github.com/KratosMultiphysics/Kratos/blob/master/.github/workflows/configure.sh#L67

loumalouomega:
> Not sure what you mean, you can add all the flags you want, but they are already included in the -Wall

But not as error

roigcarlo:
Yes, the werror comes from the KRATOS_CMAKE_FLAGS which is added before the Wall of the configure, so every warning is still an error.

loumalouomega:
> Yes, the werror comes from the KRATOS_CMAKE_FLAGS which is added before the Wall of the configure, so every warning is still an error.

Okay, so I will cherrypick the fix of this PR and close this

PR #10130, opened 2022-08-11
[Core][MPI] Adding SynchronizeCurrentDataToMax and SynchronizeNonHistoricalDataToMax to MPI communicator

### LINKED ISSUE #10129: MPI - How to synchronize non historical data to max?
**Description**
I have the following code working in SMP (in the original this is cpp), but I think this cannot be synchronized right now in MPI because there is no SynchronizeNonHistoricalDataToMax (only ToMin). Maybe there is some reason for this as this can be done in a different way but I have not been able to find it yet. This is the pseudo-code.
```
for node in model_part.Nodes:
        node.SetValue(VARIABLE_1, 0.0)

for condition in model_part.Conditions:
    if condition1(condition):
         for node in condition:
                if condition2(node):
                     new_value = function(condition,node)
                     tmp = max(node.GetValue(VARIABLE_1), new_value)
                     node.SetValue(VARIABLE_1, tmp)

#Nodes are not synchronized in MPI at this point

```

Is there any solution to this apart from adding SynchronizeNonHistoricalDataToMax? Any reason why this was not added before?

@KratosMultiphysics/mpi @KratosMultiphysics/technical-committee @KratosMultiphysics/implementation-committee @roigcarlo 

**Scope**
- KratosMPI

loumalouomega:
**📝 Description**

Adding SynchronizeCurrentDataToMax and SynchronizeNonHistoricalDataToMax. Fixes https://github.com/KratosMultiphysics/Kratos/issues/10129

**🆕 Changelog**

- Added max value in ReduceValues
- Adding SynchronizeCurrentDataToMax and SynchronizeNonHistoricalDataToMax
- Added tests

### DIFF
--- a/kratos/includes/communicator.h
+++ b/kratos/includes/communicator.h
@@ -347,13 +347,27 @@ public:
 
     virtual bool SynchronizeNonHistoricalVariable(Variable<Quaternion<double>> const& rThisVariable);
 
-    /// Synchronize variable in nodal solution step data to the minimum value across all processes.
-    /** @param ThisVariable The variable to be synchronized.
+    /** 
+     * @brief Synchronize variable in nodal solution step data to the maximum value across all processes.
+     * @param ThisVariable The variable to be synchronized.
+     */
+    virtual bool SynchronizeCurrentDataToMax(Variable<double> const& ThisVariable);
+
+    /** 
+     * @brief Synchronize variable in nodal data to the maximum value across all processes.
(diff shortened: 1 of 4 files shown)


loumalouomega:
Ping @roigcarlo

roigcarlo (approved):
OK. BTW  I think the we did not implement the max because you can just synch min to the neg value. But I don't remember so...

PR #10131, opened 2022-08-12
[Core] Zero initialization of array_1d

loumalouomega:
**📝 Description**

Zero initialization of `array_1d`. This avoids some error compilation if `-Werror=unused-variable`

**🆕 Changelog**

- Zero initialization of array_1d

### DIFF
--- a/kratos/containers/array_1d.h
+++ b/kratos/containers/array_1d.h
@@ -722,7 +722,7 @@ private:
     ///@name Member Variables
     ///@{
 
-    array_type data_;
+    array_type data_{};
 
     ///@}
     ///@name Private Operators


sunethwarna:
Wouldn't this make all the `array_1d` vairables to initialize with zeros which is redundant and time consuming if someone declares the `array_1d` variable within a gauss loop and assigns its components in next lines?

RiccardoRossi (changes requested):
in my opinion this is wrong.

you are initializing an array, say if size 3, with a list of size1.

if we were in devig we should actually throw an error.

on the ither hand we do not want to initialize always to zero, as it is time consuming

loumalouomega:
> in my opinion this is wrong.
> 
> you are initializing an array, say if size 3, with a list of size1.
> 
> if we were in devig we should actually throw an error.
> 
> on the ither hand we do not want to initialize always to zero, as it is time consuming

I will close, but then we cannot activate `-Werror=unused-variable` as compilation flag

PR #10133, opened 2022-08-13
[Core] Adding possibility to have hints for cpp libraries in IDEs

sunethwarna:
**📝 Description**
This PR adds the possibility to automate stub `pyi` file creation for pybind11 libraries for each app, making possible to have hints given by IDEs for cpp libraries as well.  This works with all the application extensions as well.

following json settings can be put in `.vscode/settings.json` if you have installed kratos in `/software/kratos/master/bin/gcc/Release`, then you will see hints for all the cpp libraries you have compiled along with the hints for the existing python libraries as well.

```json
{
    "python.analysis.extraPaths": [
        "/software/kratos/master/bin/gcc/Release"
    ]
}
```

** Todo:
- [x]  Add to windows compilation defaults
- [x]  Add to mac os compilation defaults
- [x]  Add this to python wheel generation
- [x]  Add an automated way to detect installation directory

**🆕 Changelog**
- Added `stub_generation.py` script
- Added `pyi` stub file generation to `standard_configure.sh`

fyi @matekelemen


sunethwarna:
@roigcarlo Is it possible to always have the installation directory set in the configuration script and not to set it in `CMakeLists.txt` (where  a default is assumed if `CMAKE_INSTALL_PREFIX` is not given in cmake options)?

roigcarlo:
Well yes and no. Is possible to remove the default in the CMakeLists.txt but cmake will set the default somewhere else

roigcarlo:
But you can always set it i the configure and that one will take priority

matekelemen:
Great find, thanks @sunethwarna!

matekelemen:
Regarding the to-do list:

With some modifications, the generation script could be invoked from CMake. It would have all the information it needs in that case; the question is how people feel about adding this step to the main ```CMakeLists.txt```.

sunethwarna:
@matekelemen Thats a really good idea. May be we can make it optional by adding a cmake flag which will by default will make these files, but in the case of CI, we can omit creating these stub files to avoid extra CI time in each PR.

matekelemen:
I managed to attach the generated stubs to Kratos package, so there's no need to edit anything in *.vscode*. One catch is that I don't really understand how/where auxiliary libraries (such as ```FluidDynamicsTrilinosExtension```) get loaded, so I haven't touched those.

matekelemen:
Stubs generated from the binaries shadow the rest of the symbols in the package if they're in the same directory, so I reverted moving them. Unless we come up with a better idea, @sunethwarna's solution is still the way to go (add the extra paths to the VSCode config file).

scripts/post_install/stub_generation.py
```diff
@@ -0,0 +1,116 @@
...
+    generate_stubs(options)
+
+    # if sub modules are found, include them in the __init__.pyi
+    submodule_path = kratos_library_path / cpp_module_name
+    __AddSubmodules(submodule_path)
+
+    return cpp_module_name
+
+
+def __MoveGeneratedStub(source: Path, destination: Path) -> None:
```
sunethwarna: do we need this still?

matekelemen: I'd like to revisit this later, so I'd keep it until then. Hopefully I'll find a way to make python and binary symbols to coexist.

scripts/post_install/CMakeLists.txt
```diff
@@ -1,10 +1,12 @@
 set(KRATOS_GENERATE_PYTHON_STUBS ON CACHE BOOL "Generate hints for binary python packages")
 
-if (${KRATOS_GENERATE_PYTHON_STUBS})
-    if (NOT DEFINED PYTHON_EXECUTABLE)
-        set(PYTHON_EXECUTABLE "python")
-    endif()
+if (${CMAKE_SYSTEM_NAME} MATCHES "Linux") # todo: test the stub generation on other platforms too
```
sunethwarna: Can we leave out the temporary restriction? @roigcarlo  will check it in windows.

roigcarlo:
From my side ok :+1: . I would love to give you more feedback but my dev invironment generate all of this from the sources I work on automatically, so I don't want to tamper with it. If it is usefull green light from my side.

@sunethwarna I saw that you marked "add to wheels" but I don't see any change related to this. Do you intend for this pyi to get shipped in the pip package?

scripts/post_install/stub_generation.py
```diff
@@ -33,102 +32,166 @@ def __AddSubmodules(output_path: Path):
             __AddSubmodules(cpp_sub_module_path)
 
 
-def __binary_module_from_name(module_name: str) -> str:
-    """Get the binary module's stub from its name."""
-    return "Kratos" + module_name
+def __FindCPPModuleImportsInPythonModules(current_path: Path, cpp_module_names_list: list[str], cpp_python_modules_dict: dict[str, Path]):
```
matekelemen:
```suggestion
def __FindCPPModuleImportsInPythonModules(current_path: Path, cpp_module_names_list: "list[str]", cpp_python_modules_dict: "dict[str, Path]") -> None:

```

scripts/post_install/stub_generation.py
```diff
@@ -33,102 +32,166 @@ def __AddSubmodules(output_path: Path):
...
-    if binary_module_name.startswith("Kratos"):
-        return binary_module_name[6:]
-    else:
-        return binary_module_name
+        for sub_path in current_path.iterdir():
+            if sub_path.is_dir():
+                __FindCPPModuleImportsInPythonModules(sub_path, cpp_module_names_list, cpp_python_modules_dict)
+
+
+def __GetPythonModulesImportingCppModules(kratos_python_module_path: Path, kratos_library_path: Path) -> dict[Path, Path]:
```
matekelemen:
```suggestion
def __GetPythonModulesImportingCppModules(kratos_python_module_path: Path, kratos_library_path: Path) -> "dict[Path, Path]":

```

sunethwarna:
@roigcarlo 

> @sunethwarna I saw that you marked "add to wheels" but I don't see any change related to this. Do you intend for this pyi to get shipped in the pip package?

These pyi stub files are created automatically unless `KRATOS_GENERATE_PYTHON_STUBS=OFF` is given to the cmake configuration as in the CI configure files. So wheels also should have it automatically. Is there a way to test this?

scripts/post_install/stub_generation.py
```diff
@@ -33,102 +32,166 @@ def __AddSubmodules(output_path: Path):
...
+    # generate list of cpp modules
+    list_of_binary_modules = []
+    for custom_library_path in kratos_library_path.iterdir():
+        custom_library_name = str(custom_library_path.relative_to(custom_library_path.parent))
```
matekelemen:
```suggestion
        custom_library_name = custom_library_path.name

```

scripts/post_install/stub_generation.py
```diff
@@ -33,102 +32,166 @@ def __AddSubmodules(output_path: Path):
...
         kratos_library_path: Path,
         output_path: Path,
-        cpp_module_name: str) -> None:
-    args = ["-o", str(output_path.absolute()), "-p", cpp_module_name]
-    options = parse_options(args)
-    generate_stubs(options)
+        cpp_module_name: str,
+        python_import_module_path_dict: dict[str, Path]) -> None:
+
+    def __generate(output_path: Path, cpp_module_name: str, custom_args: list[str]):
```
matekelemen:
```suggestion
    def __generate(output_path: Path, cpp_module_name: str, custom_args: "list[str]"):

```

roigcarlo:
Oh, I thought it was disabled by default. Usually all options (except for build testing) default to off, specially if they depend on external libs/packages.

Anyway, the wheel builder machines does not have `mypy` I can confirm you that so it won't be generated.

matekelemen (commented):
type hints with arguments (eg.: ```list[str]```) are a python 3.9 feature. I usually put them in quotes until we drop support for older python versions (```"list[str]"```).

scripts/post_install/stub_generation.py
```diff
@@ -0,0 +1,206 @@
...
+        print("                    " + "\n                    ".join(list_of_binary_modules))
+
+    return copy_list_of_binary_modules, cpp_python_modules_dict
+
+def __GenerateStubFilesForModule(
+        kratos_python_module_path: Path,
+        kratos_library_path: Path,
+        output_path: Path,
+        cpp_module_name: str,
+        python_import_module_path_dict: dict[str, Path]) -> None:
```
matekelemen:
```suggestion
        python_import_module_path_dict: "dict[str, Path]") -> None:

```

sunethwarna:
@roigcarlo I think this is ready for your review. :)

roigcarlo:
@sunethwarna Looks good, but I think the pip module for the linux docker is missing

sunethwarna:
@roigcarlo I am not 100% sure about the wheel changes I did, I added mypy for linux wheels also. Could you double check?

scripts/docker_files/docker_file_wheelbuilder_linux_mpi/Dockerfile
```diff
@@ -34,6 +34,8 @@ RUN git clone -b 'v5.5.1' --depth 1 https://github.com/MmgTools/mmg /workspace/m
     rm -r /workspace/mmg_5_5_1 && \
     cd /
 
+RUN python3.6 -m pip install mypy
```
roigcarlo: that's it, but you need to do it for 3.6, 3.7, 3.8, 3.9 and 3.10

I think the binary are located at `/opt/python/VERSION/bin/python`

loumalouomega: 3.6 is already legacy...

sunethwarna: I am doing it in the following place also. 
https://github.com/KratosMultiphysics/Kratos/blob/950b67f72ebe36c1127f9f4c52814aaf1d760da3/scripts/wheels/linux/build.sh#L161

I saw the core is built against pytho3.6, and thats why i included it in there as well. Am i wrong?

sunethwarna: https://github.com/KratosMultiphysics/Kratos/blob/950b67f72ebe36c1127f9f4c52814aaf1d760da3/scripts/wheels/linux/build.sh#L149

roigcarlo: No you both are right. Sadly 3.6 is still not legacy for us (but it soon will be if it depends on me :imp: ).

The build process in the wheel builder is different from the one you are used in regular builds. In order to save time we make a single compilation of the "Core" .so and we use those to link agains different version of the "application" ,so. The core elements do not depend on python, but I didn't want to rewrite all the cmake files so I simply pass a placeholder python version so it does not complain.

So we build:
KratosCore.so

Then
Kratos.p36.so, Kratos.p37.so ... Kratos.p310.so

Which all link to the same Kratos.so. But the commands in the cmake builf will be invoked with the proper version of python (the one inside the loop below the build_core instruction)

sunethwarna: In that case we don't need to have the following ryt?
`RUN python3.6 -m pip install mypy`, because it is anyways installed for the respective version from `$PYTHON_LOCATION -m pip install mypy ` ryt?

roigcarlo: aaa you are right sorry, didn't see that change! Exactly as you say.

sunethwarna: @roigcarlo  done :)

roigcarlo (approved):
Good! If I see that something fails on the wheel side I will fix it myself and ping you. Gj!!

sunethwarna:
@roigcarlo  Thanks :)

PR #10136, opened 2022-08-14
[Core][CMake] Adding additional checks C++17

loumalouomega:
**📝 Description**

Adding additional checks C++17

**🆕 Changelog**

- Adding additional checks C++17


philbucher:
why is this necessary?
We enforce C++17, it is not optional

loumalouomega:
> why is this necessary? We enforce C++17, it is not optional

A priori checks the compiler before it starts to compile (in my experience, just the cmake set doesn't work). But this PR requires a newer version of CMake and the CI is failing, closing.

loumalouomega:
> > why is this necessary? We enforce C++17, it is not optional
> 
> A priori checks the compiler before it starts to compile (in my experience, just the cmake set doesn't work). But this PR requires a newer version of CMake and the CI is failing, closing.

I simplified the thing @philbucher

PR #10138, opened 2022-08-15
[Fluid] Changes the severity of FluidDynamicsApplication tests and add cpp tests run.

sunethwarna:
**📝 Description**
Changes the logger severity for fluid tests and makes sure cpp tests are also executed when `test_FluidDynamicsApplication.py` is executed.

**🆕 Changelog**
- Changed default severity of `FluidDynamicsApplication` tests
- Make sure cpp tests are also run


applications/FluidDynamicsApplication/tests/test_FluidDynamicsApplication.py
```diff
@@ -160,4 +163,12 @@ def AssembleTestSuites():
     return suites
 
 if __name__ == '__main__':
+    KratosMultiphysics.Logger.GetDefaultOutput().SetSeverity(
+        KratosMultiphysics.Logger.Severity.WARNING)
+    KratosMultiphysics.Logger.PrintInfo("Unittests", "\nRunning cpp unit tests ...")
+    run_cpp_unit_tests.run()
+    KratosMultiphysics.Logger.PrintInfo("Unittests", "Finished running cpp unit tests!")
```
rubenzorrilla: This will crash if cpp tests are not compiled right? (@roigcarlo I think this is what we were discussing a few days ago).

sunethwarna: That is true. Sorry, I will remove it. I remember now, this will make cpp tests to run twice also ryt @roigcarlo ?

philbucher: yes they are run separately hence this should not be here

roigcarlo: More ore less what has been said, we can not add it automatically because there are instances of compilations where that does not exists and breaks.

If its really needed a simple try catch would do, but I recommend not to do it. If cpp are needed they should be run from the run_tests_scripts with the `-a` flag. Probably we should lift the restriction of having to run the core tests in that case from our side, but that's another discussion....

sunethwarna: Done and removed 👍

roigcarlo:
To me looks ok!

roigcarlo (approved):

PR #10142, opened 2022-08-16
[Several applications] Adding `if constexpr` from C++17 to optimize code

loumalouomega:
**📝 Description**

Adding `if constexpr` from C++17 to optimize code in the parts of the code where is possible. Mostly in the templated classes.

This improves performance as the if is computed in compilation time, so the if is not executed in running time, saving time specially in parts of the code executed recursively.

**🆕 Changelog**

- Adding `if constexpr` from C++17 to optimize code in the parts of the code where is possible

### DIFF
--- a/applications/StructuralMechanicsApplication/custom_conditions/small_displacement_line_load_condition.cpp
+++ b/applications/StructuralMechanicsApplication/custom_conditions/small_displacement_line_load_condition.cpp
@@ -143,8 +143,8 @@ void SmallDisplacementLineLoadCondition<TDim>::CalculateAll(
     Vector pressure_on_nodes = ZeroVector( number_of_nodes );
 
     // Pressure applied to the element itself
-    double pressure_on_condition = 0.0;
     if constexpr (TDim == 2) {
+        double pressure_on_condition = 0.0;
         if( this->Has( PRESSURE ) ) {
             pressure_on_condition += this->GetValue( PRESSURE );
         }
--- a/applications/StructuralMechanicsApplication/custom_conditions/small_displacement_line_load_condition.cpp
+++ b/applications/StructuralMechanicsApplication/custom_conditions/small_displacement_line_load_condition.cpp
@@ -144,7 +144,7 @@ void SmallDisplacementLineLoadCondition<TDim>::CalculateAll(
 
     // Pressure applied to the element itself
     double pressure_on_condition = 0.0;
-    if (TDim == 2) {
+    if constexpr (TDim == 2) {
         if( this->Has( PRESSURE ) ) {
             pressure_on_condition += this->GetValue( PRESSURE );
         }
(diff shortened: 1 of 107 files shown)


loumalouomega:
@roigcarlo @rubenzorrilla if we want to avoid conflicts I would merge this ASAP

applications/StructuralMechanicsApplication/custom_conditions/small_displacement_line_load_condition.cpp
```diff
@@ -143,8 +143,8 @@ void SmallDisplacementLineLoadCondition<TDim>::CalculateAll(
     Vector pressure_on_nodes = ZeroVector( number_of_nodes );
 
     // Pressure applied to the element itself
-    double pressure_on_condition = 0.0;
-    if (TDim == 2) {
+    if constexpr (TDim == 2) {
+        double pressure_on_condition = 0.0;
```
roigcarlo, line 147: Why is this inside the if now? (not familiar with the code, but in some other similar codes you left it ouside)

kratos/processes/apply_periodic_boundary_condition_process.cpp
```diff
@@ -226,7 +226,7 @@ void ApplyPeriodicConditionProcess::ConstraintSlaveNodeWithConditionForVectorVar
             mrMasterModelPart.AddMasterSlaveConstraint(constraint6);
 
 
-            if (TDim == 3)
+            if (TDim == 3) // TODO: This function can be optimized using template specialization. if constexpr will fail due to unused variables declared
```
roigcarlo, line 229: Either way we should decide on going the constexp way or specializing, but both at the same time are ortogonal.

roigcarlo (commented):
I've gone through and, in principle, I see no problem merging, but please 2nd review. I am afraid of long PR.

Just to confirm, the only change is `if` <-> `if constexp`

kratos/processes/apply_periodic_boundary_condition_process.cpp
loumalouomega: In this particular piece of code I would go for template specialization, otherwise will be a bit ugly. Just adding the comment for the future

applications/StructuralMechanicsApplication/custom_conditions/small_displacement_line_load_condition.cpp
loumalouomega: Because otherwise clang complains (because unused variable if TDim is different from 2). The variable is only used in the if anyway

roigcarlo (approved):

rubenzorrilla (approved):
I went one by one and I think everything is correct.

pooyan-dadvand:
I think a good addition would be add it to the DEBUG_ERROR macros to be sure it does not affect the performance in release.

loumalouomega:
> I think a good addition would be add it to the DEBUG_ERROR macros to be sure it does not affect the performance in release.

It is not required, as it uses the post processor which is also resolved in compilation time

PR #10143, opened 2022-08-16
[Core] Allowing to consider two different types for scalar AtomicAdd

loumalouomega:
**📝 Description**

Allowing to consider two different types for scalar AtomicAdd. This is done in the vector types as well.

**🆕 Changelog**

-  [Allowing to consider two different types for scalar AtomicAdd](https://github.com/KratosMultiphysics/Kratos/commit/92c9253e1e7c1b8da4480e886ddb0dd84161d438)
- [Check not overflow in debug](https://github.com/KratosMultiphysics/Kratos/commit/dcaa4e7d1d33d99f474dc21f3967b3f9005c6e1a)

### DIFF
--- a/kratos/utilities/atomic_utilities.h
+++ b/kratos/utilities/atomic_utilities.h
@@ -37,7 +37,7 @@ namespace Kratos
  * @param target variable being atomically updated by doing target += value
  * @param value value being added
  */
-template<class TDataType>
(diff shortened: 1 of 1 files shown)


loumalouomega:
This is not needed anymore after the fix in #10144

PR #10144, opened 2022-08-16
[Core] fix error in omp in monotonicity preserving in OMP

ddiezrod:
**📝 Description**
I found that there is an error in monotonicity preserving solver. In the for loop it makes sure that for a given (i,j) that j is greater than i. However it might that for the same i, there are two j values that satisfies rA(i,j) > 0.0 and therefore, there might be two threads that try to modify at the same time rA(i,i) or rA(j,j) and same for rB. Adding a critical to avoid this.

**🆕 Changelog**
E.g.
- Fixing OMP bug in monotonicity preserving solver

### DIFF
--- a/kratos/linear_solvers/monotonicity_preserving_solver.h
+++ b/kratos/linear_solvers/monotonicity_preserving_solver.h
@@ -210,12 +210,15 @@ public:
                     if (value > 0.0) {
                         const auto j = index2_vector[k];
                         if (j > i) {
-                            rA(i,i) += value;
                             rA(i,j) -= value;
                             rA(j,i) -= value;
-                            rA(j,j) += value;
-                            rB[i] += value*dofs_values[j] - value*dofs_values[i];
-                            rB[j] += value*dofs_values[i] - value*dofs_values[j];
+                            #pragma omp critical
+                            {
+                                rA(i,i) += value;
+                                rA(j,j) += value;
+                                rB[i] += value*dofs_values[j] - value*dofs_values[i];
+                                rB[j] += value*dofs_values[i] - value*dofs_values[j];
+                            }
                         }
                     }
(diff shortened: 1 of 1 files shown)


roigcarlo (commented):
Should be done with locks on the matrix cell, otherwise it won't work if we decide to swap omp backends

loumalouomega (approved):

roigcarlo:
@loumalouomega I think we should at least discuss about the problem of using `omp critial` inside one of the backend agnostic before approving...

ddiezrod:
> @loumalouomega I think we should at least discuss about the problem of using `omp critial` inside one of the backend agnostic functions before approving...

He replaced the critical with AtomicAdd. Do we still need https://github.com/KratosMultiphysics/Kratos/pull/10143?

ddiezrod:
> > @loumalouomega I think we should at least discuss about the problem of using `omp critial` inside one of the backend agnostic functions before approving...
> 
> He replaced the critical with AtomicAdd. @loumalouomega Do we still need #10143?

Ok, I see we do not.

loumalouomega:
> > @loumalouomega I think we should at least discuss about the problem of using `omp critial` inside one of the backend agnostic functions before approving...
> 
> He replaced the critical with AtomicAdd. @loumalouomega Do we still need #10143?

Nope, closed

roigcarlo:
oh, I missed the commit, my bad

loumalouomega:
> @loumalouomega I think we should at least discuss about the problem of using `omp critial` inside one of the backend agnostic functions before approving...

I approved after using AtomicAdd

loumalouomega:
> oh, I missed the commit, my bad

I think everything is happening so fast we are confused and we attack ourselves

PR #10147, opened 2022-08-17
[StructuralMechanicsApp] Bugfix and some optimizations

AlejandroCornejo:
In this PR I am solving a bug in which the initial state was not correctly set in parallel if you perform a custom elem->SetValue of the initial stress or strains. Now it is solved.

Besides I have added some optimizations of the elements (noalias missing mainly)


loumalouomega (approved):
Very nice

PR #10149, opened 2022-08-17
[StructuralMechanicsApplication] Added moving load with rotational and translational DOF

aronnoordam:
**📝 Description**

WIP

Adding a moving load

**🆕 Changelog**

- Adding a moving load


applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,319 @@
...
+
+// Project includes
+#include "set_moving_load_process.h"
+#include "utilities/interval_utility.h"
+#include "structural_mechanics_application_variables.h"
+
+
+namespace Kratos
+{
+
```
rubenzorrilla, line 29: These functions are declared in the entire Kratos namespaces. These must be `private` methods of your process.

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,319 @@
...
+}
+
+
+void SetMovingLoadProcess::ExecuteInitialize()
+{
+    KRATOS_TRY
+    mLoad = mParameters["load"].GetVector();
+    const double current_time = mrModelPart.GetProcessInfo().GetValue(TIME);
+
+    vector<int> direction = mParameters["direction"].GetVector();
```
rubenzorrilla:
```suggestion
   const array_1d<double,3> direction = mParameters["direction"].GetVector();
```
These components are not necessarily unitary. Hence you should use a floating point array and then normalize it by taking its norm with `norm_2`.

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,319 @@
...
+    
+
+    return sorted_conditions;
+}
+
+
+void SetMovingLoadProcess::ExecuteInitialize()
+{
+    KRATOS_TRY
+    mLoad = mParameters["load"].GetVector();
```
rubenzorrilla: Why saving both the `mParameters` and `mLoad` as member variables?

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,319 @@
...
+    // get all end node ids ( not the middle nodes, in case of line3 conditions)
+    for (auto& r_cond : mrModelPart.Conditions()) {
+
+        auto geom = r_cond.GetGeometry();
+        node_id_vector.push_back(geom[0].Id());
+        node_id_vector.push_back(geom[1].Id());
+    }
+
+    // find non repeating node ids
+    std::vector<int> non_repeating_node_ids = FindNonRepeatingIndices(node_id_vector);
```
rubenzorrilla: I think you can easily achieve the same with an `unordered_set`.

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,319 @@
...
+    for (auto& r_cond : mSortedConditions)
+    {
+        auto& r_geom = r_cond.GetGeometry();
+        double element_length = r_geom.Length();
+
+        // if moving load is located at current condition element, apply moving load, else apply a zero load
+        if ((distance_cond + element_length >= mCurrentDistance) && (distance_cond <= mCurrentDistance) && !is_moving_load_added)
+        {
+            
+            r_cond.SetValue(MOVING_LOAD, mLoad);
```
rubenzorrilla: I don't think it is a good idea to create a new `MOVING_LOAD` variable as you'll require to update all the conditions in order to assemble it to the local system. Current Kratos load variables should be used instead.

applications/StructuralMechanicsApplication/python_scripts/set_moving_load_process.py
```diff
@@ -0,0 +1,86 @@
...
+
+        #The value can be a double or a string (function)
+        default_settings = KratosMultiphysics.Parameters("""
+        {
+            "help"            : "This process applies a moving load condition belonging to a modelpart. The load moves over line elements.",
+            "model_part_name" : "please_specify_model_part_name",
+            "variable_name"   : "MOVING_LOAD",
+            "load"            : [0.0, 1.0, 0.0],
+            "direction"       : [1,1,1],
+            "velocity"        : 1
```
rubenzorrilla: Why not adding the velocity vector instead? If doing it in that way we can eventually add a function depending on x,y,z and t for each component.

applications/StructuralMechanicsApplication/python_scripts/set_moving_load_process.py
```diff
@@ -0,0 +1,86 @@
...
+
+        KratosMultiphysics.Process.__init__(self)
+
+        #The value can be a double or a string (function)
+        default_settings = KratosMultiphysics.Parameters("""
+        {
+            "help"            : "This process applies a moving load condition belonging to a modelpart. The load moves over line elements.",
+            "model_part_name" : "please_specify_model_part_name",
+            "variable_name"   : "MOVING_LOAD",
+            "load"            : [0.0, 1.0, 0.0],
```
rubenzorrilla: Shouldn't we ask for the origin point as well?

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
rubenzorrilla: OK. Now I see, you created your custom condition.

rubenzorrilla: In any case, I'd still use `POINT_LOAD`.

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,270 @@
...
+// System includes
+
+
+// External includes
+
+
+// Project includes
+#include "custom_conditions/moving_load_condition.h"
+
+#include <includes/checks.h>
```
loumalouomega:
```suggestion
#include "includes/checks.h"
```

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,270 @@
...
+    // Resizing as needed the LHS
+    unsigned int MatSize = NumberOfNodes * block_size;
+   /* for (const auto& r_node : GetGeometry().Points()) {
+        if (r_node.HasDofFor(ROTATION_X)) { MatSize += 1; }
+        if (r_node.HasDofFor(ROTATION_Y)) { MatSize += 1; }
+        if (r_node.HasDofFor(ROTATION_Z)) { MatSize += 1; }
+    }*/
+
+
+    if ( CalculateStiffnessMatrixFlag == true ) //calculation of the matrix is required
```
loumalouomega: The style is 
if (...) {
} else {
}

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,270 @@
...
+    if( this->Has(MOVING_LOAD ) )
+    {
+        noalias(MovingLoad) = this->GetValue( MOVING_LOAD );
+    }
+
+    // check if cond should be calculated
+    bool is_moving_load = false;
+    for (int i = 0; i < TDim; ++i)
+    {
+        if (abs(MovingLoad[i]) > DBL_EPSILON)
```
loumalouomega:
```suggestion
        if (std::abs(MovingLoad[i]) > DBL_EPSILON)
```
This is very dangerous. abs returns a integer, 0.01 will be 0

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+//************************************************************************************
+//************************************************************************************
+template< std::size_t TDim, std::size_t TNumNodes >
+Condition::Pointer MovingLoadCondition< TDim, TNumNodes>::Create( IndexType NewId, NodesArrayType const& ThisNodes,  PropertiesType::Pointer pProperties ) const
+{
+    return Kratos::make_intrusive<MovingLoadCondition>( NewId, GetGeometry().Create( ThisNodes ), pProperties );
+}
+
+/***********************************************************************************/
```
loumalouomega, line 57: Unify separeator style

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.h
```diff
@@ -0,0 +1,323 @@
...
+
+///@}
+///@name  Functions
+///@{
+
+///@}
+///@name Kratos Classes
+///@{
+
+/// Short class definition.
```
loumalouomega: Add description

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,319 @@
...
+
+    // Traverse through map only and
+    std::vector<int> non_repeating_indices;
+    for (auto x : mp)
+        if (x.second == 1)
+            non_repeating_indices.push_back(x.first);
+
+    return non_repeating_indices;
+          
+    }
```
loumalouomega: Extra spaces
```suggestion
}
```

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,319 @@
...
+        return end_conditions[0];
+    }
+    	return end_conditions[1];
+
+}
+
+Condition& GetFirstCondition(Point first_point, Point second_point, vector<int> direction, std::vector<Condition>& end_conditions)
+{
+    // sort on x-coord, if x coords are equal, sort on y coord, if y coord is equal sort on z-coord
+    if (abs(first_point[0] - second_point[0]) > DBL_EPSILON)
```
loumalouomega:
```suggestion
    if (std::abs(first_point[0] - second_point[0]) > DBL_EPSILON)
```

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,319 @@
...
+    {
+        return GetFirstConditionFromCoord(first_point[0], second_point[0], direction[0], end_conditions);
+    }
+    if (abs(first_point[1] - second_point[1]) > DBL_EPSILON)
```
loumalouomega:
```suggestion
    if (std::abs(first_point[1] - second_point[1]) > DBL_EPSILON)
```

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,319 @@
...
+}
+
+
+std::vector<Condition> BubbleSortConditions(ModelPart::ConditionsContainerType& unsorted_conditions, Condition& first_condition)
+{
+    //ModelPart::ConditionsContainerType& sorted_conditions = unsorted_conditions;
+    std::vector<Condition> unsorted_conditions_v(unsorted_conditions.begin(), unsorted_conditions.end());
+
+    for (int i = 0; i < unsorted_conditions_v.size(); ++i)
+    {
```
loumalouomega: Style of {} please (in all of them)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
loumalouomega:
```suggestion
    std::vector<int> direction = mParameters["direction"].GetVector();
```
Otherwise it will consider the first vector it founds

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,319 @@
...
+                {
+                    end_conditions.push_back(r_cond);
+                }
+            }    
+        }
+    }
+
+
+    // find start condition 
+    Point center_1 = end_conditions[0].GetGeometry().Center();
```
loumalouomega:
```suggestion
    const Point center_1 = end_conditions[0].GetGeometry().Center();
```

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,319 @@
...
+    Point center_2 = end_conditions[1].GetGeometry().Center();
```
loumalouomega:
```suggestion
    const Point center_2 = end_conditions[1].GetGeometry().Center();
```

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,319 @@
...
+
+    Condition& r_first_cond = GetFirstCondition(center_1, center_2, direction, end_conditions);
+
+    mSortedConditions = BubbleSortConditions(mrModelPart.Conditions(), r_first_cond);
+
+    auto tmp = r_first_cond.GetGeometry();
+
+    int a = 1 + 1;
+    //for (int i = 0; i < arr_size; i++) {
+    //    node_id_vector[node_id_vector[i] % arr_size]
```
loumalouomega: Remove if unused

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,319 @@
...
+            r_cond.SetValue(MOVING_LOAD_LOCAL_DISTANCE, mCurrentDistance - distance_cond);
+            is_moving_load_added = true;
+        }
+        else
+        {
+            r_cond.SetValue(MOVING_LOAD, ZeroVector(3));
+        }
+        distance_cond += element_length;
+    }
+ /*   const double current_time = mrModelPart.GetProcessInfo().GetValue(TIME);
```
loumalouomega: Remove if unuded

applications/StructuralMechanicsApplication/python_scripts/set_moving_load_process.py
```diff
@@ -0,0 +1,86 @@
...
+        """ This method is executed in order to initialize the current step
+
+        Keyword arguments:
+        self -- It signifies an instance of a class.
+        """
+
+        self.process.ExecuteInitialize()
+
+    def ExecuteInitializeSolutionStep(self):
+    #     """ This method is executed in order to initialize the current step
```
loumalouomega: Can the documentation follow the style of the previous block?

applications/StructuralMechanicsApplication/python_scripts/set_moving_load_process.py
```diff
@@ -0,0 +1,86 @@
...
+        params_load.AddValue("velocity", settings["velocity"])
+
+        # Set process
+        self.model_part = Model.GetModelPart(params_load["model_part_name"].GetString())
+        self.process = KSM.SetMovingLoadProcess(self.model_part, params_load)
+
+    def ExecuteInitialize(self):
+        """ This method is executed in order to initialize the current step
+
+        Keyword arguments:
```
loumalouomega: Like this

applications/StructuralMechanicsApplication/python_scripts/set_moving_load_process.py
```diff
@@ -0,0 +1,86 @@
...
+    #     """ This method is executed in order to initialize the current step
+    #
+    #     Keyword arguments:
+    #     self -- It signifies an instance of a class.
+    #     """
+        self.process.ExecuteInitializeSolutionStep()
+
+    def ExecuteFinalizeSolutionStep(self):
+    #     """ This method is executed in order to initialize the current step
+    #
```
loumalouomega: Idem

loumalouomega (changes requested):
Some comments. test is missing

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
aronnoordam: thanks, I felt like there was a better way than I was doing

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
aronnoordam: I will look into that

applications/StructuralMechanicsApplication/python_scripts/set_moving_load_process.py
aronnoordam: My idea was to give a velocity table eventualy, such that every time step, the velocity can be different. The direction is meant to be either positive or negative (per axis). But it will always follow the line of the condition. An exact direction is thus dependent on the mesh.

applications/StructuralMechanicsApplication/python_scripts/set_moving_load_process.py
aronnoordam: Indeed, a starting point is to be added. I had in mind to set a starting distance relative the first condition element. In this way the load is always placed on a condition element.

applications/StructuralMechanicsApplication/python_scripts/set_moving_load_process.py
aronnoordam: but giving a function might be better indeed

aronnoordam: do you have an idea on how to give such a function as an input?

aronnoordam:
> Some comments. test is missing

Thanks

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
aronnoordam: actually, I can not use unordered_set, as my intention is to find the node id's which are not duplicated. I don't want to just find the unique id's

rubenzorrilla: OK. Now I see your point, you need the "init and end points" id's. Maybe you can store all the ids in a vector, sort them and use `std::unique`.

applications/StructuralMechanicsApplication/python_scripts/set_moving_load_process.py
rubenzorrilla: You can check what we do with the value field  in the `AssignVectorVariableProcess`. I'd use a similar mechanism.

applications/StructuralMechanicsApplication/python_scripts/set_moving_load_process.py
rubenzorrilla: OK. But if we do it in that way the distance is relative to one of the condition ends, which depends on the meshing. This is quite misleading for the user. Instead I'd ask for the origin point and check that such point lies in the condition (this OFC comes with the tolerance issues of the `IsInside` check).

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
rubenzorrilla: Just mentioning that we've never positioned about this. The musts are collected in the [style guide](https://github.com/KratosMultiphysics/Kratos/wiki/Style-Guide).

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
aronnoordam: If I have 10 line2D2 condition elements, I get 20 points. Using std::unique results in 11 unique points. What i want is to only find the very first and last node id, so 2 points. 

I also do a check if 2 points are found. If more than 2 points are found, it means that some conditions are not connected, thus i raise an error.

applications/StructuralMechanicsApplication/python_scripts/set_moving_load_process.py
aronnoordam: thanks, i will check it out

applications/StructuralMechanicsApplication/python_scripts/set_moving_load_process.py
aronnoordam: yes, maybe thats better indeed

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
rubenzorrilla: True. `std::unique` keeps the first of the repeated entries... Then nothing comes to my mind.

applications/StructuralMechanicsApplication/python_scripts/set_moving_load_process.py
aronnoordam: I was rethinking this comment. I think adding a function for the velocity is great, but in my opinion it should only be based on time, not on x,y,z. Because the moving load is only applied on line elements. In this case, it is still required to keep the "direction" variable, since it is important to identify which direction the load has to move in in the initial solution step. Further directions are not constant, since the load follows the line element. This is handled in the process, since i sort the condition element in the correct order

aronnoordam: an other thought: maybe in the future a moving load on a surface/volume can be added, where indeed a velocity vector can be set as a function of x,y,z and time. But for now, a moving load on a continuous line has my priority.

loumalouomega:
> @loumalouomega and @rubenzorrilla In each of my python tests i get the following error in the CI "RuntimeError: Error: bad argument",
> 
> Locally (on windows) my tests work (except 1 expected failure), do you have an idea what could be the problem and solution?

ERROR: test_SetMovingLoadWithVelocityFunction (test_set_moving_load_process.TestSetMovingLoadProcess)
  File "/__w/Kratos/Kratos/bin/Custom/applications/StructuralMechanicsApplication/tests/test_set_moving_load_process.py", line 685, in test_SetMovingLoadWithVelocityFunction
    self._TestSetMovingLoadWithVelocityFunction()
  File "/__w/Kratos/Kratos/bin/Custom/applications/StructuralMechanicsApplication/tests/test_set_moving_load_process.py", line 630, in _TestSetMovingLoadWithVelocityFunction
    cond.CalculateLocalSystem(lhs, rhs, mp.ProcessInfo)
RuntimeError: Error: bad argument
in /__w/Kratos/Kratos/applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp:223: virtual void MovingLoadCondition<2, 2>::CalculateAll(Condition::MatrixType &, Condition::VectorType &, const ProcessInfo &, const bool, const bool) [TDim = 2, TNumNodes = 2]

Looks like two booleans are missing (maybe Windows is deducing the arguments but not Linux)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
```diff
@@ -0,0 +1,220 @@
...
+	 * \param FirstCoord one of the end x,y or z coordinate of the condition model part
+	 * \param SecondCoord one of the end x,y or z coordinate of the condition model part
+	 * \param Direction direction of the moving load, positive or negative with respect to the x,y or z axis
+	 * \param EndConditions vector of the two end conditions within the model part
+	 * \return the first condition within the model part with respect to the direction of the moving load
+	 */
+    static Condition& GetFirstConditionFromCoord(const double FirstCoord, const double SecondCoord, const int Direction, std::vector<Condition>& EndConditions);
+
+    /**
+	 * \brief Checks if the points within a condition are to be swapped, based on the direction of the moving load
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
```diff
@@ -0,0 +1,220 @@
...
+	 * \param FirstCoord  x,y or z coordinate of the first point within the condition element
+	 * \param SecondCoord x,y or z coordinate of the second point within the condition element
+	 * \param Direction  direction of the moving load, positive or negative with respect to the x,y or z axis
+	 * \return bool which indicates if the condition's points are to be swapped
+	 */
+    static bool IsSwapPoints(const double FirstCoord, const double SecondCoord, const int Direction);
+
+    /**
+	 * \brief Initializes the global distance of the load based on the origin point within the sorted condition vector, starting from the first
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
```diff
@@ -0,0 +1,220 @@
...
+
+    // vector of load functions
+    std::vector<BasicGenericFunctionUtility> mLoadFunctions;
+
+    ///@}
+    ///@name Private Operations
+    ///@{
+
+    /**
+	 * \brief Finds indices of the nodes within the conditions model part, which do not share are condition within the model part.
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
```diff
@@ -0,0 +1,220 @@
...
+	 * \param IndicesVector vector of all the node indices, which are found in the condition elements.
+	 * \return vector of non repeating node indices within the model part
+	 */
+    static std::vector<IndexType> FindNonRepeatingIndices(const std::vector<IndexType> IndicesVector);
+
+    /**
+	 * \brief Finds condition elements which are at the spatial ends of the conditions vector. This function checks which condition's points are not repeated
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
```diff
@@ -0,0 +1,220 @@
...
+	 * in this conditions model part. If nodes are not repeated, it means that the condition is at one of the spatial ends.
+	 * \return Vector of the two end conditions.
+	 */
+    std::vector<Condition> FindEndConditions();
+
+    /**
+	 * \brief Sorts conditions vector. Sorts conditions in the direction of the moving load. The sorting starts from the first condition. Repeated nodes
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
```diff
@@ -0,0 +1,220 @@
...
+	 * ares searched in the conditions vector. If a node is repeated, it means the condition is connected to the previous condition. 
+	 * \param rUnsortedConditions container of the unsorted conditions within the model part
+	 * \param rFirstCondition first condition in the to be sorted conditions vector
+	 * \return vector of sorted conditions
+	 */
+    std::vector<Condition> SortConditions(ModelPart::ConditionsContainerType& rUnsortedConditions, Condition& rFirstCondition);
+
+    /**
+	 * \brief Check if conditions points are to be flipped in the direction of the moving load. The points are sorted based on the x-coordinates; if the x-coordinates are equal,
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
```diff
@@ -0,0 +1,220 @@
...
+	 * sorting is done based on the y-coordinates; if y-coordinates are equal, sorting is done based on the z-coordinates
+	 * \param rCondition reference of the current condition
+	 * \param Direction direction of the moving load
+	 * \return bool which indicates if the condition's points are to be flipped
+	 */
+    static bool IsConditionReversed(const Condition& rCondition, const array_1d<int, 3> Direction);
+
+    /**
+	 * \brief Finds the first condition within the model part, based on the direction of the moving load.
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
```diff
@@ -0,0 +1,220 @@
...
+	 * \param  FirstPoint, one of the end points of the condition model part
+	 * \param  SecondPoint, one of the end points of the condition model part
+	 * \param  Direction, vector of the direction of the moving load, positive values indicate positive direction with respect to the global axis
+	 * \param  rEndConditions vector of the two end conditions within the model part
+	 * \return the first condition within the model part with respect to the direction of the moving load
+	 */
+    static Condition& GetFirstCondition(const Point FirstPoint, const Point SecondPoint, const array_1d<int, 3> Direction, std::vector<Condition>& rEndConditions);
+
+    /**
+	 * \brief Finds the first condition within the model part, based on the direction of the moving load. In this function, a single direction is checked, either
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,485 @@
+// KRATOS  ___|  |                   |                   |
+//       \___ \  __|  __| |   |  __| __| |   |  __| _` | |
+//             | |   |    |   | (    |   |   | |   (   | |
+//       _____/ \__|_|   \__,_|\___|\__|\__,_|_|  \__,_|_| MECHANICS
+//
+//  License:		 BSD License
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,485 @@
...
+
+    // Vector with a loading applied to the condition
+    array_1d<double, TDim > moving_load = ZeroVector(TDim);
+    if( this->Has(POINT_LOAD) ){
+        noalias(moving_load) = this->GetValue(POINT_LOAD);
+    }
+
+    const double local_x_coord = this->GetValue(MOVING_LOAD_LOCAL_DISTANCE);
+
+	// check if cond should be calculated
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,485 @@
...
+    MatrixType& rLeftHandSideMatrix, VectorType& rRightHandSideVector,
+    const ProcessInfo& rCurrentProcessInfo,
+    const bool CalculateStiffnessMatrixFlag,
+    const bool CalculateResidualVectorFlag
+    )
+{
+    KRATOS_TRY
+
+    const SizeType number_of_nodes = GetGeometry().size();
+	const SizeType block_size = this->GetBlockSize();
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,485 @@
...
+
+        BoundedMatrix<double, TDim, TNumNodes> local_load_matrix = ZeroMatrix(TDim, TNumNodes);
+        BoundedMatrix<double, TDim, TNumNodes> global_load_matrix = ZeroMatrix(TDim, TNumNodes);
+
+        Matrix global_moment_matrix;
+
+        global_moment_matrix.resize(block_size - TDim, TNumNodes, false);
+
+
+        for (IndexType nod = 0; nod < TNumNodes; ++nod){
```
loumalouomega: i_nod

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,485 @@
...
+            global_moment_matrix(1, 0) = rotational_shape_functions_vector[0] * local_moving_load[2];
+            global_moment_matrix(2, 0) = rotational_shape_functions_vector[0] * local_moving_load[1];
+
+            global_moment_matrix(0, 1) = 0;
+            global_moment_matrix(1, 1) = rotational_shape_functions_vector[1] * local_moving_load[2];
+            global_moment_matrix(2, 1) = rotational_shape_functions_vector[1] * local_moving_load[1];
+
+        }
+        // rotation around z axis (2D)
+    	else if (block_size > TDim){
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,485 @@
...
+}
+
+
+template< std::size_t TDim, std::size_t TNumNodes >
+void MovingLoadCondition< TDim, TNumNodes>::CalculateExactNormalShapeFunctions(VectorType& rShapeFunctionsVector, const double LocalXCoord) const
+{
+    if (rShapeFunctionsVector.size() != TNumNodes) {
+        rShapeFunctionsVector.resize(TNumNodes, false);
+    }
+    auto& rGeom = this->GetGeometry();
```
loumalouomega: r_geometry

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,485 @@
...
+}
+
+
+template< std::size_t TDim, std::size_t TNumNodes >
+void MovingLoadCondition< TDim, TNumNodes>::CalculateExactShearShapeFunctions(VectorType& rShapeFunctionsVector, const double LocalXCoord) const
+{
+    if (rShapeFunctionsVector.size() != TNumNodes) {
+        rShapeFunctionsVector.resize(TNumNodes, false);
+    }
+    auto& rGeom = this->GetGeometry();
```
loumalouomega: r_geometry

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,485 @@
...
+}
+
+
+template< std::size_t TDim, std::size_t TNumNodes >
+void MovingLoadCondition< TDim, TNumNodes>::CalculateExactRotationalShapeFunctions(VectorType& rShapeFunctionsVector, const double LocalXCoord) const
+{
+    if (rShapeFunctionsVector.size() != TNumNodes) {
+        rShapeFunctionsVector.resize(TNumNodes, false);
+    }
+    auto& rGeom = this->GetGeometry();
```
loumalouomega: r_geometry

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,485 @@
...
+    const double length = rGeom.Length();
+
+    rShapeFunctionsVector[0] = LocalXCoord + (pow(LocalXCoord, 3) / pow(length, 2)) - 2 * (pow(LocalXCoord, 2) / length);
```
loumalouomega:
```suggestion
    rShapeFunctionsVector[0] = LocalXCoord + (std::pow(LocalXCoord, 3) / std::pow(length, 2)) - 2 * (std::pow(LocalXCoord, 2) / length);
```

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,485 @@
...
+    rShapeFunctionsVector[1] = (pow(LocalXCoord, 3) / pow(length, 2)) - (pow(LocalXCoord, 2) / length);
```
loumalouomega:
```suggestion
    rShapeFunctionsVector[1] = (std::pow(LocalXCoord, 3) / std::pow(length, 2)) - (std::pow(LocalXCoord, 2) / length);
```

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,485 @@
...
+//************************************************************************************
+//************************************************************************************
+
+
+/// Calculates rotation matrix for line2d2 element
+template<>
+void MovingLoadCondition<2, 2>::CalculateRotationMatrix(BoundedMatrix<double, 2, 2>& rRotationMatrix, const GeometryType& rGeom)
+{
+    constexpr double tolerance = 1e-8;
+	//Unitary vector in local x direction
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.h
```diff
@@ -0,0 +1,350 @@
+// KRATOS  ___|  |                   |                   |
+//       \___ \  __|  __| |   |  __| __| |   |  __| _` | |
+//             | |   |    |   | (    |   |   | |   (   | |
+//       _____/ \__|_|   \__,_|\___|\__|\__,_|_|  \__,_|_| MECHANICS
+//
+//  License:		 BSD License
+//					 license: structural_mechanics_application/license.txt
```
loumalouomega: Correct spaces

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.h
```diff
@@ -0,0 +1,350 @@
...
+//
+//  Main authors:    Aron Noordam
+//
+
+// System includes
+#if !defined(KRATOS_MOVING_LOAD_CONDITION_H_INCLUDED )
+#define  KRATOS_MOVING_LOAD_CONDITION_H_INCLUDED
```
loumalouomega:
```suggestion
#pragma once  

// System includes
```

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.h
```diff
@@ -0,0 +1,350 @@
...
+     * @return a Pointer to the new condition
+     */
+    Condition::Pointer Clone (
+        IndexType NewId,
+        NodesArrayType const& ThisNodes
+        ) const override;
+
+
+    /**
+	 * @brief Check if Rotational Dof existant
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.h
```diff
@@ -0,0 +1,350 @@
...
+    void CalculateAll(
+        MatrixType& rLeftHandSideMatrix,
+        VectorType& rRightHandSideVector,
+        const ProcessInfo& rCurrentProcessInfo,
+        const bool CalculateStiffnessMatrixFlag,
+        const bool CalculateResidualVectorFlag
+        ) override;
+
+    /**
+	 * \brief Calculates exact shape functions for a local load in normal direction
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.h
```diff
@@ -0,0 +1,350 @@
...
+	 * \param rShapeFunctionsVector  vector of exact shape functions
+	 * \param LocalXCoord local x coordinate within condition element
+	 */
+    void CalculateExactNormalShapeFunctions(VectorType& rShapeFunctionsVector, const double LocalXCoord) const;
+
+    /**
+	 * \brief Calculates exact shape functions for a local load in perpendicular direction
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.h
```diff
@@ -0,0 +1,350 @@
...
+	 * \param rShapeFunctionsVector vector of exact shape functions
+	 * \param LocalXCoord local x coordinate within condition element
+	 */
+    void CalculateExactShearShapeFunctions(VectorType& rShapeFunctionsVector, const double LocalXCoord) const;
+
+    /**
+	 * \brief Calculates exact shape functions for a local moment around z-axis.
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.h
```diff
@@ -0,0 +1,350 @@
...
+	 * \param rShapeFunctionsVector vector of exact shape functions
+	 * \param LocalXCoord local x coordinate within condition element
+	 */
+    void CalculateExactRotationalShapeFunctions(VectorType& rShapeFunctionsVector, const double LocalXCoord) const;
+
+    /**
+	 * \brief Calculates rotation matrix
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,413 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:		 BSD License
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,413 @@
...
+                            sorted_conditions.push_back(r_cond);
+                            visited_indices.push_back(i);
+                        }
+                    } else {
+                        // sort nodes in condition, such that new node is connected to previous condition
+                        std::set<IndexType>::iterator prev_id;
+                        if (is_cond_reversed){
+                            prev_id = node_id_vector.begin();
+                        } else
+                        {
```
loumalouomega:
```suggestion
                        } else {
```

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,413 @@
...
+        const array_1d<double, 3> origin_point = mParameters["origin"].GetVector();
+
+        // if origin point is within the current condition, set the global distance of the load, else continue the loop
+        if (r_geom.IsInside(origin_point, local_point)){
+            const double local_to_global_distance = (local_point[0] + 1) / 2 * element_length;
+
+            if (mIsCondReversedVector[i]){
+                mCurrentDistance = global_distance + element_length - local_to_global_distance;
+            }
+            else{
```
loumalouomega:
```suggestion
            } else {
```

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,413 @@
...
+
+
+void SetMovingLoadProcess::ExecuteInitialize()
+{
+    KRATOS_TRY
+
+    // clear load functions vector
+    mLoadFunctions.clear();
+
+	// check if load input is a function or numeric and add load to member variable
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,413 @@
...
+
+    // get the two line condition elements at both sides of the model part
+    std::vector<Condition> end_conditions = FindEndConditions();
+
+    // find start condition 
+    const Point center_1 = end_conditions[0].GetGeometry().Center();
+    const Point center_2 = end_conditions[1].GetGeometry().Center();
+    Condition& r_first_cond = GetFirstCondition(center_1, center_2, direction, end_conditions);
+
+	// Initialise vector which indicates if nodes in condition are in direction of movement
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,413 @@
...
+	// retrieve load from load function if given
+    if (mUseLoadFunction){
+        // get current time
+        const double current_time = this->mrModelPart.GetProcessInfo().GetValue(TIME);
+
+        for (IndexType i =0; i< mLoadFunctions.size();++i){
+            load_vector[i] = mLoadFunctions[i].CallFunction(0, 0, 0, current_time, 0, 0, 0);
+        }
+    } else {
+	    load_vector = mParameters["load"].GetVector();
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,413 @@
...
+    KRATOS_CATCH("")
+}
+
+
+void SetMovingLoadProcess::ExecuteInitializeSolutionStep()
+{
+
+    array_1d<double, 3> load_vector;
+
+	// retrieve load from load function if given
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -0,0 +1,413 @@
...
+{
+    double load_velocity;
+    // retrieve load velocity from velocity function if given
+    if (mUseVelocityFunction){
+        // get current time
+        const double current_time = this->mrModelPart.GetProcessInfo().GetValue(TIME);
+
+        BasicGenericFunctionUtility velocity_function = BasicGenericFunctionUtility(mParameters["velocity"].GetString());
+
+		// update velocity value
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
```diff
@@ -0,0 +1,220 @@
...
+//                   Multi-Physics
+//
+//  License:		 BSD License
+//					 Kratos default license: kratos/license.txt
+//
+//  Main authors:    Aron Noordam
+//
+
+#if !defined(KRATOS_SET_MOVING_LOAD_PROCESS_H_INCLUDED )
+#define  KRATOS_SET_MOVING_LOAD_PROCESS_H_INCLUDED
```
loumalouomega:
```suggestion
#pragma once
```

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
```diff
@@ -0,0 +1,220 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:		 BSD License
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
```diff
@@ -0,0 +1,220 @@
...
+
+    SetMovingLoadProcess(ModelPart& rModelPart,
+                                  Parameters Parameters);
+
+    ///@}
+    ///@name Operations
+    ///@{
+
+	 /**
+	 * \brief  Initializes the set moving load process. Check if load functions and a velocity function are present in the parameters.
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
```diff
@@ -0,0 +1,220 @@
...
+	 * Sort vector of conditions, and find the start position of the moving load, within the conditions vector.
+	 */
+    void ExecuteInitialize() override;
+
+
+    /**
+	 * \brief Initialize solution step. Calculate the load based on the load functions if present, else retrieve the load from the input parameters.
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
```diff
@@ -0,0 +1,220 @@
...
+	 * Loop over the conditions and find, on which condition the load is located. Then set the load on the condition element, if the load is located
+	 * within the element. If the moving load is not located on the condition element, set the load to zero.
+	 */
+    void ExecuteInitializeSolutionStep() override;
+
+    /**
+	 * \brief Finalizes solution step. Sets load velocity based on load velocity function if present, else load velocity is retrieved from the input values.
```
loumalouomega: Tabs (replace spaces)

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
```diff
@@ -0,0 +1,220 @@
...
+    void InitializeDistanceLoadInSortedVector();
+    ///@}
+
+}; // Class SetMovingLoadProcess
+
+///@}
+
+}  // namespace Kratos.
+
+#endif // KRATOS_SET_MOVING_LOAD_PROCESS_H_INCLUDED  defined
```
loumalouomega:
```suggestion
```

aronnoordam:
> > @loumalouomega and @rubenzorrilla In each of my python tests i get the following error in the CI "RuntimeError: Error: bad argument",
> > Locally (on windows) my tests work (except 1 expected failure), do you have an idea what could be the problem and solution?
> 
> ## ======================================================================
> ERROR: test_SetMovingLoadWithVelocityFunction (test_set_moving_load_process.TestSetMovingLoadProcess)
> Traceback (most recent call last): File "/__w/Kratos/Kratos/bin/Custom/applications/StructuralMechanicsApplication/tests/test_set_moving_load_process.py", line 685, in test_SetMovingLoadWithVelocityFunction self._TestSetMovingLoadWithVelocityFunction() File "/__w/Kratos/Kratos/bin/Custom/applications/StructuralMechanicsApplication/tests/test_set_moving_load_process.py", line 630, in _TestSetMovingLoadWithVelocityFunction cond.CalculateLocalSystem(lhs, rhs, mp.ProcessInfo) RuntimeError: Error: bad argument in /__w/Kratos/Kratos/applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp:223: virtual void MovingLoadCondition<2, 2>::CalculateAll(Condition::MatrixType &, Condition::VectorType &, const ProcessInfo &, const bool, const bool) [TDim = 2, TNumNodes = 2]
> 
> Looks like two booleans are missing (maybe Windows is deducing the arguments but not Linux)

"CalculateAll" is a function called by the function "CalculateLocalSystem" in the base "base_load_condtion.cpp", The CalculateLocalSystem function fills in these booleans. So i dont think this is the problem

loumalouomega (changes requested):
Some comments

loumalouomega:
Looks like doesn't work on Windows any more. I don't understand why doesn't work. It does the same as other tests

loumalouomega:
.

> Looks like doesn't work on Windows any more. I don't understand why doesn't work. It does the same as other tests

Can  you try to compile from scratch?, sometimes Windows doesn't take into account code changes when compiling

aronnoordam:
hmm strange. After recompiling from scratch, it still works locally, but not on the CI. However locally I've used python 3.8 and in the CI I see 3.10. Maybe this is the error, however this doesnt explain why the other tests do work.

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,570 @@
+// KRATOS  ___|  |                   |                   |
+//       \___ \  __|  __| |   |  __| __| |   |  __| _` | |
+//             | |   |    |   | (    |   |   | |   (   | |
+//       _____/ \__|_|   \__,_|\___|\__|\__,_|_|  \__,_|_| MECHANICS
+//
+//  License:         BSD License
+//	                 license: structural_mechanics_application/license.txt
```
loumalouomega:
```suggestion
//  License:         BSD License
//                        license: structural_mechanics_application/license.txt
```

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,570 @@
...
+                const IndexType index = base + TDim + k;
+                rRightHandSideVector[index] = global_moment_matrix(k, ii);
+            }
+        }
+    }
+    KRATOS_CATCH( "" )
+}
+
+
+template< >
```
loumalouomega: You can reduce this explicit instantations using c++17 if constexpr

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,570 @@
...
+//************************************************************************************
+//************************************************************************************
+
+
+/// Calculates rotation matrix for line2d2 element
+template<>
+void MovingLoadCondition<2, 2>::CalculateRotationMatrix(BoundedMatrix<double, 2, 2>& rRotationMatrix, const GeometryType& rGeom)
+{
+    KRATOS_TRY
+    constexpr double tolerance = 1e-8;
```
loumalouomega, line 296: Idem, try to use if constexpr to reduce code duplication

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,570 @@
...
+  
+    //Unitary vector in local x direction
+    array_1d<double, 3> vx;
+    noalias(vx) = rGeom.GetPoint(1) - rGeom.GetPoint(0);
+    const double inv_norm_x = 1.0 / norm_2(vx);
+    vx[0] *= inv_norm_x;
+    vx[1] *= inv_norm_x;
+    vx[2] *= inv_norm_x;
+
+    //initialise vectors in local y and z direction
```
loumalouomega: The same as the previous comment

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,570 @@
...
+    noalias(vx) = rGeom.GetPoint(1) - rGeom.GetPoint(0);
+    const double inv_norm_x = 1.0 / norm_2(vx);
+    vx[0] *= inv_norm_x;
+    vx[1] *= inv_norm_x;
+
+    rRotationMatrix(0, 0) = vx[0];
+    rRotationMatrix(0, 1) = vx[1];
+
+    array_1d<double, 3> vy = ZeroVector(3);
+    array_1d<double, 3> vy_tmp = ZeroVector(3);
```
loumalouomega: Idem

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,570 @@
...
+    noalias(vx) = rGeom.GetPoint(1) - rGeom.GetPoint(0);
+    const double inv_norm_x = 1.0 / norm_2(vx);
+    vx[0] *= inv_norm_x;
+    vx[1] *= inv_norm_x;
+    vx[2] *= inv_norm_x;
+
+    //initialise vectors in local y and z direction
+    array_1d<double, 3> vy = ZeroVector(3);
+    array_1d<double, 3> vz = ZeroVector(3);
+
```
loumalouomega: Idem

loumalouomega (changes requested):
Comments concerning code duplication

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,570 @@
...
+}
+
+template< std::size_t TDim, std::size_t TNumNodes >
+void MovingLoadCondition< TDim, TNumNodes>::CalculateExactNormalShapeFunctions(VectorType& rShapeFunctionsVector, const double LocalXCoord) const
+{
+    KRATOS_TRY
+    if (rShapeFunctionsVector.size() != TNumNodes) {
+        rShapeFunctionsVector.resize(TNumNodes, false);
+    }
+    auto& r_geometry = this->GetGeometry();
```
AlejandroCornejo: const?

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,570 @@
...
+
+
+template< std::size_t TDim, std::size_t TNumNodes >
+void MovingLoadCondition< TDim, TNumNodes>::CalculateExactShearShapeFunctions(VectorType& rShapeFunctionsVector, const double LocalXCoord) const
+{
+    KRATOS_TRY
+    if (rShapeFunctionsVector.size() != TNumNodes) {
+        rShapeFunctionsVector.resize(TNumNodes, false);
+    }
+    auto& r_geometry = this->GetGeometry();
```
AlejandroCornejo: const

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,570 @@
...
+    rRotationMatrix(2, 2) = vz[2];
+}
+
+
+/// Calculates rotation matrix for line2d3 element
+template<>
+void MovingLoadCondition<2, 3>::CalculateRotationMatrix(BoundedMatrix<double, 2, 2>& rRotationMatrix, const GeometryType& rGeom)
+{
+    KRATOS_TRY
+    constexpr double tolerance = 1e-8;
```
AlejandroCornejo: this should be defined in the .h file I think

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
```diff
@@ -0,0 +1,216 @@
...
+    double mCurrentDistance;
+
+    // bool which indicates if a load function is used to determine the value of the load
+    bool mUseLoadFunction;
+
+    // bool which indicates of velocity function is used to determine the value of the load velocity
+    bool mUseVelocityFunction;
+
+    // vector of load functions
+    std::vector<BasicGenericFunctionUtility> mLoadFunctions;
```
AlejandroCornejo, line 137: I think you need to implement the `Serializer` in case that you want to use it together with a restart

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
aronnoordam: ah nice, I didn't know

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
aronnoordam: Do you mean that i have to save and load the "mCurrentDistance",(the parameter which changes each time step)  or also the other parameters?

loumalouomega (dismissed):

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
aronnoordam: Hi @AlejandroCornejo, I've added the serializer. Do you have an idea when you have time to review that part?

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.cpp
```diff
@@ -407,4 +409,25 @@ void SetMovingLoadProcess::ExecuteFinalizeSolutionStep()
...
+    rSerializer.save("UseVelocityFunction", mUseVelocityFunction);
+    rSerializer.save("CurrentDistance", mCurrentDistance);
+
+}
+
+void SetMovingLoadProcess::load(Serializer& rSerializer)
+{
+    KRATOS_SERIALIZE_LOAD_BASE_CLASS(rSerializer, Process);
+    rSerializer.load("SortedConditions", mSortedConditions);
+    rSerializer.load("IsCondReversedVector", mIsCondReversedVector);
```
AlejandroCornejo, line 427: Indeed! now it's ok

applications/StructuralMechanicsApplication/custom_conditions/moving_load_condition.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+        noalias( rRightHandSideVector ) = ZeroVector(mat_size); //resetting RHS
+    }
+
+    // Vector with a loading applied to the condition
+    array_1d<double, TDim > moving_load = ZeroVector(TDim);
+    if( this->Has(POINT_LOAD) ){
+        array_1d<double, 3 > point_load = this->GetValue(POINT_LOAD);
+        for (IndexType i = 0; i < TDim; ++i)
+        {
```
AlejandroCornejo, line 123: this bracket should be in line 122 for unifying style

applications/StructuralMechanicsApplication/custom_processes/set_moving_load_process.h
philbucher: FYI currently we only serialize the `ModelPart` for restarts, everything else gets reconstructed

AlejandroCornejo (approved):
seems ok now!!

aronnoordam:
allright thanks for the reviews

PR #10152, opened 2022-08-17
[Fluid] Update fluid calculation utils to C++17

sunethwarna:
**📝 Description**
This extends the capabilities of `FluidCalculationUtilities` to compute gauss point values and gradients using historical and non historical values of geometries. This also adapts the old way to new fold expressions in c++17.

**🆕 Changelog**
- Added `EvaluateGradientInPoint` method
- Added `EvaluateNonHistoricalGradientInPoint` method
- Added `EvaluateNonHistoricalInPointMethod` method
- Added cpp tests
- Updated to c++ 17 fold expressions


rubenzorrilla:
This is the stuff we kept commented for years right?

sunethwarna:
@rubenzorrilla  yes, finally because of the lovely addition of `std::apply` with tuples in c++17, we can now simplify them and make them more human readable. :). I started changing, i will push them shortly.

applications/FluidDynamicsApplication/custom_utilities/fluid_calculation_utilities.cpp
```diff
@@ -65,6 +65,159 @@ void FluidCalculationUtilities::UpdateValue(
...
 
+template<>
+void FluidCalculationUtilities::AssignGradientValue(
+    const double& rInput,
+    const double rShapeFunctionDerivative,
+    const IndexType DirectionIndex,
+    array_1d<double, 3>& rOutput)
+{
+    KRATOS_TRY
+
```
roigcarlo:
```suggestion
    KRATOS_DEBUG_ERROR_IF(DirectionIndex > 2)
        << "Unsupported Direction Index. DirectionIndex should be  less than 3 "
           "[ DirectionIndex = "
        << DirectionIndex << " ].\n";
```

Shouldn't be a direction check here too?

applications/FluidDynamicsApplication/custom_utilities/fluid_calculation_utilities.cpp
```diff
@@ -65,6 +65,159 @@ void FluidCalculationUtilities::UpdateValue(
...
+        << "Unsupported Direction Index. DirectionIndex should be  less than 2 "
+           "[ DirectionIndex = "
+        << DirectionIndex << " ].\n";
+
+    rOutput(0, DirectionIndex) += rInput[0] * rShapeFunctionDerivative;
+    rOutput(1, DirectionIndex) += rInput[1] * rShapeFunctionDerivative;
+
+    KRATOS_CATCH("");
+}
+
```
roigcarlo: I wild shot (maybe not possible)

Couldn't a `boost::matrix_column<BoundedMatrix<double, x, y>>` be used here instead of `BoundedMatrix<double, x, y>`? 

I understand that you need to specialize all those combinations instead of making a template instantiation over the `TOutputDataType` because partial specialization for functions in not yet allowed and you need to assign the values differently if its a matrix or a vector, but using the proxy should allow you to access the `DirectionIndex` of the matrix your are interested in and just pass `rOutput[DirectionIndex]` as your output vector, so all functions would have essentially the same code, and you just need to explicitly instantiate at the end of the file.

applications/FluidDynamicsApplication/custom_utilities/fluid_calculation_utilities.cpp
sunethwarna: got rid of these template specializations, now it is more simpler with `std::apply`

applications/FluidDynamicsApplication/custom_utilities/fluid_calculation_utilities.cpp
sunethwarna: got rid of these template specializations, now it is more simpler with `std::apply`

sunethwarna:
@roigcarlo I updated the PR. I got rid of most of the template specializations and introduced `std::apply` which made this more simplified. Could you take a look?

roigcarlo (approved):

sunethwarna:
thanks @roigcarlo

PR #10154, opened 2022-08-18
Changing source of boost

roigcarlo:
**📝 Description**
Seems more stable than sourceforge, so, in the meantime...

**🆕 Changelog**
- Changed boost source for win.

### DIFF
--- a/.github/workflows/nightly_build.yml
+++ b/.github/workflows/nightly_build.yml
@@ -101,7 +101,7 @@ jobs:
 
     - name: Download boost
       run: |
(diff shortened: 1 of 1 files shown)


loumalouomega (commented):
In fact i did this for the non nightly build!

loumalouomega (approved):

loumalouomega:
> In fact i did this for the non nightly build!

https://github.com/KratosMultiphysics/Kratos/commit/d405427d3757b048b001fe8125ce88347c1d74e1

It was you who approved!, that's why I was confused about the boost because I thought I already fixed it!

roigcarlo:
Yep I did not realize the nightly was not changed. My bad. Hopefully it will work now

PR #10162, opened 2022-08-18
[CI] adding docker for Jammy (Ubuntu 22.04)

philbucher:
see title

this only adds the new image
I will update the CI in an upcoming PR

### DIFF
--- a/.github/workflows/build_docker_images_for_ci.yml
+++ b/.github/workflows/build_docker_images_for_ci.yml
@@ -20,6 +20,16 @@ on:
       - 'scripts/docker_files/docker_file_ci_centos_7/*'
       - '.github/workflows/build_docker_images_for_ci.yml'
 
+  workflow_dispatch:
+
+# for cancelling redundant runs
+concurrency:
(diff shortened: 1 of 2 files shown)


philbucher:
@marcnunezc @loumalouomega seems like MMG 5,4 does not work with the latest ubuntu
I remember that we discusses already a long time ago to drop support for it, now would be a good time, do you agree?

loumalouomega:
> @marcnunezc @loumalouomega seems like MMG 5,4 does not work with the latest ubuntu
> I remember that we discusses already a long time ago to drop support for it, now would be a good time, do you agree?

Okay

marcnunezc:
> @marcnunezc @loumalouomega seems like MMG 5,4 does not work with the latest ubuntu I remember that we discusses already a long time ago to drop support for it, now would be a good time, do you agree?

Yes, I agree as well.

roigcarlo (approved):
Do we want to use this in the CI?

philbucher:
> Do we want to use this in the CI?

yes :)

PR #10165, opened 2022-08-19
[MeshMoving] FM-ALE custom variables projection

rubenzorrilla:
**📝 Description**
In this PR I'm adding some usability options. Most important one is the possibility to use custom variables in the projection. Besides, I also added the possibility to change the search settings after @RiccardoRossi 's changes. Last but not least, I removed the old explicit FM-ALE as it was only adding an extra complexity with no benefit (we developed in the past but the results were not satisfactory enough, I note here that we've never used it again).

@miguelmaso this is one of the changes we need for what we're working on.
In a subsequent PR I'd introduce the option to calculate the mesh motion based on the origin mesh velocity to complement the structure-based we have right now (@miguelmaso @uxuech ).


philbucher:
I have some comments, will add them over the WE

applications/MeshMovingApplication/custom_utilities/fixed_mesh_ale_utilities.cpp
```diff
@@ -194,24 +194,20 @@ namespace Kratos
...
-        double search_tol = 1e-5;
-        block_for_each( rOriginModelPart.Nodes(), typename BinBasedFastPointLocator<TDim>::ResultContainerType(max_results),
-            [&]( Node<3>& rNode,
-                 typename BinBasedFastPointLocator<TDim>::ResultContainerType& search_results
-             )
-            {
+        block_for_each(
+            rOriginModelPart.Nodes(),
+            typename BinBasedFastPointLocator<TDim>::ResultContainerType(mSearchMaxResults),
+            [&](Node<3>& rNode, typename BinBasedFastPointLocator<TDim>::ResultContainerType& rSearchResults){
```
philbucher:
```suggestion
            [&](auto& rNode, auto& rSearchResults){
```

try this magic :star_struck:

rubenzorrilla: Cool! Thanks for the hint.

philbucher: C++14 btw

rubenzorrilla:
> I have some comments, will add them over the WE

@philbucher do you still have those comments in mind?

philbucher (approved):

PR #10166, opened 2022-08-19
[Core][MPI] Skin detection process in MPI

loumalouomega:
**📝 Description**

This PR fixes SkinDetectionProcess so it works in MPI

**🆕 Changelog**

- Fixes SkinDetectionProcess so it works in MPI
- Update test to be MPI compatible


kratos/processes/skin_detection_process.cpp
```diff
@@ -338,6 +353,135 @@ void SkinDetectionProcess<TDim>::SetUpAdditionalSubModelParts(const ModelPart& r
...
+            if (to_remove) {
+                faces_to_remove.push_back(r_vector_ids);
+            }            
+        }
+
+        /* Not all the faces are going to be removed, only the ones which are repeated in different processes. So we need to filter then. */
+        
+        // First we determine the rank and the size of the world
+        auto& r_communicator = mrModelPart.GetCommunicator();
+        auto& r_data_communicator = r_communicator.GetDataCommunicator();
```
philbucher:
```suggestion
        const auto& r_communicator = mrModelPart.GetCommunicator();
        const auto& r_data_communicator = r_communicator.GetDataCommunicator();
```

kratos/processes/skin_detection_process.cpp
```diff
@@ -215,8 +224,13 @@ void SkinDetectionProcess<TDim>::FillAuxiliaryModelPart(
...
     // Now we set the flag on the nodes. The list of nodes of the auxiliar model part
     auto& r_nodes_array = rAuxiliaryModelPart.Nodes();
-
     VariableUtils().SetFlag(INTERFACE, true, r_nodes_array);
+
+    // In case we are in MPI we syncronize the INTERFACE flag
+    if (mrModelPart.IsDistributed()) {
+        auto& r_communicator = mrModelPart.GetCommunicator();
+        r_communicator.SynchronizeOrNodalFlags(INTERFACE);
+    }
```
philbucher:
```suggestion
     mrModelPart.GetCommunicator().SynchronizeOrNodalFlags(INTERFACE);
```

if possible avoid this check as it is cluttering the code
it is designed to work in both cases

kratos/tests/test_skin_detection_process.py
```diff
@@ -122,10 +96,74 @@ def test_NotOnSubModelPartSkinDetectionProcess(self):
...
-        self.assertEqual(model_part.NumberOfConditions(), 112)
+        # Check the number of conditions created
+        data_comm = self.model_part.GetCommunicator().GetDataCommunicator()
+        if data_comm.IsDistributed():
+            number_of_conditions = 0
+            number_of_conditions += self.model_part.NumberOfConditions()
+            global_number_of_conditions = data_comm.SumAll(number_of_conditions)
+            self.assertEqual(global_number_of_conditions, 112)
+        else:
+            self.assertEqual(self.model_part.NumberOfConditions(), 112)
```
philbucher, line 89:
```suggestion
        # Check the number of conditions created
        self.assertEqual(model_part.GetCommunicator().GlobalNumberOfConditions(), 112)
```

kratos/tests/test_skin_detection_process.py
```diff
@@ -122,10 +96,74 @@ def test_NotOnSubModelPartSkinDetectionProcess(self):
...
+                                        }
+                                        """)
+                                    )
+
+        gid_output.ExecuteInitialize()
+        gid_output.ExecuteBeforeSolutionLoop()
+        gid_output.ExecuteInitializeSolutionStep()
+        gid_output.PrintOutput()
+        gid_output.ExecuteFinalizeSolutionStep()
+        gid_output.ExecuteFinalize()
```
philbucher: why not using the respective python processes to avoid code duplication

or even better, remove the post stuff altogether?

loumalouomega: But but is useful...

loumalouomega:
I don't know why the compiler fails. Any idea @roigcarlo ?

roigcarlo:
No idea, let me check back in the office :S

kratos/tests/test_skin_detection_process.py
loumalouomega: Doesn't work, I don't know why, maybe I need to use DistributedModelPartInitializer first, not very practical

kratos/tests/test_skin_detection_process.py
```diff
@@ -79,7 +79,14 @@ def test_NotOnSubModelPartSkinDetectionProcess(self):
         detect_skin.Execute()
 
         # Check the number of conditions created
-        self.assertEqual(self.model_part.GetCommunicator().GlobalNumberOfConditions(), 112)
```
rubenzorrilla: Why is this not working? Internally it does exactly the same operation you're doing.

loumalouomega: I don't know. It says it is zero. I don't know if I need to do something else to update the database

philbucher: If it is not working then the communicator is also bot properly initialized
Might be related to other problems

loumalouomega: AS you see, I adapted the test to the same format used in other dual serial/MPI tests

loumalouomega: I tried with:

~~~sh
# Check the number of conditions created
        if self.model_part.IsDistributed():
            import KratosMultiphysics.mpi as KratosMPI
            KratosMPI.ParallelFillCommunicator(self.model_part, self.model_part.GetCommunicator().GetDataCommunicator()).Execute()
            self.assertEqual(self.model_part.GetCommunicator().GlobalNumberOfConditions(), 112)
        else:
            self.assertEqual(self.model_part.NumberOfConditions(), 112)
~~~

but it fails with more than 3 cores

loumalouomega: If we avoid this, the PR is ready to merge (compilation fixed)

roigcarlo: Most likely the comm is the placeholder one for the serial cases, so it just does not have any mLocalMesh initialized.

loumalouomega: Yes, I tried to define ParallelFillCommunicator in the process, but the compilation fails because is not in the /mpi folder. And moving the process to MPI or separate it between MPI and non MPI is something I want to avoid

loumalouomega: Are we happy like this?

roigcarlo: My 2 cents, for me its ok to leave it like this. But we should start considering if its worth to pay the overhead of initializing a communicator with 1 partition for serial cases as well. Some tests could benefit from this and it should be much more intuitive to use (As a user, I would expect that function to work without prior setup)

philbucher: I checked, the communication plan is properly initialized, hence I am not sure why the fct fails

I would investigate it deeper

loumalouomega:
Ping @KratosMultiphysics/mpi

loumalouomega:
> Ping @KratosMultiphysics/mpi

I would like to join @KratosMultiphysics/mpi

kratos/processes/skin_detection_process.cpp
```diff
@@ -338,6 +350,135 @@ void SkinDetectionProcess<TDim>::SetUpAdditionalSubModelParts(const ModelPart& r
 /***********************************************************************************/
 /***********************************************************************************/
 
+template<SizeType TDim>
+void SkinDetectionProcess<TDim>::GenerateSetNodeIdsInterface(std::unordered_set<IndexType>& rSetNodeIdsInterface)
+{
+    // Only in case we are executing in MPI
+    if (mrModelPart.IsDistributed()) {
```
roigcarlo, line 357: In general, if the function is only used in MPI, I would just put the `if` outside and leave the body with the code.

roigcarlo (commented):
I am not 100% sure about what this process does, but after reading a bit, I understand that the problem is that some faces are generated in multiple processes.

Instead of removing the faces once created, couldn't you use a criteria to select if the face needs to be created or not based on the process is creating it?

For example:

```C++
for (IndexType i_face = 0; i_face < potential_number_neighbours; ++i_face) {

    /* FACES/EDGES */
   const SizeType number_nodes = r_boundary_geometries[i_face].size();
   VectorIndexType vector_ids(number_nodes);
   VectorIndexType ordered_vector_ids(number_nodes);

   /* FACE/EDGE */
   for (IndexType i_node = 0; i_node < number_nodes; ++i_node) {
       vector_ids[i_node] = r_boundary_geometries[i_face][i_node].Id();
       ordered_vector_ids[i_node] = vector_ids[i_node];
    }

    /*** THE ARRAY OF IDS MUST BE ORDERED!!! ***/
    std::sort(vector_ids.begin(), vector_ids.end());
    // Check if the elements already exist in the HashMapVectorIntType
    HashMapVectorIntTypeIteratorType it_check = rInverseFaceMap.find(vector_ids);

    if(it_check == rInverseFaceMap.end() ) {
        // If it doesn't exist it is added to the database
        rInverseFaceMap.insert(std::pair<VectorIndexType, VectorIndexType>(vector_ids, ordered_vector_ids));
        rPropertiesFaceMap.insert(std::pair<VectorIndexType, IndexType>(vector_ids, (it_elem->pGetProperties())->Id()));
        }
    }
}
```
Here you are building the db of the new faces. We could exploit the fact that the faces have hashes and those are ordered to exclude them form being added t the db

Pseudocode:
```c++
std::unoderder_set<int, node::Pointer> inverse_node_map;

/* FACE/EDGE */
for (IndexType i_node = 0; i_node < number_nodes; ++i_node) {
    vector_ids[i_node] = r_boundary_geometries[i_face][i_node].Id();
    ordered_vector_ids[i_node] = vector_ids[i_node];
    inverse_node_map.insert(std::pair<int, node::Pointer>( r_boundary_geometries[i_face][i_node].Id(),  r_boundary_geometries[i_face][i_node]))
}

....

HashMapVectorIntTypeIteratorType it_check = rInverseFaceMap.find(vector_ids);
bool process_check = true;

// I am not sure if the condition is so simplistic, but if there is an element capable of generating faces in one partition, we should be able the filter the id's of such faces so the "master" process that would physically owns the face is identified.
if (mrModelPart->GetCommunicator()->IsDistributed()) {
    process_check = (inverse_node_map[vector_ids[0]]->GetSolutionStepValue(PARTITION_INDEX) == my_rank);
}

if(it_check == rInverseFaceMap.end() && process_check ) {

```

Then, the filter function for MPI is not necessary, as you can ensure there are no repeated faces through the domains.

kratos/processes/skin_detection_process.cpp
loumalouomega: Okay

loumalouomega:
> I am not 100% sure about what this process does, but after reading a bit, I understand that the problem is that some faces are generated in multiple processes.
> 
> Instead of removing the faces once created, couldn't you use a criteria to select if the face needs to be created or not based on the process is creating it?
> 
> For example:
> 
> ```c++
> for (IndexType i_face = 0; i_face < potential_number_neighbours; ++i_face) {
> 
>     /* FACES/EDGES */
>    const SizeType number_nodes = r_boundary_geometries[i_face].size();
>    VectorIndexType vector_ids(number_nodes);
>    VectorIndexType ordered_vector_ids(number_nodes);
> 
>    /* FACE/EDGE */
>    for (IndexType i_node = 0; i_node < number_nodes; ++i_node) {
>        vector_ids[i_node] = r_boundary_geometries[i_face][i_node].Id();
>        ordered_vector_ids[i_node] = vector_ids[i_node];
>     }
> 
>     /*** THE ARRAY OF IDS MUST BE ORDERED!!! ***/
>     std::sort(vector_ids.begin(), vector_ids.end());
>     // Check if the elements already exist in the HashMapVectorIntType
>     HashMapVectorIntTypeIteratorType it_check = rInverseFaceMap.find(vector_ids);
> 
>     if(it_check == rInverseFaceMap.end() ) {
>         // If it doesn't exist it is added to the database
>         rInverseFaceMap.insert(std::pair<VectorIndexType, VectorIndexType>(vector_ids, ordered_vector_ids));
>         rPropertiesFaceMap.insert(std::pair<VectorIndexType, IndexType>(vector_ids, (it_elem->pGetProperties())->Id()));
>         }
>     }
> }
> ```
> 
> Here you are building the db of the new faces. We could exploit the fact that the faces have hashes and those are ordered to exclude them form being added t the db
> 
> Pseudocode:
> 
> ```c++
> std::unoderder_set<int, node::Pointer> inverse_node_map;
> 
> /* FACE/EDGE */
> for (IndexType i_node = 0; i_node < number_nodes; ++i_node) {
>     vector_ids[i_node] = r_boundary_geometries[i_face][i_node].Id();
>     ordered_vector_ids[i_node] = vector_ids[i_node];
>     inverse_node_map.insert(std::pair<int, node::Pointer>( r_boundary_geometries[i_face][i_node].Id(),  r_boundary_geometries[i_face][i_node]))
> }
> 
> ....
> 
> HashMapVectorIntTypeIteratorType it_check = rInverseFaceMap.find(vector_ids);
> bool process_check = true;
> 
> // I am not sure if the condition is so simplistic, but if there is an element capable of generating faces in one partition, we should be able the filter the id's of such faces so the "master" process that would physically owns the face is identified.
> if (mrModelPart->GetCommunicator()->IsDistributed()) {
>     process_check = (inverse_node_map[vector_ids[0]]->GetSolutionStepValue(PARTITION_INDEX) == my_rank);
> }
> 
> if(it_check == rInverseFaceMap.end() && process_check ) {
> ```
> 
> Then, the filter function for MPI is not necessary, as you can ensure there are no repeated faces through the domains.

In fact I tried that in the first iteration, but it didn't work. I will give a second chance

roigcarlo:
If you tried already, maybe it can't be done, but it really smells to me that it should be possible

loumalouomega:
> If you tried already, maybe it can't be done, but it really smells to me that it should be possible

Nope, it didn't work, sorry. Can we merge this?

loumalouomega:
Hello @roigcarlo

roigcarlo (approved):

roigcarlo:
I forgot this existed sorry

PR #10167, opened 2022-08-19
[Core][Serializer] using data

philbucher:
follow up of #9572, since we now have C++17 we can use `data` which makes more sense than `front`


loumalouomega:
> follow up of #9572, since we now have C++17 we can use `data` which makes more sense than `front`

I just remembered that now we can simplify pybind11 code duplication

roigcarlo (approved):

PR #10168, opened 2022-08-19
[Do not merge] Simplifying core interfaces

roigcarlo:
**📝 Description**
This for discussion. Since now we can make use of variadric template iterators through ..., we can simplify all our python interfaces

This is just a simple example. Also notice that we have implemented a lot of missing entry points for int, matrices, vectros, etc... which were not defined (inconsistenly) for many types while the functionality was there.

In an ideal world, the iterators can be reused through multiple interfaces. Here I show the usage for variable types, but it can be combined with container types, or made only for containers etc... you see the idea.


loumalouomega (approved):
Okay

roigcarlo:
I honestly doubt it passes the ci in this state, I just wanted to discuss the idea :P

roigcarlo:
Mm that was better than expected :O

RiccardoRossi:
pybind has a overload m3chanism for accomplishing this.

can u take q look at

https://pybind11.readthedocs.io/en/stable/classes.html

roigcarlo:
@RiccardoRossi As I understand, pybind11 interfaces allows to implement overloads, with some suggar if you use C++14.

so if you have a class with:
```c++
class MyClass {
    void foo(int a) {}
    void foo(float a) {}
    void foo(double a) {}
}
```

You can use:

```c++
py::class_<MyClass >(m, "MyClass ")
    .def("foo", py::overload_cast<int>(&MyClass::foo), "Foo for ints")
    .def("foo", py::overload_cast<float>(&MyClass::foo), "Foo for floats");
```

This way you have control over which overloadeds are exposed, in this exampel `double` is not.

What I do is create a common interface for different classes, as they are instanced with different types:

```c++
class MyClass {
    template<class TDataType>
    void foo(TDataType a) {}
}
```

Which needs to be exported as (follwing the example):

```c++
py::class_<MyClass>(m, "MyClass ")
    .def("foo", (&MyClass::foo<int>), "Foo for int function")
    .def("foo", (&MyClass::foo<float>), "Foo for float function");
````

If more than one function is needed for every type of the class template you end up with something like this:

```c++
py::class_<MyClass>(m, "MyClass ")
    .def("foo1", (&MyClass::foo1<int>), "Foo1 for int function")
    .def("foo2", (&MyClass::foo2<int>), "Foo2 for int function")
    .def("foo3", (&MyClass::foo3<int>), "Foo3 for int function")
    .def("foo4", (&MyClass::foo4<int>), "Foo4 for int function")
    .def("foo5", (&MyClass::foo5<int>), "Foo5 for int function")
    .def("foo1", (&MyClass::foo1<float>), "Foo1 for float function")
    .def("foo2", (&MyClass::foo2<float>), "Foo2 for float function")
    .def("foo3", (&MyClass::foo3<float>), "Foo3 for float function")
    .def("foo4", (&MyClass::foo4<float>), "Foo4 for float function")
    .def("foo5", (&MyClass::foo5<float>), "Foo5 for float function");
````

This PR changes the later, such you can define a list of classes which will be added to the interface (not overloaded verions, but we could extend it as well for that...) so it becomes:

```c++
template<class TDataType>
void AddInterface(pythonclass pc) {
    pc.def("foo1", (&MyClass::foo1<TDataType>), "Foo1 for TDataType function")
    pc.def("foo2", (&MyClass::foo2<TDataType>), "Foo2 for TDataType function")
    pc.def("foo3", (&MyClass::foo3<TDataType>), "Foo3 for TDataType function")
    pc.def("foo4", (&MyClass::foo4<TDataType>), "Foo4 for TDataType function")
    pc.def("foo5", (&MyClass::foo5<TDataType>), "Foo5 for TDataType function")
}

auto pythonclass = py::class_<MyClass>(m, "MyClass ");

AddInterface<int>(pythonclass);
AddInterface<float>(pythonclass);
````

For example, this is what we already had with ``AddCopyModelPartFlaggedInterface` which was already part of the existing code.
Because C++17 allows it, I passed the types as a list instead of calling the function twice:

```c++
AddInterface<int, float, ....>(pythonclass);
````

RiccardoRossi:
i see.

thx for the explanation

roigcarlo:
Closing because it has error and I don't have time to review this.

I will retake it once I have time.

PR #10173, opened 2022-08-21
[Core][VariableUtils] Variadic set to zero methods

rubenzorrilla:
**📝 Description**
In this PR I'm adding two new `SetNonHistoricalVariablesToZero` and `SetHistoricalVariablesToZero` methods. Nice thing is that thanks to fold expressions feature from c++17 we can easily implement these with variadic templates, so now we can set more than one variable to zero when looping the nodes.

Bad thing is that the variables to be set to zero need to be known at compile time, so we (or at least I don't know how) cannot export it to Python.

In the future we can do something similar with the method to set values in the `VariableUtils` by passing variable-value pairs in a variadic manner.

Finally, I point that I decided to (finally) make these `static` as they will be only used in c++.

In subsequent PRs we can start populating these.

### DIFF
--- a/kratos/tests/cpp_tests/utilities/test_variable_utils.cpp
+++ b/kratos/tests/cpp_tests/utilities/test_variable_utils.cpp
@@ -1,0 +1,154 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
+//
+//  Main authors:    Ruben Zorrilla
+//
+
+// System includes
+#include <limits>
+
+// External includes
+
+// Project includes
+#include "containers/model.h"
+#include "testing/testing.h"
+#include "utilities/variable_utils.h"
+
+namespace Kratos
+{
+namespace Testing
+{
+
+    KRATOS_TEST_CASE_IN_SUITE(VariableUtilsSetHistoricalVariablesToZero, KratosCoreFastSuite)
+    {
+        // Set auxilary nodal structure
+        Model test_model;
+        auto& r_test_model_part = test_model.CreateModelPart("TestModelPart");
+        r_test_model_part.AddNodalSolutionStepVariable(PRESSURE);
+        r_test_model_part.AddNodalSolutionStepVariable(VELOCITY);
+        r_test_model_part.AddNodalSolutionStepVariable(TEMPERATURE);
+        r_test_model_part.AddNodalSolutionStepVariable(DISPLACEMENT);
+        r_test_model_part.AddNodalSolutionStepVariable(DEFORMATION_GRADIENT);
+        for (std::size_t i = 0; i < 10; ++i) {
+            r_test_model_part.CreateNewNode(i,0.0,0.0,0.0);
+        }
+
+        // Set fake values in the historical database
+        array_1d<double,3> aux_vect;
+        Matrix aux_mat = ZeroMatrix(3,3);
+        for (std::size_t i = 0; i < 3; ++i) {
+            aux_vect[i] = 1.0;
+            for (std::size_t j = 0; j < 3; ++j) {
+                aux_mat(i,j) = 1.0;
+            }
+        }
+        for (auto& r_node : r_test_model_part.Nodes()) {
+            r_node.FastGetSolutionStepValue(PRESSURE) = 1.0;
+            r_node.FastGetSolutionStepValue(TEMPERATURE) = 1.0;
+            noalias(r_node.FastGetSolutionStepValue(VELOCITY)) = aux_vect;
+            noalias(r_node.FastGetSolutionStepValue(DISPLACEMENT)) = aux_vect;
+            r_node.FastGetSolutionStepValue(DEFORMATION_GRADIENT) = aux_mat;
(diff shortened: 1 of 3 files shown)


kratos/tests/cpp_tests/utilities/test_variable_utils.cpp
```diff
@@ -0,0 +1,154 @@
...
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
+//
+//  Main authors:    Ruben Zorrilla
+//
+
```
loumalouomega, line 12: Strange this wasn't tested

loumalouomega (dismissed):
Nice

loumalouomega:
Ci compilation error: /__w/Kratos/Kratos/kratos/utilities/variable_utils.h:755:64: error: dependent-name 'TVariableArgs::Type' is parsed as a non-type, but instantiation yields a type

kratos/tests/cpp_tests/utilities/test_variable_utils.cpp
rubenzorrilla: All tests are in Python. I think this one has a pretty good coverage.

rubenzorrilla:
> Ci compilation error: /__w/Kratos/Kratos/kratos/utilities/variable_utils.h:755:64: error: dependent-name 'TVariableArgs::Type' is parsed as a non-type, but instantiation yields a type

I see... For first time in life I did it in Windows and this didn't pop up :smile:

kratos/utilities/variable_utils.cpp
```diff
@@ -323,6 +323,24 @@ KRATOS_API(KRATOS_CORE) const ModelPart::ConditionsContainerType& VariableUtils:
...
+    rNode.FastGetSolutionStepValue(rVariable) = rValue;
+}
+
+template<>
+void VariableUtils::AuxiliaryHistoricalValueSetter(
+    const Variable<array_1d<double,3>>& rVariable,
+    const array_1d<double,3>& rValue,
+    NodeType& rNode)
+{
+    noalias(rNode.FastGetSolutionStepValue(rVariable)) = rValue;
```
matekelemen, line 341: Does ```noalias``` make a difference in a simple assignment?

roigcarlo: Agree, shouldn't the `noalias` be in the above function?

rubenzorrilla: > Does `noalias` make a difference in a simple assignment?

It does.

rubenzorrilla: > Agree, shouldn't the `noalias` be in the above function?

Nope, as it doesn't work with all variable types. This is why I needed this auxiliary function.

roigcarlo: I don't understand, afaik `noalias` is used to prevent operations like `A = A + B`, from being executed as `temp = A + B, A = temp` (and 2nd+ level pointer indirections but we don't have variables like those don't we?).

Setting a `rValue` is already that operation. The only scenario in which this could fail, is if you do something like this:

```c++
// This would forcefully modify MATRIX_VAR while calculating it making it yield and incorrect value during the op
AuxiliaryHistoricalValueSetter(MATRIX_VAR, myNode.FastGetSolutionStepValue(MATRIX_VAR) * myNode.FastGetSolutionStepValue(MATRIX_VAR_2), myNode)
```

Am I missing something? Which is the case or the type of the var that is making no-alias behave differently?

kratos/utilities/variable_utils.h
```diff
@@ -721,6 +739,23 @@ class KRATOS_API(KRATOS_CORE) VariableUtils
...
+     * @param rNodes Nodes container to be initialized to zero
+     * @param rVariableArgs Variables to be set to zero
+     */
+    template<class... TVariableArgs>
+    static void SetHistoricalVariablesToZero(
+        NodesContainerType& rNodes,
+        const TVariableArgs&... rVariableArgs)
+    {
+        block_for_each(rNodes, [&](NodeType& rNode){(
+            AuxiliaryHistoricalValueSetter<typename TVariableArgs::Type>(rVariableArgs, rVariableArgs.Zero(), rNode), ...);
```
sunethwarna, line 755:
```suggestion
            (rNode.FastGetSolutionStepValue(rVariableArgs) = rVariableArgs.Zero(), ...);
```

Wouldn't this work without any problems since `rNode.FastGetSolutionStepValue(rVariableArgs)` anyways returns the reference? Do we still need the `noalias`?

rubenzorrilla: I did the auxiliary method to enable using `noalias` with the `array_1d`. This would work as well but wouldn't be as efficient.

kratos/utilities/variable_utils.cpp
rubenzorrilla: More than behaving differently, the  `noalias` doesn't work with `double` type variables. This is why we need the auxiliary function. I'm not aware about the technicalities behind, for this we'd better ask @RiccardoRossi .

rubenzorrilla:
@loumalouomega the Linux CI is now fixed.

loumalouomega (approved):

loumalouomega:
> @loumalouomega the Linux CI is now fixed.

BTW, I never blocked the merge

PR #10176, opened 2022-08-22
[StructuralMechanicsApplication] Adding patch test for axisymmetry

loumalouomega:
**📝 Description**

Adding patch test for axisymmetry

**🆕 Changelog**

- Adding patch test for axisymmetry

@AlejandroCornejo the test fails, by a lot, any idea why?


AlejandroCornejo:
Can it be because the geometry has non-positive `x` components?

AlejandroCornejo:
actually not, I've checked

loumalouomega:
> Can it be because the geometry has non-positive `x` components?

Huuuuuuuuum, let me see

> actually not, I've checked

Faster than me

applications/StructuralMechanicsApplication/tests/test_patch_test_small_strain.py
```diff
@@ -136,10 +138,10 @@ def _check_results(self,mp,A,b):
             d = node.GetSolutionStepValue(KratosMultiphysics.DISPLACEMENT)
             for i in range(3):
                 if abs(u[i]) > 0.0:
-                    error = (d[i] - u[i])/u[i]
-                    if error > 1.0e-6:
+                    error = abs((d[i] - u[i])/u[i])
```
loumalouomega: BTW, this is a bug in the check

AlejandroCornejo: because of not using the `std::abs` ?

AlejandroCornejo:
The element I'm pretty sure that is fine cause I've compared the result with another solver (Ramseries) and works fine. Maybe I'm wrong hehe

loumalouomega:
> The element I'm pretty sure that is fine cause I've compared the result with another solver (Ramseries) and works fine. Maybe I'm wrong hehe

Could be..., what I don't understand is why your cpp test works

applications/StructuralMechanicsApplication/tests/test_patch_test_small_strain.py
loumalouomega: Because not using abs (could give false negative if the error is negative)

AlejandroCornejo:
> > The element I'm pretty sure that is fine cause I've compared the result with another solver (Ramseries) and works fine. Maybe I'm wrong hehe
> 
> Could be..., what I don't understand is why your cpp test works

well my cpp is simpler, I'm only comparing stresses comparing with the current status, not against any analytical value. (I assumed that was ok)

AlejandroCornejo:
I suppose that the ordering of the elements is consistent and not generating negative jacobians

loumalouomega:
> I suppose that the ordering of the nodes within the elements is consistent and not generating negative jacobians

A priori is the same geometry as the regular 2D tests

AlejandroCornejo:
And what about the 2*PI factor? does it have to be taken into account somehow in the test?

loumalouomega:
> And what about the 2*PI factor? does it have to be taken into account somehow in the test?

I would say that the displacements are the same. The 2Pi is to integrate

loumalouomega:
Any news?

loumalouomega:
@AlejandroCornejo could you think about this?

loumalouomega:
> @AlejandroCornejo could you think about this?

👋

applications/StructuralMechanicsApplication/tests/test_patch_test_small_strain.py
```diff
@@ -52,10 +52,12 @@ def _apply_material_properties(self,mp,dim):
...
         mp.GetProperties()[1].SetValue(KratosMultiphysics.VOLUME_ACCELERATION,g)
 
-        if(dim == 2):
+        if dim == 2:
             cl = StructuralMechanicsApplication.LinearElasticPlaneStress2DLaw()
-        else:
+        elif dim == 3:
             cl = StructuralMechanicsApplication.LinearElastic3DLaw()
+        else:
+            cl = StructuralMechanicsApplication.LinearElasticAxisym2DLaw()
```
AlejandroCornejo, line 60: then we are not going to enter here right?

applications/StructuralMechanicsApplication/tests/test_patch_test_small_strain.py
```diff
@@ -292,6 +295,41 @@ def test_SmallDisplacementElement_2D_triangle(self):
 
         #self.__post_process(mp)
 
+    def test_SmallDisplacementElement_2D_triangle_axisymm(self):
+        dim = 2
```
AlejandroCornejo, line 299: here we set the dim as 2

applications/StructuralMechanicsApplication/tests/test_patch_test_small_strain.py
```diff
@@ -292,6 +295,41 @@ def test_SmallDisplacementElement_2D_triangle(self):
...
+        KratosMultiphysics.VariableUtils().AddDof(KratosMultiphysics.DISPLACEMENT_X, KratosMultiphysics.REACTION_X,mp)
+        KratosMultiphysics.VariableUtils().AddDof(KratosMultiphysics.DISPLACEMENT_Y, KratosMultiphysics.REACTION_Y,mp)
+        KratosMultiphysics.VariableUtils().AddDof(KratosMultiphysics.DISPLACEMENT_Z, KratosMultiphysics.REACTION_Z,mp)
+
+        #create a submodelpart for boundary conditions
+        bcs = mp.CreateSubModelPart("BoundaryCondtions")
+        bcs.AddNodes([1,2,3,4])
+
+        #create Element
+        mp.CreateNewElement("AxisymSmallDisplacementElement2D3N", 1, [1,2,5], mp.GetProperties()[1])
```
AlejandroCornejo, line 321: I suppose that the ordering is ok regarding the Jacobian

applications/StructuralMechanicsApplication/tests/test_patch_test_small_strain.py
loumalouomega: Wait...

applications/StructuralMechanicsApplication/tests/test_patch_test_small_strain.py
loumalouomega: I feel stupid

loumalouomega: See line 303,  I am setting 2.5

loumalouomega: I feel less stupid

PR #10180, opened 2022-08-22
[All] Most extended typo fix: auxiliar -> auxiliary

loumalouomega:
**📝 Description**

This PR fixes the most extended typo in every app. auxiliar -> auxiliary

(Yes, I know this is a pain in the ass to review. I did a commit for every app. I can do it in different steps, but this is so trivial that I find it better this way)

**🆕 Changelog**

- auxiliar -> auxiliary


matekelemen:
You're my hero! <3

loumalouomega:
> You're my hero! <3

Most of the typos are my fault. Sorry, auxiliary is auxiliar in Spanish, and I always thought it was the same word until a PR of @rubenzorrilla where he wrote well the word

matekelemen:
That doesn't matter; the point is you fixed them!

kratos/tests/cpp_tests/utilities/test_auxiliar_model_part_utilities.cpp
```diff
@@ -18,7 +18,7 @@
 #include "containers/model.h"
 #include "testing/testing.h"
 #include "includes/model_part.h"
-#include "utilities/auxiliar_model_part_utilities.h"
+#include "utilities/auxiliary_model_part_utilities.h"
```
roigcarlo, line 21: This filename should be changed as well. I mean the file containing this change, not the include. I don't know to attach the comment to the file name :S

roigcarlo (commented):
Just a missing file name to change.

I will approve once its changed, but please try not to make such big PR it takes forever to review.

kratos/tests/cpp_tests/utilities/test_auxiliar_model_part_utilities.cpp
loumalouomega: The name is changed

loumalouomega: Ahh, the test name...

loumalouomega: Done

roigcarlo (approved):

roigcarlo (changes requested):
Also, please, add a note or something to tell people that `auxiliar_process_list` is now `auxiliary_process_list`

I didn't realized before, but this is an API breaker, so it must be proper documented.

loumalouomega:
> `auxiliar_process_list` is now `auxiliary_process_list`

Warning added

roigcarlo:
I meant something more on the lines of a warning when used inside the code. API breaking change normally are announced with 6 mont grace period for people to update. In this status we cannot merge, as the change potentially breaks all existing *.json files.

Let us check this before in the @KratosMultiphysics/technical-committee on how to proceed.

loumalouomega:
> I meant something more on the lines of a warning when used inside the code. API breaking changes normally are announced with 6 months grace period for people to update. In this status we cannot merge, as the change potentially breaks all existing *.json files.
> 
> Let us check this before in the @KratosMultiphysics/technical-committee on how to proceed.

Now I understand. FYI I was the only one using it, as I was the one who introduced... I will add the check

applications/FSIApplication/python_scripts/fsi_analysis.py
```diff
@@ -111,7 +111,8 @@ def _GetOrderOfProcessesInitialization(self):
                 "fluid_gravity",
                 "fluid_initial_conditions_process_list",
                 "fluid_boundary_conditions_process_list",
-                "fluid_auxiliary_process_list"]
```
loumalouomega: @roigcarlo I did this commit in order to do the transition and avoid a behaviour change

loumalouomega:
> I meant something more on the lines of a warning when used inside the code. API breaking changes normally are announced with 6 months grace period for people to update. In this status we cannot merge, as the change potentially breaks all existing *.json files.
> 
> Let us check this before in the @KratosMultiphysics/technical-committee on how to proceed.

See [KratosMultiphysics/Kratos@`e64abef` (#10180)](https://github.com/KratosMultiphysics/Kratos/pull/10180/commits/e64abefec281a5476a93572304c49f0d83e4942b)

miguelmaso:
Is it possible to merge this? Just to avoid conflicts...

loumalouomega:
> Is it possible to merge this? Just to avoid conflicts...

Please @roigcarlo I added the retro-compatibility for the parameters

roigcarlo:
@KratosMultiphysics/technical-committee ?

I've gone through all the code I know and there should be no more problems (that I've detected).

RiccardoRossi:
well, to give my 2 cents i would rather live with "auxiliar" rather than rename files and hence lose all of their history.

honestly i see this as a rather minor mistake... (auxiliar exist as archaic/obsolete form, and it sounds quite acceptable to non english speakers...)

matekelemen:
> rename files and hence lose all of their history.

Git [detects moves/renames](https://stackoverflow.com/a/2314757/12350793) up to some similarity, and ```auxiliar``` => ```auxiliary``` is definitely a small enough change for that to succeed. If you take a look at the diff, none of the files are deleted and all of them are correctly detected as moved.

philbucher:
Renaming files is ok I think 

I a much more worried about changing and possibly breathing Input files for a typo (that I didn't even realize)

applications/CoSimulationApplication/tests/fsi_sdof/ProjectParametersCFD.json
```diff
@@ -180,7 +180,7 @@
             }
 
         }],
-        "auxiliar_process_list"            : [{
```
RiccardoRossi, line 183: to me this kind of changes is a blocker. IT would cause cases that were prevously running to fail.

you can (and you did) fix the tests, but this also breaks existing examples that users have on their own computer

RiccardoRossi (changes requested):
Dear @loumalouomega i took some time to look into this, and we also had a brief online discussinon in the @KratosMultiphysics/technical-committee, so here i am writing on their behalf

we all think that it is NOT a good idea to make this change for the following reasons:
1 - way too many files accepted => very easy to introduce errors
2 - no real advantage after the merging (this, although nice, it is a cosmetic change))
3 - changes in json input files break existing working examples, and put a burden onto the users

we do appreciate that you put a lot of effort into this ... but we really think that this sort of changes should not be admitted without a strong reason. For the future please propose and let's discuss the change before it is actually made ... we would really like not to waste your time and effort...

applications/CoSimulationApplication/tests/fsi_sdof/ProjectParametersCFD.json
loumalouomega: I know, that's why I added the "patchs". I will separate this in smaller PR...

PR #10182, opened 2022-08-23
[Stat] Update spatial statistics process

sunethwarna:
**📝 Description**
This PR re-aranges how the output is printed to file by `SpatialStatisticsProcess`. It re-arranges the json settings as well.

**🆕 Changelog**
- Updated the `SpatialStatisticsProcess`
- Updated tests and cleaned lots of files


applications/StatisticsApplication/python_scripts/spatial_statistics_process.py
```diff
@@ -39,23 +40,38 @@ def __init__(self, model, settings):
...
-            "model_part_name" : "PLEASE_SPECIFY_MODEL_PART_NAME",
+            "model_part_name"         : "PLEASE_SPECIFY_MODEL_PART_NAME",
+            "echo_level"              : 0,
+            "computation_processes"   : [],
+            "computation_points"      : [
+                "ExecuteInitialize",
+                "ExecuteInitializeSolutionStep",
+                "Execute",
+                "ExecuteFinalizeSolutionStep"
+            ],
```
rubenzorrilla, line 51: I think we should specify `Execute` as unique computation point by default (not 100% sure).

rubenzorrilla (commented):
I'm not very familiar with the app but in any case changes seem reasonable to me. My unique observation is that we will be breaking current API right?

applications/StatisticsApplication/python_scripts/spatial_statistics_process.py
sunethwarna: I am not sure about this since some processes have defined multiple steps. Just in case some one uses this with some other process and they don't know the exact computation points. I left the flexibility in there also.

mpentek:
I think the user base is so limited and specific, I would go ahead with merging and stress less that it is API-breaking.

@mtnzguillermo

applications/StatisticsApplication/python_scripts/spatial_statistics_process.py
rubenzorrilla: Fair enough. My point was to promote the use of  `Execute` processes only, but at this point this is probably way too strict.

rubenzorrilla:
> I think the user base is so limited and specific, I would go ahead with merging and stress less that it is API-breaking.
> 
> @mtnzguillermo

Agree. I think it's worth in this case.

rubenzorrilla (approved):
Fine for me. @mpentek @mtnzguillermo  feel free to dismiss my approval if you have any further consideration.

PR #10184, opened 2022-08-23
[Core] Clean up legacy code from FromJsonCheckResultProcess

loumalouomega:
**📝 Description**

Clean up legacy code from FromJsonCheckResultProcess. The old python code was replaced 2 years ago by C++

**🆕 Changelog**

- Clean up legacy code from FromJsonCheckResultProcess


loumalouomega:
Okay now I remember why we keep it, it was because the @KratosMultiphysics/mpm application, any plan to migrate to C++?, can we just move the code there so at least the legacy code is not there?

loumalouomega:
> Okay now I remember why we keep it, it was because the @KratosMultiphysics/mpm application, any plan to migrate to C++?, can we just move the code there so at least the legacy code is not there?

Just to say that the migration would be relatively easy, but for the moment keeping as legacy code

loumalouomega:
Ping @philbucher

philbucher (approved):

PR #10188, opened 2022-08-24
[DEM] Add a parallel bond model for cohesive material and update the rolling friction calculation

ChengshunShang1996:
**📝 Description**
This PR added a Parallel Bond Model (PBM) for cohesive material and updated the rolling friction calculation. 

**🆕 Changelog**
- Fixed the rotation moment calculation for continnuum DEM
- Added the Parallel Bond Model (PBM) 
- Added a Quadratic contact model for conical contact 
- Added a classical Linear contact model 
- Separated the rolling friction calculation from [spheric_particle.cpp](https://github.com/KratosMultiphysics/Kratos/blob/dem/bondedParticleModel/applications/DEMApplication/custom_elements/spheric_particle.cpp) and [spheric_continuum_particle.cpp](https://github.com/KratosMultiphysics/Kratos/blob/dem/bondedParticleModel/applications/DEMApplication/custom_elements/spheric_continuum_particle.cpp)
- Separated the moment calculation from [spheric_particle.cpp](https://github.com/KratosMultiphysics/Kratos/blob/dem/bondedParticleModel/applications/DEMApplication/custom_elements/spheric_particle.cpp) and [spheric_continuum_particle.cpp](https://github.com/KratosMultiphysics/Kratos/blob/dem/bondedParticleModel/applications/DEMApplication/custom_elements/spheric_continuum_particle.cpp)
- Added a constant torque rolling friction model
- Added a [DEM_compound_constitutive_law_for_PBM.h](https://github.com/KratosMultiphysics/Kratos/blob/dem/bondedParticleModel/applications/DEMApplication/custom_constitutive/DEM_compound_constitutive_law_for_PBM.h) file for combining the PBM with different base models
- Added the PBM test files


maceligueta:
Hello @ChengshunShang1996 . This is a big pull request, with many files modified, so it will take a time to review.

maceligueta (changes requested):
Please do not change .gitignore, it is common for the whole Kratos community.
Instead, you can ignore files in your computer by adding them to the file `kratos/.git/info/exclude`.
This file is yours and you can add any file name or folder name you want.

applications/DEMApplication/custom_constitutive/DEM_D_Hertz_viscous_Coulomb_2D_CL.cpp
```diff
@@ -44,9 +47,14 @@ namespace Kratos {
         const double my_shear_modulus = 0.5 * my_young / (1.0 + my_poisson);
         const double walls_shear_modulus = 0.5 * walls_young / (1.0 + walls_poisson);
         const double equiv_shear = 1.0 / ((2.0 - my_poisson) / my_shear_modulus + (2.0 - walls_poisson) / walls_shear_modulus);
-
-        mKn = 0.7854 * equiv_young;     //KRATOS_M_PI_4 = 0.7854
+        const double element_radius = element->GetRadius();
+        mKn = 2.0 * element_radius * equiv_young;
```
maceligueta: In order to avoid the future question 'why this?' we could paste here the reference of the paper or book used for this. It will be very helpful if possible.

applications/DEMApplication/custom_constitutive/DEM_D_Hertz_viscous_Coulomb_2D_CL.cpp
```diff
@@ -44,9 +47,14 @@ namespace Kratos {
...
+        //mKn = 0.7854 * equiv_young;     //KRATOS_M_PI_4 = 0.7854
         mKt = 4.0 * equiv_shear * mKn / equiv_young;
     }
 
+    double DEM_D_Hertz_viscous_Coulomb2D::CalculateNormalForce(const double indentation) {
```
maceligueta, line 55: is this necessary? If base class has the same function there is no need to override.

maceligueta: Sorry, base class does a different thing. Solved.

applications/DEMApplication/custom_constitutive/DEM_D_Linear_HighStiffness_CL.h
```diff
@@ -23,11 +23,16 @@ namespace Kratos {
...
+        void Check(Properties::Pointer pProp) const override;
+
         void InitializeContact(SphericParticle* const element1, SphericParticle* const element2, const double indentation) override;
 
         void InitializeContactWithFEM(SphericParticle* const element, Condition* const wall, const double indentation, const double ini_delta = 0.0) override;
 
-    };
+    protected:
 
+        Properties::Pointer mpProperties;
```
maceligueta: Discontinuum walls do not have member variables. Are we sure we need this one? We can find the properties of this contact by doing:
`Properties& properties_of_this_contact = element1->GetProperties().GetSubProperties(element2->GetProperties().Id());`

applications/DEMApplication/custom_constitutive/DEM_D_Linear_classic_CL.h
```diff
@@ -0,0 +1,42 @@
...
+
+        std::unique_ptr<DEMDiscontinuumConstitutiveLaw> CloneUnique() override;
+
+        void InitializeContact(SphericParticle* const element1, SphericParticle* const element2, const double indentation) override;
+
+        void InitializeContactWithFEM(SphericParticle* const element, Condition* const wall, const double indentation, const double ini_delta = 0.0) override;
+
+    protected:
+
+        Properties::Pointer mpProperties;
```
maceligueta: Same comment

applications/DEMApplication/custom_constitutive/DEM_D_Linear_viscous_Coulomb_2D_CL.cpp
```diff
@@ -23,30 +23,44 @@ namespace Kratos {
...
         const double walls_young      = wall->GetProperties()[YOUNG_MODULUS];
         const double my_poisson       = element->GetPoisson();
         const double walls_poisson    = wall->GetProperties()[POISSON_RATIO];
 
         const double equiv_young      = my_young * walls_young / (walls_young * (1.0 - my_poisson * my_poisson) + my_young * (1.0 - walls_poisson * walls_poisson));
 
-        const double my_shear_modulus = 0.5 * my_young / (1.0 + my_poisson);
-        const double walls_shear_modulus = 0.5 * walls_young / (1.0 + walls_poisson);
-        const double equiv_shear = 1.0 / ((2.0 - my_poisson)/my_shear_modulus + (2.0 - walls_poisson)/walls_shear_modulus);
+        //const double my_shear_modulus = 0.5 * my_young / (1.0 + my_poisson);
```
maceligueta: Commented code can be removed in general. It can be visited by checking previous versions with git

applications/DEMApplication/custom_constitutive/DEM_D_Linear_viscous_Coulomb_CL.cpp
```diff
@@ -211,8 +211,7 @@ namespace Kratos {
 
         //const double alpha = 0.05;
         const double modified_radius = effective_radius * 0.31225;
-        const double mKn_augmenter = 1.0; // 100.0;
-        mKn = mKn_augmenter * equiv_young * Globals::Pi * modified_radius; // 2.0 * equiv_young * sqrt_equiv_radius;
+        mKn = equiv_young * Globals::Pi * modified_radius; // 2.0 * equiv_young * sqrt_equiv_radius;
```
maceligueta, line 214: Nice cleanup

applications/DEMApplication/custom_constitutive/DEM_KDEM_with_damage_CL.cpp
```diff
@@ -99,7 +99,9 @@ namespace Kratos {
                 sliding,
                 r_process_info);
 
-        FindMaximumValueOfNormalAndTangentialDamageComponents();
+        // FindMaximumValueOfNormalAndTangentialDamageComponents();
+        
+        CalculationOfNormalAndTangentialDamageComponents();
```
maceligueta: Name of functions must be an imperative verb. `CalculateNormalAndTangentialDamageComponents`

applications/DEMApplication/custom_constitutive/DEM_compound_constitutive_law_for_PBM.h
```diff
@@ -0,0 +1,56 @@
...
+        return mCCL.CalculateNormalForce(indentation);
+    }
+
+    void InitializeContact(SphericParticle* const element1, SphericParticle* const element2, const double indentation) override {
+        return mCCL.InitializeContact(element1, element2, indentation);
+    }
+
+private:
+
+    UnbondCL mCCL;
```
maceligueta, line 39: This could be a pointer instead of a full object. We can discuss it.

applications/DEMApplication/python_scripts/DEM_analysis_stage.py
```diff
@@ -242,14 +240,89 @@ def FillAnalyticSubModelPartsWithNewParticles(self):
         #analytic_particle_ids = [elem.Id for elem in self.spheres_model_part.Elements]
         #self.analytic_model_part.AddElements(analytic_particle_ids)
 
+    def SkinForCube(self):
```
maceligueta: Let's remove these two functions added. It seems nobody is calling them, and they should have never been in this file, but in a derived from this.

applications/DEMApplication/python_scripts/DEM_procedures.py
```diff
@@ -159,8 +158,8 @@ def UpdateData(self, domain_volume):
 
         self.voids_volume = domain_volume - self.solid_volume
         self.global_porosity = self.voids_volume / domain_volume
-        if self.do_print_results_option:
-            self.PrintCurrentData()
+
+        self.PrintCurrentData()
```
maceligueta, line 162: Is this function being called always? Or is it just available for future calls? We don't want to print files always...

applications/DEMApplication/python_scripts/DEM_procedures.py
```diff
@@ -1034,6 +1038,17 @@ def FinalizeGraphs(self, rigid_face_model_part):
         if self.TestType == "None":
             self.close_graph_files(rigid_face_model_part)
 
+    def PrintAdditionalGraphs(self, time, solver):
```
maceligueta, line 1041: Let's check that we are not printing graphs always with this function. Printing should always be an option marked by the user, not hardcoded and always being called...

applications/PoromechanicsApplication/CMakeLists.txt
```diff
@@ -2,6 +2,10 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
 message("**** configuring KratosPoromechanicsApplication ****")
 
+kratos_add_dependency(${KRATOS_SOURCE_DIR}/applications/StructuralMechanicsApplication)
+kratos_add_dependency(${KRATOS_SOURCE_DIR}/applications/LinearSolversApplication)
+kratos_add_dependency(${KRATOS_SOURCE_DIR}/applications/FluidDynamicsApplication)
```
maceligueta, line 7: @ipouplana Are we sure the Poromechanics app requires the FluidDynamics App? I guess that someone put it here for a reason (maybe I put it),

applications/DEMApplication/custom_constitutive/DEM_D_Linear_classic_CL.h
ChengshunShang1996: Thanks for your comments. This variable is unnecessary here and has been removed.

applications/DEMApplication/custom_constitutive/DEM_D_Linear_viscous_Coulomb_2D_CL.cpp
ChengshunShang1996: Thanks. I will remove the commented code.

applications/DEMApplication/custom_constitutive/DEM_D_Linear_viscous_Coulomb_CL.cpp
ChengshunShang1996: Thanks

applications/DEMApplication/custom_constitutive/DEM_KDEM_with_damage_CL.cpp
ChengshunShang1996: Modified. Thanks!

applications/DEMApplication/custom_constitutive/DEM_D_Linear_HighStiffness_CL.h
ChengshunShang1996: Thanks for the comment. `Properties::Pointer mpProperties;` has been removed. For ball-ball and ball-wall contacts, the `Properties& properties_of_this_contact = element1->GetProperties().GetSubProperties(element2->GetProperties().Id());` is used.

applications/DEMApplication/custom_constitutive/DEM_D_Hertz_viscous_Coulomb_2D_CL.cpp
ChengshunShang1996: Thanks! Maybe @salvalatorre knows about this equation.

applications/DEMApplication/custom_constitutive/DEM_compound_constitutive_law_for_PBM.h
ChengshunShang1996: Thanks! Discontinuum laws do not save memory because they do not have memory variables, so it is similar to save a point or an instance. Also: what would be the pointer pointing at?

applications/DEMApplication/python_scripts/DEM_procedures.py
ChengshunShang1996: Thanks! This function only works at the begining and prints once. I think we can leave it here.

applications/DEMApplication/python_scripts/DEM_procedures.py
ChengshunShang1996: Yes, it only prints the coordination number only if the user ask for it.

applications/CoSimulationApplication/tests/dem_fem_cable_net/MaterialsDEM.json
```diff
@@ -23,6 +23,7 @@
             "COEFFICIENT_OF_RESTITUTION": 0.2,
             "STATIC_FRICTION": 0.9999999999999999,
             "DYNAMIC_FRICTION": 0.9999999999999999,
+            "DEM_ROLLING_FRICTION_MODEL_NAME": "DEMRollingFrictionModelConstantTorque",
```
philbucher: Can you please remove this here and make it a default?
Otherwise every existing case is broken

ChengshunShang1996: Yes, I can. I have tested that making this parameter a default can works well. But listing it in the input file maybe more clear because we will have several different rolling friction models in the future. What is your opinion about this? @maceligueta 
I have also tested that adding this parameter in the input file can works well with the existing cases.

maceligueta: If the case runs fine without this line (because the default value is ok) I would remove that from the CoSimulation application. However, in the DEM and SwimmingDEM app I would like to see which model is being used, so I would keep this line in DEM and SDEM.

ChengshunShang1996: Okay. I will make it a default parameter and print some information in screen to tell user that they are using a default rolling friction model.

applications/DEMApplication/custom_constitutive/DEM_compound_constitutive_law_for_PBM.h
maceligueta: Right, if we use a pointer the structure will be harder to understand, and the memory overhead is small or zero.

applications/DEMApplication/python_scripts/sphere_strategy.py
```diff
@@ -718,6 +718,10 @@ def ModifySubProperties(self, properties, param = 0):
         if not properties.Has(ROLLING_FRICTION_WITH_WALLS):
             properties[ROLLING_FRICTION_WITH_WALLS] = properties[ROLLING_FRICTION]
 
+        if not properties.Has(DEM_ROLLING_FRICTION_MODEL_NAME):
+            properties[DEM_ROLLING_FRICTION_MODEL_NAME] = "DEMRollingFrictionModelConstantTorque"
+            self.Procedures.KratosPrintWarning("Using a default rolling friction model [DEMRollingFrictionModelConstantTorque].")
```
maceligueta: Here we could print the properties id, so the message can be understood without debugging

applications/DEMApplication/custom_constitutive/DEM_rolling_friction_model_constant_torque.cpp
```diff
@@ -0,0 +1,87 @@
...
+    void DEMRollingFrictionModelConstantTorque::Check(Properties::Pointer pProp) const {
+        
+        if(!pProp->Has(DEM_ROLLING_FRICTION_MODEL_NAME)) 
+        {
+            KRATOS_ERROR << "Variable DEM_ROLLING_FRICTION_MODEL_NAME should be presented."<<std::endl;
+        }
+        
+    }
+
+    void DEMRollingFrictionModelConstantTorque::ComputeRollingFriction(SphericParticle* p_element, SphericParticle* p_neighbor, double LocalContactForce[3], array_1d<double, 3>& mContactMoment)
```
joaquinirazabal: Hi, @ChengshunShang1996. I implemented the rolling friction a long time ago, and maybe I am missing something, but here I see that here you are omitting the check to avoid the particles starting to rotate in the opposite direction.
If you look in the previous code, there is an if (MR_max > MR_now)
I also attach a file where the way of applying the rolling friction is explained.
Let me know if I am wrong.
[2017_IRAZABAL_BROF_MODEL_LATERAL_RESISTANCE.pdf](https://github.com/KratosMultiphysics/Kratos/files/9468847/2017_IRAZABAL_BROF_MODEL_LATERAL_RESISTANCE.pdf)

applications/DEMApplication/custom_constitutive/DEM_D_Hertz_viscous_Coulomb_2D_CL.cpp
ChengshunShang1996: This question has been solved.

applications/DEMApplication/python_scripts/sphere_strategy.py
ChengshunShang1996: Updated.

applications/DEMApplication/custom_constitutive/DEM_rolling_friction_model_constant_torque.cpp
ChengshunShang1996: Hi, @joaquinirazabal. Currently, the bounded rolling friction model proposed in your paper has been recovered. You can find it here [DEM_rolling_friction_model_bounded.cpp](https://github.com/KratosMultiphysics/Kratos/blob/dem/bondedParticleModel/applications/DEMApplication/custom_constitutive/DEM_rolling_friction_model_bounded.cpp). By setting the material relationship in json file  as below, we can use it.
`"DEM_ROLLING_FRICTION_MODEL_NAME"        : "DEMRollingFrictionModelBounded"`

joaquinirazabal: Great, @ChengshunShang1996, thank you. Nice and huge work!

ChengshunShang1996: Thanks, @joaquinirazabal . I have tested this rolling friction model and it works well with a collapse test. If you find any unexpected results for special cases, we can improve it continuously.

ChengshunShang1996:
Hi @maceligueta. I think this PR is very close to being merged.

applications/DEMApplication/python_scripts/sphere_strategy.py
```diff
@@ -1,3 +1,4 @@
+from pickle import FALSE, TRUE
```
maceligueta: Is this import necessary?

applications/DEMApplication/custom_elements/spheric_particle.cpp
```diff
@@ -1659,42 +1635,47 @@ void SphericParticle::ComputeDifferentialStrainTensor(const ProcessInfo& r_proce
...
-            CoefficientsMatrix(j,i) += relative_position[j] * relative_position[i];
-            RightHandSide(j,i) += relative_delta_displacement[i] * relative_position[j];
-        }
-    }
-
-    for (unsigned int i = 0; i < mNeighbourElements.size(); i++) {
-        if (!mNeighbourElements[i]) continue;
-        relative_position = mNeighbourElements[i]->GetGeometry()[0].Coordinates() - assembly_centroid;
-        relative_delta_displacement = mNeighbourElements[i]->GetGeometry()[0].FastGetSolutionStepValue(DELTA_DISPLACEMENT) - assembly_average_delta_displacement;
+        array_1d<double, 3> relative_position = this->GetGeometry()[0].Coordinates() - assembly_centroid;
```
maceligueta, line 1647: `const array_1d<double, 3>& relative_position`
(if possible)

applications/DEMApplication/python_scripts/sphere_strategy.py
ChengshunShang1996: This import was automatically added when I was trying to use FALSE and TRUE. It is not necessary now and has been removed.

maceligueta (dismissed):

maceligueta (approved):

PR #10189, opened 2022-08-24
Fixing MPI slow tests in CI

### LINKED ISSUE #10169: Mpi tests taking long time
**Description**
Investigating the CI times, I've noticed that GCC mpi tests when oversubscriving are having a huge overhead. 
```
Gcc   3 cores -> 10-15 min
Clang 3 cores -> 3 min
Gcc   4 cores -> 15-20 min
Clang 4 cores -> 4 min
```

@KratosMultiphysics/mpi We need to investigate this. I am very worried about it :S

For example, a random build https://github.com/KratosMultiphysics/Kratos/runs/7914657048?check_suite_focus=true

**Scope**
- CI

roigcarlo:
Description:

Fixes #10169. Two problems were interacting to cause problems in the MPI CI runs when over-subscription was happening:
- Backend was unaware of such over-subscription
- OpenMP was not disabled during the execution of MPI runs.

This was causing, for example, that Gcc with MPI 4 and OMP 2 was spawning 4 competing processes which each occupying all the machine resources. 

Clang for some reason generated a code which is much more aggressive in limiting the resources available to each thread, which caused this problem not to happen. I will investigate the details of this as I am not 100% aware of which is the mechanism they use and could be interesting.

With this fix the GCC testing time (and hence CI time since it was bounded by these tests) will be reduced in about ~50 min per run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,11 @@ jobs:
         export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:${GITHUB_WORKSPACE}/bin/${{ matrix.build-type }}/libs
         python3 kratos/python_scripts/testing/run_tests.py -v 2 -l nightly -c python3
 
+    - name: Prepare Parallel Env
+      shell: bash
+      run: |
+        echo "localhost slots=2" >> ci_hostfile
+
     - name: Running MPICore C++ tests (2 Cores)
       shell: bash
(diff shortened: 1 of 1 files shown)


loumalouomega:
But I want to approve!!

roigcarlo:
@loumalouomega you can approve now :heart:

loumalouomega (approved):
Approving

PR #10191, opened 2022-08-25
[ConstitutiveLawsApplication] Registering isotropic damage laws properly

AlejandroCornejo:
**📝 Description**
This task has been in my TODO list for several years. 

I am renaming the Isotropic damage template registrations to ONLY the useful cases. Now the damage is registered once per yield surface, before I registered all the combinations even if they were not used.

Additionally, **I have added the plane stress damage** and registered them in a very intuitive way according to its yield surface: `SmallStrainIsotropicDamage3DVonMises`, `SmallStrainIsotropicDamagePlaneStrainVonMises` and `SmallStrainIsotropicDamagePlaneStressVonMises`, for example. This means that the compilation time will be reduced (1/3 of registrations in damage laws).

@lagoncalvesjr @SergioJimenezReyes @balcayde @atfard This means that from this PR on the damage laws will be called in the `Materials.json` like I am mentioning here. @SergioJimenezReyes I had te register accordingly your HCF laws to be consistent with the damage ones. Now it will be called like `SmallStrainHighCycleFatigue3DLawVonMises`

@loumalouomega In order to do this I had to move the Elastic Matrix calculation in the linear elastic cl's inside the `ConstitutiveLawsUtilities` so I could call the static method from the plane stress damage (derived from generic 3D damage). No change of behaviour in this sense, only moving the source impleemntation. Actually the plane stress damage law is empty, only overrides the `CalculateValue `method of the `CONSTITUTIVE_MATRIX`.

@maceligueta I finally did it ;)


loumalouomega:
~~~
/__w/Kratos/Kratos/applications/ConstitutiveLawsApplication/custom_constitutive/small_strains/damage/generic_small_strain_isotropic_damage_plane_stress.h:65:22: error: expected a qualified name after 'typename'
    typedef typename GenericSmallStrainIsotropicDamage<TConstLawIntegratorType> BaseType;
~~~

applications/ConstitutiveLawsApplication/custom_constitutive/small_strains/damage/generic_small_strain_isotropic_damage_plane_stress.h
```diff
@@ -0,0 +1,209 @@
...
+//                                                                                         |_|   |_|
+//
+//  License:         BSD License
+//                   license: structural_mechanics_application/license.txt
+//
+//  Main author:    Alejandro Cornejo
+//
+//
+
+#if !defined(KRATOS_GENERIC_SMALL_STRAIN_ISOTROPIC_DAMAGE_PLANE_STRESS_H_INCLUDED)
```
loumalouomega: Use pragma once instead

loumalouomega (changes requested):
Minor comment (and fix CI)

applications/ConstitutiveLawsApplication/custom_constitutive/small_strains/damage/generic_small_strain_isotropic_damage_plane_stress.h
AlejandroCornejo: you mean adding `#pragma once` before that?

loumalouomega: Replace it with #pragma once

loumalouomega (dismissed):

applications/ConstitutiveLawsApplication/tests/cpp_tests/test_high_cycle_fatigue_constitutive_law_3d.cpp
```diff
@@ -36,7 +36,7 @@ namespace Testing
 {
 // We test the associated damage Constitutive laws...
 typedef Node<3> NodeType;
-typedef GenericSmallStrainHighCycleFatigueLaw<GenericConstitutiveLawIntegratorDamage<VonMisesYieldSurface<VonMisesYieldPlasticPotential<6>>>> FatigueLawType;
+typedef GenericSmallStrainHighCycleFatigueLaw<GenericConstitutiveLawIntegratorDamage<VonMisesYieldSurface<VonMisesPlasticPotential<6>>>> FatigueLawType;
```
loumalouomega, line 38: image

SergioJimenezReyes (commented):
Thanks!

AlejandroCornejo:
can anyone reapprove? :)

loumalouomega (approved):

PR #10201, opened 2022-08-29
[Release] Enable FemToDem

roigcarlo:
**📝 Description**
Some groups have shown interest in this application to be in release, as it passes the CI I see no problem in adding it. 

Sadly a flaw still exists and having this app requires us to provide access to DelaunayMeshing through pip. This app depends on tetgen and cannot be released under BSD. I have added a mechanism to licence individual packages in Kratos with different licences until either PfemFluid stops depending on tetgen, or FemToDem stops depending on PfemFluid.

**Under no circumstance** this opens the door to releasing Kratos under a licence which is not BSD. This solutions comes as a recommendation from Tetgen devs for us to act as projects like [this](https://github.com/JuliaGeometry/TetGen.jl):

**🆕 Changelog**
- Added FemToDem to package release.
- Added PfemFluid to pakcage release.
- Added custom licencing mechanism for packages.


philbucher:
side comment, I guess the wheels folder can also be excluded from the CI

roigcarlo:
> side comment, I guess the wheels folder can also be excluded from the CI

That's very much true... let's merge #10203 to check there are no problems with the checks and I will change.

roigcarlo:
Ping @KratosMultiphysics/technical-committee do you agree with this? Can we proceed?

rubenzorrilla:
Just to confirm that I didn't misunderstood what we'll be doing here. Does this mean that we'll be releasing the non-BSD apps with their corresponding license but the Core and rest of applications will remain as BSD right?

roigcarlo:
> Just to confirm that I didn't misunderstood what we'll be doing here. Does this mean that we'll be releasing the non-BSD apps with their corresponding license but the Core and rest of applications will remain as BSD right?

Yes. As a matter of fact only Delaunay app will be released under AGPL3

rubenzorrilla:
> > Just to confirm that I didn't misunderstood what we'll be doing here. Does this mean that we'll be releasing the non-BSD apps with their corresponding license but the Core and rest of applications will remain as BSD right?
> 
> Yes. As a matter of fact only Delaunay app will be released under AGPL3

So what you get from doing `pip install KratosMultiphysics-all` is a mixture of one AGPL3 application and lots of BSD applications right?

roigcarlo:
Yep. If this is a concern we can move FemToDem outside Kratos-all.

rubenzorrilla:
> Yep. If this is a concern we can move FemToDem outside Kratos-all.

I think we should. If there is a non-BSD application, this must be a separate package in my opinion.

roigcarlo:
> I think we should. If there is a non-BSD application, this must be a separate package in my opinion.

Done!

PR #10207, opened 2022-08-31
Extending the list of exclusions

roigcarlo:
**📝 Description**
As suggested in #10201. 

@sunethwarna I've notice that part of the CMake/python files for the generation of the python hints exist in this folder. We should move that inside the source, scripts folder content should not interact with the actual compilation (at most trigger it).

Dockerfile builds will still happen in other workflows.

**🆕 Changelog**
- Extending CI exclusion to scripts and docker files.

### DIFF
--- a/.github/workflows/ci-dummy.yml
+++ b/.github/workflows/ci-dummy.yml
@@ -7,9 +7,10 @@ on:
   pull_request:
     branches:
       - master
-    paths: # this must be the same as .github/workflows/ci.yml - "paths-ignore" !
+    paths:              # This must be the same as .github/workflows/ci.yml - "paths-ignore" !
       - '**.md'
       - '/documents/**'
+      - '/scripts/**'
 
 jobs:
   ubuntu:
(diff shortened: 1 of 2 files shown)


matekelemen:
> part of the CMake/python files for the generation of the python hints exist in this folder.

They can be moved if you have a better place in mind. Don't forget to update the root ```CMakeLists.txt``` too: *scripts/post_install* is added as a subdirectory [at the very end of the root ```CMakeLists.txt```](https://github.com/KratosMultiphysics/Kratos/blob/8364089d69a2daecab9604f16308ad0335ee81b1/CMakeLists.txt#L652) (this is necessary to make sure that CMake executes the stub generating python script **after** everything else got installed).

sunethwarna:
I would say, lets keep it in the scripts folder because, those scripts are not executed when CI is run hence does not need to trigger the CI for changes.

sunethwarna (approved):

PR #10212, opened 2022-09-01
disable C++ non-standard compiler specific extensions

philbucher:
**📝 Description**
Wanted to do this for a while, this should help with portability

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,8 @@
 project (KratosMultiphysics)
 cmake_minimum_required (VERSION 2.8.6)
 set (CMAKE_CXX_STANDARD 17)
+set (CMAKE_CXX_STANDARD_REQUIRED ON)
+set (CMAKE_CXX_EXTENSIONS OFF)
 
 # Setting some policies
 # No recursive dereferencing


RiccardoRossi:
i am not really positive about deactivating it by default...let s.doscuss it in tje next TC meeting

RiccardoRossi:
i am not really positive about deactivating it by default...let s.doscuss it in tje next TC meeting

RiccardoRossi:
in essence i would make it very clear how to deactivate them...bit i would not change the default...

philbucher:
I am not sure I follow

Why would you allow them in the first place? I.e. it is frustrating to develop only to realize when CI is not passing that some compiler extensions were unknowingly used :thinking: 

Or even worse, some extensions are used that are not caught by the CI, but then it fails on the cluster that has an other/older toolchain

RiccardoRossi:
it looks like i posted in the wrong pr.

i am not positove about tests being turned off in release

philbucher:
So do you agree with this PR then?

loumalouomega (commented):
I agree

RiccardoRossi (approved):
@KratosMultiphysics/technical-committee i am approving since i think we all agree about this.

@philbucher please wait until tomorrow night to merge it in case someone in @KratosMultiphysics/technical-committee wants to rise a last minute red flag

RiccardoRossi:
we spoke about in @KratosMultiphysics/technical-committee and we are all positive, so please go ahead with the merging

PR #10214, opened 2022-09-01
[Core] Simplified interface to BinBasedFastPointLocator

RiccardoRossi:
**📝 Description**
This exposes a python interface to FindPointOnMesh which allows usage as

~~~py
is_found,N,elem = locator.FindPointOnMesh(coords)
~~~

essential if is_found==true

N contains the shape functions within the geometry of the element elem

Without this the FindPointOnMesh is not usable from python

**🆕 Changelog**

- Expose FindPointOnMesh to python interface


kratos/python/add_geometrical_utilities_to_python.cpp
```diff
@@ -198,12 +198,24 @@ void AddGeometricalUtilitiesToPython(pybind11::module &m)
...
         ;
 
     py::class_< BinBasedFastPointLocator < 3 >, BinBasedFastPointLocator < 3 >::Pointer >(m,"BinBasedFastPointLocator3D")
         .def(py::init<ModelPart&  >())
         .def("UpdateSearchDatabase", &BinBasedFastPointLocator < 3 > ::UpdateSearchDatabase)
         .def("FindPointOnMesh", &BinBasedFastPointLocator < 3 > ::FindPointOnMeshSimplified)
+        .def("FindPointOnMesh", [](BinBasedFastPointLocator < 3 >& self, const array_1d<double,3>& coords ){
+            Element::Pointer pelem=nullptr;
+            Vector N;
+            bool is_found = self.FindPointOnMeshSimplified(coords,N,pelem);
```
loumalouomega:
```suggestion
            const bool is_found = self.FindPointOnMeshSimplified(coords,N,pelem);
```

kratos/python/add_geometrical_utilities_to_python.cpp
```diff
@@ -198,12 +198,24 @@ void AddGeometricalUtilitiesToPython(pybind11::module &m)
         .def("UpdateSearchDatabase", &BinBasedFastPointLocator < 2 > ::UpdateSearchDatabase)
         .def("UpdateSearchDatabaseAssignedSize", &BinBasedFastPointLocator < 2 > ::UpdateSearchDatabaseAssignedSize)
         .def("FindPointOnMesh", &BinBasedFastPointLocator < 2 > ::FindPointOnMeshSimplified)
+        .def("FindPointOnMesh", [](BinBasedFastPointLocator < 2 >& self, const array_1d<double,3>& coords ){
+            Element::Pointer pelem=nullptr;
+            Vector N;
+            bool is_found = self.FindPointOnMeshSimplified(coords,N,pelem);
```
loumalouomega:
```suggestion
            const bool is_found = self.FindPointOnMeshSimplified(coords,N,pelem);
```

loumalouomega (dismissed):
Thanks!

kratos/python/add_geometrical_utilities_to_python.cpp
```diff
@@ -192,19 +191,31 @@ void AddGeometricalUtilitiesToPython(pybind11::module &m)
         .def("CreateNodesArray", &IsosurfacePrinterApplication::CreateNodesArray)
         ;
 
-    //binbased locators
+    // BinBasedFastPointLocator
     py::class_< BinBasedFastPointLocator < 2 >, BinBasedFastPointLocator < 2 >::Pointer >(m,"BinBasedFastPointLocator2D")
         .def(py::init<ModelPart& >())
         .def("UpdateSearchDatabase", &BinBasedFastPointLocator < 2 > ::UpdateSearchDatabase)
         .def("UpdateSearchDatabaseAssignedSize", &BinBasedFastPointLocator < 2 > ::UpdateSearchDatabaseAssignedSize)
         .def("FindPointOnMesh", &BinBasedFastPointLocator < 2 > ::FindPointOnMeshSimplified)
```
loumalouomega: What's the difference with this one?

kratos/python/add_geometrical_utilities_to_python.cpp
```diff
@@ -192,19 +191,31 @@ void AddGeometricalUtilitiesToPython(pybind11::module &m)
...
+            return std::tuple<bool,Vector,Element::Pointer>{is_found,N,p_elem};
+        })
         ;
 
     py::class_< BinBasedFastPointLocator < 3 >, BinBasedFastPointLocator < 3 >::Pointer >(m,"BinBasedFastPointLocator3D")
         .def(py::init<ModelPart&  >())
         .def("UpdateSearchDatabase", &BinBasedFastPointLocator < 3 > ::UpdateSearchDatabase)
-        .def("FindPointOnMesh", &BinBasedFastPointLocator < 3 > ::FindPointOnMeshSimplified)
         .def("UpdateSearchDatabaseAssignedSize", &BinBasedFastPointLocator < 3 > ::UpdateSearchDatabaseAssignedSize)
+        .def("FindPointOnMesh", &BinBasedFastPointLocator < 3 > ::FindPointOnMeshSimplified)
```
loumalouomega: What's the difference with this one?

loumalouomega (changes requested):
I realized that maybe it already existed. What is the difference?

kratos/python/add_geometrical_utilities_to_python.cpp
roigcarlo: The dimension

roigcarlo (commented):
To me look ok...

kratos/python/add_geometrical_utilities_to_python.cpp
```diff
@@ -192,19 +191,31 @@ void AddGeometricalUtilitiesToPython(pybind11::module &m)
...
+        .def("FindPointOnMesh", [](BinBasedFastPointLocator < 3 >& rSelf, const array_1d<double,3>& rCoords ){
+            Element::Pointer p_elem = nullptr;
+            Vector N;
+            const bool is_found = rSelf.FindPointOnMeshSimplified(rCoords,N,p_elem);
+            return std::tuple<bool,Vector,Element::Pointer>{is_found,N,pelem};
```
roigcarlo:
```suggestion
            return std::tuple<bool,Vector,Element::Pointer>{is_found,N,p_elem};
```

rubenzorrilla:
I think it would be a good idea to add a small test showing how to use it.

kratos/python/add_geometrical_utilities_to_python.cpp
RiccardoRossi: no, not just the dimension. To use this one one would need to pass a pointer to the element to be returned. To be honest i don't have a clue about how we can accomplish this from python. That is, from what i see that function cannot be used.
I am not removing it since someone might be using it...(although i don't know how, otherwise i would have done the same)

in any case the new interface is way simpler:

~~~py
[found,N,pelem] = locator.FindPointOnMesh(pos)
~~~

(where N and pelem only make sense if found==True, the same way as would happen in c++)

kratos/python/add_geometrical_utilities_to_python.cpp
RiccardoRossi: answered below (in short, the previous one is unusable or at least very difficult to use)

RiccardoRossi (commented):
from my point of view this is complete @loumalouomega ?

loumalouomega (dismissed):

kratos/python/add_geometrical_utilities_to_python.cpp
```diff
@@ -193,33 +192,57 @@ void AddGeometricalUtilitiesToPython(pybind11::module &m)
...
+            return std::tuple<bool,Vector,Condition::Pointer>{is_found,N,p_cond};
+        })
         ;
 
     py::class_< BinBasedFastPointLocatorConditions < 3 > >(m,"BinBasedFastPointLocatorConditions3D")
         .def(py::init<ModelPart&  >())
         .def("UpdateSearchDatabase", &BinBasedFastPointLocatorConditions < 3 > ::UpdateSearchDatabase)
-        .def("FindPointOnMesh", &BinBasedFastPointLocatorConditions < 3 > ::FindPointOnMeshSimplified)
         .def("UpdateSearchDatabaseAssignedSize", &BinBasedFastPointLocatorConditions < 3 > ::UpdateSearchDatabaseAssignedSize)
+        .def("FindPointOnMesh", &BinBasedFastPointLocatorConditions < 3 > ::FindPointOnMeshSimplified)
```
loumalouomega:
```suggestion
```
If unusable, we should simply remove it

kratos/python/add_geometrical_utilities_to_python.cpp
```diff
@@ -193,33 +192,57 @@ void AddGeometricalUtilitiesToPython(pybind11::module &m)
...
+            const bool is_found = rSelf.FindPointOnMeshSimplified(rCoords,N,p_elem);
+            return std::tuple<bool,Vector,Element::Pointer>{is_found,N,p_elem};
+        })
         ;
 
     py::class_< BinBasedFastPointLocatorConditions < 2 > >(m,"BinBasedFastPointLocatorConditions2D")
         .def(py::init<ModelPart& >())
         .def("UpdateSearchDatabase", &BinBasedFastPointLocatorConditions < 2 > ::UpdateSearchDatabase)
         .def("UpdateSearchDatabaseAssignedSize", &BinBasedFastPointLocatorConditions < 2 > ::UpdateSearchDatabaseAssignedSize)
         .def("FindPointOnMesh", &BinBasedFastPointLocatorConditions < 2 > ::FindPointOnMeshSimplified)
```
loumalouomega:
```suggestion
```
If unusable, we should simply remove it

kratos/python/add_geometrical_utilities_to_python.cpp
```diff
@@ -193,33 +192,57 @@ void AddGeometricalUtilitiesToPython(pybind11::module &m)
...
+            return std::tuple<bool,Vector,Element::Pointer>{is_found,N,p_elem};
+        })
         ;
 
     py::class_< BinBasedFastPointLocator < 3 >, BinBasedFastPointLocator < 3 >::Pointer >(m,"BinBasedFastPointLocator3D")
         .def(py::init<ModelPart&  >())
         .def("UpdateSearchDatabase", &BinBasedFastPointLocator < 3 > ::UpdateSearchDatabase)
-        .def("FindPointOnMesh", &BinBasedFastPointLocator < 3 > ::FindPointOnMeshSimplified)
         .def("UpdateSearchDatabaseAssignedSize", &BinBasedFastPointLocator < 3 > ::UpdateSearchDatabaseAssignedSize)
+        .def("FindPointOnMesh", &BinBasedFastPointLocator < 3 > ::FindPointOnMeshSimplified)
```
loumalouomega:
```suggestion
```
If unusable, we should simply remove it

kratos/python/add_geometrical_utilities_to_python.cpp
```diff
@@ -193,33 +192,57 @@ void AddGeometricalUtilitiesToPython(pybind11::module &m)
         .def("CreateNodesArray", &IsosurfacePrinterApplication::CreateNodesArray)
         ;
 
-    //binbased locators
+    // BinBasedFastPointLocator
     py::class_< BinBasedFastPointLocator < 2 >, BinBasedFastPointLocator < 2 >::Pointer >(m,"BinBasedFastPointLocator2D")
         .def(py::init<ModelPart& >())
         .def("UpdateSearchDatabase", &BinBasedFastPointLocator < 2 > ::UpdateSearchDatabase)
         .def("UpdateSearchDatabaseAssignedSize", &BinBasedFastPointLocator < 2 > ::UpdateSearchDatabaseAssignedSize)
         .def("FindPointOnMesh", &BinBasedFastPointLocator < 2 > ::FindPointOnMeshSimplified)
```
loumalouomega:
```suggestion
```
If unusable, we should simply remove it

loumalouomega (dismissed):

roigcarlo (approved):

PR #8142, opened 2021-01-24
[Fluid] Feature - QSVMS adjoints

sunethwarna:
**Description**
This PR adds adjoint formulation of QSVMS to Fluid App.

Requires #8255

**Changelog**
- Add `QSVMSAdjointElement`
- Change `adjoint_vms_monolithic_solver` to `adjoint_monolithic_solver`
- Add stabilization of `qs_vms` and `vms` for `AdjointMonolithicSolver`
- Add element level finite difference tests
- Add solver level finite difference tests
- Add mpi tests

PR #10215, opened 2022-09-02
[Core] Indirect Scalar Variable for Adjoints

sunethwarna:
**📝 Description**
This is a strange PR, and please consider the following reasoning behind the implementations.

In the adjoint formulations we have the following equation:
$$\underline{L}\left(\underline{w}\right) = \underline{\mathbf{M}}\left(\underline{w}\right)\cdot\underline{v}$$

And when we build the adjoint system matrix, we use the same implementation to the above equation where $$\underline{w} = \underline{w}^n$$ where $n$ represents current time step and $$\underline{w} = \underline{w}^{n+1}$$ where $n+1$ represents next time step (transient adjoints are solved reverse in time). What we do to have $\underline{L}\left(\underline{w}^{n+1}\right)$ is, we store it in a nodal variable after each time step solve so that can be used in the next solve time iteration. The problem is, some of the values in $\underline{L}\left(\underline{w}^{n+1}\right)$ are always zero (like the dof position corresponding to `PRESSURE`). Therefore, we do not reserve space for these since they are always zero. In order to have the generality in the adjoint schemes, currently we use `IndirectScalar` in the following way (conceptually).

we pass list of indirect scalars like the following in `AdjointExtensions`
```c++
std::vector<IndirectScalar> GetIndirectScalars()
{
return std::vector<IndirectScalar>{
     IndirectScalar(node, ADJOINT_VAR_1, Step),
     IndirectScalar(node, ADJOINT_VAR_2, Step),
     IndirectScalar()} // represents corresponding `PRESSURE` dof
}
// then we use it in another place like this

const auto& L = prod(M, v)
// the following line is called for each node seperately
auto indirect_scalars = GetIndirectScalars()
for (int i = 0; i < L.Size(); ++i)
{
     indirect_scalars[i] = L[i];
}
```

At the end, we update `ADJOINT_VAR_1` and `ADJOINT_VAR_2` in `node`, and disregarding `L[2]` update since it is zero, and it is never stored.

Then we use it again in `CalculateLocalSystem` like the following to construct the system matrix
```c++
// the following line is called for each node seperately
auto indirect_scalars = GetIndirectScalars()
for (int i = 0; i < L.Size(); ++i)
{
     RHS[i] = B * indirect_scalars[i];
}
```
`IndirectScalars` are bound to specific nodes, step, and variable always. Therefore, they are generated on the fly in `AdjointExtensions` class. Instead, I want to create `IndirectScalarVariable` class in this PR which is only bound to variable.

The use case is as follows then.
```c++
std::vector<IndirectScalarVariable> GetIndirectScalarVariables()
{
return std::vector<IndirectScalarVariable*>{
     &INDIRECT_ADJOINT_VAR_1,
     &INDIRECT_ADJOINT_VAR_2,
     &INDIRECT_SCALAR_ZERO} // represents corresponding `PRESSURE` dof
}
// then we use it in another place like this

const auto& L = prod(M, v)
// following line is called once per element
auto indirect_scalars = GetIndirectScalarVariables()
for node in geometry
{
    for (int i = 0; i < L.Size(); ++i)
    {
         (*indirect_scalars[i])(node, Step) = L[i];
    }
}
```
Then it can be used like the following
```c++
// following line is called once per element
auto indirect_scalars = GetIndirectScalarVariables()
for node in geometry
{
    for (int i = 0; i < L.Size(); ++i)
    {
         RHS[i] = B * (*indirect_scalars[i])(node, Step);
    }
}
```

Followings are taken into account:

1. This is required and used for `double` values, therefore `IndirectScalarVariable` only supports `double` (not required to template)
2. It works with `OpenMP` and `CXX SMP` parallelism
3. The `IndirectScalarVariable` is not copy or move constructible, and not assignable or movable as the `Variable`

If we are not  to use `IndirectScalar` or `IndirectScalarVariable`, then we won't be able to make a generic `ResidualBasedAdjointBossakScheme` which works with structural, fluid, rans elements because in the fluid and rans case, we have some rows in $\underline{\mathbf{M}}\cdot\underline{v}$ which are guranteed to be zero, and required to be stored so that they can be used in the next iteration. The naive option would be to create a variable for dofs corresponding to `PRESSURE` and reserve space, which is waste of memory or to re-implement specialized `ResidualBasedAdjointBossakScheme` for different elements in Fluid, Struct, RANS. Having different schemes is no go for coupled adjoint simulations such as RANS already, and is planned to expand coupling for Struc-Fluid-RANS adjoints. That is because we can solve adjoints for now using the monolithic construction only, and when we have monolithic adjoint coupled construction, these zero guranteed places can occure at different dof positions.

This PR is created only to introduce `IndirectScalarVariable` as a replacement for `IndirectScalar` which is already being used.

This PR can simplify implementations in #8142 a lot and will be easy to read as well.

**🆕 Changelog**
- Add `IndirectScalarVariable`
- Add macro definitions
- Add adjoint indirect variables
- Expose to python
- Add tests


kratos/tests/cpp_tests/containers/test_indirect_scalar_variables.cpp
```diff
@@ -0,0 +1,91 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:		 BSD License
+//                   Kratos default license: kratos/license.txt
```
rubenzorrilla, line 8:
```suggestion
//  License:		 BSD License
//                   Kratos default license: kratos/license.txt
```
Minor in the header formatting.

rubenzorrilla (commented):
I'm not negative about this addition. However I've a few comments:
1. Even though there's already a cpp test, as far as I understand this will be created from Python, so I'd add a small Python test as well.
2. Where the `GetIndirectScalarVariables()` will be implemented? And from where is called? I understand that this is not relevant for this specific PR but I think it is important to clarify the discussion at hand.
3. If we were creating an extra adjoint variable for the pressure, we would need this anymore right? However this would imply implementing the zero pressure contribution in the elements as well as the memory overhead. Am I wrong?

sunethwarna:
Hi Ruben,

> Even though there's already a cpp test, as far as I understand this will be created from Python, so I'd add a small Python test as well.

Will do. But this is mainly intended to be used in C++, python is also allowed.

> Where the GetIndirectScalarVariables() will be implemented? And from where is called? I understand that this is not relevant for this specific PR but I think it is important to clarify the discussion at hand.

At the moment, `GetIndirectScalarVariables` are implemented indirectly via `AdjointExtensions` in the core. But I want to remove the `AdjointExtensions` complication and use these `IndirectScalars` instead within adjoint `Schemes`. The elements will hold list of `IndirectScalars`

> If we were creating an extra adjoint variable for the pressure, we would **not** need this anymore right? However this would imply implementing the zero pressure contribution in the elements as well as the memory overhead. Am I wrong?

If we introduce adjoint variable for the pressure, then we don't need `IndirectZeroVariable` anymore. But It will be a waste of memory because we have to reserve space for a variable in adjoints which is always zero (with buffer size = 2). Adjoints already has a really huge memory impact because they require all the variables from primal, and additional variables from adjoint solve. Hence I would like to avoid reserving additional memory for this zero value.

The position of this adjoint pressure can change (when used with/without RANS adjoints) hence it is harder to indicate where the row of zeros will be found from matrix multiplications.

I changed the way I was creating these Indirect variables. Now, those variables are created on the fly. (but zero variable memory allocation is static, so it will behave like a singleton). So it will be easier to use, and less error prone.

I hope that I explained with enough details. :D

kratos/containers/indirect_scalar_variable.h
```diff
@@ -0,0 +1,218 @@
...
+    {
+        rOStream << Info();
+    }
+
+    ///@}
+    ///@name Operators
+    ///@{
+
+    inline double& operator()(NodeType& rNode) const
+    {
```
pooyan-dadvand, line 130: Maybe a naïve question, but why we need this version of the operator?

pooyan-dadvand:
@KratosMultiphysics/technical-committee would suggest to close this one in favor of substituting the B&S with scheme after moving to the native CSR implementation

PR #10216, opened 2022-09-02
Cosim/rigid body solver unittest

mtnzguillermo:
**📝 Description**
Adding unittests for the Rigid Body Solver

Only a couple of questions:
- Should the big tests run only for a few time steps so the size of the files is not that big?
- Can the file `test_rigid_body_solver.py` be reduced?


applications/CoSimulationApplication/tests/test_rigid_body_solver.py
```diff
@@ -0,0 +1,2047 @@
...
+             [   0.,    0.,    0.,  200.,    0.,    0.],
+             [   0.,    0.,    0.,    0., 1000.,    0.],
+             [   0.,    0.,    0.,    0.,    0.,  500.]]
+
+        for i, dof in enumerate(['displacement_x', 'displacement_y', 'displacement_z', 'rotation_x', 'rotation_y', 'rotation_z']):
+            self.assertEqual(simulation.is_constrained[dof], Constrains[dof])
+            for j in range(6):
+                self.assertEqual(simulation.M[i][j], M[i][j])
+                self.assertEqual(simulation.C[i][j], C[i][j])
+                self.assertEqual(simulation.K[i][j], K[i][j])
```
sunethwarna: you can avoid the last for loop and use `self.assertMatrixAlmostEqual(m1,m2, precision)`

sunethwarna: It is there for vectors as well

applications/CoSimulationApplication/tests/test_rigid_body_solver.py
```diff
@@ -0,0 +1,2047 @@
...
+              [    0. ,     0. , 20000.5,     0. ,     0. ,     0. ],
+              [    0. ,     0. ,     0. , 20000.5,     0. ,     0. ],
+              [    0. ,     0. ,     0. ,     0. , 20000.5,     0. ],
+              [    0. ,     0. ,     0. ,     0. ,     0. , 20000.5]]
+
+
+        for i in range(6):
+            for j in range(6):
+                self.assertAlmostEqual(simulation.LHS[i][j], LHS[i][j])
+
```
sunethwarna: can be simlplified

applications/CoSimulationApplication/tests/test_rigid_body_solver.py
```diff
@@ -0,0 +1,2047 @@
...
+              [    0. ,     0. , 2136682.06896552,     0. ,     0. ,     0. ],
+              [    0. ,     0. ,     0. , 43144.4137931,     0. ,     0. ],
+              [    0. ,     0. ,     0. ,     0. , 215722.06896552,     0. ],
+              [    0. ,     0. ,     0. ,     0. ,     0. , 21344.86206897]]
+
+
+        for i in range(6):
+            for j in range(6):
+                self.assertAlmostEqual(simulation.LHS[i][j], LHS[i][j])
+
```
sunethwarna: Can be simplified

applications/CoSimulationApplication/tests/test_rigid_body_solver.py
```diff
@@ -0,0 +1,2047 @@
...
+        
+        # Specific variables for the model part RootPoint
+        self.assertTrue(simulation.root_point_model_part.HasNodalSolutionStepVariable(KM.REACTION))
+        self.assertTrue(simulation.root_point_model_part.HasNodalSolutionStepVariable(KM.REACTION_MOMENT))
+        self.assertTrue(simulation.root_point_model_part.HasNodalSolutionStepVariable(KMC.PRESCRIBED_DISPLACEMENT))
+        self.assertTrue(simulation.root_point_model_part.HasNodalSolutionStepVariable(KMC.PRESCRIBED_ROTATION))
+
+
+    def test_SetCompleteVector_rigid_body(self):
+        simulation = RigidBodySolver(self.model, self.default_parameters)
```
sunethwarna: since you are using this a lot (in all the tests), you can moce it to `def setUp(self)` method

applications/CoSimulationApplication/tests/test_rigid_body_solver.py
```diff
@@ -0,0 +1,2047 @@
...
+
+        simulation.AdvanceInTime(simulation.main_model_part.ProcessInfo[KM.TIME])
+        simulation.InitializeSolutionStep()
+        simulation.Predict()
+        simulation.SolveSolutionStep()
+
+        reference = [14.47730397, -954.52993411, 30.97172112, 56.62280302, -20.61982872, 98.25390568]
+        simulation_effective_load = simulation._GetCompleteVector("rigid_body", KMC.EFFECTIVE_FORCE, KMC.EFFECTIVE_MOMENT)
+
+        self.assertVectorAlmostEqual(reference, simulation_effective_load)
```
sunethwarna: As far as I see, you are running a whole simulation here ryt? I would create the json file seperately and run it as a normal simulation rather than running it with a skeleton like simulation setup.

applications/CoSimulationApplication/tests/test_rigid_body_solver.py
```diff
@@ -0,0 +1,2047 @@
...
+                "load_restart_files_from_folder": true,
+                "restart_load_file_label": "0.0"
+            },
+            "time_integration_parameters": {
+                "rho_inf": 0.16,
+                "time_step": 0.01
+            }
+        }''')
+        
+        self.assertEqual(ref_solver_settings.PrettyPrintJsonString(), checked_solver_settings.PrettyPrintJsonString())
```
sunethwarna: Usually we don't have to test for Validate and Assign, because on the long run, you may add new settings which does not change the old ones therefore extending features. Which will breeak this test. Even if you want to check, then you can use the following 

```python
self.assertTrue(ref_solver_settings.IsEquivalentTo(checked_solver_settings))
```

sunethwarna: don't use `PrettyPringJsonString()`

applications/CoSimulationApplication/tests/test_rigid_body_solver.py
```diff
@@ -0,0 +1,2047 @@
...
+                }
+            },
+            "output_processes": [
+            ],
+            "processes": {
+            }
+        }''')
+        solver_settings = Parameters["solver_settings"]
+        
+        self.assertRaises(Exception, input_check._ValidateAndAssignRigidBodySolverDefaults, solver_settings)
```
sunethwarna: I would try to capture the exact exception in here like 
```python
self.assertRaises(Exception("The error given in the place"), input_check._ValidateAndAssignRigidBodySolverDefaults, solver_settings)
```

sunethwarna (changes requested):
Nice work. It is ok to have `test_rigid_body_solver.py` large, but it can be still simplified a little bit. I have small comments which i did inplace, and overall following concerns.

1. You have some binary and `.cnd` files. please remove them from the PR
2. The `Barc.mdpa` is too big. You have to make it coarset as much as possible. It would be nice if we can coarsen it to max 500 nodes, and compare the results with some ref json file.
3. For sure there are lots of time steps in the barc case. Checking for 10 time steps would be fine with higher tolerance.

applications/CoSimulationApplication/tests/test_rigid_body_solver.py
andimkatili: I changed it into assertRaisesRegex with the exact exception

applications/CoSimulationApplication/tests/test_rigid_body_solver.py
andimkatili: I will revise this test

applications/CoSimulationApplication/tests/test_rigid_body_solver.py
andimkatili: I was unable to use assertMatrixAlmostEqual because the matrix that it evaluates uses numpy.ndarray and it doesn't have Size1 and Size2 in its attribute. But it is possible to use just one loop and use assertVectorAlmostEqual instead of assertMatrixAlmostEqual.

sunethwarna: In that case, I would keep it like this.

applications/CoSimulationApplication/tests/test_rigid_body_solver.py
andimkatili: I replaced PrettyPringJsonString() with IsEquivalentTo() for now. Should I remove this test instead?

rubenzorrilla:
Again, why the rigid body solver capability is not being added to the Core? Is there any technical limitation?

PR #10221, opened 2022-09-05
[CMake] Fix bug with compiler flags

loumalouomega:
**📝 Description**

Fix bug with compiler flags. When LTO linker is considered, the compiler flags are overwritten

**🆕 Changelog**

- [Fix bug with flags](https://github.com/KratosMultiphysics/Kratos/commit/489a0363b84e16d470fd5aac556bac6b27483e9f)


roigcarlo (approved):
Approve but LTO is in gray status, we made changes to the pybind linkage options as well. I am not sure if it is currently working (just to warn you if you are using this options, changes are ok)

RiccardoRossi:
lto was problematic with the gid output @maceligueta can u comment?

maceligueta:
Well, I remember that we had like two different gidpost libraries linked when we activated LTO, so it did weird things. I remember this was noticed when running pfem cases (PfemFluidApplication by @AFranci ). So, if we want to try again, we should try one of his examples in debug and release versions (they were behaving differently).

PR #10224, opened 2022-09-05
[Core] Adding KRATOS_ARCHITECTURE_EXTENSION to easily define the architecture extensions

loumalouomega:
**📝 Description**

Adding KRATOS_ARCHITECTURE_EXTENSION to easily define the architecture extensions. I added macros to easily define the architecture extensions of x86_64. Adding to the default scripts the "native" option, which for *Clang/LLVM* and *Gcc* automatically detects the best architecture for your machine. Also, there is a macro which will determine which extension will consider the "native" option. This will provide a performance boost in developers compilations.

In order to use:

~~~sh
-DKRATOS_ARCHITECTURE_EXTENSION="native" \
~~~

Or a specific extension:

~~~sh
-DKRATOS_ARCHITECTURE_EXTENSION="avx2" \
~~~

Includes #10221

**🆕 Changelog**

- Adding KRATOS_ARCHITECTURE_EXTENSION to easily define the architecture extensions
- Adding extension to Kratos Version
- Update standard configures


roigcarlo (changes requested):
Let us discuss in @KratosMultiphysics/technical-committee

philbucher:
How did you test this?

Are there measurable performance gains?

loumalouomega:
> How did you test this?
> 
> Are there measurable performance gains?

Unfortunately for now will be limited effect because the compiler cannot determine many SIMD operations automatically (it can, but limited) and our main linear algebra library (where the SIMD heavy lifting is) UBlas has poor support for SIMD. I assume at some point we will fully migrate to Eigen (http://eigen.tuxfamily.org/index.php?title=FAQ see Vectorization) or Blaze 
(https://bitbucket.org/blaze-lib/blaze/src/master/ vectorization is the second point in the README). In any case adding support is relevant and in the future if we do the proper changes will help improve performance (AVX2 is already X86-64 quite standard, and even if AVX512 is available is x2 for algebra operations) significantly

loumalouomega:
FYI @jrubiogonzalez

PR #10225, opened 2022-09-05
[CoSimulationApplication] Allowing direct parameters instead of input_file

loumalouomega:
**📝 Description**

Allowing direct parameters instead of input_file. This will allow defining everything in one file (or using the future includes). This is because in @KratosMultiphysics/altair we like to have all the settings in one Parameter object that at the end of the simulation is exported (with all the missing parameters added), so we can know exactly what consisted the whole input.

**🆕 Changelog**

- [Allowing direct parameters instead of input_file](https://github.com/KratosMultiphysics/Kratos/commit/3f355780661d4059a8adf38e387b827796076c40)


philbucher (approved):
:+1: 

as discussed with @pooyan-dadvand

PR #10238, opened 2022-09-07
[Core] Move ExecuteBeforeSolutionLoop just before executing solution loop in base analysis

loumalouomega:
**📝 Description**

As in #10217, this PR is a minor change in the base analysis class in order to unify what we do in @KratosMultiphysics/altair. A priori with this change there should not be any significant change. 

**🆕 Changelog**

- Move ExecuteBeforeSolutionLoop just before executing solution loop in base analysis

### DIFF
--- a/applications/CoSimulationApplication/python_scripts/co_simulation_analysis.py
+++ b/applications/CoSimulationApplication/python_scripts/co_simulation_analysis.py
@@ -53,6 +53,9 @@ class CoSimulationAnalysis(AnalysisStage):
         else:
             # flush by default only in OpenMP, can decrease performance in MPI
             self.flush_stdout = (self.parallel_type == "OpenMP")
+        
+        # Define a dummy list of processes, since the baseclass expects it
+        self._list_of_processes = []
 
         self._GetSolver() # this creates the solver
 
--- a/kratos/python_scripts/analysis_stage.py
+++ b/kratos/python_scripts/analysis_stage.py
@@ -58,6 +58,11 @@ class AnalysisStage(object):
         """This function executes the solution loop of the AnalysisStage
         It can be overridden by derived classes
         """
+        # We execute the processes before the solution loop
+        for process in self._GetListOfProcesses():
+            process.ExecuteBeforeSolutionLoop()
+
+        # Now we execute the solution loop
         while self.KeepAdvancingSolutionLoop():
             self.time = self._GetSolver().AdvanceInTime(self.time)
             self.InitializeSolutionStep()
@@ -94,9 +99,6 @@ class AnalysisStage(object):
         self.Check()
 
         self.ModifyAfterSolverInitialize()
-
-        for process in self._GetListOfProcesses():
-            process.ExecuteBeforeSolutionLoop()
 
         ## Stepping and time settings
         self.end_time = self.project_parameters["problem_data"]["end_time"].GetDouble()


rubenzorrilla (changes requested):
Change might have sense. However it might break custom analysis stages deriving from the base one. We should be careful with this. To be discussed in @KratosMultiphysics/technical-committee.

miguelmaso:
I like this change because the `initial_time` is set before ExecuteBeforeSolutionLoop

On the other hand, I feel like some tests are failing due to the same reason than in #10234

loumalouomega:
The problem is in the CoSimulation. This can be solved or creating a dummy list of processes in the Cosimulation analysis or overriding the RunSolutionLoop. I will do the first, so if in the future we want to incorporate processes in the CoSimulation. Opinion @philbucher

rubenzorrilla:
> The problem is in the CoSimulation. This can be solved or creating a dummy list of processes in the Cosimulation analysis or overriding the RunSolutionLoop. I will do the first, so if in the future we want to incorporate processes in the CoSimulation. Opinion @philbucher

Can you elaborate a bit? As far as I understand co-simulation assumes current process execution order.

loumalouomega:
> > The problem is in the CoSimulation. This can be solved or creating a dummy list of processes in the Cosimulation analysis or overriding the RunSolutionLoop. I will do the first, so if in the future we want to incorporate processes in the CoSimulation. Opinion @philbucher
> 
> Can you elaborate a bit? As far as I understand co-simulation assumes current process execution order.

CoSimulation analysis does not have processes list (the sub-analysis have, but not the main CoSimulation analysis)

rubenzorrilla:
> > > The problem is in the CoSimulation. This can be solved or creating a dummy list of processes in the Cosimulation analysis or overriding the RunSolutionLoop. I will do the first, so if in the future we want to incorporate processes in the CoSimulation. Opinion @philbucher
> > 
> > 
> > Can you elaborate a bit? As far as I understand co-simulation assumes current process execution order.
> 
> CoSimulation analysis does not have processes list (the sub-analysis have, but not the main CoSimulation analysis)

And how this turns into a problem?

philbucher:
This is complex, I will answer later

But I don't like changing such basic things since we cannot know what will break

loumalouomega:
> > > > The problem is in the CoSimulation. This can be solved or creating a dummy list of processes in the Cosimulation analysis or overriding the RunSolutionLoop. I will do the first, so if in the future we want to incorporate processes in the CoSimulation. Opinion @philbucher
> > > 
> > > 
> > > Can you elaborate a bit? As far as I understand co-simulation assumes current process execution order.
> > 
> > 
> > CoSimulation analysis does not have processes list (the sub-analysis have, but not the main CoSimulation analysis)
> 
> And how this turns into a problem?

Because the method which retrieves the list of processes has an exception implemented if not initialized

miguelmaso:
In addition to CoSimulation and GeomMechanics, there is a failing test in the Core:
`testInitialTime` This test always tell the truth, it has two independent simulations which must reproduce the same results regardless of the `initial_time`

loumalouomega:
> In addition to CoSimulation and GeomMechanics, there is a failing test in the Core: `testInitialTime` This test always tell the truth, it has two independent simulations which must reproduce the same results regardless of the `initial_time`

Interesting. @KratosMultiphysics/altair do we have any strong reason to move the block  ExecuteBeforeSolutionLoop?

miguelmaso:
But knowing why the test is failing is interesting too. I like your proposal but there is a design issue which we should address

ddiezrod:
The failure in test_initial_time exposes one problem in the current implementation. TIME is not correctly set at the moment we call ExecuteBeforeSolutionLoop which can be quite problematic as it is used for example in assign_scalar_variable_process to check if we are inside specified interval time.

miguelmaso:
> The failure in test_initial_time exposes one problem in the current implementation. TIME is not correctly set at the moment we call ExecuteBeforeSolutionLoop which can be quite problematic as it is used for example in assign_scalar_variable_process to check if we are inside specified interval time.

Agree

rubenzorrilla:
> The failure in test_initial_time exposes one problem in the current implementation. TIME is not correctly set at the moment we call ExecuteBeforeSolutionLoop which can be quite problematic as it is used for example in assign_scalar_variable_process to check if we are inside specified interval time.

The point on this is that now we set `TIME` when doing the very first `CloneSolutionStep`, which occurs in the first iteration of the run solution loop. Most probably, we should complementary set the `TIME` to the initial time in the `problem_data` when filling the buffer.

(somehow related to #9592).

miguelmaso:
> > The failure in test_initial_time exposes one problem in the current implementation. TIME is not correctly set at the moment we call ExecuteBeforeSolutionLoop which can be quite problematic as it is used for example in assign_scalar_variable_process to check if we are inside specified interval time.
> 
> The point on this is that now we set `TIME` when doing the very first `CloneSolutionStep`, which occurs in the first iteration of the run solution loop. Most probably, we should complementary set the `TIME` to the initial time in the `problem_data` when filling the buffer.
> 
> (somehow related to #9592).

This approach is wrong (the time step may be wrongly computed) and has been fixed in #9267. Furthermore, the test introduced in #9267 (`testInitialTime`) would raise an error if the time is set at the very first `CloneSolutionStep`.

@rubenzorrilla I strongly recommend to implement an analogous `testInitialTime` in FluidDynamicsApp

rubenzorrilla:
> > > The failure in test_initial_time exposes one problem in the current implementation. TIME is not correctly set at the moment we call ExecuteBeforeSolutionLoop which can be quite problematic as it is used for example in assign_scalar_variable_process to check if we are inside specified interval time.
> > 
> > 
> > The point on this is that now we set `TIME` when doing the very first `CloneSolutionStep`, which occurs in the first iteration of the run solution loop. Most probably, we should complementary set the `TIME` to the initial time in the `problem_data` when filling the buffer.
> > (somehow related to #9592).
> 
> This approach is wrong (the time step may be wrongly computed) and has been fixed in #9267. Furthermore, the test introduced in #9267 (`testInitialTime`) would raise an error if the time is set at the very first `CloneSolutionStep`.
> 
> What @ddiezrod says is true (see #9645).
> 
> @rubenzorrilla I strongly recommend to implement an analogous `testInitialTime` in FluidDynamicsApp before merging #9592



> > > The failure in test_initial_time exposes one problem in the current implementation. TIME is not correctly set at the moment we call ExecuteBeforeSolutionLoop which can be quite problematic as it is used for example in assign_scalar_variable_process to check if we are inside specified interval time.
> > 
> > 
> > The point on this is that now we set `TIME` when doing the very first `CloneSolutionStep`, which occurs in the first iteration of the run solution loop. Most probably, we should complementary set the `TIME` to the initial time in the `problem_data` when filling the buffer.
> > (somehow related to #9592).
> 
> This approach is wrong (the time step may be wrongly computed) and has been fixed in #9267. Furthermore, the test introduced in #9267 (`testInitialTime`) would raise an error if the time is set at the very first `CloneSolutionStep`.
> 
> What @ddiezrod says is true (see #9645).

I said this by heart and I didn't remember #9267. It is true that `TIME` is already there when the first `CloneSolutionStep` is performed. I also agree that current approach doesn't solve the issue as you both @ddiezrod and @miguelmaso pointed out.

Having said so, forget about my previous comment.

> 
> @rubenzorrilla I strongly recommend to implement an analogous `testInitialTime` in FluidDynamicsApp before merging #9592

About this, #9592 shouldn't IMO be affected by this, since what we are basically doing is to go backwards to go forward, instead of directly going forward.

miguelmaso:
> > @rubenzorrilla I strongly recommend to implement an analogous `testInitialTime` in FluidDynamicsApp before merging #9592
> 
> About this, #9592 shouldn't IMO be affected by this, since what we are basically doing is to go backwards to go forward, instead of directly going forward.

The computation of the first and old TIME_STEP is something crucial and it may be affected by going backwards and forward. Hopefully, https://github.com/KratosMultiphysics/Kratos/pull/9592 shouldn't be affected by this, just to be on the safe side

philbucher:
I gave this some thought and I am not sure what to think abt it.

On one hand I agree that if it fixes a but (regarding `TIME`, I didn't read exactly the other posts) then it is great

However, as was pointed out, this might break some existing code that is hard to know. Changing the order in which the processes are executed is very dangerous IMO (there are even some failing tests hence not sure how to proceed)

Regarding CoSim, @loumalouomega did you run all the tests? Looking at the error I am afraid that this might break more tests. Figuring out the call order is one of the nastiest problem in CoSim, hence this must be checked carefully

Overall I don't have a strong opinion, but adapting the base for a special app (CoSim) at the cost of potentially breaking stuff is dangerous

I will also comment in #10242 regarding the treatment of CoSim

loumalouomega:
> Regarding CoSim, @loumalouomega did you run all the tests? Looking at the error I am afraid that this might break more tests. Figuring out the call order is one of the nastiest problem in CoSim, hence this must be checked carefully

The idea of this PR is to discuss first toward a unification. I will not try to fix failing tests until consensus.

loumalouomega:
I will try to unify from our side, closing

loumalouomega:
(Keeping the branch for now)

PR #10242, opened 2022-09-08
[CoSimulationApplication] Adding custom base class to coupling solvers

loumalouomega:
**📝 Description**

Please @philbucher don't kill me. In order to use the CoSimulationApplication in the internal implementations of @KratosMultiphysics/altair I need to be able to select a custom base class in the coupling solvers, as our strategies differ from the standard one (see #10238 and #10217, and many more, unify this will take forever), and therefore we are not able to use the CoSimulationApplication without this change.

**🆕 Changelog**

- [Adding custom base class to coupling solvers](https://github.com/KratosMultiphysics/Kratos/commit/94312e4cfc3252c1610d4f0f625340f7bcea9279)
- [Improve the solver_wrapper_factory to import custom base classes](https://github.com/KratosMultiphysics/Kratos/commit/ef2f7b2b99d0261366ba4bcd82b2b93a56038fec)

### DIFF
--- a/applications/CoSimulationApplication/python_scripts/base_classes/co_simulation_coupled_solver.py
+++ b/applications/CoSimulationApplication/python_scripts/base_classes/co_simulation_coupled_solver.py
@@ -14,6 +14,8 @@ from KratosMultiphysics.CoSimulationApplication.coupling_interface_data import B
 # Other imports
 from collections import OrderedDict
 
+def Create(settings = None, models = None, solver_name = None, base_type = None):
+    return CoSimulationCoupledSolver(settings, models, solver_name)
 
 class UndefinedSolver:
     def __init__(self, name, settings):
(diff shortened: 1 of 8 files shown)


rubenzorrilla (changes requested):
We (together with @RiccardoRossi) had a short discussion about this. We decided to discuss it first at the @KratosMultiphysics/technical-committee. We would appreciate @philbucher thoughts for it.

loumalouomega:
> We (together with @RiccardoRossi) had a short discussion about this. We decided to discuss it first at the @KratosMultiphysics/technical-committee. We would appreciate @philbucher thoughts for it.

Any case I will fix first the failing tests

philbucher:
Hm 😅

I will habe a closer look in the next days

How many of those big changes are you planning to do? 😅

loumalouomega:
> How many of those big changes are you planning to do? 😅

I wish I knew it...

philbucher:
First of all I am no longer using Kratos professionally and currently only to keep up with some C++ skills and because I like it / have fun maintaining it. Hence I will not block any changes. 
Therefore the following is my personal opinion and any decisions how to proceed should be taken by active members of @KratosMultiphysics/cosimulation and @KratosMultiphysics/technical-committee 

I looked at the changes and I am not sure what the intention is. To me it looks like a lot of obfuscation of an already complicated structure/implementation. CoSimulation was and is a difficult subject when it comes to the orchestration/integration of the solvers involved in it. The structure that is currently used might seem simple/straight forward, but it took a lot of work by many ppl to arrive to the current state.

Due to its nature it has always been a bit different compared to the other Kratos Apps. That mostly being that it uses the other Apps as black-box, which implies that each of the solvers is coupled on the `AnalysisStage` level and not the solver level, which includes the solvers having their own set of `ProjectParameters` as well as their own `Model`
Note that Optimization also uses the Kratos solvers as black-box and follows a similar approach

Without knowing the implementation of @KratosMultiphysics/altair, could you describe the problem a bit more that you are facing when using CoSim? Perhaps then I can suggest some alternative solutions (same for #10238)

Also this brings up a philosophical question, should Kratos adapt to the needs of Altair at the cost of other users (e.g. due to the proposed changes making further developments more complicated for other users) or the other way round? I perfectly understand that you want to unify things to make development easier and smoother, but maybe it can be done in a better way that benefits everyone. So far I think we managed that quite well

loumalouomega:
> I looked at the changes and I am not sure what the intention is. To me it looks like a lot of obfuscation of an already complicated structure/implementation. CoSimulation was and is a difficult subject when it comes to the orchestration/integration of the solvers involved in it. The structure that is currently used might seem simple/straight forward, but it took a lot of work by many ppl to arrive to the current state.

First, I am not saying the implementation is wrong. All the contrary. We can adapt the code from @KratosMultiphysics/altair to be CoSim compatible:

- Copy and paste whole application and modify what is necessary. This will make the code incompatible in  a very short time.
- Reach a consensus of the changing classes between the FOSS and our internal version. This will take a very long time and discussion. So the middle solution is to modify certain classes to inject modifications without changing everything. This is the middle solution, and it is not forever, it is a patch until the analysis is unified, but we need these changes in order to have a functional version before that.

> Due to its nature it has always been a bit different compared to the other Kratos Apps. That mostly being that it uses the other Apps as black-box, which implies that each of the solvers is coupled on the `AnalysisStage` level and not the solver level, which includes the solvers having their own set of `ProjectParameters` as well as their own `Model` Note that Optimization also uses the Kratos solvers as black-box and follows a similar approach

I know, read my previous comment.

> Without knowing the implementation of @KratosMultiphysics/altair, could you describe the problem a bit more that you are facing when using CoSim? Perhaps then I can suggest some alternative solutions (same for #10238)

I can assure that without tell you confidential information, this is the best solution.

> Also this brings up a philosophical question, should Kratos adapt to the needs of Altair at the cost of other users (e.g. due to the proposed changes making further developments more complicated for other users) or the other way round? I perfectly understand that you want to unify things to make development easier and smoother, but maybe it can be done in a better way that benefits everyone. So far I think we managed that quite well

I am not forcing that, this is a patch until a consensus is reached. Furthermore, I can assure you that people like @ddiezrod are desiring to unify, but deadlines are now priority. We have deadlines and cannot wait to that. So this solution is the simplest one. 100% retro-compatible and the one which will require the minimal amount of changes.

mpentek (changes requested):
Blocking for now to prevent an accidental merge.

I suggest a community discussion as certain aspects of this PR raise some concerns. These are not related to the technical solution, but the argumentation.

As Kratos Multiphysics is BSD open source and many universities are involved, I think we need to have an opporunity to reflect on this and see what types of reasoning and deadlines are appropriate. In my opinion not being able to disclose the reasons and further steps which might be taken for development does not fit the scope and license of the project. Also, we all have deadlines.

loumalouomega:
> Blocking for now to prevent an accidental merge.
> 
> I suggest a community discussion as certain aspects of this PR raise some concerns. These are not related to the technical solution, but the argumentation.
> 
> As Kratos Multiphysics is BSD open source and many universities are involved, I think we need to have an opporunity to reflect on this and see what types of reasoning and deadlines are appropriate. In my opinion not being able to disclose the reasons and further steps which might be taken for development does not fit the scope and license of the project. Also, we all have deadlines.

As I said we are looking for a solution from our side. I will close this (keeping the branch) to avoid confusions

mpentek:
> Blocking for now to prevent an accidental merge.
> 
> I suggest a community discussion as certain aspects of this PR raise some concerns. These are not related to the technical solution, but the argumentation.
> 
> As Kratos Multiphysics is BSD open source and many universities are involved, I think we need to have an opporunity to reflect on this and see what types of reasoning and deadlines are appropriate. In my opinion not being able to disclose the reasons and further steps which might be taken for development does not fit the scope and license of the project. Also, we all have deadlines.

That is appreciated!

loumalouomega:
> > Blocking for now to prevent an accidental merge.
> > I suggest a community discussion as certain aspects of this PR raise some concerns. These are not related to the technical solution, but the argumentation.
> > As Kratos Multiphysics is BSD open source and many universities are involved, I think we need to have an opporunity to reflect on this and see what types of reasoning and deadlines are appropriate. In my opinion not being able to disclose the reasons and further steps which might be taken for development does not fit the scope and license of the project. Also, we all have deadlines.
> 
> That is appreciated!

I was able today of finally fix the issue from our side after many patches

loumalouomega:
Deleting branch, I have a local copy

Issue #7544, opened 2020-09-30
How to set initial conditions?


ddiezrod:
I was discussing today with @mohittyagi-k how to set initial conditions. If I'm not wrong, surprisingly there is no standard way to impose them using processes. The trick we used was applying "interval" : [0.0 , 0.0] in the "assign_.._process.py". This is working right now but to do comparisons between doubles does not seem very safe and also looks quite dirty for such a basic thing as setting initial conditions. I think using "interval" : [0.0 , 1e-10] is not much better neither... 

Maybe there is another process to do this that we did not see? If not, I think we should have a common solution for this. though it might be a little painful considering how the number of "assign" processes grew.

@KratosMultiphysics/altair @KratosMultiphysics/design-committee @KratosMultiphysics/interface-committee

loumalouomega:
We can add to the processes the flag `initial_condition`, false by default, that will obviate the interval and applied only in the `BeforeSolutionLoop`

ddiezrod:
Well, it seems @vdivine found the solution. There is a ApplyConstantScalar/VectorProcess that does what we need, calling Execute only in Initialize. Its in cpp so that's why we forgot about it. I think we will use that one if it's not deprecated, sorry for the noise. Hopefully, this issue will help others in the future :)

rubenzorrilla:
We had the same discussion not long ago. I see two possible straighforward solutions:
1. what @loumalouomega suggests
2. from each one of the main processes derive it's corresponding initial condition version that only implements the `ExecuteInitialize` and `ExecuteBeforeSolutionLoop` methods.

Another thing that comes to my mind is to create a new `initial_process` class. This class would have the same API as the `process` one but all the methods but the `ExecuteInitialize` and the `ExecuteBeforeSolutionLoop` must be non-virtual so they can be called using the standard process list but enforcing an empty implementation. Then, in the derived classes (e.g. assign_scalar_value_process_initial) we can make the standard processes (e.g. assign_scalar_value_process) friends of the initial classes, so we can reuse the standard processes implementation.

mohittyagi-k:
Also, (in addition to what @ddiezrod mentioned), I have the following doubt. 

Shouldn't the following two processes have the same behavior, since the `Parameters` are identical (assuming we assume `constrained`  would imply the same thing as `is_fixed`, which may not be the case):

Process 1

           "kratos_module": "KratosMultiphysics",
            "python_module": "assign_scalar_variable_process",
[...]

PR #10243, opened 2022-09-08
[IntervalUtility] Time comparison

miguelmaso:
**📝 Description**
This fixes a test in #10238 (`testInitialTime`). I haven't pushed directly to #10238 because this should be discussed in another PR.

If this PR is rejected, a better solution for applying initial conditions should be found (see #7544)

**🆕 Changelog**
- Update `double` comparison

### DIFF
--- a/kratos/tests/test_interval_utility.py
+++ b/kratos/tests/test_interval_utility.py
@@ -1,0 +1,39 @@
+import KratosMultiphysics
+import KratosMultiphysics.KratosUnittest as KratosUnittest
+
+class TestInitialTime(KratosUnittest.TestCase):
+
+    def testIntervalUtilityDefault(self):
+        settings = KratosMultiphysics.Parameters()
+
+        # Default interval [0, 1e30]
+        interval = KratosMultiphysics.IntervalUtility(settings)
+        self.assertTrue(interval.IsInInterval(0.0))
+
+        # Initial interval [0, 0]
+        settings["interval"][0].SetDouble(0.0)
+        settings["interval"][1].SetDouble(0.0)
+        interval = KratosMultiphysics.IntervalUtility(settings)
+        self.assertTrue(interval.IsInInterval(0.0))
+
+        # Initial interval [1000, 1000]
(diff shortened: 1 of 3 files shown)


loumalouomega (dismissed):
Okay from my side. TODO: move this piece of code to cpp

miguelmaso:
@loumalouomega , the float comparison is fixed and a test is added. Ready for review

loumalouomega (approved):
Okay, thanks

matekelemen:
I've updated ```IntervalUtility``` in [this branch](https://github.com/KratosMultiphysics/Kratos/tree/core/interval-utility-2) but I couldn't open a PR for it yet because it relies on #10052, which has been stuck in *core PR review limbo* for ages. That branch fixes this issue, as well as adds the possibility to use it with integers.

I'd appreciate it if I didn't have to completely rewrite the stuff I've done, but I'm open to suggestions.

miguelmaso:
Thanks @matekelemen , your [branch](https://github.com/KratosMultiphysics/Kratos/tree/core/interval-utility-2) is very nice.

The aim of this PR is to fix a float comparison, which produces wrong results in non-zero initial times.

What do you prefer, merge this PR or close it and wait for #10052 and [branch](https://github.com/KratosMultiphysics/Kratos/tree/core/interval-utility-2)?

matekelemen:
@miguelmaso well, if another PR depends on this one, I won't hold you back; who knows when mine will get reviewed.

miguelmaso:
@matekelemen sorry for the delay

#10238 depends on this, but is closed. So, this can also be closed

If possible, add a test like this (if I'm not wrong, it fails with the current master)
```py
    def test_LargeBoundaries(self) -> None:
        for begin, end in ((0.0, 1000.0), (1000.0, 1000.0)):
```
or even, with larger values...

miguelmaso:
To avoid messing up this [branch](https://github.com/KratosMultiphysics/Kratos/tree/core/interval-utility-2)

PR #10246, opened 2022-09-08
[Core] cleaning some includes

philbucher:
see title


kratos/sources/cfd_variables.cpp
```diff
@@ -11,16 +11,13 @@
...
-#include <string>
-#include <iostream>
-#include <vector>
 
 // External includes
 
 // Project includes
 #include "includes/define.h"
 #include "includes/cfd_variables.h"
-#include "includes/kernel.h"
```
roigcarlo, line 23: This one is very abused indeed, and it even could go completely off if the variables were registered in the kratos_application source.

roigcarlo (dismissed):

loumalouomega (approved):
Thanks

PR #10252, opened 2022-09-09
[Poro] Fixed 2D interface element rotation matrix and other small changes

ipouplana:
**📝 Description**
We detected a wrong sign in some components of the 2D interface element rotation matrix. Now we build such a matrix assuming that the nodes in the 2D interface elements are ordered clockwise (as GiD does). We also added the CONTACT_STRESS_VECTOR variable with the contact stresses in the interface written for the global coordinate system. We also fixed the result shown in the post-processing of the variable JOINT_WIDTH.

**🆕 Changelog**
- Fixed the sign of some components of the rotation matrix for the 2D interface elements
- Fixed the post-processing of the JOINT_WIDTH variable
- Added CONTACT_STRESS_VECTOR in the interface elements


ipouplana:
@joaquinirazabal I also changed the sign in the rotation matrix of the 2D solid interface element of the DamApp. Please, if you have some time, take a look at it and decide whether you want to implement this change or not.

JoseMGonzalez-UPC (approved):

PR #10253, opened 2022-09-09
[dem] First 1 to 4 Chung Ooi tests update

salvalatorre:
Adjusting the older Chung-Ooi tests to newer code. Only the first 4 tests (out of 8).


maceligueta:
My only comment is related to all those Codacy complaints: we should avoid `import *` as much as possible. Using star imports makes the code less readable (it is not easy to know from which import one variable became available). All the new work shoud follow this recommendation, regardless of what 'the old code does'.
The PR is nice for the rest. Thanks @salvalatorre!!

maceligueta (dismissed):

salvalatorre:
Please, re-review after modifications due to merging conflicts...

salvalatorre:
@maceligueta

maceligueta (approved):

PR #10256, opened 2022-09-12
[CMake] Using option to simplify code

loumalouomega:
**📝 Description**

Using option to simplify code

**🆕 Changelog**

- [Using option to simplify code](https://github.com/KratosMultiphysics/Kratos/commit/9daa3d0ac96017c7421d1a224042bfcbc2191d30)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,7 +51,7 @@ endif(NOT DEFINED ENV{KRATOS_INSTALL_PYTHON_USING_LINKS})
 set(CMAKE_INSTALL_RPATH "$ORIGIN/../libs")
 
 # If no test policy enable by default
-option(KRATOS_BUILD_TESTING "KRATOS_BUILD_TESTING defines if the C++ tests are compiled. These increases compilation time, but if not set only python tests can be executed. Default setting is ON" ON)
+option(KRATOS_BUILD_TESTING "KRATOS_BUILD_TESTING defines if the C++ tests are compiled. These increase compilation time, but if not set only python tests can be executed. Default setting is ON" ON)
 
 # If not defined mute the very verbose installation messages
 if(NOT DEFINED CMAKE_INSTALL_MESSAGE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,7 +51,7 @@ endif(NOT DEFINED ENV{KRATOS_INSTALL_PYTHON_USING_LINKS})
 set(CMAKE_INSTALL_RPATH "$ORIGIN/../libs")
 
 # If no test policy enable by default
-option(KRATOS_BUILD_TESTING "-- No KRATOS_BUILD_TESTING is defined. Setting to ON" ON)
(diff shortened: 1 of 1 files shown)


roigcarlo (dismissed):

CMakeLists.txt
```diff
@@ -51,15 +51,12 @@ endif(NOT DEFINED ENV{KRATOS_INSTALL_PYTHON_USING_LINKS})
 set(CMAKE_INSTALL_RPATH "$ORIGIN/../libs")
 
 # If no test policy enable by default
-if(NOT DEFINED KRATOS_BUILD_TESTING)
-message("-- No KRATOS_BUILD_TESTING is defined. Setting to ON")
-SET (KRATOS_BUILD_TESTING ON)
-endif(NOT DEFINED KRATOS_BUILD_TESTING)
+option(KRATOS_BUILD_TESTING "-- No KRATOS_BUILD_TESTING is defined. Setting to ON" ON)
```
philbucher: This is not the same, the string is the name of the option, but it will now no longer issue the message

If you use it then I would improve the message

loumalouomega: In the doc only speaks about <help_text>, I don't know how it was defined

loumalouomega:
```suggestion
option(KRATOS_BUILD_TESTING "KRATOS_BUILD_TESTING defines if the C++ tests are compiled. These increases compilation time, but if not set only python tests can be executed. Default setting is ON" ON)
```
@philbucher better like this?

roigcarlo: @philbucher Oh you are right

CMakeLists.txt
```diff
@@ -51,15 +51,12 @@ endif(NOT DEFINED ENV{KRATOS_INSTALL_PYTHON_USING_LINKS})
 set(CMAKE_INSTALL_RPATH "$ORIGIN/../libs")
 
 # If no test policy enable by default
-if(NOT DEFINED KRATOS_BUILD_TESTING)
-message("-- No KRATOS_BUILD_TESTING is defined. Setting to ON")
-SET (KRATOS_BUILD_TESTING ON)
-endif(NOT DEFINED KRATOS_BUILD_TESTING)
+option(KRATOS_BUILD_TESTING "KRATOS_BUILD_TESTING defines if the C++ tests are compiled. These increases compilation time, but if not set only python tests can be executed. Default setting is ON" ON)
```
philbucher: Can you please add the message back?
I think it is important

rubenzorrilla:
```suggestion
option(KRATOS_BUILD_TESTING "KRATOS_BUILD_TESTING defines if the C++ tests are compiled. These increase compilation time, but if not set only python tests can be executed. Default setting is ON" ON)
```

rubenzorrilla: Typo.

loumalouomega: But I thought the text was informative, do you mean to check the variable and throw the message?

philbucher: Yes

The text as is is only shown in the cmake GUI afaik
It will for sure not be issues

philbucher (approved):

PR #10257, opened 2022-09-12
[Core] Remove unused Distance method from MathUtils

loumalouomega:
**📝 Description**

I realized that the Distance method from MathUtils is not used anywhere in the code, and it doesn't even makes sense as there is no Distance method in double

**🆕 Changelog**

- [Remove unused Distance method](https://github.com/KratosMultiphysics/Kratos/commit/ce14b25e77e14889fe0abb21a5a1de80be6c6b3a)


roigcarlo (dismissed):
Mark it as deprecated, then we can delete it.

We cannot be sure if someone outside Kratos is using it..., and deleting could cause am API break.

loumalouomega:
> Mark it as deprecated, then we can delete it.
> 
> We cannot be sure if someone outside Kratos is using it..., and deleting could cause am API break.

But the code doesn't even make sense. I realized this trying to do an explicit instantiation and being unable to compile. MathUtils is supposed to be used with double or float, and none of these two types has a method called Distance. Try to use it, and you will see it doesn't even make sense.

loumalouomega:
> Mark it as deprecated, then we can delete it.
> 
> We cannot be sure if someone outside Kratos is using it..., and deleting could cause am API break.

The method is very, very old, with the git blame it is not possible to track who added originally

RiccardoRossi:
i would be positive about deleting this straight. The code really does not make sense...

RiccardoRossi:
if we want to be purist we could throw an error if used...

RiccardoRossi:
but no, better let's remove it for good, and if anyone complains about a compilation error than we'll know it

rubenzorrilla:
Agree. This cannot be used in current state so an eventual user of it (which in my opinion is nonexistent) must had edited it in a custom repo.

loumalouomega:
> Agree. This cannot be used in current state so an eventual user of it (which in my opinion is nonexistent) must had edited it in a custom repo.

Then approval?

rubenzorrilla (approved):
Approval.

loumalouomega:
@roigcarlo can you remove the block?

PR #10263, opened 2022-09-13
[thermal dem] Improvements in Energy Computation

rlrangel:
In DEMApp:
* Computing energy dissipation by rolling friction.
* Added option to compute gravitational potential energy in function SphericParticle::Calculate.
* Added RollingResistance to particle data buffer.
* Changed input arguments of function StoreBallToBallContactInfo and StoreBallToRigidFaceContactInfo.

In ThermalDEMApp:
* Modified the way heat is calculated from dissipated energy.
* New graph for energy contributions.


applications/DEMApplication/DEM_application.cpp
```diff
@@ -313,6 +313,7 @@ KRATOS_CREATE_VARIABLE(double, PARTICLE_ROTATIONAL_KINEMATIC_ENERGY)
 KRATOS_CREATE_VARIABLE(double, PARTICLE_GRAVITATIONAL_ENERGY)
 KRATOS_CREATE_VARIABLE(double, PARTICLE_INELASTIC_VISCODAMPING_ENERGY)
 KRATOS_CREATE_VARIABLE(double, PARTICLE_INELASTIC_FRICTIONAL_ENERGY)
+KRATOS_CREATE_VARIABLE(double, PARTICLE_INELASTIC_ROLLRESIST_ENERGY)
```
AlejandroCornejo: the name is already long, you can write the full name I think

applications/DEMApplication/custom_elements/analytic_spheric_particle.cpp
```diff
@@ -210,7 +210,6 @@ void AnalyticSphericParticle::EvaluateBallToBallForcesForPositiveIndentiations(S
 void AnalyticSphericParticle::ComputeBallToRigidFaceContactForce(SphericParticle::ParticleDataBuffer & data_buffer,
                                                                 array_1d<double, 3>& r_elastic_force,
                                                                 array_1d<double, 3>& r_contact_force,
-                                                                double& RollingResistance,
```
AlejandroCornejo, line 213: do you need to set a default for the provious behaviour? @maceligueta

applications/DEMApplication/custom_elements/analytic_spheric_particle.h
```diff
@@ -170,7 +170,6 @@ void EvaluateBallToBallForcesForPositiveIndentiations(SphericParticle::ParticleD
 void ComputeBallToRigidFaceContactForce(SphericParticle::ParticleDataBuffer & data_buffer,
                                         array_1d<double, 3>& r_elastic_force,
                                         array_1d<double, 3>& r_contact_force,
-                                        double& RollingResistance,
```
AlejandroCornejo, line 173: `r_rolling_resistance` should be the style according to what I see in DEMApp. Don you have any style convention in DEM @maceligueta ?

applications/ThermalDEMApplication/custom_elements/thermal_spheric_particle.cpp
```diff
@@ -148,15 +137,35 @@ namespace Kratos
     SetGenerationModel(generation_model);
     SetRealContactModel(real_contact_model);
 
+    // Set flags
+    mHasMotion = r_process_info[MOTION_OPTION];
+
+    if ((GetProperties().Has(THERMAL_EXPANSION_COEFFICIENT) && GetProperties()[THERMAL_EXPANSION_COEFFICIENT] != 0.0) ||
```
AlejandroCornejo: you can declare the properties `r_properties` above and use the same one in all `GetProperties()` that you have there

AlejandroCornejo (commented):
That's a lot of stuff. Next time try to do more and smaller PR, they are easier to review

@maceligueta This affects some parts of the DEM...

AlejandroCornejo:
"Added RollingResistance to particle data buffer."

This may be an issue for the DEM guys?

applications/DEMApplication/custom_elements/analytic_spheric_particle.cpp
maceligueta: No, it is ok to remove this argument. Actually, @ChengshunShang1996  is taking steps in the same direction in PR #10188

applications/DEMApplication/custom_elements/analytic_spheric_particle.h
maceligueta: no problem, we are close to removing it forever.

applications/DEMApplication/custom_elements/spheric_particle.cpp
```diff
@@ -1935,9 +1940,9 @@ std::unique_ptr<DEMDiscontinuumConstitutiveLaw> SphericParticle::pCloneDiscontin
 
 void SphericParticle::ComputeOtherBallToBallForces(array_1d<double, 3>& other_ball_to_ball_forces) {}
 
-void SphericParticle::StoreBallToBallContactInfo(const ProcessInfo& r_process_info, SphericParticle::ParticleDataBuffer& data_buffer, double GlobalContactForceTotal[3], double LocalContactForceDamping[3], bool sliding) {}
+void SphericParticle::StoreBallToBallContactInfo(const ProcessInfo& r_process_info, SphericParticle::ParticleDataBuffer& data_buffer, double GlobalContactForceTotal[3], double LocalContactForceTotal[3], double LocalContactForceDamping[3]) {}
 
-void SphericParticle::StoreBallToRigidFaceContactInfo(const ProcessInfo& r_process_info, SphericParticle::ParticleDataBuffer& data_buffer, double GlobalContactForceTotal[3], double LocalContactForceDamping[3], bool sliding) {}
+void SphericParticle::StoreBallToRigidFaceContactInfo(const ProcessInfo& r_process_info, SphericParticle::ParticleDataBuffer& data_buffer, double GlobalContactForceTotal[3], double LocalContactForceTotal[3], double LocalContactForceDamping[3]) {}
```
maceligueta: I am not against removing sliding, but it seems a relevant feature of the contact between two particles. Why removing it?

maceligueta:
> "Added RollingResistance to particle data buffer."
> 
> This may be an issue for the DEM guys?

As long as it works, no problem. But anyways, we are migrating towards a Base Class Rolling Friction Model with a clear API which will allow several rolling friction models. In that case, we will need @rlrangel 's approval, because energy will need to be calculated for every model, hopefully in a unified way.

applications/DEMApplication/DEM_application.cpp
rlrangel: Changed to full name

applications/ThermalDEMApplication/custom_elements/thermal_spheric_particle.cpp
rlrangel: Done

applications/DEMApplication/custom_elements/spheric_particle.cpp
rlrangel: Back with sliding as an input. I removed simply because I was not using it anymore.

rlrangel:
> > "Added RollingResistance to particle data buffer."
> > This may be an issue for the DEM guys?
> 
> As long as it works, no problem. But anyways, we are migrating towards a Base Class Rolling Friction Model with a clear API which will allow several rolling friction models. That will allow those models to store member variables, so no need for them to use the data buffer. Once finished, we will ask for @rlrangel 's approval, because energy will need to be calculated for every model, hopefully in a unified way.

The RollingResistance was a local variable and I added to the DataBuffer to avoid having to pass this variable as an argument to the functions where I need it.
That's why the RollingResistance  was removed from the argument list of some functions, as ntoed by @AlejandroCornejo.
All these functions already had the DataBuffer  as an input.

It is interesting to hear that there will be a class for rolling resistance models.
Actually, I was wondering where does the current model come from, as it is different from what I have seen in the literature.
I am currently investigating heat generation by energy dissipation, and rolling friction is one of the dissipative mechanisms, so I need it to be computed correctly.
However, I'm still a bit concerned about the way I'm computing it in this PR, maybe it will be improved soon.

maceligueta:
> > > "Added RollingResistance to particle data buffer."
> > > This may be an issue for the DEM guys?
> > 
> > 
> > As long as it works, no problem. But anyways, we are migrating towards a Base Class Rolling Friction Model with a clear API which will allow several rolling friction models. That will allow those models to store member variables, so no need for them to use the data buffer. Once finished, we will ask for @rlrangel 's approval, because energy will need to be calculated for every model, hopefully in a unified way.
> 
> The RollingResistance was a local variable and I added to the DataBuffer to avoid having to pass this variable as an argument to the functions where I need it. That's why the RollingResistance was removed from the argument list of some functions, as ntoed by @AlejandroCornejo. All these functions already had the DataBuffer as an input.
> 
> It is interesting to hear that there will be a class for rolling resistance models. Actually, I was wondering where does the current model come from, as it is different from what I have seen in the literature. I am currently investigating heat generation by energy dissipation, and rolling friction is one of the dissipative mechanisms, so I need it to be computed correctly. However, I'm still a bit concerned about the way I'm computing it in this PR, maybe it will be improved soon.

Yes, it will (hopefully) soon be merged by PR #10188.

maceligueta (approved):

PR #10268, opened 2022-09-14
[Core] Recursive Symlink Resolution

matekelemen:
Cherry-picked from #10004 at the request of @philbucher.

### DIFF
--- a/kratos/includes/kratos_filesystem.h
+++ b/kratos/includes/kratos_filesystem.h
@@ -106,7 +106,7 @@ void KRATOS_API(KRATOS_CORE) MPISafeCreateDirectories(const std::string& rPath);
  *  @note The existence of the final result is not checked and is up to the user.
  *  @note The input is returned if it is not a symlink.
  */
-std::string ResolveSymlinks(const std::string& rPath);
+std::filesystem::path ResolveSymlinks(const std::filesystem::path& rPath);
 
 } // namespace FilesystemExtensions
 } // namespace Kratos
--- a/kratos/includes/kratos_filesystem.h
+++ b/kratos/includes/kratos_filesystem.h
@@ -98,6 +98,16 @@ std::vector<std::string> KRATOS_API(KRATOS_CORE) ListDirectory(const std::string
  */
 void KRATOS_API(KRATOS_CORE) MPISafeCreateDirectories(const std::string& rPath);
 
+/** @brief Resolve symlinks recursively.
+ *
+ *  @param rPath: path to a symbolic link.
+ *  @return The result of the recursive dereferencing.
+ *  @throws If the input path does not exist or the symlink is cyclic.
+ *  @note The existence of the final result is not checked and is up to the user.
+ *  @note The input is returned if it is not a symlink.
+ */
+std::string ResolveSymlinks(const std::string& rPath);
+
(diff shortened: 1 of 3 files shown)


kratos/includes/kratos_filesystem.h
```diff
@@ -98,6 +98,16 @@ std::vector<std::string> KRATOS_API(KRATOS_CORE) ListDirectory(const std::string
...
 
+/** @brief Resolve symlinks recursively.
+ *
+ *  @param rPath: path to a symbolic link.
+ *  @return The result of the recursive dereferencing.
+ *  @throws If the input path does not exist or the symlink is cyclic.
+ *  @note The existence of the final result is not checked and is up to the user.
+ *  @note The input is returned if it is not a symlink.
+ */
+std::string ResolveSymlinks(const std::string& rPath);
```
philbucher:
```suggestion
std::filesystem::path KRATOS_API(KRATOS_CORE) ResolveSymlinks(const std::filesystem::path& rPath);
```

will refactor the other methods also soon

kratos/sources/kratos_filesystem.cpp
```diff
@@ -132,5 +133,23 @@ void MPISafeCreateDirectories(const std::string& rPath)
...
+
+std::string ResolveSymlinks(const std::string& rPath)
+{
+    std::filesystem::path path = rPath;
+    KRATOS_ERROR_IF_NOT(std::filesystem::exists(path)) << "File not found: " << rPath;
+
+    std::set<std::filesystem::path> symlinks;
+
+    while (std::filesystem::is_symlink(path))
+    {
```
philbucher:
```suggestion
    while (std::filesystem::is_symlink(path)) {
```

kratos/sources/kratos_filesystem.cpp
```diff
@@ -132,5 +133,23 @@ void MPISafeCreateDirectories(const std::string& rPath)
     }
 }
 
+
+std::string ResolveSymlinks(const std::string& rPath)
+{
+    std::filesystem::path path = rPath;
+    KRATOS_ERROR_IF_NOT(std::filesystem::exists(path)) << "File not found: " << rPath;
+
+    std::set<std::filesystem::path> symlinks;
```
philbucher, line 143: why not unordered?

philbucher (commented):
please also add 2 tests, one with and one without cyclic symlinks

kratos/sources/kratos_filesystem.cpp
matekelemen: I don't expect many items in that set (probably in the single digits) so a hash table would be an overkill. In fact, I have a hard time imagining a scenario in which more than 2-3 indirections would be reasonable. A couple of lexicographical string comparisons in ```std::set``` surely outperform computing the hash of a lengthy absolute path.

matekelemen:
@philbucher good thing you asked for the tests, ```std::filesystem``` is an inconsistent mess when it comes to following symlinks.

philbucher (dismissed):
Perfect 👍🏻

matekelemen:
fixed a small mistake in the tests; @philbucher can you please take a quick peek again?

philbucher (approved):

PR #10271, opened 2022-09-15
Forwarding 10201 to v9.2

roigcarlo:
**📝 Description**
Forward porting the changes from #10201, as Release can't wait more and I need some mechanism there to be in place.

**🆕 Changelog**
- Kratos now uses C++17 by default.


roigcarlo:
Note: As this is internal I will just merge it directly to the Release brach.

PR #10277, opened 2022-09-15
adding a function to effectively map skin velocity to embedded velocity

RiccardoRossi:
**📝 Description**
adding a function to effectively map skin velocity to embedded velocity

i wrongly pushed this in a branch that was already merged...but we can still use it 

- Feature
- 
**🆕 Changelog**
adding a new function

### DIFF
--- a/applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
+++ b/applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
@@ -446,8 +446,9 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
             }
 
             if(npos>0 && nneg>0){ //element is split
-                for(unsigned int i=0; i<geom.size(); ++i)
-                    cut_elem_nodes.push_back(geom(i));
+                for(unsigned int i=0; i<r_geom.size(); ++i) {
+                    cut_elem_nodes.push_back(r_geom(i));
+                }
             }
         }
         cut_elem_nodes.Unique();
--- a/applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
+++ b/applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
@@ -420,7 +420,7 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
     rpModShapeFunc->ComputeNegativeExteriorFaceAreaNormals(rNormals, FaceId, GeometryData::IntegrationMethod::GI_GAUSS_2);
 }
 
-    void FluidAuxiliaryUtilities::EmbeddedSkinToVolumeMapping(
+    void FluidAuxiliaryUtilities::MapEmbeddedVelocityFromSkinToVolume(
         ModelPart& rVolumeModelPart,
         ModelPart& rSkinModelPart,
         const double SearchRadius
@@ -482,7 +482,7 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
                 //assign interpolation
                 auto& r_v = rNode.GetValue(EMBEDDED_VELOCITY); //we know it is zero since we zeroed all the nodes at the beginning of the function
                 for(unsigned int i=0; i<nfound; ++i){
-                    noalias(r_v) += shape_functions[i]*neighbours[i]->FastGetSolutionStepValue(VELOCITY);
+                    noalias(r_v) += shape_functions[i]*r_neighbours[i]->FastGetSolutionStepValue(VELOCITY);
                 }
             });
     }
--- a/applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
+++ b/applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
@@ -439,7 +439,7 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
         //detect all the nodes that belong to cut elements
         ModelPart::NodesContainerType cut_elem_nodes;
         for(const auto& rElem : rVolumeModelPart.Elements()){
-            const auto& geom = rElem.GetGeometry();
+            const auto& r_geom = rElem.GetGeometry();
             
             unsigned int npos=0, nneg=0;
             for (const auto& r_node : geom) {
--- a/applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
+++ b/applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
@@ -22,6 +22,9 @@
 #include "modified_shape_functions/tetrahedra_3d_4_modified_shape_functions.h"
 #include "utilities/parallel_utilities.h"
 #include "utilities/reduction_utilities.h"
+#include "spatial_containers/bins_dynamic.h"
+#include "utilities/rbf_shape_functions_utility.h"
+#include "utilities/parallel_utilities.h"
 
 // Application includes
 #include "fluid_auxiliary_utilities.h"
@@ -417,5 +420,76 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
     rpModShapeFunc->ComputeNegativeExteriorFaceShapeFunctionsAndGradientsValues(rShapeFunctions, n_pos_DN_DX, rWeights, FaceId, GeometryData::IntegrationMethod::GI_GAUSS_2);
     rpModShapeFunc->ComputeNegativeExteriorFaceAreaNormals(rNormals, FaceId, GeometryData::IntegrationMethod::GI_GAUSS_2);
 }
+
+    void FluidAuxiliaryUtilities::EmbeddedSkinToVolumeMapping(
+        ModelPart& rVolumeModelPart,
+        ModelPart& rSkinModelPart,
+        const double SearchRadius
+        )
+    {
+        //put the nodes in the skin into a search database
+        typedef BinsDynamic<3, Node<3>, std::vector<Node<3>::Pointer>> DynamicBins;
+        DynamicBins locator(rSkinModelPart.Nodes().GetContainer().begin(), rSkinModelPart.Nodes().GetContainer().end());
+
+        //reset the EMBEDDED_VELOCITY
+        block_for_each(rVolumeModelPart.Nodes(), [](auto& rNode){
+            rNode.GetValue(EMBEDDED_VELOCITY).clear();
+        });
+
+        //detect all the nodes that belong to cut elements
+        ModelPart::NodesContainerType cut_elem_nodes;
+        for(const auto& rElem : rVolumeModelPart.Elements()){
+            const auto& geom = rElem.GetGeometry();
+            
+            unsigned int npos=0, nneg=0;
+            for(unsigned int i=0; i<geom.size(); ++i){
+                if(geom[i].FastGetSolutionStepValue(DISTANCE)>0)
+                    npos++;
+                else
+                    nneg++;
+            }
+
+            if(npos>0 && nneg>0){ //element is split
+                for(unsigned int i=0; i<geom.size(); ++i)
+                    cut_elem_nodes.push_back(geom(i));
+            }
+        }
+        cut_elem_nodes.Unique();
+
+        unsigned int MaxResults = 100; //deliberately low, we do not want to interpolate from more than this nodes
+        auto TLS = std::make_pair(std::vector<Node<3>::Pointer>(MaxResults), std::vector<double>(MaxResults));
+        
+        //for every interface node (nodes in cut elements)
+        block_for_each(cut_elem_nodes, TLS, [&locator, SearchRadius](auto& rNode, auto& rTLS){
+                //find all neighbours within a radius
+                auto& neighbours = rTLS.first;
+                auto& distances  = rTLS.second;
+                const unsigned int nfound = locator.SearchInRadius( rNode, 
+                                                                SearchRadius, 
+                                                                neighbours.begin(), 
+                                                                distances.begin(), 
+                                                                neighbours.size());
+                if(nfound == 0)
+                    std::cout << "not found any neighbour!!! for node " << rNode.Id() << std::endl;
+
+                //compute RBF basis
+                Matrix cloud_coords(nfound,3);
+                Vector shape_functions(nfound);
+                for(unsigned int i=0; i<nfound; ++i){
+                    cloud_coords(i,0)=neighbours[i]->X(); 
+                    cloud_coords(i,1)=neighbours[i]->Y(); 
+                    cloud_coords(i,2)=neighbours[i]->Z();
+                }
+                RBFShapeFunctionsUtility::CalculateShapeFunctions(cloud_coords,rNode.Coordinates(), shape_functions);
+
+                //assign interpolation
+                auto& v = rNode.GetValue(EMBEDDED_VELOCITY); //we know it is zero since we zeroed all the nodes at the beginning of the function
+                for(unsigned int i=0; i<nfound; ++i){
+                    noalias(v) += shape_functions[i]*neighbours[i]->FastGetSolutionStepValue(VELOCITY);
+                }
+            });
+    }
+
+
 
 } // namespace Kratos
--- a/applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
+++ b/applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
@@ -420,7 +420,7 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
     rpModShapeFunc->ComputeNegativeExteriorFaceAreaNormals(rNormals, FaceId, GeometryData::IntegrationMethod::GI_GAUSS_2);
 }
 
-    void FluidAuxiliaryUtilities::MapEmbeddedVelocityFromSkinToVolume(
+    void FluidAuxiliaryUtilities::MapVelocityFromSkinToVolumeRBF(
         ModelPart& rVolumeModelPart,
         ModelPart& rSkinModelPart,
         const double SearchRadius
--- a/applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
+++ b/applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
@@ -459,22 +459,23 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
         //for every interface node (nodes in cut elements)
         block_for_each(cut_elem_nodes, TLS, [&locator, SearchRadius](auto& rNode, auto& rTLS){
                 //find all neighbours within a radius
-                auto& neighbours = rTLS.first;
-                auto& distances  = rTLS.second;
-                const unsigned int nfound = locator.SearchInRadius( rNode, 
-                                                                SearchRadius, 
-                                                                neighbours.begin(), 
-                                                                distances.begin(), 
-                                                                neighbours.size());
+                auto& r_neighbours = rTLS.first;
+                auto& r_distances  = rTLS.second;
+                const unsigned int nfound = locator.SearchInRadius(
+                    rNode, 
+                    SearchRadius, 
+                    r_neighbours.begin(), 
+                    r_distances.begin(), 
+                    r_neighbours.size());
                 KRATOS_ERROR_IF(nfound == 0) << "Not found any neighbor for node " << rNode.Id() << " in skin model part." << std::endl;
 
                 //compute RBF basis
                 Matrix cloud_coords(nfound,3);
                 Vector shape_functions(nfound);
                 for(unsigned int i=0; i<nfound; ++i){
-                    cloud_coords(i,0)=neighbours[i]->X(); 
-                    cloud_coords(i,1)=neighbours[i]->Y(); 
-                    cloud_coords(i,2)=neighbours[i]->Z();
+                    cloud_coords(i,0)=r_neighbours[i]->X(); 
+                    cloud_coords(i,1)=r_neighbours[i]->Y(); 
+                    cloud_coords(i,2)=r_neighbours[i]->Z();
                 }
                 RBFShapeFunctionsUtility::CalculateShapeFunctions(cloud_coords,rNode.Coordinates(), shape_functions);
 
--- a/applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
+++ b/applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
@@ -453,8 +453,8 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
         }
         cut_elem_nodes.Unique();
 
-        unsigned int MaxResults = 100; //deliberately low, we do not want to interpolate from more than this nodes
-        auto TLS = std::make_pair(std::vector<Node<3>::Pointer>(MaxResults), std::vector<double>(MaxResults));
+        unsigned int max_results = 100; //deliberately low, we do not want to interpolate from more than this nodes
+        auto TLS = std::make_pair(std::vector<Node<3>::Pointer>(max_results), std::vector<double>(max_results));
         
         //for every interface node (nodes in cut elements)
         block_for_each(cut_elem_nodes, TLS, [&locator, SearchRadius](auto& rNode, auto& rTLS){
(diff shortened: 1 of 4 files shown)


RiccardoRossi (commented):
@rubenzorrilla this is directly targeted at working together

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
```diff
@@ -418,4 +421,75 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
...
+                else
+                    nneg++;
+            }
+
+            if(npos>0 && nneg>0){ //element is split
+                for(unsigned int i=0; i<geom.size(); ++i)
+                    cut_elem_nodes.push_back(geom(i));
+            }
+        }
+        cut_elem_nodes.Unique();
```
matekelemen, line 454: ```cut_elem_nodes``` should be a set or hash set.

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
```diff
@@ -418,4 +421,75 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
...
+        for(const auto& rElem : rVolumeModelPart.Elements()){
+            const auto& geom = rElem.GetGeometry();
+            
+            unsigned int npos=0, nneg=0;
+            for(unsigned int i=0; i<geom.size(); ++i){
+                if(geom[i].FastGetSolutionStepValue(DISTANCE)>0)
+                    npos++;
+                else
+                    nneg++;
+            }
```
matekelemen:
```suggestion
            for (const auto& r_node : geom) {
                r_node.FastGetSolutionStepValue(DISTANCE) > 0 ? ++npos : ++nneg;
            }
```

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
```diff
@@ -418,4 +421,75 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
...
+                //find all neighbours within a radius
+                auto& neighbours = rTLS.first;
+                auto& distances  = rTLS.second;
+                const unsigned int nfound = locator.SearchInRadius( rNode, 
+                                                                SearchRadius, 
+                                                                neighbours.begin(), 
+                                                                distances.begin(), 
+                                                                neighbours.size());
+                if(nfound == 0)
+                    std::cout << "not found any neighbour!!! for node " << rNode.Id() << std::endl;
```
matekelemen: Is this a leftover from debugging? If not, I guess it should go in the logger.

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
RiccardoRossi: i am not very clear you are right about this one specifically. The cost of what i have here is nlogn, and the cost of insering n times in a set should also be the same.

RiccardoRossi: point is, that later on a need to loop in parallel over this. Neither std::set nor std::unordered_set are suitable to that, so anyway i would need to copy back into a vector to be able to further process it in parallel. At that moment i prefer to push back to a vector and make it unique

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
RiccardoRossi: ok, i transformed it into an error

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
RiccardoRossi: > i am not very clear you are right about this one specifically. The cost of what i have here is nlogn, and the cost of insering n times in a set should also be the same.

i correct myself. It is nlogn in a set, n in a unordered_set

RiccardoRossi: @matekelemen this is ready

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
```diff
@@ -22,6 +22,9 @@
 #include "modified_shape_functions/tetrahedra_3d_4_modified_shape_functions.h"
 #include "utilities/parallel_utilities.h"
 #include "utilities/reduction_utilities.h"
+#include "spatial_containers/bins_dynamic.h"
+#include "utilities/rbf_shape_functions_utility.h"
+#include "utilities/parallel_utilities.h"
```
rubenzorrilla:
```suggestion
```
Parallel utilities are already include above.

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
```diff
@@ -418,4 +421,71 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
...
+
+        //reset the EMBEDDED_VELOCITY
+        block_for_each(rVolumeModelPart.Nodes(), [](auto& rNode){
+            rNode.GetValue(EMBEDDED_VELOCITY).clear();
+        });
+
+        //detect all the nodes that belong to cut elements
+        ModelPart::NodesContainerType cut_elem_nodes;
+        for(const auto& rElem : rVolumeModelPart.Elements()){
+            const auto& geom = rElem.GetGeometry();
```
rubenzorrilla:
```suggestion
            const auto& r_geom = rElem.GetGeometry();
```

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
```diff
@@ -418,4 +421,71 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
...
+            rNode.GetValue(EMBEDDED_VELOCITY).clear();
+        });
+
+        //detect all the nodes that belong to cut elements
+        ModelPart::NodesContainerType cut_elem_nodes;
+        for(const auto& rElem : rVolumeModelPart.Elements()){
+            const auto& r_geom = rElem.GetGeometry();
+            
+            unsigned int npos=0, nneg=0;
+            for (const auto& r_node : geom) {
```
rubenzorrilla:
```suggestion
            for (const auto& r_node : r_geom) {
```

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
```diff
@@ -418,4 +420,71 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
...
+            const auto& r_geom = rElem.GetGeometry();
+            
+            unsigned int npos=0, nneg=0;
+            for (const auto& r_node : r_geom) {
+                r_node.FastGetSolutionStepValue(DISTANCE) > 0 ? ++npos : ++nneg;
+            }
+
+            if(npos>0 && nneg>0){ //element is split
+                for(unsigned int i=0; i<geom.size(); ++i)
+                    cut_elem_nodes.push_back(geom(i));
```
rubenzorrilla:
```suggestion
                for(unsigned int i=0; i<r_geom.size(); ++i) {
                    cut_elem_nodes.push_back(r_geom(i));
                }
```

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
```diff
@@ -418,4 +420,72 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
...
+            if(npos>0 && nneg>0){ //element is split
+                for(unsigned int i=0; i<r_geom.size(); ++i) {
+                    cut_elem_nodes.push_back(r_geom(i));
+                }
+            }
+        }
+        cut_elem_nodes.Unique();
+
+        unsigned int MaxResults = 100; //deliberately low, we do not want to interpolate from more than this nodes
+        auto TLS = std::make_pair(std::vector<Node<3>::Pointer>(MaxResults), std::vector<double>(MaxResults));
```
rubenzorrilla:
```suggestion
        unsigned int max_results = 100; //deliberately low, we do not want to interpolate from more than this nodes
        auto TLS = std::make_pair(std::vector<Node<3>::Pointer>(max_results), std::vector<double>(max_results));
```

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
```diff
@@ -418,4 +420,72 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
...
+        block_for_each(cut_elem_nodes, TLS, [&locator, SearchRadius](auto& rNode, auto& rTLS){
+                //find all neighbours within a radius
+                auto& neighbours = rTLS.first;
+                auto& distances  = rTLS.second;
+                const unsigned int nfound = locator.SearchInRadius( rNode, 
+                                                                SearchRadius, 
+                                                                neighbours.begin(), 
+                                                                distances.begin(), 
+                                                                neighbours.size());
+                KRATOS_ERROR_IF(nfound == 0) << "not found any neighbour!!! for node " << rNode.Id() << std::endl;
```
rubenzorrilla:
```suggestion
                KRATOS_ERROR_IF(nfound == 0) << "Not found any neighbor for node " << rNode.Id() << " in skin model part." << std::endl;
```

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.h
```diff
@@ -174,6 +174,21 @@ class KRATOS_API(FLUID_DYNAMICS_APPLICATION) FluidAuxiliaryUtilities
...
     static ModifiedShapeFunctionsFactoryType GetStandardModifiedShapeFunctionsFactory(const GeometryType& rGeometry);
 
+    /**
+     * @brief This function maps the velocity of the nodes "on the skin" by the use of RBF interpolation
+     * to the variable EMBEDDED_VELOCITY of the nodes which correspond to the cut cells in the volume.
+     * It is designed for use together with the "embedded solver"
+     * @param rVolumeModelPart is the destination domain on which EMBEDDED_VELOCITY will be calculated
+     * @param rSkinModelPart is the skin of the object from which the velocity will be taken
+     * @param SearchRadius is the radius which will be used in searching the neighbours. It needs to be sufficiently large otherwise the 
+     *                     method will fail in the calculation of the RBF basis
```
rubenzorrilla:
```suggestion
     * @param SearchRadius is the radius which will be used in searching the neighbours. It needs to be sufficiently large otherwise the method will fail in the calculation of the RBF basis
```

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
```diff
@@ -418,4 +420,72 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
...
+                KRATOS_ERROR_IF(nfound == 0) << "Not found any neighbor for node " << rNode.Id() << " in skin model part." << std::endl;
+
+                //compute RBF basis
+                Matrix cloud_coords(nfound,3);
+                Vector shape_functions(nfound);
+                for(unsigned int i=0; i<nfound; ++i){
+                    cloud_coords(i,0)=neighbours[i]->X(); 
+                    cloud_coords(i,1)=neighbours[i]->Y(); 
+                    cloud_coords(i,2)=neighbours[i]->Z();
+                }
```
rubenzorrilla:
```suggestion
                auto& r_neighbours = rTLS.first;
                auto& r_distances  = rTLS.second;
                const unsigned int nfound = locator.SearchInRadius(
                    rNode, 
                    SearchRadius, 
                    r_neighbours.begin(), 
                    r_distances.begin(), 
                    r_neighbours.size());
                KRATOS_ERROR_IF(nfound == 0) << "Not found any neighbor for node " << rNode.Id() << " in skin model part." << std::endl;

                //compute RBF basis
                Matrix cloud_coords(nfound,3);
                Vector shape_functions(nfound);
                for(unsigned int i=0; i<nfound; ++i){
                    cloud_coords(i,0)=r_neighbours[i]->X(); 
                    cloud_coords(i,1)=r_neighbours[i]->Y(); 
                    cloud_coords(i,2)=r_neighbours[i]->Z();
                }
```

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
```diff
@@ -418,4 +420,73 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
...
+                    cloud_coords(i,1)=r_neighbours[i]->Y(); 
+                    cloud_coords(i,2)=r_neighbours[i]->Z();
+                }
+                RBFShapeFunctionsUtility::CalculateShapeFunctions(cloud_coords,rNode.Coordinates(), shape_functions);
+
+                //assign interpolation
+                auto& v = rNode.GetValue(EMBEDDED_VELOCITY); //we know it is zero since we zeroed all the nodes at the beginning of the function
+                for(unsigned int i=0; i<nfound; ++i){
+                    noalias(v) += shape_functions[i]*neighbours[i]->FastGetSolutionStepValue(VELOCITY);
+                }
```
rubenzorrilla:
```suggestion
                auto& r_v = rNode.GetValue(EMBEDDED_VELOCITY); //we know it is zero since we zeroed all the nodes at the beginning of the function
                for(unsigned int i=0; i<nfound; ++i){
                    noalias(r_v) += shape_functions[i]*neighbours[i]->FastGetSolutionStepValue(VELOCITY);
                }
```

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
```diff
@@ -418,4 +420,73 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
...
+
+        //reset the EMBEDDED_VELOCITY
+        block_for_each(rVolumeModelPart.Nodes(), [](auto& rNode){
+            rNode.GetValue(EMBEDDED_VELOCITY).clear();
+        });
+
+        //detect all the nodes that belong to cut elements
+        ModelPart::NodesContainerType cut_elem_nodes;
+        for(const auto& rElem : rVolumeModelPart.Elements()){
+            const auto& r_geom = rElem.GetGeometry();
```
rubenzorrilla:
```suggestion
        for(const auto& r_elem : rVolumeModelPart.Elements()){
            const auto& r_geom = r_elem.GetGeometry();
```

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.h
```diff
@@ -174,6 +174,20 @@ class KRATOS_API(FLUID_DYNAMICS_APPLICATION) FluidAuxiliaryUtilities
      */
     static ModifiedShapeFunctionsFactoryType GetStandardModifiedShapeFunctionsFactory(const GeometryType& rGeometry);
 
+    /**
+     * @brief This function maps the velocity of the nodes "on the skin" by the use of RBF interpolation
+     * to the variable EMBEDDED_VELOCITY of the nodes which correspond to the cut cells in the volume.
+     * It is designed for use together with the "embedded solver"
```
rubenzorrilla, line 180:
```suggestion
     * It is designed for use together with the "embedded solver".
     * Also note that this function assumes EMBEDDED_VELOCITY to be already allocated in the non-historical database of all the volume mesh nodes.
     * It also assumes that VELOCITY is present in the skin model part historical database.
```

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.h
```diff
@@ -174,6 +174,20 @@ class KRATOS_API(FLUID_DYNAMICS_APPLICATION) FluidAuxiliaryUtilities
...
+     * @param rVolumeModelPart is the destination domain on which EMBEDDED_VELOCITY will be calculated
+     * @param rSkinModelPart is the skin of the object from which the velocity will be taken
+     * @param SearchRadius is the radius which will be used in searching the neighbours. It needs to be sufficiently large otherwise the method will fail in the calculation of the RBF basis
+     */
+    static void EmbeddedSkinToVolumeMapping(
```
rubenzorrilla:
```suggestion
    static void EmbeddedVelocityFromSkinToVolumeMapping(
```
IMO the function name is not very well descriptive as it seems to be valid to any set of variables. Indeed it can be easily misunderstood with the `CalculateEmbeddedNodalVariableFromSkinProcess`. I'd suggest to rename it to something like `CalculateEmbeddedVelocityFromSkinVelocityRBF`.

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.h
```diff
@@ -244,6 +258,7 @@ class KRATOS_API(FLUID_DYNAMICS_APPLICATION) FluidAuxiliaryUtilities
         std::vector<array_1d<double,3>>& rNormals,
         Vector& rWeights);
 
+
```
rubenzorrilla:
```suggestion
```

rubenzorrilla (dismissed):
Two minor comments:
- The naming as this can be easily misunderstood what we currently have
- We're assuming that `EMBEDDED_VELOCITY` is already allocated (i.e. someone has already done the `SetValue`) and that `VELOCITY` is in the historical database of the skin. I think it's fair to assume both things but we should at least document it (I already did).

Besides, test is missing.

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
matekelemen: > later on a need to loop in parallel over this.

Good point. We should be able to partition any container that can be accessed by at least forward iterators. The STL can do it so it shouldn't be a problem for us either.

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
```diff
@@ -418,4 +420,73 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
...
+
+                //compute RBF basis
+                Matrix cloud_coords(nfound,3);
+                Vector shape_functions(nfound);
+                for(unsigned int i=0; i<nfound; ++i){
+                    cloud_coords(i,0)=r_neighbours[i]->X(); 
+                    cloud_coords(i,1)=r_neighbours[i]->Y(); 
+                    cloud_coords(i,2)=r_neighbours[i]->Z();
+                }
+                RBFShapeFunctionsUtility::CalculateShapeFunctions(cloud_coords,rNode.Coordinates(), shape_functions);
```
philbucher, line 480: hm this is like half of an RBF mapper, maybe I should use this to implement it ...

RiccardoRossi: well, more than half...it IS a RBF mapping...except that we do not return the interpolation matrix but just the application of it to some data. (with the new matrix returning T would be quite easy)

philbucher: I think there is a second step required to compute the coefficients for the T matrix

RiccardoRossi: the coerficients of T are the shape functions...

philbucher: Hmm I have to check

Will get back to you :D

RiccardoRossi: maybe we should have a short conversation about this. send me a telegram when you have some time

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
RiccardoRossi: to be honest i am not sure of how we can achieve this for an unorder_set without assumptions on the way the set is actually stored. (possibly stl does make those assumptions). In any case i agree that it would be interesting to have this capability...although of course not in this PR :-)

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.h
RiccardoRossi: i called it MapEmbeddedVelocityFromSkinToVolume

RiccardoRossi (commented):
@rubenzorrilla adopted your comments and added a test

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
```diff
@@ -418,4 +420,73 @@ void FluidAuxiliaryUtilities::CalculateSplitConditionGeometryData<false>(
     rpModShapeFunc->ComputeNegativeExteriorFaceAreaNormals(rNormals, FaceId, GeometryData::IntegrationMethod::GI_GAUSS_2);
 }
 
+    void FluidAuxiliaryUtilities::MapEmbeddedVelocityFromSkinToVolume(
```
rubenzorrilla:
```suggestion
    void FluidAuxiliaryUtilities::MapVelocityFromSkinToVolumeRBF(
```
If you want to call it map `EmbeddedVelocity` refers to the skin variable, which is `VELOCITY`. Besides I think we should make clear that RBF are used internally.

applications/FluidDynamicsApplication/tests/test_fluid_auxiliary_utilities.py
```diff
@@ -113,6 +113,39 @@ def testCalculateFlowRateNegativeSkin(self):
         negative_flow_rate_top = KratosFluid.FluidAuxiliaryUtilities.CalculateFlowRateNegativeSkin(top_skin_model_part)
         self.assertAlmostEqual(negative_flow_rate_top, ref_negative_flow_rate_top, 12)
 
+    def testCalculateFluidVolume(self):
```
rubenzorrilla: I think there is a copy&paste error in the test name.

applications/FluidDynamicsApplication/tests/test_fluid_auxiliary_utilities.py
```diff
@@ -113,6 +113,39 @@ def testCalculateFlowRateNegativeSkin(self):
...
+        fluid_model_part = self.model.GetModelPart("FluidModelPart")
+        fluid_model_part.AddNodalSolutionStepVariable(Kratos.DISTANCE)
```
rubenzorrilla: You are adding a nodal variable to a model part which is already populated with nodes.

applications/FluidDynamicsApplication/tests/test_fluid_auxiliary_utilities.py
```diff
@@ -113,6 +113,39 @@ def testCalculateFlowRateNegativeSkin(self):
...
+        for node in fluid_model_part.Nodes:
+            node.GetValue(Kratos.EMBEDDED_VELOCITY)
```
rubenzorrilla, line 119: Which is the purpose of this `GetValue`?

applications/FluidDynamicsApplication/tests/test_fluid_auxiliary_utilities.py
RiccardoRossi: ok, anyway the distance was there so it was not needed

applications/FluidDynamicsApplication/tests/test_fluid_auxiliary_utilities.py
RiccardoRossi: i need to allocate the EMBEDDED_VELOCITY...

applications/FluidDynamicsApplication/tests/test_fluid_auxiliary_utilities.py
RiccardoRossi: true

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.cpp
RiccardoRossi: done

applications/FluidDynamicsApplication/tests/test_fluid_auxiliary_utilities.py
```diff
@@ -113,6 +113,38 @@ def testCalculateFlowRateNegativeSkin(self):
         negative_flow_rate_top = KratosFluid.FluidAuxiliaryUtilities.CalculateFlowRateNegativeSkin(top_skin_model_part)
         self.assertAlmostEqual(negative_flow_rate_top, ref_negative_flow_rate_top, 12)
 
+    def testMapEmbeddedVelocityFromSkinToVolume(self):
```
rubenzorrilla:
```suggestion
    def testMapVelocityFromSkinToVolumeRBF(self):
```

rubenzorrilla (approved):
This is ready IMO. Approving to merge it if CI passes. Thanks @RiccardoRossi.

PR #10278, opened 2022-09-16
Release 9.2

roigcarlo:
**📝 Description**
Upgraded version number and some other fixes. Please notice that the changes in the package construction DO NOT AFFECT the released apps. Only the ones that are compiled internally.


scripts/wheels/windows/build.ps1
```diff
@@ -1,7 +1,7 @@
 param([System.String]$cotire="OFF")
 
 $pythons = "36", "37", "38", "39"
```
loumalouomega, line 3: With the last pybind11 should be 3.7 minimum

scripts/wheels/setup.py
```diff
@@ -39,6 +39,31 @@ def replaceKeywords(stringArray):
     f.writelines(list(map(lambda x: replaceKeyword(x) + "\n", conf["excluded_binaries"])))
     f.close()
 
+package_classifiers = [
+        "Programming Language :: Python :: 3",
+        "Programming Language :: C++",
+        "Programming Language :: Python",
+        "Programming Language :: Python :: 3.6",
```
loumalouomega, line 46: Idem pybind11 is stating with 3.7

roigcarlo: Ok, I will update for all packages after merging this :+1:

roigcarlo: But we are still able to generate 3.6 packages correctly, what is the problem?

loumalouomega: If it works, it works. But there is no official support anymore

rubenzorrilla (approved):

PR #9549, opened 2022-01-18
[CI] Adding sympy to centos and windows CI

EduardGomezEscandell:
**📝 Description**
Changes missed in #9544.

> There are some scripts within Kratos where sympy is imported (for instance here).
>
> So far this hadn't been an issue because no file that used them was unit-tested. Now, however, the FluidDynamicsApplication will start testing its symbolic file generation in #9426.

**🆕 Changelog**
- Added `sympy`to centos CI docker 
- Added `sympy` to windows CI run

PR #10288, opened 2022-09-19
[CI] Adding missing scipy to CI

loumalouomega:
**📝 Description**

I realized that we installed sympy some time ago in the CI:

https://github.com/KratosMultiphysics/Kratos/pull/9549

For me this was strange, as we thought it was a confusion because we use it scipy. Then I realized that we use sympy as well is some tests, so it makes sense. But It would make even more sense to also install scipy, as it is used in many tests that right now the CI must be skipping:

https://github.com/KratosMultiphysics/Kratos/search?q=scipy

**🆕 Changelog**

- [Adding missing scipy to CI](https://github.com/KratosMultiphysics/Kratos/commit/90cd49aef07a23faa23f6f5a0d35c3be69777abf)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -153,6 +153,7 @@ jobs:
       run: |
         pip install numpy
         pip install sympy
+        pip install scipy
         pip install h5py
 
     - name: Build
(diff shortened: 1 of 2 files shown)


roigcarlo:
:+1: From my side if no one has anything to say

roigcarlo (approved):

PR #10294, opened 2022-09-21
[DEM] Remove unused includes

loumalouomega:
**📝 Description**

Remove unused includes

**🆕 Changelog**

- [Remove unused includes](https://github.com/KratosMultiphysics/Kratos/commit/1f894cfc5272ae4465d331d6db8b60d634f702a8)

### DIFF
--- a/applications/DEMApplication/custom_utilities/particle_data.h
+++ b/applications/DEMApplication/custom_utilities/particle_data.h
@@ -21,7 +21,6 @@
 
 //Database includes
 #include "spatial_containers/spatial_containers.h"
-#include "containers/vector_map.h"
 #include "containers/pointer_vector_set.h"
 #include "containers/variables_list_data_value_container.h"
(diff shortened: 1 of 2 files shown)


maceligueta (approved):
Thanks @loumalouomega !! How did you find out we were not using this include? I guess it is happening with many more and finding them would certainly decrease the compilation time.

loumalouomega:
We were including this and having problems of compilation with C++17. The class was not actually used, and I assumed it was not used either in here, and I was right. I will open now a PR to remove this class or at least adding some tests. But the code is 10 years old and is not used anywhere

loumalouomega:
> We were including this and having problems of compilation with C++17. The class was not actually used, and I assumed it was not used either in here, and I was right. I will open now a PR to remove this class or at least adding some tests. But the code is 10 years old and is not used anywhere

See https://github.com/KratosMultiphysics/Kratos/pull/10299

PR #10295, opened 2022-09-21
[Core][CoSimulationApplication] Adding BeforeSolutionLoop to PythonSolver

loumalouomega:
**📝 Description**

In aim to unify with @KratosMultiphysics/altair (we are doing our best to unify from our side, but there are things that are complex to unify, and in addition I consider interesting to bring to Kratos), I propose to add the method BeforeSolutionLoop  to the base python solver (by default it does nothing, so keeping exactly the same behaviour). I extend then this to the @KratosMultiphysics/cosimulation application for consistency.

**🆕 Changelog**

- [Adding BeforeSolutionLoop to PythonSolver](https://github.com/KratosMultiphysics/Kratos/commit/46a1d1c8c0a0b5d3fa522a820df217b989e0386a)
- [Call solver BeforeSolutionLoop in analysis](https://github.com/KratosMultiphysics/Kratos/commit/5a17e18c20799afdd05b6af6d1884d3cdaa0f700)
- [Extend to CoSimulationApplication](https://github.com/KratosMultiphysics/Kratos/commit/62b97a54c9e030cee45f9d83b9a124aab0659e4a)

### DIFF
--- a/kratos/python_scripts/analysis_stage.py
+++ b/kratos/python_scripts/analysis_stage.py
@@ -54,6 +54,15 @@ class AnalysisStage(object):
         """
         return self.time < self.end_time
 
+    def RunBeforeSolutionLoop(self):
+        """This function executes the processes before solution loop of the AnalysisStage
+        It can be overridden by derived classes.
+        """
+
+        for process in self._GetListOfProcesses():
+            process.ExecuteBeforeSolutionLoop()
+        self._GetSolver().BeforeSolutionLoop()
+
     def RunSolutionLoop(self):
         """This function executes the solution loop of the AnalysisStage
         It can be overridden by derived classes
@@ -95,8 +104,8 @@ class AnalysisStage(object):
 
         self.ModifyAfterSolverInitialize()
 
-        for process in self._GetListOfProcesses():
-            process.ExecuteBeforeSolutionLoop()
+        # Execute before the solution loop
+        self.RunBeforeSolutionLoop()
 
         ## Stepping and time settings
         self.end_time = self.project_parameters["problem_data"]["end_time"].GetDouble()
--- a/applications/CoSimulationApplication/python_scripts/base_classes/co_simulation_coupled_solver.py
+++ b/applications/CoSimulationApplication/python_scripts/base_classes/co_simulation_coupled_solver.py
@@ -171,6 +171,16 @@ class CoSimulationCoupledSolver(CoSimulationSolverWrapper):
 
         return self.time
 
+    def BeforeSolutionLoop(self):
+        for solver in self.solver_wrappers.values():
+            solver.BeforeSolutionLoop()
+            
+        for predictor in self.predictors_list:
+            predictor.BeforeSolutionLoop()
+
+        for coupling_operation in self.coupling_operations_dict.values():
+            coupling_operation.BeforeSolutionLoop()
+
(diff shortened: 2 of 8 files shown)


maceligueta:
Two comments:
- Isn't 'before solution loop' something which could be added to the end of `Initialize`?
- As a function name, shouldn't it contain a verb (in imperative form)?

loumalouomega:
> Two comments:
> 
>     * Isn't 'before solution loop' something which could be added to the end of `Initialize`?

Yes, to be consistent should be run before RunSolutionLoop and after Initialize. I don't want to break compatibility, so right now only adding the method and keeping current behaviour.

>     * As a function name, shouldn't it contain a verb (in imperative form)?

Like RunBeforeSolutionLoop?

maceligueta:
> > Two comments:
> > ```
> > * Isn't 'before solution loop' something which could be added to the end of `Initialize`?
> > ```
> 
> Yes, to be consistent should be run before RunSolutionLoop and after Initialize. I don't want to break compatibility, so right now only adding the method and keeping current behaviour.
I meant at the end of Initialize but inside Initialize. 
> 
> > ```
> > * As a function name, shouldn't it contain a verb (in imperative form)?
> > ```
> 
> Like RunBeforeSolutionLoop?
Sure

ddiezrod:
For me the important part is that processes.ExecuteBeforeSolutionLoop is called after time is set, so after https://github.com/KratosMultiphysics/Kratos/blob/df854e335117c12aed33d3d942cbc425861c2d77/kratos/python_scripts/analysis_stage.py#L108

loumalouomega:
> For me the important part is that processes.ExecuteBeforeSolutionLoop is called after time is set, so after
> 
> https://github.com/KratosMultiphysics/Kratos/blob/df854e335117c12aed33d3d942cbc425861c2d77/kratos/python_scripts/analysis_stage.py#L108

In this PR I am not trying to change behaviour

kratos/python_scripts/analysis_stage.py
```diff
@@ -54,6 +54,15 @@ def KeepAdvancingSolutionLoop(self):
         """
         return self.time < self.end_time
 
+    def RunBeforeSolutionLoop(self):
```
loumalouomega, line 57: Like in here @maceligueta ?

loumalouomega:
> > 
> > Like RunBeforeSolutionLoop?
> 
> Sure

See my comment

maceligueta:
> > > Like RunBeforeSolutionLoop?
> > 
> > 
> > Sure
> 
> See my comment

I meant here: https://github.com/KratosMultiphysics/Kratos/blob/1f43350abefb08dfc3a02d32b945bf2b4a0146e2/applications/CoSimulationApplication/python_scripts/base_classes/co_simulation_coupled_solver.py#L174

loumalouomega:
> > > > Like RunBeforeSolutionLoop?
> > > 
> > > 
> > > Sure
> > 
> > 
> > See my comment
> 
> I meant here:
> 
> https://github.com/KratosMultiphysics/Kratos/blob/1f43350abefb08dfc3a02d32b945bf2b4a0146e2/applications/CoSimulationApplication/python_scripts/base_classes/co_simulation_coupled_solver.py#L174

The coupled solver is like a python solver, so the methods should have the same name as in a python solver

PR #10298, opened 2022-09-22
[GeoMechanicsApplication] Progress reporting in geomechanics flow analysis

carloslubbers:
**Description**
Adds callback methods to flow analysis which can report back progress (numerical and textual) to the caller.

**Changelog**
Adds the following parameters to the flow analysis method:
* ReportProgress(std::function<void(double)>)
* ReportTextualProgress(std::function<void(char*)>)


mcgicjn2 (changes requested):
Could you add some progress tests, or point to where it's functionality is tested. I see the test is updated, but don't see a test for the progress reporting. 

@carloslubbers I could add this if required.

mcgicjn2 (dismissed):

mcgicjn2 (approved):
These changes are good from the GeoMechanicsApplication stand point

PR #10301, opened 2022-09-22
[Core] Update IntervalUtility

matekelemen:
- **behaviour change**: default interval begin is now the most negative value that double can represent instead of 0
- **behaviour change**: when passing ```"End"``` to the interval, it is converted to the maximum value that ```double``` can represent, instead of ```1e30```
- add ```DiscreteIntervalUtility``` (```IntervalUtility``` for ```ints```)
- move non-performance-critical code to sources
- fix inconsistencies around interval boundaries
- add ```"Begin"``` as an option to the interval begin
- add tests

depends on #10052
adopts extra tests from #10243


matekelemen:
FYI @miguelmaso (sorry, messed up the base branch in the previous PR)
I copy-pasted your tests. There's only one assertion that I removed due to the behaviour change (default interval begin is now ```std::numeric_limits<double>::lowest``` instead of 0).

miguelmaso:
> FYI @miguelmaso (sorry, messed up the base branch in the previous PR)

This gives me peace of mind. I'm not the only one who sometimes mess up something

miguelmaso (commented):
I'm in favour of this PR, especially because it fixes some inconsistencies around boundaries. Moreover, implementation is very clean

Waiting to approve after dependencies have been merged

matekelemen:
Ready for review @miguelmaso @KratosMultiphysics/technical-committee

PR #10303, opened 2022-09-23
[Core] Consistent number retrieve in Parameters

### LINKED ISSUE #10302: Inconsistent Parameter Type for Floats
## TL;DR
```Parameters::IsDouble``` returns ```false``` for integers, but ```Parameters::GetDouble``` parses the value without exceptions anyway.

## Details
The question is how the following piece of code should behave:
```py
parameters = KratosMultiphysics.Parameters("""{"a_numeric_value" : 42}""")
parameters["a_numeric_value"].IsDouble()
parameters["a_numeric_value"].GetDouble()
```
Currently, ```IsDouble()``` returns ```false```. While this guarantees that exactly one of ```Parameters::Is...()``` returns ```true``` for any valid input (== no ambiguity about the type of any value), it makes parameter checking/validation a bit more messy (see de422c0f4c5719ae5a20649829f7b4640cbb0c5b).

However, ```GetDouble()``` yields ```42.0``` without issues, which is a bit surprising. In general, if an object is not of type ```T```, it shouldn't be possible to parse it as ```T``` directly.

## Opinion
I don't think the type system imposed on JSON structures should be too strict but it should absolutely be consistent, so I'd change ```IsDouble()``` to return ```true``` for integers as well. I know this behaviour comes directly from ```nlohmann::json```, but I don't see why ```Parameters``` should behave identically to it. This only implies the following change:

```kratos_parameters.cpp```
```diff
bool Parameters::IsDouble() const
{
-    return mpValue->is_number_float();
+    return mpValue->is_number();
}
```

The alternative would that ```IsDouble()``` yields ```false``` and ```GetDouble()``` throws an exception, which isn't a great idea in my opinion.

loumalouomega:
**📝 Description**

Fixes #10302. This makes consistent the `Is` and `Get` methods for double and integers

**🆕 Changelog**

- [Consistent number retrieve in Parameters](https://github.com/KratosMultiphysics/Kratos/commit/ed07f9dfcd8f884dba71c506e2b875eec18d05ba)


matekelemen (commented):
Fine by me, but this is technically a behaviour change and will break a couple of tests (speaking from experience 😢).

loumalouomega:
> Fine by me, but this is technically a behaviour change and will break a couple of tests (speaking from experience 😢).

I know, but this is more consistent, otherwise we can do what you suggest, but it will be more inconsistent.

RiccardoRossi (changes requested):
blocking since you are chaning the behaviour while the current one is by design

loumalouomega:
> blocking since you are chaning the behaviour while the current one is by design

Yes I know, see my comment

loumalouomega:
I applied @matekelemen suggestion

kratos/sources/kratos_parameters.cpp
```diff
@@ -542,7 +542,7 @@ double Parameters::GetDouble() const
 
 int Parameters::GetInt() const
 {
-    KRATOS_ERROR_IF_NOT(mpValue->is_number()) << "Argument must be a number" << std::endl;
+    KRATOS_ERROR_IF_NOT(mpValue->is_number_integer()) << "Argument must be an integer number" << std::endl;
```
loumalouomega:
```suggestion
    KRATOS_ERROR_IF_NOT(mpValue->is_number()) << "Argument must be a number" << std::endl;
```
Revert

kratos/sources/kratos_parameters.cpp
```diff
@@ -438,7 +438,7 @@ bool Parameters::IsNumber() const
 
 bool Parameters::IsDouble() const
 {
-    return mpValue->is_number_float();
```
RiccardoRossi, line 441: why are u changing this?

it was correct before...

in essence getDouble is like doing a static_cast<double>

loumalouomega:
@RiccardoRossi @matekelemen are we happy with the proposal changes?

kratos/sources/kratos_parameters.cpp
loumalouomega: See https://github.com/KratosMultiphysics/Kratos/issues/10302

loumalouomega: In addition to that, there is a comment in the change

matekelemen (commented):
This is exactly what I want so I'm happy with it, but it's ultimately up to what the technical committee thinks.

The reason why I think this is the way to go is validation: if we want to catch bugs due to misconfiguration as early as possible, the first thing we can do is validate the input each object gets. If we ever want to implement a robust input validation scheme, we need a consistent, logical system, and a clear separation of responsibilities.

```Parameters``` is our de-facto input interface, so imagine we want to instantiate an object with it. ```Parameters``` cannot validate the **value** of each of its items (that's the object's responsibility), but it *can* validate their **types** based on a default ```Parameters``` instance for the given object. Systematically, if an input passes **type** validation, the object doesn't have to check that again and can continue validating the **values** of the items within. For this to be reliable though, we need to be consistent about querying the **type** of an item (```Parameters::Is...()```) and parsing its **value** (```Parameters::Get...()```).

The current behaviour in master is that ```Parameters::IsInt()``` and ```Parameters::IsDouble()``` can fail while ```Parameters::GetInt()``` and ```Parameters::GetDouble()``` succeed for the same input. As a result, if we want to add systematic validation to existing objects, they'll likely break tests/user scripts that otherwise work correctly. To avoid that, we need to implement workarounds for ```int``` and ```double```, which leads to code duplication, increased complexity, and makes maintenance more painful.

loumalouomega:
> This is exactly what I want so I'm happy with it, but it's ultimately up to what the technical committee thinks.
> 
> The reason why I think this is the way to go is validation: if we want to catch bugs due to misconfiguration as early as possible, the first thing we can do is validate the input each object gets. If we ever want to implement a robust input validation scheme, we need a consistent, logical system, and a clear separation of responsibilities.
> 
> `Parameters` is our de-facto input interface, so imagine we want to instantiate an object with it. `Parameters` cannot validate the **value** of each of its items (that's the object's responsibility), but it _can_ validate their **types** based on a default `Parameters` instance for the given object. Systematically, if an input passes **type** validation, the object doesn't have to check that again and can continue validating the **values** of the items within. For this to be reliable though, we need to be consistent about querying the **type** of an item (`Parameters::Is...()`) and parsing its **value** (`Parameters::Get...()`).
> 
> The current behaviour in master is that `Parameters::IsInt()` and `Parameters::IsDouble()` can fail while `Parameters::GetInt()` and `Parameters::GetDouble()` succeed for the same input. As a result, if we want to add systematic validation to existing objects, they'll likely break tests/user scripts that otherwise work correctly. To avoid that, we need to implement workarounds for `int` and `double`, which leads to code duplication, increased complexity, and makes maintenance more painful.

Thanks for the explanation FYI @RiccardoRossi

loumalouomega:
As #10302 is closed I close this

RiccardoRossi:
well here i added only my point of view. mentioning @KratosMultiphysics/technical-committee in case the view is different

loumalouomega:
Any agreement? @RiccardoRossi

loumalouomega:
> Any agreement? @RiccardoRossi

Hello 👋

RiccardoRossi (commented):
@KratosMultiphysics/technical-committee  agrees that a double is a double and an int is not ... 

that is, sorry but we disagree with this behaviour change

PR #10311, opened 2022-09-25
[Core][GeometricalObject] Add IsActive

philbucher:
**📝 Description**
Adding `IsActive` to `GeometricalObject` (and hence to `Element` and `Condition`)

I think this makes sense since this is duplicated all over the place (12 times only in the Block B&S)

I will add tests if this gets accepted

FYI @KratosMultiphysics/geomechanics I think you use this quite a lot


RiccardoRossi:
honestly we should remove the need of "isdefined" (although admittedly it is my bad ot is there).

that way flags would occupy one double instead of 2

kratos/sources/geometrical_object.cpp
```diff
@@ -0,0 +1,28 @@
...
+// System includes
+
+// External includes
+
+// Project includes
+#include "includes/geometrical_object.h"
+#include "includes/kratos_flags.h"
+
+namespace Kratos {
+
```
loumalouomega, line 22: I would move more things to the cpp to reduce compilation times

loumalouomega (commented):
Okay from my side

RiccardoRossi:
i would also be ok with  this. @KratosMultiphysics/technical-committee?

rubenzorrilla:
> i would also be ok with this. @KratosMultiphysics/technical-committee?

Makes also sense for me.

philbucher:
Pung @roigcarlo @sunethwarna @pooyan-dadvand

kratos/sources/geometrical_object.cpp
pooyan-dadvand: But in a separate PR. It is important to keep the code reorganization PR separate from the otheres.

pooyan-dadvand:
BTW, why we need to check if is defined?

RiccardoRossi:
if it is not defined it would default to false...

philbucher:
Any more comments?
I would like to use this in some refactorings :)

kratos/sources/geometrical_object.cpp
```diff
@@ -0,0 +1,28 @@
...
+bool GeometricalObject::IsActive() const 
+{
+    return IsDefined(ACTIVE) ? Is(ACTIVE) : true;
```
pooyan-dadvand, line 25: In checking for true, the `IsDefined` call is redundant. The `IsDefined` shall be use if we want to check for false in which we don't know if the flag is unset or is really set to false.

philbucher: I am sorry I am a bit lost :sweat_smile: 
Can you please post your suggestion?

pooyan-dadvand: It is correct! Sorry for the noise...

pooyan-dadvand (approved):
I am approving as @KratosMultiphysics/technical-committee members already gave their thumbs ups!

PR #10312, opened 2022-09-26
[Adjoint][Fluid] Adjoint advance in time minors

rubenzorrilla:
**📝 Description**
Following @sunethwarna indications, in here I'm slightly changing the time marching of the adjoint fluid solvers. The main changes are
- Now we use `start_time` and `end_time` as input parameters (no more `nsteps` and start/end step).
- The `time_step` is now a positive magnitude and is internally reversed by the adjoint solver.

Long story short, thanks to this the user doesn't need to change the primal problem time stepping settings to solve the adjoint one. Also note that we first check if the old input is provided to throw the corresponding message to the user.

As soon as we merge this, I will update the cases in the example repository.

Related to #9592.


sunethwarna (approved):
Nice... Thanks @rubenzorrilla :)

PR #10322, opened 2022-09-29
[Core] Moving GetDefaultParameters() and Name() from Access to Inquiry block in ImplicitSolvingStrategy

loumalouomega:
**📝 Description**

Moving GetDefaultParameters() and Name() from Access to Operations block in ImplicitSolvingStrategy. Also using pragma once (the t is not capital, so I could correct it or just replace it with pragma once)

**🆕 Changelog**

- [Moving GetDefaultParameters() and Name() from Access to Operations block in ImplicitSolvingStrategy](https://github.com/KratosMultiphysics/Kratos/commit/8782bc137f232da85119cbb30884d03bd526dbf6)
- Also using pragma once

### DIFF
--- a/kratos/solving_strategies/strategies/implicit_solving_strategy.h
+++ b/kratos/solving_strategies/strategies/implicit_solving_strategy.h
@@ -11,8 +11,7 @@
 //
 //
 
-#if !defined(KRATOS_IMPLICIt_SOLVING_STRATEGY)
-#define KRATOS_IMPLICIt_SOLVING_STRATEGY
+#pragma once
 
 /* System includes */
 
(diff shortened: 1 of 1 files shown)


rubenzorrilla:
If we're moving them I think these would be better placed at the `Inquiry` section.

loumalouomega:
> If we're moving them I think these would be better placed at the `Inquiry` section.

True

loumalouomega:
Done

loumalouomega:
> Done

@rubenzorrilla

rubenzorrilla (approved):

PR #10325, opened 2022-09-29
[CoSimulationApplication] Improve generic Kratos wrapper

loumalouomega:
**📝 Description**

Improving generic Kratos wrapper. This wrapper can generate on the fly a Kratos wrapper by providing the analysis stage file.

For example:

~~~json
"type" : "KratosMultiphysics.CoSimulationApplication.solver_wrappers.kratos.convection_diffusion_wrapper",
"solver_wrapper_settings" : {
    "analysis_stage_module" : "KratosMultiphysics.ConvectionDiffusionApplication.convection_diffusion_analysis"
}
~~~

can be replaced with:

~~~json
"type" : "solver_wrappers.kratos.kratos_base_wrapper",
"solver_wrapper_settings" : {
    "analysis_stage_module" : "KratosMultiphysics.ConvectionDiffusionApplication.convection_diffusion_analysis",
	"analysis_stage_name"   : "ConvectionDiffusionAnalysis",
    "input_file"  : "ProjectParameters"
}
~~~

or simply with:

~~~json
"type" : "solver_wrappers.kratos.kratos_base_wrapper",
"solver_wrapper_settings" : {
    "analysis_stage_module" : "KratosMultiphysics.ConvectionDiffusionApplication.convection_diffusion_analysis",
    "input_file"  : "ProjectParameters"
}
~~~

Because if `analysis_stage_name` is not provided, then it will try to determine the name of the analysis class by [pascalicing](https://www.theserverside.com/answer/Pascal-case-vs-camel-case-Whats-the-difference) the name of the module.

Of course, this only works if your wrapper doesn't do any relevant change in the base class.

**🆕 Changelog**

- [Enhance the wrapper](https://github.com/KratosMultiphysics/Kratos/pull/10325/commits/4051254c7aa20c217c5185dfa528a1ffa3a4e4ed)


applications/CoSimulationApplication/python_scripts/base_classes/co_simulation_solver_wrapper.py
```diff
@@ -160,6 +160,8 @@ def _GetDataCommunicator(self):
     def _GetDefaultParameters(cls):
         return KM.Parameters("""{
             "type"                    : "",
+            "analysis_module"         : "",
```
rubenzorrilla: This field already exist in the current `ProjectParameters.json` and we call it `analysis_stage`. I suggest to be consistent with that in here.

applications/CoSimulationApplication/python_scripts/base_classes/co_simulation_solver_wrapper.py
```diff
@@ -160,6 +160,8 @@ def _GetDataCommunicator(self):
...
+            "analysis_name"           : "",
```
rubenzorrilla: I wouldn't introduce this field. After assuming that each Python module implements only one class, you can "pascalize" the module name to take it (OFC we are assuming a naming convention such that `fsi_analysis` module implements `FsiAnalysis` class). Note that this is what we already did in order to unify the `MainKratos.py`.

applications/CoSimulationApplication/python_scripts/base_classes/co_simulation_solver_wrapper.py
loumalouomega: Done

applications/CoSimulationApplication/python_scripts/base_classes/co_simulation_solver_wrapper.py
loumalouomega: I do this because there are some which are not consistent in our side (few files, but anyway). I can remove it, for me is simpler, but this is more generic

philbucher:
Pls give me the weekend to review :)

loumalouomega:
> Pls give me the weekend to review :)

No hurries, I already implemented this in our side

loumalouomega:
> Pls give me the weekend to review :)

There is a big controversy about adding `_GetDataCommunicator` or not, what is your opinion about this?, to add a block like this one?

~~~py
    def _GetDataCommunicator(self):
        if not KM.IsDistributedRun():
            return KM.ParallelEnvironment.GetDataCommunicator("Serial")

        # now we know that Kratos runs in MPI
        parallel_type = self.project_parameters["problem_data"]["parallel_type"].GetString()

        # first check if the solver uses MPI
        if parallel_type != "MPI":
            return data_communicator_utilities.GetRankZeroDataCommunicator()

        # now we know that the solver uses MPI, only question left is whether to use all ranks or a subset
        self._CheckDataCommunicatorIsConsistentlyDefined(self.project_parameters["solver_settings"]["model_import_settings"], self.settings["mpi_settings"])

        return super()._GetDataCommunicator()
~~~

loumalouomega:
> > you can already use a custom analysis stage:
> > https://github.com/KratosMultiphysics/Kratos/blob/97e59b1b25854b618a3c04b3f143511d44639253/applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py#L99-L104
> > 
> > this is what you need right?
> 
> Yes, but if the analysis stage doesn't have Create method it will not work

I will refactor

philbucher:
> > you can already use a custom analysis stage:
> > https://github.com/KratosMultiphysics/Kratos/blob/97e59b1b25854b618a3c04b3f143511d44639253/applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py#L99-L104
> > 
> > this is what you need right?
> 
> Yes, but if the analysis stage doesn't have Create method it will not work

true, but I think it should be easy to add
same concept as for processes

loumalouomega:
> > > you can already use a custom analysis stage:
> > > https://github.com/KratosMultiphysics/Kratos/blob/97e59b1b25854b618a3c04b3f143511d44639253/applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py#L99-L104
> > > 
> > > this is what you need right?
> > 
> > 
> > Yes, but if the analysis stage doesn't have Create method it will not work
> 
> true, but I think it should be easy to add same concept as for processes

Huuuuum, I agree, but in our custom version I think I will use the trick I presented here

loumalouomega:
> > > > you can already use a custom analysis stage:
> > > > https://github.com/KratosMultiphysics/Kratos/blob/97e59b1b25854b618a3c04b3f143511d44639253/applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py#L99-L104
> > > > 
> > > > this is what you need right?
> > > 
> > > 
> > > Yes, but if the analysis stage doesn't have Create method it will not work
> > 
> > 
> > true, but I think it should be easy to add same concept as for processes
> 
> Huuuuum, I agree, but in our custom version I think I will use the trick I presented here

(mostly because right now the analysis don't implement the Create)

loumalouomega:
Let me present an alternative solution

loumalouomega:
> Let me present an alternative solution

@philbucher do you agree with the alternative?, I just improve current one

philbucher:
I would be ok with it, although I don't love it tbh

Why cannot you add the Create methods? This would the the standard approach that we apply for more and more things

loumalouomega:
> I would be ok with it, although I don't love it tbh
> 
> Why cannot you add the Create methods? This would the the standard approach that we apply for more and more things

You don't always have the control of the app, and this is more flexible. I added a warning, so if the user is also a developer of the app can add the Create method

loumalouomega:
> > I would be ok with it, although I don't love it tbh
> > Why cannot you add the Create methods? This would the the standard approach that we apply for more and more things
> 
> You don't always have the control of the app, and this is more flexible. I added a warning, so if the user is also a developer of the app can add the Create method

Do you agree @philbucher ?

applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py
```diff
@@ -98,8 +102,21 @@ def _CreateAnalysisStage(self):
...
     def __GetAnalysisStage(self):
         if self.settings["solver_wrapper_settings"].Has("analysis_stage_module"):
-            analysis_stage_module = import_module(self.settings["solver_wrapper_settings"]["analysis_stage_module"].GetString())
-            return analysis_stage_module.Create(self.model, self.project_parameters)
+            module_name = self.settings["solver_wrapper_settings"]["analysis_stage_module"].GetString()
+            analysis_stage_module = import_module(module_name)
+            if hasattr(analysis_stage_module, "Create"):
+                return analysis_stage_module.Create(self.model, self.project_parameters)
+            else:
+                KM.Logger.PrintWarning("KratosBaseWrapper", "The analysis_stage_module (" + module_name + ") does not have a \"Create\" function, trying to create the AnalysisStage directly...")
```
philbucher:
```suggestion
                KM.Logger.PrintWarning("KratosBaseWrapper", f'The analysis_stage_module "{module_name}" does not have a "Create" function, trying to create the AnalysisStage directly...')
```

applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py
```diff
@@ -98,8 +102,21 @@ def _CreateAnalysisStage(self):
...
+                if self.settings["solver_wrapper_settings"].Has("analysis_name"):
+                    analysis_stage_name = self.settings["solver_wrapper_settings"]["analysis_name"].GetString()
+                else:
+                    # We assume that the name of the AnalysisStage is the same as the name of the module in PascalCase instead of cammel_case
+                    file_name = module_name.split(".")[-1]
+                    # Convert Snake case to Pascal case
+                    analysis_stage_name = string.capwords(file_name.replace("_", " ")).replace(" ", "")
+                analysis = getattr(analysis_stage_module, analysis_stage_name)
```
philbucher: please add a proper error here if this does not exist

philbucher (commented):
> You don't always have the control of the app, and this is more flexible. I added a warning, so if the user is also a developer of the app can add the Create method

fair enough, this is a good argument

Minor comments, then I will approve

applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py
loumalouomega: Okay

loumalouomega: Done

applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py
```diff
@@ -98,8 +102,25 @@ def _CreateAnalysisStage(self):
...
+            module_name = self.settings["solver_wrapper_settings"]["analysis_stage_module"].GetString()
+            analysis_stage_module = import_module(module_name)
+            if hasattr(analysis_stage_module, "Create"):
+                return analysis_stage_module.Create(self.model, self.project_parameters)
+            else:
+                KM.Logger.PrintWarning("KratosBaseWrapper", f'The analysis_stage_module "{module_name}" does not have a "Create" function, trying to create the AnalysisStage directly...')
+                if self.settings["solver_wrapper_settings"].Has("analysis_name"):
+                    analysis_stage_name = self.settings["solver_wrapper_settings"]["analysis_name"].GetString()
+                else:
+                    # We assume that the name of the AnalysisStage is the same as the name of the module in PascalCase instead of cammel_case
```
rubenzorrilla:
```suggestion
                    # We assume that the name of the AnalysisStage is the same as the name of the module in PascalCase instead of snake_case
```

applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py
```diff
@@ -98,8 +102,25 @@ def _CreateAnalysisStage(self):
...
-            return analysis_stage_module.Create(self.model, self.project_parameters)
+            module_name = self.settings["solver_wrapper_settings"]["analysis_stage_module"].GetString()
+            analysis_stage_module = import_module(module_name)
+            if hasattr(analysis_stage_module, "Create"):
+                return analysis_stage_module.Create(self.model, self.project_parameters)
+            else:
+                KM.Logger.PrintWarning("KratosBaseWrapper", f'The analysis_stage_module "{module_name}" does not have a "Create" function, trying to create the AnalysisStage directly...')
+                if self.settings["solver_wrapper_settings"].Has("analysis_name"):
+                    analysis_stage_name = self.settings["solver_wrapper_settings"]["analysis_name"].GetString()
+                else:
```
rubenzorrilla: I think it should be in the other way around. You assume the name is the module name in UpperCamelCase. If not you search for a custom one. This is the same we assume in the standard MainKratos.py.

loumalouomega: I don't understand, what should I check?

applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py
```diff
@@ -98,8 +102,25 @@ def _CreateAnalysisStage(self):
...
-            analysis_stage_module = import_module(self.settings["solver_wrapper_settings"]["analysis_stage_module"].GetString())
-            return analysis_stage_module.Create(self.model, self.project_parameters)
+            module_name = self.settings["solver_wrapper_settings"]["analysis_stage_module"].GetString()
+            analysis_stage_module = import_module(module_name)
+            if hasattr(analysis_stage_module, "Create"):
+                return analysis_stage_module.Create(self.model, self.project_parameters)
+            else:
+                KM.Logger.PrintWarning("KratosBaseWrapper", f'The analysis_stage_module "{module_name}" does not have a "Create" function, trying to create the AnalysisStage directly...')
+                if self.settings["solver_wrapper_settings"].Has("analysis_name"):
+                    analysis_stage_name = self.settings["solver_wrapper_settings"]["analysis_name"].GetString()
```
rubenzorrilla:
```suggestion
                    analysis_stage_name = self.settings["solver_wrapper_settings"]["analysis_stage_name"].GetString()
```
For consistency with the `analysis_stage_module` field. Besides, as I pointed before, in our current ProjectParameters.json this field is named `analysis_stage`. I think it would be good to use the same key.

applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py
rubenzorrilla: You do a try with what you put in the else. If doesn't succeed, you try to parse a custom analysis name.

applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py
rubenzorrilla: Please, note my comment. I think we should agree towards standarizing this with current `ProjectParameters.json`.

applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py
loumalouomega: But I think it more consistent the other way around, if my name is not "standard" I will provide it manually, otherwise I will be doing two checks instead of one

applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py
loumalouomega: Yes, but the current key was already there, if @philbucher agrees I will change it

rubenzorrilla: OK. Now I see. You meant changing `analysis_stage_module`. If this is the case, the discussion is not that easy. Lets leave it as it is for the moment.

rubenzorrilla: (resolving again)

applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py
rubenzorrilla: The point is that as it is right now, I can provide the correct naming and nothing will happen (e.g. I provide `ToteAnalysisStage` from `tote_analysis_stage` module). So we're implicitly promoting using the extra `analysis_stage_name` field.

If we do in the other way  around, we only retrieve the `analysis_stage_name` if the class naming is not standard, so you can throw a warning saying something like a custom analysis stage name is being used and that the convention is bla bla bla so in the long run you can safely remove support of the `analysis_stage_nanme`.

loumalouomega: Okay....

loumalouomega: Huuuuuum.....

loumalouomega: Done

applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py
```diff
@@ -108,18 +108,24 @@ def __GetAnalysisStage(self):
...
+
                 # Getting the analysis class
                 if hasattr(analysis_stage_module, analysis_stage_name):
                     analysis = getattr(analysis_stage_module, analysis_stage_name)
                 else:
-                    Exception("\"" + module_name + "\" does not have a \"" + analysis_stage_name + "\" class!")
+                    if self.settings["solver_wrapper_settings"].Has("analysis_name"):
+                        analysis_stage_name = self.settings["solver_wrapper_settings"]["analysis_stage_name"].GetString()
+                        if hasattr(analysis_stage_module, analysis_stage_name):
+                            analysis = getattr(analysis_stage_module, analysis_stage_name)
```
rubenzorrilla, line 123: Here I'd throw a warning saying that you're using a custom name that doesn't follow the convention.

applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py
```diff
@@ -108,18 +108,24 @@ def __GetAnalysisStage(self):
...
+                file_name = module_name.split(".")[-1]
+                # Convert Snake case to Pascal case
+                analysis_stage_name = string.capwords(file_name.replace("_", " ")).replace(" ", "")
+
                 # Getting the analysis class
                 if hasattr(analysis_stage_module, analysis_stage_name):
                     analysis = getattr(analysis_stage_module, analysis_stage_name)
                 else:
-                    Exception("\"" + module_name + "\" does not have a \"" + analysis_stage_name + "\" class!")
+                    if self.settings["solver_wrapper_settings"].Has("analysis_name"):
```
rubenzorrilla, line 121: Indeed, I throw the warning saying that a non standard analysis stage name is to be used  in here.

loumalouomega:
> > Done @rubenzorrilla
> 
> I don't see the warning. Maybe GitHub messed up some changes.

Line 120

applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py
```diff
@@ -98,8 +102,32 @@ def _CreateAnalysisStage(self):
...
+                if hasattr(analysis_stage_module, analysis_stage_name):
+                    analysis = getattr(analysis_stage_module, analysis_stage_name)
+                else:
+                    KM.Logger.PrintWarning("KratosBaseWrapper", f'The analysis_stage_module "{module_name}" does not follow the standard way to define the analysis stage name "{analysis_stage_name}" . Trying to retrieve from a custom definition')
+                    if self.settings["solver_wrapper_settings"].Has("analysis_name"):
+                        analysis_stage_name = self.settings["solver_wrapper_settings"]["analysis_stage_name"].GetString()
+                        if hasattr(analysis_stage_module, analysis_stage_name):
+                            analysis = getattr(analysis_stage_module, analysis_stage_name)
+                        else:
+                            Exception("\"" + module_name + "\" does not have a \"" + analysis_stage_name + "\" class!")
```
philbucher: please use f-strings

and `raise` is missing

rubenzorrilla:
> > > Done @rubenzorrilla
> > 
> > 
> > I don't see the warning. Maybe GitHub messed up some changes.
> 
> Line 120

Ouch, I didn't see that yesterday. Having said this, this is fine for me. I let @philbucher to give the final approval. Thanks @loumalouomega .

applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py
```diff
@@ -98,8 +102,32 @@ def _CreateAnalysisStage(self):
...
+                # We assume that the name of the AnalysisStage is the same as the name of the module in PascalCase instead of snake_case
+                file_name = module_name.split(".")[-1]
+                # Convert Snake case to Pascal case
+                analysis_stage_name = string.capwords(file_name.replace("_", " ")).replace(" ", "")
+
+                # Getting the analysis class
+                if hasattr(analysis_stage_module, analysis_stage_name):
+                    analysis = getattr(analysis_stage_module, analysis_stage_name)
+                else:
+                    KM.Logger.PrintWarning("KratosBaseWrapper", f'The analysis_stage_module "{module_name}" does not follow the standard way to define the analysis stage name "{analysis_stage_name}" . Trying to retrieve from a custom definition')
```
loumalouomega, line 120: It is here @rubenzorrilla

loumalouomega:
This is ready @philbucher

philbucher (approved):

PR #10337, opened 2022-10-03
[Core] MLS shape functions enhancements

rubenzorrilla:
**📝 Description**
In this PR I'm enhancing the current MLS shape functions capabilities. The main improvements are:
- Avoiding the transpose matrix multiplication in the L2 minimization problem. Instead a QR decomposition is used to solve the rectangular matrix. This should make the resolution of the problem more robust in presence of ill-conditioned cases.
- Adding the capability of using a 2nd order interpolation basis.
- Kernel function simplification (no need to normalize with Pi).

Note that these required quite a lot of refactoring (this is why there are so much changes).

@frawahl FYI, this are the changes I told you long time ago.

### DIFF
--- a/kratos/utilities/mls_shape_functions_utility.h
+++ b/kratos/utilities/mls_shape_functions_utility.h
@@ -12,9 +12,7 @@
 //                   Zhiming Guo
 //
 
-#if !defined(KRATOS_MLS_SHAPE_FUNCTIONS_UTILITY_H_INCLUDED)
-#define  KRATOS_MLS_SHAPE_FUNCTIONS_UTILITY_H_INCLUDED
-
+#pragma once
 
 // System includes
 
@@ -23,7 +21,7 @@
 // Project includes
 #include "containers/array_1d.h"
 #include "includes/define.h"
-#include "spaces/ublas_space.h"
+#include "includes/ublas_interface.h"
 
 namespace Kratos
 {
@@ -181,5 +179,3 @@ public:
 };
 
 }  // namespace Kratos.
-
-#endif // KRATOS_MLS_SHAPE_FUNCTIONS_UTILITY_H_INCLUDED  defined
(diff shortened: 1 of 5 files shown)


kratos/utilities/mls_shape_functions_utility.h
```diff
@@ -37,6 +38,8 @@ class MLSShapeFunctionsUtility
 
 public:
 
+    using DenseSpace = UblasSpace<double, Matrix, Vector>;
```
philbucher: Seems unused here
Maybe move to cpp?

Or did you mean to use the ublas interface instead?

rubenzorrilla: I think I added it when trying different QR decompositions. Removing it.

rubenzorrilla: OFC, it is used in the cpp to instantiate the QR :facepalm:

RiccardoRossi (commented):
why are u changing from stack matrices and vectors to tje dynqmoc versions?

rubenzorrilla:
> why are u changing from stack matrices and vectors to tje dynqmoc versions?



> why are u changing from stack matrices and vectors to tje dynqmoc versions?

Note that we're changing the way we calculate the approximants (no transpose multiplication. Now it is impossible as the problem size is dependent on the  number of points in the cloud.

kratos/utilities/mls_shape_functions_utility.h
```diff
@@ -23,6 +23,7 @@
 // Project includes
 #include "containers/array_1d.h"
 #include "includes/define.h"
+#include "spaces/ublas_space.h"
```
philbucher:
```suggestion
#include "includes/define.h"
#include "includes/ublas_interface.h"
```

the ublas_space should be moved to the cpp

rubenzorrilla: Done.

RiccardoRossi (approved):
9k to me

PR #10338, opened 2022-10-04
[Core] Some performance improvements in ApplyPeriodicConditionProcess

loumalouomega:
**📝 Description**

Minor performance improvement in ApplyPeriodicConditionProcess. Moving declaration to outside loop to improve memory cache.

**🆕 Changelog**

- [Minor performance improvement in ApplyPeriodicConditionProcess](https://github.com/KratosMultiphysics/Kratos/commit/b8ebcdf388bcc6ea2c942c2977c9b1e99b29144a)


kratos/processes/apply_periodic_boundary_condition_process.cpp
```diff
@@ -149,38 +149,30 @@ template <int TDim>
...
 
-    IndexType num_slaves_found = 0;
-
-    num_slaves_found = block_for_each<SumReduction<IndexType>>(mrSlaveModelPart.Nodes(), [&](Node<3>& rNode){
-        IndexType counter = 0;
-        Condition::Pointer p_host_cond;
-        VectorType shape_function_values;
-        array_1d<double, 3 > transformed_slave_coordinates;
-        TransformNode(rNode.Coordinates(), transformed_slave_coordinates);
+    struct auxiliary {IndexType counter = 0; Condition::Pointer p_host_cond; VectorType shape_function_values; array_1d<double, 3 > transformed_slave_coordinates;};
```
philbucher:
```suggestion
    struct auxiliar {IndexType counter = 0; Condition::Pointer p_host_cond; VectorType shape_function_values; array_1d<double, 3 > transformed_slave_coordinates;};
```

rubenzorrilla: Jokes aside, I'd call this `TLS` or `TLS_container`.

rubenzorrilla (changes requested):
Please avoid the `{` aesthetic changes. Aside of this, I think changes are correct.

kratos/processes/apply_periodic_boundary_condition_process.cpp
```diff
@@ -149,38 +149,30 @@ template <int TDim>
...
+            for (unsigned int j = 0; j < num_vars; j++) {
                 const std::string var_name = mParameters["variable_names"][j].GetString();
                 // Checking if the variable is a vector variable
-                if (KratosComponents<Variable<array_1d<double, 3>>>::Has(var_name))
-                {   // TODO: Look for a better alternative to do this.
-                    ConstraintSlaveNodeWithConditionForVectorVariable<TDim>(rNode, p_host_cond->GetGeometry() , shape_function_values, var_name);
-                } else if (KratosComponents<VariableType>::Has(var_name))
-                {
-                    ConstraintSlaveNodeWithConditionForScalarVariable<TDim>(rNode, p_host_cond->GetGeometry() , shape_function_values, var_name);
+                if (KratosComponents<Variable<array_1d<double, 3>>>::Has(var_name)) {   // TODO: Look for a better alternative to do this.
```
philbucher: This is expensive 
Can it be done outside the loop?

loumalouomega: That's why there is a TODO, I can guess a way

rubenzorrilla:
> > Please avoid the `{` aesthetic changes. Aside of this, I think changes are correct.
> 
> But that is our style...

There is no mention in [here](https://github.com/KratosMultiphysics/Kratos/wiki/Style-Guide).

loumalouomega:
> > > Please avoid the `{` aesthetic changes. Aside of this, I think changes are correct.
> > 
> > 
> > But that is our style...
> 
> There is no mention in [here](https://github.com/KratosMultiphysics/Kratos/wiki/Style-Guide).

We have been using this style for a long time... (@roigcarlo @philbucher do you remember when we set the bracket style?)

loumalouomega:
> > > > Please avoid the `{` aesthetic changes. Aside of this, I think changes are correct.
> > > 
> > > 
> > > But that is our style...
> > 
> > 
> > There is no mention in [here](https://github.com/KratosMultiphysics/Kratos/wiki/Style-Guide).
> 
> We have been using this style for a long time... (@roigcarlo @philbucher do you remember when we set the bracket style?)

(I am reverting it BTW, but also improving performance in general)

loumalouomega:
Well, what it started with a simple change to show an example to @ddiezrod it has become a full refactor, good news is that now the bottlenecks from before are no more, the bad news is that now this is a complex PR to review

kratos/processes/apply_periodic_boundary_condition_process.cpp
loumalouomega: I fixed it

rubenzorrilla:
At this point, I think that @RiccardoRossi should be the one reviewing this as he was the original author.

loumalouomega:
> At this point, I think that @RiccardoRossi should be the one reviewing this as he was the original author.

In the header also points @adityaghantasala

loumalouomega:
> > At this point, I think that @RiccardoRossi should be the one reviewing this as he was the original author.
> 
> In the header also points @adityaghantasala

BTW, I found another way to even improve more the process, but that would be too many changes (basically using the critical at the end of the loop)

kratos/processes/apply_periodic_boundary_condition_process.cpp
```diff
@@ -149,92 +150,125 @@ template <int TDim>
...
 {
     const auto timer = BuiltinTimer();
-    const int num_vars = mParameters["variable_names"].size();
+    const unsigned int num_vars = mParameters["variable_names"].size();
     BinBasedFastPointLocatorConditions<TDim> bin_based_point_locator(mrMasterModelPart);
     bin_based_point_locator.UpdateSearchDatabase();
 
-    IndexType num_slaves_found = 0;
+    // Define auxiliary functions
+    const std::function<void(NodeType&, const GeometryType&, const VectorType&, const std::vector<const VariableType*>&)> function_for_vector_variable =
```
matekelemen: I think this would be much more readable with a local alias, for example:
```cpp
using VectorFunctor = std::function<void(NodeType&, const GeometryType&, const VectorType&, const std::vector<const VariableType*>&)>;
const VectorFunctor function_for_vector_variable = [...](...){...};
...
std::vector<VectorFunctor> functions_required(...);
```

kratos/processes/apply_periodic_boundary_condition_process.cpp
```diff
@@ -149,92 +150,125 @@ template <int TDim>
...
+    [this](NodeType& rSlaveNode, const GeometryType& rHostedGeometry, const VectorType& rWeights, const std::vector<const VariableType*>& rVars) {ConstraintSlaveNodeWithConditionForVectorVariable<TDim>(rSlaveNode, rHostedGeometry, rWeights, rVars);};
+    const std::function<void(NodeType&, const GeometryType&, const VectorType&, const std::vector<const VariableType*>&)> function_for_scalar_variable =
+    [this](NodeType& rSlaveNode, const GeometryType& rHostedGeometry, const VectorType& rWeights, const std::vector<const VariableType*>& rVars) {ConstraintSlaveNodeWithConditionForScalarVariable<TDim>(rSlaveNode, rHostedGeometry, rWeights, rVars);};
+
+    // Fill an auxiliary vector with the functions
+    std::vector<const std::function<void(NodeType&, const GeometryType&, const VectorType&, const std::vector<const VariableType*>&)>*> functions_required(num_vars, nullptr);
```
matekelemen: If you have a setup where you need to allocate some objects that should be in an invalid / uninitialized state before they get assigned, you can use [```std::optional```](https://en.cppreference.com/w/cpp/utility/optional) instead of raw pointers and ```nullptr```. 

```std::optional``` is stored on the stack and only introduces one additional byte (a boolean variable that indicates whether the stored object has a valid value), so it's a minimal overhead but much safer than gambling with ```nullptr```s.

File: kratos/processes/apply_periodic_boundary_condition_process.cpp [hunk 5]
```diff
@@ -149,92 +150,125 @@ template <int TDim>
...
-            mrMasterModelPart.AddMasterSlaveConstraint(constraint1);
-            mrMasterModelPart.AddMasterSlaveConstraint(constraint2);
-            mrMasterModelPart.AddMasterSlaveConstraint(constraint3);
-            mrMasterModelPart.AddMasterSlaveConstraint(constraint4);
-            mrMasterModelPart.AddMasterSlaveConstraint(constraint5);
-            mrMasterModelPart.AddMasterSlaveConstraint(constraint6);
-
-
-            if (TDim == 3) // TODO: This function can be optimized using template specialization. if constexpr will fail due to unused variables declared
+            #pragma omp critical
```
matekelemen: @philbucher do we have something in mind to emulate ```#pragma omp critical``` generically in ```ParallelUtilities```?

kratos/processes/apply_periodic_boundary_condition_process.cpp
loumalouomega: It could be

kratos/processes/apply_periodic_boundary_condition_process.cpp
loumalouomega: Sorry, I don't get how to use optional in this situation https://www.cppstories.com/2018/05/using-optional/

kratos/processes/apply_periodic_boundary_condition_process.cpp
philbucher: this bad boy:
https://github.com/KratosMultiphysics/Kratos/blob/fef4fb6be42bf22839de8219c1fc61eef9e8d795/kratos/sources/parallel_environment.cpp#L249

Although generally I recommend to use reductions (which internally would use the same)

matekelemen: but ```LockObject``` is not implemented for *C++11* parallelism (yet?) right?

kratos/processes/apply_periodic_boundary_condition_process.cpp
matekelemen: Example:
```diff
-    std::vector<const MyFunction*> functions_required(num_vars, nullptr);
+    std::vector<std::optional<std::reference_wrapper<const MyFunction>>> functions_required(num_vars);
    std::vector<std::vector<const VariableType*>> variables_vector(num_vars, std::vector<const VariableType*>());
    for (unsigned int j = 0; j < num_vars; j++) {
        const std::string& r_var_name = mParameters["variable_names"][j].GetString();
        if (KratosComponents<Variable<array_1d<double, 3>>>::Has(r_var_name)) {
-            functions_required[j] = &function_for_vector_variable;
+            functions_required[j] = function_for_vector_variable;
            variables_vector[j].push_back(&KratosComponents<VariableType>::Get(r_var_name + "_X"));
            variables_vector[j].push_back(&KratosComponents<VariableType>::Get(r_var_name + "_Y"));
            variables_vector[j].push_back(&KratosComponents<VariableType>::Get(r_var_name + "_Z"));
        } else if (KratosComponents<VariableType>::Has(r_var_name)) {
-            functions_required[j] = &function_for_scalar_variable;
+            functions_required[j] = function_for_scalar_variable;
```
but now that I'm looking at it, you could also just use the classic ```reserve``` and ```push_back``` combo in this case, because you're just iterating contiguously.

kratos/processes/apply_periodic_boundary_condition_process.cpp
philbucher: almost nothing is fully functional with C++11 based parallelism
The big fat warning in CMake is there for a reason :D

But hopefully soon it will

kratos/processes/apply_periodic_boundary_condition_process.cpp
```diff
@@ -149,92 +150,126 @@ template <int TDim>
...
+                auto p_constraint5 = r_clone_constraint.Create(++current_num_constraint, r_master_node, r_var_y, rSlaveNode, r_var_y, master_weight * mTransformationMatrixVariable(1,1), constant_y);
+                auto p_constraint6 = r_clone_constraint.Create(++current_num_constraint, r_master_node, r_var_z, rSlaveNode, r_var_y, master_weight * mTransformationMatrixVariable(1,2), constant_y);
+
+                auto p_constraint7 = r_clone_constraint.Create(++current_num_constraint, r_master_node, r_var_x, rSlaveNode, r_var_z, master_weight * mTransformationMatrixVariable(2,0), constant_z);
+                auto p_constraint8 = r_clone_constraint.Create(++current_num_constraint, r_master_node, r_var_y, rSlaveNode, r_var_z, master_weight * mTransformationMatrixVariable(2,1), constant_z);
+                auto p_constraint9 = r_clone_constraint.Create(++current_num_constraint, r_master_node, r_var_z, rSlaveNode, r_var_z, master_weight * mTransformationMatrixVariable(2,2), constant_z);
```
matekelemen: I'd do this in a nested loop to reduce copy-pasta. ```TDim``` is known at compile time so the loops would automatically get unrolled by the compiler anyway.

not too important though

matekelemen (dismissed):

matekelemen:
what's up with the CI?

adityaghantasala:
Hi Guys ... thank you for the improvements. 
I did implement the original version of the process. But at this point, I cannot spend time for the review. So please continue with the changes suggested. 

Going with the title of the PR, I am curious to know and see how much is the performance improvement !!

matekelemen (dismissed):

loumalouomega:
@rubenzorrilla do you agree?

rubenzorrilla:
> @rubenzorrilla do you agree?

As I pointed above I'd like @RiccardoRossi to review this.

loumalouomega:
@rubenzorrilla still blocking?

loumalouomega:
@rubenzorrilla hello

loumalouomega:
> @rubenzorrilla hello

👋

loumalouomega:
@rubenzorrilla hello

loumalouomega:
@rubenzorrilla this is still blocked

loumalouomega:
> @rubenzorrilla this is still blocked

👋 @rubenzorrilla

rubenzorrilla:
> > @rubenzorrilla do you agree?
> 
> As I pointed above I'd like @RiccardoRossi to review this.

As I said in my last comment. @RiccardoRossi  is the original author of this,.

loumalouomega:
> > > @rubenzorrilla do you agree?
> > 
> > 
> > As I pointed above I'd like @RiccardoRossi to review this.
> 
> As I said in my last comment. @RiccardoRossi is the original author of this,.

Author is @adityaghantasala https://github.com/KratosMultiphysics/Kratos/commits/core/minor-improvement-ApplyPeriodicConditionProcess/kratos/processes/apply_periodic_boundary_condition_process.cpp

loumalouomega:
👋 @RiccardoRossi @adityaghantasala

loumalouomega:
> 👋 @RiccardoRossi @adityaghantasala

Hello

RiccardoRossi (commented):
at a quick glance i have no objection...but are you sure it still works?

loumalouomega:
> at a quick glance i have no objection...but are you sure it still works?

Test passes

PR #10339, opened 2022-10-04
[Core][FindGlobalNodalEntityNeighboursProcess] fix race condition

philbucher:
**📝 Description**
Maps cannot be filled in parallel, hence I added a custom reducer

@mpentek this fixes the parallel version of the surface smoothing


sunethwarna (approved):
Thanks a lot @philbucher :+1:  It seems either there was no test, or they were done in serial :/

philbucher:
I think there is a test but it is too small to detect this

PR #10342, opened 2022-10-05
[FastPR][Core] Generic condition and element registry

rubenzorrilla:
**📝 Description**
This PR  registers a generic geometry condition and element. These can be useful for the testing of KratosCore features that require conditions and elements with non-specified (i.e. generic) number of nodes implemented in applications.


loumalouomega:
Do you need approval?

rubenzorrilla:
> Do you need approval?

I think so. But I think it needs to be from one of the @KratosMultiphysics/technical-committee .

roigcarlo (approved):

PR #10344, opened 2022-10-06
add particle utilities

RiccardoRossi:
**📝 Description**
some useful utilities to classify and deal with particle distributions

- feature...

**🆕 Changelog**


kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+
+    /// Default constructor.
+    ParticlesUtilities(){};
+
+    /// Destructor.
+    virtual ~ParticlesUtilities(){};
+
+    template<unsigned int TDim>
+    static void CountParticlesInNodes(
+            BinBasedFastPointLocator<TDim>& rLocator,
```
roigcarlo: This could be const no?

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+            )
+        {
+
+            //reset the EMBEDDED_VELOCITY
+            block_for_each(rVolumeModelPart.Nodes(), [&rCounterVariable](auto& rNode){
+                rNode.GetValue(rCounterVariable) = 0.0;
+            });
+
+
+            unsigned int max_results = 10000;
```
roigcarlo: Why do we need so many results? I would leave it as a parameter with a default, but forcing to reserve 10K for every mesh seems overkill to me.

roigcarlo (commented):
Looks good to me. Just  two comments.

kratos/utilities/particles_utilities.h
RiccardoRossi: i will try to make it c9nst. i think ot 2as but it was giving a compilation prpblem, but i will verify.

kratos/utilities/particles_utilities.h
RiccardoRossi: i had pretty bad experiences with this...so i would leave it pretty large

roigcarlo: ok

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:		 BSD License
+//					 Kratos default license: kratos/license.txt
+//
+//  Main authors:    Riccardo Rossi
```
rubenzorrilla: Header formatting.

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+///@{
+
+///@}
+///@name Kratos Classes
+///@{
+
+/// Collection of utilities to compute statistic of particles
+/** Detail class definition.
+*/
+class ParticlesUtilities
```
rubenzorrilla, line 55: I'd call this `ParticleTrackingUtilities` of something of this sort to avoid developers being confused with any PFEM-related stuff.

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+    template<unsigned int TDim>
+    static void CountParticlesInNodes(
+            BinBasedFastPointLocator<TDim>& rLocator,
+            ModelPart& rVolumeModelPart,
+            const ModelPart& rParticlesModelPart,
+            const Variable<double>& rCounterVariable
+            )
+        {
+
+            //reset the EMBEDDED_VELOCITY
```
rubenzorrilla:
```suggestion
            //reset the counter variable
```

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+            block_for_each(rVolumeModelPart.Nodes(), [&rCounterVariable](auto& rNode){
+                rNode.GetValue(rCounterVariable) = 0.0;
```
rubenzorrilla: In here you rely that a `SetValue(rCounterVariable)` has been already performed. Maybe you can do the `SetValue` in here just in case. Not sure, just pointing it out.

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+
+            unsigned int max_results = 10000; 
+            typename BinBasedFastPointLocator<TDim>::ResultContainerType TLS(max_results);
+            
+            //for every interface node (nodes in cut elements)
+            block_for_each(rParticlesModelPart.Nodes(), TLS, [&rLocator, &rCounterVariable](const auto& rNode, auto& rTLS){
+
+                Vector shape_functions;
+                Element::Pointer p_element;
+                bool is_found = rLocator.FindPointOnMesh(rNode.Coordinates(), shape_functions, p_element, rTLS.begin(), rTLS.size(), 1e-5);
```
rubenzorrilla:
```suggestion
                const bool is_found = rLocator.FindPointOnMesh(rNode.Coordinates(), shape_functions, p_element, rTLS.begin(), rTLS.size(), 1e-5);
```
I'd also save the tolerance in an auxiliary variable like you did with `max_results` for a better code readability.

kratos/utilities/particles_utilities.h
rubenzorrilla:
```suggestion
            const unsigned int max_results = 10000; 
```

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+            //reset the EMBEDDED_VELOCITY
+            block_for_each(rVolumeModelPart.Nodes(), [&rCounterVariable](auto& rNode){
+                rNode.GetValue(rCounterVariable) = 0.0;
+            });
+
+
+            unsigned int max_results = 10000; 
+            typename BinBasedFastPointLocator<TDim>::ResultContainerType TLS(max_results);
+            
+            //for every interface node (nodes in cut elements)
```
rubenzorrilla:
```suggestion
            //for each node in the mesh, count the number of particles in the neighbouring elements
```

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+    ///@}
+    ///@name Life Cycle
+    ///@{
+
+    /// Default constructor.
+    ParticlesUtilities(){};
+
+    /// Destructor.
+    virtual ~ParticlesUtilities(){};
+
```
rubenzorrilla: Missing Doxygen documentation.

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+                    }
+                }
+
+            });
+
+        }
+
+    /**
+    this function looks if particle is found in the locator. If it is not, rVariable is marked with the value "outsider_value"
+    */
```
rubenzorrilla: Missing Doxygen documentation for the arguments.

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+    template<unsigned int TDim, class TScalarType>
+    static void MarkOutsiders(
+            BinBasedFastPointLocator<TDim>& rLocator,
+            ModelPart& rParticlesModelPart,
+            const Variable<TScalarType>& rVariable,
+            const TScalarType outsider_value
```
rubenzorrilla:
```suggestion
            const TScalarType OutsiderValue
```
[Style guide](https://github.com/KratosMultiphysics/Kratos/wiki/Style-Guide) suggestion.

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+            )
+        {
+            unsigned int max_results = 10000;
```
rubenzorrilla:
```suggestion
            const unsigned int max_results = 10000; 
```

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+        {
+            unsigned int max_results = 10000; 
+            typename BinBasedFastPointLocator<TDim>::ResultContainerType TLS(max_results);
+            
+            //for every interface node (nodes in cut elements)
+            block_for_each(rParticlesModelPart.Nodes(), TLS, [&rLocator, &rVariable, outsider_value](auto& rNode, auto& rTLS){
+
+                Vector shape_functions;
+                Element::Pointer p_element;
+                bool is_found = rLocator.FindPointOnMesh(rNode.Coordinates(), shape_functions, p_element, rTLS.begin(), rTLS.size(), 1e-5);
```
rubenzorrilla:
```suggestion
                const bool is_found = rLocator.FindPointOnMesh(rNode.Coordinates(), shape_functions, p_element, rTLS.begin(), rTLS.size(), 1e-5);
```

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+
+                if(!is_found){
+                    rNode.SetValue(rVariable, outsider_value);
```
rubenzorrilla: Shouldn't you use a flag for this purpose? Besides, this would make possible to get rid of the `TScalarType` template argument.

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+            BinBasedFastPointLocator<TDim>& rLocator,
+            ModelPart& rParticlesModelPart,
+            const Variable<TScalarType>& rVariable,
+            const TScalarType outsider_value
+            )
+        {
+            unsigned int max_results = 10000; 
+            typename BinBasedFastPointLocator<TDim>::ResultContainerType TLS(max_results);
+            
+            //for every interface node (nodes in cut elements)
```
rubenzorrilla:
```suggestion
            //for every particle search if it is inside or outside the volume mesh
```

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+                bool is_found = rLocator.FindPointOnMesh(rNode.Coordinates(), shape_functions, p_element, rTLS.begin(), rTLS.size(), 1e-5);
+
+                if(!is_found){
+                    rNode.SetValue(rVariable, outsider_value);
+                }
+            });
+        }
+
+    /**@brief provides the value of the interpolated variable "rInterpolationVariable" at all the positions indicated in rCoordinates
+    */
```
rubenzorrilla: Missing Doxygen documentation for function arguments.

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+    template<unsigned int TDim, class TScalarType>
+    static std::pair< std::vector<bool>, std::vector<TScalarType> > InterpolateValuesAtPoints(
+            BinBasedFastPointLocator<TDim>& rLocator,
+            const Matrix& rCoordinates,
+            const Variable<TScalarType>& rInterpolationVariable
+            )
+        {
+            unsigned int max_results = 10000;
```
rubenzorrilla:
```suggestion
            const unsigned int max_results = 10000; 
```

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+            typename BinBasedFastPointLocator<TDim>::ResultContainerType TLS(max_results);
+
+            auto interpolations = std::make_pair(std::vector<bool>(rCoordinates.size1()), std::vector<TScalarType>(rCoordinates.size1()));            
+            
+            //for every interface node (nodes in cut elements)
+            IndexPartition(rCoordinates.size1()).for_each(TLS, [&rLocator, &rCoordinates, &interpolations, &rInterpolationVariable](const auto& i, auto& rTLS){
+
+                Vector shape_functions;
+                Element::Pointer p_element;
+                bool is_found = rLocator.FindPointOnMesh(row(rCoordinates,i), shape_functions, p_element, rTLS.begin(), rTLS.size(), 1e-5);
```
rubenzorrilla:
```suggestion
                const bool is_found = rLocator.FindPointOnMesh(row(rCoordinates,i), shape_functions, p_element, rTLS.begin(), rTLS.size(), 1e-5);
```

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+
+                (interpolations.first)[i] = is_found;
+                if(is_found){
+
+                    auto& r_geom = p_element->GetGeometry();
+                    (interpolations.second)[i] = TScalarType();
```
rubenzorrilla:
```suggestion
                    (interpolations.second)[i] = rInterpolationVariable.Zero();
```
I'm not sure if all the data types default constructors return their corresponding zero.

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+            const Matrix& rCoordinates,
+            const Variable<TScalarType>& rInterpolationVariable
+            )
+        {
+            unsigned int max_results = 10000; 
+            typename BinBasedFastPointLocator<TDim>::ResultContainerType TLS(max_results);
+
+            auto interpolations = std::make_pair(std::vector<bool>(rCoordinates.size1()), std::vector<TScalarType>(rCoordinates.size1()));            
+            
+            //for every interface node (nodes in cut elements)
```
rubenzorrilla:
```suggestion
            //for each particle interpolate the given variable from the volume mesh
```

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+
+                if(!is_found){
+                    rNode.SetValue(rVariable, outsider_value);
+                }
+            });
+        }
+
+    /**@brief provides the value of the interpolated variable "rInterpolationVariable" at all the positions indicated in rCoordinates
+    */
+    template<unsigned int TDim, class TScalarType>
```
rubenzorrilla:
```suggestion
    template<unsigned int TDim, class TDataType>
```
Couldn't you extend this to non-scalar types? e.g. array  types to interpolate the velocity of the particles?

kratos/utilities/particles_utilities.h
RiccardoRossi: no, here it has to be a variable. The point is that i am assigning a "color" to the particles, and here i assign a special color to the ones that become "lost" to the simulation.

A more correct discussion however is that we should be able to set this either in the Database with history or in the one without history. I need to look how this was done e.g. in the variable utils

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+    static std::pair< std::vector<bool>, std::vector<TScalarType> > InterpolateValuesAtPoints(
+            BinBasedFastPointLocator<TDim>& rLocator,
+            const Matrix& rCoordinates,
```
rubenzorrilla: Wouldn't it be better to pass an `std::vector<array_1d<double,3>>`?  Then the user only needs to `push_back` (or `insert`) the result of calling `Coordinates()`.

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+
+            });
+
+        }
+
+    /**
+    this function looks if particle is found in the locator. If it is not, rVariable is marked with the value "outsider_value"
+    */
+    template<unsigned int TDim, class TScalarType>
+    static void MarkOutsiders(
```
rubenzorrilla:
```suggestion
    static void MarkOutsiderParticles(
```

kratos/utilities/particles_utilities.h
RiccardoRossi: this one cannot. I am using the is?found many times in the substepping

kratos/utilities/particles_utilities.h
```diff
@@ -0,0 +1,330 @@
...
+                if(!is_found){
+                    rNode.SetValue(rVariable, outsider_value);
+                }
+            });
+        }
+
+    /**@brief provides the value of the interpolated variable "rInterpolationVariable" at all the positions indicated in rCoordinates
+    */
+    template<unsigned int TDim, class TScalarType>
+    static std::pair< std::vector<bool>, std::vector<TScalarType> > InterpolateValuesAtPoints(
```
rubenzorrilla:
```suggestion
    static std::pair< std::vector<bool>, std::vector<TScalarType> > InterpolateValuesAtParticleCoordinates(
```

rubenzorrilla (changes requested):
The implementation of the three functions is correct (only a few comments). However, testing and cpp implementation is missing.

kratos/utilities/particles_utilities.h
RiccardoRossi: why not TScalarType()? it should be the same here ... and i suspect that it can be faster since i am not accessing million of times to the same variable in the memory to get the zero

kratos/utilities/particles_utilities.h
RiccardoRossi: you would have to export std::vector<array...> to python and to do an interface to accomplish this.

kratos/utilities/particles_utilities.h
RiccardoRossi: i may agree however that using std::vector,bool> can be dangerous... not sure what to use insteads however (do recall that we need python interface for it...for std::vector<bool> it is provided by pybind, but not for custom types)

RiccardoRossi (commented):
accepting some of the changes. other are pending

kratos/utilities/particles_utilities.h
RiccardoRossi: ?

kratos/utilities/particles_utilities.h
rubenzorrilla: My suggestion was on the naming of the method...

kratos/utilities/particles_utilities.h
rubenzorrilla: Doesn't pybind automatically cast it to a Python list?

kratos/utilities/particles_utilities.h
RiccardoRossi: could be done, but it is not as simple...for example as you would need to export new interfaces for vector<array_1d ... to python

kratos/utilities/particles_utilities.h
RiccardoRossi: actually it could even be constexpr

kratos/utilities/particles_utilities.h
rubenzorrilla: My though was in the direction of extending it to non scalar types. In these cases I'm not sure what the default constructor of the type does.

kratos/utilities/particles_utilities.h
rubenzorrilla: OK. I see your point. Resolving the conversation.

kratos/utilities/particles_utilities.h
rubenzorrilla: Understood. In this case we need it to be a variable. About the `VariablesUtils` comment, it depends on the method you need. Most of them work for both historical and non-historical database, so I'd try to stick to the non-historical one as much as possible.

RiccardoRossi (commented):
comments

kratos/utilities/particles_utilities.h
philbucher: It should work, we do that all the time

philbucher (changes requested):
Please separate into header/source and add tests

kratos/utilities/particles_utilities.h
miguelmaso: ParticleTrackingUtilities remids me of pfem2...

kratos/utilities/particles_utilities.h
RiccardoRossi: ok, i will try it out

RiccardoRossi:
> Please separate into header/source and add tests

methods here are templated, so header only is ok

ok about the tests

rubenzorrilla:
> > Please separate into header/source and add tests
> 
> methods here are templated, so header only is ok

Well, there is no problem on doing the explicit template instantiation in here.

> 
> ok about the tests

kratos/utilities/particles_utilities.h
RiccardoRossi: I also do not really love this renaming...honestly i would keep the name as it is

kratos/utilities/particles_utilities.h
RiccardoRossi: ok

kratos/utilities/particles_utilities.h
RiccardoRossi: ok

kratos/utilities/particles_utilities.h
RiccardoRossi: ok

kratos/utilities/particles_utilities.h
RiccardoRossi: ok, done

kratos/utilities/particles_utilities.h
RiccardoRossi: it does work (tested) however i would prefer the current version as i can do the following:

~~~py
coords = np.array(self.var_utils.GetInitialPositionsVector(self.all_nodes,3)).reshape(len(self.all_nodes),3)
~~~

kratos/utilities/particles_utilities.h
RiccardoRossi: ok

RiccardoRossi:
> > > Please separate into header/source and add tests
> > 
> > 
> > methods here are templated, so header only is ok
> 
> Well, there is no problem on doing the explicit template instantiation in here.
> 
> > ok about the tests

well... i DO NOT want to explicitly instantiate all the different types...

kratos/utilities/particles_utilities.h
rubenzorrilla: Fine for me then.

kratos/utilities/particles_utilities.h
rubenzorrilla: Very minor, there are mixing tabs and spaces. In principle our header has `BSD License` and `Kratos default...` aligned with the first R in `Riccardo Rossi`.

philbucher (dismissed):

philbucher (approved):

kratos/utilities/particles_utilities.h
RiccardoRossi: ok, corrected

kratos/utilities/particles_utilities.h
RiccardoRossi: done

rubenzorrilla (approved):
Approving.

PR #10346, opened 2022-10-06
[Core] Shifted boundary meshless interface utility

rubenzorrilla:
**📝 Description**
This PR adds a new utility to be used for the calculation of the extension opeartor in a new flavor of the SBM method. The changes can be summarized in
- The creation of two new variables `SHAPE_FUNCTIONS_VECTOR` and `SHAPE_FUNCTIONS_GRADIENT_MATRIX`
- The addition of a new utility to calculate the unfitted boundary shape functions and shape functions gradient *a la* meshless. The new utility supports MLS and RBF shape functions as well as a gradient-based extension operator, which stems from the idea of the original SBM method. Tests have been added accordingly.

The rationale of putting this in the `KratosCore` is that it is physics agnostic so it could be used later on to extend any of the Kratos Multiphysics applications capabilities.


kratos/utilities/shifted_boundary_meshless_interface_utility.h
```diff
@@ -0,0 +1,363 @@
+// KRATOS ___ ___  _  ___   __   ___ ___ ___ ___
+//       / __/ _ \| \| \ \ / /__|   \_ _| __| __|
+//      | (_| (_) | .` |\ V /___| |) | || _|| _|
+//       \___\___/|_|\_| \_/    |___/___|_| |_|  APPLICATION
+//
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
```
philbucher: wrong header

kratos/utilities/shifted_boundary_meshless_interface_utility.h
```diff
@@ -0,0 +1,363 @@
...
+
+#pragma once
+
+// System includes
+
+// External includes
+
+// Project includes
+#include "containers/model.h"
+#include "containers/pointer_vector.h"
```
philbucher:
```suggestion

```

kratos/utilities/shifted_boundary_meshless_interface_utility.h
```diff
@@ -0,0 +1,363 @@
...
+    /// @param ThisParameters Parameters object encapsulating the settings
+    ShiftedBoundaryMeshlessInterfaceUtility(
+        Model& rModel,
+        Parameters ThisParameters);
+
+    /// Copy constructor.
+    ShiftedBoundaryMeshlessInterfaceUtility(ShiftedBoundaryMeshlessInterfaceUtility const& rOther) = delete;
+
+    /// Destructor.
+    virtual ~ShiftedBoundaryMeshlessInterfaceUtility() = default;
```
philbucher:
```suggestion
```

unless you actually derive from it I would remove it
The compiler does this job for you

kratos/utilities/shifted_boundary_meshless_interface_utility.h
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    /// Print information about this object.
+    virtual void PrintInfo(std::ostream& rOStream) const
+    {
+        rOStream << "ShiftedBoundaryMeshlessInterfaceUtility";
+    }
+
+    /// Print object's data.
+    virtual void PrintData(std::ostream& rOStream) const
```
philbucher:
```suggestion
    std::string Info() const
    {
        return "ShiftedBoundaryMeshlessInterfaceUtility";
    }

    /// Print information about this object.
    void PrintInfo(std::ostream& rOStream) const
    {
        rOStream << "ShiftedBoundaryMeshlessInterfaceUtility";
    }

    /// Print object's data.
    void PrintData(std::ostream& rOStream) const

```

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
```diff
@@ -0,0 +1,1124 @@
+// KRATOS ___ ___  _  ___   __   ___ ___ ___ ___
```
philbucher: old header

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
```diff
@@ -0,0 +1,1124 @@
...
+            "sbm_interface_condition_name" : "",
+            "conforming_basis" : true,
+            "extension_operator_type" : "MLS",
+            "mls_extension_operator_order" : 1
+        })" );
+
+        return default_parameters;
+    }
+
+    void ShiftedBoundaryMeshlessInterfaceUtility::CalculateGradientBasedConformingExtensionBasis()
```
philbucher, line 212: longest function in the wild west

Seems that this could be shortened

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
```diff
@@ -0,0 +1,1124 @@
...
+    }
+
+    void ShiftedBoundaryMeshlessInterfaceUtility::CalculateMeshlessBasedNonConformingExtensionBasis()
+    {
+        // Set the required interface flags
+        SetInterfaceFlags();
+
+        // Set the modified shape functions factory
+        // Note that unique geometry in the mesh is assumed
+        const auto& r_begin_geom = mpModelPart->ElementsBegin()->GetGeometry();
```
philbucher, line 688: boom if the modelpart has no elements

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
```diff
@@ -0,0 +1,1124 @@
...
+                        if (p_neigh_elem->Is(ACTIVE)) {
+                            p_neigh_elem->Set(INTERFACE, true);
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    bool ShiftedBoundaryMeshlessInterfaceUtility::IsSplit(const GeometryType& rGeometry)
```
philbucher:
```suggestion
    bool ShiftedBoundaryMeshlessInterfaceUtility::IsSplit(const GeometryType& rGeometry) const

```

or better, make it a free function at the top of the file

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
```diff
@@ -0,0 +1,1124 @@
...
+            if (r_node.FastGetSolutionStepValue(DISTANCE) < 0.0) {
+                n_neg++;
+            } else {
+                n_pos++;
+            }
+        }
+        return (n_pos != 0 && n_neg != 0);
+    }
+
+    bool ShiftedBoundaryMeshlessInterfaceUtility::IsNegative(const GeometryType& rGeometry)
```
philbucher: idem

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
```diff
@@ -0,0 +1,1124 @@
...
+        std::size_t n_neg = 0;
+        for (const auto& r_node : rGeometry) {
+            if (r_node.FastGetSolutionStepValue(DISTANCE) < 0.0) {
+                n_neg++;
+            }
+        }
+        return (n_neg == rGeometry.PointsNumber());
+    }
+
+    void ShiftedBoundaryMeshlessInterfaceUtility::SetNodalDistancesVector(
```
philbucher: idem

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
```diff
@@ -0,0 +1,1124 @@
...
+        if (rNodalDistances.size() != n_nodes) {
+            rNodalDistances.resize(n_nodes);
+        }
+        std::size_t i = 0;
+        for (const auto& r_node : rGeometry) {
+            rNodalDistances[i++] = r_node.FastGetSolutionStepValue(DISTANCE);
+        }
+    }
+
+    ShiftedBoundaryMeshlessInterfaceUtility::ModifiedShapeFunctionsFactoryType ShiftedBoundaryMeshlessInterfaceUtility::GetStandardModifiedShapeFunctionsFactory(const GeometryType& rGeometry)
```
philbucher: idem

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
```diff
@@ -0,0 +1,1124 @@
...
+                    return Kratos::make_unique<Triangle2D3ModifiedShapeFunctions>(pGeometry, rNodalDistances);};
+            case GeometryData::KratosGeometryType::Kratos_Tetrahedra3D4:
+                return [](const GeometryType::Pointer pGeometry, const Vector& rNodalDistances)->ModifiedShapeFunctions::UniquePointer{
+                    return Kratos::make_unique<Tetrahedra3D4ModifiedShapeFunctions>(pGeometry, rNodalDistances);};
+            default:
+                KRATOS_ERROR << "Asking for a non-implemented modified shape functions geometry.";
+        }
+    }
+
+    ShiftedBoundaryMeshlessInterfaceUtility::MLSShapeFunctionsAndGradientsFunctionType ShiftedBoundaryMeshlessInterfaceUtility::GetMLSShapeFunctionsAndGradientsFunction()
```
philbucher:
```suggestion
    auto ShiftedBoundaryMeshlessInterfaceUtility::GetMLSShapeFunctionsAndGradientsFunction() const

```

philbucher (commented):
functions should be subdivided

also make the `const` of possible
you could also benefit from using `auto` for the return types

kratos/utilities/shifted_boundary_meshless_interface_utility.h
rubenzorrilla: Ouch. At first I implemented it in the `ConvectionDiffusionApplication` and this sneaked in. Thanks.

kratos/utilities/shifted_boundary_meshless_interface_utility.h
rubenzorrilla: Moved to the cpp.

kratos/utilities/shifted_boundary_meshless_interface_utility.h
rubenzorrilla: OK. Removing.

kratos/utilities/shifted_boundary_meshless_interface_utility.h
rubenzorrilla: Done.

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
rubenzorrilla: Done.

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
rubenzorrilla: I'll try... But at some point I thought it will be easier to keep it like this.

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
rubenzorrilla: Nice catch. Now an error is thrown (also in the other places this is done).

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
rubenzorrilla: Nice suggestion. I put them as free functions in an anonymous namespace. Indeed, my idea is to migrate these to an `DistanceFunctionUtilities` or something of this sort (they are repeated in many places...).

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
rubenzorrilla: Done.

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
rubenzorrilla: Done.

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
rubenzorrilla: Done.

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
rubenzorrilla: It seems that `auto` return type cannot be used in this case (some lambda-related stuff). OK about the `conts`.

kratos/tests/cpp_tests/utilities/test_shifted_boundary_meshless_interface_utility.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include "geometries/quadrilateral_2d_4.h"
+#include "processes/find_global_nodal_neighbours_process.h"
+#include "processes/generic_find_elements_neighbours_process.h"
+#include "processes/structured_mesh_generator_process.h"
+#include "utilities/shifted_boundary_meshless_interface_utility.h"
+
+namespace Kratos
+{
+namespace Testing
+{
```
philbucher:
```suggestion
namespace Kratos::Testing
{

```

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
```diff
@@ -0,0 +1,1164 @@
...
+                        KRATOS_ERROR << "Wrong MLS extension operator order. Only linear (1) and quadratic (2) are supported.";
+                }
+            default:
+                KRATOS_ERROR << "Wrong domain size.";
+        }
+    }
+
+    std::map<std::size_t, std::map<std::size_t, Vector>> ShiftedBoundaryMeshlessInterfaceUtility::SetSurrogateBoundaryNodalGradientWeights()
+    {
+        std::map<std::size_t, std::map<std::size_t, Vector>> sur_bd_nodes_map;
```
philbucher: maybe using unordered_map would be better?

philbucher (dismissed):

kratos/tests/cpp_tests/utilities/test_shifted_boundary_meshless_interface_utility.cpp
rubenzorrilla: Done.

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
rubenzorrilla: Agree. For the outer one it is possible. Not for the inner one.

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
rubenzorrilla: I went through it, the largest function is around 250 lines, which is admittedly larger than the standard rule of thumb of 200 lines. However, in this specific case, and taking into account that the implementation of the extension operators is a bit tricky, I think it would be better to leave it like this to ease following the steps one by one.

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
```diff
@@ -0,0 +1,1164 @@
...
+#include "utilities/parallel_utilities.h"
+#include "utilities/reduction_utilities.h"
+#include "utilities/shifted_boundary_meshless_interface_utility.h"
+
+namespace Kratos
+{
+
+namespace
+{
+    using GeometryType = ShiftedBoundaryMeshlessInterfaceUtility::GeometryType;
```
RiccardoRossi, line 35: could we pass the variable instead of using DISTANCE by default?

this goes in the direciton of using a different distance variable for example for free surface and for embedded, with the ultimate goal of being able to mix the two things

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
```diff
@@ -0,0 +1,1164 @@
...
+    }
+
+    /**
+     * @brief Check if current geometry is in negative side
+     * This method checks if current geometry has negative DISTANCE value in all the nodes
+     * @param rGeometry Geometry to be checked
+     * @return true If all nodes are negative
+     * @return false If there is one or more positive nodes
+     */
+    bool IsNegative(const GeometryType& rGeometry)
```
RiccardoRossi: same

~~~cpp
bool IsNegative(const GeometryType& rGeometry, const Variable<double>& rDistanceVar)
~~~

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
rubenzorrilla: Makes sense. Good observation.

rubenzorrilla: Done!

kratos/utilities/shifted_boundary_meshless_interface_utility.cpp
rubenzorrilla: Done!

RiccardoRossi (approved):
ok, this is fine to me.

philbucher (approved):

PR #10352, opened 2022-10-08
[HDF5] Model Predicates

matekelemen:
## Description
I'm *still* working on proper checkpointing, and I realized that Kratos could use a **flexible** system that checks the state of an analysis and triggers a process if some condition is met. Enter ```ModelPredicate```, which is a functor that takes a ```Model``` and returns a ```bool```.

## Details
Modularity comes from *pipes*: compile-time interchangable functors that take an immutable input, produce an output, and can be fed into each other (a vaguely similar idea to pipes on Unix/Linux). For example, checking whether ```TIME``` is within an interval would semantically look something like this:
```
Model => get ModelPart => get ProcessInfo => get TIME => IsInInterval
```
Pipes use static polymorphism so it's a bit of a template mayhem; a result of lacking C++20 concepts. However, it's straightforward and no workarounds are used.

I won't go into detail because, in my experience, PRs with long descriptions don't get reviewed; but ask and I shall answer.

## Notes
depends on templated ```Parameters``` getters #10052 and the updated ```IntervalUtility``` #10301.


matekelemen:
Closing for now. I'll open it up again as part of a chain of PRs to show how it's used, after the requirements are merged.

PR #10353, opened 2022-10-09
[DEMApplication] Energy dissipation and heat generation by rolling friction

rlrangel:
* Created methods to compute energy dissipation in the rolling friction models (missing Bounded model);
* Fixed the recently added viscous model of rolling friction;
* Removed two unnecessary energy operations in DEMApp: initialization of PARTICLE_INELASTIC_FRICTIONAL_ENERGY and call to CleanEnergies();
* Changed some function/variable names related to energy dissipation by rolling resistance;
* Made ThermalDEMApp compatible with recent changes on rolling friction models + General improvements in heat generation models;


ChengshunShang1996 (commented):
Well done! No problem from my side.

applications/DEMApplication/custom_strategies/strategies/explicit_solver_strategy.cpp
```diff
@@ -1794,8 +1790,8 @@ namespace Kratos {
         total_inelastic_frictional_energy = 0.0;
         double& total_inelastic_viscodamping_energy = r_process_info[PARTICLE_INELASTIC_VISCODAMPING_ENERGY];
```
maceligueta, line 1791: It's not to be fixed in this PR, but I consider using the ProcessInfo a bit ugly here. It is not meant to save physical quantities, but variables related to the process. We need to think a different place to put this information...

maceligueta (approved):

applications/DEMApplication/custom_strategies/strategies/explicit_solver_strategy.cpp
rlrangel: I agree.
However, the ProcessInfo is never used to save or retrieve the energy values, only in this function (ExplicitSolverStrategy::CleanEnergies) that it is setting the energies to zero.
This is why I removed the call to this function in the beginning of the time step, as it was useless.
I would go further and remove the function (I can do it on the next PR).

Issue #10148, opened 2022-08-17
Add Parameterize package to python 


mcgicjn2:
Hi all,

We would like to request the inclusion of parameterized to the python packages to enable the parameterization of our python tests. This allows us to produce a lot of acceptance tests very quickly to test large ranges of values.

Can this be considered.

Regards

Jon (mcgicjn2)

mcgicjn2:
![MicrosoftTeams-image](https://user-images.githubusercontent.com/9568573/185175814-44504ed9-7439-438d-b1e9-250e8b144199.png)

roigcarlo:
Hi @mcgicjn2 

I see no problem, but let me discuss first with the @KratosMultiphysics/technical-committee

philbucher:
Probably time to add a proper requirements.txt and setup.py

I think this package is a good addition (would be a testing dependency so that is does not get installed when distributing it)

roigcarlo:
> Probably time to add a proper requirements.txt and setup.py

Agree. Btw, we have a [setup.py](https://github.com/KratosMultiphysics/Kratos/blob/master/scripts/wheels/setup.py), but its only used through the wheelbuilders if we put some defaults we can move to the root.

mcgicjn2:
Cheers, we have skipped the test for now, and will apply it again after the update

roigcarlo:
@KratosMultiphysics/technical-committee In principle is favorable to install it. I will do it during this week.

carloslubbers:
Hi @roigcarlo did you have any time to work on this?

roigcarlo:
Hi, yes we agreed on adding it. Let me make the PR.

PR #10354, opened 2022-10-10
Adding parametrized package to CI

roigcarlo:
**📝 Description**
Implements #10148. It lets you decorate tests with parameter permutations.

**🆕 Changelog**
- Adding parametrized parameters

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,11 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
+    - name: Installing dependencies
+      shell: bash
+      run: |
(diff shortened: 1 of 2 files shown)


carloslubbers:
Is the package also useful in windows-core-without-unity and ubuntu-core-without-unity since they run tests too?

roigcarlo:
Not sure, tests of those are very minimal, the scope of the non-unity test is to check that we are not missing any inclusion by generating the source bundles or any of the other side effects that may cause. 

If we need to add tests / refactor existing ones that depend on it we can add it later. @KratosMultiphysics/technical-committee Opinion?

carloslubbers (approved):

PR #10358, opened 2022-10-11
Updating CD

roigcarlo:
**📝 Description**
Updates some dockerfiles with new sources and default versions.

**🆕 Changelog**
- Adds Python 310 enabled by default
- Updating to 9.2.1 for future release.


philbucher (approved):
the wheels files should also be excluded from the CI

roigcarlo:
I think they are, but I have touched the Cmake as well. I have to find a better solution to set the version...

PR #10367, opened 2022-10-13
Fixing HDF5 app compilation for windows

roigcarlo:
**📝 Description**
Mostly symbol exports, but also a odd bug in the mpi communicator.

Needs -DHDF5_ROOT to point at the compilation / installation dir.

**🆕 Changelog**
- Hdf5 application can now be compiled in win.


applications/HDF5Application/custom_io/hdf5_container_gauss_point_output.cpp
```diff
@@ -315,7 +315,7 @@ void SetDataBuffer(Variable<TDataType> const& rVariable,
     }
 
     NumberOfGaussPoints = rDataCommunicator.MaxAll(NumberOfGaussPoints);
-    flat_data_size = rDataCommunicator.MaxAll(flat_data_size);
+    flat_data_size = rDataCommunicator.MaxAll((int)flat_data_size);
```
roigcarlo, line 318: This is the odd fix. I am not sure why it is complaining, but it cannot deduce the type automatically from std::size_t

sunethwarna (approved):
Nice addition. Thanks @roigcarlo . Hmm... The CI failure is not related this ryt?

roigcarlo:
Nope, that's from another PR. @RiccardoRossi I think this is from the Particle utilities. The first comparison is always true if the type is std::size_t

RiccardoRossi (approved):
looks ok to me

RiccardoRossi:
the failure for the intel compiler is addressed in #10368

PR #10380, opened 2022-10-20
[HDF5] Journal

matekelemen:
Add ~~```RegistryFile```~~ ```Journal```, a class for keeping track of administrational data during an analysis. It's meant to keep track of output files generated during a simulation but is a bit more flexible to allow extending the output data as necessary.

For example, we no longer need to encode the step index or time stamp of the simulation step in the output file's name.


matekelemen:
### ToDo
- add base class with raw string interface (the current ```Journal``` will derive from it and keep it ```Parameters``` interface)
- make sure that the extractor gets called only from the thread that set it. This should avoid locking up when the extractor is a python object.
- provide some extensibility for formatting?

RiccardoRossi:
this is kind of cool..but whwt about backward compqtibility? is this
breaking all old code?

matekelemen:
Weeell, it's up to everyone using HDF5 to discuss whether they'll want existing processes to use ```Journal``` - and thus **kinda** (but not necessarily) break backward compatibility - or not. This PR just introduces the idea but doesn't apply it anywhere, so it won't break anything just yet. I don't want to push anyone to start using this; I'm just planning to rely on it later on myself.

Of course, even if we do eventually convert existing processes to use ```Journal```, data could still be encoded into file names if someone so desires.

matekelemen:
Ready for review.

PR #7296, opened 2020-07-28
[Core] delete Variable assignment operator

philbucher:
**Description**
This PR delete the assignment operator of `Variable`, which should help mitigate the memory problems we have with Variables =>  #3789 

Only minor updates should be necessary because we did already a large effort in cleaning (I found one in the core which was also discovered by @loumalouomega)
@loumalouomega you didn't make a separate commit for the fix, you mind pushing it here? I don't wanna meddle with https://github.com/KratosMultiphysics/Kratos/pull/7290/commits/fabb23002537fed5d40800f721ea50280741d9a3

PR #10381, opened 2022-10-20
[Core] Make private copy constructor of Variable

loumalouomega:
**📝 Description**

Making private* copy constructor of Variable. For memory safety.

FYI @ddiezrod, related #7296

*So it can be used by a friend class. Required for new resgitry FYI @pooyan-dadvand 

**🆕 Changelog**

- [Remove copy constructor of Variable](https://github.com/KratosMultiphysics/Kratos/commit/b9e047bcf7112a796ea3cbcceb25958faa1be18d)
- [Better private so it be declared as friend class for new registry](https://github.com/KratosMultiphysics/Kratos/pull/10381/commits/ba70f311fa1aaca6009bc05fa7f6e64654626f58)
- [Passing by reference in VariableUtils](https://github.com/KratosMultiphysics/Kratos/commit/4503ccd5322ee1e4f5cf0979dfd435285c56d62a)
- [Pass by reference in debug_utilities](https://github.com/KratosMultiphysics/Kratos/commit/802c1b925324e0cb1a221415064b6b3c67263016)
- [Fix MeshingApp](https://github.com/KratosMultiphysics/Kratos/commit/cdd9864d2cee53ec3d05b0b00a57f0860a3bd6d0)
- [Fix fluid app](https://github.com/KratosMultiphysics/Kratos/commit/de0d283f0383ee4c898b1577e9629b8f3a7da7bc)
- [Fix Structural app](https://github.com/KratosMultiphysics/Kratos/commit/dff5d8a35ca5327eb324685e2278afc79fbe4a77)
- [Using reference in StatisticsApplication](https://github.com/KratosMultiphysics/Kratos/pull/10381/commits/3c0ac69cad95cb7a82a2c7766f0a7e5572449b66)

### DIFF
--- a/applications/FluidDynamicsApplication/custom_utilities/statistics_utilities.h
+++ b/applications/FluidDynamicsApplication/custom_utilities/statistics_utilities.h
@@ -648,7 +648,7 @@ namespace Internals {
 class MakeSamplerAtLocalCoordinate {
 public:
     static std::function<double(const Geometry< Node<3> >& rGeometry, const Vector& rShapeFunctions, const Matrix& rShapeDerivatives)> ValueGetter(const Variable<double>& rVariable) {
-        return [rVariable](const Geometry< Node<3> >& rGeometry, const Vector& rShapeFunctions, const Matrix& rShapeDerivatives) -> double {
+        return [&rVariable](const Geometry< Node<3> >& rGeometry, const Vector& rShapeFunctions, const Matrix& rShapeDerivatives) -> double {
             KRATOS_DEBUG_ERROR_IF(rGeometry.size() != rShapeFunctions.size()) << "Number of nodes in provided geometry does not match number of shape functions" << std::endl;
             double value = 0.0;
             for (unsigned int i =  0; i < rGeometry.size(); i++) {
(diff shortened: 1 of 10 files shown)


matekelemen:
Weeeell, if we were to follow along these lines, all constructors should be made private to avoid creating dangling references.

loumalouomega:
> Weeeell, if we were to follow along these lines, all constructors of `Variable` should be made private to avoid creating dangling references.

And tried, but then macros don't work

matekelemen:
Yep, you'd have to define some ```VariableFactory``` friend class that has access to the private constructors of ```Variable```. Technically, it would be the only class authorized to instantiate ```Variable```, so the macros would need to be changed accordingly.

I'm not too serious about it though, and don't want to force you to do it.

loumalouomega:
> Yep, you'd have to define some `VariableFactory` friend class that has access to the private constructors of `Variable`. Technically, it would be the only class authorized to instantiate `Variable`, so the macros would need to be changed accordingly.
> 
> I'm not too serious about it though, and don't want to force you to do it.

@pooyan-dadvand maybe we can solve the problem we mentioned instead of delete with private+friend

loumalouomega:
Okay, looks like this is ready @KratosMultiphysics/technical-committee

RiccardoRossi (commented):
i think this is a wise change, and i would be favorable to have it in.

@KratosMultiphysics/technical-committee do you have any resistance to this?

rubenzorrilla (commented):
I also agree with the change. Besides, the other changes fix some potential memory issues so IMO this is ready to be merged.

Just mentioning that the variable factory @matekelemen is pointing will (soon) be the `Registry`. Hence, making the registry friend of the `Variable` should be the unique requirement as observed by @matekelemen .

rubenzorrilla (approved):
I think this should go in.

PR #10170, opened 2022-08-19
[CI] use ubuntu 22

philbucher:
see title

PR #10390, opened 2022-10-21
[AMGCL] add missing initialization

philbucher:
Minor fix, porting from https://github.com/ddemidov/amgcl/pull/241

### DIFF
--- a/external_libraries/amgcl/adapter/block_matrix.hpp
+++ b/external_libraries/amgcl/adapter/block_matrix.hpp
@@ -73,7 +73,7 @@ struct block_matrix_adapter {
         typedef ptrdiff_t col_type;
         typedef BlockType val_type;
 
-        std::array<char, sizeof(Base) * BlockSize> buf;
+        std::array<char, sizeof(Base) * BlockSize> buf {};
         Base * base;
 
         bool done;
--- a/external_libraries/amgcl/adapter/block_matrix.hpp
+++ b/external_libraries/amgcl/adapter/block_matrix.hpp
@@ -73,16 +73,16 @@ struct block_matrix_adapter {
         typedef ptrdiff_t col_type;
         typedef BlockType val_type;
 
-        std::array<char, sizeof(Base) * BlockSize> buf {};
+        std::array<char, sizeof(Base) * BlockSize> buf;
         Base * base;
 
         bool done;
         col_type cur_col;
         val_type cur_val;
 
-        row_iterator(const Matrix &A, col_type row)
-            : base(reinterpret_cast<Base*>(buf.data())), done(true)
+        row_iterator(const Matrix &A, col_type row) : done(true)
         {
+            base = reinterpret_cast<Base*>(buf.data());
             for(int i = 0; i < BlockSize; ++i) {
                 new (base + i) Base(backend::row_begin(A, row * BlockSize + i));
 


external_libraries/amgcl/adapter/block_matrix.hpp
```diff
@@ -73,7 +73,7 @@ struct block_matrix_adapter {
         typedef ptrdiff_t col_type;
         typedef BlockType val_type;
 
-        std::array<char, sizeof(Base) * BlockSize> buf;
+        std::array<char, sizeof(Base) * BlockSize> buf {};
```
RiccardoRossi: iisn t this a memory error?

RiccardoRossi: i mean, you are initializing a block to nothing...

philbucher: if I understand correctly, then before it was not initialized at all (so basically what you want to achieve with the `noinit` with std::vector)
This adds default initialization

RiccardoRossi: but...do you need to initialize it? mentioning also @ddemidov

ddemidov: - The `= {}` does nothing to the array except filling it with zeros, so it is only used here to silence the compiler.
- You don't have to initialize it here, as the contents of the array is indirectly initialized at line 87 below.

philbucher: yes I need to initialize it, otherwise the compiler complains. It is used already in line 84, but the initialization happens only from line 87

@ddemidov maybe you know a better solution?

IMO the compiler is correct in this case

ddemidov: In line 84 the pointer to the contents of the buffer is assigned to the `base` pointer. The contents is then filled out in line 87. The memory is allocated "statically" and then filled out using in-place constructor. 

To clarify, I do not object against the change in this PR. The "you don't need it" bit was to say that the program works correctly without it, but I agree silencing the compiler is a good enough reason.

RiccardoRossi: well...generally you do not want to i initialize it with zeros as it takes time...not sure in this specific case

RiccardoRossi: well...but silencing the co0iler doesn t happen a t the orice of a Ero initialization? if so...i would keep the warning

RiccardoRossi: possibly silence it locally by pragma

ddemidov: Ok, this does the same thing as the original code, but the compiler does not complain:

```diff
diff --git a/amgcl/adapter/block_matrix.hpp b/amgcl/adapter/block_matrix.hpp
index ea7eedd..febb4be 100644
--- a/amgcl/adapter/block_matrix.hpp
+++ b/amgcl/adapter/block_matrix.hpp
@@ -80,9 +80,9 @@ struct block_matrix_adapter {
         col_type cur_col;
         val_type cur_val;

-        row_iterator(const Matrix &A, col_type row)
-            : base(reinterpret_cast<Base*>(buf.data())), done(true)
+        row_iterator(const Matrix &A, col_type row) : done(true)
         {
+            base = reinterpret_cast<Base*>(buf.data());
             for(int i = 0; i < BlockSize; ++i) {
                 new (base + i) Base(backend::row_begin(A, row * BlockSize + i));
```

RiccardoRossi: ok, admittedly i am a bit lost ... and it looks like that both @ddemidov and @philbucher understand what you are doing so ... i'd better silence myself. Thx to both of you, and sorry for adding unnecessary noise

ddemidov: @RiccardoRossi , I think you helped to come up with a better solution, which is my comment above. @philbucher , if you do not object, I would replace your last change with it?

philbucher: Sure!
I am already testing it: https://github.com/KratosMultiphysics/Kratos/pull/10170/commits/3860691e93ca1b54d0f601c0beae4df433e7999e

philbucher: works, thx @ddemidov

ddemidov (approved):

PR #10392, opened 2022-10-23
[Core][ParallelUtils] Adding KRATOS_CRITICAL_SECTION

philbucher:
This macro is to be used instead of `omp critical`. It is a shorter version of the `std::lock_guard` and more user friendly. @KratosMultiphysics/technical-committee [we discussed this a long time ago](https://github.com/orgs/KratosMultiphysics/teams/technical-committee/discussions/5/comments/7).

Usage:
~~~c++
block_for_each {
    ...
    // old version with omp critical
    #pragma omp critical
    {
        // some critical section
        sum += 1;
    }
     
    // new version with lock_guard and GlobalLock, this is already used in some places
    {
        // another critical section
        const std::lock_guard scope_lock(ParallelUtilities::GetGlobalLock());
        sum += 1;
    }

    // improved solution with Macro. 
    // (This way it would also be possible to change it in the background if necessary)
    KRATOS_CRITICAL_SECTION(
        sum += 1;
    )
}
~~~


RiccardoRossi:
mmm

no {}?

i think mantaining them wpuld help in keeping under control the length of the critical zone (which does not need to coincide with the length of the loop body)

philbucher:
no, no `{}` needed, the syntax is slightly different from `omp critical`

see the [docs](https://en.cppreference.com/w/cpp/thread/lock_guard)

an alternative impl could be `KRATOS_CRITICAL_BEGIN` combined with `KRATOS_CRITICAL_END`, then it could be made consistent with `omp critical`

philbucher:
@RiccardoRossi I improved the solution, now it is very similar to `omp critical`
I edited the first post to update the example

RiccardoRossi:
now i like it!

kratos/utilities/parallel_utilities.h
```diff
@@ -36,18 +35,17 @@
 #include "includes/global_variables.h"
 #include "includes/lock_object.h"
 
+#define KRATOS_CRITICAL_SECTION(critical_section) {const std::lock_guard scope_lock(ParallelUtilities::GetGlobalLock()); critical_section }
```
matekelemen: This will produce an error on comma separators because there's only one argument. We'd need ```...``` and ```__VA_ARGS__``` to deal with those cases. [Example on godbolt](https://godbolt.org/#g:!((g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:2,endLineNumber:21,positionColumn:2,positionLineNumber:21,selectionStartColumn:2,selectionStartLineNumber:21,startColumn:2,startLineNumber:21),source:'%23include+%3Cvector%3E%0A%23include+%3Ciostream%3E%0A%23include+%3Cnumeric%3E%0A%0A//+This+will+result+in+an+error+if+there+are+separators+(commas)+in+critical_section.%0A%23define+KRATOS_CRITICAL_SECTION(critical_section)+%7B+critical_section+%7D%0A%0A//+This+one+allows+comma+separators.%0A%23define+KRATOS_CRITICAL_SECTION_2(...)+%7B__VA_ARGS__%7D%0A%0Aint+main()%0A%7B%0A++++std::vector%3Cint%3E+vec(10)%3B%0A++++KRATOS_CRITICAL_SECTION(%0A++++++++int+i+%3D+1,+j+%3D+2%3B+//+%3C%3D%3D+error+if+KRATOS_CRITICAL_SECTION%0A++++++++std::iota(vec.begin(),+vec.end(),+i+%2B+j)%3B+//+%3C%3D%3D+this+is+fine+(commas+enclosed+between+parentheses)%0A++++)%0A%0A++++for+(int+item+:+vec)%0A++++++++std::cout+%3C%3C+item+%3C%3C+std::endl%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),header:(),k:51.34680134680136,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:executor,i:(argsPanelShown:'1',compilationPanelShown:'0',compiler:gsnapshot,compilerOutShown:'0',execArgs:'',execStdin:'',fontScale:14,fontUsePx:'0',j:3,lang:c%2B%2B,libs:!(),options:'-std%3Dc%2B%2B20',source:1,stdinPanelShown:'1',tree:'1',wrap:'1'),l:'5',n:'0',o:'Executor+x86-64+gcc+(trunk)+(C%2B%2B,+Editor+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:51.34680134680136,l:'3',n:'0',o:'',t:'0'),(g:!((g:!((h:executor,i:(argsPanelShown:'1',compilationPanelShown:'0',compiler:clang1400,compilerOutShown:'0',execArgs:'',execStdin:'',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-std%3Dc%2B%2B20',source:1,stdinPanelShown:'1',tree:'1',wrap:'1'),l:'5',n:'0',o:'Executor+x86-64+clang+14.0.0+(C%2B%2B,+Editor+%231)',t:'0')),header:(),l:'4',m:37.833378084126586,n:'0',o:'',s:0,t:'0'),(g:!((h:executor,i:(argsPanelShown:'1',compilationPanelShown:'0',compiler:vcpp_v19_latest_x86,compilerOutShown:'0',execArgs:'',execStdin:'',fontScale:14,fontUsePx:'0',j:2,lang:c%2B%2B,libs:!(),options:'/std:c%2B%2B20',source:1,stdinPanelShown:'1',tree:'1',wrap:'1'),l:'5',n:'0',o:'Executor+x86+msvc+v19.latest+(C%2B%2B,+Editor+%231)',t:'0')),header:(),l:'4',m:62.16662191587342,n:'0',o:'',s:0,t:'0')),k:48.65319865319866,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

```suggestion
#define KRATOS_CRITICAL_SECTION(...) {const std::lock_guard scope_lock(ParallelUtilities::GetGlobalLock()); __VA_ARGS__}
```

matekelemen (changes requested):

matekelemen (dismissed):

RiccardoRossi:
the onlny problem is () vs {}

bud admittedly i have no solution for it

kratos/utilities/parallel_utilities.h
philbucher: Good catch!

philbucher:
> the onlny problem is () vs {}
> 
> 
> 
> bud admittedly i have no solution for it

I guess this is unfortunate but acceptable 

If it is the inly remaining issue and you like it, @RiccardoRossi feel free to approve ;)

RiccardoRossi:
@KratosMultiphysics/technical-committee i would approve it AS IS. Is it ok to you?

kratos/utilities/parallel_utilities.h
philbucher: now also covered by the test

philbucher:

pooyan-dadvand:
@KratosMultiphysics/technical-committee discussed this one and we think that having the macro inside the braces just to define the lock can be more clear. That is we would prefer to write:

```c++
block_for_each {
    ...

    // new version with lock_guard and GlobalLock, this is already used in some places
    {
        // another critical section
        KRATOS_CRITICAL_SECTION
        sum += 1;
    }

}

void f()
{
    KRATOS_CRITICAL_SECTION
}

```

were KRATOS_CRITICAL_SECTION is defined as:

```c++
#define KRATOS_CRITICAL_SECTION const std::lock_guard scope_lock(ParallelUtilities::GetGlobalLock());
```

pooyan-dadvand:
The reason that @KratosMultiphysics/technical-committee didn't vote for the proposed one is the parenthesis which is not very clear and a small other reason is making a function critical.

philbucher:
done @KratosMultiphysics/technical-committee, thanks for the review

Should be ready now :)

rubenzorrilla (approved):

PR #10395, opened 2022-10-23
[Core][ParUtils] make atomic-utils compatible with C++ based parallelization

philbucher:
This PR makes the `atomic_utilities` compatible with C++ based parallelization.

It uses `std::atomic_ref` (part of C++20) / `boost::atomic_ref` (available from boost v1.74) depending on what is available


CMakeLists.txt
```diff
@@ -385,6 +385,14 @@ else()
   add_definitions( -DBOOST_UBLAS_MOVE_SEMANTICS )
 endif ()
 
+# std::atomic_ref is part of C++20, hence using boost::atomic_ref
+# boost::atomic_ref has full support from version 1.74
```
RiccardoRossi, line 389: overall ok to me...but i am wondering if we should not make it forward compatible to
c++20, by defining a kratos atomic_ref.

what do you think?

philbucher: Look at the code, this is basically what I did :)

RiccardoRossi: well, i may misunderstand the code, but i  would say that you always require boost 1.74, even if c++20 is being used...

kratos/utilities/atomic_utilities.h
```diff
@@ -152,8 +178,15 @@ inline void AtomicSubMatrix(TMatrixType1& target, const TMatrixType2& value)
 template<class TDataType>
 inline void AtomicMult(TDataType& target, const TDataType& value)
 {
+#ifdef KRATOS_SMP_OPENMP
     #pragma omp atomic
     target *= value;
+#elif defined(KRATOS_SMP_CXX11)
+    AtomicRef<TDataType> at_ref{target};
```
RiccardoRossi, line 185: just for the sake of curiosity ... but why here not using *=?

CMakeLists.txt
philbucher: ah got it, good catch

kratos/utilities/atomic_utilities.h
philbucher: std::atomic ref doesn't have `operator*`

RiccardoRossi (dismissed):

philbucher:
ping @RiccardoRossi , this needs a reapproval :)

RiccardoRossi (approved):

PR #10397, opened 2022-10-24
[CMake][TrilinosApplication] Improve Trilinos find

loumalouomega:
**📝 Description**

Using `find_package`* will improve and make more standard the *Trilinos* find. With this it will be easier to use *Trilinos*, for example, I did the test using [*Spack*](https://github.com/spack/spack) with great results in an ARM64 machine.

* We should use `find_package` always when possible

**🆕 Changelog**

- [Improve Trilinos find](https://github.com/KratosMultiphysics/Kratos/commit/218b5f538f9d1bbbe50d05197155660bf9d8b5b2)
- [Improvement in Trilinos libraries message](https://github.com/KratosMultiphysics/Kratos/commit/9a213403a2dbecaf8e486a730e39670bd996425b)
- [Improve documentation](https://github.com/KratosMultiphysics/Kratos/pull/10397/commits/07bc57ede6ee3103679c04815ce410407313d813)
- [Adding Spack documentation](https://github.com/KratosMultiphysics/Kratos/pull/10397/commits/d91aa4d14da50de136a52d54ce2b69cc737a6348)


roigcarlo (approved):
Thx!

applications/TrilinosApplication/CMakeLists.txt
```diff
@@ -96,8 +96,8 @@ kratos_python_install(${INSTALL_PYTHON_USING_LINKS} ${CMAKE_CURRENT_SOURCE_DIR}/
 get_filename_component (CURRENT_DIR_NAME ${CMAKE_CURRENT_SOURCE_DIR} NAME)
 kratos_python_install_directory(${INSTALL_PYTHON_USING_LINKS} ${CMAKE_CURRENT_SOURCE_DIR}/python_scripts KratosMultiphysics/${CURRENT_DIR_NAME} )
 
-# TODO: Move this to the new system. Not sure how it will work on windows but being trillinos should give no problem.
-message("Trilinos libs to be installed" ${TRILINOS_LIBRARIES})
+# TODO: Move this to the new system. Not sure how it will work on windows but being Trilinos should give no problem.
+message(STATUS "Trilinos libs to be installed: " ${TRILINOS_LIBRARIES})
```
rubenzorrilla, line 100: Could it be that this change turned into a super verbose output?

roigcarlo: Shouldn't be, in any case STATUS is in the same level as no mode output :S

loumalouomega: It is basically the same with a minor decorator

rubenzorrilla: The point is that now the `Trilinos libs to be installed: ` is super verbose in my machine (maybe I didn't note it before).

loumalouomega: I just added STATUS, is not that different. You can check to remove STATUS locally. If that's the reason I will revert.

PR #10400, opened 2022-10-24
[CMake] Remove old CMake module. Pastix is not longer compatible

loumalouomega:
**📝 Description**

We need to rewrite the wrapper from scratch

**🆕 Changelog**

- [Remove old CMake module. Pastix is no longer compatible](https://github.com/KratosMultiphysics/Kratos/commit/da5eaf278eb7bc49796212803bb8e4c64cb969a4)


RiccardoRossi:
but...doesn't this work with old versions of pastix?

loumalouomega:
> but...doesn't this work with old versions of pastix?

Nope, we removed the solvers with the old LinearSolversApplication: https://github.com/KratosMultiphysics/Kratos/search?q=Pastix

We should create a new wrapper...or wait until they add the interface to Amesos (Trilinos), which it is being done right now according to Sandia Lab: https://trilinos.github.io/pdfs/PARA06-amesos.pdf

loumalouomega:
> @loumalouomega Is right, we tried to compile with pastix some time ago and its broken. For me this is 👍

@RiccardoRossi do you agree to clean up?, Maybe we can give support to pastix with Trilinos: https://trilinos.github.io/pdfs/4-UsingTrilinosLinearSolvers.pdf, at the end the version we supported was fully serial, and the newer versions work in MPI

(license is a different topic...)

philbucher:
afaik Eigen also has an interface to Pastix

BTW the original pastix interface was part of the ExternalSolversApp, which was removed long time ago

philbucher (approved):
Not sure if anyone would still need this now that we have Pardiso

PR #10401, opened 2022-10-26
[StructuralMechanicsApp] Minor inconsistencies in rotations in Init and Fin solution step methods

AlejandroCornejo:
**📝 Description**
Together with @lagoncalvesjr we noticed that, for cases in which we rotate the local axis of the elements a few rotations when missing when doing the `InitializeSolutionStep`, `FinalizeSolutionStep` and `CalculateOnConstitutiveLaw` methods.

I've solved this issues together with removing redundant checks in the `InitializeSolutionStep` and in the `FinalizeSolutionStep`.

### DIFF
--- a/applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
+++ b/applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
@@ -827,6 +827,8 @@ void BaseSolidElement::CalculateOnIntegrationPoints(
                 for (IndexType point_number = 0; point_number < number_of_integration_points; point_number++) {
                     // Compute element kinematics B, F, DN_DX ...
                     CalculateKinematicVariables(this_kinematic_variables, point_number, this->GetIntegrationMethod());
+                    // Compute constitutive law variables
+                    SetConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points);
 
                     // Compute material reponse
                     CalculateConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points, GetStressMeasure());
@@ -888,7 +890,8 @@ void BaseSolidElement::CalculateOnIntegrationPoints(
             for (IndexType point_number = 0; point_number < number_of_integration_points; ++point_number) {
                 // Compute element kinematics B, F, DN_DX ...
                 CalculateKinematicVariables(this_kinematic_variables, point_number, this->GetIntegrationMethod());
-
+                // Compute constitutive law variables
+                SetConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points);
                 // Compute material reponse, not encessary to rotate since it's an invariant
                 CalculateConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points, GetStressMeasure());
 
@@ -1037,6 +1040,8 @@ void BaseSolidElement::CalculateOnIntegrationPoints(
             for ( IndexType point_number = 0; point_number < number_of_integration_points; ++point_number ) {
                 // Compute element kinematics B, F, DN_DX ...
                 CalculateKinematicVariables(this_kinematic_variables, point_number, this->GetIntegrationMethod());
+                // Compute constitutive law variables
+                SetConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points);
 
                 if (is_rotated)
                     RotateToLocalAxes(Values, this_kinematic_variables);
@@ -1089,7 +1094,8 @@ void BaseSolidElement::CalculateOnIntegrationPoints(
             for ( IndexType point_number = 0; point_number < number_of_integration_points; ++point_number ) {
                 // Compute element kinematics B, F, DN_DX ...
                 CalculateKinematicVariables(this_kinematic_variables, point_number, this->GetIntegrationMethod());
-
+                // Compute constitutive law variables
+                SetConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points);
                 // Compute material reponse
                 CalculateConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points, this_stress_measure);
 
@@ -1216,7 +1222,8 @@ void BaseSolidElement::CalculateOnIntegrationPoints(
             for ( IndexType point_number = 0; point_number < mConstitutiveLawVector.size(); ++point_number ) {
                 // Compute element kinematics B, F, DN_DX ...
                 CalculateKinematicVariables(this_kinematic_variables, point_number, this->GetIntegrationMethod());
-
+                // Compute constitutive law variables
+                SetConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points);
                 if (is_rotated)
                     RotateToLocalAxes(Values, this_kinematic_variables);
 
@@ -1552,9 +1559,6 @@ void BaseSolidElement::CalculateConstitutiveVariables(
     const ConstitutiveLaw::StressMeasure ThisStressMeasure
     )
 {
-    // Setting the variables for the CL
-    SetConstitutiveVariables(rThisKinematicVariables, rThisConstitutiveVariables, rValues, PointNumber, IntegrationPoints);
-
     // Actually do the computations in the ConstitutiveLaw in local axes
     mConstitutiveLawVector[PointNumber]->CalculateMaterialResponse(rValues, ThisStressMeasure); //here the calculations are actually done
 }
--- a/applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
+++ b/applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
@@ -128,6 +128,9 @@ void BaseSolidElement::InitializeSolutionStep( const ProcessInfo& rCurrentProces
                 // Compute constitutive law variables
                 SetConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points);
 
+                // rotate to local axes strain/F
+                RotateToLocalAxes(rValues, this_kinematic_variables);
+
                 // Call the constitutive law to update material variables
                 mConstitutiveLawVector[point_number]->InitializeMaterialResponse(Values, GetStressMeasure());
 
--- a/applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
+++ b/applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
@@ -725,7 +725,6 @@ void BaseSolidElement::CalculateOnIntegrationPoints(
     )
 {
     const GeometryType::IntegrationPointsArrayType &integration_points = this->IntegrationPoints(this->GetIntegrationMethod());
-    const bool is_rotated = IsElementRotated();
 
     const std::size_t number_of_integration_points = integration_points.size();
     const auto& r_geometry = GetGeometry();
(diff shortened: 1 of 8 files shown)


applications/StructuralMechanicsApplication/custom_elements/base_solid_element.h
```diff
@@ -1053,6 +1053,9 @@ class KRATOS_API(STRUCTURAL_MECHANICS_APPLICATION) BaseSolidElement
             // Compute material reponse
             this->SetConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points);
 
+            // rotate to local axes strain/F
```
loumalouomega, line 1058:
```suggestion
            // Rotate to local axes strain/F
```

Should we check first that the element is rotated to avoid overhead?

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
```diff
@@ -209,19 +196,20 @@ void BaseSolidElement::FinalizeSolutionStep( const ProcessInfo& rCurrentProcessI
...
+            CalculateKinematicVariables(this_kinematic_variables, point_number, mThisIntegrationMethod);
 
-                // Compute constitutive law variables
-                SetConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points);
+            // Compute constitutive law variables
+            SetConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points);
 
-                // Call the constitutive law to update material variables
-                mConstitutiveLawVector[point_number]->FinalizeMaterialResponse(Values, GetStressMeasure());
+            // rotate to local axes strain/F
```
loumalouomega:
```suggestion
            // Rotate to local axes strain/F
```

Iden, no check?

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
```diff
@@ -121,19 +114,20 @@ void BaseSolidElement::InitializeSolutionStep( const ProcessInfo& rCurrentProces
...
+            CalculateKinematicVariables(this_kinematic_variables, point_number, mThisIntegrationMethod);
 
-                // Compute constitutive law variables
-                SetConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points);
+            // Compute constitutive law variables
+            SetConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points);
 
-                // Call the constitutive law to update material variables
-                mConstitutiveLawVector[point_number]->InitializeMaterialResponse(Values, GetStressMeasure());
+            // rotate to local axes strain/F
```
loumalouomega:
```suggestion
            // Rotate to local axes strain/F
```

Idem ch

loumalouomega (changes requested):
Comments

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.h
AlejandroCornejo: At that time we did it like this to try to encapsulate... I can try to separate the check

AlejandroCornejo: Ideally this check should be done once per simulation but IDK when to do it... In the Initialize we may have the problem that the element has not been assigned the rotation.

loumalouomega: This is important, otherwise could have a significative impact

AlejandroCornejo:
okkey @loumalouomega , this took longer than expected. 

In order to do the check only once per element (before was once per IP) I had to de-encapsulate the rotations to be exposed in the upper method. It's less clean but more efficient. 

Next I had to to minor modifications to the solid elements (we lost some encapsulation).
The tests are passing in my computer. 

Additionally i found some inconsistencies when printing stresses (rotations were missing).

loumalouomega (commented):
Now I agree, I will unblock if @RiccardoRossi agrees

AlejandroCornejo:
> Now I agree, I will unblock if @RiccardoRossi agrees

ok let's wait for @RiccardoRossi

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
```diff
@@ -82,15 +82,9 @@ void BaseSolidElement::Initialize(const ProcessInfo& rCurrentProcessInfo)
...
 {
-    // We initialize the material reponse if required
-    bool required = false;
-    for ( IndexType point_number = 0; point_number < mConstitutiveLawVector.size(); ++point_number ) {
-        if (mConstitutiveLawVector[point_number]->RequiresInitializeMaterialResponse()) {
-            required = true;
-            break;
-        }
-    }
+    const bool required = mConstitutiveLawVector[0]->RequiresInitializeMaterialResponse();
```
RiccardoRossi: you can have different CLs within the element...i think that the previosu code was better here

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
```diff
@@ -121,19 +115,21 @@ void BaseSolidElement::InitializeSolutionStep( const ProcessInfo& rCurrentProces
...
 
-                // Compute constitutive law variables
-                SetConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points);
+            // Compute constitutive law variables
+            SetConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points);
 
-                // Call the constitutive law to update material variables
-                mConstitutiveLawVector[point_number]->InitializeMaterialResponse(Values, GetStressMeasure());
+            // Rotate to local axes strain/F
+            if (is_rotated)
```
RiccardoRossi, line 131: what are u trying to do here?

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
```diff
@@ -171,15 +167,9 @@ void BaseSolidElement::FinalizeNonLinearIteration( const ProcessInfo& rCurrentPr
...
 {
-    // We finalize the material reponse if required
-    bool required = false;
-    for ( IndexType point_number = 0; point_number < mConstitutiveLawVector.size(); ++point_number ) {
-        if (mConstitutiveLawVector[point_number]->RequiresFinalizeMaterialResponse()) {
-            required = true;
-            break;
-        }
-    }
+    const bool required = mConstitutiveLawVector[0]->RequiresFinalizeMaterialResponse();
```
RiccardoRossi: same, here previous code is better

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
```diff
@@ -895,8 +889,9 @@ void BaseSolidElement::CalculateOnIntegrationPoints(
             for (IndexType point_number = 0; point_number < number_of_integration_points; ++point_number) {
                 // Compute element kinematics B, F, DN_DX ...
                 CalculateKinematicVariables(this_kinematic_variables, point_number, this->GetIntegrationMethod());
-
-                // Compute material reponse
+                // Compute constitutive law variables
+                SetConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points);
```
RiccardoRossi: ?? what is this

RiccardoRossi (commented):
this looks a pretty big change to me.

why do you need to cal lthe "SetConstitutiveVariables" so many times now?

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
AlejandroCornejo: ok! i did not think that this could happen

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
AlejandroCornejo: the main issue is that the check `IsElementRotated()` was done every time the `CalculateConstitutiveVariables` (so at each IP) and that was more expensive than putting the check outside the IP loop. I have an idea to encapsulate this better

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
AlejandroCornejo: done

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
AlejandroCornejo: now i've added a bool to the `CalculateConstitutiveVariables` in order to know if it's rotated. Now they are back as it was

AlejandroCornejo:
With the latests changes the `CalculateConstitutiveVariables` method remained as it was but adding a const bool to know if it's rotated or not.

AlejandroCornejo:
Besides I corrected the `FinalizeSolutionStep` and `InitializeSolutionStep` cause they were inconsistent (no ratations were performed)

AlejandroCornejo:
ping

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
```diff
@@ -836,7 +840,7 @@ void BaseSolidElement::CalculateOnIntegrationPoints(
                     CalculateKinematicVariables(this_kinematic_variables, point_number, this->GetIntegrationMethod());
 
                     // Compute material reponse
-                    CalculateConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points, GetStressMeasure());
+                    CalculateConstitutiveVariables(this_kinematic_variables, this_constitutive_variables, Values, point_number, integration_points, GetStressMeasure(), false);
```
RiccardoRossi, line 843: but then, don't you need to change all of the derived elements?

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.h
```diff
@@ -774,7 +774,8 @@ class KRATOS_API(STRUCTURAL_MECHANICS_APPLICATION) BaseSolidElement
         ConstitutiveLaw::Parameters& rValues,
         const IndexType PointNumber,
         const GeometryType::IntegrationPointsArrayType& IntegrationPoints,
-        const ConstitutiveLaw::StressMeasure ThisStressMeasure = ConstitutiveLaw::StressMeasure_PK2
+        const ConstitutiveLaw::StressMeasure ThisStressMeasure = ConstitutiveLaw::StressMeasure_PK2,
```
RiccardoRossi, line 777: this function is virtual, so making this change will imply changing all of the derived elements

applications/StructuralMechanicsApplication/custom_elements/small_displacement_bbar.cpp
```diff
@@ -112,7 +113,7 @@ void SmallDisplacementBbar::CalculateAll(
         if ( rRightHandSideVector.size() != mat_size )
             rRightHandSideVector.resize( mat_size, false );
 
-        rRightHandSideVector = ZeroVector( mat_size ); //resetting RHS
+        noalias(rRightHandSideVector) = ZeroVector( mat_size ); //resetting RHS
```
RiccardoRossi, line 116: just to tell (not really a review here, just to let you know) that you can use  the .clear()) method to do the same shorter

RiccardoRossi (approved):
overall ok to me. 

I am more and more convinced that our derivation strategy is too long here, and that we would be better off with standalone functions instead of implementing everything in the base class, but this is definitely off topic.

approving on my side

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
AlejandroCornejo: I did it, i changed the UL, TL and small displ elements :)

AlejandroCornejo:
@loumalouomega

loumalouomega:
I am out until Monday, does @RiccardoRossi agrees?

AlejandroCornejo:
> I am out until Monday, does @RiccardoRossi agree?

he approved the PR

loumalouomega:
Okay from my side, feel free to merge

AlejandroCornejo:
@loumalouomega you should remove the block :)

loumalouomega (approved):
Approving after approval from @RiccardoRossi

PR #10402, opened 2022-10-26
[Core] Run tests script - fix applications argument

ddiezrod:
**📝 Description**
The option to provide applications as arguments in run_tests files was not working properyl.
- choices should not only include the list of applications but also any combination of them, so better to remove it.
- type check was not correct, resulting in an error when the argument was provided


loumalouomega (changes requested):
You should adapt the run_test_mpi as well. Thanks

loumalouomega (approved):

PR #10403, opened 2022-10-26
[Core] Minor `Point` clean up and adding tests

loumalouomega:
**📝 Description**

Minor `Point` clean up and adding tests

- Remove duplicated includes (already defined in `define.h`)
- Consistent definition of Dimension (defining just once)
- Adding Point tests

**🆕 Changelog**

- [Remove duplicated includes (already defined in `define.h`)](https://github.com/KratosMultiphysics/Kratos/commit/14d63873d74f26365aa00d2cf80ffdb7d3dbc89c)
- [Consistent definition of Dimension (defining just once)](https://github.com/KratosMultiphysics/Kratos/commit/9b9f28d1ac7ac155fb67396a77690f4f15f5cfe7)
- [Adding Point tests](https://github.com/KratosMultiphysics/Kratos/commit/0f61dbc561f7f0e1f9e1933a295784067be1395b)

PR #10404, opened 2022-10-26
[Core] Adding `Distance` method to `Point`

loumalouomega:
**📝 Description**

I just realized that every time we want to compute the distance between two points, we need to compute it manually. Even if trivial, it is very easy to avoid this "code duplication"  by adding a `Distance` method to the `Point`. This PR adds `Distance` method plus includes #10403

**🆕 Changelog**

- [Remove duplicated includes (already defined in `define.h`)](https://github.com/KratosMultiphysics/Kratos/commit/14d63873d74f26365aa00d2cf80ffdb7d3dbc89c)
- [Consistent definition of Dimension (defining just once)](https://github.com/KratosMultiphysics/Kratos/commit/9b9f28d1ac7ac155fb67396a77690f4f15f5cfe7)
- [Adding Point tests](https://github.com/KratosMultiphysics/Kratos/commit/0f61dbc561f7f0e1f9e1933a295784067be1395b)
- [Adding Distance definition](https://github.com/KratosMultiphysics/Kratos/commit/48251a3d41bc7bf04eca0374b4cd947dabe9110b)
- [Adding SquaredDistance](https://github.com/KratosMultiphysics/Kratos/pull/10404/commits/a655d7d855727e005b36cc16e72420bdd178903d) By recommendation of @matekelemen to avoid square root calculation in faster computations
- [Test for Distance](https://github.com/KratosMultiphysics/Kratos/commit/105ab58524cefd4f8456ec4980ec8abe920bb290)
- [Adding test SquaredDistance](https://github.com/KratosMultiphysics/Kratos/pull/10404/commits/18f918578acebf96155f0c8c46b9c7afd893c563)


kratos/geometries/point.h
```diff
@@ -150,13 +146,23 @@ class Point : public array_1d<double, 3>
     ///@name Operations
     ///@{
 
+    /**
+     * @brief This method computes the distance betwee this point and another one
+     * @param rOtherPoint The other point to compute the distance
+     * @return The distance between this and another point     
+     */
+    double Distance(const Point& rOtherPoint)
```
philbucher: const!

kratos/tests/cpp_tests/geometries/test_point.cpp
```diff
@@ -0,0 +1,109 @@
...
+        Point point;
+        KRATOS_CHECK_STRING_EQUAL(point.Info(), "Point");
+    }
+
+    /**
+     *  Here the Distance is test
+     */
+    KRATOS_TEST_CASE_IN_SUITE(Distance, KratosCoreGeometriesFastSuite)
+    {
+        Point point;
```
philbucher: const

philbucher (changes requested):
Missing const

kratos/geometries/point.h
tteschemacher: I suppose this could be similarly using the CalculateDistance function.

Not sure about the efficiency, though.

loumalouomega: That function is in the geometry. A point is not a geometry (Point2D and Point3D)

philbucher (approved):

PR #10405, opened 2022-10-26
[Core] Adding FastMinimalDistanceOnLine to GeometricalProjectionUtilities

loumalouomega:
**📝 Description**

Adding FastMinimalDistanceOnLine to GeometricalProjectionUtilities

**🆕 Changelog**

- [Adding FastMinimalDistanceOnLine](https://github.com/KratosMultiphysics/Kratos/commit/319f945f8f6ff4d11f2b1e18a5e58e0351badaa9)
- [Adding corresponding test for FastMinimalDistanceOnLine](https://github.com/KratosMultiphysics/Kratos/commit/11bc0d28bb62b6b0ff963053febea3098755295e)

### DIFF
--- a/kratos/utilities/geometrical_projection_utilities.h
+++ b/kratos/utilities/geometrical_projection_utilities.h
@@ -221,8 +221,8 @@ public:
         if (rGeometry.IsInside(projected_point.Coordinates(), projected_local, Tolerance)) {
             return projected_distance;
         } else {
-            const double distance_a = norm_2(rPoint.Coordinates() - rGeometry[0].Coordinates()); /// TODO: Replace with distance method if ever exists
-            const double distance_b = norm_2(rPoint.Coordinates() - rGeometry[1].Coordinates()); /// TODO: Replace with distance method if ever exists
+            const double distance_a = rPoint.Distance(rGeometry[0]);
+            const double distance_b = rPoint.Distance(rGeometry[1]);
             return std::min(distance_a, distance_b);
         }
     }
--- a/kratos/utilities/geometrical_projection_utilities.h
+++ b/kratos/utilities/geometrical_projection_utilities.h
@@ -11,8 +11,7 @@
 //                   Philipp Bucher
 //
 
-#if !defined(KRATOS_GEOMETRICAL_PROJECTION_UTILITIES)
-#define KRATOS_GEOMETRICAL_PROJECTION_UTILITIES
+#pragma once
 
 // System includes
 
@@ -183,7 +182,7 @@ public:
      */
     template<class TGeometryType>
     static inline double FastProjectOnLine(const TGeometryType& rGeometry,
-                                           const Point& rPointToProject,
+                                           const PointType& rPointToProject,
                                            PointType& rPointProjected)
     {
         const array_1d<double, 3>& r_p_a = rGeometry[0].Coordinates();
@@ -197,6 +196,35 @@ public:
         rPointProjected.Coordinates() = r_p_a + factor * ab;
 
         return norm_2(rPointProjected.Coordinates()-p_c);
+    }
+
+
+    /**
+     * @brief Computes the minimal distance to a line
+     * @details Projects over a line and if the point porjected is inside the line that distance is taken into consideration, otherwise the minimal between the two points in the line is considered
+     * @tparam TGeometryType The type of the line
+     * @param rGeometry The line where compute the distance
+     * @param rPoint The point to compute the distance
+     * @param Tolerance Tolerance to check it falls inside the line
+     * @return Distance The distance between point and line
+     */
+    template<class TGeometryType>
+    static inline double FastMinimalDistanceOnLine(
+        const TGeometryType& rGeometry,
+        const PointType& rPoint,
+        const double Tolerance = 1.0e-9
+        )
+    {
+        PointType projected_point;
+        const double projected_distance = FastProjectOnLine(rGeometry, rPoint, projected_point);
+        typename TGeometryType::CoordinatesArrayType projected_local;
+        if (rGeometry.IsInside(projected_point.Coordinates(), projected_local, Tolerance)) {
+            return projected_distance;
+        } else {
+            const double distance_a = norm_2(rPoint.Coordinates() - rGeometry[0].Coordinates()); /// TODO: Replace with distance method if ever exists
+            const double distance_b = norm_2(rPoint.Coordinates() - rGeometry[1].Coordinates()); /// TODO: Replace with distance method if ever exists
+            return std::min(distance_a, distance_b);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


kratos/utilities/geometrical_projection_utilities.h
```diff
@@ -199,6 +198,35 @@ class GeometricalProjectionUtilities
         return norm_2(rPointProjected.Coordinates()-p_c);
     }
 
+
+    /**
+     * @brief Computes the minimal distance to a line
+     * @details Projects over a line and if the point porjected is inside the line that distance is taken into consideration, otherwise the minimal between the two points in the line is considered
```
philbucher: Isnt this the same a the new project methods that @tteschemacher added?

Also there is a typo

loumalouomega: If it is already implemented perfect, can you tell me where?

loumalouomega:
```suggestion
     * @details Projects over a line and if the point projected is inside the line that distance is taken into consideration, otherwise the minimal between the two points in the line is considered
```

tteschemacher: I don't think I have implemented the exact same function. However, you can actually check to include this in the respectively derived function under:
https://github.com/KratosMultiphysics/Kratos/blob/8a9721a05c73c8976139705dc1880d50389a6e74/kratos/geometries/geometry.h#L2660
or 
https://github.com/KratosMultiphysics/Kratos/blob/8a9721a05c73c8976139705dc1880d50389a6e74/kratos/geometries/geometry.h#L2687

tteschemacher: wait. sorry I was confused.

This one should be the correct interface, no?:
https://github.com/KratosMultiphysics/Kratos/blob/8a9721a05c73c8976139705dc1880d50389a6e74/kratos/geometries/geometry.h#L2881

loumalouomega: No, it is not the same, here I am computing the closest distance between a line and a point

tteschemacher: > wait. sorry I was confused.
> 
> This one should be the correct interface, no?:
> 
> https://github.com/KratosMultiphysics/Kratos/blob/8a9721a05c73c8976139705dc1880d50389a6e74/kratos/geometries/geometry.h#L2881

But this one, then?

loumalouomega: I didn't know about that, I will see what can I refactor

loumalouomega: I cannot derive this method because the definition is:

~~~c++
    /**
    * @brief Computes the distance between an point in
    *        global coordinates and the closest point
    *        of this geometry.
    *        If projection fails, double::max will be returned.
    * @param rPointGlobalCoordinates the point to which the
    *        closest point has to be found.
    * @param Tolerance accepted orthogonal error.
    * @return Distance to geometry.
    *         positive -> outside of to the geometry (for 2D and solids)
    *         0        -> on/ in the geometry.
    */
~~~

In my case I project, but if it fails I return the closest distance between the nodes of the line. Not exactly the same, if you want to create new method like `CalculateProjectedDistance` or `ClaculateClosestDistance`, etc..., Ok. But this method for me doesn't meet requirements

kratos/utilities/geometrical_projection_utilities.h
```diff
@@ -199,6 +198,35 @@ class GeometricalProjectionUtilities
...
+    {
+        PointType projected_point;
+        const double projected_distance = FastProjectOnLine(rGeometry, rPoint, projected_point);
+        typename TGeometryType::CoordinatesArrayType projected_local;
+        if (rGeometry.IsInside(projected_point.Coordinates(), projected_local, Tolerance)) {
+            return projected_distance;
+        } else {
+            const double distance_a = norm_2(rPoint.Coordinates() - rGeometry[0].Coordinates()); /// TODO: Replace with distance method if ever exists
+            const double distance_b = norm_2(rPoint.Coordinates() - rGeometry[1].Coordinates()); /// TODO: Replace with distance method if ever exists
+            return std::min(distance_a, distance_b);
```
matekelemen: Can you please add a version that doesn't take the square root? I imagine there are a lot of cases in which you just need sorting (closest point, farthest point, ...) but not the actual distance. We'd be saving a lot of unnecessary work in those cases. This also applies to the other PR.

This version (the one that computes the distance) could just call the one that computes the square, and take its root so there'd be no code duplication (+ there'd be no performance overhead because this can surely be inlined).
```cpp
double FastMinimalSquaredDistanceOnLine(...)
{...}

double FastMinimalDistanceOnLine(...)
{
    return std::sqrt(FastMinimalSquaredDistanceOnLine(...));
}
```

matekelemen: Actually, this comes up so often that I'd just define a dedicated ```Distance``` class that stores the squared distance, and produces its square root on demand only. I think it'd be a good idea to return such an object from every distance calculation.
```cpp
template <class TValue>
class Distance
{
public:
    Distance(TValue squared) : mSquared(squared) {}

    ...

    bool operator<(Distance Left, Distance Right) {return Left.mSquared < Right.mSquared;}

    explicit operator TValue() const noexcept {return std::sqrt(mSquared);}

private:
    TValue mSquared;
};
```

*Having an additional ```bool``` member would also allow storing the root if it's already been computed once, which would help to avoid computing the root repeatedly.*

loumalouomega: See https://github.com/KratosMultiphysics/Kratos/pull/10404

matekelemen: > See #10404

I'm talking about avoiding the square root calculation unless it's absolutely necessary. Does that PR address that?

loumalouomega: Well, can be adapted

loumalouomega: Done @matekelemen

matekelemen: thanks ❤️

kratos/utilities/geometrical_projection_utilities.h
```diff
@@ -199,6 +198,35 @@ class GeometricalProjectionUtilities
...
+        )
+    {
+        PointType projected_point;
+        const double projected_distance = FastProjectOnLine(rGeometry, rPoint, projected_point);
+        typename TGeometryType::CoordinatesArrayType projected_local;
+        if (rGeometry.IsInside(projected_point.Coordinates(), projected_local, Tolerance)) {
+            return projected_distance;
+        } else {
+            const double distance_a = norm_2(rPoint.Coordinates() - rGeometry[0].Coordinates()); /// TODO: Replace with distance method if ever exists
+            const double distance_b = norm_2(rPoint.Coordinates() - rGeometry[1].Coordinates()); /// TODO: Replace with distance method if ever exists
```
philbucher: the method exists now :)

loumalouomega:
Yes, but I am one week off :)

philbucher (approved):

PR #10408, opened 2022-10-26
Update templates

philbucher:
More updates to the templates with some modernizations


rubenzorrilla (approved):
I think all these make sense. Thanks for the clean up.

matekelemen:
This comment isn't really related to this PR but more to the class "template":

I don't think having boilerplate for a class gives the right idea to people. Constructors and destructors should be defined/removed as necessary for the given situation, and having a copy-pasta boilerplate actively discourages the developer of thinking through what they need.

philbucher:
I agree in general, but IMO the templates are pretty lightweight and give more a general structure of how the file should look like rather than a complete implementation

PR #10411, opened 2022-10-27
[FastPR][Core] Operations addition

rubenzorrilla:
**📝 Description**
This PR adds the base class for `Operations`. This will be used later on in the multistage.

@KratosMultiphysics/technical-committee I decided to separate it from the registry / multistage stuff to make the final review easier.

EDIT: @KratosMultiphysics/technical-committee note that I already added the registry macro at the bottom of `operation.h`.

### DIFF
--- a/kratos/operations/operation.h
+++ b/kratos/operations/operation.h
@@ -1,0 +1,203 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
+//
+//  Main authors:    Carlos Roig
+//                   Ruben Zorrilla
+//
+
+#pragma once
+
+// System includes
+#include <string>
+#include <iostream>
+
+// External includes
+
+// Project includes
+#include "containers/model.h"
+#include "includes/define.h"
+#include "includes/kratos_flags.h"
+#include "includes/kratos_parameters.h"
+
+namespace Kratos
+{
+
+///@name Kratos Classes
+///@{
+
+/**
+ * @class Operation
+ * @ingroup KratosCore
+ * @brief The base class for all operations in Kratos.
+ * @details The operation is the base class for all operations and defines the interface for them.
+ * Execute method is used to execute all the Operation algorithms. Operation parameters must be passed at construction time.
+  @author Carlos Roig
+  @author Ruben Zorrilla
+*/
+class Operation
+{
+public:
+    ///@name Type Definitions
+    ///@{
+
+    /// Pointer definition of Operation
+    KRATOS_CLASS_POINTER_DEFINITION(Operation);
+
+    ///@}
+    ///@name Life Cycle
+    ///@{
+
+    /// Default constructor.
+    explicit Operation() {}
+
+    /// Destructor.
+    virtual ~Operation() {}
+
+    /// Copy constructor.
+    //TODO: Check. It is required by the registry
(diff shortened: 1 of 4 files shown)


philbucher (dismissed):

kratos/operations/operation.h
```diff
@@ -0,0 +1,191 @@
...
+    {                                                                                                           \
+        std::string all_path = std::string("Operations.All.") + operation_name;                                 \
+        if (!Registry::HasItem(all_path)) {                                                                     \
+            auto& r_operation_item = Registry::AddItem<RegistryItem>(all_path);                                 \
+            r_operation_item.AddItem<RegistryValueItem<Operation>>("Prototype", operation_prototype);           \
+        } else {                                                                                                \
+            KRATOS_ERROR << "Operation '" << operation_name << "' is already registered." << std::endl;         \
+        }                                                                                                       \
+        std::string module_path = std::string("Operations.") + module_name + std::string(".") + operation_name; \
+        if (!Registry::HasItem(module_path)) {                                                                  \
```
pooyan-dadvand, line 166: Shall we do this check? or we should leave it to registry to send an error?

kratos/operations/operation.h
```diff
@@ -0,0 +1,191 @@
...
+///@{
+
+#ifndef KRATOS_REGISTER_OPERATION
+#define KRATOS_REGISTER_OPERATION(                                                                              \
+    module_name,                                                                                                \
+    operation_name,                                                                                             \
+    operation_prototype)                                                                                        \
+    {                                                                                                           \
+        std::string all_path = std::string("Operations.All.") + operation_name;                                 \
+        if (!Registry::HasItem(all_path)) {                                                                     \
```
pooyan-dadvand, line 159: I understand that you are checking to send a custom error. Isn't it?

If this is the case, we may need to create (not in this PR) a AddItem(path, error_lambda) to make this more straightforward

pooyan-dadvand (commented):
I don't revoke the approval but there are some minor comments

kratos/operations/operation.h
rubenzorrilla: (see above answer)

kratos/operations/operation.h
rubenzorrilla: True. As it is right now we're checking twice, and the custom error isn't that custom... About the the `AddItem` with error thrown, it is not a bad idea.

Considering that we're about to start the PR with all the new registry stuff, and that this macro is almost identical to that of the `Process` that I'll add in there, I think we'd better discuss it in the upcoming one.

sunethwarna (approved):

rubenzorrilla:
`ShallowWaterApplication` tests are failing. Strange as these have nothing to do with this PR. @miguelmaso FYI.

kratos/operations/operation.h
pooyan-dadvand: Agree that we shall do it in a separate PR

kratos/operations/operation.h
pooyan-dadvand: But here you are avoiding the error. Is it intentional?

loumalouomega:
> `ShallowWaterApplication` tests are failing. Strange as these have nothing to do with this PR. @miguelmaso FYI.

Timeout, maybe some small tests must be moved to nightly

miguelmaso:
> > `ShallowWaterApplication` tests are failing. Strange as these have nothing to do with this PR. @miguelmaso FYI.
> 
> Timeout, maybe some small tests must be moved to nightly

Does anyone has more information about the failing test? It seems to be a specific test (Boussinesq) with clang but I don't know how to reproduce it

philbucher:
no clue
did you try to reproduce it locally?

roigcarlo:
Trying in my machine but so far works...

PR #10413, opened 2022-10-28
[Core] Adding ComputeClosestLineLineIntersection to IntersectionUtilities

loumalouomega:
**📝 Description**

Adding ComputeClosestLineLineIntersection to IntersectionUtilities. Implements the shortest line between two lines in 3D following http://paulbourke.net/geometry/pointlineplane/

**🆕 Changelog**

- [Adding ComputeClosestLineLineIntersection to IntersectionUtilities](https://github.com/KratosMultiphysics/Kratos/commit/2405d179b88e5eea3fe89248955da11669170543)
- [Adding test](https://github.com/KratosMultiphysics/Kratos/commit/de2d91899f8b48117f8ac5aaa56c5b5a9d88ebcb)


rubenzorrilla (changes requested):
Fine in general terms. I only have a few comments.
- About the style guide, though this doesn't follow ours, it matches the one in the algorithm documentation, so I'd leave it as you did.
- The name is a bit misleading for me, I'd call it `ComputeShortestLineBetweenTwoLines` since by line-line intersection (though there is the closest in front) what one could expect as result is a point.
- As far as I understand, current implementation only works for `Line2D3` and  `Line3D3` although one can pass a quadratic line as argument. To avoid this, I'd specialize the method for these two types in the cpp. In the 2D case, we can even make the implementation simpler by omitting all the z-dependent values.
- Concerning the return, I've two comments. First one is that there are some cases in which we're returning two points with (0,0,0) coordinates. I guess these are the cases in which the lines effectively intersect. I think this can be misleading. Second one is, is it ensured that the points in the return line are such that the first point is the one corresponding to the first line in the arguments and so for the second?

rubenzorrilla:
BTW, which is the use case for this?

loumalouomega:
> BTW, which is the use case for this?

A signed distance to line process that is on the works

loumalouomega:
> Fine in general terms. I only have a few comments.
> 
>     * About the style guide, though this doesn't follow ours, it matches the one in the algorithm documentation, so I'd leave it as you did.

A priori, I adapted to our style

>     * The name is a bit misleading for me, I'd call it `ComputeShortestLineBetweenTwoLines` since by line-line intersection (though there is the closest in front) what one could expect as result is a point.

Okay

>     * As far as I understand, current implementation only works for `Line2D3` and  `Line3D3` although one can pass a quadratic line as argument. To avoid this, I'd specialize the method for these two types in the cpp. In the 2D case, we can even make the implementation simpler by omitting all the z-dependent values.

Better to do a static check with C++17 and avoid extra classes. It works with linear lines

>     * Concerning the return, I've two comments. First one is that there are some cases in which we're returning two points with (0,0,0) coordinates. I guess these are the cases in which the lines effectively intersect. I think this can be misleading. Second one is, is it ensured that the points in the return line are such that the first point is the one corresponding to the first line in the arguments and so for the second?

I will try to figure out a better way

loumalouomega:
>     * The name is a bit misleading for me, I'd call it `ComputeShortestLineBetweenTwoLines` since by line-line intersection (though there is the closest in front) what one could expect as result is a point.

Done

>     * As far as I understand, current implementation only works for `Line2D3` and  `Line3D3` although one can pass a quadratic line as argument. To avoid this, I'd specialize the method for these two types in the cpp. In the 2D case, we can even make the implementation simpler by omitting all the z-dependent values.

Check added

>     * Concerning the return, I've two comments. First one is that there are some cases in which we're returning two points with (0,0,0) coordinates. I guess these are the cases in which the lines effectively intersect. I think this can be misleading. Second one is, is it ensured that the points in the return line are such that the first point is the one corresponding to the first line in the arguments and so for the second?

I took a look at the code, and you can see that if  it doesn't fill the points until the end, otherwise if fails it will return an empty PointerVectorSet of points, not 2 points at (0,0,0), if you try to build a geometry from this you will have an execution error. The implementation using this method should check the size of the vector before using it to avoid this problem (In my implementation I do that)

rubenzorrilla:
> I took a look at the code, and you can see that if it doesn't fill the points until the end, otherwise if fails it will return an empty PointerVectorSet of points, not 2 points at (0,0,0), if you try to build a geometry from this you will have an execution error. The implementation using this method should check the size of the vector before using it to avoid this problem (In my implementation I do that)

Now I see. But what about my second observation?

loumalouomega:
> > I took a look at the code, and you can see that if it doesn't fill the points until the end, otherwise if fails it will return an empty PointerVectorSet of points, not 2 points at (0,0,0), if you try to build a geometry from this you will have an execution error. The implementation using this method should check the size of the vector before using it to avoid this problem (In my implementation I do that)
> 
> Now I see. But what about my second observation?

Can you reformulate?

rubenzorrilla:
> > > I took a look at the code, and you can see that if it doesn't fill the points until the end, otherwise if fails it will return an empty PointerVectorSet of points, not 2 points at (0,0,0), if you try to build a geometry from this you will have an execution error. The implementation using this method should check the size of the vector before using it to avoid this problem (In my implementation I do that)
> > 
> > 
> > Now I see. But what about my second observation?
> 
> Can you reformulate?

I refer to this "is it ensured that the points in the return line are such that the first point is the one corresponding to the first line in the arguments and so for the second? ".

Your return type is a two-point line. My question is, it is guaranteed that the first point is located in the first argument line and so for the second or this is not deterministic?

loumalouomega:
> > > > I took a look at the code, and you can see that if it doesn't fill the points until the end, otherwise if fails it will return an empty PointerVectorSet of points, not 2 points at (0,0,0), if you try to build a geometry from this you will have an execution error. The implementation using this method should check the size of the vector before using it to avoid this problem (In my implementation I do that)
> > > 
> > > 
> > > Now I see. But what about my second observation?
> > 
> > 
> > Can you reformulate?
> 
> I refer to this "is it ensured that the points in the return line are such that the first point is the one corresponding to the first line in the arguments and so for the second? ".
> 
> Your return type is a two-point line. My question is, it is guaranteed that the first point is located in the first argument line and so for the second or this is not deterministic?

A priori, is deterministic, and the points belong to the lines (not our object line, to the infinite lines defined by the two points which define the line objects). If there is no solution, with the corresponding checks, it will return an empty array of points.

kratos/utilities/intersection_utilities.h
```diff
@@ -233,6 +234,89 @@ class IntersectionUtilities
...
+        const TGeometryType& rSegment1,
+        const TGeometryType& rSegment2
+        )
+    {
+        // Zero tolerance
+        const double zero_tolerance = std::numeric_limits<double>::epsilon();
+
+        // Check geometry type
+        const auto geom_type = rSegment1.GetGeometryType();
+        KRATOS_ERROR_IF((geom_type != GeometryData::KratosGeometryType::Kratos_Line2D2) && geom_type != GeometryData::KratosGeometryType::Kratos_Line3D2) << "The geometry type is not correct, it is suppossed to be a linear line" << std::endl;
```
rubenzorrilla: And if we do an `if consexpr` to do the check "at compile time"?

loumalouomega: It is not possible as GetGeometryType is not a static method (I already tried)

rubenzorrilla: But you can't use the type from the template argument?

loumalouomega: No, the template argument only ensures that you can pass a geometry of nodes or points or any derived point class, if you enforce manually setting Line3D2<Point> or Line2D2<Point> for example, that will not work anyway because we don't know in compilation time that type, and `GetGeometryType` is not a static method. If you want to do that, that it will practical I admit, we need to add a static method to every geometry time returning it's geometry type. We will be using static polimorphism and I guess this will not work well with the dynamic polimorphism we use for geometries. @matekelemen do you think so?

matekelemen: @loumalouomega  is unfortunately right, it can't be `constexpr` because `Geometry::GetGeometryType` is `virtual` (dynamic poly), which is a purely runtime abstraction. You *do* have some other options though.

You could use [`std::enable_if`](https://en.cppreference.com/w/cpp/types/enable_if) to restrict the types of geometries `ComputeShortestLineBetweenTwoLines` can be called with. Here's an [example on godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIAOykrgAyeAyYAHI%2BAEaYxCCSAJykAA6oCoRODB7evgGp6ZkCoeFRLLHxSbaY9o4CQgRMxAQ5Pn6BdpgOWQ1NBCWRMXEJyQqNza15HeP9YYPlw0kAlLaoXsTI7BzmAMxhyN5YANQmO25jxGHAp9gmGgCCu/uHmCdnTheYrDd3jw%2B/Fy8DiOAHFMKg2ARiABPX4mfxWB5HZFHABueGaXjERzG6BAIAuVyOwEwBAAKtCUpgIEsjmgGGMTgjiKT1gwTmYzGCIaSYeYzKcrP4ACInf4iwVw/4PQHAgYaI4gUHgyF88WI%2B4onEEPEEqFEknkynU2n0xmoVFxS7HeEWFkENkcszy/mSiXi4WS6V/e6yghHAZcRXKnlQ2Hq35a3H4wlGYmkilUml0gTmy3Ea2vW32x38wOunZCz0er0%2BgFQoH%2BoQ8gDyBAQcW5quhwabvPDjwRkZR0b1lzjhsTJpTDP9FqteBtzNZxHZ/OrbDrDeIbbDBaLcIlhalPoImBYKQMe7ebm7WvPdIMCgURzJq5hpDPF57OvxjCY0XoAH08FQvwRTnOV8QDwBQvwUVhMC/VFALvFV21IF0dmwJlT38NxtV1UDwMg6DYPvaFELmLgblIaJUE8G43lFCtMCfc8fgeVFUEnI50FQBdSQQK4awYe4BggM1/Tg0M%2BTMAA2I5iC/ElROhJZNw1KNgLQLx/UAwCpJk%2BCwwAOkHY1kw0s5MLfBh0FoUt4WLMsHjCf0WCYMIaUUp8CPjOTSy1eUjloOYvw0LyUUDXz/JI7ckRRTil0bHSYSOVB6ziKzIuRAB6NL2M4%2BseL4gTZObBTCyODKT1OT0dlFK0SCVBhUCORyCGQbi4yoLwGG6AQnyy0MWuAXj%2BLmCA/PCAKiqUlEeshPqBoEkaoK4canwyqauNywbwggRLl3Gkq0rKyryqOar4iOOqGqYJq%2BqONqOrqBhN2FDgVloTgAFZeD8DgtFIVBOFPSxrBxNYNizMwdh4UgCE0Z6VgbJgsHiGlSAAaxAN6uF0nZJHEgAODQuGxvGCZ2cT9E4SReBYCQNA0Ugvp%2Bv6OF4BQQDp6Hvue0g4FgGBEBATBVC6NSSHISgmmABRlEMGohAQVAAHcvshtADzoS6sml8JaDlxWGd4VWUjoYZgC4TlSEN43iBrNTdaVmGgiF5B7mISXOF4QWugafAvt4fhBBEMR2CkGRBEUFR1E50hdC4fRDGMaxrH0PBolZyAVlQFJ7tZjgAFpcXK0xAcsLh/BZkHNj0XEwi12X5ft7heChTAtkhhXiCYFJOB4F73s%2Bh2mewJ2ReII5VFx8Tc/EyRiWQZAjjN3SzCOCAAasSxSCOXBCBIDkIaWJuYaWOGvkRyhe44SnSGpnZ/F03Gdkf/x8dxyQzfEt6J/pgf3dsNmoaPtzPmEAkCe2QCPMWEAJZSxljreuysDYQiNkeTWsC7b6wtkgq2KADBGC4G9WmmC1b0GtrbeBDswEuzdszR2XsdRhF/v7YQohxAhyYeHNQDtdA7DjkYFAicN5%2BVTvADOWcsg53zjqQuAiLCl3LusSusdq4MFrnAvW3cm4slbrwduncNEXw%2Bt/KOg9h5EFHuPSe09LzxwXm9XSGh7ErzXknLe%2BAzF70WofTmx9SDwzPsjV6l8qYgAhvYh%2BEMNCJDev4RIpN/CxwwUzFm/8OZaBWDzfmYCIEUCga7GB2t0EaKIcgjWAhVGFKjpbEhIBUTIBSCkaCXBEhfiPJgMYX4LHFKtjbYQ5Co6ULyb/MB3sGE0KYYHVh0h2FKE4VHXQZheEJ2LjYIRadkaZ2zpwNK0Yi7r1kWXVYCjg62HoSotBfTIbN20aQXRXdG4GP7sYzgQ9hbuIsQ1BQtS0RNN0q0xkq8ZGb23u43YniAHeJPgjYYASKbBNCRocJXBInRNieJeJRjGa/1ZuzI%2BKw0aSFxrpD%2BENEgTzeiSieZNAk7EeZimhqTYbkw4GYWlvAkngrSSsdMGRnCSCAA%3D%3D). The advantage of this approach is that you'll get errors at compile time if you try to misuse it, but the drawback is that you either have to store the derived types directly (`Line2D2`s or `Line3D2`s), or `dynamic_cast` the base class pointers before you invoke this function on them. I'm not a fan of any type of casting and it's unlikely to be any faster than checking `GetGeometryType`, but at least the user has to go out of their way to misuse this function, so it's a bit safer. (*btw C++20 concepts would make the unholy `enable_if`s much nicer*)

The other solution, my preferred one, is to just loop through all edges in the geometry.

rubenzorrilla: OK. OFC you're both right. 

In any case I think that in this case the good approach would be to override the method with `Line2D2` and `Line3D2` being the node type the template argument. In that way you don't need to do the type check and you can optimize the 2D implementation by suppressing all the z-component operations.

loumalouomega: The only way I could do it was with explicit specialization, explicit instantiation was giving me compilation errors I could not solve

loumalouomega:
I just realized this only makes sense in 3D, because if the lines are parallel it will fails as the number of possible lines is infinite, therefore in 2D it will return always the same two points (intersection point), and this is already implemented. I will revert the specialization because breaks the philosophy of these utilities

loumalouomega:
@rubenzorrilla after the discussion I think the best solution is to keep it as it was, the template specialization duplicated code and in addition specialize for a 2D case was completely unnecessary as this is always a point and there is already a method for that

matekelemen:
> in 2D it will return always the same two points (intersection point), and this is already implemented.

So why not invoke the already implemented function in the 2D case? I think it would be nice if the user didn't have to use different functions in 2D and 3D.

loumalouomega:
> > in 2D it will return always the same two points (intersection point), and this is already implemented.
> 
> So why not invoke the already implemented function in the 2D case? I think it would be nice if the user didn't have to use different functions in 2D and 3D.

But the function I am adding is to find the closest line between two lines, in 2D does not make sense, as the only option is a perpendicular line with infinite solutions

matekelemen:
> But the function I am adding is to find the closest line between two lines, in 2D does not make sense, as the only option > is a perpendicular line with infinite solutions

Well it's and edge case. I'd keep it as well because someone might require it for a generalized algorithm, but it's your call.

loumalouomega:
> > But the function I am adding is to find the closest line between two lines, in 2D does not make sense, as the only option  is a perpendicular line with infinite solutions
> 
> Well it's and edge case. I'd keep it as well because someone might require it for a generalized algorithm, but it's your call.

I could define a perpendicular line, for example from the centre of one of the lines, and in that case the 2D will make sense. In any case the template specialization is a no-go IMHO, and template instantiation does not work due to the template structure of the geometry classes, which makes this extremely complex

loumalouomega:
@matekelemen I added perpendicular line for parallel lines, any case I would left the code as it is

loumalouomega:
Hello @rubenzorrilla

loumalouomega:
Hello @rubenzorrilla

loumalouomega:
Hello @rubenzorrilla

kratos/utilities/intersection_utilities.h
```diff
@@ -233,6 +237,118 @@ class IntersectionUtilities
...
+    static PointerVector<Point> ComputeShortestLineBetweenTwoLines(
+        const TGeometryType& rSegment1,
+        const TGeometryType& rSegment2
+        )  
+    {
+        // Zero tolerance
+        const double zero_tolerance = std::numeric_limits<double>::epsilon();
+
+        // Check geometry type
+        KRATOS_ERROR_IF(rSegment1.GetGeometryType() != GeometryData::KratosGeometryType::Kratos_Line3D2) << "The geometry type is not correct, it is suppossed to be a linear 3D line" << std::endl;
```
rubenzorrilla: 1. If the implementation only supports `Line3D2` the template argument must be in my opinion the point type. 

2. On top of this, maybe I'm missing something, but isn't the `Line2D2` a particular in-plane case of the 3D one? As far as I understand the same algorithm should work in both cases.

loumalouomega: Nope, in 2D the closest line is always a perpendicular line or intersection, I tried this is simpler than specialize (I am the first that if I could use constexpr or templates I would use it)

rubenzorrilla: Yes, we already discussed about specializing the implementation. My comment was regarding the error thrown, why are we forbidding using this with 2D lines? Shouldn't it work as in the 3D case with intersected or aligned lines?

loumalouomega: Technically, yes, I can add them. I will see if I can check geometry family instead

loumalouomega: Okay, @rubenzorrilla last change should work both 2D and 3D

rubenzorrilla: Perfect. This was exactly my point.

rubenzorrilla (approved):
Ready.

PR #10416, opened 2022-10-30
[Ci] some cleanups

philbucher:
This is not necessary any more since we now use Ubuntu 22.04 in the Ci
see #10170

Also excluding some stuff from being run in the CI


roigcarlo (approved):

roigcarlo (approved):
I like windows change :P

PR #10418, opened 2022-11-02
[FastPR][Core] Minor in kernel info formatting

rubenzorrilla:
**📝 Description**
Minor change in the formatting of the kernel output.

### DIFF
--- a/kratos/sources/kernel.cpp
+++ b/kratos/sources/kernel.cpp
@@ -97,14 +97,12 @@ void Kernel::PrintData(std::ostream& rOStream) const {
     rOStream << std::endl;
     rOStream << "Modelers:" << std::endl;
     KratosComponents<Modeler>().PrintData(rOStream);
-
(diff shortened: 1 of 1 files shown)


roigcarlo (approved):

kratos/sources/kernel.cpp
```diff
@@ -97,14 +97,12 @@ void Kernel::PrintData(std::ostream& rOStream) const {
     rOStream << std::endl;
     rOStream << "Modelers:" << std::endl;
     KratosComponents<Modeler>().PrintData(rOStream);
-
+    rOStream << std::endl;
     rOStream << "Loaded applications:" << std::endl;
```
philbucher, line 101:
```suggestion
    rOStream << "\nLoaded applications:" << std::endl;
```

rubenzorrilla: I'll leave it as it is to keep the way its done above.

PR #10419, opened 2022-11-02
[FastPR][Core]Missing Create() in modelers

rubenzorrilla:
**📝 Description**
This adds the missing `Create` method to these two Python modelers. These will be used later on by the multistage factory call.


rubenzorrilla:
Closing as we'll probably build the Python `Modelers` from the Python module.

PR #9318, opened 2021-11-10
[Core] Add variables to registry

pooyan-dadvand:
**📝 Description**
This is a continuation of #9308 and should be merged after that

Here I am adding the automatic registration directly to the variables constructors. This is assuming that we don't construct variables in the performance-intensive parts. 

The copy constructor doesn't register because it assumes that the variable should have been already registered.

The registry will keep a copy of that variable so we can use this copy in our data value container to avoid problems with local variables that may go out of scope:  #3789

**🆕 Changelog**
- Adding RegisterVariables to constructor of the Variables

PR #9308, opened 2021-11-08
[Core] Add registry

pooyan-dadvand:
**📝 Description**
This PR adds a new singleton `Registry` class.  It stores a tree of `RegistryItem` objects and can access them via "." connected labels like the way `Model` accesses its sub model parts.  

The `RegistryItem` base class can be used to store sub-items and provides the common interface. The `RegistryValueItem` is derived from `RegistryItem` and will store **a copy** of the given value and destroy it at its end of life.

**🆕 Changelog**
- Adding `Registry` Class
- Adding `RegistryItem` and `RegistryValueItem`
- Small change to the run_cpp_tests.py to take the tests to run as its argukment

PR #10420, opened 2022-11-02
Python registry

rubenzorrilla:
**📝 Description**
This finally adds the Python extension to the `Registry`. In order to see the usages, I've also added the new model-parameters factory as well as the changes in our current process one (which should be eventually substituted by the former).

This is 100% compatible with our ongoing multi-stage, but I decided to split the PR to ease the merging. 

Also note that this has been extensively discussed by the @KratosMultiphysics/technical-committee so it is not my proposal but the one of the KTC.

**IMPORTANT**: I required some features from Python 3.8 so I'm increasing the minimum version.


kratos/processes/process.h
```diff
@@ -253,6 +254,26 @@ class Process : public Flags
 ///@name Type Definitions
 ///@{
 
+#ifndef KRATOS_REGISTER_PROCESS
+#define KRATOS_REGISTER_PROCESS(                                                                             \
```
pooyan-dadvand: I think we should call this KRATOS_REGISTER_PROCESS_WITH_PROTOTYPE to keep open the door for eventually having processes registered with create function (no standard constructor) 

And the name helps to identify the arguments

Disclamer: I remember from the discussions that we want to register only the standard ones but you never know the future.

pooyan-dadvand: I would create a template method:

``` c++
template<typename PrototypeType> 
RegisterProcessWithPrototype(std::string const& ModuleName, std::string const& ProcessName, PrototypeType ProtoType)
´´´

File: kratos/processes/process.h [hunk 2]
```diff
@@ -253,6 +254,26 @@ class Process : public Flags
...
+            r_process_item.AddItem<RegistryValueItem<Process>>("Prototype", process_prototype);              \
+        } else {                                                                                             \
+            KRATOS_ERROR << "Process '" << process_name << "' is already registered." << std::endl;          \
+        }                                                                                                    \
+        std::string module_path = std::string("Processes.") + module_name + std::string(".") + process_name; \
+        if (!Registry::HasItem(module_path)) {                                                               \
+            auto& r_process_item = Registry::AddItem<RegistryItem>(module_path);                             \
+            r_process_item.AddItem<RegistryValueItem<Process>>("Prototype", process_prototype);              \
+        }                                                                                                    \
+    }
```
pooyan-dadvand: We can return the reference to the Item. Then we can register it automatically in each class by adding the following line to the macro:

`static const auto& mrRegistryItem = RegisterProcessWithPrototype(module_name, process_name, process_prototype)`

Then we don't need the manual registering.

kratos/python_interface/python_registry.py
```diff
@@ -0,0 +1,370 @@
...
+        py_values = None
+        if self.HasItems(Name, RegistryContext.PYTHON):
+            py_values = self.__InternalGetPythonItem(Name).values()
+
+        # Return the c++ and Python registry values in a tuple
+        # Note that we return None if the item was not present in one of the registries
+        return py_values, cpp_values
+
+    def items(self, Name):
+        # Check if Name registry level is registered and if it is a value
```
pooyan-dadvand, line 117: It should be value?

kratos/python_interface/python_registry.py
```diff
@@ -0,0 +1,370 @@
...
+    def NumberOfPythonItems(self, Name):
+        '''Returns the number of Python registered items
+        '''
+        py_keys, _ = self.keys(Name)
+        total_items = 0
+        if py_keys is not None:
+            total_items += len(py_keys)
+        return total_items
+
+    def NumberOfCppItems(self, Name):
```
pooyan-dadvand: Todo: We may add this method in CPP to be more efficient. (In a separate PR)

kratos/python_scripts/model_parameters_factory.py
```diff
@@ -0,0 +1,88 @@
...
+                        create = KM.Registry[f"{registry_entry}.CreateFunction"]
+                        instance = create(self.model, parameters)
+                    # Get prototype from stored Python module
+                    elif KM.Registry.HasItem(f"{registry_entry}.ModuleName"):
+                        class_name = registry_entry.split(".")[-1]
+                        module_name = KM.Registry[f"{registry_entry}.ModuleName"]
+                        module = import_module(module_name)
+                        # Check if the standard class name is available
+                        # Note that in here we assume that the registry last key matches the class name
+                        if hasattr(module, class_name):
```
pooyan-dadvand, line 62: Are you sure that if shouldn't be first if has ClassName?

I mean if we want to register two classes in the same module in this way it won't work...

kratos/python_scripts/process_factory.py
```diff
@@ -13,33 +13,64 @@ def ConstructListOfProcesses( self, process_list ):
...
+                registry_entry = item["name"].GetString()
+                if KM.Registry.HasItem(registry_entry):
+                    # Get already stored prototype
+                    if KM.Registry.HasItem(f"{registry_entry}.Prototype"):
+                        prototype = KM.Registry[f"{registry_entry}.Prototype"]
+                    # Get prototype from stored Python module
+                    elif KM.Registry.HasItem(f"{registry_entry}.ModuleName"):
+                        class_name = registry_entry.split(".")[-1]
+                        module_name = KM.Registry[f"{registry_entry}.ModuleName"]
+                        module = import_module(module_name)
```
pooyan-dadvand, line 28: shall we check before import?

pooyan-dadvand (commented):
Very nice work!! Some minor comments plus an automatic way of registering in CPP

philbucher (changes requested):
Awesome work!

I have a few comments, will post them in the next days.

Out of curiosity, why the py version bump to 3.8? You have my full support ;)

rubenzorrilla:
> Awesome work!
> 
> I have a few comments, will post them in the next days.
> 
> Out of curiosity, why the py version bump to 3.8? You have my full support ;)

OK. But please note that we have a very tight deadline. We'll need this to be merged by the end of this week.

kratos/python_interface/python_registry.py
rubenzorrilla: I think the comment is correct. We first check if `Name` is registered and then if it is a value item of a subitem. If the former we throw an error.

In any case, thanks to this observation I found a copy & paste error in the error thrown messages below.

kratos/python_interface/python_registry.py
rubenzorrilla: Agree. I'll add a TODO comment as reminder.

kratos/python_scripts/model_parameters_factory.py
rubenzorrilla: mmm... I think I'm not getting your point at all.

For the registry entry
```
"Processes.KratosMultiphysics.FluidDynamicsApplication.ApplyInletProcess" : {
    "ModuleName" : "KratosMultiphysics.FluidDynamicsApplication.apply_inlet_process"
}
```
`module_name` would be `"KratosMultiphysics.FluidDynamicsApplication.apply_inlet_process"`and module is the result of importing that. Then we retrieve the `class_name` from the registry entry last word, which in this case is `ApplyInletProcess`. Alternatively, we give the option of providing the `ClassName` together with the `ModuleName`.

Having said so, I think that this should work provided that there is no more than one class with the same name, which is something that according to my understanding we've always assumed in Kratos.

rubenzorrilla: Adding to my comment that I've just checked (again) that this mechanism works with that very same `ApplyInletProcess` and a CFD simulation.

kratos/python_scripts/process_factory.py
rubenzorrilla: I'm not getting your point at all...

If you mean checking that current item has `name` (i.e. is to be constructed from the registry) it is done in line 18. If you mean checking that provided `name` is registered, it is done in line 20.

Alternatively, if you mean a try-except in the module import, I think the `import_module` handles this.

kratos/processes/process.h
rubenzorrilla: I agree with both observations. Indeed, I'll create a `registry_utilities.h` so we can collect all these functions in one place.

kratos/processes/process.h
rubenzorrilla: About this, I'm not getting your point at all about "manual registering"... You mean adding this line to the pointer definition macro or to the already existing one? Because the latter would still require to call it somewhere.

rubenzorrilla: You mean something of this sort 
```c++
template<typename TPrototypeType>
void RegisterProcessWithPrototype(
    const std::string RegistryEntryName,
    TPrototypeType rProcessPrototype)
{
    if (!Registry::HasItem(RegistryEntryName)) {
        auto& r_process_item = Registry::AddItem<RegistryItem>(RegistryEntryName);
        r_process_item.AddItem<RegistryValueItem<TPrototypeType>>("Prototype", rProcessPrototype);
    } else {
        KRATOS_ERROR << "Process '" << RegistryEntryName << "' is already registered." << std::endl;
    }
}

#ifndef KRATOS_REGISTER_PROCESS_WITH_PROTOTYPE
#define KRATOS_REGISTER_PROCESS_WITH_PROTOTYPE(                                                              \
    module_name,                                                                                             \
    process_name,                                                                                            \
    process_prototype)                                                                                       \
    {                                                                                                        \
        std::string all_path = std::string("Processes.All.") + process_name;                                 \
        RegisterProcessWithPrototype(all_path, process_prototype);                                           \
        std::string module_path = std::string("Processes.") + module_name + std::string(".") + process_name; \
        RegisterProcessWithPrototype(module_path, process_prototype);                                        \
    }
#endif
```

CMakeLists.txt
```diff
@@ -616,6 +616,7 @@ install(CODE "file(REMOVE_RECURSE \"${CMAKE_INSTALL_PREFIX}/libs\")")
 # Install core files for the KratosMultiphysics python module
 kratos_python_install(${INSTALL_PYTHON_USING_LINKS} "${KRATOS_SOURCE_DIR}/kratos/python_interface/__init__.py" "KratosMultiphysics/__init__.py" )
 kratos_python_install(${INSTALL_PYTHON_USING_LINKS} "${KRATOS_SOURCE_DIR}/kratos/python_interface/kratos_globals.py" "KratosMultiphysics/kratos_globals.py" )
+kratos_python_install(${INSTALL_PYTHON_USING_LINKS} "${KRATOS_SOURCE_DIR}/kratos/python_interface/python_registry.py" "KratosMultiphysics/python_registry.py" )
```
philbucher, line 619: please add this to `python_scripts` instead such that this magic is not required.

applications/FluidDynamicsApplication/FluidDynamicsApplication.py
```diff
@@ -1,7 +1,16 @@
...
 application_name = "KratosFluidDynamicsApplication"
 
 _ImportApplication(application, application_name)
+
+#TODO: We should add a RegistryProcess function to the python_registry.py to put the keyword in front (we don't want the user to do this)
+KratosMultiphysics.Registry.AddItem(
+    "Processes.KratosMultiphysics.FluidDynamicsApplication.ApplyInletProcess",
+    {
+        "ClassName" : "ApplyInletProcess",
+        "ModuleName" : "KratosMultiphysics.FluidDynamicsApplication.apply_inlet_process"
```
philbucher, line 14:
```suggestion
        "class_name" : "ApplyInletProcess",
        "module_name" : "KratosMultiphysics.FluidDynamicsApplication.apply_inlet_process"
```

please follow our naming convention

applications/FluidDynamicsApplication/FluidDynamicsApplication.py
```diff
@@ -1,7 +1,16 @@
 from KratosFluidDynamicsApplication import *
 
+import KratosMultiphysics
 from KratosMultiphysics import _ImportApplication
```
philbucher:
```suggestion
from KratosMultiphysics import Registry, _ImportApplication
```

kratos/processes/register_operations.cpp
```diff
@@ -0,0 +1,35 @@
...
+//
+
+// System includes
+#include <string>
+#include <iostream>
+
+// External includes
+
+// Project includes
+#include "includes/kernel.h"
```
philbucher:
```suggestion
#include "includes/kratos_application.h"
```

kratos/processes/register_processes.cpp
```diff
@@ -0,0 +1,36 @@
...
+//
+
+// System includes
+#include <string>
+#include <iostream>
+
+// External includes
+
+// Project includes
+#include "includes/kernel.h"
```
philbucher:
```suggestion
#include "includes/kratos_application.h"
```

kratos/python/add_registry_to_python.cpp
```diff
@@ -0,0 +1,79 @@
...
+    namespace py = pybind11;
+
+    py::class_<RegistryItem, RegistryItem::Pointer>(m, "RegistryItem")
+        .def("Name", &RegistryItem::Name)
+        .def("HasItems", &RegistryItem::HasItems)
+        .def("HasValue", &RegistryItem::HasValue)
+        .def("keys", &registry_item_keys)
+        .def("size", &RegistryItem::size)
+        .def("__iter__", [](RegistryItem& rSelf){return py::make_iterator(rSelf.key_begin(), rSelf.key_end());}, py::keep_alive<0,1>())
+        // .def("__str__", [](const RegistryItem& rSelf){rSelf.Name();})
```
philbucher:
```suggestion
        .def("__str__", PrintObject<RegistryItem>)
```

kratos/python/add_registry_to_python.h
```diff
@@ -0,0 +1,29 @@
...
+//
+//  Main authors:    Suneth Warnakulasuriya
+//                   Carlos Roig
+//                   Ruben Zorrilla
+//
+
+#pragma once
+
+// System includes
+#include <pybind11/pybind11.h>
```
philbucher, line 18:
```suggestion
#include "includes/define_python.h"
```

kratos/python_interface/__init__.py
```diff
@@ -1,10 +1,10 @@
 import os
-import re
 import sys
 from . import kratos_globals
+from . import python_registry
 
-if sys.version_info < (3, 6):
-    raise Exception("Kratos only supports Python version 3.6 and above")
+if sys.version_info < (3, 8):
+    raise Exception("Kratos only supports Python version 3.8 and above")
```
philbucher: Nice!

This check also exists in CMake, please update it too

kratos/python_interface/python_registry.py
```diff
@@ -0,0 +1,370 @@
+import KratosMultiphysics
+from KratosMultiphysics.kratos_utilities import GetListOfAvailableApplications
+
+class RegistryContext():
+    ALL = 0
+    CPP = 1
+    PYTHON = 2
```
philbucher: please use an `Enum` here

kratos/python_interface/python_registry.py
philbucher: I think what he means is that the naming is confusing with `dict`

kratos/python_interface/python_registry.py
```diff
@@ -0,0 +1,370 @@
+import KratosMultiphysics
```
philbucher, line 3: please move this file to `python_scripts`

philbucher (commented):
given the time constraint I only did a brief review addressing my main concerns.

Overall I really like the changes and this is done well and goes into a good direction :tada: 

Still my **personal** opinion is to wait (at least) a week with merging such big and important changes. While I fully understand that deadlines are pressing, going too fast might lead to non-optimal solutions that will haunt us in the future. Especially for components like this which are decisive for the future of Kratos. 
Also consider that you are missing out on the help from the community! I think not few ppl are interested in this and would give their helpful comments and suggestions, but they need some time (e.g. afaik @loumalouomega is on vacation this week)
I am writing this as we made similar mistakes in the past :/

kratos/python_interface/python_registry.py
```diff
@@ -0,0 +1,372 @@
...
+        cpp_items = None
+        if cpp_keys is not None:
+            cpp_items = []
+            for cpp_key, cpp_value in zip(cpp_keys, cpp_values):
+                cpp_items.append((cpp_key, cpp_value))
+
+        # Return the c++ and Python registry items in a tuple
+        return py_items, cpp_items
+
+    def NumberOfItems(self, Name):
```
roigcarlo: I would use `Context=RegistryContext.XXX` instead of having three exposed functions for the user and leave `NumberOfPythonItems` and `NumberOfCppItems` as interal.

kratos/python/add_registry_to_python.h
roigcarlo: Why we need the full define_python.h here? We are not using custom holders or register functions

kratos/includes/registry_item.h
```diff
@@ -66,6 +67,131 @@ class RegistryItem
...
+    public:
+        ///@name Type Definitions
+        ///@{
+
+        using BaseIterator      = SubRegistryItemType::iterator;
+        using iterator_category = BaseIterator::iterator_category;
+        using difference_type   = BaseIterator::difference_type;
+        using value_type        = BaseIterator::value_type;
+        using pointer           = BaseIterator::pointer;
+        using reference         = BaseIterator::reference;
```
matekelemen: ```value_type``` is the type the iterator is iterating over. In this case, it's an ```std::string``` key, which is definitely not the same as the ```map```'s iterator. Same goes for ```pointer``` and ```reference```.
```diff
-        using value_type        = BaseIterator::value_type;
-        using pointer           = BaseIterator::pointer;
-        using reference         = BaseIterator::reference;
+        using value_type        = std::string;
+        using pointer           = const value_type*;
+        using reference         = const value_type&;
```

kratos/includes/registry_item.h
```diff
@@ -66,6 +67,131 @@ class RegistryItem
...
+    using SubRegistryItemType = std::unordered_map<std::string, Kratos::shared_ptr<RegistryItem>>;
+
+    /// Custom iterator with key as return type to be used in the Python export
+    class key_return_iterator
+    {
+    public:
+        ///@name Type Definitions
+        ///@{
+
+        using BaseIterator      = SubRegistryItemType::iterator;
```
matekelemen: As far as I can tell, ```key_return_iterator``` is not supposed to mutate the underlying keys, so ```BaseIterator``` should reflect that:
```suggestion
        using BaseIterator      = SubRegistryItemType::const_iterator;
```

kratos/includes/registry_item.h
```diff
@@ -66,6 +67,131 @@ class RegistryItem
     /// Pointer definition of RegistryItem
     KRATOS_CLASS_POINTER_DEFINITION(RegistryItem);
 
+    /// Subregistry item type definition
+    using SubRegistryItemType = std::unordered_map<std::string, Kratos::shared_ptr<RegistryItem>>;
+
+    /// Custom iterator with key as return type to be used in the Python export
+    class key_return_iterator
```
matekelemen: Iterating over the keys of a map is a fairly common task, so I think this iterator should be templated and moved to a separate header file. By the way I already happen to have an [implementation](https://github.com/KratosMultiphysics/Kratos/blob/749ba528d6976a0703d91192f9ee54d2eeab7184/kratos/utilities/ranges.h#L95) in a pending PR.

kratos/includes/registry_item.h
```diff
@@ -66,6 +67,131 @@ class RegistryItem
...
+        ///@name Operators
+        ///@{
+
+        key_return_iterator& operator=(const key_return_iterator& rIterator)
+        {
+            this->mIterator = rIterator.mIterator;
+            return *this;
+        }
+
+        std::string operator*() const
```
matekelemen:
```suggestion
        reference operator*() const
```

kratos/includes/registry_item.h
```diff
@@ -66,6 +67,131 @@ class RegistryItem
...
+        {
+            return mIterator->first;
+        }
+
+        std::string operator->()
+        {
+            return mIterator->first;
+        }
```
matekelemen:
```suggestion
        pointer operator->() const
        {
            return &mIterator->first;
        }
```

kratos/includes/registry_item.h
```diff
@@ -66,6 +67,131 @@ class RegistryItem
...
+
+    /// Custom iterator with key as return type to be used in the Python export
+    class key_return_iterator
+    {
+    public:
+        ///@name Type Definitions
+        ///@{
+
+        using BaseIterator      = SubRegistryItemType::iterator;
+        using iterator_category = BaseIterator::iterator_category;
```
matekelemen: ```std::map```'s iterator is ~~random access~~ **bi-directional**, but ```key_return_iterator``` only implements increment operators.
```suggestion
        using iterator_category = std::forward_iterator_tag;
```

kratos/includes/registry_item.h
```diff
@@ -167,7 +317,7 @@ class RegistryItem
 
     std::string mName;
     const void* mpValue;
```
matekelemen, line 318: I don't like this. It's basically raping the C++ type system, which is one of the most basic abstractions the entire language is built on.

kratos/includes/registry_item.h
```diff
@@ -118,12 +256,24 @@ class RegistryItem
 
     RegistryItem& GetItem(std::string const& rItemName);
 
+    template<typename TDataType> TDataType const& GetValue() const
+    {
+        KRATOS_ERROR_IF(mpValue == nullptr) << "Item " << Name() << " does not have value to be returned." << std::endl;
+        return *static_cast<const TDataType*>(mpValue);
```
matekelemen, line 261: Again, I don't like the ```void*``` + ```static_cast``` combo and don't think this does anything that isn't already possible within the C++ type system.

kratos/includes/registry_item.h
```diff
@@ -97,18 +223,30 @@ class RegistryItem
         std::string const& ItemName,
         TArgumentsList&&... Arguments)
     {
-        KRATOS_ERROR_IF(this->HasItem(ItemName)) << "The RegistryItem " << this->Name() << " already has an item with name " << ItemName << std::endl;
+        KRATOS_ERROR_IF(this->HasItem(ItemName)) << "The RegistryItem '" << this->Name() << "' already has an item with name " << ItemName << "." << std::endl;
+        KRATOS_ERROR_IF(this->HasValue()) <<
+            "Trying to add '"<< ItemName << "' item to the RegistryItem '" << this->Name() << "' but this already has value. Items cannot have both value and subitem." << std::endl;
         auto insert_result = mSubRegistryItem.emplace(std::make_pair(ItemName, Kratos::make_unique<TItemType>(ItemName, std::forward<TArgumentsList>(Arguments)...)));
-        KRATOS_ERROR_IF_NOT(insert_result.second) << "Error in inserting " << ItemName << " in registry item with name " << this->Name() << std::endl;
+        KRATOS_ERROR_IF_NOT(insert_result.second) << "Error in inserting '" << ItemName << "' in registry item with name '" << this->Name() << "'." << std::endl;
```
matekelemen: This check is unnecessary; it checks the exact same thing as ```HasValue```.

kratos/includes/registry_item.h
```diff
@@ -97,18 +223,30 @@ class RegistryItem
...
+
+    SubRegistryItemType::const_iterator cbegin() const;
+
+    SubRegistryItemType::iterator end();
+
+    SubRegistryItemType::const_iterator cend() const;
+
+    key_return_iterator key_begin();
+
+    key_return_iterator key_end();
```
matekelemen: ```key_return_iterator```, ```key_begin```, and ```key_end``` are not standard C++ names, so they should follow the Kratos naming conventions. Other than that, ```key_return_iterator``` never mutates the ```map``` it iterates on so these functions can be ```const```.
```suggestion
    KeyReturnIterator KeyBegin() const;

    KeyReturnIterator KeyEnd() const;
```

kratos/python/add_registry_to_python.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include "includes/registry_value_item.h"
+#include "operations/operation.h"
+#include "processes/process.h"
+
+namespace Kratos::Python
+{
+
+namespace
+{
+    pybind11::list registry_item_keys(const RegistryItem& rSelf)
```
matekelemen: Naming.
```suggestion
    pybind11::list GetRegistryItemKeys(const RegistryItem& rSelf)
```

kratos/python/add_registry_to_python.cpp
```diff
@@ -0,0 +1,79 @@
...
+    {
+        pybind11::list t;
+        KRATOS_ERROR_IF(!rSelf.HasItems()) << "Asking for the keys of " << rSelf.Name() << "which has no subitems." << std::endl;
+        for (auto it = rSelf.cbegin(); it != rSelf.cend(); ++it) {
+            t.append(it->first);
+        }
+        return t;
+    }
+
+    pybind11::list registry_keys()
```
matekelemen: Naming.
```suggestion
    pybind11::list GetRegistryKeys()
```

kratos/python/add_registry_to_python.cpp
```diff
@@ -0,0 +1,79 @@
...
+{
+    namespace py = pybind11;
+
+    py::class_<RegistryItem, RegistryItem::Pointer>(m, "RegistryItem")
+        .def("Name", &RegistryItem::Name)
+        .def("HasItems", &RegistryItem::HasItems)
+        .def("HasValue", &RegistryItem::HasValue)
+        .def("keys", &registry_item_keys)
+        .def("size", &RegistryItem::size)
+        .def("__iter__", [](RegistryItem& rSelf){return py::make_iterator(rSelf.key_begin(), rSelf.key_end());}, py::keep_alive<0,1>())
```
matekelemen: ```RegistryItem```s are static, right? If they are, ```py::keep_alive<0,1>()``` is unnecessary.

kratos/python_interface/__init__.py
```diff
@@ -77,6 +83,15 @@ def CustomExceptionHook(exc_type, exc_value, exc_traceback):
...
 KratosGlobals = __ModuleInitDetail()
 
+# Create the Python global registry
+# Note that this interfaces the c++ registry.
+Registry = python_registry.PythonRegistry()
+RegisterPrototype = python_registry.RegisterPrototype
+
+# Remove CppRegistry from locals() in order to give preference to the exception thrown in __getattr__
+# This is required since we cannot use properties as usual due to the fact that we have no instance of CppRegistry (it is a static variable in c++)
+locals().pop("CppRegistry")
```
matekelemen, line 93: Um, what? Can someone explain this to me in more detail?

kratos/python_interface/python_registry.py
```diff
@@ -0,0 +1,372 @@
...
+
+    #TODO: Change this by a frozenmap() once this is available
+    CppGetFunctionNamesMap = {
+        "Operations": "GetOperation",
+        "Processes": "GetProcess"
+    }
+
+    def __init__(self):
+        # We get a "private" pointer to the c++ registry
+        self.__cpp_registry = KratosMultiphysics.CppRegistry
```
matekelemen, line 48: I'm confused. ```KratosMultiphysics.CppRegistry``` is a class, isn't it?. Furthermore, it's never supposed to be instantiated, so I don't get the comment about the *private pointer*.

kratos/python_interface/python_registry.py
```diff
@@ -0,0 +1,372 @@
...
+
+    @staticmethod
+    def __GetCppRegistryGetFunctionNameFromItemKeyword(ItemKeyword):
+        if ItemKeyword in PythonRegistry.CppGetFunctionNamesMap:
+            return PythonRegistry.CppGetFunctionNamesMap[ItemKeyword]
+        else:
+            err_msg = f"'{ItemKeyword}' has not been addded to 'CppGetFunctionNamesMap'."
+            raise Exception(err_msg)
+
+    def __InternalHasPythonItem(self, Name):
```
matekelemen: The leading ```__``` already implies that the function is *"private"* and mangles the function name outside the class' scope, so no need for the ```Internal``` tag. This goes for all other function names as well.

applications/FluidDynamicsApplication/FluidDynamicsApplication.py
rubenzorrilla: The point is that we are registering modules using CamelCase (as you'd import them in Python). Then if we were using snake_case for the other items we'd end up with something of this sort `processes.KratosMultiphysics.FluidDynamicsApplication.ApplyInletProcess.module_name`. We considered that it would be better to always use camel case in the keys, meaning that the one above is retrieved as `Processes.KratosMultiphysics.FluidDynamicsApplication.ApplyInletProcess.ModuleName`.

rubenzorrilla:
> Awesome work!
> 
> I have a few comments, will post them in the next days.
> 
> Out of curiosity, why the py version bump to 3.8? You have my full support ;)

We need it to override the `__getattr__` at the module level so we can throw an exception if tries to directly access the `CppRegistry` from Python.

loumalouomega:
> > given the time constraint I only did a brief review addressing my main concerns.
> > 
> > Overall I really like the changes and this is done well and goes into a good direction tada
> > 
> > Still my **personal** opinion is to wait (at least) a week with merging such big and important changes. While I fully understand that deadlines are pressing, going too fast might lead to non-optimal solutions that will haunt us in the future. Especially for components like this which are decisive for the future of Kratos. Also consider that you are missing out on the help from the community! I think not few ppl are interested in this and would give their helpful comments and suggestions, but they need some time (e.g. afaik @loumalouomega is on vacation this week) I am writing this as we made similar mistakes in the past :/
> 
> I completely agree. Nonetheless, this is a new feature, meaning that future changes will be not as painful as changing production-ready stuff. Indeed, chances are that as soon as we start populating this minor things will pop up. On top of that, I'd like to make clear that this is not a personal deadline but a Kratos one affecting @KratosMultiphysics/all , as from the @KratosMultiphysics/technical-committee we want this to be ready for the workshop.

I will try to take a look tomorrow

applications/FluidDynamicsApplication/FluidDynamicsApplication.py
rubenzorrilla: Done.

kratos/python/add_registry_to_python.cpp
rubenzorrilla: Done.

kratos/processes/register_processes.cpp
rubenzorrilla: Done.

kratos/processes/register_operations.cpp
rubenzorrilla: Done.

kratos/python_interface/__init__.py
rubenzorrilla: Done. Thanks for the reminder!

roigcarlo:
@matekelemen I am not sure to follow your last comment:

```C++
 /// Constructor with the name
    RegistryItem(std::string Name) : mName(Name), mpValue(nullptr){}
```
That's not a `RegistryValueItem`. `RegistryItem` do not get deleted in the destructor because it does not hold a copy of the object.

While I would prefer the registry to be strongly typed as well, I don't see any way to store different copies of types without using `void*`, `std::any` or `std::optional` which at the end are just pretty `void*`

kratos/python_interface/python_registry.py
rubenzorrilla: Done.

kratos/python_interface/python_registry.py
rubenzorrilla: We discussed about this at the very beginning. Considering that this is quite internal and that it can be understood as an extension of the c++ registry we (the KTC) decided to place it in here (OFC at the cost of "manually" installing it as you pointed above). Personally, I'm still of the opinion of keeping it like this. However if it is a blocker for you we can discuss it further (@KratosMultiphysics/technical-committee).

kratos/includes/registry_item.h
rubenzorrilla: This was introduced in #9308. Unfortunately, as far as I know, this is the unique way we have to do a type-agnostic registry.

kratos/includes/registry_item.h
rubenzorrilla: Considering that this is a direct consequence of the `void*`, I'd resolve this to continue the discussion in your comment above.

kratos/includes/registry_item.h
rubenzorrilla: Agree. Having said this, I'd merge this one (note that it is pretty internal) to then use (if possible) the final common one you implemented once it is merged. In this regard, I'd say that after including your suggestions, it is not as hacky as my very first version.

kratos/python/add_registry_to_python.cpp
rubenzorrilla: Done.

kratos/python/add_registry_to_python.cpp
rubenzorrilla: Done.

kratos/includes/registry_item.h
rubenzorrilla: Fair enough. About the class, I'd call it `KeyReturnConstIterator`. Concerning the method names, I suggest calling them  `KeyConstBegin` and `KeyConstEnd`. I think this makes clear that we're working with a const iterator.

kratos/includes/registry_item.h
rubenzorrilla: Agree and done. Thanks for the comment.

kratos/includes/registry_item.h
rubenzorrilla: Done. Changed to `const_pointer`. `const` specifier also included.

kratos/includes/registry_item.h
rubenzorrilla: Done. Changed to `const_reference`.

kratos/includes/registry_item.h
rubenzorrilla: Agree and done. Thanks for the observation.

kratos/includes/registry_item.h
rubenzorrilla: Agree. Previous version was assuming that the key type would always be `std::string`. Although this is always the case in the particular case, now I'm using as `value_type` the `SubRegistryItemType::key_type`. Reference and pointer types have been updated accordingly. Note that now these are `const_reference` and `const_pointer`.

matekelemen:
```RegistryValueItem``` is derived from ```RegistryItem```, so the constructor of ```RegistryItem``` is definitely invoked when a ```RegistryValueItem``` is instantiated.

The problem is, one of ```RegistryValueItem```'s constructor results in ```mpValue``` being ```nullptr```.

File: kratos/processes/process.h (hunk 2 above)
Comment (pooyan-dadvand) [reply kratos/processes/process.h]: Yes in this line! Something like:

```c++
template<typename TPrototypeType>
void RegisterProcessWithPrototype(
    const std::string RegistryEntryName,
    TPrototypeType rProcessPrototype)
{
    if (!Registry::HasItem(RegistryEntryName)) {
        auto& r_process_item = Registry::AddItem<RegistryItem>(RegistryEntryName);
        r_process_item.AddItem<RegistryValueItem<TPrototypeType>>("Prototype", rProcessPrototype);
    } else {
        KRATOS_ERROR << "Process '" << RegistryEntryName << "' is already registered." << std::endl;
    }
}

template<typename TPrototypeType>
void RegisterProcessWithPrototype(
    const std::string ModuleName
    const std::string RegistryEntryName,
    TPrototypeType rProcessPrototype)
{
        std::string all_path = std::string("Processes.All.") + process_name;
        RegisterProcessWithPrototype(all_path, process_prototype);
        std::string module_path = std::string("Processes.") + module_name + std::string(".") + process_name;
        RegisterProcessWithPrototype(module_path, process_prototype);    
}

#ifndef KRATOS_PROCESS_DEFAULT_BODY
#define KRATOS_PROCESS_DEFAULT_BODY(                                                               \
    module_name,                                                                                   \
    process)                                                                                       \
    purblic:                                                                                       \
    virtual Process::Pointer Create(Model& rThisModel, Parameters TheParameters) {                 \
        return make_shared<process>(rThisModel, TheParameters);                                    \
    }                                                                                              \
    private:                                                                                       \
    static auto& mrRegistryItem = RegisterProcessWithPrototype(module_name, #process, process());   
    #endif


class AnyProcess{
    private:
    AnyProcess() // Private constructor that we may even add it to the same macro
    KRATOS_PROCESS_DEFAULT_BODY("any_module", AnyProcess)
}
```

kratos/python_scripts/process_factory.py
pooyan-dadvand: I am saying if we should check if the module is already imported. Indeed, very small detail....

kratos/python_scripts/model_parameters_factory.py
pooyan-dadvand: OK, I see. BTW, this can be revised easily afterward.

kratos/python_interface/python_registry.py
rubenzorrilla: Fair enough. Done in all cases.

matekelemen:
Another thing: why's Registry polymorphic? All its member variables are static, so derived classes will write into the same global registry structure. Is that intended?

kratos/includes/registry_item.h
rubenzorrilla: Not exactly. We added the `HasValue` check because, even though it's technically possible, we don't want items to have both value and subitem. The one you suggest to eliminate checks that the insertion (of either a value or a subitem) is correct.

kratos/python/add_registry_to_python.cpp
rubenzorrilla: mmmm... The main `Registry` is a static singleton, but I'm not so sure about how this is transferred to its `RegistryItem`(s). Considering that it doesn't harm, I'd keep it (at least for the moment).

kratos/includes/registry_item.h
matekelemen: My bad, I meant it's the same as the `HasItem` check. Both fail iff `mSubRegistryItem` already has an entry with `ItemName` key.

kratos/python_interface/python_registry.py
rubenzorrilla: This is a mechanism to have an alternative access to the `CppRegistry`. After doing this one could still access the `CppRegistry` from python by doing `KratosMultiphysics.CppRegistry.WatheverCppFunction()`. This is something we want to be explicitly forbidden in order to force people to access the c++ registry through the Python extension, implying that, the user doesn't even realize that there is a c++ layer and a Python extension one but a unique registry.

Besides, we do the `__getattr__` override at the module level in order to throw a meaningful exception if the user still calls `KratosMultiphysics.CppRegistry`.

Once we've saved this pointer, we do the `pop` in your comment above in order to explicitly remove the `KratosMultiphysics.CppRegistry` access so we only keep the `self__cpp_registry` inside the Python layer. We need to do this `pop` instead of using pybind properties as we do in normal classes because the `CppRegistry` is indeed a static global variable. If we were not doing this `pop` the exception would never be thrown.

I tried to be as clear as possible but maybe it's still a bit blurry. Maybe @roigcarlo can give a better explanation.

kratos/python_interface/__init__.py
rubenzorrilla: See my [comment below](https://github.com/KratosMultiphysics/Kratos/pull/10420#discussion_r1013950859). I hope this clarifies both points.

rubenzorrilla:
> Another thing: why's `Registry` polymorphic? All its member variables are static, so derived classes will write into the same global registry structure. Is that intended?

About this, I don't think there will be any derived class, so maybe it's easier to make it `final`.

kratos/python_interface/python_registry.py
matekelemen: Got it, thanks!

kratos/python_interface/python_registry.py
rubenzorrilla: Sorry guys, but I still don't gent your point at all... :confused: :facepalm:

kratos/python_interface/python_registry.py
rubenzorrilla: OK. I remembering wondering about your point when implementing it.

rubenzorrilla: Done.

kratos/python_interface/python_registry.py
rubenzorrilla: Resolving!

kratos/python_interface/python_registry.py
philbucher: that makes sense, hadn't considered that!

applications/FluidDynamicsApplication/FluidDynamicsApplication.py
philbucher: ok, up to xou

kratos/python/add_registry_to_python.h
philbucher: good point
I guess I would not include the ext lib stuff diretly, but it is used in this file

philbucher (dismissed):

kratos/processes/process.h
rubenzorrilla: After giving it a try I see two limitations:
- I'm not sure if it is technically possible as it implies using the default constructor before the class is completed.
- In case we have a template process, as far as I understand, we'll need to register each of the possible specializations. Hence, it is impossible to do it in the header.

rubenzorrilla: Finally I decided to create an `RegistryAuxiliary` static class. In here we place the functions above, which are to be called later on in the corresponding `register_processes.cpp` and `register_operations.cpp`.

kratos/includes/registry_item.h
rubenzorrilla: Ah! Agree. I think we can safely remove it.

rubenzorrilla:
I went through the vast majority of points raised during the review. From the @KratosMultiphysics/technical-committee we think that this is now in very good shape to do in. The missing comments can IMO be handled in upcoming PRs.

INSTALL.md
```diff
@@ -506,7 +506,7 @@ The result should be:
  . \  |   (   | |   (   |\__ \  
 _|\_\_|  \__,_|\__|\___/ ____/
            Multi-Physics 9.2."0"-4afb88094a-Release-ARM64
-           Compiled for GNU/Linux and Python3.6 with GCC-8.5
+           Compiled for GNU/Linux and Python3.8 with GCC-8.5
```
loumalouomega: jiji, super replacement

kratos/includes/registry.h
```diff
@@ -121,29 +122,60 @@ class Registry
...
+        return GetItem(rItemFullName).GetValue<TDataType>();
+    }
 
     static void RemoveItem(std::string const& ItemName);
 
     ///@}
     ///@name Inquiry
     ///@{
 
+    static std::size_t size();
```
loumalouomega, line 159: Should be Size?, for consistency with other classes. size is actually more STL

loumalouomega (dismissed):
Okay, my only concern is python 3.8 which is quite recent. My Oracle VM works with python 3.6

Question, not in this PR, but I assume that the idea is to remove KratosComponents, and therefore the strategies factories will be back adapted to the registry?

kratos/includes/registry.h
rubenzorrilla: I did it like this because at the end this is a sort of extension of an stl unordered map. I don't mind changing it. The one people likes the most will be fine for me.

rubenzorrilla:
Well, it is more than three years old right now, but considering that we moved to c++17 not long ago, this could be admittedly considered rather new... In any case, the only thing you need to make it work in your VM is to remove the `__getattr__` override in the `__init__.py` (at the price of not having the meaningful exception thrown).

Regarding the `KratosComponents`, hopefully in the long term yes. Indeed, there is #9318 which aims at this direction. Nice thing of this is, even though we're only registering prototypes, we can easily extend this design to store pointers to custom create functions, which is what we'll need for the strategies.

matekelemen:
Python 3.6 reached EoL almost a year ago, so bumping to 3.8 is a good idea in my opinion.

kratos/includes/registry_auxiliaries.h
```diff
@@ -0,0 +1,195 @@
...
+    ///@name Operators
+    ///@{
+
+
+    ///@}
+    ///@name Operations
+    ///@{
+
+    template<typename TPrototypeType>
+    static void RegisterProcessWithPrototype(
```
sunethwarna, line 87: Are we gonna allow `Process` also registered via Prototypes and Create methods?

sunethwarna (dismissed):

kratos/includes/registry_auxiliaries.h
pooyan-dadvand: We don't, but we name it just in case.

pooyan-dadvand:
About the python dependency, we shall add (in a future PR) a backward compatibility for python < 3.8 taking out the getattribute.

rubenzorrilla:
> `RegistryValueItem` is derived from `RegistryItem`, so the constructor of `RegistryItem` is definitely invoked when a `RegistryValueItem` is instantiated.
> 
> The problem is, one of `RegistryValueItem`'s constructor results in `mpValue` being `nullptr`.
> 

About this, it should be as easy as removing the name-only constructor from the `RegistryValueItem` and deleting the default one (which is already done) right? Indeed, I don't see a situation in which we may need to create an empty `RegistryValueItem` so I think that, at least for the moment, we can safely remove it. Any comment @pooyan-dadvand ?

rubenzorrilla:
> Python 3.6 reached EoL almost a year ago, so bumping to 3.8 is a good idea in my opinion (although it probably should have been done in a separate PR).

Fair enough... Let me port these specific changes to an independent PR.

rubenzorrilla:
> > `RegistryValueItem` is derived from `RegistryItem`, so the constructor of `RegistryItem` is definitely invoked when a `RegistryValueItem` is instantiated.
> > The problem is, one of `RegistryValueItem`'s constructor results in `mpValue` being `nullptr`.
> 
> About this, it should be as easy as removing the name-only constructor from the `RegistryValueItem` and deleting the default one (which is already done) right? Indeed, I don't see a situation in which we may need to create an empty `RegistryValueItem` so I think that, at least for the moment, we can safely remove it. Any comment @pooyan-dadvand ?

Indeed, after having briefly discussed about with @pooyan-dadvand we both think that this could have sneaked in when copying and pasting `registry_item.h` class to create the `registry_value_item.h` one. I'll delete it right away. Nice catch @matekelemen.

I'll take the chance to add the retrocompatibility in the `__getattr__`.

roigcarlo (approved):
It should be ready then as a base implementation. We can improve on this bit by bit.

PR #10421, opened 2022-11-03
Avoid calling base kernel registry in apps

roigcarlo:
**📝 Description**
Should fix import errors from #10420.

I suppose it also fixes a bunch of other problems with those three apps?

**🆕 Changelog**
- Prevent kernel from being registered multiple times in selected apps.


philbucher (approved):
I thought I removed that years ago 🤔

pooyan-dadvand (approved):
👍

PR #10422, opened 2022-11-03
[PfemFluid] MuIrheology and FrictionalViscoplastic models with nodal integration

AFranci:
This PR introduces the Mu(I)-rheology and Frictional-Viscoplastic models for the nodal integration algorithm with nodal integration.

Besides this, a general cleaning of the nodal integration builder and solvers is carried out. The unused Block B&S has been removed and nodal-FSI B&Ss have been derived from the general B&S, avoiding duplication of the code.

The PR also removes mu(i)-rheology constitutive models that were used only for academic purposes. A unique constitutive model is maintained, the PapanastasiouMuIrheology which is now simply called MuIrheology.


AlejandroCornejo (dismissed):

rlrangel (dismissed):
Ok

AlejandroCornejo (approved):

PR #10430, opened 2022-11-07
[Core] Using Distance instead of manual calculation

loumalouomega:
**📝 Description**

Using Distance instead of manual calculation

**🆕 Changelog**

- Using Distance instead of manual calculation

### DIFF
--- a/kratos/utilities/exact_mortar_segmentation_utility.h
+++ b/kratos/utilities/exact_mortar_segmentation_utility.h
@@ -530,7 +530,7 @@ protected:
     {
 //         const double tolerance = std::numeric_limits<double>::epsilon(); // NOTE: Giving some problems, too tight
         const double tolerance = 1.0e-15;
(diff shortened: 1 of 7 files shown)


kratos/utilities/exact_mortar_segmentation_utility.h
```diff
@@ -530,7 +530,7 @@ class KRATOS_API(KRATOS_CORE) ExactMortarIntegrationUtility
     {
 //         const double tolerance = std::numeric_limits<double>::epsilon(); // NOTE: Giving some problems, too tight
         const double tolerance = 1.0e-15;
-        return (norm_2(rPointDest.Coordinates() - rPointOrig.Coordinates()) < tolerance) ? true : false;
+        return (rPointDest.Distance(rPointOrig) < tolerance) ? true : false;
```
philbucher:
```suggestion
        return rPointDest.Distance(rPointOrig) < tolerance;
```

philbucher (approved):

PR #10432, opened 2022-11-07
[Core] Multistage analysis

rubenzorrilla:
**📝 Description**
New multistage class.  #10431 should be merge first (this includes that leftover). Creating the PR to start the discussion.

PD: Note that this is not restart compatible yet. From the @KratosMultiphysics/technical-committee we decided to add it in a second stage.


loumalouomega (commented):
I would add more documentation to the different methods

loumalouomega:
> ```json
> execution_list
> ```

Just to say that in @KratosMultiphysics/altair we execute in order of declaration, so we don't define `execution_list`. Can we define the execution list automatically if not defined just taking the order of declaration?

rubenzorrilla:
> > ```json
> > execution_list
> > ```
> 
> Just to say that in @KratosMultiphysics/altair we execute in order of declaration, so we don't define `execution_list`. Can we define the execution list automatically if not defined just taking the order of declaration?

I don't see any problem on this. We do a `Has` and if not provided we create a fake one (e.g. for this case `["stage_1", "stage_2"]`).

loumalouomega:
> > > ```json
> > > execution_list
> > > ```
> > 
> > 
> > Just to say that in @KratosMultiphysics/altair we execute in order of declaration, so we don't define `execution_list`. Can we define the execution list automatically if not defined just taking the order of declaration?
> 
> I don't see any problem on this. We do a `Has` and if not provided we create a fake one (e.g. for this case `["stage_1", "stage_2"]`).

rubenzorrilla:
> > > > ```json
> > > > execution_list
> > > > ```
> > > 
> > > 
> > > Just to say that in @KratosMultiphysics/altair we execute in order of declaration, so we don't define `execution_list`. Can we define the execution list automatically if not defined just taking the order of declaration?
> > 
> > 
> > I don't see any problem on this. We do a `Has` and if not provided we create a fake one (e.g. for this case `["stage_1", "stage_2"]`).
> 
>  [ 68747470733a2f2f692e6b796d2d63646e2e636f6d2f70686f746f732f696d616765732f6e657773666565642f3030312f3432312f3739372f6635612e676966 ](https://camo.githubusercontent.com/b338e8219b0ea21bb021b768a1931f8435d2bfb0cbc97ddcc988e32beaafb6a0/68747470733a2f2f692e6b796d2d63646e2e636f6d2f70686f746f732f696d616765732f6e657773666565642f3030312f3432312f3739372f6635612e676966) [ ](https://camo.githubusercontent.com/b338e8219b0ea21bb021b768a1931f8435d2bfb0cbc97ddcc988e32beaafb6a0/68747470733a2f2f692e6b796d2d63646e2e636f6d2f70686f746f732f696d616765732f6e657773666565642f3030312f3432312f3739372f6635612e676966)

Done!

loumalouomega:
> > I would add more documentation to the different methods
> 
> Done. Although it is admittedly naive at the moment...

image

kratos/python_scripts/multistage_analysis.py
```diff
@@ -0,0 +1,181 @@
...
+        return stages_map
+
+    #TODO: Think about this
+    #TODO: I think we should be able to import a custom module in the current work directory
+    def __CreateAnalysisStageWithFlushInstance(self, cls, global_model, parameters):
+        '''This method emulates the old single-stage MainKratos.py.
+        Note that it exploits dynammic inheritance to extend the corresponding analysis stage class.
+        '''
+
+        class AnalysisStageWithFlush(cls):
```
philbucher: Can you please put this into a separate PR?
I think this should be integrated with the baseclass, I have somme comments/ideas (but lets not clutter this important PR) :)

rubenzorrilla: So you mean creating the stage as it is (without the flush) in this PR and then extending it?

rubenzorrilla: Removed :wink: .

ddiezrod:
Hi @rubenzorrilla I will give this a try today to check if it works with our structure and will let you know if there is anything we can improve.

ddiezrod:
@rubenzorrilla In order to make this work for us we will need to create a derived class from this one to add some more things that we need in our calculations (adding license and other stuff). To be able to do this I need ` __current_stage_name` and `__GetExecutionList` to not be private as I will need to call them from our derived class, is that ok?

rubenzorrilla:
> @rubenzorrilla In order to make this work for us we will need to create a derived class from this one to add some more things that we need in our calculations (adding license and other stuff). To be able to do this I need ` __current_stage_name` and `__GetExecutionList` to not be private as I will need to call them from our derived class, is that ok?

OK. My proposal would be:
1. Keeping the `__current_stage_name` as "private" but add a `GetCurrentStageName`.
2. Make the `__GetExecutionList` "protected" as `_GetExecutionList`.

Is this OK?

ddiezrod:
> > @rubenzorrilla In order to make this work for us we will need to create a derived class from this one to add some more things that we need in our calculations (adding license and other stuff). To be able to do this I need ` __current_stage_name` and `__GetExecutionList` to not be private as I will need to call them from our derived class, is that ok?
> 
> OK. My proposal would be:
> 
> 1. Keeping the `__current_stage_name` as "private" but add a `GetCurrentStageName`.
> 2. Make the `__GetExecutionList` "protected" as `_GetExecutionList`.
> 
> Is this OK?

I think I would also need a `SetCurrentStageName` so I can do a for loop in the stages.

rubenzorrilla:
> > > @rubenzorrilla In order to make this work for us we will need to create a derived class from this one to add some more things that we need in our calculations (adding license and other stuff). To be able to do this I need ` __current_stage_name` and `__GetExecutionList` to not be private as I will need to call them from our derived class, is that ok?
> > 
> > 
> > OK. My proposal would be:
> > 
> > 1. Keeping the `__current_stage_name` as "private" but add a `GetCurrentStageName`.
> > 2. Make the `__GetExecutionList` "protected" as `_GetExecutionList`.
> > 
> > Is this OK?
> 
> I think I would also need a `SetCurrentStageName` so I can do a for loop in the stages.

I've just realized that the get is already there. The set is intentionally hardcoded within the `Run` when the stages are looped. Which is the use case that requires looping the stages out of the `Run` method?

ddiezrod:
@rubenzorrilla Ok nevermind, I found that I can just override RunCurrentStagePreprocess to do our magic there so we do not actually need any change.

rubenzorrilla:
> @rubenzorrilla Ok nevermind, I found that I can just override RunCurrentStagePreprocess to do our magic there so we do not actually need any change.

Cool. I assume that this means that we have green light from your side.

kratos/python_scripts/multistage_analysis.py
```diff
@@ -0,0 +1,149 @@
+import importlib
+
+import KratosMultiphysics
+from KratosMultiphysics.model_parameters_factory import KratosModelParametersFactory
+
+class MultistageAnalysis(object):
```
sunethwarna:
```suggestion
class MultistageAnalysis:
```

It is anyways derrived from `Object` ryt?

kratos/python_scripts/multistage_analysis.py
```diff
@@ -0,0 +1,149 @@
...
+        if not hasattr(self, "__execution_list"):
+            # Default case in which the execution list is provided by the user
+            if self.settings.Has("execution_list"):
+                self.__execution_list = self.settings["execution_list"].GetStringArray()
+            # If not provided, create an auxiliary execution list from the stages declaration order
+            else:
+                KratosMultiphysics.Logger.PrintInfo("'execution_list' is not provided. Stages will be executed according to their declaration order.")
+                self.__execution_list = []
+                for stage_name in self.settings["stages"].keys():
+                    self.__execution_list.append(stage_name)
```
sunethwarna:
```suggestion
                self.__execution_list = list(self.settings["stages"].keys())
```

The keys insertion order is only guranteed for python >= 3.7. In the case of python 3.6, only CPython implementations gurantee it. Since we are moving to python 3.8, i think this is ok. Just wanted to mention if we are gonna stay in python3.6 for little bit longer.

https://stackoverflow.com/questions/5629023/order-of-keys-in-dictionaries-in-old-versions-of-python

kratos/python_scripts/multistage_analysis.py
```diff
@@ -0,0 +1,149 @@
...
+
+        factory = KratosModelParametersFactory(self.model)
+        return factory.ConstructListOfItems(execution_point_settings["modelers"])
+
+    def __GetOperations(self, execution_point):
+        '''This method creates the operations at any execution point.'''
+
+        if execution_point not in ["stage_preprocess","stage_postprocess"]:
+            err_msg = f"Wrong execution point '{execution_point}'. Supported ones are 'stage_preprocess' and 'stage_postprocess'."
+            raise Exception(err_msg)
```
sunethwarna, line 122: Is this error necessary? This is a private member method and we only call it specifically with those two strings. As far as i see, it won't be triggered ryt?

kratos/python_scripts/multistage_analysis.py
```diff
@@ -0,0 +1,149 @@
...
+        The values of the map are the stages instances, which are created by importing its corresponding analysis stage Python module.
+        '''
+
+        stages_map = {}
+        for stage_name in self.__GetExecutionList():
+            # Get the Python module name and class name implementing each analysis stage
+            # Note that we assume that the class name is the provided module (file) name in CamelCase
+            analysis_stage_module_name = self.settings["stages"][stage_name]["analysis_stage"].GetString()
+            analysis_stage_class_name = analysis_stage_module_name.split('.')[-1]
+            analysis_stage_class_name = ''.join(x.title() for x in analysis_stage_class_name.split('_'))
```
sunethwarna, line 140: Woudn't it be better to introduce the Factory methods to `AnalysisStage` derrived classes as we usually do in other classes since this is going to be used heavily also (when we want to simplify the MainKratos.py as well)

sunethwarna (commented):
Minor comments

kratos/python_scripts/multistage_analysis.py
rubenzorrilla: Agree. Makes no sense as we no longer support Python 2.

kratos/python_scripts/multistage_analysis.py
rubenzorrilla: As pointed by @ddiezrod, this is important for the @KratosMultiphysics/altair guys so another good reason for getting rid of the EOL Python 3.6.

kratos/python_scripts/multistage_analysis.py
rubenzorrilla: Actually this is not right now, but considering that this is an still under development feature I put it in there just in case. Note that this error is checked twice per stage, so its performance affectation is indeed (almost) nothing.

kratos/python_scripts/multistage_analysis.py
rubenzorrilla: So you mean having a `CreateMethod` in our current `MainKratos.py` files. Doesn't sound bad to me as these were unified thanks to the  `analysis_stage` field. However, I wouldn't make it mandatory just in case someone is currently using custom `MainKratos.py` files. 

Hence, we can do both options, first we check for a `Create` in the corresponding `analysis_stage` Python module. If not found, we try to create it like we are doing in here. Makes sense for you?

rubenzorrilla: @sunethwarna I added the `Create` option. However, we do it in the other way around, first we check if the "official" naming convention is followed. If not, we try to use the `Create`. If none of these work an error is thrown.

roigcarlo (approved):

loumalouomega:
Yay, this is ready ping @KratosMultiphysics/altair

PR #10433, opened 2022-11-07
[Core] Cpp registry enhancement

sunethwarna:
**📝 Description**
This PR removes the use of `void*`and uses `std::any`. I simplified the `RegistryItem` and `RegistryValueItem` classes to one `RegistryItem` class. All the tests are passing.

**🆕 Changelog**
- Updated `void*` to `std::any`
- Removed `RegistrValueItem`
- Removed storage of `void*` and `std::unordered_map<std::string, std::shared_ptr<RegistryItem>>` and uses one `std::any` for both.


rubenzorrilla:
This is a potential improvement of what we merged recently. However, at first glance I think there's a minor behavior change, right now you can only iterate the registry if current level holds a subitem. Maybe I'm missing something, but after this change, it seems to me that it can always be iterated. How is this translated to the Python export considering that the getter(s) are templated?

kratos/includes/registry_item.h
```diff
@@ -200,33 +144,51 @@ class KRATOS_API(KRATOS_CORE) RegistryItem
     RegistryItem() = delete;
 
     /// Constructor with the name
-    RegistryItem(std::string Name) : mName(Name), mpValue(nullptr){}
+    RegistryItem(const std::string& rName)
+        : mName(rName), mpValue(Kratos::make_shared<SubRegistryItemType>()) {}
```
matekelemen: I know `std::any` requires its stored type to be copy constructible, but I don't think that alone justifies using `std::shared_ptr` as a wrapper here. How about some wrapper class that has a copy constructor but throws an error if called (that would cut unnecessary heap allocations in half).

kratos/sources/registry_item.cpp
```diff
@@ -69,45 +74,105 @@ namespace Kratos
...
     }
 
-    std::string RegistryItem::ToJson(std::string const& rIndentation) const
+    std::string RegistryItem::GetValueString() const
     {
-        KRATOS_ERROR_IF(HasValue()) << "For storing a value you should use the RegistryValueItem" << std::endl;
+        return mpValue.type().name();
+    }
+
+    std::string RegistryItem::ToJson(std::string const& rTabSpacing, const std::size_t Level) const
```
matekelemen, line 91: I think it'd be easier to build a `Parameters` and let that take care of formatting. I know it's inefficient, but how often do we really want to print the registry?

kratos/includes/registry_item.h
rubenzorrilla: We've been discussing about this in the @KratosMultiphysics/technical-committee. @roigcarlo would change this to lambda functions to get rid of shared pointers in his upcoming upgrade.

rubenzorrilla: Adding to our previous comment that we would merge this as it is so @roigcarlo can build his developments on top of  @sunethwarna 's enhancements.

kratos/sources/registry_item.cpp
rubenzorrilla: @KratosMultiphysics/technical-committee considers that formatting is not sufficient reason to introduce the need of parameters (remind the pointer issues).

rubenzorrilla:
> This is a potential improvement of what we merged recently. However, at first glance I think there's a minor behavior change, right now you can only iterate the registry if current level holds a subitem. Maybe I'm missing something, but after this change, it seems to me that it can always be iterated. How is this translated to the Python export considering that the getter(s) are templated?

This comment is outdated. 👍

rubenzorrilla (dismissed):

sunethwarna:
@rubenzorrilla @roigcarlo I added two more ctors, and removed the private `RegistryValueItem` class. Can you take a look?

roigcarlo (approved):

PR #10437, opened 2022-11-11
[Core] Removing 32 bit compilation support

loumalouomega:
**📝 Description**

Removing 32 bit compilation support (@roigcarlo we decided to remove this in Delf)

**🆕 Changelog**

- [Removing 32 bit compilation support](https://github.com/KratosMultiphysics/Kratos/commit/ca93994d66e8bb1125ece5e9bddbeaac0a31fd58)
- [Revert 32 bit support change](https://github.com/KratosMultiphysics/Kratos/commit/cd4a413be69d2b890db8366441d7f14dd520829b)

### DIFF
--- a/kratos/includes/define.h
+++ b/kratos/includes/define.h
@@ -40,12 +40,14 @@
      #define KRATOS_ENV64BIT
    #else
      #define KRATOS_ENV32BIT
+     #error 32 bit system are not supported anymore. Please consider a 64 bits system
   #endif
 #else // It is POSIX (Linux, MacOSX, BSD...)
   #if defined(__x86_64__) || defined(__ppc64__) || defined(__aarch64__)
     #define KRATOS_ENV64BIT
   #else // This includes __arm__ and __x86__
     #define KRATOS_ENV32BIT
+     #error 32 bit system are not supported anymore. Please consider a 64 bits system
   #endif
 #endif
 
--- a/kratos/includes/dof.h
+++ b/kratos/includes/dof.h
@@ -448,11 +448,7 @@ private:
     int mIndex : 6;
 
     /** Equation identificator of the degree of freedom */
-#ifdef KRATOS_ENV32BIT // Required to avoid overflow on 32 bit systems
-    EquationIdType mEquationId : 32;
-#else
     EquationIdType mEquationId : 48;
-#endif
 
     /** A pointer to nodal data stored in node which is corresponded to this dof */
     NodalData* mpNodalData;


matekelemen:
rip rpi 🫡

loumalouomega:
Rip 32bit pi, rpi 3 and 4 now are 64 by default with a significant increase
on performance

matekelemen:
> rpi 3 and 4 now are 64 by default

Ah I didn't know that

kratos/includes/define.h
```diff
@@ -40,12 +40,14 @@
...
    #else
      #define KRATOS_ENV32BIT
+     #error 32 bit system are not supported anymore. Please consider a 64 bits system
   #endif
 #else // It is POSIX (Linux, MacOSX, BSD...)
   #if defined(__x86_64__) || defined(__ppc64__) || defined(__aarch64__)
     #define KRATOS_ENV64BIT
   #else // This includes __arm__ and __x86__
     #define KRATOS_ENV32BIT
+     #error 32 bit system are not supported anymore. Please consider a 64 bits system
```
philbucher, line 50: Can this be checked with CMake?
Seems like the better place to do this

Plus then we could get rid of these macros

loumalouomega: We can...

loumalouomega: Do you know how to check that in CMake?, I can't find an easy manner

roigcarlo: I would just not check it, 32 bit users are scarse

loumalouomega: But without this those scarce developers will be able to compile, and they will have an execution error not explained when using a variable and the hash induced and overflow

rubenzorrilla: Maybe we can just inform about this in the wiki compilation guide.

loumalouomega: You know that our main problem is that people don't read the documentation, I think the check is relevant and must be done

rubenzorrilla: Well, I think that the how to compile is one of the very few that people always check... :confused:

loumalouomega: But still we have the possibility that someone directly compiles without reading it. It is safer this way

roigcarlo: Idk to be honest, I see @loumalouomega points, i don't like it but if you really want to leave it I have no problems with it. Maybe we can remove it later. What do you think?

loumalouomega:
What is the status of this?

roigcarlo (approved):
I think we can approve. I still think we could just assume that 32bit does not exist but its ok.

PR #10440, opened 2022-11-12
[Core] removing no longer needed macros

philbucher:
This has been deprecated for a while and is no longer used


loumalouomega (approved):
Thanks

PR #7809, opened 2020-11-16
[MetisApp] drop support for version 4

philbucher:
**Description**
Using Metis 4 has been long deprecated (Metis 5 was released 7 years ago)
Now using metis 4 is no longer supported, for a while also a proper error is thrown
This is another step towards cleaning the MetisApp

I would wait with the merge until after the release

FYI @KratosMultiphysics/mpi @KratosMultiphysics/altair

PR #10441, opened 2022-11-13
[Metis] remove Metis 4 related code

philbucher:
Support for Metis 4 was dropped almost 2 years ago: #7809

Now I am removing the old code


rubenzorrilla (approved):
Thanks for the clean up!

PR #10442, opened 2022-11-14
[Core] Making ModelPart final. Also adding some missing tests

loumalouomega:
**📝 Description**

Making ModelPart final. Also adding some missing tests

**🆕 Changelog**

- [Making ModelPart final](https://github.com/KratosMultiphysics/Kratos/commit/0a0b6cb614e8bf95866124d954498857e51fed68)
- [Missing tests ModelPart](https://github.com/KratosMultiphysics/Kratos/commit/25232611be9cf2175ccb8367ae7a8f2120bcfd6d)
- [Improve ModelPart test](https://github.com/KratosMultiphysics/Kratos/commit/88844367e73a655b707c53f60f0a6c0d5cb99a77)

PR #10443, opened 2022-11-14
[Core] Adding DataValueContainer to Model

loumalouomega:
**📝 Description**

Adding DataValueContainer to Model. This can be used to manage data during  multistage. Includes #10442

**🆕 Changelog**

- [Making Model derivaded from dataValueContainer](https://github.com/KratosMultiphysics/Kratos/commit/7cc80cd5a5653183c63ad585f61fcc938209c5bb)
- [Expose to python](https://github.com/KratosMultiphysics/Kratos/commit/fcb6b5e4b51f1ff3efeb16031e31bb72e412fc85)
- [Adding test to Model](https://github.com/KratosMultiphysics/Kratos/commit/32c57e4c3d2e6a3e6ee3747e4508f29552299b99)
- [Add cpp test](https://github.com/KratosMultiphysics/Kratos/commit/fc3985703ff9ba240651a28cfc62776aeb704715)


roigcarlo (changes requested):
Wait wait, this is a **very** major change. Blocking it until we can discuss

rubenzorrilla:
For the discussion at the @KratosMultiphysics/technical-committee it would be useful to know which is the use case of this.

loumalouomega:
> For the discussion at the @KratosMultiphysics/technical-committee it would be useful to know which is the use case of this.

This can be used to manage data during multistage. We discussed this in the workshop.

sunethwarna:
I am also interested in this. This is useful in anycase where there are some common data to be stored which does not belong to any model part, but model parts also needs to access them.

Currently this can be achieved by creating duplicates in every model part and managing and synchronizing these data within model parts.

loumalouomega:
Thanks @sunethwarna

kratos/python/add_model_to_python.cpp
```diff
@@ -31,7 +31,7 @@ ModelPart& Model_GetModelPart(Model& rModel, const std::string& rFullModelPartNa
 void  AddModelToPython(pybind11::module& m)
 {
     namespace py = pybind11;
-    py::class_<Model>(m, "Model")
+    py::class_<Model, Kratos::shared_ptr<Model>, DataValueContainer>(m, "Model")
```
sunethwarna:
```suggestion
    py::class_<Model, Model::Pointer, DataValueContainer>(m, "Model")
```

sunethwarna:
I think the `Serialization` will fail with this if there exists data in the container, you may have to call the base class serialization to serialize the data in `DataValueContainer`. Could you check it by adding some variables to the serialization in the following test?
https://github.com/KratosMultiphysics/Kratos/blob/8a9275999e52e82150859538d016bd997ade9670/kratos/tests/test_model.py#L70

kratos/containers/model.h
```diff
@@ -57,6 +56,7 @@ namespace Kratos
 * @author Riccardo Rossi
 */
 class KRATOS_API(KRATOS_CORE) Model final
+    : public DataValueContainer
```
RiccardoRossi: Model is a very basic class. Our (@KratosMultiphysics/technical-committee ) proposal would be to have the model to "have" a data_value_container instead of "being" a datavalue container.

this way one would use it on the line of 

~~~cpp
current_model.GetDataValueContainer()....
~~~

loumalouomega: Take a look now, @RiccardoRossi

kratos/containers/model.h
```diff
@@ -152,15 +151,76 @@ class KRATOS_API(KRATOS_CORE) Model final
      */
     std::vector<std::string> GetModelPartNames() const;
 
+   /**
+     * @brief Gets the value associated with a given variable.
+     * @tparam TDataType The data type of the variable.
+     * @param rThisVariable The variable for which the value is to be retrieved.
+     * @return Reference to the value of the provided variable.
+     */
+    template<class TDataType>
```
RiccardoRossi: the proposal of using the DataValueContainer as an object, was that instead of enrichning the interface of the model one would use the 

~~~py
model.GetDataValueContainer().GetValue(...)
~~~

this allows removing the functions from 154 to 190 as well as the Has method below

kratos/containers/model.h
```diff
@@ -152,15 +151,76 @@ class KRATOS_API(KRATOS_CORE) Model final
...
+     * @return The data value container of the model
+     */
+    DataValueContainer& GetDataValueContainer();
+
+    /**
+     * @brief This method returns the data value container of the model
+     * @details This method is used to access the data value container of the model.
+     * @return The data value container of the model
+     */
+    const DataValueContainer& GetDataValueContainer() const;
```
RiccardoRossi, line 170: with this method you need to add no other methods

kratos/python/add_model_to_python.cpp
```diff
@@ -40,7 +56,22 @@ void  AddModelToPython(pybind11::module& m)
         .def("HasModelPart", &Model::HasModelPart)
         .def("GetModelPartNames", &Model::GetModelPartNames)
         .def("__getitem__", &Model_GetModelPart, py::return_value_policy::reference_internal)
-        .def("__str__", PrintObject<Model>);
+        .def("__str__", PrintObject<Model>)
+        ;
+    VariableIndexingUtility< ModelBinder, Variable<bool> >(model_binder);
```
RiccardoRossi: all of this is not needed as you will use the interface of the DataValueContainer

kratos/tests/test_model.py
```diff
@@ -53,6 +53,9 @@ def test_model(self):
     def _create_and_save_model(self,file_name, serializer_flag):
         current_model = KratosMultiphysics.Model()
 
+        # Setting the data value container
+        current_model.SetValue(KratosMultiphysics.TEMPERATURE, 293.0)
```
RiccardoRossi: this for example becomes

~~~py
    current_model.GetDataValueContainer().SetValue(KratosMultiphysics.TEMPERATURE, 293.0)
~~~

kratos/containers/model.h
loumalouomega: Are you sure about that?, with this you will add more complexity to its usage. To set values in nodes or elements we do not call the container

rubenzorrilla: Just saying that we agreed this on the @KratosMultiphysics/technical-committee. The rationale is to avoid duplicating the interface of the `DataValueContainer` within the `Model`.

rubenzorrilla (approved):
I think now it's ready. Thanks @loumalouomega.

loumalouomega:
> I think now it's ready. Thanks @loumalouomega.

I need @roigcarlo to remove its block

roigcarlo (approved):

loumalouomega:
Finally 

imagen

PR #10428, opened 2022-11-07
[FastPR][All] Upgrade minimum Python version to 3.8

rubenzorrilla:
**📝 Description**
Dear @KratosMultiphysics/all , as described in #10420 we require to upgrade minimum Python version to 3.8. Note that the current version we are using, the 3.6, reached it's end of life past December.

PR #10446, opened 2022-11-14
Relaxing 3.8 errors

roigcarlo:
**📝 Description**
If defined `KRATOS_ENFORCE_PYTHON_VERSION` to false will ignore the python version check (which should be causing your CI to fail)

Should fix problems in #10428. @KratosMultiphysics/altair  We did not understand that it was a complex change to make from your side.

**🆕 Changelog**
Using versions prior to 3.8 should only issue a warning.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -292,7 +292,7 @@ message("AR VERSION: ${CMAKE_AR}")
 
 # check version of Python, needs to be done after including pybind
 if(${PYTHON_VERSION_MAJOR} LESS 3 OR (${PYTHON_VERSION_MAJOR} EQUAL 3 AND ${PYTHON_VERSION_MINOR} LESS 8))
-    message( FATAL_ERROR "Kratos only supports Python version 3.8 and above")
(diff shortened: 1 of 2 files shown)


rubenzorrilla (changes requested):
This would only throw a warning if Python 3.5 is used, which is effectively porting our standard backwards to what we used to had before the registry-related changes.

I also tend to agree @roigcarlo about this. I think this should be done at the @KratosMultiphysics/altair in accordance to your CI requirements.

pooyan-dadvand:
I don't agree with this observation. In fact the one of the CI is something that we can fix. But there are things that are out of our hands. For example, In this case we are also limiting the Cosimulation application in connecting to some pythons (that admitly are EOL) but still in use. Moving officially to 3.8 (even not enforcing it) only complicates such efforts and for me the use case was not paying the price in this case. 

Also the planning and timing weren't optimal. In Kratos we usually notify early and provide an adaption period for such a change. 

IMO, one of the important tasks of the @KratosMultiphysics/technical-committee, is seeing the all scenarios and not few (official) ones and try to balance between cons and pros in a global way.

PR #10447, opened 2022-11-14
Adding PATH parser for windows

roigcarlo:
**📝 Description**
Adds the PATH directories to the search table of shared libs for python >= 3.8 in windows.

This is a quick fix until a better solution is found. Should fix MKL / PARDISO / FEAST / HDF5, etc... problems in win with 3.8

**🆕 Changelog**
- Adding PATH entries to the shared lib lookup locations for python 3.8

### DIFF
--- a/kratos/python_interface/__init__.py
+++ b/kratos/python_interface/__init__.py
@@ -1,5 +1,17 @@
 import os
 import sys
+
+# This is a "dirty" fix to force python to keep loading shared libraries from
+# the PATH in windows (See https://docs.python.org/3/library/os.html#os.add_dll_directory)
+# THIS NEEDS TO BE EXECUTED BEFORE ANY DLL / DEPENDENCY IS LOADED.
+if os.name == 'nt': # This means "Windows"
+    for lib_path in os.environ["PATH"].split(os.pathsep):
+        try:
+            os.add_dll_directory(lib_path.replace("\\","/"))
+        except Exception as e:
+            # We need to capture possible invalid paths.
+            pass
+
 from . import kratos_globals
 from . import python_registry
 


loumalouomega (approved):

AlejandroCornejo (changes requested):
Gonna try it now

matekelemen:
oh god ...
what happened?

AlejandroCornejo:
> oh god ... what happened?

Windows peasants like me could not use MKL hehehe ;)

roigcarlo:
> oh god ... what happened?

https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew

TL;DR, if you have your DLL paths in PATH they are not loaded anymore.

I think it has to do with the fact that both bin and DLL were being loaded from path and is somehow exploitable, but haven't dig into the details. 

Not sure yet up to which point it can be fixed by us alone. In the case of MKL for example the problem is that, even if we copy the DLL into Kratos folder in the install stage, the dependencies of the dependencies are not copied, and since the direct dependency is a entry-point that dynamically loads the correct lib with `LoadLibrary` directly from the code is impossible (or I don't know how to, not gonna parse the strings of the binary....) catch it at that stage.

So that's the better I can come up with for now :S

AlejandroCornejo (approved):
It's working!! :)

Back to the future again :D

AlejandroCornejo:
@lagoncalvesjr FYI

AlejandroCornejo:
BTW we should update the README of the LinearSolvers then

AlejandroCornejo:
@roigcarlo should we merge?

roigcarlo:
Yup, if no one else sees any problem it can go in.

I promise we will look for a better solution

PR #10456, opened 2022-11-15
[Metis] more cleanup

philbucher:
This PR contains:
- very minor cleanup in existing files
- addition of empty files for utilities. I will add the functionality in future PRs. I decided to do it slow to make the review as simple as possible


rubenzorrilla (approved):
Fine for me, but creating a new file to add legacy utilities seems a bit odd to me (I guess this will be clarified in the upcoming PRs).

philbucher:
> Fine for me, but creating a new file to add legacy utilities seems a bit odd to me (I guess this will be clarified in the upcoming PRs).

It is, and I would not have done it usually. In this case I want to split and simplify some refactorings as much as possible to avoid breaking stuff. I can point it out in the next PRs ;)

PR #10457, opened 2022-11-15
[Metis] split header source

philbucher:
Looks like a lot, but I only moved the implementation to the source files


rubenzorrilla (approved):
Seems so. Thanks.

PR #10458, opened 2022-11-16
[Metis] moving legacy utility functions

philbucher:
next step of refactoring, moving some utility functions

I know they are ugly, but in order to avoid breaking stuff I move them ass is and refactor them later


roigcarlo:
Does this have to do with the reordering thing I Presume?

philbucher:
Not yet
I want to add some functionality, but currently the app is quite a mess

I am only restructuring the code a bit, bit not refactoring it

Will let you know if I change/improve sth

roigcarlo (approved):

PR #10460, opened 2022-11-17
[CI] compiling centos in release

ddiezrod:
**📝 Description**
I found why centos was failing in our CI if we did not remove the maybe-uninitialized warnings. It seems that this warning only appears when compiling in Release mode. When I created the centos CI I used Custom as I saw that this was used in other compilations. Is there any reason why we should use Custom? 

@roigcarlo @philbucher @KratosMultiphysics/technical-committee ?

### DIFF
--- a/.github/workflows/centos_configure.sh
+++ b/.github/workflows/centos_configure.sh
@@ -42,8 +42,8 @@ cmake -H"${KRATOS_SOURCE}" -B"${KRATOS_BUILD}/${KRATOS_BUILD_TYPE}" \
 ${KRATOS_CMAKE_OPTIONS_FLAGS} \
 -DUSE_MPI=OFF \
 -DPYBIND11_PYTHON_VERSION="3.8" \
--DCMAKE_CXX_FLAGS="${KRATOS_CMAKE_CXX_FLAGS} -O0 -Wall" \
--DCMAKE_UNITY_BUILD=ON                              \
+-DCMAKE_CXX_FLAGS="${KRATOS_CMAKE_CXX_FLAGS} -O0 -Wall -Werror=maybe-uninitialized" \
+-DCMAKE_UNITY_BUILD=ON
 
 # Buid
 cmake --build "${KRATOS_BUILD}/${KRATOS_BUILD_TYPE}" --target install -- -j2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -179,7 +179,7 @@ jobs:
   centos:
     runs-on: ubuntu-latest
     env:
-      KRATOS_BUILD_TYPE: Custom
+      KRATOS_BUILD_TYPE: Release
 
     container:
       image: kratosmultiphysics/kratos-image-ci-centos7:latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -179,7 +179,7 @@ jobs:
   centos:
     runs-on: ubuntu-latest
     env:
-      KRATOS_BUILD_TYPE: Release
+      KRATOS_BUILD_TYPE: Custom
 
     container:
       image: kratosmultiphysics/kratos-image-ci-centos7:latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -179,7 +179,7 @@ jobs:
   centos:
     runs-on: ubuntu-latest
     env:
-      KRATOS_BUILD_TYPE: Custom
+      KRATOS_BUILD_TYPE: Release
 
     container:
       image: kratosmultiphysics/kratos-image-ci-centos7:latest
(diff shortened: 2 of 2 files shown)


philbucher:
one word: compilation time :)
Tends to take muuuuch longer (hours) with all optimizations enabled

But lets see how long it takes

roigcarlo:
In any case, if time is an issue (I suspect it will) we can identify the specific flag that makes that warning appear and add it in the custom collection. It should cause the same effect.

The problem with time tends to be the collection of optimizations inside -02

ddiezrod:
Here is the error [ 55%] Building CXX object kratos/CMakeFiles/Kratos.dir/python/add_geometrical_utilities_to_python.cpp.o
In file included from /__w/Kratos/Kratos/kratos/geometries/point.h:25,
                 from /__w/Kratos/Kratos/kratos/integration/integration_point.h:23,
                 from /__w/Kratos/Kratos/kratos/geometries/geometry_data.h:23,
                 from /__w/Kratos/Kratos/kratos/geometries/geometry.h:29,
                 from /__w/Kratos/Kratos/kratos/geometries/line_2d_2.h:26,
                 from /__w/Kratos/Kratos/kratos/geometries/triangle_2d_3.h:26,
                 from /__w/Kratos/Kratos/applications/FluidDynamicsApplication/fluid_dynamics_application.cpp:19,
                 from /__w/Kratos/Kratos/build/Release/applications/FluidDynamicsApplication/CMakeFiles/KratosFluidDynamicsCore.dir/Unity/unity_0_cxx.cxx:3:
/__w/Kratos/Kratos/kratos/containers/array_1d.h: In member function 'void Kratos::Serializer::load(const string&, Kratos::DenseVector<TDataType>&) [with TDataType = Kratos::array_1d<double, 2>]':
/__w/Kratos/Kratos/kratos/containers/array_1d.h:143:23: error: '<anonymous>' may be used uninitialized in this function [-Werror=maybe-uninitialized]
         data_ (v.data_) {}

ddiezrod:
> one word: compilation time :) Tends to take muuuuch longer (hours) with all optimizations enabled
> 
> But lets see how long it takes

Those are two words : P. Let's fix the issue first and see how long it takes, I hope it is not that bad....

roigcarlo:
If this is the error, my suggestion would be to change https://github.com/KratosMultiphysics/Kratos/blob/master/.github/workflows/centos_configure.sh

```Bash
# Configure
cmake -H"${KRATOS_SOURCE}" -B"${KRATOS_BUILD}/${KRATOS_BUILD_TYPE}" \
${KRATOS_CMAKE_OPTIONS_FLAGS} \
-DUSE_MPI=OFF \
-DPYBIND11_PYTHON_VERSION="3.8" \
-DCMAKE_CXX_FLAGS="${KRATOS_CMAKE_CXX_FLAGS} -O0 -Wall" \
-DCMAKE_UNITY_BUILD=ON                              \
```

to:

```Bash
# Configure
cmake -H"${KRATOS_SOURCE}" -B"${KRATOS_BUILD}/${KRATOS_BUILD_TYPE}" \
${KRATOS_CMAKE_OPTIONS_FLAGS} \
-DUSE_MPI=OFF \
-DPYBIND11_PYTHON_VERSION="3.8" \
-DCMAKE_CXX_FLAGS="${KRATOS_CMAKE_CXX_FLAGS} -O0 -Wall -Werror=maybe-uninitialized" \
-DCMAKE_UNITY_BUILD=ON
```

And remove the last \ because it will give problems in the future

ddiezrod:
@roigcarlo That is what we did on our side. I do not know how serious this warning is, but if it's okay for you then I will do that.

ddiezrod:
> @roigcarlo That is what we did on our side. I do not know how serious this warning is, but if it's okay for you then I will do that.

I realized I misread your comment, you meant to force this to be an error without changing build type. Lets see.

ddiezrod:
It is still not detecting the warning if we compile in Custom. I guess at the end the point is if we want to silence this error or not.

roigcarlo:
Then I'd beat it means that some optimization may be reordering the code, but I am not sure. In that case go ahead and use release if you want, since its only the code we can assume the cost.

ddiezrod:
So I found where the error is coming from (when compiling without unity it becomes much more clear).
All the d_vms elements from FluidDynamicsApp load the OldSubscaleVelocity.

https://github.com/KratosMultiphysics/Kratos/blob/0ab138248bb906ea965d3029a6d9803318dd2801/applications/FluidDynamicsApplication/custom_elements/d_vms.cpp#L894

but it seems that just with this it does not detect the DenseVector was initialized
https://github.com/KratosMultiphysics/Kratos/blob/0ab138248bb906ea965d3029a6d9803318dd2801/applications/FluidDynamicsApplication/custom_elements/d_vms.cpp#L32

(definition here):
https://github.com/KratosMultiphysics/Kratos/blob/0ab138248bb906ea965d3029a6d9803318dd2801/applications/FluidDynamicsApplication/custom_elements/d_vms.h#L279

I will try to find the way to initialize this correctly, if anyone has any suggestions I can give it a try in my docker. Pinging also @rubenzorrilla as this affect FluidDynApp

PR #10461, opened 2022-11-17
Fix compilation with python 3.11 and greater

roigcarlo:
**📝 Description**
SO was deprecated from 3.11 so extension appeared as None and caused compiled libs to have incorrect names. Porting the solution from Pybind.

Ping @AlejandroCornejo (I think you also mentioned problems with this) and @tteschemacher. That should solve your problems. 

**🆕 Changelog**
- Using `USE_SYSCONFIG`to obtain the module extension.


tteschemacher:
Nice. Thank you!

AlejandroCornejo:
I KNEW it!!! :) 

imagen

AlejandroCornejo (approved):
niceee

PR #10463, opened 2022-11-17
[Core] Adding intersection line-tetrahedra

loumalouomega:
**📝 Description**

Adding intersection line-tetrahedra in IntersectionUtilities. Depends on https://github.com/KratosMultiphysics/Kratos/pull/10413

**🆕 Changelog**

- [Adding intersection line-tetrahedra](https://github.com/KratosMultiphysics/Kratos/commit/77f005d6fae0ad246d621871bd0d5d1ac41e8fd5)
- [Added test](https://github.com/KratosMultiphysics/Kratos/commit/1cd6b292ee8680d28a5cf391e2160b5fa7d82e39)


AlejandroCornejo:
just for my curiosity, this is a way of finding intersection points between line elements and tets? There is any plan for hexas?

loumalouomega:
> just for my curiosity, this is a way of finding intersection points between line elements and tets? There is any plan for hexas?

Not for now

loumalouomega:
Hello, anyone? @rubenzorrilla ?

loumalouomega:
Hello this is blocking other developments @KratosMultiphysics/technical-committee

kratos/utilities/intersection_utilities.h
```diff
@@ -233,6 +237,335 @@ class IntersectionUtilities
...
+     *         2 (intersect in one point)
+     *         3 (intersect in two points inside the tetrahedra)
+     *         4 (intersect in two points, one inside the tetrahedra)
+     *         5 (intersect in the first corner of the tetrahedra)
+     *         6 (intersect in the second corner of the tetrahedra)
+     *         7 (intersect in the thid corner of the tetrahedra)
+     *         8 (intersect in the fourth corner of the tetrahedra)
+     */
+    template <class TGeometryType, bool TConsiderInsidePoints = true>
+    static int ComputeTetrahedraLineIntersection(
```
pooyan-dadvand: Wouldn't be better to return a flag and marking where it is collide:

Is(IntersectionUtilities::INTERSECTED)
Is(IntersectionUtilities::INSIDE)
Is(IntersectionUtilities::EDGE)
Is(IntersectionUtilities::FACE)
etc.....

kratos/utilities/intersection_utilities.h
```diff
@@ -233,6 +237,335 @@ class IntersectionUtilities
...
+        const TGeometryType& rTetrahedraGeometry,
+        const array_1d<double,3>& rLinePoint1,
+        const array_1d<double,3>& rLinePoint2,
+        array_1d<double,3>& rIntersectionPoint1,
+        array_1d<double,3>& rIntersectionPoint2,
+        const double Epsilon = 1e-12
+        ) 
+    {
+        // Zero tolerance
+        const double zero_tolerance = std::numeric_limits<double>::epsilon();
```
pooyan-dadvand: Why are you using a different tolerance than the given one?

kratos/utilities/intersection_utilities.h
```diff
@@ -233,6 +237,335 @@ class IntersectionUtilities
...
+     *         4 (intersect in two points, one inside the tetrahedra)
+     *         5 (intersect in the first corner of the tetrahedra)
+     *         6 (intersect in the second corner of the tetrahedra)
+     *         7 (intersect in the thid corner of the tetrahedra)
+     *         8 (intersect in the fourth corner of the tetrahedra)
+     */
+    template <class TGeometryType, bool TConsiderInsidePoints = true>
+    static int ComputeTetrahedraLineIntersection(
+        const TGeometryType& rTetrahedraGeometry,
+        const array_1d<double,3>& rLinePoint1,
```
pooyan-dadvand: Making this type a template argument would help to pass node directly

kratos/utilities/intersection_utilities.h
```diff
@@ -233,6 +237,335 @@ class IntersectionUtilities
...
+        const array_1d<double,3>& rLinePoint2,
+        array_1d<double,3>& rIntersectionPoint1,
```
pooyan-dadvand: Same here, would be good to have it as template argument.

kratos/utilities/intersection_utilities.h
```diff
@@ -233,6 +237,335 @@ class IntersectionUtilities
...
+
+        // Lambda function to check the inside of a line corrected
+        auto is_inside_projected = [] (auto& rGeometry, const array_1d<double,3>& rPoint) -> bool {
+            // We compute the distance, if it is not in the plane we project
+            const Point point_to_project(rPoint);
+            Point point_projected;
+            const double distance = GeometricalProjectionUtilities::FastProjectOnLine(rGeometry, point_to_project, point_projected);
+
+            // We check if we are on the plane
+            if (std::abs(distance) > 1.0e-6 * rGeometry.Length()) {
```
pooyan-dadvand: Epsilon?

kratos/utilities/intersection_utilities.h
```diff
@@ -233,6 +237,335 @@ class IntersectionUtilities
...
+                return false;
+            }
+            array_1d<double, 3> local_coordinates;
+            return rGeometry.IsInside(point_projected, local_coordinates);
+        };
+
+        int solution = 0;
```
pooyan-dadvand: I would start with the point inside check because it is faster than the rest and can be concluding. If both are out I would check the faces.

kratos/utilities/intersection_utilities.h
```diff
@@ -233,6 +237,335 @@ class IntersectionUtilities
...
+            if (face_solution == 1) { // The line intersects the face
+                if (solution == 0) {
+                    noalias(rIntersectionPoint1) = intersection_point;
+                    solution = 2;
+                } else {
+                    if (norm_2(rIntersectionPoint1 - intersection_point) > Epsilon) { // Must be different from the first one
+                        noalias(rIntersectionPoint2) = intersection_point;
+                        solution = 1;
+                        break;
+                    }
```
pooyan-dadvand: else?!?

For a sliver and relax tolerance, the else may happen

kratos/utilities/intersection_utilities.h
```diff
@@ -233,6 +237,335 @@ class IntersectionUtilities
...
+                }
+            } else if (face_solution == 2) { // The line is coincident with the face
+                // Compute intersection with the edges
+                for (auto& r_edge : r_face.GenerateEdges()) {
```
pooyan-dadvand: Are you sure that you should do this loop? Is not cheaper to calculate the edge by the shape functions of the face?

kratos/utilities/intersection_utilities.h
```diff
@@ -233,6 +237,335 @@ class IntersectionUtilities
...
+     * @return The intersection type index:
+     *         0 (disjoint - no intersection)
+     *         1 (intersect in two points)
+     *         2 (intersect in one point)
+     *         3 (intersect in two points inside the tetrahedra)
+     *         4 (intersect in two points, one inside the tetrahedra)
+     *         5 (intersect in the first corner of the tetrahedra)
+     *         6 (intersect in the second corner of the tetrahedra)
+     *         7 (intersect in the thid corner of the tetrahedra)
+     *         8 (intersect in the fourth corner of the tetrahedra)
```
pooyan-dadvand: Would you please add the reference of your algorithm?

Meanwhile, I know this one for ray and tet which can be interesting: https://github.com/erich666/jgt-code/blob/e67d05e4398c737abc40744cf3984c64b7df1e84/Volume_08/Number_4/Platis2003/RayTetraPluecker.cpp#L52

kratos/utilities/intersection_utilities.h
```diff
@@ -233,6 +237,335 @@ class IntersectionUtilities
...
+    /**
+     * @brief Calculates the line to line intersection (shortest line). If line is length 0, it is considered a point and therefore there is intersection (3D version)
+     * @details Calculate the line segment PaPb that is the shortest route between two lines P1P2 and P3P4. Calculate also the values of mua and mub where
+     *    Pa = P1 + mua (P2 - P1)
+     *    Pb = P3 + mub (P4 - P3)
+     *    http://paulbourke.net/geometry/pointlineplane/
+     * @param rSegment1 The first segment
+     * @param rSegment2 The second segment
+     * @tparam TGeometryType The geometry type
+     * @return Return empty points array if no solution exists. Otherwise returns the line intersection points array
```
pooyan-dadvand: hmmm, is there a situation that shortest line doesn't exists? I assume that for parallel lines we can choose any.

kratos/utilities/intersection_utilities.h
pooyan-dadvand: I will revise this part when we conclude that this is the algorithm to go

kratos/utilities/intersection_utilities.h
pooyan-dadvand: Please add your reference in the documenation

kratos/utilities/intersection_utilities.h
```diff
@@ -233,6 +237,335 @@ class IntersectionUtilities
...
+                    GeometricalProjectionUtilities::FastProjectOnLine(rSegment1, rSegment2[0], projected_point);
+                    if (rSegment1.IsInside(projected_point, local_coords)) {
+                        pa->Coordinates() = rSegment2[0].Coordinates();
+                        pb->Coordinates() = projected_point;
+                    } else {
+                        GeometricalProjectionUtilities::FastProjectOnLine(rSegment1, rSegment2[1], projected_point);
+                        if (rSegment1.IsInside(projected_point, local_coords)) {
+                            pa->Coordinates() = rSegment2[1].Coordinates();
+                            pb->Coordinates() = projected_point;
+                        } else { // Parallel and not projection possible
```
pooyan-dadvand: I don't understand, you already knew that they are parallel

Meanwhile, two parallel line segments have also shortest line connection

pooyan-dadvand (commented):
I am not sure if the algorithms are the most efficient ones and I think the one of the shortest line is not very well defined.

kratos/utilities/intersection_utilities.h
loumalouomega: Okay

kratos/utilities/intersection_utilities.h
loumalouomega: This is to check that the norm is not 0, so it will not produce NaN

kratos/utilities/intersection_utilities.h
loumalouomega: Okay

kratos/utilities/intersection_utilities.h
loumalouomega: It's a different criteria, but I can consider Epsilon as well

kratos/utilities/intersection_utilities.h
loumalouomega: Okay, as I added corner cases while I was founding them maybe I produced inefficient code

kratos/utilities/intersection_utilities.h
loumalouomega: In that case, it would continue iterating

kratos/utilities/intersection_utilities.h
loumalouomega: > Are you sure that you should do this loop? Is not cheaper to calculate the edge by the shape functions of the face?

That could be a good idea...

kratos/utilities/intersection_utilities.h
loumalouomega: This is added originally in #10413, we can discuss there. When you say documentation, which one do you mean?

kratos/utilities/intersection_utilities.h
loumalouomega: > I don't understand, you already knew that they are parallel
> 
> Meanwhile, two parallel line segments have also shortest line connection

Better to discuss in #10413. I added the parallel case by suggestion of @matekelemen . A priori the last case should never happen, I may remove the else

kratos/utilities/intersection_utilities.h
pooyan-dadvand: I understood, but the question is rather subjective, shall we use the same Epsilon here to be coherent?

kratos/utilities/intersection_utilities.h
pooyan-dadvand: I mean in the comment in the code

kratos/utilities/intersection_utilities.h
loumalouomega: I would not, because the norm_2 equal to zero is a more numerical check than a tolerance check

kratos/utilities/intersection_utilities.h
loumalouomega: I was thinking about this. We are not checking if we intersect (this is already implemented in HasIntersection method. which actually follows this strategy). We are determining the actual intersection points. The points in the faces. corners, etc... are actually inside, so if we follow this strategy we cannot determine in which scenario are we (in addition to the intersection points we inform about the intersection scenario)

kratos/utilities/intersection_utilities.h
loumalouomega: > Are you sure that you should do this loop? Is not cheaper to calculate the edge by the shape functions of the face?

At the end the problem is the same as said before. The devil is in details, and we want to capture the kind of intersection we are. With the shape functions, we cannot capture the type of intersection

kratos/utilities/intersection_utilities.h
loumalouomega: After check, I see that this return in the else case is an empty vector and therefore informing that there is no solution possible

loumalouomega:
👋

kratos/tests/cpp_tests/utilities/test_intersection_utilities.cpp
```diff
@@ -584,6 +585,95 @@ namespace Testing {
         KRATOS_CHECK_EQUAL(int_id, 1);
     }
 
+    KRATOS_TEST_CASE_IN_SUITE(ComputeTetrahedraLineIntersection, KratosCoreFastSuite)
```
rubenzorrilla: I think this is a pretty large test. I'd suggest separating in smaller ones. This eases a lot (eventual) debugging.

kratos/utilities/intersection_utilities.h
```diff
@@ -38,6 +38,19 @@ namespace Kratos
 ///@name  Enum's
 ///@{
 
+enum class IntersectionUtilitiesLineIntersection
```
rubenzorrilla:
```suggestion
enum class IntersectionUtilitiesTetrahedraLineIntersectionStatus
```
Besides I don't really love the idea of defining a new enum in the entire scope of Kratos. I'd do as we did so far, return an integer with the intersection status and properly document it.

rubenzorrilla (changes requested):
Some  comments.

loumalouomega:
> Some comments.

Done

kratos/utilities/intersection_utilities.h
rubenzorrilla: Note the comment below. This is not resolved.

loumalouomega: Before I did it with an integer, but then it was complex to understand the code when reading it. Can we pass with the enum?

loumalouomega: Hello @rubenzorrilla

rubenzorrilla: I think that if it is properly documented it is not very much difficult to understand the code. Besides it would be much less verbose, consistent with the other intersections and we avoid having a hanging enum in the complete Kratos namespace (I don't want misuses of this). I'm of course considering that, in my opinion, this is a pretty internal function that would potentially be used by very few developers.

I'd only be in favor of the enum if we agreed a standard intersection status valid for all intersections. Not sure if possible.

loumalouomega: Okay, I will do something that may satisfy everyone

kratos/tests/cpp_tests/utilities/test_intersection_utilities.cpp
rubenzorrilla: This is not done yet.

kratos/utilities/intersection_utilities.h
```diff
@@ -237,6 +235,266 @@ class KRATOS_API(KRATOS_CORE) IntersectionUtilities
...
+     * @param Epsilon The tolerance
+     */
+    template <class TGeometryType, class TCoordinatesType>
+    static void ComputeTriangleLineIntersectionInTheSamePlane(
+        const TGeometryType& rTriangleGeometry,
+        const TCoordinatesType& rLinePoint1,
+        const TCoordinatesType& rLinePoint2,
+        TCoordinatesType& rIntersectionPoint1,
+        TCoordinatesType& rIntersectionPoint2,
+        int& rSolution,//IntersectionUtilitiesTetrahedraLineIntersectionStatus& rSolution,
```
rubenzorrilla, line 261:
```suggestion
        int& rSolution,//IntersectionUtilitiesTetrahedraLineIntersectionStatus& rSolution,
```
Wouldn't it be better to return the intersection status rather than passing it as a reference? This is what we normally do in the other methods.

loumalouomega: This method is used as an auxiliary method as well, must be as a reference

kratos/tests/cpp_tests/utilities/test_intersection_utilities.cpp
loumalouomega: Okay

kratos/utilities/intersection_utilities.h
rubenzorrilla: But why needs to be a reference when is used as an auxiliary method?

loumalouomega: Because the previous information is relevant to determine the intersection status

kratos/tests/cpp_tests/utilities/test_intersection_utilities.cpp
loumalouomega: Done

kratos/utilities/intersection_utilities.h
loumalouomega: Now also returns the int

rubenzorrilla: OK. Now I understand your point. Thanks.

rubenzorrilla:
Why leaving the enum code? Are you planning to implement a generic stuff?

loumalouomega:
> Why leaving the enum code? Are you planning to implement a generic stuff?

Because the code is easier to understand that way

loumalouomega:
This is ready @rubenzorrilla

loumalouomega:
> This is ready @rubenzorrilla

Ping

loumalouomega:
imagen

@rubenzorrilla

loumalouomega:
> imagen
> 
> @rubenzorrilla

👋

rubenzorrilla (approved):

PR #10468, opened 2022-11-18
[Core] Adding combine model part modeler

ddiezrod:
**📝 Description**
This implements a modeler to combine model parts into a new one. The input is a list of model parts with origin and destination path (inside origin and destination model part structure) so you have complete control on how the structure of your combined model part looks.

The list of origin model parts should have the same root model part. This root model part will be used to create process info and variable list.

ping @KratosMultiphysics/technical-committee @KratosMultiphysics/implementation-committee


kratos/modeler/combine_model_part_modeler.cpp
```diff
@@ -0,0 +1,413 @@
...
+{
+    KRATOS_TRY;
+    const auto& r_new_model_part_name = mParameters["combined_model_part_name"].GetString();
+
+    auto& r_combined_model_part = mpModel->HasModelPart(r_new_model_part_name) ? 
+        mpModel->GetModelPart(r_new_model_part_name) : mpModel->CreateModelPart(r_new_model_part_name);
+
+    //TODO: RESET combined model part
+
+    this->CopyCommonData(r_combined_model_part);
```
roigcarlo, line 52: I think this will break.

In the example you are relaying on the fact that the default `main_model_part_name` is `main_model_part`, and all your sources are in fact `main_model_part`.

But you can easily make all the functions below return wrong data if you for example try to 
```json
{
    "combined_model_part_name" : "thermal_model_part",
    "condition_name" : "SurfaceCondition3D3N",
    "model_part_list" : [
        {
            "origin_model_part": "main_model_part.submodelpart_solid.chiller",
            "destination_model_part": "thermal_model_part.submodelpart_solid.chiller"
        },
        {
            "origin_model_part": "not_my_main_model_part.submodelpart_solid.contact_solid2liquid.contact_chiller_62-2liquid",
             "destination_model_part": "thermal_model_part.submodelpart_solid.contact_solid2liquid.contact_chiller_62-2liquid"
        }
    ]
}
```

I think that we should check that none of the modelparts in the combine list comes from a origin different from the one specified (even if its the default one) or either create a map that handles multiple origin/destination permutations if you want the process to be able to work on the above case.

I think the same happens if one specifies a different output, but for that I would leave it as it is, as its odd that you want to have multiple outputs (still the check would be nice)

ddiezrod: Having the possibility of combining model parts from different places is very useful  so I will think about that. In our cases in altair everything is read inside main model part but I think it would make this modeler mucho mor interesting for the rest of the community if we can read from several input model parts.

ddiezrod: I was discussing this with @pooyan-dadvand and @jrubiogonzalez and we think that it might be quite tricky to do it correctly with model parts that have different origins. (Variable list should be the same, which process info should we use etc.) . So, as a first iteration, I am checking that all origin model parts have the same root.

roigcarlo: Perfect. I did not want you to implement the other, just to add the check :P

ddiezrod: done ;)

ddiezrod:
Any thoughts, suggestions, insults...? @KratosMultiphysics/implementation-committee @KratosMultiphysics/technical-committee @philbucher @KratosMultiphysics/altair

loumalouomega (approved):

ddiezrod:
If someone wants to take a look before merging this let me know. Otherwise I will merge this tomorrow.

kratos/modeler/combine_model_part_modeler.cpp
rubenzorrilla: > I was discussing this with @pooyan-dadvand and @jrubiogonzalez and we think that it might be quite tricky to do it correctly with model parts that have different origins. (Variable list should be the same, which process info should we use etc.) . So, as a first iteration, I am checking that all origin model parts have the same root.

I did not follow the discussion at all but I think that this does pretty much sense. :+1:

PR #10471, opened 2022-11-18
[Fluid][Bugfix] Outlet inflow contribution enhancements

rubenzorrilla:
**📝 Description**
This fixes a long term problem with the reversed flows contribution in the NS wall condition. Basically there was the chance to end with a memory leak if tagging the condition as `OUTLET` but without using the  `apply_outlet_process.py`. This is now solved by adding a switch variable that defaults to `false` if not provided.

Besides I took the chance to add some missing testing and enhance the usage of this.


rubenzorrilla:
In an upcoming PR I'll do similar stuff with the `SLIP` boundaries and their contributions in the `NavierStokesWallCondition`.

applications/FluidDynamicsApplication/python_scripts/apply_outlet_process.py
```diff
@@ -24,7 +25,10 @@ def __init__(self, Model, settings ):
             "value"              : 0.0,
             "interval"           : [0.0,"End"],
             "hydrostatic_outlet" : false,
-            "h_top"              : 0.0
+            "h_top"              : 0.0,
+            "outlet_inflow_contribution" : false,
```
ddiezrod, line 28: Are you sure you prefer false as default?

rubenzorrilla: I was wondering whether we should put it active or inactive by default... I convinced myself because this, of course, slightly affects the mass conservation, something that might confuse the user if specifically checks for this without being notified that this extra contribution is added.

ddiezrod: Ok, I see, it was more out of curiosity.

rubenzorrilla:
Anyone @KratosMultiphysics/fluid-dynamics ?

ddiezrod (approved):

rubenzorrilla:
Thanks @ddiezrod!

PR #10472, opened 2022-11-18
[Core][ModelPartIO] unify DivideInputToPartitions

philbucher:
The two methods `DivideInputToPartitions` (one with files and one with streams as input) duplicate most of their logic.
I refactored to remove this duplication. This way it will be easier to extend (e.g. for extending to geometries, which I am currently looking into)


rubenzorrilla (approved):
As noted in the PR description, this is just refactoring. Thanks.

PR #8599, opened 2021-04-08
[CI][Intel] enable MetisApplication

philbucher:
**Description**
This enables the MetisApp in the CI that uses the IntelCompiler

I am adding it first as it is needed for many MPI tests in the Core (for the partitioning)

PR #10475, opened 2022-11-20
[CI][Intel] enable MetisApp

philbucher:
second attempt to activate the MetisApp in the Intel build
Hopefully now that some time has passed and the versions of some compilers were updated it would work now

see #8599


philbucher:
Still doesn't work 

Lets hope it will work with icx

PR #10476, opened 2022-11-20
[CI][Intel] using icx

philbucher:
The `icc` compiler is deprecated, this PR updates the Intel CI to use the new `icx` compiler


philbucher:
Build times legacy intel compiler (ICC/ICPC): 1h 15min
Build times new LLVM-based Intel compiler (ICX/ICPX): 22min

Since the legacy one will be removed soon from what I understand, I would like to merge this as is and enable the tests (which currently segfault for some reason) in a future PR

roigcarlo (commented):
Makes sense to me, 

I would leave the old one because some people still uses it (maybe not mandatory), but agree with moving the "official" one to the new.

philbucher:
I did not remove the old one. I would leave it for a while longer, at least if it does not cause us any problems :)

roigcarlo (approved):

PR #10474, opened 2022-11-20
[CI] 1/2 Add mmg and parmmg stable versions to ubuntu dockerfiles

marcnunezc:
**📝 Description**

Adding a stable version of ParMmg (v1.4.0) which requires version 5.6.0 of MMG.

The idea is to do the upgrade in three different PRs:
- [ ]  Add new mmg and parmmg to Dockerfiles
- [ ]  Change versions used in ci.yml and nightly.yml
- [ ] Remove old versions from Dockerfile.

It may help #10450. This upgrade also removes all the annoying warnings that currently appear in the logs.

**🆕 Changelog**
Please summarize the changes in one list to generate the changelog:
E.g.
- Adding ParMmg v1.4.0 and Mmg v5.5.6

PR #10477, opened 2022-11-20
[CI] 2/2 Use ParMmg 1.4.0 and Mmg 5.6.0 in CI

marcnunezc:
**📝 Description**
continuation of #10474 

- [x]  Add new mmg and parmmg to Dockerfiles
- [ ]  Change versions used in ci.yml and nightly.yml
- [ ] Remove old versions from Dockerfile.

**🆕 Changelog**

- Use new versions in ci files. 
- Remove old versions from docker


philbucher:
I think you could already remove the old version from the docker in this PR, since it will be pushed only after merging

marcnunezc:
> I think you could already remove the old version from the docker in this PR, since it will be pushed only after merging

True, just did it. Older branches will need to merge master when opening a PR though, as their CI files will be outdated. (the other option is to wait some days to remove the older versions, but I agree to do it now)

loumalouomega (approved):

loumalouomega:
@marcnunezc it would be a good opportunity to update MMG 5.7

marcnunezc:
Yes I think I will close this and directly upgrade to 5.7.0.

Issue #10464, opened 2022-11-17
Nearest element mapper - non-consistent value n MPI


ddiezrod:
I was doing a test for the nearest element mapper. I took a case where the mesh was conformant and moved one node a little so the nearest neighbour mapper would detect it as non-conformant.

What I have found is that while running in mpi the value on the other side of the mesh (the side that was not modified) is changing depending on the number of cores (very significantly). Is this an "expected" behaviour? I tried to debug it a little but I could not completely follow the code, especially in MPI. @philbucher Could you give me any guidance on how to debug it or anything I could try?

I upload the test with the mdpa just for reference.

ddiezrod:
[test_mapper.zip](https://github.com/KratosMultiphysics/Kratos/files/10032657/test_mapper.zip)

philbucher:
so just to understand, the problem is that values in the origin modelpart are being modified during mapping?
In your example you use inversemap, so the values there should change

Can you also comment on how the values change and what can be done to reproduce the problem?

ddiezrod:
Sorry for the poor explanation. I basically have two conforming meshes except for two nodes (id 313 on solid side and 122 on liquid side) which are in a slightly different position. I am using the nearest element mapper to map values from solid to liquid. After mapping, all values are exactly the same as the one on the mold side except for 122. This is normal as it does not coincide exactly with its correspondent node.

If I run the problem in serial the value on this node (122) is 0.00093982 which makes sense more or less as the value on 313 is 0.001. The problem is that when I run this in MPI the value of this node changes depending on the number of cores.
- With 8 cores, the value is 0.00093982 which is correct in principle as it is the same value that we got in serial.
- With 4 cores, the value I get is -4.754481e-05 so obviously the test fails there.

I guess that this means that in MPI it is reading its value from a different element so that looks like a possible error in the code. Maybe I am missing something in the test too... Is there a way to see which element is being used for each node? I dont mind  debugging a long print I do not even know where to put those prints :D

mpentek:
I am interested in this issue. Throwing one more aspect into the discussion, could this be related also to small numbers?...

philbucher:
It could be that the search is only finding an approximation 

Did you increase the echo level and check the search output?

Try increasing the search radius

ddiezrod:
Ok, I think I got some information about what is going on. I ran again the cases with 4 and 8 cores. The status at the end seem to be fine for both cases.

With 8 cores:
```
[...]

PR #10481, opened 2022-11-22
[Mapping] sync init search radius

ddiezrod:
I open this PR as a possible solution to be discussed for https://github.com/KratosMultiphysics/Kratos/issues/10464 . The idea is just to have the same initial search radius in all partitions so results between mpi with any number of cores and smp match.

I did not go very deep to see if all this actually makes complete sense, as I said, I open this just to have a place to discuss.

### DIFF
--- a/applications/MappingApplication/custom_searching/interface_communicator.cpp
+++ b/applications/MappingApplication/custom_searching/interface_communicator.cpp
@@ -66,23 +66,6 @@ T MinAll(
     return Value;
 }
 
-// Since either ModelPart might not be part of this rank,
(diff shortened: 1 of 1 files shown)


philbucher:
Hey,
I think I have an idea why it is not consistent in MPI
I will take a look in the next days

ddiezrod:
> Hey, I think I have an idea why it is not consistent in MPI I will take a look in the next days

Hi Ok, I was playing a little with the way to compute the initial search radius here but it seems it is still failing for some reason...

PR #10487, opened 2022-11-24
[MappingApplication] Adding 3D/2D metamapper

loumalouomega:
**📝 Description**

### Explanation

In @KratosMultiphysics/altair, @pooyan-dadvand has found that in the specific case where the origin model part is 2D, or an extruded 3D solution, the mapper may not be as robust as it should. In order to tackle this @pooyan-dadvand suggested creating a metamapper which obtains the solution for the 3D destination model part from the original 2D solution. This in addition to be more robust it will be cheaper in case of a 2D origin model part is taken as the number of nodes will be significantly lower.

For doing that, this metamapper projects the destination nodes to the origin 2D model part, and then creates the chosen "base_mapper" (only interpolation mappers).

**Mapping example**

image

### Usage

It is used as any other mapper, with the following specific settings:

- `base_mapper`: The mapper used as actual mapper, default is `nearest_neighbor`

### Notes

**NOTE:** Sorry @philbucher, but I needed to modify a bit the base interpolation mapper, please don't kill me. Many methods and variables are private and not exposed and doing a meta-mapper I needed to access some data, I tried to minimize the modifications.
**NOTE 2:** Sorry @philbucher, but I did a bit of copypaste in the tests because in the last derivation of the tests class didn't work, and the code to duplicate wasn't that much.

**🆕 Changelog**

- [Adding projection_3D_2D_mapper](https://github.com/KratosMultiphysics/Kratos/commit/d5d0f15744f66613c583af0705eac7990a29f5c6)
- [Register in application](https://github.com/KratosMultiphysics/Kratos/commit/c16aece8873f2960590ee2e4b5db5f190b26e7e3)
- [Register in python](https://github.com/KratosMultiphysics/Kratos/commit/0e4bdfda800299f9f89b45ae5da30025a7d9276a) 
- [Adding tests](https://github.com/KratosMultiphysics/Kratos/commit/4316d5a5516b6335d1de5889ae43b784835649fd)
- [Adding method for inverse mapping parameters](https://github.com/KratosMultiphysics/Kratos/commit/70b5c5ff0e70885af729ffbe17f47b73c82a71e7)
- [Adding access methods](https://github.com/KratosMultiphysics/Kratos/commit/54e2e4bba443f645bad2be6985250a8c5d454226)


philbucher:
ping me once you are done, then i can take a look :)

loumalouomega:
> ping me once you are done, then i can take a look :)

Almost done, there is a bug in the inverse map. I cannot find why fails sometimes

loumalouomega:
> ping me once you are done, then i can take a look :)

Now done

philbucher:
Segfault in MPI tests

Are you sure this works in MPI? You code does not do any global operations 🤔

loumalouomega (commented):
> Segfault in MPI tests
> 
> Are you sure this works in MPI? You code does not do any global operations thinking

I need to take a look at the MPI side, as I work now in Windows by default it takes me extra effort to check this. A priori I am not doing anything that should conflict with MPI... a priori...Okay maybe the auxiliary model parts are crashing between them, I will take a look

philbucher:
> as I work now in Windows

My condolences 😄😆

Well, usually making it work in MPI requires a lot more work. 

Why don't you add the serial version for now and do MPI later?

loumalouomega:
> > as I work now in Windows
> 
> My condolences smilelaughing
> 
> Well, usually making it work in MPI requires a lot more work.
> 
> Why don't you add the serial version for now and do MPI later?

I will try to make it work, I think it will be relatively easy to do

loumalouomega:
@philbucher a priori is fixed, it took me some time because first I needed to compile Trilinos in WSL and that takes a while and second I was trying to minimize the quantity of code to modify.

loumalouomega:
> @philbucher a priori is fixed, it took me some time because first I needed to compile Trilinos in WSL and that takes a while and second I was trying to minimize the quantity of code to modify.

Marvelous, it work with two cores but not three, I will check it

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
```diff
@@ -0,0 +1,673 @@
...
+/**
+ * @brief This function returns the corresponding model part considering the parameters
+ * @param rOriginModelPart The model part to be considered
+ * @param ThisParameters The configuration parameters
+*/
+ModelPart& GetOriginModelPart(
+    ModelPart& rOriginModelPart,
+    Parameters ThisParameters = Parameters(R"({})")
+    )
+{
```
pooyan-dadvand, line 69: I am not sure if I get the logic correctly but as I understand you are assuming that the origin is the 2D one.

If it is correct why we need a new origin_2d keyword?

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
```diff
@@ -0,0 +1,673 @@
...
+    }
+}
+} // namespace Utility
+
+///@}
+///@name Kratos Classes
+///@{
+
+// Definition of projection variables
+struct ProjectionVariables
```
pooyan-dadvand: I would make it a private member struct if is not used in the interface

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
```diff
@@ -0,0 +1,673 @@
...
+{
+    // We get the model part name
+    const std::string origin_2d_sub_model_part_name = ThisParameters.Has("origin_2d_sub_model_part_name") ? ThisParameters["origin_2d_sub_model_part_name"].GetString() : "";
+
+    // We check if the submodelpart exists
+    if (origin_2d_sub_model_part_name != "") {
+        // We check if the submodelpart is the modelpart
+        if (rOriginModelPart.Name() == origin_2d_sub_model_part_name) {
+            return rOriginModelPart;
+        } else { // We retrieve the orginal modelpart if not
```
pooyan-dadvand: original

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
pooyan-dadvand: IMO, we should assume that always one modelpart is 2d and another one is 3d. We can recognize them internally which one is the 2D.

Then we can reuse the same origin and destination modelpart like any other mapper.

pooyan-dadvand (commented):
Without going into detailed review I would take out the mapping of 3d to 3d using reference surface and delaunator.

loumalouomega:
Still work in progress, I am pushing to test MPI in Linux

philbucher:
I will have a look beginning of next week

loumalouomega:
A priori this is fixed and comments applied @pooyan-dadvand . The final code is cleaner and simpler

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
```diff
@@ -0,0 +1,748 @@
...
+{
+    NEAREST_NEIGHBOR,
+    NEAREST_ELEMENT,
+    BARYCENTRIC
+};
+
+/**
+ * @brief Entity type mesh considered
+ */
+enum class EntityTypeMesh
```
pooyan-dadvand: We don't have this one anywhere? BTW, I would give it a context being global enum in entire Kratos.

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
```diff
@@ -0,0 +1,748 @@
...
+///@{
+
+///@}
+///@name  Enum's
+///@{
+
+/**
+ * @brief Type of meta mapper considered
+ */
+enum class MetaMapperType
```
pooyan-dadvand: I would use the string conversion in the context of the factory without using and additional enum which is used only in 1 if.

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
```diff
@@ -0,0 +1,748 @@
...
+    // In MPI could be a bit troublesome
+    if (rFirstModelPart.IsDistributed()) { 
+        const auto& r_communicator = rFirstModelPart.GetCommunicator();
+        const auto& r_data_communicator = r_communicator.GetDataCommunicator();
+        const int mpi_rank = r_data_communicator.Rank();
+        const int mpi_size = r_data_communicator.Size();
+
+        const int partition_entity_1 = DeterminePartitionWithEntities(rFirstModelPart);
+        const int partition_entity_2 = DeterminePartitionWithEntities(rSecondModelPart);
+        KRATOS_ERROR_IF(partition_entity_1 == -1 && partition_entity_2 == -1) << "ERROR:: Both model parts are empty (no elements or not conditions)" << std::endl;
```
pooyan-dadvand: This may happen regulary, no? I mean, in some ranks there might be nothing to map at all.

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
```diff
@@ -0,0 +1,748 @@
...
+
+    ///@}
+    ///@name Inquiry
+    ///@{
+
+    int AreMeshesConforming() const override
+    {
+        KRATOS_WARNING_ONCE("Mapper") << "Developer-warning: \"AreMeshesConforming\" is deprecated and will be removed in the future" << std::endl;
+        return BaseType::mMeshesAreConforming;
+    }
```
philbucher:
```suggestion
```

this is already a PITA for me, please don't add it to new code

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
```diff
@@ -0,0 +1,748 @@
...
+            if (mCopiedParameters.Has("interpolation_type")) mCopiedParameters.RemoveValue("interpolation_type");
+            if (mCopiedParameters.Has("local_coord_tolerance")) mCopiedParameters.RemoveValue("local_coord_tolerance");
+            mpBaseMapper = Kratos::make_unique<NearestNeighborMapperType>(r_origin_model_part, r_destination_model_part, mCopiedParameters);
+        } else if (mMetaMapperType == MetaMapperType::NEAREST_ELEMENT) {
+            if (mCopiedParameters.Has("interpolation_type")) mCopiedParameters.RemoveValue("interpolation_type");
+            mpBaseMapper = Kratos::make_unique<NearestElementMapperType>(r_origin_model_part, r_destination_model_part, mCopiedParameters);
+        } else if (mMetaMapperType == MetaMapperType::BARYCENTRIC) {
+            mpBaseMapper = Kratos::make_unique<BarycentricMapperType>(r_origin_model_part, r_destination_model_part, mCopiedParameters);
+        } else {
+            KRATOS_ERROR << "ERROR:: Mapper " << mCopiedParameters["base_mapper"].GetString() << " is not available as base mapper for projection" << std::endl;
```
philbucher:
```suggestion
            KRATOS_ERROR << "Mapper " << mCopiedParameters["base_mapper"].GetString() << " is not available as base mapper for projection" << std::endl;
```

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
```diff
@@ -0,0 +1,748 @@
...
+    void GetEntityMeshType()
+    {
+        KRATOS_TRY;
+
+        // A priori in the constructor we have already checked that the 2D is not empty
+        const auto& r_2d_model_part = this->Get2DModelPart();
+        if (r_2d_model_part.NumberOfElements() > 0) {
+            mEntityTypeMesh = EntityTypeMesh::ELEMENTS;
+        } else if (r_2d_model_part.NumberOfConditions() > 0) {
+            mEntityTypeMesh = EntityTypeMesh::CONDITIONS;
```
philbucher: Usually Conditions have priority in the mapper, is it possible to change this?

philbucher (commented):
minor comments, overall looks pretty good to me (I did not review/understand the plane-reconstruction algorithm in detail)

One request, please make sure to run the `runscript_test_MappingApplication.py`, which runs the tests with up to 50 processes, to ensure that the MPI version is fully working

philbucher:
BTW codacy is complaining about trailing whitespaces, is it possible that you forgot to set your IDE to remove whitespaces on save?

loumalouomega:
> BTW codacy is complaining about trailing whitespaces, is it possible that you forgot to set your IDE to remove whitespaces on save?

I use VS Code with the extension for trailing whitespaces

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
loumalouomega: Okay

loumalouomega:
> One request, please make sure to run the `runscript_test_MappingApplication.py`, which runs the tests with up to 50 processes, to ensure that the MPI version is fully working

Except this I have take into account your suggestions. I don't know how to run that script and where (which machine)

philbucher:
> > One request, please make sure to run the `runscript_test_MappingApplication.py`, which runs the tests with up to 50 processes, to ensure that the MPI version is fully working
> 
> 
> 
> Except this I have take into account your suggestions. I don't know how to run that script and where (which machine)
> 
> 

python runscript test_MappingApplication.py ;)

loumalouomega:
> > runscript
> 
> Some combinations are failing and I don't know why, because larger number of cores then pass... I will research

@philbucher Okay, actually there are only two tests failing with more than 8 cores. If I skip these two tests it passes:

image

The tests failing are when using `closest_neighbour` when mapping from the coarsest mesh (super coarse). I have tried with more iterations and greater search radius, but looks like this is a corner case, so may be interesting to dig deeper. With my current experience, I cannot fix that, which would imply modify `closest_neighbour`. This may be the problem found by @ddiezrod. If I understand correctly, this is because the initial radius is computed in each partition. I may have the solution to that, but I prefer to take a look at that in an independent PR.

applications/MappingApplication/tests/test_projection_3d_2d_mapper.py
```diff
@@ -0,0 +1,255 @@
...
+        return os.path.join("result_files", self.mapper_type, self.__class__.__name__ + "_" + file_appendix)
+
+    def test_Projection3D2DMapper_Map_non_constant_scalar(self):
+        SetHistoricalNonUniformSolutionScalar(self.model_part_origin.Nodes, KM.PRESSURE)
+        self.mapper.Map(KM.PRESSURE, KM.TEMPERATURE)
+        mapper_test_case.CheckHistoricalNonUniformValues(self.model_part_destination, KM.TEMPERATURE, GetFilePath(self._GetFileName("map_scalar")))
+        #mapper_test_case.VtkOutputNodesHistorical(self.model_part_destination, KM.TEMPERATURE)
+
+    @KratosUnittest.skipUnless(IsDistributedRun() and ParallelEnvironment.GetDefaultSize() <= 7,  "Test designed to be run with max. 7 ranks.")
+    def test_Projection3D2DMapper_InverseMap_non_constant_scalar(self):
```
loumalouomega, line 166: The two tests in question

applications/MappingApplication/tests/test_projection_3d_2d_mapper.py
```diff
@@ -0,0 +1,255 @@
...
+        #mapper_test_case.VtkOutputNodesHistorical(self.model_part_origin, KM.PRESSURE)
+
+    def test_Projection3D2DMapper_Map_non_constant_vector(self):
+        SetHistoricalNonUniformSolutionVector(self.model_part_origin.Nodes, KM.FORCE)
+        self.mapper.Map(KM.FORCE, KM.VELOCITY)
+        mapper_test_case.CheckHistoricalNonUniformValues(self.model_part_destination, KM.VELOCITY, GetFilePath(self._GetFileName("map_vector")))
+        #mapper_test_case.VtkOutputNodesHistorical(self.model_part_destination, KM.VELOCITY)
+
+    @KratosUnittest.skipUnless(IsDistributedRun() and ParallelEnvironment.GetDefaultSize() <= 7,  "Test designed to be run with max. 7 ranks.")
+    def test_Projection3D2DMapper_InverseMap_non_constant_vector(self):
```
loumalouomega, line 179: And here

loumalouomega:
> > > runscript
> > 
> > 
> > Some combinations are failing and I don't know why, because larger number of cores then pass... I will research
> 
> @philbucher Okay, actually there are only two tests failing with more than 8 cores. If I skip these two tests it passes:
> 
> image
> 
> The tests failing are when using `closest_neighbour` when mapping from the coarsest mesh (super coarse). I have tried with more iterations and greater search radius, but looks like this is a corner case, so may be interesting to dig deeper. With my current experience, I cannot fix that, which would imply modify `closest_neighbour`. This may be the problem found by @ddiezrod. If I understand correctly, this is because the initial radius is computed in each partition. I may have the solution to that, but I prefer to take a look at that in an independent PR.

I found something funny, the coarse mesh (the problematic one) is only 6 elements. Seeing this mesh after partition, there are partitions without elements but with nodes. I assume this should not happen and may be a potential cause of the problem.

loumalouomega:
@philbucher I am stuck looking for the bug in the mapping, any suggestion where the problem could be?

philbucher:
> @philbucher I am stuck looking for the bug in the mapping, any suggestion where the problem could be?

I can try, but can you please explain what exactly the MPI part does?

I didn't really get what you do there

In serial everything works fine right?

loumalouomega:
> > @philbucher I am stuck looking for the bug in the mapping, any suggestion where the problem could be?
> 
> I can try, but can you please explain what exactly the MPI part does?
> 
> I didn't really get what you do there
> 
> In serial everything works fine right?

There are two tests that are failing when CPU number in MPI > 8. This is the case of the coarsest mesh. In my research looks like is due to the search step. As I am understanding the whole mechanisms of the mapping app on the way, I am still not sure about the problem. I don't know if related with @ddiezrod  problem.

applications/MappingApplication/tests/test_projection_3d_2d_mapper.py
loumalouomega: @philbucher the test

applications/MappingApplication/tests/test_projection_3d_2d_mapper.py
loumalouomega: @philbucher and here (is the same as the previous one but with an array)

philbucher:
The problem of @ddiezrod is not really related I think 

You can check by setting yourself a large search radius 

I mean that I don't understand your algorithm thus it is difficult to help

If it does not work (only different values or segfault) with more cores it might be a design problem

loumalouomega:
> You can check by setting yourself a large search radius

Already tried that and not helping

> If it does not work (only different values or segfault) with more cores it might be a design problem

Takes instead of the values of one node, another different node

philbucher:
Then I think sth is wrong with your algorithm 🤔 

Have you checked that you get the same results from your algorithm? I mean before the other mappers are called

loumalouomega:
> Then I think sth is wrong with your algorithm 🤔

It only fails in that specific coarse case with nearest neighbour....

> Have you checked that you get the same results from your algorithm? I mean before the other mappers are called

Yes, I already checked...

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
```diff
@@ -0,0 +1,743 @@
...
+/**
+ * @brief This method determines the partition where the model part has at least one entity
+ * @param rModelPart The input model part where determine the partition with at least one entity
+ * @return A partition with at least one partition
+ */
+int DeterminePartitionWithEntities(const ModelPart& rModelPart)
+{
+    const bool is_distributed = rModelPart.IsDistributed();
+    Geometry<Node<3>>::Pointer p_geometry = rModelPart.NumberOfElements() > 0 ? rModelPart.ElementsBegin()->pGetGeometry() : rModelPart.NumberOfConditions() > 0 ? rModelPart.ConditionsBegin()->pGetGeometry() : nullptr;
+
```
pooyan-dadvand, line 57: I think an alternative way of doing this:
```c++
local_dimension = (p_geometry == nullptr) ? 0 : p_geometry->LocalDimension()
max_local_dimension = r_data_communicator.Max(local_dimension)
this_partition_is_2d = (max_local_dimension == 2);
```

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
```diff
@@ -0,0 +1,743 @@
...
+}
+
+/**
+ * @brief This method determines the 2D model part
+ * @param rFirstModelPart The first ModelPart
+ * @param rSecondModelPart The second ModelPart
+ * @return The 2D model part 
+ */
+ModelPart& Determine2DModelPart(ModelPart& rFirstModelPart, ModelPart& rSecondModelPart)
+{
```
pooyan-dadvand, line 129: Please check my comment above to replace this implemenation

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
```diff
@@ -0,0 +1,743 @@
...
+/**
+ * @brief This method determines the 3D model part
+ * @details The counter part of the previous method
+ * @param rFirstModelPart The first ModelPart
+ * @param rSecondModelPart The second ModelPart
+ * @return The 3D model part 
+ */
+ModelPart& Determine3DModelPart(ModelPart& rFirstModelPart, ModelPart& rSecondModelPart)
+{
+    // It is the counter part of the previous method
```
pooyan-dadvand: Please check my comment above

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
```diff
@@ -0,0 +1,743 @@
...
+                normal_check(array_1d<double, 3>& rNormal) : reference_normal(rNormal) {};
+                array_1d<double, 3> reference_normal;
+                GeometryType::CoordinatesArrayType aux_coords;
+            };
+            if (mEntityTypeMesh == EntityTypeMesh::CONDITIONS) {
+                check_normal = block_for_each<SumReduction<std::size_t>>(r_2d_model_part.Conditions(), normal_check(mNormalPlane), [&numerical_limit](auto& r_cond, normal_check& nc) {
+                    auto& r_geom = r_cond.GetGeometry();
+                    r_geom.PointLocalCoordinates(nc.aux_coords, r_geom.Center());
+                    const auto normal = r_geom.UnitNormal(nc.aux_coords);
+                    if (norm_2(normal - nc.reference_normal) > numerical_limit) { return 1; } else { return 0; }
```
pooyan-dadvand: `return (norm_2(normal - nc.reference_normal) > numerical_limit)`

pooyan-dadvand (commented):
Some minor comments and improvements

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
loumalouomega: Okay, moving inside class

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
loumalouomega: Refactoring

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
loumalouomega: But if you check `DeterminePartitionWithEntities` function, you will see that there is a check between all partitions. I copy paste the function so is easier to see:

~~~c++
/**
 * @brief This method determines the partition where the model part has at least one entity
 * @param rModelPart The input model part where determine the partition with at least one entity
 * @return A partition with at least one partition
 */
int DeterminePartitionWithEntities(const ModelPart& rModelPart)
{
    const bool is_distributed = rModelPart.IsDistributed();
    Geometry<Node<3>>::Pointer p_geometry = rModelPart.NumberOfElements() > 0 ? rModelPart.ElementsBegin()->pGetGeometry() : rModelPart.NumberOfConditions() > 0 ? rModelPart.ConditionsBegin()->pGetGeometry() : nullptr;

    // MPI data
    int aux_partition_entity = -1;
    const auto& r_communicator = rModelPart.GetCommunicator();
    const auto& r_data_communicator = r_communicator.GetDataCommunicator();
    const int mpi_rank = r_data_communicator.Rank();
    const int mpi_size = r_data_communicator.Size();

    // Get maximum rank
    if (p_geometry != nullptr) aux_partition_entity = mpi_rank;

    // Determine root's rank
    const int root_rank = 0;

    // Getting the partition with entities index
    int partition_entity = r_data_communicator.Max(aux_partition_entity, root_rank);

    // Define the send tag
    const int tag_send_index = 1;

    // Transfer to other partitions
    if (is_distributed) {
        if (mpi_rank == root_rank) {
            for (int i_rank = 1; i_rank < mpi_size; ++i_rank) {
                r_data_communicator.Send(partition_entity, i_rank, tag_send_index);
            }
        } else {
            r_data_communicator.Recv(partition_entity, root_rank, tag_send_index);
        }
    }

    return partition_entity;
}
~~~

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
loumalouomega: Let me see

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
loumalouomega: True

applications/MappingApplication/custom_mappers/projection_3D_2D_mapper.h
loumalouomega: Done

loumalouomega:
Code duplication reduced

pooyan-dadvand (approved):
Thanks for the refactory! Becomes much better!

PR #10488, opened 2022-11-24
[Core] KRATOS_WATCH_LINE

matekelemen:
- Add a handy utility macro for old-fashioned print-debugging, that prints the code location and function signature along with your message. *Super useful for MPI debugging!*
- Change `CodeLocation` such that its output is followable (Ctrl+LMB click => jump to location) in VSCode's embedded terminal.

### DIFF
--- a/kratos/includes/debug_helpers.h
+++ b/kratos/includes/debug_helpers.h
@@ -51,4 +51,7 @@
                   << values.str() << std::endl;                         \
     }
 
+// Cout a message with appended line and function information.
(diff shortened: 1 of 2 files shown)


roigcarlo (approved):

rubenzorrilla:
Just to make it clear, this is like a combo of current `KRATOS_WATCH` plus `KRATOS_WATCH(__LINE__)` right?

matekelemen:
> Just to make it clear, this is like a combo of current `KRATOS_WATCH` plus `KRATOS_WATCH(__LINE__)` right?

Yup, it's exactly that (+ some extra stuff you may want to prepend it with).

PR #10489, opened 2022-11-24
[HDF5] Fix some random HDF5 deadlocks

matekelemen:
## Changes
Sort variable names before executing IO operations.

## Reason
For any given HDF5 IO operation, the set of variable names is expected to be identical on all ranks, but their order might not always be the same. This can happen if the variable list was not provided by the user but is the result of some rank-local scanning and a related MPI synchronization, which guarantees that every item is on every rank, but makes no guarantees about their order.

As a result, it might happen that different ranks are trying to write different variables at the same time, resulting in a deadlock. Sorting the incoming name list solves this issue.

### DIFF
--- a/applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
+++ b/applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
@@ -386,6 +386,8 @@ ContainerComponentIO<TContainerType, TContainerItemType, TComponents...>::Contai
     for (std::size_t i = 0; i < num_components; ++i)
         mComponentNames[i] = Settings["list_of_variables"].GetArrayItem(i).GetString();
 
+    std::sort(mComponentNames.begin(), mComponentNames.end());
+
     KRATOS_CATCH("");
 }
 
--- a/applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
+++ b/applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
@@ -377,14 +377,7 @@ ContainerComponentIO<TContainerType, TContainerItemType, TComponents...>::Contai
     Settings.ValidateAndAssignDefaults(default_params);
 
     mComponentPath = Settings["prefix"].GetString() + rComponentPath;
-
-    const std::size_t num_components = Settings["list_of_variables"].size();
-
-    if (mComponentNames.size() != num_components)
-        mComponentNames.resize(num_components);
-
-    for (std::size_t i = 0; i < num_components; ++i)
-        mComponentNames[i] = Settings["list_of_variables"].GetArrayItem(i).GetString();
+    mComponentNames = Settings["list_of_variables"].GetStringArray();
 
     // Sort component names to make sure they're in the same order on each rank.
     // The basic assumption is that the set of components is identical on every
--- a/applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
+++ b/applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
@@ -386,6 +386,11 @@ ContainerComponentIO<TContainerType, TContainerItemType, TComponents...>::Contai
     for (std::size_t i = 0; i < num_components; ++i)
         mComponentNames[i] = Settings["list_of_variables"].GetArrayItem(i).GetString();
 
+    // Sort component names to make sure they're in the same order on each rank.
+    // The basic assumption is that the set of components is identical on every
+    // rank, but they may not be in the same order (which would lead to ranks
+    // trying to write different variables at the same time, resulting in
+    // a deadlock), hence the sorting.
     std::sort(mComponentNames.begin(), mComponentNames.end());
 
     KRATOS_CATCH("");
--- a/applications/HDF5Application/custom_io/hdf5_nodal_solution_step_data_io.cpp
+++ b/applications/HDF5Application/custom_io/hdf5_nodal_solution_step_data_io.cpp
@@ -6,6 +6,7 @@
 #include "includes/communicator.h"
 #include "custom_utilities/registered_component_lookup.h"
 #include "custom_utilities/local_ghost_splitting_utility.h"
+#include "includes/debug_helpers.h"
 
 namespace Kratos
 {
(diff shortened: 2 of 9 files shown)


philbucher:
How did you discover/debug this?

I also had sometimes that it would hang, but while writing the mesh, maybe there is also a bug there

Other than that great job 👏🏻

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
```diff
@@ -386,6 +386,8 @@ ContainerComponentIO<TContainerType, TContainerItemType, TComponents...>::Contai
     for (std::size_t i = 0; i < num_components; ++i)
         mComponentNames[i] = Settings["list_of_variables"].GetArrayItem(i).GetString();
```
philbucher: Here you could refactor with GetStringArray

philbucher (dismissed):
Minor suggestion for improvement

applications/HDF5Application/custom_io/hdf5_nodal_solution_step_data_io.cpp
```diff
@@ -6,6 +6,7 @@
 #include "includes/communicator.h"
 #include "custom_utilities/registered_component_lookup.h"
 #include "custom_utilities/local_ghost_splitting_utility.h"
+#include "includes/debug_helpers.h"
```
sunethwarna: this canbe removed ryt?

sunethwarna (dismissed):
Nice :)

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
```diff
@@ -386,6 +386,8 @@ ContainerComponentIO<TContainerType, TContainerItemType, TComponents...>::Contai
...
 
+    std::sort(mComponentNames.begin(), mComponentNames.end());
```
sunethwarna, line 387: May be itisa good idea to put a comment why the sorting is necessary.

matekelemen:
> How did you discover/debug this?

5 hours of print-debugging with @sunethwarna ... 😓
I was (un)lucky enough to write a tiny test case that hung pretty reliably about ~50% of the time.

Debugging it with the usual separate-terminal-for-each-rank approach wouldn't work because each process had to be launched manually, which guaranteed a consistent execution order and made the bug impossible to reproduce. If you have some good tips on how to deal with MPI a bit more comfortably, please let me know, I'm begging you 🥲

> I also had sometimes that it would hang, but while writing the mesh, maybe there is also a bug there

I hope this fixes that too, but I can't guarantee that this bug was responsible for all HDF5 deadlocks :(
If it still happens with this fix, and you manage to narrow it down to a reasonably sized test case, send it my way too plz!

#

P.S.: I just realized that I should:
- add some comments why sorting is necessary
- add a test case that deliberately passes variable names in mismatched orders

so I'll add them tomorrow.

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
matekelemen: Yup, I just realized that too! + test case

philbucher:
> If you have some good tips on how to deal with MPI a bit more comfortably, please let me know

Sadly I do not. I contemplated many times why I did not do research on locking mechanisms of a very specialized shell element, where I can implement everything in my matlab code and never deal with these issues :smiling_face_with_tear: 

But more seriously, it is just very painful. I still have not found _the_ way for debugging MPI. I never looked into debuggers really, because you can make it working locally and then it crashes on the cluster - back to square zero. There are some solutions for this I think too, but nothing that was easy to use for me.

What helped me:
- Experience! I faced so many problems over the years, that at some point I had an idea what was happening. This is kinda why I started following all the developments in Kratos closely to understand what is happening.
- Being super pedantic when developing any kind of code, in serial and MPI. Having something that barely works in serial will certainly fail on the cluster ;)
- Algorithms in MPI are just completely different and require much thinking, especially if they have to be efficient. I needed years to fully get into it and understand it. Therefore start small :)
- Well, try to test soon (on the cluster!) and with representative examples.
- Don't assume that existing things do what you need. Many Kratos devs have no idea about MPI, therefore test it yourself if you want to be sure. This PR is a nice example :) Was developed by very knowledgeable devs, but still there are some bugs here and there

In a nutshell: know what you are doing :see_no_evil: 
Not really good advice, I know, but the more familiar you are with everything the easier also other things become :man_shrugging: 

Don't loose faith, it is a difficult and hard topic!

loumalouomega:
> > If you have some good tips on how to deal with MPI a bit more comfortably, please let me know
> 
> Sadly I do not. I contemplated many times why I did not do research on locking mechanisms of a very specialized shell element, where I can implement everything in my matlab code and never deal with these issues 🥲
> 
> But more seriously, it is just very painful. I still have not found _the_ way for debugging MPI. I never looked into debuggers really, because you can make it working locally and then it crashes on the cluster - back to square zero. There are some solutions for this I think too, but nothing that was easy to use for me.
> 
> What helped me:
> 
>     * Experience! I faced so many problems over the years, that at some point I had an idea what was happening. This is kinda why I started following all the developments in Kratos closely to understand what is happening.
> 
>     * Being super pedantic when developing any kind of code, in serial and MPI. Having something that barely works in serial will certainly fail on the cluster ;)
> 
>     * Algorithms in MPI are just completely different and require much thinking, especially if they have to be efficient. I needed years to fully get into it and understand it. Therefore start small :)
> 
>     * Well, try to test soon (on the cluster!) and with representative examples.
> 
>     * Don't assume that existing things do what you need. Many Kratos devs have no idea about MPI, therefore test it yourself if you want to be sure. This PR is a nice example :) Was developed by very knowledgeable devs, but still there are some bugs here and there
> 
> 
> In a nutshell: know what you are doing 🙈 Not really good advice, I know, but the more familiar you are with everything the easier also other things become 🤷‍♂️
> 
> Don't loose faith, it is a difficult and hard topic!

matekelemen:
@philbucher @sunethwarna Sorry for the wait, can you please take a quick look again?

For the record, `TestHDF5NodalDataValueIO_WriteNodalResults1` from `applications/HDF5Application/tests/test_hdf5_nodal_data_value_io.cpp` segfaults on MPI runs on up-to-date Manjaro systems (on master too). I haven't found the source yet, but it's pretty scary that the CI passes just fine, which can mean a couple of things:
- newer Manjaro packages are responsible => sooner or later, we'll have to find the segfault anyway because other distros will eventually get updated too
- the test is not run in the CI
- the test is not run in MPI in the CI

sunethwarna (approved):
I am ok with the changes. thanks @matekelemen :)

PR #10492, opened 2022-11-25
[CLApp] Adding the capability to add prestress loads in the Serial Parallel Rule of Mixtures

AlejandroCornejo:
**📝 Description**
In this PR I am adding the capability of adding a prestressing load in composite materials (pre-stressed concrete) by means of the SP-RoM as we did in https://www.sciencedirect.com/science/article/abs/pii/S0263822318301144 in another code.

Basically checking only once if the geometry `Has(SP_PRESTRESS)` the pre-stressing load and the adding it to the system.

The behaviour of the CL is not affected. The `SP_PRESTRESS` has to be defined at the geometry, for example by adding:

~~~txt
Begin ElementalData SP_PRESTRESS
1    1.0e-5
2    1.0e-5
3    1.0e-5
4    1.0e-5
End ElementalData
~~~
in the mdpa file


AlejandroCornejo:
@RiccardoRossi we are getting more and more closer to the capabilities that we had in PLCd...

applications/ConstitutiveLawsApplication/constitutive_laws_application.cpp
```diff
@@ -365,6 +365,7 @@ void KratosConstitutiveLawsApplication::Register()
     KRATOS_REGISTER_VARIABLE(ADVANCE_STRATEGY_APPLIED);
 
     // Constitutive laws variables
+    KRATOS_REGISTER_VARIABLE(SP_PRESTRESS)
```
loumalouomega: Can you use a more descriptive name SERIAL_PARALLEL_PRESTRESS?

applications/ConstitutiveLawsApplication/custom_constitutive/composites/serial_parallel_rule_of_mixtures_law.cpp
```diff
@@ -532,9 +534,18 @@ void SerialParallelRuleOfMixturesLaw::CalculateStrainsOnEachComponent(
...
     noalias(rStrainVectorMatrix) = prod(rParallelProjector, r_total_parallel_strain_vector) + prod(trans(rSerialProjector), rSerialStrainMatrix);
-    noalias(rStrainVectorFiber)  = prod(rParallelProjector, r_total_parallel_strain_vector) + prod(trans(rSerialProjector), (1.0 / kf * r_total_serial_strain_vector) - (km / kf * rSerialStrainMatrix));
+    if (mIsPrestressed) {
+        Vector aux(1);
+        aux[0] = rValues.GetElementGeometry().GetValue(SP_PRESTRESS);
+        if (Iteration > 0)
+            aux[0] += r_total_parallel_strain_vector[0];
+        noalias(rStrainVectorFiber)  = prod(rParallelProjector, aux) + prod(trans(rSerialProjector), (1.0 / kf * r_total_serial_strain_vector) - (km / kf * rSerialStrainMatrix));
+    } else {
+        noalias(rStrainVectorFiber)  = prod(rParallelProjector, r_total_parallel_strain_vector) + prod(trans(rSerialProjector), (1.0 / kf * r_total_serial_strain_vector) - (km / kf * rSerialStrainMatrix));
```
loumalouomega, line 547: I assume kf is never zero?

loumalouomega (changes requested):
Change variable name please

applications/ConstitutiveLawsApplication/custom_constitutive/composites/serial_parallel_rule_of_mixtures_law.cpp
AlejandroCornejo: correct

loumalouomega (approved):

PR #10495, opened 2022-11-26
[Metis] improve InMemoryPartitioner to remove MPI dependency

philbucher:
Until recently, the `MetisDivideHeterogeneousInputInMemoryProcess` was one of the remaining strongholds in Kratos, a mighty beast of manual memory allocation, lots of code duplication and an unnecessary MPICore dependency. 
Then I came along and cut it down to a mere 60 line long function :D

Jokes aside, this is a pretty cool cleanup! It includes:
- removing the MPICore dependency of the MetisApp! This is achieved by using the `DataCommunicator` directly
- It removes every manual memory allocation (I think there were some leaks there which are now removed)
- The `Scatterv` interface of the `DataCommunicator` was extended for `char`, which is super simple thanks to the well done design
- Overall the process was simplified, it now contains basically only the data exchange with the other partitions


kratos/includes/data_communicator.h
```diff
@@ -360,6 +359,7 @@ class KRATOS_API(KRATOS_CORE) DataCommunicator
     KRATOS_BASE_DATA_COMMUNICATOR_DECLARE_PUBLIC_INTERFACE_FOR_TYPE(unsigned int)
     KRATOS_BASE_DATA_COMMUNICATOR_DECLARE_PUBLIC_INTERFACE_FOR_TYPE(long unsigned int)
     KRATOS_BASE_DATA_COMMUNICATOR_DECLARE_PUBLIC_INTERFACE_FOR_TYPE(double)
+    KRATOS_BASE_DATA_COMMUNICATOR_DECLARE_SCATTER_INTERFACE_FOR_TYPE(char)
```
loumalouomega, line 362: You can do an independent PR just for this

loumalouomega (commented):
Very nice

loumalouomega:
> Very nice

Pure C people won't like to replace char ** with std::vector<std::vector<char>>

loumalouomega:
> > Very nice
> 
> Pure C people won't like to replace char ** with std::vector<std::vector>

image

kratos/mpi/sources/mpi_data_communicator.cpp
```diff
@@ -265,6 +265,7 @@ KRATOS_MPI_DATA_COMMUNICATOR_DEFINE_PUBLIC_INTERFACE_FOR_TYPE(int)
 KRATOS_MPI_DATA_COMMUNICATOR_DEFINE_PUBLIC_INTERFACE_FOR_TYPE(unsigned int)
 KRATOS_MPI_DATA_COMMUNICATOR_DEFINE_PUBLIC_INTERFACE_FOR_TYPE(long unsigned int)
 KRATOS_MPI_DATA_COMMUNICATOR_DEFINE_PUBLIC_INTERFACE_FOR_TYPE(double)
+KRATOS_MPI_DATA_COMMUNICATOR_DEFINE_SCATTER_INTERFACE_FOR_TYPE(char)
```
roigcarlo, line 268: Wouldn't be better to just define the full interface for chars?

roigcarlo (approved):
Aside from a minor comment OK. 

Approving anyway,

kratos/mpi/sources/mpi_data_communicator.cpp
philbucher: Makes sense, will do in a follow up PR

It will anyway not be the same as e.g. max of a char does not make sense

PR #10496, opened 2022-11-26
[Core][MPI] const_cast no longer needed with C++17

philbucher:
see title
C++17 added the non-const version of `data` for `std::string`


sunethwarna (approved):
Thanks @philbucher

PR #10497, opened 2022-11-26
[Metis] remove unused inheritance and old code

philbucher:
This removes the inheritance of `MetisDivideHeterogeneousInputProcess` from `MetisDivideInputToPartitionsProcess`, as it is not used at all.

In fact, `MetisDivideInputToPartitionsProcess` is not working since we moved to use Metis 5 (which happened years ago), and is hence removed, together with the also broken `MetisGraphPartitioningProcess`


rubenzorrilla (approved):
Makes sense for me. Thanks.

PR #10500, opened 2022-11-27
[Core][IO] WriteModelPart - correcting const

philbucher:
The `WriteModelPart` of `IO` takes a non-const `ModelPart` as input, which is wrong and should be corrected. I added a new overload which takes it as `const`. It is fully backward compatible, same principle as in #7381 is applied, which has proven to work very well

This way the derived classes can be adapted one-by-one

I also fixed the deprecation start/stop macros for the intel compiler


loumalouomega (approved):
Even more thanks

rubenzorrilla (approved):
Makes perfect sense for me. Thanks!

PR #10445, opened 2022-11-14
[All][KratosCore] Deprecate Dimension()

rubenzorrilla:
**📝 Description**
Deprecating the method `Dimension` in `Geometry`, `GeometryData` and `GeometryDimension` in favor of the`WorkingSpaceDimension` and `LocalSpaceDimension` ones. This is the first step in the roadmap established in #9369.

PR #10501, opened 2022-11-28
[Core] Dimension() deprecation warnings

rubenzorrilla:
**📝 Description**
This PR amends the deprecation warnings in Kratos core arising from #10445. Besides, I added the link to the corresponding PR in the `CMakeLists.txt`.


loumalouomega (dismissed):

rubenzorrilla:
Discussion in  #10503 must be solved before continuing forward with this one.

tteschemacher:
My issue is the following I have geometries that are embedded within others. That would mean that those are actually containing a global space which is 3D. Then they are containing a local space which actually may be a surface that is 2D. Then I have geometries that are in 1D or maybe 0D. This is in fact very nice to contain this as then e.g. loads or supports can very easy be unified.

My proposal would be that we get now rid of Dimension, now. Then we can decide if we will need something similar again, but something which will be easier to understand. Probably like an enum naming the entities like solid, surface, curve, and point.

roigcarlo:
We cannot "delete" the function now because we would break other people code, but the idea is getting rid of it yes.

roigcarlo (approved):

PR #10503, opened 2022-11-28
[Core] Curves on surface LocalSpaceDimension

rubenzorrilla:
**📝 Description**
This fixes the CI problems in #10501. Considering that this might be a delicate change, I preferred to isolated from the rest. As far as I understand, the `LocalSpaceDimension` of the curves on surfaces is one, but of course I might be wrong.


rubenzorrilla:
Just mentioning that after having discussed this briefly with @tteschemacher it might not be as straightforward.

tteschemacher (approved):
I do not want to block this actually. If this is needed for a general advantage I guess we should go for it. It will be a problem later though.

rubenzorrilla:
After having discussed it again with @tteschemacher, we both think that we can go ahead with this.

If we eventually need an extra dimension for the IGA geometries we can add a new method in the base geometry, which will throw an error and a very self-descriptive name (not as `Dimension`), to ONLY override it in the IGA geometries.

PR #10511, opened 2022-11-29
[MPM] Newtonian fluid irreducible

lauramoremar:
**📝 Description**
Newtonian fluid constitutive law for irreducible formulation added.
For the UP element (in general) the possibility of unable the stabilization has been added with a new flag.

**🆕 Changelog**
- New Newtonian fluid constitutive law.
- Unit test associated to this new law.
- Possibility of choose stabilize or not the up formulation.


applications/ParticleMechanicsApplication/custom_elements/updated_lagrangian_UP.cpp
```diff
@@ -504,23 +505,6 @@ double& UpdatedLagrangianUP::CalculatePUCoefficient(double& rCoefficient, Genera
 }
 
 
-//************************************************************************************
-//************************************************************************************
-
-double& UpdatedLagrangianUP::CalculatePUDeltaCoefficient(double &rDeltaCoefficient, GeneralVariables & rVariables)
```
andinoboerst: Maybe this can be useful in case the PU Delta Coefficient is to be changed in the future?
Not sure exactly what the PU Delta Coefficient is used for but I guess there are cases where it is not 1?
If it is definitely always 1 then I approve

lauramoremar: Hi! Yes, it's true. Although, in all the cases performed with this type of stabilization this parameter has been always =1.  I think that this stabilization is not very "robust" for doing an study in the future changing this parameter, and for that I deleted it... But I don't know, may be it's better leave it as it was, just in case

andinoboerst: Ok, up to you then. If you think it will stay 1 then it will be better to clean up the code. Otherwise, I would maybe leave it for now.

andinoboerst (dismissed):
Looks good to me, considering the discussion with the PU Delta Coefficient.

lauramoremar:
Thank you @andinoboerst . Finally I decided leave the CalculatePUDeltaCoefficient in the code for now. Could you reapproved the pull request? :)

andinoboerst (approved):
Ok sounds good, approved! :)

PR #10518, opened 2022-12-01
Embedded modeler

miguelmaso:
**📝 Description**
Just adding a handy tool for embedded simulations. I've added it to the core because it's importing a core process.

Maybe, in the future, modelers could implement and `UpdateModelPart` or `UpdateGeometry`. But it should be discussed in a separate issue.

**🆕 Changelog**
- Add embedded modeler

### DIFF
--- a/kratos/python_scripts/modelers/embedded_geometry_modeler.py
+++ b/kratos/python_scripts/modelers/embedded_geometry_modeler.py
@@ -1,0 +1,56 @@
+import KratosMultiphysics as KM
+
+class EmbeddedGeometryModeler(KM.Modeler):
+
+    def __init__(self, model, settings):
+        super().__init__(model, settings)
+
+        # Cannot validate as settings may differ ammong input types
+        settings.AddMissingParameters(self.__GetDefaultSettings())
+
+        # Declare required member variables
+        self.model_part = None
+        self.settings = settings
+
+        # Get the model parts
+        self.volume_model_part = model.CreateModelPart(settings["volume_model_part_name"].GetString())
+        self.embedded_model_part = model.CreateModelPart(settings["embedded_model_part_name"].GetString())
+
+    def SetupGeometryModel(self):
+        super().SetupGeometryModel()
+
+    def PrepareGeometryModel(self):
+        super().PrepareGeometryModel()
+
+    def SetupModelPart(self):
+        super().SetupModelPart()
+
+        # Compute the level-set function
+        domain_size = self.volume_model_part.ProcessInfo[KM.DOMAIN_SIZE]
+        if domain_size == 2:
+            if self.settings["discontinuous_distance"].GetBool():
+                KM.CalculateDiscontinuousDistanceToSkinProcess2D(self.volume_model_part, self.embedded_model_part).Execute()
+            else:
+                KM.CalculateDistanceToSkinProcess2D(self.volume_model_part, self.embedded_model_part).Execute()
+        elif domain_size == 3:
+            if self.settings["discontinuous_distance"].GetBool():
+                KM.CalculateDiscontinuousDistanceToSkinProcess3D(self.volume_model_part, self.embedded_model_part).Execute()
+            else:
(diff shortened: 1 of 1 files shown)


kratos/python_scripts/modelers/embedded_geometry_modeler.py
```diff
@@ -0,0 +1,56 @@
...
+class EmbeddedGeometryModeler(KM.Modeler):
+
+    def __init__(self, model, settings):
+        super().__init__(model, settings)
+
+        # Cannot validate as settings may differ ammong input types
+        settings.AddMissingParameters(self.__GetDefaultSettings())
+
+        # Declare required member variables
+        self.model_part = None
```
sunethwarna, line 12: Why do you need this?

kratos/python_scripts/modelers/embedded_geometry_modeler.py
```diff
@@ -0,0 +1,56 @@
...
+                KM.CalculateDistanceToSkinProcess3D(self.volume_model_part, self.embedded_model_part).Execute()
+        else:
+            raise Exception(f"Unknown DOMAIN_SIZE = {domain_size}")
+
+        KM.Logger.PrintInfo("EmbeddedGeometryModeler", "Finished computing distance.")
+
+    @classmethod
+    def __GetDefaultSettings(cls):
+        default_settings = KM.Parameters('''{
+            "echo_level"               : 0,
```
sunethwarna, line 49: The `echo_level` is not used ryt?

sunethwarna (commented):
This is only useful in the case where the initial distance is not changing throughout the simulation ryt? Wouldn't it be better if we can have this as a `Process` (where we can define where this needs to be execute either in `ExecuteInitialize`, `ExecuteInitializeSolutionStep`....) or `Operation`? @rubenzorrilla

kratos/python_scripts/modelers/embedded_geometry_modeler.py
```diff
@@ -0,0 +1,56 @@
...
+        # Cannot validate as settings may differ ammong input types
+        settings.AddMissingParameters(self.__GetDefaultSettings())
+
+        # Declare required member variables
+        self.model_part = None
+        self.settings = settings
+
+        # Get the model parts
+        self.volume_model_part = model.CreateModelPart(settings["volume_model_part_name"].GetString())
+        self.embedded_model_part = model.CreateModelPart(settings["embedded_model_part_name"].GetString())
```
rubenzorrilla:
```suggestion
        self.model_part = model.CreateModelPart(settings["model_part_name"].GetString())
        self.skin_model_part = model.CreateModelPart(settings["skin_model_part_name"].GetString())

```
To be consistent to the usual notation in Kratos.

kratos/python_scripts/modelers/embedded_geometry_modeler.py
```diff
@@ -0,0 +1,56 @@
...
+    @classmethod
+    def __GetDefaultSettings(cls):
+        default_settings = KM.Parameters('''{
+            "echo_level"               : 0,
+            "volume_model_part_name"   : "",
+            "embedded_model_part_name" : "",
+            "discontinuous_distance"   : false
+        }''')
+        return default_settings
+
```
rubenzorrilla, line 55:
```suggestion

#TODO: Remove this once the modelers use the new model-parameters factory
```

rubenzorrilla:
> This is only useful in the case where the initial distance is not changing throughout the simulation ryt? Wouldn't it be better if we can have this as a `Process` (where we can define where this needs to be execute either in `ExecuteInitialize`, `ExecuteInitializeSolutionStep`....) or `Operation`? @rubenzorrilla

Well, it would be more flexible, but I think this would add an extra complexity for a very rare use case. Note that the distance only changes if you're doing FSI. If it is a rigid body movement, you'd still need to specify it it some place. If FSI coupling, you'd need to do it at the coupling level. This contribution makes very much sense for rapid pre-processing or "simple" unfitted simulations.

rubenzorrilla (changes requested):
Testing is missing.

PR #7401, opened 2020-09-01
[Core] Line/triangle IsInside removing projection

philbucher:
**Description**
This PR removes the projection that is happening in the `IsInside` function of `Line2D2` and `Triangle3D3`
Reasons:
- Inconsistent with other geometries, there only the basic check is done
- Projection is sort of an informative/safety feature, but it punishes users that use the function correctly as then projection is unnecessary!
- Affects the shared memory parallelism very much. Even for my simple example as shown the performance takes a serious hit!!! I don't even want to think what can happen in bigger examples. I think this is manly related to the Logger macros that have a shared counter (IIRC @roigcarlo)

results for 1e7 executions **with** the projection
|   |Line (time [s]) | Triangle (time [s]) |
|---|---|---|
| serial | 0.251562 | 0.709768 |
| parallel (2 threads) | 0.432373 | 0.578865 |
| parallel (3 threads) | 0.333716 | 0.453672 |
| parallel (4 threads) | 0.323167 | 0.39459 |

results for 1e7 executions **without** the projection (here it pretty much scales linear, as expected)
|   |Line (time [s]) | Triangle (time [s]) |
|---|---|---|
| serial | 0.134439 | 0.424116 |
| parallel (2 threads) | 0.0565617 | 0.212631 |
| parallel (3 threads) | 0.0381214 | 0.144309 |
| parallel (4 threads) | 0.0295059 | 0.110287 |

PR #10520, opened 2022-12-01
[Core] Missing distance check in Line3D2 `IsInside`

loumalouomega:
**📝 Description**

Missing distance check in Line3D2 `IsInside`. With this the `IsInside` could give false positive. Consistent with 2D line https://github.com/KratosMultiphysics/Kratos/blob/5c587c1a8f9c6f4adf7502b960709572f0345156/kratos/geometries/line_2d_2.h#L950

Tests are passing

**🆕 Changelog**

- [Missing check in Line3D2 LocalCoordinates](https://github.com/KratosMultiphysics/Kratos/commit/6e3eb8be4b5265be07f32cddee93e75e633ebf34)
- [Moving to IsInside (consistent with 2D line)](https://github.com/KratosMultiphysics/Kratos/commit/faa711a3d5d0defc0c49bc7001bac5c6336badf8)
- [Minor clean upo line 2D (fix typo + pragma once)](https://github.com/KratosMultiphysics/Kratos/commit/aac99a30cc58a5090a43c91988c661b5438e26de)


philbucher (changes requested):
this is conflicting with #7401, lets discuss first what should be the desired behavior

loumalouomega:
> this is conflicting with #7401, lets discuss first what should be the desired behavior

Okay, now I remember the PR, thanks. In any case this is inconsistent, if we fix we should fix globally. BTW, conflict in that PR

loumalouomega:
> this is conflicting with #7401, lets discuss first what should be the desired behavior

Maybe the best solution would be to create a method IsInsideWithChecks, @pooyan-dadvand suggested something like that

loumalouomega:
> > this is conflicting with #7401, lets discuss first what should be the desired behavior
> 
> Maybe the best solution would be to create a method IsInsideWithChecks, @pooyan-dadvand suggested something like that

Better thinking about FastInside in fact

loumalouomega:
For the moment I will solve my issue with a lambda, so this doesn't block my work

pooyan-dadvand:
I was saying something like `ProjectionIsInside` because is not the question if you do checks or if you do it faster. The question is if you consider out of plan or our of line as out or you want to check if projected point is inside.

Another option is to distinguish it with the tolerance. (In this case the tolerance should be the same in plana and out plane as we have only one argument)

loumalouomega:
Closing this as @KratosMultiphysics/implementation-committee reached a consensus and this will be replaced by a new method

PR #10525, opened 2022-12-03
[Core] Extend set historical variables to zero in Variable utils

sunethwarna:
**📝 Description**
As stated in the title, to support step as well in `VariableUtils::SetHistoricalVariablesToZero`.

** Todo:

- [x]  Add test

**🆕 Changelog**
- Extend  `VariableUtils::SetHistoricalVariablesToZero` to support `Step`

### DIFF
--- a/kratos/utilities/variable_utils.h
+++ b/kratos/utilities/variable_utils.h
@@ -767,7 +767,7 @@ public:
         const int Step,
         const TVariableArgs&... rVariableArgs)
     {
-        block_for_each(rNodes, [&](NodeType& rNode){(
+        block_for_each(rNodes, [&](auto& rNode){(
             AuxiliaryHistoricalValueSetter<typename TVariableArgs::Type>(rVariableArgs, rVariableArgs.Zero(), rNode, Step), ...);
         });
     }
--- a/kratos/utilities/variable_utils.h
+++ b/kratos/utilities/variable_utils.h
@@ -752,6 +752,25 @@ public:
             AuxiliaryHistoricalValueSetter<typename TVariableArgs::Type>(rVariableArgs, rVariableArgs.Zero(), rNode), ...);
         });
     }
+    
+    /**
+     * @brief Set the Historical Variables To Zero
+     * This method sets the provided list of variables to zero in the nodal historical database
+     * @tparam TVariableArgs Variadic template argument representing the variables types
+     * @param rNodes Nodes container to be initialized to zero
+     * @param Step Step of the solution step data which needs to be set to zero
+     * @param rVariableArgs Variables to be set to zero
+     */
+    template<class... TVariableArgs>
+    static void SetHistoricalVariablesToZero(
+        NodesContainerType& rNodes,
+        const int Step,
+        const TVariableArgs&... rVariableArgs)
+    {
+        block_for_each(rNodes, [&](NodeType& rNode){(
+            AuxiliaryHistoricalValueSetter<typename TVariableArgs::Type>(rVariableArgs, rVariableArgs.Zero(), rNode, Step), ...);
+        });
+    }
 
     /**
      * @brief Sets the container value of any type of non historical variable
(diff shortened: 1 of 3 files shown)


loumalouomega (approved):

loumalouomega:
Tests are failing: TestVariableUtilsSetHistoricalVariablesToZero

loumalouomega (changes requested):
Block until test fix

loumalouomega (dismissed):
Okay

kratos/utilities/variable_utils.h
```diff
@@ -1728,7 +1728,7 @@ class KRATOS_API(KRATOS_CORE) VariableUtils
         NodeType& rNode,
         const int Step = 0)
     {
-        if constexpr(std::is_integral_v<TDataType> || std::is_floating_point_v<TDataType>) {
+        if constexpr(std::is_integral_v<TDataType> || std::is_floating_point_v<TDataType> || std::is_same_v<TDataType, Matrix> || std::is_same_v<TDataType, Vector>) {
```
philbucher, line 1731: but `Matrix` and `Vector` should use `noalias` no? 🤔

sunethwarna: Thats what i thought too, but they fail if they are not properly initialized in the first place with the error bad index

kratos/utilities/variable_utils.h
```diff
@@ -753,6 +753,25 @@ class KRATOS_API(KRATOS_CORE) VariableUtils
...
+     * @param Step Step of the solution step data which needs to be set to zero
+     * @param rVariableArgs Variables to be set to zero
+     */
+    template<class... TVariableArgs>
+    static void SetHistoricalVariablesToZero(
+        NodesContainerType& rNodes,
+        const int Step,
+        const TVariableArgs&... rVariableArgs)
+    {
+        block_for_each(rNodes, [&](NodeType& rNode){(
```
philbucher:
```suggestion
        block_for_each(rNodes, [&](auto& rNode){(
```

kratos/utilities/variable_utils.h
```diff
@@ -1703,10 +1722,18 @@ class KRATOS_API(KRATOS_CORE) VariableUtils
     const TContainerType& GetContainer(const ModelPart& rModelPart);
 
     template<class TDataType>
-    static void AuxiliaryHistoricalValueSetter(
+    static void inline AuxiliaryHistoricalValueSetter(
```
matekelemen, line 1725: that `inline` doesn't do anythin

matekelemen, line 1725: that `inline` doesn't do anything

kratos/utilities/variable_utils.h
matekelemen: `noalias` would be useful if we could guarantee that the object on the left doesn't appear in the expression on the right. However, both are inputs in this case so we can't guarantee anything => the branch with `noalias`  should be deleted and we should always assign without `noalias`.

Btw this is a simple assignment; there's no point in using `noalias` because there's no intermediate step here, no potential temporary objects.

kratos/utilities/variable_utils.h
```diff
@@ -753,6 +753,25 @@ class KRATOS_API(KRATOS_CORE) VariableUtils
...
+    static void SetHistoricalVariablesToZero(
+        NodesContainerType& rNodes,
+        const int Step,
+        const TVariableArgs&... rVariableArgs)
+    {
+        block_for_each(rNodes, [&](auto& rNode){(
+            AuxiliaryHistoricalValueSetter<typename TVariableArgs::Type>(rVariableArgs, rVariableArgs.Zero(), rNode, Step), ...);
+        });
+    }
+
```
matekelemen, line 774: How about being a bit more general and make it something like `SetHistoricalVariablesTo`, and then provide the value to set them to?

RiccardoRossi (approved):
well this one is old...but i would have no issue if it gets merged

PR #10535, opened 2022-12-06
[Core][CMake] Some fixes related with Trilinos and MPI finding

loumalouomega:
**📝 Description**

Some fixes related with Trilinos and MPI finding

**🆕 Changelog**

- [Checking if MPI application is compiled to set ON USE_MPI](https://github.com/KratosMultiphysics/Kratos/commit/3b1afd719d7c450247934707c342c1684195fc30)
- [Fix random failing find](https://github.com/KratosMultiphysics/Kratos/commit/5c28a53f69ebfc4b478071570315abe0c78fb6f9)
- Some clean up (missing STATUS and remove empty messages and replace with \n)


CMakeLists.txt
```diff
@@ -484,7 +484,21 @@ if(${USE_TETGEN_NONFREE_TPL} MATCHES ON)
     endif(tetgen_FOUND)
 endif(${USE_TETGEN_NONFREE_TPL} MATCHES ON)
 
+# Checking if the MPI specific apps are compiled
```
philbucher: Metis does not require MPI any more (well my PR needs to be merged but hopefully thats only a matter of days)

In general I would prefer this mechanism to be in the apps themselves, same as we do with kratos_add_dependency

Other opinions?
@roigcarlo ?

CMakeLists.txt
```diff
@@ -484,7 +484,21 @@ if(${USE_TETGEN_NONFREE_TPL} MATCHES ON)
...
+        if(${USE_MPI} MATCHES OFF )
+            SET(USE_MPI ON)
+            MESSAGE(WARNING "${APPLICATION_MPI} application requires of USE_MPI. Setting to ON")
+        ENDIF(${USE_MPI} MATCHES OFF )
+    ENDIF (${index} GREATER -1)
+ENDFOREACH(APPLICATION_MPI ${LIST_OF_MPI_SPECIFIC_APPS})
+
+# Setting MPI
 if(${USE_MPI} MATCHES ON )
+  message(STATUS "Compiling with MPI support")
```
philbucher: This message is already in the MPIcore

loumalouomega: Reverted

CMakeLists.txt
loumalouomega: The problem is, MPI is check here, so if Trilinos is compiled and MPI is not selected it will fail. If the check is done later it would be too late

loumalouomega: Maybe if the MPI check is separated in a CMake module it could be done in a more modular way, and then call it from the app

loumalouomega:
@roigcarlo

roigcarlo (commented):
Since we are making an effort to remove mpi as a hard dependency I would rather move the finding code into the mpi extension of the core, but changes in the PR are OK so imo we can merge it and move them if we can make it work (in another PR)

CMakeLists.txt
roigcarlo: I would put the check in the app itself, as other requirements.

I would not create a cmake module for that as its simply an if, and trilinos is the only app that has that dependency. But if you want to do it I won't block it.

loumalouomega: But here not only the check is done, is activating it in case it is not

roigcarlo: that's the problem, we should not assume that someone wants to enable MPI because it has enabled trilinos. It could be what the user wanted or it could simply be because the user did not remember to remove it.

In general I prefer this kind of errors to show up, because then is easier to diagnose. 

I am thinking on (for example) the case of someone trying to compile HDF5Application with the serial lib and having the Trilinos app enabled as well. In that case it would crash because the libhdf5-mpi not found, even when MPI is off because it has been auto-activated by trilinos, which is not obvious unless you know what is happening underneath.

loumalouomega: Okay, in that case I will put an error in trilinos instead of activating, because Trilinos app requires MPI 100%

loumalouomega: Changes suggested applied

roigcarlo (approved):

PR #10533, opened 2022-12-06
[Core] Adding CopyValuesFromExistingParameters to Kratos Parameters

loumalouomega:
**📝 Description**

 Adding CopyValuesFromExistingParameters to Kratos Parameters. This will allow reducing a lot of verbose code when manually copying parameters 

**🆕 Changelog**

- [Adding CopyValuesFromExistingParameters](https://github.com/KratosMultiphysics/Kratos/commit/15e524d0469134cf5d7b507c4a67c4884bb72d85)
- [Expose to python](https://github.com/KratosMultiphysics/Kratos/commit/aa84268cb5a833a6bd5828028b751876535f1069)
- [Adding test](https://github.com/KratosMultiphysics/Kratos/commit/019b8c3774def5963277d549be032af500edf47b)
- Minor clean up

PR #10539, opened 2022-12-09
[Core] Adding RemoveValues to Kratos Parameters

loumalouomega:
**📝 Description**

I add this method in order to reduce a lot of duplicated code when removing several keys at once. In the same style as https://github.com/KratosMultiphysics/Kratos/pull/10533

**🆕 Changelog**

- [Proper category](https://github.com/KratosMultiphysics/Kratos/commit/57fdfb228875fca0e9628e55fd2cc2f29599474f)
- [Adding RemoveValues](https://github.com/KratosMultiphysics/Kratos/commit/0de7c4760b4054e4d68dafcb728cba098f630e0f)
- [Expose to python](https://github.com/KratosMultiphysics/Kratos/commit/f0d97a92fadf4e773c81670ee3648fe202fef9aa)
- [Adding test](https://github.com/KratosMultiphysics/Kratos/commit/abff192811cf12c8c756faade389664429be72f9)


kratos/sources/kratos_parameters.cpp
```diff
@@ -483,6 +483,19 @@ bool Parameters::RemoveValue(const std::string& rEntry)
...
 
+bool Parameters::RemoveValues(const std::vector<std::string>& rEntries)
+{
+    bool sucess;
+    for (auto& r_entry : rEntries) {
+        sucess = this->RemoveValue(r_entry);
+        if (!sucess) return false;
+    }
+    return true;
+}
```
sunethwarna: This removal is order dependent from the `rEntries`. If a mid item is not removed successfully, it will exit having `Parameters` removed half of the entries and we let the user of this function to decide on to throw an error or not.
I would do it in either of the following ways.
1. First check whether all the `rEntries` present in `Parameters`, if so then removing all.
2. do not return if the item was not successfully removed, return `success` at the end. Example: `bool sucess = true; success = success && this->RemoveValue(r_entry); return success` in respective lines.

sunethwarna (commented):
Minor comment

kratos/sources/kratos_parameters.cpp
loumalouomega: Okay

loumalouomega: Done

kratos/sources/kratos_parameters.cpp
```diff
@@ -483,6 +483,24 @@ bool Parameters::RemoveValue(const std::string& rEntry)
...
+    for (auto& r_entry : rEntries) {
+        sucess = this->Has(r_entry);
+        if (!sucess) break;
+    }
+    if (sucess) {
+        for (auto& r_entry : rEntries) {
+            this->RemoveValue(r_entry);
+        }
+    }
+    return sucess;
```
matekelemen:
```suggestion
    for (const auto& r_entry : rEntries) {
        if (!this->Has(r_entry)) return false;
    }
    for (const auto& r_entry : rEntries) {
        this->RemoveValue(r_entry);
    }
    return true;
```

`const` + don't need the extra `bool`

kratos/tests/test_kratos_parameters.py
```diff
@@ -497,6 +497,17 @@ def test_remove_value(self):
         self.assertFalse(kp.Has("int_value"))
         self.assertFalse(kp.Has("level1"))
 
+    def test_remove_values(self):
+        kp = Parameters(json_string)
+        self.assertTrue(kp.Has("int_value"))
+        self.assertTrue(kp.Has("level1"))
+
+        list_remove = ["int_value", "level1"]
+        kp.RemoveValues(list_remove)
```
matekelemen, line 512: Can you plz add a quick test for a failure too?

loumalouomega: I can...

loumalouomega: Done

sunethwarna (approved):
Thanks @loumalouomega :+1: Ok from my side.

Issue #2817, opened 2018-09-06
ProcessInfo is not being passed to Initialize of Elements/Conditions


philbucher:
@KratosMultiphysics/technical-committee 

Title says all. Is this intentional? I was having some minor issues with this in #2349 

I see that it is a but a pain to change but maybe it would be a good thing to do bcs it is being passed to pretty much all the other functions too

maceligueta:
As far as I remember from conversations with @RiccardoRossi some years back, it was intentional. Anyways, the Technical Committe will discuss this issue and post an official answer here.

RiccardoRossi:
I don't remember the details, i do remember there were some issues. However i am not opposing this change...

pooyan-dadvand:
@KratosMultiphysics/technical-committee: To be passed to initialize. @philbucher you may proceed with it

philbucher:
done in #4938

PR #10542, opened 2022-12-12
[MPM] Restart Application

### LINKED ISSUE #10664: [Core] Passing the ProcessInfo to the InitializeMaterial() function for the Constitutive Laws
**Description**
I want to suggest to the @KratosMultiphysics/technical-committee to pass the `ProcessInfo` as an additional parameter in the function call of the `InitializeMaterial()` function for the Constitutive Laws.

Currently, the `ProcessInfo` is not being passed to the `InitializeMaterial()` function for the Constitutive Laws. The `ProcessInfo` contains information, among other things, on whether the simulation was restarted or not.
This information is required in the `ParticleMechanicsApplication` for example, as only some parts of the `InitializeMaterial()` function need to be called on a restart.

For this reason, I would suggest adjusting the `InitializeMaterial()` function in the `consitutive_law.h` file to pass the `ProcessInfo` in the function call as well.
To not break the other applications, which are not passing the `ProcessInfo`, I think the best solution would be to overload this function.
The reason for requesting this change, as well as an implementation of the suggested changes, can be seen in this branch: #10542.

andinoboerst:
**📝 Description**
Implementing the restart application for use with the Particle Mechanics Application.

**🆕 Changelog**
- Enabled the mpm_solver to identify and accept restart files as model imports
- Forced reassignment of the material points to the corresponding grid elements on restarts


andinoboerst:
*To-Do:*
Write a test for the mpm restart application.

applications/ParticleMechanicsApplication/custom_utilities/mpm_search_element_utility.h
```diff
@@ -429,8 +440,24 @@ namespace MPMSearchElementUtility
...
-        NeighbourSearchElements(rMPMModelPart, rBackgroundGridModelPart, missing_elements, Tolerance);
-        NeighbourSearchConditions(rMPMModelPart, rBackgroundGridModelPart, missing_conditions, Tolerance);
+        if (!rMPMModelPart.GetProcessInfo()[IS_RESTARTED]) {
+            NeighbourSearchElements(rMPMModelPart, rBackgroundGridModelPart, missing_elements, Tolerance);
+            NeighbourSearchConditions(rMPMModelPart, rBackgroundGridModelPart, missing_conditions, Tolerance);
+        } else {
+            #pragma omp parallel for
+            for (int i = 0; i < static_cast<int>(rMPMModelPart.Elements().size()); ++i) {
+                auto element_itr = (rMPMModelPart.ElementsBegin() + i);
+                #pragma omp critical
```
philbucher: this serializes the entire loop, making it even slower than without OMP :)

I recommend to directly use `std::vector<typename Condition::Pointer> = rMPMModelPart.Conditions()`. AFAIK this is not possible (due to type mismatch), hence the easiest and fastest solution is probably the following:
~~~c++
missing_elements.resize(rMPMModelPart.NumberOFElements());
IndexPartition(rMPMModelPart.NumberOFElements()).for_each([](std::size_t i){
    missing_elements[i] = &*(rMPMModelPart.ElementsBegin() + i); // maybe add/remove *&?
});

andinoboerst: True, good point.
I will try that solution and update, thank you! :)

andinoboerst: Ok done :thumbsup:

applications/ParticleMechanicsApplication/custom_constitutive/hyperelastic_3D_law.cpp
```diff
@@ -153,10 +153,19 @@ void HyperElastic3DLaw::InitializeMaterial( const Properties& rMaterialPropertie
         const GeometryType& rElementGeometry,
         const Vector& rShapeFunctionsValues )
 {
-  mDeterminantF0                = 1;
-  mInverseDeformationGradientF0 = IdentityMatrix(3);
-  mStrainEnergy                 = 0;
+    bool restarted = false;
+    
+    if (rShapeFunctionsValues.size() > 0){
+        if (rShapeFunctionsValues(0) != -888 ){
```
philbucher: this is quite a hack :thinking: 
@AlejandroCornejo how do you handle this in the CLawsApp?
This function seems to be one of the few (if not the only one) that does not have access to the `ProcessInfo`

andinoboerst: Yes 😅 i've been trying to figure out a way to solve this differently, for example by adding a "IS_RESTARTED" value to the Material Properties but I was running into issues when running it with multiple threads in OMP. So any other suggestions are welcome, I just used this as a temporary solution until I could come up with something different.

philbucher: I did not look at it in detail, but I know that this would not be the first function in Elements/Conditions/Constitutive-Laws that is missing the `ProcessInfo

Lets wait for the response of @AlejandroCornejo
The proper solution would probably be to update the interface of this function. This involves the core and hence takes some time, but we have a workflow/knowledge how to do it backward compatible (e.g. see #2817)

philbucher: > Material Properties but I was running into issues when running it with multiple threads in OMP

yes this is not allowed, and will lead to race conditions as you experienced :)

applications/ParticleMechanicsApplication/custom_constitutive/hencky_plastic_3D_law.cpp
```diff
@@ -189,7 +189,7 @@ void HenckyElasticPlastic3DLaw::InitializeMaterial(const Properties& rProps,
     bool restarted = false;
 
     if (rShapeFunctionsValues.size() > 0){
-        if (rShapeFunctionsValues(0) != -888 ){
+        if (rShapeFunctionsValues(0) == -888 ){
```
AlejandroCornejo: I do not understand why you need to set anything related to the ShapeFunctions in the CL, can you develop a bit?

AlejandroCornejo: @philbucher don't we have a `rProcessInfo[IS_RESTARTED]` capability?

rubenzorrilla: Yes. Indeed, this is the purposed it is intended to be used for.

andinoboerst: Yes however, I don't know how to access the ProcessInfo from within this function, as it is not passed in the function call and I could not find it being saved within the class (I might be wrong).

rubenzorrilla: True. I think you'll need to do it through the properties...

rubenzorrilla: Indeed, I can't understand why we are not passing the `ProcessInfo` in here. @KratosMultiphysics/technical-committee .

AlejandroCornejo: Theoretically, the Initialize is done only once. If it is restarted it does not make sense to do it again (IMO). All the parameters to be initialized are serialized

andinoboerst: I thought so too, but if I don't reinitialize the `mpMPMFlowRule` in the hencky constitutive law the restart does not work. In the case of using a linear elastic law the whole initialize can be skipped, but for the hencky law it didn't work.

philbucher: I would pass the ProcessInfo.

Then it is consistent with other parts of the code and everyone can do what they need

To me this looks like a leftover

rubenzorrilla: Aside of the `ProcessInfo` discussion. This seems a bug in the serialization of one of the constitutive laws...

andinoboerst: True. I did not verify the serialization within the constitutive laws, but I will check if the `mpMPMFlowRule` is skipped and in that case if I can add it.

andinoboerst: This seems to be the problem here:

https://github.com/KratosMultiphysics/Kratos/blob/057edcf143648d0ece0df3595d50bd8f0c44063a/applications/ParticleMechanicsApplication/custom_constitutive/hardening_laws/particle_hardening_law.hpp#L259-L273

andinoboerst: I cannot come up with a solution on how to solve the serializer problem for the particle hardening law (I don't know if someone else had this problem with the recursive call before?). To be fair the information saved there only includes the material properties from the `Materials.json` file, so even when restarting these values can be imported again.
However, this means it needs to be initialized, which goes back to the initial problem of this strange workaround. To solve this workaround it's necessary to either pass the `rProcessInfo` to the `InitializeMaterial()` function or to include a new function for example `InitializeMaterialRestart()` in the `constitutive_law.h`. Maybe the second option does not make so much sense if this application is the only one that needs it, even though in my case it would be better since I would not need to include an if-statement in cases where the whole initialization can be skipped. Both of these solutions include modifying the core though.

andinoboerst: @philbucher @rubenzorrilla @AlejandroCornejo If you all agree I would open an issue assigning the technical committee to decide what is the best way to proceed. I would suggest just passing the `ProcessInfo` as it will be the easiest way to implement this and could be useful in the future as well.

kratos/sources/constitutive_law.cpp
```diff
@@ -548,13 +548,21 @@ bool ConstitutiveLaw::IsIncremental()
...
  * @param rShapeFunctionsValues the shape functions values in the current integration point
+ * @param rCurrentProcessInfo the current process information
  */
 void ConstitutiveLaw::InitializeMaterial(const Properties& rMaterialProperties,
         const GeometryType& rElementGeometry,
         const Vector& rShapeFunctionsValues)
 {
 }
 
+void ConstitutiveLaw::InitializeMaterial(const Properties& rMaterialProperties,
```
philbucher, line 559: LGTM!

Please add a deprecation warning to the other function (KRATOS_DEPRECATED_MESSAGE)

kratos/python/add_constitutive_law_to_python.cpp
```diff
@@ -209,7 +228,8 @@ void  AddConstitutiveLawToPython(pybind11::module& m)
     .def("FinalizeMaterialResponsePK2",&ConstitutiveLaw::FinalizeMaterialResponsePK2)
     .def("FinalizeMaterialResponseKirchhoff",&ConstitutiveLaw::FinalizeMaterialResponseKirchhoff)
     .def("FinalizeMaterialResponseCauchy",&ConstitutiveLaw::FinalizeMaterialResponseCauchy)
-    .def("InitializeMaterial",&ConstitutiveLaw::InitializeMaterial)
+    .def("InitializeMaterial",&ConstitutiveLawInitializeMaterialStandard, py::return_value_policy::reference_internal)
+    .def("InitializeMaterial",&ConstitutiveLawInitializeMaterialProcessInfo, py::return_value_policy::reference_internal)
```
philbucher: Please use py::overload_cast (and remove the return value stuff since it is a void fct)

philbucher (commented):
Nice!

Btw regarding the original problem with the serialzation, I think that CLaws should not save Properties, but use them as passed from outside

kratos/python/add_constitutive_law_to_python.cpp
andinoboerst: Yes thanks, done!

applications/ParticleMechanicsApplication/custom_constitutive/hencky_plastic_3D_law.cpp
matekelemen: > This seems to be the problem here:
> 
> https://github.com/KratosMultiphysics/Kratos/blob/057edcf143648d0ece0df3595d50bd8f0c44063a/applications/ParticleMechanicsApplication/custom_constitutive/hardening_laws/particle_hardening_law.hpp#L259-L273

**Ugh**, if that can't be solved, I'd at least do one of the following:
- throw an exception when `save` is called (best indicator of failure in my opinion: fails exactly when the problem arises)
- throw an exception when `load` is called (not great, not terrible: fails when your results are about to get messed up)
- issue warnings in `save` and `load` (suboptimal: noone cares about warnings until they've hunted the bug for a couple of days)

andinoboerst: The problem with throwing an exception is that the restart application would never work in this case. When the restart application tries to save or load this information it would then fail every time.
As @philbucher mentioned in the next comment, the other applications do not save these properties, but instead, pass the properties whenever necessary from the outside.
However, if this is to be changed, then all calls involving the `particle_hardening_law` would need to be changed.

matekelemen: Well, if there's no way of omitting them during serialization/deserialization, all that's left is at least to issue a warning.

matekelemen: By the way, I think it would be best to follow up on why serialization is not possible. The current behaviour is a bug and no amount of workarounds will change that. Furthermore, it might be straightforward for us right now to follow how the restarted initialization happens, but I guarantee that it won't be for others who take a look at it for the first time.

I haven't looked into it yet, but I have a sneaking suspicion that we'll find other bugs if we start looking into this more closely. Storing `shared_ptr`s to each other definitely smells like a memory leak.

andinoboerst: I agree, I will look into this. I will try to change the current implementation to avoid storing the material properties in this hardening_law class and instead pass the properties to all of the functions that need them. Maybe this way it would also not be necessary to have an `is_restarted` flag in the `InitializeMaterial()` function.

I will follow up on that.

RiccardoRossi:
about the multithreading issue, the point is that you need to ensure that the variable IS_RESTARTED is defined in the Properties OUTSIDE OF A PARALLEL SECTION.

andinoboerst:
> about the multithreading issue, the point is that you need to ensure that the variable IS_RESTARTED is defined in the Properties OUTSIDE OF A PARALLEL SECTION.

Yes, that is true, but this only works when running a simulation that was not restarted. The `Properties` are being serialized inside `element.h` so when I am running a restarted simulation the properties are loaded directly into a parallel section so that I cannot modify the `IS_RESTARTED` flag anymore.

andinoboerst:
I now worked on an alternative solution that should take care of the serialization bug inside the `particle_hardening_law`. I was hesitant about making such a core change to the entire class, but based on the discussions here it seems that it might be the better solution after all.
An alternative draft can be found in PR #10667.
This alternative solution would also not need a `is_restarted` flag inside the constitutive law function anymore.

applications/ParticleMechanicsApplication/custom_constitutive/hencky_plastic_3D_law.cpp
andinoboerst: Ok so I deleted the attribute `mpProperties` inside the `particle_hardening_law` now and instead pass it as a parameter to all the functions that need it. This way the `mpProperties` do not need to be serialized anymore and thus avoid storing pointers to each other.
See PR #10667.

PR #10546, opened 2022-12-12
[Core][Geometry] Directly call internal methods for Inverse of Jacobian

loumalouomega:
**📝 Description**

Directly call internal methods for Inverse of Jacobian instead of directly compute inverse (reduce code duplication). Also, when not calling `InverseOfJacobian`, using `GeneralizedInvertMatrix` instead of `InvertMatrix` for the sake of consistency with `InverseOfJacobian`

**🆕 Changelog**

- `pragma once`
- [Directly call internal methods for Inverse of Jacobian](https://github.com/KratosMultiphysics/Kratos/commit/32f2aa7ca4548b1012f5fd0a73e54b1e4e7dad13)
- Using `GeneralizedInvertMatrix` instead of `InvertMatrix`. Consistent with  `InverseOfJacobian`


rubenzorrilla (approved):
Makes sense.

PR #10548, opened 2022-12-13
[Core] Improve AuxiliarModelPartUtilities

philbucher:
the Get- & Setfunctions now also work with `Vector`

FYI @DSchmoelz

### DIFF
--- a/kratos/utilities/auxiliar_model_part_utilities.h
+++ b/kratos/utilities/auxiliar_model_part_utilities.h
@@ -293,11 +293,11 @@ public:
 
 
     /// To Export a Scalar data (Double/int/...)
-    template<typename TDataType>
+    template<class TContainerType>
     void GetScalarData(
-        const Variable<TDataType>& rVariable,
+        const Variable<typename TContainerType::value_type>& rVariable,
         const DataLocation DataLoc,
-        std::vector<TDataType>& data) const
+        TContainerType& data) const
     {
         switch (DataLoc)
         {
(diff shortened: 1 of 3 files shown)


kratos/utilities/auxiliar_model_part_utilities.h
```diff
@@ -350,11 +350,11 @@ class KRATOS_API(KRATOS_CORE) AuxiliarModelPartUtilities
     }
 
     /// To Export a Vector data (std::vector/array/..)
-    template<class TDataType>
+    template<class TContainerType, class TVarType>
```
sunethwarna, line 355: Curious: Wouldn't it be better to deduce `TVarType` from `TContainerType::value_type` as in the scalar case?

kratos/utilities/auxiliar_model_part_utilities.h
```diff
@@ -694,17 +694,17 @@ class KRATOS_API(KRATOS_CORE) AuxiliarModelPartUtilities
...
-    template<typename TDataType, class TContainerType>
-    void GetScalarDataFromContainer(const TContainerType& rContainer, const Variable<TDataType>& rVariable, std::vector<TDataType>& data) const
+    template<typename TDataType, class TContainerType, class TDataContainerType>
+    void GetScalarDataFromContainer(const TContainerType& rContainer, const Variable<TDataType>& rVariable, TDataContainerType& data) const
     {
         IndexPartition<std::size_t>(rContainer.size()).for_each([&](std::size_t index){
             const auto& r_entity = *(rContainer.begin() + index);
             data[index] = r_entity.GetValue(rVariable);
         });
     }
```
sunethwarna: No size checks. Was it done purposely?

sunethwarna (commented):
Minor comments only except for one major comment. What is the reason behind not performing size checks within the function?

loumalouomega (dismissed):
Okay from my side

kratos/utilities/auxiliar_model_part_utilities.h
philbucher: because I need to turn Vector variables (such as array_3) into a vector

kratos/utilities/auxiliar_model_part_utilities.h
philbucher: added

loumalouomega (approved):

PR #10552, opened 2022-12-13
[Core] Profiler

matekelemen:
# TL;DR

Add a basic thread-safe profiler that measures total time spent in a scope, as well as the number of times that scope was entered. No MPI support yet. Questions, suggestions and improvements are welcome.

Example usage:
```cpp
ModelPart::IndexType ModelPart::CloneSolutionStep()
{
    KRATOS_PROFILE_SCOPE(KRATOS_CODE_LOCATION); // <== accumulate the total runtime of the entire function
    ...
}
```

Possible output:
```json
{
    "meta": {
        "name": "total",
        "time": 2375888,
        "timeUnit": "us"
    },
    "results": [
        {
            "callCount": 1564,
            "file": "/home/kelemen/git/krex/external/kratos/kratos/sources/model_part.cpp",
            "function": "ModelPart::IndexType Kratos::ModelPart::CloneSolutionStep()",
            "line": 123,
            "time": "4811"
        }
    ]
}
```

### Misc Changes
- `const` qualifier for `Parameters::Clone` (API change)
- replace `CodeLocation`'s copy constructor with a `default` generated one
- add the `KRATOS_ENABLE_PROFILING` flag in CMake that ensures empty macros unless the user opts in for profiling
- require #10052

# Details

### Overview

Each **first** call to the `KRATOS_PROFILE_SCOPE` macro creates a *static* struct ([Item](https://github.com/KratosMultiphysics/Kratos/blob/ebc1d843fc9f221774f81d363c90b7e26bebb84f/kratos/utilities/profiler.h#L43)) that aggregates stats for that specific scope. Additionally, **on each call** a *non-static* scope guard ([Scope](https://github.com/KratosMultiphysics/Kratos/blob/ebc1d843fc9f221774f81d363c90b7e26bebb84f/kratos/utilities/profiler.h#L65)) is constructed which updates the stats when the scope is left.

### Implementation

`Item`s are `thread_local` (also implying `static`) and are only constructed once, when the macro is first encountered at runtime, and stored in a container that ensures **thread-safe** and **persistent** access. Subsequent calls just create a `Scope` that takes a *raw reference* to the existing `Item` and updates it during destruction.

I know what you're thinking: *Oh god, playing with `static` variables and raw references in a multithreaded environment is a recipe for disaster*, but keep in mind that the static `Item`s are stored in linked lists unique to each thread, so cross-thread mutation and iterator invalidation is impossible. The benefit of this nasty-looking system is that the overhead is front-loaded into the first pass, and for the rest of the passes it's just a single dereferencing.

# ToDo

Either sum times and call counts across MPI ranks and write one aggregate file, or create unique files for each rank separately.


kratos/includes/code_location.h
```diff
@@ -33,7 +33,9 @@ namespace Kratos
 
 		CodeLocation();
 
-		CodeLocation(CodeLocation const & Other);
+        CodeLocation(CodeLocation&& rOther) noexcept = default;
+
+		CodeLocation(CodeLocation const & Other) = default;
```
loumalouomega: Tabs (replace spaces)

loumalouomega:
This is nice, but I will block to avoid accidental merge

loumalouomega (changes requested):
Blocking

roigcarlo (commented):
To be honest I see no reason to block this. In fact I see it very useful, for me it can go in, but If you want we can discuss first in the @KratosMultiphysics/technical-committee

pooyan-dadvand:
Very nice addition and very easy interface! 

Some questions:
1. Why aren't you using directly TLS variable for the items rather than the map of the [thread, items]?
2. I assume that you are using the list rather than directly summing the results to avoid search overhead. Nevertheless, it makes one allocation per push_back which is very time consuming. (order of 1000 cycles) Is it really faster than having an unorder map and adding it? Have you compared the overhead? Shall we go for TLS statics and keep track of them in the singleton list?
3. Shall we have different levels of profiling macro? So we can keep some upper levels on in production and only mute the lower level one? (like the warning levels)

Apart from the questions, I think it may be useful to give the load balancing info as you have the times per thread!

Todo in a separate PR: Making the Timer work in the same way. It is way cleaner than the start stop of timer! (My fault!)

kratos/sources/profiler.cpp
```diff
@@ -0,0 +1,236 @@
...
+{
+    return "ns";
+}
+} // unnamed namespace
+
+
+template <class T>
+Profiler<T>::~Profiler()
+{
+    // Stop the total timer
```
pooyan-dadvand: I would put this function in a ToJson() function so we can call it on demand. (for example in each stage or time step)

matekelemen: Yep, that's a good idea.

Also, writing to a file is not MPI safe in general. I'm thinking about appending the file name using the default communicator, but I noticed using that is discouraged. Any tips or better ideas on how to address this?

matekelemen:
> 1) Why aren't you using directly TLS variable for the items rather than the map of the [thread, items]?

I'm not sure I get it, but if you mean `thread_local`, I don't use it because
1) it also implies `static` and I'd rather avoid making the `Profiler` class itself `static`. I prefer having the class non-static and provide access to a static instance through a singleton (`ProfilerSingleton` in this case). This allows people to use the class locally as well if they so desire.
2) `thread_local` is not accessible from other threads (at least I think it isn't), but I need to reduce the results from all threads onto the main thread in the end.

> 2) I assume that you are using the list rather than directly summing the results to avoid search overhead. Nevertheless, it makes one allocation per push_back which is very time consuming. (order of 1000 cycles) Is it really faster than having an unorder map and adding it? Have you compared the overhead? Shall we go for TLS statics and keep track of them in the singleton list?

You're right about avoiding the lookup, but not about the reason for using the `std::list`. I'm using an `std::list` because it never gets resized/reallocated; once an item is pushed to it, that item will remain at that memory address until it is erased from the list or the list is destroyed. This allows me to store a **raw reference** to that memory address and safely use it afterwards, knowing that it will never get invalidated.

This last bit also answers your concern about performance. I only call `std::list::push_back` once per profiled item: the first time that specific macro is encountered. Subsequent accesses always happen using that raw reference. 

You're right about running some performance tests though. It'd be nice to verify that what's happening actually coincides with what I meant to do + quantify the overhead of using the profiler.

> 3) Shall we have different levels of profiling macro? So we can keep some upper levels on in production and only mute the lower level one? (like the warning levels)

Sure, I'm open to suggestions. It would probably involve creating different singletons and their related macros.

> Apart from the questions, I think it may be useful to give the load balancing info as you have the times per thread!

You mean writing the individual data (call counts, durations) from each thread as well?

kratos/sources/profiler.cpp
philbucher: I think in this particular case it might be of to use the default data comm

Or can you pass one at construction? That would be the preferred way

rubenzorrilla:
We had a look to this in the last @KratosMultiphysics/technical-committee meeting.

We all consider this a very nice addition. However, we're afraid of the performance implications. @matekelemen would you please do an overhead analysis using a profiler? Thanks.

matekelemen:
## TL;DR

Up to the resolution of `perf`, the overhead incurred by `Profiler` is exclusively dominated by `std::chrono::high_resolution_clock::now`. As far as I can tell, the only way of further reducing the overhead is by somehow quantifying performance **without measuring time**.

This doesn't mean that profiling comes for free, but it **does** mean that the extra infrastructure beyond what's necessary is minimal.

## Setup

### Case
I chose to run the [body fitted cylinder](https://github.com/KratosMultiphysics/Examples/tree/master/fluid_dynamics/validation/body_fitted_cylinder_100Re) from the examples repo, but restricted `"end_time"` to `2.0` in order to keep `prof`'s output size manageable.

### `KRATOS_PROFILE_SCOPE`

I chose the worst possible places to insert `KRATOS_PROFILE_SCOPE` that I could come up with: every getter in the `VariableListDataValueContainer` class, effectively triggering a profiled scope every single time any of a `Node`'s variables are queried. Here's the diff with respect to the current state of the PR:
```diff
diff --git a/kratos/containers/variables_list_data_value_container.h b/kratos/containers/variables_list_data_value_container.h
index 42eb3dfc06..d1f1688b08 100644
--- a/kratos/containers/variables_list_data_value_container.h
+++ b/kratos/containers/variables_list_data_value_container.h
@@ -11,6 +11,7 @@
 //
 //
 
+#include "utilities/profiler.h"
 #if !defined(KRATOS_VARIABLES_LIST_DATA_VALUE_CONTAINER_H_INCLUDED )
 #define KRATOS_VARIABLES_LIST_DATA_VALUE_CONTAINER_H_INCLUDED
 
@@ -27,6 +28,7 @@
 #include "containers/variable.h"
 #include "containers/variables_list.h"
 #include "includes/global_variables.h"
+#include "utilities/profiler.h"
 
 namespace Kratos
 {
@@ -280,6 +282,7 @@ public:
     template<class TDataType>
     TDataType& GetValue(const Variable<TDataType>& rThisVariable)
     {
+        KRATOS_PROFILE_SCOPE(KRATOS_CODE_LOCATION);
         KRATOS_DEBUG_ERROR_IF(!mpVariablesList) << "This container don't have a variables list assigned. A possible reason is creating a node without a model part." << std::endl;
         KRATOS_ERROR_IF_NOT(mpVariablesList->Has(rThisVariable)) << "This container only can store the variables specified in its variables list. The variables list doesn't have this variable:" << rThisVariable << std::endl;
         return *(reinterpret_cast<TDataType*>(Position(rThisVariable)) + rThisVariable.GetComponentIndex());
@@ -288,6 +291,7 @@ public:
     template<class TDataType>
     TDataType& GetValue(const Variable<TDataType>& rThisVariable, SizeType QueueIndex)
     {
+        KRATOS_PROFILE_SCOPE(KRATOS_CODE_LOCATION);
         KRATOS_DEBUG_ERROR_IF(!mpVariablesList) << "This container don't have a variables list assigned. A possible reason is creating a node without a model part." << std::endl;
         KRATOS_ERROR_IF_NOT(mpVariablesList->Has(rThisVariable)) << "This container only can store the variables specified in its variables list. The variables list doesn't have this variable:" << rThisVariable << std::endl;
         return *(reinterpret_cast<TDataType*>(Position(rThisVariable, QueueIndex)) + rThisVariable.GetComponentIndex());
@@ -296,6 +300,7 @@ public:
     template<class TDataType>
     const TDataType& GetValue(const Variable<TDataType>& rThisVariable) const
     {
+        KRATOS_PROFILE_SCOPE(KRATOS_CODE_LOCATION);
         KRATOS_DEBUG_ERROR_IF(!mpVariablesList) << "This container don't have a variables list assigned. A possible reason is creating a node without a model part." << std::endl;
         KRATOS_ERROR_IF_NOT(mpVariablesList->Has(rThisVariable)) << "This container only can store the variables specified in its variables list. The variables list doesn't have this variable:" << rThisVariable << std::endl;
         return *(reinterpret_cast<const TDataType*>(Position(rThisVariable)) + rThisVariable.GetComponentIndex());
@@ -304,6 +309,7 @@ public:
     template<class TDataType>
     const TDataType& GetValue(const Variable<TDataType>& rThisVariable, SizeType QueueIndex) const
     {
+        KRATOS_PROFILE_SCOPE(KRATOS_CODE_LOCATION);
         KRATOS_DEBUG_ERROR_IF(!mpVariablesList) << "This container don't have a variables list assigned. A possible reason is creating a node without a model part." << std::endl;
         KRATOS_ERROR_IF_NOT(mpVariablesList->Has(rThisVariable)) << "This container only can store the variables specified in its variables list. The variables list doesn't have this variable:" << rThisVariable << std::endl;
         return *(reinterpret_cast<const TDataType*>(Position(rThisVariable, QueueIndex)) + rThisVariable.GetComponentIndex());
@@ -312,24 +318,28 @@ public:
     template<class TDataType>
     TDataType& FastGetValue(const Variable<TDataType>& rThisVariable)
     {
+        KRATOS_PROFILE_SCOPE(KRATOS_CODE_LOCATION);
         return *(reinterpret_cast<TDataType*>(Position(rThisVariable)) + rThisVariable.GetComponentIndex());
     }
 
     template<class TDataType>
     TDataType* pFastGetValue(const Variable<TDataType>& rThisVariable)
     {
+        KRATOS_PROFILE_SCOPE(KRATOS_CODE_LOCATION);
         return (reinterpret_cast<TDataType*>(Position(rThisVariable)) + rThisVariable.GetComponentIndex());
     }
 
     template<class TDataType>
     TDataType& FastGetValue(const Variable<TDataType>& rThisVariable, SizeType QueueIndex)
     {
+        KRATOS_PROFILE_SCOPE(KRATOS_CODE_LOCATION);
         return *(reinterpret_cast<TDataType*>(Position(rThisVariable, QueueIndex)) + rThisVariable.GetComponentIndex());
     }
 
     template<class TDataType>
     TDataType& FastGetValue(const Variable<TDataType>& rThisVariable, SizeType QueueIndex, SizeType ThisPosition)
     {
+        KRATOS_PROFILE_SCOPE(KRATOS_CODE_LOCATION);
         KRATOS_DEBUG_ERROR_IF(!mpVariablesList) << "This container don't have a variables list assigned. A possible reason is creating a node without a model part." << std::endl;
         KRATOS_DEBUG_ERROR_IF_NOT(mpVariablesList->Has(rThisVariable)) << "This container only can store the variables specified in its variables list. The variables list doesn't have this variable:" << rThisVariable << std::endl;
         KRATOS_DEBUG_ERROR_IF((QueueIndex + 1) > mQueueSize) << "Trying to access data from step " << QueueIndex << " but only " << mQueueSize << " steps are stored." << std::endl;
@@ -339,6 +349,7 @@ public:
     template<class TDataType>
     TDataType& FastGetCurrentValue(const Variable<TDataType>& rThisVariable, SizeType ThisPosition)
     {
+        KRATOS_PROFILE_SCOPE(KRATOS_CODE_LOCATION);
         KRATOS_DEBUG_ERROR_IF(!mpVariablesList) << "This container don't have a variables list assigned. A possible reason is creating a node without a model part." << std::endl;
         KRATOS_DEBUG_ERROR_IF_NOT(mpVariablesList->Has(rThisVariable)) << "This container only can store the variables specified in its variables list. The variables list doesn't have this variable:" << rThisVariable << std::endl;
         return *(reinterpret_cast<TDataType*>(mpCurrentPosition + ThisPosition) + rThisVariable.GetComponentIndex());
@@ -347,24 +358,28 @@ public:
     template<class TDataType>
     const TDataType& FastGetValue(const Variable<TDataType>& rThisVariable) const
     {
+        KRATOS_PROFILE_SCOPE(KRATOS_CODE_LOCATION);
         return *(reinterpret_cast<const TDataType*>(Position(rThisVariable)) + rThisVariable.GetComponentIndex());
     }
 
     template<class TDataType>
     const TDataType* pFastGetValue(const Variable<TDataType>& rThisVariable) const
     {
+        KRATOS_PROFILE_SCOPE(KRATOS_CODE_LOCATION);
         return (reinterpret_cast<const TDataType*>(Position(rThisVariable)) + rThisVariable.GetComponentIndex());
     }
 
     template<class TDataType>
     const TDataType& FastGetValue(const Variable<TDataType>& rThisVariable, SizeType QueueIndex) const
     {
+        KRATOS_PROFILE_SCOPE(KRATOS_CODE_LOCATION);
         return *(reinterpret_cast<const TDataType*>(Position(rThisVariable, QueueIndex)) + rThisVariable.GetComponentIndex());
     }
 
     template<class TDataType>
     const TDataType& FastGetValue(const Variable<TDataType>& rThisVariable, SizeType QueueIndex, SizeType ThisPosition) const
     {
+        KRATOS_PROFILE_SCOPE(KRATOS_CODE_LOCATION);
         return *(reinterpret_cast<const TDataType*>(Position(QueueIndex) + ThisPosition) + rThisVariable.GetComponentIndex());
     }
 
@@ -372,6 +387,7 @@ public:
     template<class TDataType>
     const TDataType& FastGetCurrentValue(const Variable<TDataType>& rThisVariable, SizeType ThisPosition) const
     {
+        KRATOS_PROFILE_SCOPE(KRATOS_CODE_LOCATION);
         return *(reinterpret_cast<TDataType*>(mpCurrentPosition + ThisPosition) + rThisVariable.GetComponentIndex());
     }
 
@@ -678,7 +694,7 @@ public:
     {
         KRATOS_DEBUG_ERROR_IF(!mpVariablesList) << "This container don't have a variables list assigned. A possible reason is creating a node without a model part." << std::endl;
         for(VariablesList::const_iterator it_variable = mpVariablesList->begin(); it_variable != mpVariablesList->end() ; it_variable++) {
-            it_variable->AssignZero(mpCurrentPosition + LocalOffset(*it_variable)); 
+            it_variable->AssignZero(mpCurrentPosition + LocalOffset(*it_variable));
         }
     }
```

### Environment

- OS: `5.15.94-1-MANJARO`
- python (doesn't really matter): `3.10.9`
- processor: `AMD Ryzen 9 5900X`
- compiler: `clang`
- CMake build type: `RelWithDebInfo`

### External profiler

I used `prof` to profile the run, generating `dwarf` call graphs and restricting the sampling rate (again, to reduce output size). 
```
perf record --call-graph dwarf -F 100 python MainKratos.py
```

## Results

Here are `perf`'s outputs for 1 and 12 threads. After unzipping, you can browse them with whatever `perf` output parser you fancy (I recommend [hotspot](https://github.com/KDAB/hotspot)). If you take a look at the flame graph and filter for `Profiler`, you'll see that every `Profiler` leaf span is covered by a call to `std::chrono::_V2::system_clock::now` of identical width. As far as I understand, this means that all samples that have `Profiler` in their call tree originate from `std::chrono::_V2::system_clock::now`. In other words, no other operation of `Profiler` was sampled at all. 

[`OMP_NUM_THREADS=1`](https://github.com/KratosMultiphysics/Kratos/files/10834515/perf_single_threaded.data.zip)

[`OMP_NUM_THREADS=12`](https://github.com/KratosMultiphysics/Kratos/files/10834516/perf_12_threads.data.zip)

rubenzorrilla (dismissed):
Considering that the overhead is minimal @KratosMultiphysics/technical-committee agrees with the addition. Can you please remove your block @loumalouomega ?

Thanks for the implementation.

loumalouomega (dismissed):

loumalouomega (dismissed):

kratos/sources/profiler.cpp
```diff
@@ -0,0 +1,285 @@
...
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
+//
+//  Main authors:    Máté Kelemen
+//
+
+// --- Core Includes ---
```
loumalouomega: Would you mind using the standard notation for includes?

matekelemen: Sure. This started out as a copy-pasta from one of my pet projects, so there might still be some stuff that aren't in Kratos style. Have a look plz.

kratos/sources/profiler.cpp
```diff
@@ -0,0 +1,285 @@
...
+// STL includes
+#include <cstdio>
+#include <cstring>
+#include <algorithm>
+#include <thread>
+#include <vector>
+#include <tuple>
+#include <fstream>
+#include <sstream>
+#include <atomic>
```
sunethwarna: We have the `System Includes`, `Project includes`.... Could you change those?

loumalouomega (approved):

ddiezrod:
I tried to use this in my application but I get a linker error (I have checked and this happens every time I call this in any app out of the core). Am I missing something?

```
thermomechanic_small_displacement.cpp.obj : error LNK2019: unresolved external symbol "public: static class Kratos::Internals::Profiler<class std::chrono::duration<__int64,struct std::ratio<1,1000000> > > & __cdecl Kratos::Internals::ProfilerSingleton<class std::chrono::duration<__int64,struct std::ratio<1,1000000> > >::Get(void)" (?Get@?$ProfilerSingleton@V?$duration@_JU?$ratio@$00$0PECEA@@std@@@chrono@std@@@Internals@Kratos@@SAAEAV?$Profiler@V?$duration@_JU?$ratio@$00$0PECEA@@std@@@chrono@std@@@23@XZ) referenced in function "protected: virtual class Kratos::array_1d<double,3> __cdecl Kratos::ThermomechanicSmallDisplacement::GetBodyForce(class std::vector<class Kratos::IntegrationPoint<3,double,double>,class std::allocator<class Kratos::IntegrationPoint<3,double,double> > > const &,unsigned __int64)const " (?GetBodyForce@ThermomechanicSmallDisplacement@Kratos@@MEBA?AV?$array_1d@N$02@2@AEBV?$vector@V?$IntegrationPoint@$02NN@Kratos@@V?$allocator@V?$IntegrationPoint@$02NN@Kratos@@@std@@@std@@_K@Z)
thermomechanic_small_displacement.cpp.obj : error LNK2019: unresolved external symbol "public: class Kratos::Internals::Profiler<class std::chrono::duration<__int64,struct std::ratio<1,1000000> > >::Item & __cdecl Kratos::Internals::Profiler<class std::chrono::duration<__int64,struct std::ratio<1,1000000> > >::Create(class Kratos::CodeLocation &&)" (?Create@?$Profiler@V?$duration@_JU?$ratio@$00$0PECEA@@std@@@chrono@std@@@Internals@Kratos@@QEAAAEAVItem@123@$$QEAVCodeLocation@3@@Z) referenced in function "protected: virtual class Kratos::array_1d<double,3> 
__cdecl Kratos::ThermomechanicSmallDisplacement::GetBodyForce(class std::vector<class Kratos::IntegrationPoint<3,double,double>,class std::allocator<class Kratos::IntegrationPoint<3,double,double> > > const &,unsigned __int64)const " (?GetBodyForce@ThermomechanicSmallDisplacement@Kratos@@MEBA?AV?$array_1d@N$02@2@AEBV?$vector@V?$IntegrationPoint@$02NN@Kratos@@V?$allocator@V?$IntegrationPoint@$02NN@Kratos@@@std@@@std@@_K@Z)
kratos\applications\ThermomechanicalApplication\KratosThermomechanicalCore.dll : fatal error LNK1120: 2 unresolved externals
```

matekelemen:
> I tried to use this in my application but I get a linker error (I have checked and this happens every time I call this in any app out of the core). Am I missing something?
> 
> ```
> thermomechanic_small_displacement.cpp.obj : error LNK2019: unresolved external symbol "public: static class Kratos::Internals::Profiler<class std::chrono::duration<__int64,struct std::ratio<1,1000000> > > & __cdecl Kratos::Internals::ProfilerSingleton<class std::chrono::duration<__int64,struct std::ratio<1,1000000> > >::Get(void)" (?Get@?$ProfilerSingleton@V?$duration@_JU?$ratio@$00$0PECEA@@std@@@chrono@std@@@Internals@Kratos@@SAAEAV?$Profiler@V?$duration@_JU?$ratio@$00$0PECEA@@std@@@chrono@std@@@23@XZ) referenced in function "protected: virtual class Kratos::array_1d<double,3> __cdecl Kratos::ThermomechanicSmallDisplacement::GetBodyForce(class std::vector<class Kratos::IntegrationPoint<3,double,double>,class std::allocator<class Kratos::IntegrationPoint<3,double,double> > > const &,unsigned __int64)const " (?GetBodyForce@ThermomechanicSmallDisplacement@Kratos@@MEBA?AV?$array_1d@N$02@2@AEBV?$vector@V?$IntegrationPoint@$02NN@Kratos@@V?$allocator@V?$IntegrationPoint@$02NN@Kratos@@@std@@@std@@_K@Z)
> thermomechanic_small_displacement.cpp.obj : error LNK2019: unresolved external symbol "public: class Kratos::Internals::Profiler<class std::chrono::duration<__int64,struct std::ratio<1,1000000> > >::Item & __cdecl Kratos::Internals::Profiler<class std::chrono::duration<__int64,struct std::ratio<1,1000000> > >::Create(class Kratos::CodeLocation &&)" (?Create@?$Profiler@V?$duration@_JU?$ratio@$00$0PECEA@@std@@@chrono@std@@@Internals@Kratos@@QEAAAEAVItem@123@$$QEAVCodeLocation@3@@Z) referenced in function "protected: virtual class Kratos::array_1d<double,3> 
> __cdecl Kratos::ThermomechanicSmallDisplacement::GetBodyForce(class std::vector<class Kratos::IntegrationPoint<3,double,double>,class std::allocator<class Kratos::IntegrationPoint<3,double,double> > > const &,unsigned __int64)const " (?GetBodyForce@ThermomechanicSmallDisplacement@Kratos@@MEBA?AV?$array_1d@N$02@2@AEBV?$vector@V?$IntegrationPoint@$02NN@Kratos@@V?$allocator@V?$IntegrationPoint@$02NN@Kratos@@@std@@@std@@_K@Z)
> kratos\applications\ThermomechanicalApplication\KratosThermomechanicalCore.dll : fatal error LNK1120: 2 unresolved externals
> ```

**argh** I forgot about `KRATOS_API` on windows. See  #10975.

PR #10561, opened 2022-12-15
[FluidDynamicsApplication] Old header updated

loumalouomega:
**📝 Description**

Old header updated

**🆕 Changelog**

- [Old header updated](https://github.com/KratosMultiphysics/Kratos/commit/05a4cc72c47662ebef7effbe699c5814ee32327a)


rubenzorrilla:
This is legacy code. I prefer leaving it as it is to make it clear in the history.

loumalouomega:
> This is legacy code. I prefer leaving it as it is to make it clear in the history. Thanks in any case.

Okay, I just saw the header and said, this is old!

PR #10563, opened 2022-12-16
[Core][MPI] Adding RemoveOrphanNodesFromSubModelPartsto clean up submodelparts partitions

loumalouomega:
**📝 Description**

Looking for the failing tests in #10487 I found that the submodelparts in the partitions may be empty (no elements/conditions) but still have nodes. This adds RemoveOrphanNodesFromSubModelParts so we can clean up that submodelparts

This doesn't fix my problem, but I share the changes if it is interesting

**🆕 Changelog**

- [Adding RemoveOrphanNodesFromSubModelParts](https://github.com/KratosMultiphysics/Kratos/commit/1e2f53e9c8aaf99674b61a483ebf4dcf098a21ce)
- [Using for MPI partitioning](https://github.com/KratosMultiphysics/Kratos/commit/400787c27784fea298a0d6b41048910cf3dee7e5)
- [Refactor](https://github.com/KratosMultiphysics/Kratos/commit/da39dd4debd039c044b4c80b34268424c4e5e14c)

### DIFF
--- a/kratos/mpi/python_scripts/distributed_import_model_part_utility.py
+++ b/kratos/mpi/python_scripts/distributed_import_model_part_utility.py
@@ -92,7 +92,6 @@ class DistributedImportModelPartUtility:
                     partitioner = KratosMetis.MetisDivideHeterogeneousInputInMemoryProcess(model_part_io, serial_model_part_io, self.comm , domain_size, verbosity, sync_conditions)
                     partitioner.Execute()
                     serial_model_part_io.ReadModelPart(self.main_model_part)
-                    KratosMultiphysics.AuxiliarModelPartUtilities(self.main_model_part).RemoveNodesFromSubModePartsWithoutCorrespondingEntities()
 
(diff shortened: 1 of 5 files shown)


philbucher:
hm this is not a trivial change IMO

I would only add it where you actually need it 

And more so, I would say that your algorithm should also work without this.

loumalouomega:
> hm this is not a trivial change IMO
> 
> I would only add it where you actually need it

By default, is not activated

> And more so, I would say that your algorithm should also work without this.

In fact my problem is not solved, so is not this. But it is strange that this happens

rubenzorrilla (commented):
I completely agree with @philbucher 's observations. In any case, I still think that the #10570 might be useful so I'd suggest to close this PR to only leave #10570.

loumalouomega:
> I completely agree with @philbucher 's observations. In any case, I still think that the #10570 might be useful so I'd suggest to close this PR to only leave #10570.

Okay

PR #10564, opened 2022-12-16
[HDF5] Derive HDF5 IO Operations from Kratos::Operation

matekelemen:
## Brief
IO Operations in HDF5 didn't share a common base class until now, but now they all derive from `Kratos::Operation`. Changes do not propagate outside the `HDF5Application`, except for renaming integration-point-related IO operations from `*integration_point*` to `*gauss_point*`. 

## Changes
- all operations in `python_scripts/core/operations` now derive from `Kratos::Process`
- arguments to operations must be passed in their constructors
- operations are carried out by invoking their `Execute` member function rather than `__call__`
- on the JSON level, operations are referred to by the snake_case counterparts of their CamelCase class names:
  - `element_integration_point_output` is renamed to `element_gauss_point_output` (API change)
  - `condition_integration_point_output` is renamed to `condition_gauss_point_output` (API change)
- add `ProcessInfoInput` and `ProcessInfoOutput`
- update related tests

The most significant change is that the objects required to carry out IO tasks (model part, HDF5 file) must be passed and stored in the constructor, because `Operation::Execute` does not take any arguments.

Instead of hard-coding a map between operations' snake case names and their camel case counterparts, an automatic conversion and search is carried out (which is why `*integration_point*` operations were renamed to `*gauss_point*`).

Unfortunately, tests in `HDF5Application` are tied closely to implementation details so they had to be updated (or in some cases even deleted).


applications/HDF5Application/python_scripts/core/operations/model_part.py
```diff
@@ -28,267 +32,475 @@ def Prefix(pattern, model_part, time_format=''):
...
 
 
-class ModelPartInput:
-    '''Reads a model part from a file.'''
+class IOOperation(KratosMultiphysics.Operation): # IOOperation(KratosMultiphysics.Operation, metaclass = abc.ABCMeta) # <== conflicting metaclasses
+    """ @brief Base class for HDF5 IO operations."""
+
+    # @abc.abstractmethod # <== missing the ABCMeta metaclass
+    def Execute(self) -> None:
+        pass
```
sunethwarna: One way to get around this is to put `raise RuntimeError` here. So the derrived class forsure needs to implement it.

sunethwarna (commented):
Thanks @matekelemen. I think as we discussed, this can be simplified by having a master operation which can have list of sub-operations.

applications/HDF5Application/python_scripts/core/operations/model_part.py
```diff
@@ -299,104 +325,132 @@ class PrimalBossakOutput(VariableIO):
...
-        primal_io.SetAlphaBossak(self.alpha_bossak)
-        primal_io.WriteNodalResults(model_part)
+    def Execute(self) -> None:
+        primal_io = KratosHDF5.HDF5NodalSolutionStepBossakIO(self.parameters, self.file)
+        primal_io.SetAlphaBossak(self.__alpha_bossak)
+        primal_io.WriteNodalResults(self.model_part)
 
+    @staticmethod
+    def GetDefaultParameters() -> KratosMultiphysics.Parameters:
+        parameters = super(PrimalBossakOutput, PrimalBossakOutput).GetDefaultParameters()
```
sunethwarna:
```suggestion
        parameters = super().GetDefaultParameters()
```

applications/HDF5Application/python_scripts/core/operations/model_part.py
```diff
@@ -299,104 +325,132 @@ class PrimalBossakOutput(VariableIO):
...
+    subclasses = base_class.__subclasses__()
+    for subclass in base_class.__subclasses__():
+        subclasses += GetSubclasses(subclass)
+    return subclasses
+
+
+class AggregateOperation(KratosMultiphysics.Operation):
+    """ @brief Class for aggregating HDF5 IO operations on the same file."""
+
+    def __init__(self, model: KratosMultiphysics.Model, parameters: typing.Union[KratosMultiphysics.Parameters, ParametersWrapper]):
```
sunethwarna, line 389: May be in future we can get rid of the `ParameterWrapper`. :D

applications/HDF5Application/python_scripts/core/operations/model_part.py
```diff
@@ -299,104 +325,132 @@ class PrimalBossakOutput(VariableIO):
...
+        self.__operations: "list[tuple[type, KratosMultiphysics.Parameters]]" = []
+        for i in range(parameters["list_of_operations"].size()):
+            self.Add(parameters["list_of_operations"][i])
+
+    def Execute(self) -> None:
+        with OpenHDF5File(self.__io_parameters, self.__model_part) as file:
+            for operation, operation_parameters in self.__operations:
+                operation(self.__model_part, operation_parameters, file).Execute()
+
+    def Add(self, operation_parameters: KratosMultiphysics.Parameters) -> None:
```
sunethwarna: Is there a reason why this `Add` is public? Are we gonna allow it? Does it make sense?

applications/HDF5Application/python_scripts/core/operations/model_part.py
```diff
@@ -299,104 +325,132 @@ class PrimalBossakOutput(VariableIO):
...
+        }""")
+
+    @staticmethod
+    def __SnakeToCamel(string: str) -> str:
+        return "".join(part.title() for part in string.split('_'))
+
+    @staticmethod
+    def __CamelToSnake(string: str) -> str:
+        capital_indices = [0] + [index for index, char in enumerate(string) if char.isupper()] + [len(string)]
+        return "_".join(string[begin:end].lower() for begin, end in zip(capital_indices[:-1], capital_indices[1:]) if begin != end)
```
sunethwarna: I would use the one from the core and implement the one which is missing in the core
https://github.com/KratosMultiphysics/Kratos/blob/fd0c366d716422c4e1abbab18ae050faedb64df8/kratos/utilities/string_utilities.h#L58

sunethwarna (dismissed):
Minor comments. Thanks @matekelemen

applications/HDF5Application/python_scripts/core/operations/model_part.py
matekelemen: I need to explicitly pass the class type because `GetDefaultParameters` is a `staticmethod` and thus cannot implicitly provide `self` for `super` (though it'd be much cleaner if `GetDefaultParameters` was instead a `classmethod` so I'll make that change).

applications/HDF5Application/python_scripts/core/operations/model_part.py
matekelemen: I'd very much like that. `ParametersWrapper` is a neat idea but unless `Parameters` gets replaced by it in every python script, it's just a pain to deal with.

applications/HDF5Application/python_scripts/core/operations/model_part.py
philbucher: but ...
but ...
this one is so ugly and unreadable xD

matekelemen: If you're into it, I proudly recommend my marvelous one-liner [solutions](https://github.com/matekelemen/advent-of-code/tree/master/21) to last year's *Advent of Code*.

matekelemen: @sunethwarna I'll replace these if/when #10695 gets merged.

applications/HDF5Application/python_scripts/core/operations/model_part.py
```diff
@@ -398,17 +398,17 @@ def __init__(self, model: KratosMultiphysics.Model, parameters: typing.Union[Kra
...
             for operation, operation_parameters in self.__operations:
                 operation(self.__model_part, operation_parameters, file).Execute()
 
-    def Add(self, operation_parameters: KratosMultiphysics.Parameters) -> None:
+    def __Add(self, operation_parameters: KratosMultiphysics.Parameters) -> None:
         # Convert input snake case name to the internal camel case name
-        operation_type = self.__SnakeToCamel(operation_parameters["operation_type"].GetString())
-        operation: type = next((op for op in GetSubclasses(IOOperation) if op.__name__ == operation_type), None)
+        operation_type = KratosMultiphysics.StringUtilities.ConvertSnakeCaseToCamelCase(operation_parameters["operation_type"].GetString())
+        operation: typing.Type[IOOperation] = next((op for op in GetSubclasses(IOOperation) if op.__name__ == operation_type), None)
```
philbucher, line 411:
```suggestion
        operation: IOOperation = next((op for op in GetSubclasses(IOOperation) if op.__name__ == operation_type), None)
```

why the ugly version?
same above

matekelemen: `operation: IOOperation` means that `operation` is an **instance** of `IOOperation`

`operation: typing.Type[IOOperation]` means that `operation` is the **type** `IOOperation` itself

The latter is the case here, so that's why I went with it.

philbucher: pythons type annotations 

Gotta love them 😅

matekelemen:
ping
@sunethwarna
@philbucher

applications/HDF5Application/python_scripts/core/operations/model_part.py
```diff
@@ -28,267 +35,286 @@ def Prefix(pattern, model_part, time_format=''):
...
-        KratosHDF5.HDF5ModelPartIO(
-            hdf5_file, prefix).ReadModelPart(model_part)
+        # Pick necessary entries from the input parameters
+        # (assuming they're all present)
+        self.__parameters = KratosMultiphysics.Parameters()
+        self.__parameters.AddString("prefix", Prefix(parameters["prefix"].GetString(), self.model_part, self.time_format))
 
+    #@abc.abstractmethod # <== missing the ABCMeta metaclass
+    def Execute(self) -> None:
+        raise RuntimeError("Call to a pure abstract method")
```
sunethwarna:
```suggestion
        raise NotImplementedError("Call to a pure abstract method")
```

applications/HDF5Application/python_scripts/core/operations/model_part.py
```diff
@@ -28,267 +35,286 @@ def Prefix(pattern, model_part, time_format=''):
...
 
-class ModelPartOutput:
-    '''Writes a model part to a file.'''
+    #@abc.abstractclassmethod # <== missing the ABCMeta metaclass
+    @classmethod
+    def GetDefaultParameters(cls: "typing.Type[IOOperation]") -> KratosMultiphysics.Parameters:
```
sunethwarna: Do we need to type hint the `cls`? It is just like the `self` which is always deduced automatically ryt?

sunethwarna (dismissed):
Minor comments only. Thanks @matekelemen. Could you also check whether the `create_xdmf_file.py` is also working? There are no tests for that process because at that time CI did not have `h5py`.

applications/HDF5Application/python_scripts/core/operations/model_part.py
sunethwarna: Additionally, this nicely works with VSCode as well, and it does not grey out the rest of the code after calling this if it is throwing `NotImplementedError`, otherwise, it greys out the rest of the code after a call to this method.

applications/HDF5Application/python_scripts/core/operations/model_part.py
matekelemen: tru

matekelemen:
[create_xdmf_file](https://github.com/KratosMultiphysics/Kratos/blob/8e90b1dd7e3f793bd12bfc2f8e85facd72c6edda/applications/HDF5Application/python_scripts/create_xdmf_file.py) works, but the XDMF output processes (such as [single_mesh_xdmf_output_process](https://github.com/KratosMultiphysics/Kratos/blob/8e90b1dd7e3f793bd12bfc2f8e85facd72c6edda/applications/HDF5Application/python_scripts/single_mesh_xdmf_output_process.py)) have plenty of issues that I don't think are related to this PR.
- the HDF5 operation is followed up by an XDMF operation that **inherits the file access mode of the original output operation**, which will not work in most cases (`exclusive` flat out crashes, `truncate` silently empties the file, and `read_write` is the only one that should work).
- the XDMF operation globs previously written files, but something's broken with the globbing. For example, it fails to detect any file whose name (without suffix) ends with a `5` (eg.: `output5.h5`) - weird.
- due to the mentioned globbing, all previously generated XDMFs are regenerated at every time step. This doesn't break the output, but it does mean that the XDMF output's computational cost will increase linearly over the duration of the analysis.

These issues should definitely be fixed, but not in this PR.

matekelemen:
oh god, this is still open :(
can someone approve it plz?

philbucher:
Will have a look today

applications/HDF5Application/tests/test_hdf5_core_mpi.py
```diff
@@ -1,18 +1,40 @@
...
-    return model_part
+    # Get input model part path
+    script_directory      = pathlib.Path(__file__).absolute().parent
+    kratos_root_directory = script_directory.parent.parent.parent
+    test_input_directory  = kratos_root_directory / "kratos" / "tests" / "auxiliar_files_for_python_unittest"
+    test_file_stem        = test_input_directory / "mdpa_files" / "test_processes"
+    test_file_path        = pathlib.Path(str(test_file_stem) + ".mdpa")
+
+    if not test_file_path.is_file():
+        raise FileNotFoundError("Test file not found: {}".format(test_file_path))
```
philbucher: Fstring

applications/HDF5Application/tests/test_hdf5_core_mpi.py
```diff
@@ -27,26 +49,22 @@ def test_FilenameGetterWithDirectoryInitialization_DirectoryExists(self, mock_cl
...
-        data_comm = KratosMultiphysics.Testing.GetDefaultDataCommunicator()
-        obj = file_io._FilenameGetterWithDirectoryInitialization(settings, data_comm)
-        obj.Get(_SurrogateModelPart())
-        makedirs.assert_called_once_with('/foo')
-        patcher.stop()
+        with patch('KratosMultiphysics.FilesystemExtensions.MPISafeCreateDirectories', autospec=True) as makedirs:
+            data_comm = KratosMultiphysics.Testing.GetDefaultDataCommunicator()
+            obj = file_io._FilenameGetterWithDirectoryInitialization(settings, data_comm)
+            obj.Get(_SurrogateModelPart()[1])
+            makedirs.assert_called_once_with('/foo')
```
philbucher, line 56: Hm probably better to use MPISafeCreateDirectories

applications/HDF5Application/tests/test_hdf5_processes.py
```diff
@@ -287,8 +289,8 @@ def test_SingleMeshPrimalOutputProcess(self):
         self.assertEqual(self.HDF5NodalSolutionStepBossakIO.call_count, 3)
         self.assertEqual(
             self.HDF5NodalSolutionStepBossakIO.call_args[0][0]['prefix'].GetString(), '/ResultsData')
-        self.assertEqual(
-            self.HDF5NodalSolutionStepBossakIO.call_args[0][0]['list_of_variables'].size(), 0)
+        #self.assertEqual(
+        #    self.HDF5NodalSolutionStepBossakIO.call_args[0][0]['list_of_variables'].size(), 0)
```
philbucher: Please remove commented code

philbucher (dismissed):
Not really sure what is going on but I trust you :)

applications/HDF5Application/tests/test_hdf5_core_mpi.py
matekelemen: I didn't change this test apart from scoping the patcher such that it doesn't bleed outside the function if there's an assertion failure.

Do you want me to change the original test to use `MPISafeCreateDirectories` directly instead of mocking it?

sunethwarna (approved):
Thanks @matekelemen :)

PR #10565, opened 2022-12-16
[Core][FastPR] Fix OMP issue in particles utilities

pablobecker:
**📝 Description**
Fixes the ocassional failures in the particles utilities test. See https://stackoverflow.com/questions/39037239/c-openmp-and-stdvectorbool

**🆕 Changelog**
- Adds an aux int vector that stores if the elem was found or not
-  then the int vector is stored in the bool vector to be returned, to avoid changing the interface

### DIFF
--- a/kratos/python/add_vector_to_python.cpp
+++ b/kratos/python/add_vector_to_python.cpp
@@ -452,10 +452,12 @@ void  AddVectorToPython(pybind11::module& m)
     py::implicitly_convertible<array_1d<double,3>, Vector>();
 
     //***********************************************************************************
-    auto bool_vector_binder = CreateVectorInterface<DenseVector<bool>>(m, "DenseVectorBool");
-    bool_vector_binder.def(py::init<typename DenseVector<bool>::size_type>());
-    bool_vector_binder.def(py::init<typename DenseVector<bool>::size_type, bool>());
-    bool_vector_binder.def(py::init<DenseVector<bool>>());
+        //***********************************************************************************
+    py::class_< DenseVector<bool>, std::shared_ptr<DenseVector<bool>> > bool_vector_binder(m,"BoolVector", py::buffer_protocol());
+    bool_vector_binder.def(py::init<>());
+    bool_vector_binder.def(py::init<const DenseVector<bool>&>());
+    bool_vector_binder.def(py::init< DenseVector<bool>::size_type>());
+    bool_vector_binder.def(py::init< DenseVector<bool>::size_type, bool>());
     bool_vector_binder.def(py::init( [](const py::list& input)
     {
         DenseVector<bool> tmp(input.size());
@@ -463,32 +465,34 @@ void  AddVectorToPython(pybind11::module& m)
             tmp[i] = py::cast<bool>(input[i]);
         return tmp;
     }));
-    bool_vector_binder.def(py::init( [](py::buffer b)
+    bool_vector_binder.def("Size", [](const DenseVector<bool>& self)
     {
-        py::buffer_info info = b.request();
-        KRATOS_ERROR_IF( info.format != py::format_descriptor<typename DenseVector<bool>::value_type >::value ) << "Expecting a bool vector\n";
-        KRATOS_ERROR_IF( info.ndim != 1 ) << "Buffer dimension of 1 is required, got: " << info.ndim << std::endl;
-        DenseVector<bool> vec(info.shape[0]);
-
-        std::copy(
-            static_cast<const bool*>(info.ptr),
-            static_cast<const bool*>(info.ptr) + info.shape[0],
-            vec.begin()
-        );
-
-        return vec;
-    }));
-    bool_vector_binder.def_buffer( [](DenseVector<bool>& self)-> py::buffer_info
+        return self.size();
+    } );
+    bool_vector_binder.def("Resize", [](DenseVector<bool>& self, const typename DenseVector<bool>::size_type  new_size)
     {
-        return py::buffer_info(
-            self.data().begin(),
-            sizeof(typename DenseVector<bool>::value_type),
-            py::format_descriptor<typename DenseVector<bool>::value_type>::format(),
-            1,
-        {self.size()},
-        {sizeof(typename DenseVector<bool>::value_type)}
-        );
-    });
+        if(self.size() != new_size) self.resize(new_size, false);
+    } );
+    bool_vector_binder.def("__len__", [](const DenseVector<bool>& self)
+    {
+        return self.size();
+    } );
+    bool_vector_binder.def("__setitem__", [](DenseVector<bool>& self,
+                                            const unsigned int i,
+                                            const bool value)
+    {
+        self[i] = value;
+    } );
+    bool_vector_binder.def("__getitem__", [](const DenseVector<bool>& self, const unsigned int i)
+    {
+        return self[i];
+    } );
+    bool_vector_binder.def("__iter__", [](DenseVector<bool>& self)
+    {
+        return py::make_iterator(self.begin(), self.end(), py::return_value_policy::reference_internal);
+    }, py::keep_alive<0,1>() ) ;
+    bool_vector_binder.def("__str__", PrintObject<DenseVector<bool>>);
+
     py::implicitly_convertible<py::list, DenseVector<bool>>();
 
     //***********************************************************************************
--- a/kratos/python/add_vector_to_python.cpp
+++ b/kratos/python/add_vector_to_python.cpp
@@ -465,7 +465,7 @@ void  AddVectorToPython(pybind11::module& m)
     bool_vector_binder.def(py::init( [](py::buffer b)
     {
         py::buffer_info info = b.request();
-        KRATOS_ERROR_IF( info.format != py::format_descriptor<typename DenseVector<bool>::value_type >::value ) << "Expected a double array\n";
+        KRATOS_ERROR_IF( info.format != py::format_descriptor<typename DenseVector<bool>::value_type >::value ) << "Expecting a bool vector\n";
         KRATOS_ERROR_IF( info.ndim != 1 ) << "Buffer dimension of 1 is required, got: " << info.ndim << std::endl;
         DenseVector<bool> vec(info.shape[0]);
 
--- a/kratos/python/add_vector_to_python.cpp
+++ b/kratos/python/add_vector_to_python.cpp
@@ -469,9 +469,11 @@ void  AddVectorToPython(pybind11::module& m)
         KRATOS_ERROR_IF( info.ndim != 1 ) << "Buffer dimension of 1 is required, got: " << info.ndim << std::endl;
         DenseVector<bool> vec(info.shape[0]);
 
-        for( int i=0; i<info.shape[0]; ++i ) {
-            vec[i]= static_cast<typename DenseVector<bool>::value_type *>(info.ptr)[i];
-        }
+        std::copy(
+            static_cast<const bool*>(info.ptr),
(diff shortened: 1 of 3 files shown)


kratos/utilities/particles_utilities.h
```diff
@@ -252,17 +252,18 @@ class ParticlesUtilities
         typename BinBasedFastPointLocator<TDim>::ResultContainerType TLS(max_results);
 
         auto interpolations = std::make_pair(std::vector<bool>(rCoordinates.size1()), std::vector<TDataType>(rCoordinates.size1()));
+        std::vector<int> is_found_aux_vector(rCoordinates.size1()); //aux vector using ints to allow running in omp
```
matekelemen:
```suggestion
        std::vector<char> is_found_aux_vector(rCoordinates.size1()); //aux vector using ints to allow running in omp
```

`int` is an overkill for this purpose

kratos/utilities/particles_utilities.h
```diff
@@ -280,6 +281,11 @@ class ParticlesUtilities
             }
 
         });
+        
+        //storing info in bool vector (do not use omp!)
+        for(unsigned int i=0; i<rCoordinates.size1(); i++){
+            (interpolations.first)[i] = static_cast<bool>(is_found_aux_vector[i]);
+        }
```
matekelemen:
```suggestion
        std::copy(is_found_aux_vector.begin(), is_found_aux_vector.end(), interpolations.first.begin());
```

kratos/utilities/particles_utilities.h
```diff
@@ -252,17 +252,18 @@ class ParticlesUtilities
...
-        IndexPartition(rCoordinates.size1()).for_each(TLS, [&rLocator, &rCoordinates, &interpolations, &rInterpolationVariable, &zero, SearchTolerance](const auto& i, auto& rTLS)
+        IndexPartition(rCoordinates.size1()).for_each(TLS, [&rLocator, &rCoordinates, &interpolations, &is_found_aux_vector, &rInterpolationVariable, &zero, SearchTolerance](const auto& i, auto& rTLS)
         {
 
             Vector shape_functions;
             Element::Pointer p_element;
             const bool is_found = rLocator.FindPointOnMesh(row(rCoordinates,i), shape_functions, p_element, rTLS.begin(), rTLS.size(), SearchTolerance);
 
-            (interpolations.first)[i] = is_found;
+            is_found_aux_vector[i] = static_cast<int>(is_found);
```
matekelemen:
```suggestion
            is_found_aux_vector[i] = static_cast<char>(is_found);
```

matekelemen (commented):
I've no idea how deep this function is used in the app, but if you have a bit of wiggle room with the return type, consider changing it to a vector of pairs instead of the current pair of vectors. It would eliminate the race condition and cut memory allocations in half.
```diff
- std::pair<std::vector<bool>, std::vector<T>>
+ std::vector<std::pair<bool,T>>
```

pablobecker:
> I've no idea how deep this function is used in the app, but if you have a bit of wiggle room with the return type, consider changing it to a vector of pairs instead of the current pair of vectors. It would eliminate the race condition and cut memory allocations in half.
> 
> ```diff
> - std::pair<std::vector<bool>, std::vector<T>>
> + std::vector<std::pair<bool,T>>
> ```

not my code, it has a python interface and the test is in python, so I assume that it's main usage. would that work with the python interface? I don't know if it allows it.

in the meantime i'll follow your suggestion and switch to char

ping @RiccardoRossi

kratos/utilities/particles_utilities.h
loumalouomega: This is more efficient, but using char for not text is strange for me

loumalouomega:
@pablobecker , @matekelemen is right, char is more efficient despite being a weird choice IMHO because we are not writing anything

pablobecker:
changed to char following @matekelemen 's review. I can confirm it's faster ( I run the test 10k times in fulldebug and it's noticeable faster)

matekelemen:
> not my code, it has a python interface and the test is in python, so I assume that that is its main usage. would that work with the python interface? I don't know if it allows it.

*pybind11* can bind standard data structures out-of-the box, so no problem there. The question is how @RiccardoRossi wants to use the output data. Unless he plans to use something that takes exactly a vector of `TDataType` s, it's the best solution.

kratos/utilities/particles_utilities.h
RiccardoRossi: we  could return DenseVector< bool>

kratos/utilities/particles_utilities.h
```diff
@@ -252,17 +252,18 @@ class ParticlesUtilities
         typename BinBasedFastPointLocator<TDim>::ResultContainerType TLS(max_results);
 
         auto interpolations = std::make_pair(std::vector<bool>(rCoordinates.size1()), std::vector<TDataType>(rCoordinates.size1()));
+        std::vector<char> is_found_aux_vector(rCoordinates.size1()); //aux vector using ints to allow running in omp
```
RiccardoRossi: let's return DenseVector<bool>

this is a stupid mistake from my side. Sorry about this

pablobecker:
compiles but crashes

TypeError: Unregistered type : boost::numeric::ublas::vector<bool, boost::numeric::ublas::unbounded_array<bool, std::allocator<bool> > >

pablobecker:
will try this Did you forget to `#include <pybind11/stl.h>`? Or <pybind11/complex.h>,
<pybind11/functional.h>, <pybind11/chrono.h>, etc. Some automatic
conversions are optional and require extra headers to be included
when compiling your pybind11 module.

pablobecker:
tests passed 10k times, thx @loumalouomega !

kratos/utilities/particles_utilities.h
```diff
@@ -276,26 +274,48 @@ class ParticlesUtilities
...
+     * @param rLocator this is a search structure for the volume
+     * @param rCoordinates positions at which we want to retrieve the interpolation values
+     * @param rInterpolationVariable: variable whose value that will be interpolated
+     * @param SearchTolerance: search tolerance used in the spatial search
+     * @return the function returns:
+           - is_inside a vector of bools telling if the given coordinate falls within the volume
+           - values the interpolated values (only valid if the corresponding value of is_inside is true)
+    */
+    template<unsigned int TDim, class TDataType, bool InterpolationVariableHasHistory>
+    static std::pair<std::vector<bool>, std::vector<TDataType> > InterpolateValuesAtCoordinates(
```
RiccardoRossi: here you are using again std::vector bool

RiccardoRossi (changes requested):
this PR makes a copy which should be avoided!!

kratos/utilities/particles_utilities.h
loumalouomega: It is because pybind, doesn't work if not using std::vector or any compatible class. In all cases copy is necessary

RiccardoRossi:
the correct fix is to export the DenseVector<bool> to python in 
add_vector_to_python.cpp

loumalouomega:
> the correct fix is to export the DenseVector to python in add_vector_to_python.cpp

That makes sense

RiccardoRossi:
in principle we need to add this to add_vector_to_python.cpp

    //***********************************************************************************
    auto bool_vector_binder = CreateVectorInterface<DenseVector<bool>>(m, "BoolVector");
    bool_vector_binder.def(py::init< DenseVector<bool>::size_type>());
    bool_vector_binder.def(py::init< DenseVector<bool>::size_type, bool>());
    bool_vector_binder.def(py::init( [](const py::list& input)
    {
        DenseVector<bool> tmp(input.size());
        for(unsigned int i=0; i<tmp.size(); ++i)
            tmp[i] = py::cast<bool>(input[i]);
        return tmp;
    }));
    py::implicitly_convertible<py::list, DenseVector<bool>>();

loumalouomega:
> in principle we need to add this to add_vector_to_python.cpp
> 
> ```
> //***********************************************************************************
> auto bool_vector_binder = CreateVectorInterface<DenseVector<bool>>(m, "BoolVector");
> bool_vector_binder.def(py::init< DenseVector<bool>::size_type>());
> bool_vector_binder.def(py::init< DenseVector<bool>::size_type, bool>());
> bool_vector_binder.def(py::init( [](const py::list& input)
> {
>     DenseVector<bool> tmp(input.size());
>     for(unsigned int i=0; i<tmp.size(); ++i)
>         tmp[i] = py::cast<bool>(input[i]);
>     return tmp;
> }));
> py::implicitly_convertible<py::list, DenseVector<bool>>();
> ```

Done

kratos/python/add_vector_to_python.cpp
```diff
@@ -455,6 +450,44 @@ void  AddVectorToPython(pybind11::module& m)
...
+    {
+        DenseVector<bool> tmp(input.size());
+        for(unsigned int i=0; i<tmp.size(); ++i)
+            tmp[i] = py::cast<bool>(input[i]);
+        return tmp;
+    }));
+    bool_vector_binder.def(py::init( [](py::buffer b)
+    {
+        py::buffer_info info = b.request();
+        KRATOS_ERROR_IF( info.format != py::format_descriptor<typename DenseVector<bool>::value_type >::value ) << "Expected a double array\n";
```
matekelemen:
```suggestion
        KRATOS_ERROR_IF( info.format != py::format_descriptor<typename DenseVector<bool>::value_type >::value ) << "Expecting a bool vector\n";
```

kratos/python/add_vector_to_python.cpp
```diff
@@ -455,6 +450,44 @@ void  AddVectorToPython(pybind11::module& m)
...
+        KRATOS_ERROR_IF( info.ndim != 1 ) << "Buffer dimension of 1 is required, got: " << info.ndim << std::endl;
+        DenseVector<bool> vec(info.shape[0]);
+
+        for( int i=0; i<info.shape[0]; ++i ) {
+            vec[i]= static_cast<typename DenseVector<bool>::value_type *>(info.ptr)[i];
+        }
```
matekelemen:
```suggestion
        std::copy(
            static_cast<const bool*>(info.ptr),
            static_cast<const bool*>(info.ptr) + info.shape[0],
            vec.begin()
        );
```

matekelemen (commented):
I still recommend returning `std::vector<std::pair<bool,TDataType>>`.

kratos/python/add_vector_to_python.cpp
```diff
@@ -455,6 +451,50 @@ void  AddVectorToPython(pybind11::module& m)
     py::implicitly_convertible<py::list, Vector>();
     py::implicitly_convertible<array_1d<double,3>, Vector>();
 
+    //***********************************************************************************
+        //***********************************************************************************
+    py::class_< DenseVector<bool>, std::shared_ptr<DenseVector<bool>> > bool_vector_binder(m,"BoolVector", py::buffer_protocol());
```
loumalouomega, line 456: I was going to add some constexpr to use the already existing definitions and minimize code duplication

loumalouomega (approved):
@RiccardoRossi you can remove now the block

kratos/python/add_vector_to_python.cpp
RiccardoRossi: well, the vector interface could be definitely be improved...however i would do it in a different Pr. What do you think?

RiccardoRossi (approved):

roigcarlo:
ping @loumalouomega

loumalouomega:
> ping @loumalouomega

The PR was from @pablobecker, which is currently in a plane to Argentina

roigcarlo:
oh, I assumed it was yours :P

PR #10566, opened 2022-12-16
[Core] Adding copy constructor to `BinBasedFastPointLocator`

loumalouomega:
**📝 Description**

Adding copy constructor to `BinBasedFastPointLocator`

**🆕 Changelog**

- [Adding copy constructor to `BinBasedFastPointLocator`](https://github.com/KratosMultiphysics/Kratos/commit/64516ffc8ccfae2378586535de3d20fdc5e23e4a)


kratos/utilities/binbased_fast_point_locator.h
```diff
@@ -109,11 +107,40 @@ class BinBasedFastPointLocator
...
+        UpdateSearchDatabaseAssignedSize(mCellSize);
     }
 
     /// Destructor.
     virtual ~BinBasedFastPointLocator() = default;
 
+    /// Copy constructor.
+    BinBasedFastPointLocator(BinBasedFastPointLocator const& rOther)
+        : mrModelPart(rOther.mrModelPart),
+          mCellSize(rOther.mCellSize)
```
roigcarlo: Why not calling BinsDynamicObject's: `rOther.GetCellSize()` instead of forwarding that parameter to this class?

loumalouomega: Huuuum

loumalouomega: mCellSize in BinsObjectDynamic is an array, not a double, not possible

roigcarlo: but I don't get what you are trying to do, mCellSize is an array because the cells of the Bins are cubes.

If you are copying the underlying Bins from the utility, and you only forward the int, then you are imposing the cells to have the same size, which is not correct in many cases (elongated models, planes in 3D, etc...),.

By passing the double, you are wrongly initializing the bins using this constructor:

```c++
  BinsObjectDynamic (const PointType& MinPoint, const PointType& MaxPoint, SizeType NumPoints)
      : mObjectsSize(0), mObjectsBegin(0), mObjectsEnd(0) {

    for(SizeType i = 0; i < Dimension; i++) {
      mMinPoint[i] = MinPoint[i];
      mMaxPoint[i] = MaxPoint[i];
    }

    CalculateCellSize(NumPoints);                                 // Calculate number of Cells
    AllocateContainer();                                         // Allocate cell list
  }
```

Which allocates the number of points, not the one that allocates the size of the cells.

loumalouomega: I was simply trying to do a consistent copy constructor with the existent methods

kratos/utilities/binbased_fast_point_locator.h
```diff
@@ -145,7 +172,7 @@ class BinBasedFastPointLocator
      * @brief Function to construct or update the search database
      * @param CellSize The current size of the cell used for search
      */
-    void UpdateSearchDatabaseAssignedSize(double CellSize)
+    void UpdateSearchDatabaseAssignedSize(const double CellSize)
```
loumalouomega: To be consistent with this @roigcarlo

roigcarlo: Well that method is also wrong :S

roigcarlo: O dear... I think every time that is used, its wrong....

loumalouomega: ¯\_(ツ)_/¯

roigcarlo: Most likely its a copy paste mistake because the nodes in element locator uses a StaticBins inside, which uses the Bucket size simple `std::size_t`.

Should I fix it myself or can you change it?

loumalouomega: Feel free to modify

kratos/utilities/binbased_fast_point_locator.h
```diff
@@ -151,7 +156,7 @@ class BinBasedFastPointLocator
      * @brief Function to construct or update the search database
      * @param CellSize The current size of the cell used for search
      */
-    void UpdateSearchDatabaseAssignedSize(const double CellSize)
+    void UpdateSearchDatabaseAssignedSize(const BinsCoordinateType CellSize)
```
roigcarlo, line 159: FYI @RiccardoRossi. This was wrong. Seems that nobody was using it but calling it would have updated the number of cells with the size of the cell. Probably a copy paste mistake from the other locator which uses a Bins that has a constructor with `BucketSize`

roigcarlo:
All clear from my side if compiles.

roigcarlo (approved):

PR #10562, opened 2022-12-15
[Core] Improve JsonOutputProcess

loumalouomega:
**📝 Description**

Improve JsonOutputProcess. Avoiding write when not required (not elements, not GP). Reduce code duplication. Improve dump json.

**🆕 Changelog**

- [Improve JsonOutputProcess](https://github.com/KratosMultiphysics/Kratos/commit/5a9591e37b38ce44d2247ee62dbbb58bb164963b)
- Avoiding write when not required (not elements, not GP)
- Reduce code duplication
- [Improve dump json](https://github.com/KratosMultiphysics/Kratos/pull/10562/commits/71cd223858de44760d3a95243a8954164f388010)

PR #10567, opened 2022-12-16
[All] Cleaning empty json elements using REGEX like a pro

loumalouomega:
**📝 Description**

Until now (https://github.com/KratosMultiphysics/Kratos/pull/10562) when generating results in JSON of empty values were added filling of garbage the json result files. This clean up that. I was cleaning by hand until I remembered that REGEX exists...

**🆕 Changelog**

- [Cleaning (I)](https://github.com/KratosMultiphysics/Kratos/commit/4523b69e1a277e2e6bacc422d458b4f1bd0261cf)
- [Using REGEX like a pro](https://github.com/KratosMultiphysics/Kratos/commit/91c5ba94bdb7db06a20d6f6fd6f64135182deae0)
- [Remove also empty nodes](https://github.com/KratosMultiphysics/Kratos/pull/10567/commits/65ec22091d5da0f0535b4668405c3c494fba702d)


applications/IgaApplication/tests/coupling_condition_tests/two_patch_refined_test/two_patch_refined_test_lagrange_membrane_result.json
```diff
@@ -125,29 +125,29 @@
...
+    
+    
+    
+    
+    
+    
+    
+    
+    
     "ELEMENT_25": {}
```
matekelemen, line 152: line breaks are still there, and the last entries seem to be ignored

matekelemen (commented):
https://stackoverflow.com/a/1732454/12350793 -- same for JSON; it's probably safer to just parse the JSONs, get rid of the junk, and write them anew.

Also, I doubt anyone will honestly review changes to ~200 single-line JSON files.

applications/IgaApplication/tests/coupling_condition_tests/two_patch_refined_test/two_patch_refined_test_lagrange_membrane_result.json
loumalouomega: I cleaned that too

rubenzorrilla (changes requested):
To be discussed in @KratosMultiphysics/technical-committee

RiccardoRossi:
@KratosMultiphysics/technical-committee is not positive about this large automatic refactorings unless there is an important reason. Our point of view is that the risk is not worth the advantage the change may bring onto the table as they may easily introduce bugs.

PR #10568, opened 2022-12-18
[CoSimIO] update to CoSimIO 4.2.0

philbucher:
Minor fixes for compilation with old Linux kernels

FYI @mpentek 
FYI @roigcarlo I think you reported some problems with this, but I couldn't find any more where


roigcarlo (approved):
:smile: Yep, that was a pain in the ass in some WSL systems. Thx!

PR #10569, opened 2022-12-19
[Core][FastPR] New wall flag

rubenzorrilla:
**📝 Description**
New (long term missing) `WALL` flag addition. This will make the application of any wall model much easier in all our CFD-related applications. Note that there is no similar flag in `kratos_flags.h` that could be used instead.


rubenzorrilla:
Adding to the explanation above that the option of using `BOUNDARY` or `INTERFACE` is not a good idea since these have been historically used for FSI or unfitted type applications. Note that the option of using one of these two and leave the other for such applications is IMO not a good idea since there are applications on which these two need to coexist (e.g. methods with true and surrogate boundaries).

About the possibility of only using the `SLIP` flag, which is what has been historically done in some places. Note that there are contributions that must be only added in case of a pure slip boundary, which is what one likely expects when doing `Set(SLIP,true)`. Hence, any contribution lying between the slip and no-slip limits, that is to say any CFD-related wall stuff, needs to be tagged with a different indicator. In other words, using `SLIP` for both makes the usage of pure-slip and "sticky" walls more complicated.

Finally, there is `STRUCTURE`. This in my opinion must be used in presence of domains that are actually a structure, which basically means in the coupled-problems (e.g. FSI or CHT) context.

Of course this is my personal view, and, at the end of the day any flag can be used for any purpose. However, I think it is important to use them with some coherence to favor code sustainability and collaboration.

rubenzorrilla:
BTW, I've just confirmed that there is no local flag in Kratos at this same 33 position.

pooyan-dadvand (approved):
Approving as @KratosMultiphysics/technical-committee  agrees.

PR #10575, opened 2022-12-19
[Shape opt] Curvature based adaptive vertex morphing

DSchmoelz:
**📝 Description**
Introduces a new vertex morphing mapping method which adaptively changes filter radii similar to [[Antonau]](https://link.springer.com/content/pdf/10.1007/s00158-022-03279-w.pdf?pdf=button), but according to the gaussian curvature and not the element sizes at the node. 

Three different curvature computation methods have been implemented and are chosen accordingly to the element types in the node neighborhood:
- (I) Quadratic Elements: Curvature tensor from shape functions
- (II) Bi-Linear Quadrilateral: [Taubin](http://graphics.stanford.edu/courses/cs348a-17-winter/ReaderNotes/taubin-iccv95b.pdf) / [Camprubi Estebo](https://mediatum.ub.tum.de/doc/601055/00000039.pdf)
- (III) Linear Triangle: [Meyer](https://authors.library.caltech.edu/99186/2/diffGeoOps.pdf)

If multiple different element types are around the node, the precedence of the methods is (I) > (II) > (III).

Four different functions which map the gaussian curvature to the filter radius size are available:
- "analytic"
- linear
- square root
- fourth root

Shell optimization : comparison  (left: constant radius, right: adaptive radius)
comparison_2

**🆕 Changelog**
- Added `"adaptive"` filter to ShapeOptimizationApplication
- Added Gaussian curvature computation to `geometry_utilities` in ShapeOptimizationApplication


DSchmoelz:
Could somebody please review?

applications/ShapeOptimizationApplication/custom_utilities/geometry_utilities.cpp
```diff
@@ -359,4 +382,353 @@ void GeometryUtilities::CalculateUnitNormals()
     }
 }
 
+void GeometryUtilities::CalculateGaussianCurvature() {
+
+    KRATOS_TRY
+
+    // check if surface normals are necessary (only for Taubin method)
```
Igarizza, line 410: Is it possible to incorporate this loop check into the code inside if statements below. For instance, if technique Taubin and the normalsHasn't computed -> compute Normals and set flag, NormalsComputed True. The same for Meyer and edges.

applications/ShapeOptimizationApplication/custom_utilities/geometry_utilities.cpp
```diff
@@ -359,4 +382,353 @@ void GeometryUtilities::CalculateUnitNormals()
...
+
+    block_for_each(mrModelPart.Nodes(), [&](NodeType &rNode) {
+
+        const auto& r_condition_neighbours = rNode.GetValue(NEIGHBOUR_CONDITIONS).GetContainer();
+        double& r_gaussian_curvature = rNode.FastGetSolutionStepValue(GAUSSIAN_CURVATURE);
+
+        std::string technique = this->GetCurvatureTechnique(rNode);
+
+        // For quadratic elements (e.g. 6NTriangle)
+        if (technique == "curvature_tensor") {
```
Igarizza, line 444: Can you add 3 functions: ComputeCurvatureTensorforNode(), ComputeMeyerforNode(), ComputeTaubinForNode() and replace all part of codes inside the if statements?? It should improve the code and add modularity. At least the computation part itself.

Igarizza:
@DSchmoelz, very nice PR and a nice feature. Few code improvements from my side.

applications/ShapeOptimizationApplication/custom_utilities/geometry_utilities.cpp
DSchmoelz: It's a good idea, though I am not sure how to modify the parallel loop then. Because `ExtractEdgeNodes` can't be executed twice or in parallel and `ComputeUnitSurfaceNormals` probably not as well.

applications/ShapeOptimizationApplication/custom_utilities/geometry_utilities.cpp
DSchmoelz: done

applications/ShapeOptimizationApplication/tests/test_ShapeOptimizationApplication.py
```diff
@@ -87,7 +87,7 @@ def AssembleTestSuites():
     smallSuite.addTest(curvature_6NTriangle_test('test_execution'))
     smallSuite.addTest(curvature_4NQuad_test('test_execution'))
     smallSuite.addTest(curvature_8NQuad_test('test_execution'))
-    smallSuite.addTest(mapper_adaptive_filter_curvature_test('test_execution'))
+    # smallSuite.addTest(mapper_adaptive_filter_curvature_test('test_execution'))
```
Igarizza: may be you can delete it or you want to save ??

applications/ShapeOptimizationApplication/custom_utilities/geometry_utilities.cpp
```diff
@@ -359,4 +382,353 @@ void GeometryUtilities::CalculateUnitNormals()
...
+    KRATOS_TRY
+
+    // check if surface normals are necessary (only for Taubin method)
+    bool normals_necessary = false;
+    for (const auto& rNode : mrModelPart.Nodes()) {
+        std::string technique = this->GetCurvatureTechnique(rNode);
+        if (technique == "Taubin") {
+            normals_necessary = true;
+            break;
+        }
```
Igarizza, line 417: Will it work??: 
elif (technique == "Meyer") {
             edge_necessary = true;
             break;
         }
if (normals_necessary) this->ComputeUnitSurfaceNormals();
if (edge_necessary) {
         ModelPart& r_edge_model_part = mrModelPart.HasSubModelPart(mrModelPart.Name() + "_edges") ? mrModelPart.GetSubModelPart(mrModelPart.Name() + "_edges") : mrModelPart.CreateSubModelPart(mrModelPart.Name() + "_edges");
         if (r_edge_model_part.Nodes().size() == 0) {
             this->ExtractEdgeNodes(mrModelPart.Name() + "_edges");
         }
     }

applications/ShapeOptimizationApplication/python_scripts/mapper_factory.py
```diff
@@ -39,9 +39,36 @@ def CreateMapper(origin_model_part, destination_model_part, mapper_settings):
         "revolution_settings"        : {
             "point" : [0.0, 0.0, 0.0],
             "normal": [0.0, 0.0, 0.0]
+        },
+        "adaptive_filter_settings"   : {
```
Igarizza, line 43: What do you think about adding "adaptive_filter_method"?? see Shape Module.

Then, we can have different methods later, how we compute the radius: based on curvature or element size or smth else in future.

Igarizza (dismissed):
Minor comments from my side. Nothing blocking

applications/ShapeOptimizationApplication/custom_utilities/geometry_utilities.cpp
DSchmoelz: Unfortunately not, because the loop will break, before it detects a node which has to be calculated by the other technique. 

Let's say we have 1000 nodes with Taubin method and 1000 nodes with Meyer technique. Only the technique of the first node will be detected.

Igarizza (approved):

PR #10577, opened 2022-12-19
[IO] Simplify partitioning interfaces

philbucher:
This moves an auxiliary struct to the baseclass

My main motivation is to extend the partitioning to Geometries, which without this change would require many interface changes. By adding it to the struct it can be easily achieved

The changes are fully backward compatible, I will update the ModelPartIO in a followup PR


loumalouomega (dismissed):
I was thinking about this when understanding the partitioning while debugging MPI

loumalouomega (approved):

rubenzorrilla (approved):
Makes sense.

PR #10578, opened 2022-12-19
[Core] DomainSize is not needed for Metis

philbucher:
The domain size is not used, I realized this during my recent refactorings

I will remove the leftovers soon within the scope of some other refactorings, but for now we can lift some restrictions in the tests :)


loumalouomega (approved):
Thanks

rubenzorrilla (approved):
OK.

PR #10580, opened 2022-12-20
[PfemFluid] Rotation Process edit

AFranci:
Small changes to the already existent rotation process. It has been made compatible with the GiD interface input (modified in another PR). Initial and final time instants are now taken from interval of time provided from the GID interface.


ipouplana (approved):
Great!

PR #10581, opened 2022-12-20
[MappingApplication] Adding corner case to NearestNeighbor

loumalouomega:
**📝 Description**

@philbucher here a PR that presents the corner case found in #10487, to show that the problem is coming from the base mapper and not my implementation. You can see in the following image that the node is failing is one in the middle which haves a 50% of possibilities of fall in one side or the other:

image

The problem appears when using MPI cores > 7 (in some cases work, but for example with 8 always fails)

**🆕 Changelog**

- [Adding corner case to NearestNeighbor](https://github.com/KratosMultiphysics/Kratos/commit/e6f39e7beeea057ee555ad3e1c41e708672c74b0)
- [Adding meshes from https://github.com/KratosMultiphysics/Kratos/pull/10487](https://github.com/KratosMultiphysics/Kratos/commit/c9b4d623b8ce0e0a74c8d911beeb679530821186)
- [Adding to other tests](https://github.com/KratosMultiphysics/Kratos/commit/bda0badc28c51045ec1a3c0d58a480622ddf9183)
- [More generic](https://github.com/KratosMultiphysics/Kratos/commit/1af6cb425efb6f9c621e01865de7d34fb2a579f9)


philbucher:
Ohhh I see 

I think I know how to fix it, should be easy, can take a look at some point over the holidays

loumalouomega:
> Ohhh I see
> 
> I think I know how to fix it, should be easy, can take a look at some point over the holidays

Thanks, in that case I will focus on my other developments. Have nice holidays!

loumalouomega:
> > Ohhh I see
> > I think I know how to fix it, should be easy, can take a look at some point over the holidays
> 
> Thanks, in that case I will focus on my other developments. Have nice holidays!

I have though and may have a solution. Do you think it would be a good solution to average the values in the case of same distance nodes?, this would imply change a bit in the implementation (and it would resemble to the nearest element implementation), but at least the result would be consistent.

philbucher:
Yes I also thought sth like this would make sense

I will do it in the next days, I think it should be relatively simple

loumalouomega:
> Yes I also thought sth like this would make sense
> 
> I will do it in the next days, I think it should be relatively simple

Relatively simple, but a behaviour change at the end. Luckily, cases like this are very rare

loumalouomega:
The changes from this PR were included in #10487, so now is merged with master. Tests can be found at test_nearest_neighbor_mapper.CornerCaseNearestNeighbor. Closing PR

philbucher:
@loumalouomega can you try if [this banch](https://github.com/KratosMultiphysics/Kratos/tree/mapping/nearest-neighbors-improvement) solves your issues?

I haven't tested it myself in detail, but it should now take all the closest neighbors into account if they have the same distance

loumalouomega:
> @loumalouomega can you try if [this banch](https://github.com/KratosMultiphysics/Kratos/tree/mapping/nearest-neighbors-improvement?rgh-link-date=2023-01-02T21%3A48%3A46Z) solves your issues?
> 
> I haven't tested it myself in detail, but it should now take all the closest neighbors into account if they have the same distance

I will

PR #10582, opened 2022-12-20
[CLApp] Using characteristic length on Reference configuration in small strain Damage/Plasticity

AlejandroCornejo:
**📝 Description**

That's it, now iso damage, iso plasticity, kinematic plasticity and plastic-damage will use a characteristic length calculated according to the reference configuration. This improves convergence rates since it is always constant.

@SergioJimenezReyes @atfard @lagoncalvesjr @balcayde


loumalouomega:
This has been a problem from the beginning. What I am afraid is on the increase of the computational cost (which was always the problem)

loumalouomega (dismissed):

SergioJimenezReyes:
@AlejandroCornejo, can you change the HCF CL too? I ve seen that you have updated the test according to the damage cl changes, but it would be nice to update the hcf law too.

AlejandroCornejo:
> @AlejandroCornejo, can you change the HCF CL too? I ve seen that you have updated the test according to the damage cl changes, but it would be nice to update the hcf law too.

DONE

SergioJimenezReyes (approved):
Already approved by @loumalouomega. Just updated the HCF CL.

PR #10584, opened 2022-12-21
[Core] SkinDetectionProcess MPI - bugfix - communication gets stuck if processor is empty

ddiezrod:
If rSetNodeIdsInterface  is empty in a processor it does not enter these lines and the communication later gets stuck in MPI


loumalouomega (approved):
Thanks

PR #10571, opened 2022-12-19
[Fluid] Navier-Stokes wall modelling

rubenzorrilla:
**📝 Description**
This adds the possibility of considering different wall models in the `NavierStokesWallCondition` thanks to the addition of a new variadic `TWallModel` argument. One implementation of a simple Navier-slip wall is included too to add a use case (the more elaborated linear-log one present in the legacy condition will be ported in a future PR). Note that the 2-fluids case, is also left for the future.

The value updating in the old tests comes after setting a non-unit dynamic viscosity as suggested by @sunethwarna in a previous PR. Note that we're using the same auxiliary function to generate the testing condition.

This requires merging #10569 first.

PR #10585, opened 2022-12-21
[Fluid] Linear-log wall law refactor

rubenzorrilla:
**📝 Description**
This PR refactors the linear-log law implementation in the old `MonolithicWallCondition` to be compatible with the current `NavierStokesWallCondition`. Current implementation has been checked against the old one by calculating the RHS and LHS with both conditions. Identical results are obtained.

Requires  #10571 to be merged.


ddiezrod:
There is one trick you can use to make the review easier: you can use fluid/navier-stokes-wall-modelling as target branch here so you will only see the changes from this PR and when fluid/navier-stokes-wall-modelling is merged to master, github will automatically change here the target branch to master too. We do it like this in Gitlab and I think it is quite handy.

rubenzorrilla:
> There is one trick you can use to make the review easier: you can use fluid/navier-stokes-wall-modelling as target branch here so you will only see the changes from this PR and when fluid/navier-stokes-wall-modelling is merged to master, github will automatically change here the target branch to master too. We do it like this in Gitlab and I think it is quite handy.

Nice. I'll try the next time. Thanks for the hint :wink:.

applications/FluidDynamicsApplication/custom_conditions/wall_laws/linear_log_wall_law.h
```diff
@@ -0,0 +1,382 @@
...
+
+///@}
+///@name  Enum's
+///@{
+
+///@}
+///@name  Functions
+///@{
+
+///@}
```
sunethwarna:
```suggestion
```

I would remove the empty groups.

applications/FluidDynamicsApplication/custom_conditions/wall_laws/linear_log_wall_law.h
```diff
@@ -0,0 +1,382 @@
...
+
+    /// Destructor.
+    ~LinearLogWallLaw() = default;
+
+    ///@}
+    ///@name Operators
+    ///@{
+
+
+    ///@}
```
sunethwarna:
```suggestion
```

Removing empty groups

sunethwarna (dismissed):
Thanks @rubenzorrilla :D

sunethwarna (approved):

PR #10591, opened 2022-12-22
Add rebuild level to geomechanics

### LINKED ISSUE #10067: [GeoMechanicsApplication] make "rebuild_level" adjustable in a dynamic solver strategy in GeoMechanicsApplication
currently dynamic calculations in GeoMechanicsApplication are very slow since the LHS is rebuild at each iteration.

In a linear elastic case, the LHS is constant thus has to be build only once, which greatly improves the speed in dynamic calculations

either make the "rebuild_level" property adjustable in a dynamic solver strategy in GeoMechanicsApplication. Or check if a strategy in another Application can be used for this purpose

aronnoordam:
rebuild level of LHS can now be set in geomechanics application. 

default setting is rebuild_level: 2, such that nothing breaks


Vahid-Galavi (approved):
Very good!

PR #10597, opened 2022-12-23
[GeomechanicsApplication] Correct geomechanics tests

### LINKED ISSUE #10593: [GeomechanicsApplication] Von Misis stress cannot be outputted on gauss points when UDSM interface elements are present
Von Misis stress cannot be outputted on gauss points when UDSM interface elements are present. 

Is this expected behaviour?

aronnoordam:
corrected and unignored ignored tests in Geomechanics application.

added calculation for VON_MISES_STRESS and MEAN_EFFECTIVE_STRESS on interface elements gauss points


applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_interface_element.cpp
```diff
@@ -575,58 +577,74 @@ void UPwSmallStrainInterfaceElement<TDim,TNumNodes>::
...
+        this->InterpolateOutputDoubles(rValues, GPValues);
+    } else if (rVariable == MEAN_EFFECTIVE_STRESS) {
+        //Loop over integration points
+        for (unsigned int GPoint = 0; GPoint < NumGPoints; ++GPoint) {
+            StressStrainUtilities EquivalentStress;
+            GPValues[GPoint] = EquivalentStress.CalculateMeanStress(mStressVector[GPoint]);
+        }
+    
+        this->InterpolateOutputDoubles(rValues, GPValues);
+    } else if (rVariable == MEAN_STRESS ||
```
Vahid-Galavi: MEAN_STRESS is the summation of the effective mean stress and pore water pressure. It is not not always zero.

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_interface_element.cpp
```diff
@@ -693,27 +704,20 @@ void UPwSmallStrainInterfaceElement<TDim,TNumNodes>::
 
         InterfaceElementVariables Variables;
         const PropertiesType& rProp = this->GetProperties();
-        const GeometryType& rGeom = this->GetGeometry();
-        const unsigned int NumGPoints = rGeom.IntegrationPointsNumber(mThisIntegrationMethod);
-        std::vector<double> GPValues(NumGPoints);
 
         this->InitializeElementVariables(Variables,
-            rGeom,
```
Vahid-Galavi, line 701: usually we use a prefix "r" to indicate that the variable is a reference. In the same function we have rProp. Why did you remove "r"?

applications/GeoMechanicsApplication/custom_elements/geo_linear_truss_element.cpp
```diff
@@ -129,8 +129,8 @@ void GeoLinearTrussElement<TDim,TNumNodes>::
         ConstitutiveLaw::Parameters Values(this->GetGeometry(),
                                            this->GetProperties(),
                                            rCurrentProcessInfo);
-        Vector temp_strain = ZeroVector(1);
```
Vahid-Galavi, line 132: stress and strains have always size 1.

Vahid-Galavi (commented):
Please look at my comments

applications/GeoMechanicsApplication/custom_elements/geo_linear_truss_element.cpp
aronnoordam: mStressVectorSize is set to 1 in  the .h file, I changed "1" to "mStressVectorSize" in order to stay consistent with "mInternalStressesFinalizedPrevious" as defined in "ResetConstitutiveLaw()". Like this, memory issues cannot occur when adding "temp_stress" and "mInternalStressesFinalizedPrevious"

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_interface_element.cpp
aronnoordam: I will change it back

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_interface_element.cpp
aronnoordam: I will add it

applications/GeoMechanicsApplication/custom_elements/geo_linear_truss_element.cpp
```diff
@@ -118,7 +118,7 @@ void GeoLinearTrussElement<TDim,TNumNodes>::
     }
 
     if (rVariable == FORCE) {
-        BoundedVector<double, TDim> truss_forces = ZeroVector(TDim);
+        BoundedVector<double, 3> truss_forces = ZeroVector(3);
```
mcgicjn2, line 121: Are the results always 3 dimensional even in the 2D case ? As this is a templated class. If that is the case, is the same true in other locations where TDim is used.  Perhaps we can remove the TDim templating (and hardcode 3 in more places) of this and the geo_cable_element, if they have the same structure structures regardless of dimension size. Also do we plan to have more than 2 noded versions of the 2 elements ?  (I don't quite see why we would need it)

mcgicjn2: Although, I guess length is an issue, but can't remember if the node points are still stored in 3D, if they are then that is equally not an issue

BTW this is just a discussion point and not 100% needed for the review

applications/GeoMechanicsApplication/custom_elements/geo_linear_truss_element_base.cpp
```diff
@@ -263,7 +263,20 @@ double GeoTrussElementLinearBase<TDim,TNumNodes>::CalculateLinearStrain()
...
+
+    double length_0;
+    if constexpr (TDim == 2)
+    {
+        length_0 = GeoStructuralMechanicsElementUtilities::CalculateReferenceLength2D2N(*this);
+    } else if constexpr (TDim == 3)
+    {
+        length_0 = GeoStructuralMechanicsElementUtilities::CalculateReferenceLength3D2N(*this);
+    }
+    else // 1D
```
mcgicjn2: I am not sure under which circumstance the "else" point will be touched as the only templates that appear to be possible are:

template class GeoTrussElementLinearBase<2,2>;
template class GeoTrussElementLinearBase<3,2>;

aronnoordam: yes you are right, this can be removed

applications/GeoMechanicsApplication/custom_elements/geo_linear_truss_element.cpp
aronnoordam: truss_forces is projected on rOuput which is a vector of a array_1d with size 3. To prevent memory issues, TDim in this case had to be changed to 3

applications/GeoMechanicsApplication/custom_elements/geo_linear_truss_element_base.cpp
Vahid-Galavi: In my opinion, else should be there when we have if{} else if{} . When 1D is not the case, the code should give an error message to prevent any problem in future if the element is extended to 1D geomtery.

mcgicjn2: @Vahid-Galavi Agreed, was thinking the same, @aronnoordam perhaps a not implemented error. Problem with leaving it in, is that a casual developer would assume it's implemented for 1D ..... Of course feel free to implement in 1D ;)

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_interface_element.cpp
```diff
@@ -575,68 +597,97 @@ void UPwSmallStrainInterfaceElement<TDim,TNumNodes>::
...
+    const GeometryType& rGeom = this->GetGeometry();
+    const unsigned int NumGPoints = rGeom.IntegrationPointsNumber(mThisIntegrationMethod);
+    std::vector<double> GPValues(NumGPoints);
+
+    //Printed on standard GiD Gauss points
+    const unsigned int OutputGPoints = rGeom.IntegrationPointsNumber(this->GetIntegrationMethod());
+    if (rValues.size() != OutputGPoints)
+        rValues.resize(OutputGPoints);
+
     // KRATOS_INFO("0-UPwSmallStrainInterfaceElement:::CalculateOnIntegrationPoints<double>()") << std::endl;
```
Vahid-Galavi: remove the unused line.

Vahid-Galavi (dismissed):

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_interface_element.cpp
aronnoordam: I've removed it

Vahid-Galavi:
```suggestion
  
```

aronnoordam: I've removed that line in my latest commit

aronnoordam: https://github.com/KratosMultiphysics/Kratos/pull/10597/commits/d5414e8f08012988041b7a57f61ac353a4f1e518

Vahid-Galavi (approved):

mcgicjn2 (approved):

PR #10599, opened 2022-12-23
[Core][Geometries] Reactivate methods for non-squared Jacobians

loumalouomega:
**📝 Description**

In geometries where local dimensions are smaller than the spatial dimension, the Jacobian is not squared. Until now, the following error was thrown:

~~~c++
KRATOS_ERROR << "Jacobian is not square" << std::endl;
~~~

Since we now use `GeneralizedDeterminant` and `GeneralizedInverse` this is not longer a limitation.

Includes #10546

**🆕 Changelog**

- Reactivate methods for non-squared Jacobians
- Includes #10546


rubenzorrilla (approved):
Considering that #10546, this makes pretty much sense too.

PR #10611, opened 2023-01-02
[Core][ModelPartIO] update deprecated interfaces

philbucher:
... and also update some IO interfaces regarding const-correctness


loumalouomega (approved):
Thanks

PR #10613, opened 2023-01-03
[MetisApplication] Inconsistent libraries linking in CMake

loumalouomega:
**📝 Description**

Inconsistent libraries linking in CMake. The variable defined and used in general is `METIS_LIBRARIES`, not `METIS_LIBRARY`

**🆕 Changelog**

- [Inconsistent libraries linking in CMake](https://github.com/KratosMultiphysics/Kratos/commit/bd076b7341104ce1fbbcf1ccdf5692ee9300ba2a)


philbucher:
But how was this working before? 🫣

loumalouomega:
I don't really know...

loumalouomega:
Anyone?

roigcarlo (approved):
It was working because `METIS_LIBRARY` is the internal var that the module uses to set `METIS_LIBRARIES`, but is better this way.

Most probably the FindMETIS.cmake is wrong anyway, part of the code there does not make any sense.

Anyway, this is better than what we had so...

PR #10614, opened 2023-01-04
[Core] Extend Kratos Parameters to support @include_json in array items

sunethwarna:
**📝 Description**
As stated in the title, this PR extends to support to have `@include_json` in array items in recursive manner. A test is also added.

**🆕 Changelog**
- Extended `@include_json` to support array items
- Added a test

PR #10615, opened 2023-01-04
[OptApp] Adds execution policy for analyses

sunethwarna:
**📝 Description**
This PR adds execution policies to run analyses so that, user can use `MultistageAnalysis` and their own custom `ExecutionPolicies` or `AnalysisStage` to analyse primal problems. This PR also removes the imports of other applications in `analyses_controller.py` file by making it generic to be used in any compiled application within `Kratos` as well as outside `Kratos`.

This also enables the use of `Operations` before and after an `analysis`.

The tests also have been updated.

**🆕 Changelog**
- Removes other application imports in `analyses_controller.py`
- Adds execution policies `IndependentAnalysisExecutionPolicy` and `SteppingAnalysisExecutionPolicy`
- Updates tests.

Requires #10614


applications/OptimizationApplication/tests/test_execution_policies.py
```diff
@@ -0,0 +1,32 @@
...
+
+class TestExecutionPolicies(kratos_unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.model = Kratos.Model()
+
+    def test_IndependentAnalysisExecutionPolicy(self):
+        parameters = Kratos.Parameters("""{
+            "name"                     : "test",
+            "execution_policy_type"    : "KratosMultiphysics.OptimizationApplication.execution_policies.independent_analysis_execution_policy.IndependentAnalysisExecutionPolicy",
```
RezaNajian: @sunethwarna can't we reduce/simplify the execution type ?! the input is long and might be confusing. As far as I understood, we have stepping and independent policies and why not to only say Stepping or Independent ?!
One more thing, let's set the stepping as the default type or ?!

applications/OptimizationApplication/tests/top_opt_test/optimization_parameters.json
```diff
@@ -1,114 +1,161 @@
...
-                                "gradient_mode": "semi_analytic"
-                            }                       
+            }
+        ],
+        "analyses": [
+            {
+                "name": "Structure_static",
+                "execution_policy_type": "KratosMultiphysics.OptimizationApplication.execution_policies.stepping_analysis_execution_policy.SteppingAnalysisExecutionPolicy",
+                "execution_policy_settings": {
+                    "analysis_type": "KratosMultiphysics.StructuralMechanicsApplication.structural_mechanics_analysis.StructuralMechanicsAnalysis",
```
RezaNajian: @sunethwarna The same point here, can't we simply say StructuralMechanicsAnalysis ?! 
Is there any specific reason to not use only the standard names of the applications or solvers ?!

applications/OptimizationApplication/python_scripts/execution_policies/execution_policy_wrapper.py
```diff
@@ -0,0 +1,86 @@
+# ==============================================================================
```
RezaNajian, line 1: I suggest renaming the file and class to AnalysisExecutionPolicy.... to be more specific and consistent with other classes

RezaNajian (commented):
@sunethwarna Thanks for asking me to review. This PR definitely generalizes and cleans AnalysisController.

applications/OptimizationApplication/tests/test_execution_policies.py
sunethwarna: The reasoning is, if we do it like this, users can have their own `ExecutionPolicy` made in the folder they have the optimization case for custom execution (like coupling with different solvers, OpenFOAM, etc to run some types of analysis). I can make it shorter by putting `module_name`, and `class_name` into seperate json strings.

applications/OptimizationApplication/tests/top_opt_test/optimization_parameters.json
sunethwarna: Same reasoning as before, users can have their custom `AnalysisStage` scripts at anywhere (doesn't need to be in Kratos) (such as what we have in MainKratos.py) to run an analysis in a custom way. Here also I can break it to two strings.

applications/OptimizationApplication/python_scripts/execution_policies/execution_policy_wrapper.py
sunethwarna: This class does not have an `ExecutionPolicy`, it just wraps around defined `ExecutionPolicy` such that, logging, pre, post operations are done automatically.

applications/OptimizationApplication/tests/test_execution_policies.py
sunethwarna: Would it be ok, if I shorten it but still keeping the flexibility?

applications/OptimizationApplication/tests/top_opt_test/optimization_parameters.json
sunethwarna: Would it be ok, if I shorten it but still keeping the flexibility?

applications/OptimizationApplication/tests/test_execution_policies.py
RezaNajian: Sure, it would be great.

applications/OptimizationApplication/tests/top_opt_test/optimization_parameters.json
RezaNajian: Sure, it would be great

RezaNajian (dismissed):

RezaNajian (approved):
Thanks Suneth for the updates.

PR #10617, opened 2023-01-04
[Core] Adding output process with controller

ddiezrod:
**📝 Description**
This is one of the possible additions we discussed in the last Kratos Workshop from Altair developments.

The idea is that output processes currently have two responsibilities: deciding when to print and what to print. Whenever you want to have a more complex control on when to print (e.g. every time a 1% of a part is filled), this is very inconvenient and separating the process into two different objects makes things much easier. So the point is separating the responsibilities between:
- controller: decided when to print, it only needs to implement IsOutputStep.
- print process: controls what to print and in which format. It implements the rest of the functions, especially the PrintOutput function. Note that any current output process (like gid_output_process) can be used here, it will just never call its IsOutputStep function.


ddiezrod:
To be discussed: @KratosMultiphysics/implementation-committee @KratosMultiphysics/technical-committee @KratosMultiphysics/altair @philbucher

matekelemen:
`Process`es and especially `OutputProcess`es have had a lot of revamps over the last year and it's always a pain to update everything to the new systems, so I'd appreciate it if we could settle for a long-term API.

As I see it, the `controller` you're describing is a predicate that scans the state of a `Model` and spits out a `bool`. The use cases of such an object aren't limited to IO processes, so I'd separate them into their own system and make them a bit more general (eg.: objects that can be constructed with an immutable `Model` and `Parameters`, and evaluate to a `bool`). *I had a PR with something along these lines.*

In my opinion, the part of the `Process` that actually performs the IO operation should also be delegated outside the ecosystem of `Process`es to clearly separate its responsibilities. `Process`es should just pick components and bundle them into a package that can be referenced from a JSON in a standardized manner, and not implement any functionality by themselves. In an ideal world, a single `OutputProcess` could write text, HDF5 or whatever other output format depending on how it's configured on the JSON level.

ddiezrod:
@matekelemen I agree with all your comments. Could you point me to the PR you mentioned so I have I more clear idea of what you mean?

matekelemen:
#10352

ddiezrod:
I think I get your point but it seems a little overcomplicated to me (not a real expert in c++ anyway). My idea was to create a base controller class with the basic interface, we can then maybe check that the created objects are derived from this one. I think that should be enough, although I do see the advantages of the piping system you implemented...

kratos/python_scripts/output_process_with_controller.py
```diff
@@ -0,0 +1,79 @@
...
+            settings.AddEmptyValue("Parameters")
+
+        kratos_module_name = settings["kratos_module"].GetString()
+        if not kratos_module_name.startswith("KratosMultiphysics"):
+            kratos_module_name = "KratosMultiphysics." + kratos_module_name
+
+        full_module_name = kratos_module_name + "." + python_module_name
+        python_module = import_module(full_module_name)
+
+        return python_module.Factory(settings, self.model)
```
rubenzorrilla: I'd create the controlers from the registry so we force the controlers to be registered in order to be usable.

rubenzorrilla (commented):
Some thoughts about:

- First, I'm eager to have this in the master branch 😃 
- As far as I understand from this proto-version, the input would be somehow a process (with it's current input) that has the complete input of two processes inside. Something of this sort
```json
"python_module" : "output_process_with_controller"
"kratos_module" : "KratosMultiphysics",
"Parameters" : {
    "controller_settings" : {
        "python_module" : "my_controller",
        "kratos_module" : "KratosMultiphysics.MyApplication",
        "Parameters" : {
            ....
        }
    },
    "print_process_settings" : {
        "python_module" : "my_output_process",
        "kratos_module" : "KratosMultiphysics.MyApplication",
        "Parameters" : {
            ....
        }
    }
```
If so, I think this is valuable as we can reuse much of current implementations. However, I think it could be a bit misleading for newbies. Besides, this will potentially result in a cumbersome implementation at the GUIs.

Instead, I'd register all the controllers and output tools so we just need to specify the registry name inside the `controller_settings` and `print_process_settings`. Internally we will of course use the two instances as you suggest.

- Very minor at this point, but I'd rename `print_process_settings` to `output_settings`. I think this is more Kratos-like.
- In line to @matekelemen's comments, I think that we should create a base c++ class for the controlers in order to state the API.

ddiezrod:
@rubenzorrilla I agree with your points but I am not very familiar with register so I am not sure how I should proceed.

I guess I could call the model_parameters_factory as you do in the multistage analysis and check later if the object is derived from OuputProcess/PrintController, or what did you have in mind?

rubenzorrilla:
Just to make it clear, my idea for the input would be something of this sort
```json
"python_module" : "output_process_with_controller"
"kratos_module" : "KratosMultiphysics",
"Parameters" : {
    "controller_settings" : {
        "name" : "KratosMultiphysics.OutputControllers.BaseOutputController",
        "Parameters" : {
            ....
        }
    },
    "print_process_settings" : {
        "name" : "KratosMultiphysics.OutputProcesses.GidOutputProcess",
        "Parameters" : {
            ....
        }
    }
```

About creating the instances of these two, depending on how we finally decide to register the stuff (we are about to finish this discussion) we would end up calling a factory function or a prototype. In any case, at this point we can assume that there is a mechanism that returns an instance provided the `"name"` field.

In this regard, I note that in my opinion the controllers input must be process-like, i.e. we pass a model and a json settings.

ddiezrod:
> Just to make it clear, my idea for the input would be something of this sort
> 
> ```json
> "python_module" : "output_process_with_controller"
> "kratos_module" : "KratosMultiphysics",
> "Parameters" : {
>     "controller_settings" : {
>         "name" : "KratosMultiphysics.OutputControllers.BaseOutputController",
>         "Parameters" : {
>             ....
>         }
>     },
>     "print_process_settings" : {
>         "name" : "KratosMultiphysics.OutputProcesses.GidOutputProcess",
>         "Parameters" : {
>             ....
>         }
>     }
> ```
> 
> About creating the instances of these two, depending on how we finally decide to register the stuff (we are about to finish this discussion) we would end up calling a factory function or a prototype. In any case, at this point we can assume that there is a mechanism that returns an instance provided the `"name"` field.
> 
> In this regard, I note that in my opinion the controllers input must be process-like, i.e. we pass a model and a json settings.

I think this goes in the same line as my last commit. My main doubt with register is if the same method can be called for python and cpp objects. (I am using `controller_prototype = KratosMultiphysics.Registry[f"{controller_name}.Prototype"]` here).

Thats why I thought that maybe I could call model_parameters_factory so we cover all options.

rubenzorrilla:
> > Just to make it clear, my idea for the input would be something of this sort
> > ```json
> > "python_module" : "output_process_with_controller"
> > "kratos_module" : "KratosMultiphysics",
> > "Parameters" : {
> >     "controller_settings" : {
> >         "name" : "KratosMultiphysics.OutputControllers.BaseOutputController",
> >         "Parameters" : {
> >             ....
> >         }
> >     },
> >     "print_process_settings" : {
> >         "name" : "KratosMultiphysics.OutputProcesses.GidOutputProcess",
> >         "Parameters" : {
> >             ....
> >         }
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > About creating the instances of these two, depending on how we finally decide to register the stuff (we are about to finish this discussion) we would end up calling a factory function or a prototype. In any case, at this point we can assume that there is a mechanism that returns an instance provided the `"name"` field.
> > In this regard, I note that in my opinion the controllers input must be process-like, i.e. we pass a model and a json settings.
> 
> I think this goes in the same line as my last commit. My main doubt with register is if the same method can be called for python and cpp objects. (I am using `controller_prototype = KratosMultiphysics.Registry[f"{controller_name}.Prototype"]` here).
> 
> Thats why I thought that maybe I could call model_parameters_factory so we cover all options.

This is exactly the idea of the registry, that the user doesn't need to care about cpp and py registered objects. This is why we created the [`python_registry.py`](https://github.com/KratosMultiphysics/Kratos/blob/84239308a893a185e13f1c60638fbe6265f2f655/kratos/python_interface/python_registry.py) layer. Indeed, the cpp registry is masked (and intentionally inaccessible) through this.

Long story short, provided that the `name` field is correct, we don't care if it is a Python or a cpp object :wink:.

rubenzorrilla (changes requested):
As this will be crucial for the future, I'd like to confirm our thoughts with the @KratosMultiphysics/technical-committee. In any case, I agree that we can merge this as it is right now.

rubenzorrilla:
@KratosMultiphysics/technical-committee agrees with the I/O suggested by @rubenzorrilla and @ddiezrod.

Besides, also agrees with having a base `Controller` class. These are our API suggestions for it:
- Constructor: `Model`-`Parameters`
- `bool Check()` : same as in any operation, process, ...
- `bool Evaluate()`: this is the equivalent to the `Execute` in the processes, that is to say, the method performing the evaluation of the controlling condition. OFC we're open to suggestions in this regard.

ddiezrod:
I adapted this process to use the controller interface.

kratos/python_scripts/output_process_with_controller.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    def __init__(self, model, params):
+        KratosMultiphysics.OutputProcess.__init__(self)
+        default_settings = self.GetDefaultParameters()
+        self.params = params
+        self.model = model
+        self.params.ValidateAndAssignDefaults(default_settings)
+        factory = KratosModelParametersFactory(self.model)
+        self.controller = factory.ConstructItem(self.params["controller_settings"])
+        self.print_process = factory.ConstructItem(self.params["output_settings"])
```
rubenzorrilla:
```suggestion
        self.output_process = factory.ConstructItem(self.params["output_settings"])
```
Very minor in the naming of the variables for the sake of consistency (feel free to dismiss).

rubenzorrilla (dismissed):
I think that this is basically what we need. Thanks for the implementation @ddiezrod.

kratos/python_scripts/output_process_with_controller.py
ddiezrod: done!

rubenzorrilla (approved):
:rocket:

PR #10619, opened 2023-01-05
[FastPR][Core] Add PointsNumber to Python

rubenzorrilla:
**📝 Description**
Exporting `PointsNumber` method to Python similar to the other methods.

### DIFF
--- a/kratos/python/add_mesh_to_python.cpp
+++ b/kratos/python/add_mesh_to_python.cpp
@@ -466,6 +466,7 @@ void  AddMeshToPython(pybind11::module& m)
     .def("SetValue", SetValueHelperFunction< Element, Variable< std::string > >)
     .def("GetValue", GetValueHelperFunction< Element, Variable< std::string > >)
 
+    .def("PointsNumber", [](Element& rThis){return rThis.GetGeometry().PointsNumber();})
     .def("GetNode", GetNodeFromElement )
     .def("GetNodes", GetNodesFromElement )
     .def("GetIntegrationPoints", GetIntegrationPointsFromElement )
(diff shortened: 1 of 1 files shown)


philbucher:
I would add this to Geometry to be consistent with C++

kratos/python/add_mesh_to_python.cpp
```diff
@@ -466,6 +466,7 @@ void  AddMeshToPython(pybind11::module& m)
     .def("SetValue", SetValueHelperFunction< Element, Variable< std::string > >)
     .def("GetValue", GetValueHelperFunction< Element, Variable< std::string > >)
 
+    .def("PointsNumber", [](Element& rThis){return rThis.GetGeometry().PointsNumber();})
```
philbucher, line 469: This is already possible with GetGeometry().PointsNumber(), hence I would not add this

rubenzorrilla: This was my first thought. However, all the geometry accessible methods are masked throught the condition and element. I've the feeling that this was done in the past to ease the usage from Python. Hence, I decided to do it like this to be consistent with the rest.

philbucher: Ever since we added the exposure of the Geometry for Elememts&Conditions we have added new things there

Afaik it is quite complete by now, hence I would go the opposite direction and deprecate the old style interfaces. Then the usage would be consistent from C++ and python

rubenzorrilla: Makes complete sense for me. I'll do it that way then.

rubenzorrilla:
Indeed, it was already there... :man_facepalming:. Sorry for the noise. At least we keep the decision documented :sweat_smile:

PR #10620, opened 2023-01-05
[Fluid] Apply wall law process

rubenzorrilla:
**📝 Description**
Following #10571 and #10585, this adds the possibility to use wal models from the input settings. As for the inlet, outlet, etc,  a custom process, potentially deriving from this one, must be created for the two-phase case (these will be done in future PR).

@jginternational once this is merged, we must work on adding this to the GUI.

~#10619 needs to be merged first.~

### DIFF
--- a/applications/FluidDynamicsApplication/python_scripts/apply_wall_law_process.py
+++ b/applications/FluidDynamicsApplication/python_scripts/apply_wall_law_process.py
@@ -1,7 +1,5 @@
 import KratosMultiphysics
-import KratosMultiphysics.FluidDynamicsApplication as KratosFluid
-
-from KratosMultiphysics import assign_vector_by_direction_process
+import KratosMultiphysics.FluidDynamicsApplication as KratosCFD
 
 def Factory(Settings, Model):
     if not isinstance(Settings, KratosMultiphysics.Parameters):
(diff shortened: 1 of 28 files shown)


applications/FluidDynamicsApplication/python_scripts/apply_wall_law_process.py
```diff
@@ -0,0 +1,152 @@
...
+            aux_props = condition.Properties
+            aux_cond_data_map[condition.Id] = (aux_ids, aux_props)
+        # 2. Remove current conditions
+        # Note that in here we need to assume that no submodelpart is hanging from this one as the hierarchical structure would be destroyed
+        if model_part.NumberOfSubModelParts() != 0:
+            raise Exception(f"Wall model part '{model_part.FullName}' has submodelparts.")
+        model_part.RemoveConditionsFromAllLevels(KratosMultiphysics.TO_ERASE)
+        # 3. Create new wall law conditions from the data map info
+        for key, value in aux_cond_data_map.items():
+            model_part.CreateNewCondition(cond_reg_name, key, value[0], value[1])
```
sunethwarna: Is there a specific reason not using `ReplaceElementsAndConditionsProcess`?

applications/FluidDynamicsApplication/python_scripts/apply_wall_law_process.py
```diff
@@ -0,0 +1,152 @@
...
+
+        # Set the SLIP flag in the wall model part nodes and calculate the nodal normals
+        # This is required in order to impose the no penetration condition by the rotating schemes
+        KratosMultiphysics.NormalCalculationUtils().CalculateOnSimplex(model_part, domain_size) #FIXME: This may interact with the nodal normals of slip boundaries (e.g. floor-wall in a 3D channel)
+        KratosMultiphysics.VariableUtils().SetFlag(KratosMultiphysics.SLIP, True, model_part.Nodes)
+
+        # Set the WALL flag in the wall model part new conditions
+        # This is required in order to add the wall model contribution within the condition implementation
+        KratosMultiphysics.VariableUtils().SetFlag(KratosMultiphysics.WALL, True, model_part.Conditions)
```
sunethwarna, line 134: Nice :D

sunethwarna: What if we set the Flag of the geometry as well in here? It would be useful in some cases...

sunethwarna (dismissed):
Nice. Only minor comments. Thanks @rubenzorrilla

applications/FluidDynamicsApplication/python_scripts/apply_wall_law_process.py
rubenzorrilla: I don't think this is a good idea as the same geometry can be used to create any number of conditions, which might implement a complete different physics. Think for instance in the volumetric heat + fluid coupling.

applications/FluidDynamicsApplication/python_scripts/apply_wall_law_process.py
rubenzorrilla: Nope. Done!

rubenzorrilla:
I refactored the code a bit aiming at easing the application of this process to other CFD-related stuff. Hence, the idea is to derive a custom process from this one. This derived version will only need to
- override the `_GetHelperClassSuffix` method
- implement the corresponding wall model helper classes

sunethwarna (approved):
Nice. Thanks @rubenzorrilla

PR #10622, opened 2023-01-07
[Core] Modify Properties interface to support Non-virtual interface idiom

sunethwarna:
**📝 Description**
As stated in the title, this PR makes the relevant double methods virtual. It is required to create a way to compute sensitivities w.r.t material properties which is required to perform system identification, optimization, etc.... for material properties.

Since, these won't be used with derrived classes in general cases (only in specific cases optimization cases), AFAIK this won't be adding a significant cost to the simulations.

Please see the following comments to see the performance comparison.

**🆕 Changelog**
- Make double methods virtual

### DIFF
--- a/kratos/includes/properties.h
+++ b/kratos/includes/properties.h
@@ -199,60 +199,29 @@ public:
     ///@name Operations
     ///@{
 
-    virtual void Erase(const Variable<double>& rV)
-    {
-        mData.Erase(rV);
-    }
-
     template<class TVariableType>
     void Erase(const TVariableType& rV)
     {
         mData.Erase(rV);
     }
 
-    virtual double& GetValue(const Variable<double>& rV)
-    {
-        return mData.GetValue(rV);
-    }
-
     template<class TVariableType>
     typename TVariableType::Type& GetValue(const TVariableType& rV)
     {
-        return mData.GetValue(rV);
-    }
-
-    virtual double const& GetValue(const Variable<double>& rV) const
-    {
-        return mData.GetValue(rV);
+        return GetContainer(rV).GetValue(rV);
     }
 
     template<class TVariableType>
     typename TVariableType::Type const& GetValue(const TVariableType& rV) const
     {
-        return mData.GetValue(rV);
-    }
-
-    virtual double& GetValue(const Variable<double>& rV, NodeType& rThisNode)
-    {
-        if(mData.Has(rV))
-            return mData.GetValue(rV);
-
-        return rThisNode.GetValue(rV);
+        return GetContainer(rV).GetValue(rV);
     }
 
     template<class TVariableType>
     typename TVariableType::Type& GetValue(const TVariableType& rV, NodeType& rThisNode)
     {
-        if(mData.Has(rV))
-            return mData.GetValue(rV);
-
-        return rThisNode.GetValue(rV);
-    }
-
-    virtual double const& GetValue(const Variable<double>& rV, NodeType const& rThisNode) const
-    {
-        if(mData.Has(rV))
-            return mData.GetValue(rV);
+        if(GetContainer(rV).Has(rV))
+            return GetContainer(rV).GetValue(rV);
 
         return rThisNode.GetValue(rV);
     }
@@ -260,33 +229,17 @@ public:
     template<class TVariableType>
     typename TVariableType::Type const& GetValue(const TVariableType& rV, NodeType const& rThisNode) const
     {
-        if(mData.Has(rV))
-            return mData.GetValue(rV);
+        if(GetContainer(rV).Has(rV))
+            return GetContainer(rV).GetValue(rV);
 
         return rThisNode.GetValue(rV);
-    }
-
-    virtual double& GetValue(const Variable<double>& rV, NodeType& rThisNode, IndexType SolutionStepIndex)
-    {
-        if(mData.Has(rV))
-            return mData.GetValue(rV);
-
-        return rThisNode.GetValue(rV, SolutionStepIndex);
     }
 
     template<class TVariableType>
     typename TVariableType::Type& GetValue(const TVariableType& rV, NodeType& rThisNode, IndexType SolutionStepIndex)
     {
-        if(mData.Has(rV))
-            return mData.GetValue(rV);
-
-        return rThisNode.GetValue(rV, SolutionStepIndex);
-    }
-
-    virtual double const& GetValue(const Variable<double>& rV, NodeType const& rThisNode, IndexType SolutionStepIndex) const
-    {
-        if(mData.Has(rV))
-            return mData.GetValue(rV);
+        if(GetContainer(rV).Has(rV))
+            return GetContainer(rV).GetValue(rV);
 
         return rThisNode.GetValue(rV, SolutionStepIndex);
     }
@@ -294,24 +247,19 @@ public:
     template<class TVariableType>
     typename TVariableType::Type const& GetValue(const TVariableType& rV, NodeType const& rThisNode, IndexType SolutionStepIndex) const
     {
-        if(mData.Has(rV))
-            return mData.GetValue(rV);
+        if(GetContainer(rV).Has(rV))
+            return GetContainer(rV).GetValue(rV);
 
         return rThisNode.GetValue(rV, SolutionStepIndex);
-    }
-
-    virtual void SetValue(Variable<double> const& rV, double const rValue)
-    {
-        mData.SetValue(rV, rValue);
     }
 
     template<class TVariableType>
     void SetValue(TVariableType const& rV, typename TVariableType::Type const& rValue)
     {
-        mData.SetValue(rV, rValue);
+        GetContainer(rV).SetValue(rV, rValue);
     }
 
-    virtual bool HasVariables() const
+    bool HasVariables() const
     {
         return !mData.IsEmpty();
     }
@@ -518,15 +466,10 @@ public:
     ///@name Inquiry
     ///@{
 
-    virtual bool Has(Variable<double> const& rThisVariable) const
-    {
-        return mData.Has(rThisVariable);
-    }
-
     template<class TVariableType>
     bool Has(TVariableType const& rThisVariable) const
     {
-        return mData.Has(rThisVariable);
+        return GetContainer(rThisVariable).Has(rThisVariable);
     }
 
     template<class TXVariableType, class TYVariableType>
@@ -632,6 +575,16 @@ private:
     ///@}
     ///@name Private Operations
     ///@{
+
+    virtual inline ContainerType& GetContainer(const VariableData& rVariableData)
+    {
+        return mData;
+    }
+
+    virtual inline const ContainerType& GetContainer(const VariableData& rVariableData) const
+    {
+        return mData;
+    }
 
     ///@}
     ///@name Serialization
--- a/kratos/includes/properties.h
+++ b/kratos/includes/properties.h
@@ -199,14 +199,29 @@ public:
     ///@name Operations
     ///@{
 
+    virtual void Erase(const Variable<double>& rV)
+    {
+        mData.Erase(rV);
+    }
+
     template<class TVariableType>
     void Erase(const TVariableType& rV)
     {
         mData.Erase(rV);
     }
 
+    virtual double& GetValue(const Variable<double>& rV)
+    {
+        return mData.GetValue(rV);
+    }
+
     template<class TVariableType>
     typename TVariableType::Type& GetValue(const TVariableType& rV)
+    {
+        return mData.GetValue(rV);
+    }
+
+    virtual double const& GetValue(const Variable<double>& rV) const
     {
         return mData.GetValue(rV);
     }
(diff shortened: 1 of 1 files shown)


loumalouomega:
@sunethwarna even if not derived, just being virtual has a cost (depends on the compiler of course). But you are telling that that method must be accessed by table with a significant penalty. I will block for a proper discussion

FYI @pooyan-dadvand

loumalouomega (changes requested):
Please discussion required

loumalouomega:
> @sunethwarna even if not derived, just being virtual has a cost (depends on the compiler of course). But you are telling that that method must be accessed by table with a significant penalty. I will block for a proper discussion
> 
> FYI @pooyan-dadvand

See https://web.archive.org/web/20210225014950/http://assemblyrequired.crashworks.org/how-slow-are-virtual-functions-really/

loumalouomega:
FYI @KratosMultiphysics/altair (we already have a lot of problems with the properties acess)

sunethwarna:
I also don't like the virtual methods, but, at the moment I don't have any other solution. My requirement is as follows.

1. I need to have some properties of the materials (such as `THICKNESS`, `DENSITY`) to be different for each element
2. I don't want to change how those variables are read in the `Element` classes.
3. I want to add least cost to the existing problems.

Any suggestions are highly appreciated. :D

AlejandroCornejo:
> I also don't like the virtual methods, but, at the moment I don't have any other solution. My requirement is as follows.
> 
>     1. I need to have some properties of the materials (such as `THICKNESS`, `DENSITY`) to be different for each element
> 
>     2. I don't want to change how those variables are read in the `Element` classes.
> 
>     3. I want to add least cost to the existing problems.
> 
>     4. I don't want to duplicate `Properties` and assign individual `Properties` for each `Element` which duplicates all the other tables, values constitutive laws, etc.
> 
> 
> Any suggestions are highly appreciated. :D

what if you use the same value for the property (i.e. YOUNG) and assign a sensitivity  element-wise via `element->GetValue(SENSITIVITY)`, after, within the element you can do:

~~~cpp
if (elem.HasValue(SENSITIVITY)) {
    young *= elem.GetValue(SENSITIVITY)
}
~~~

loumalouomega:
Alejandro did a good and simple solution!

AlejandroCornejo:
Noting that this `HasValue()` is better to do it in the Initialize of the element (in order to do this check once). Also you can set this element-wise value in the mdpa file or by a custom process.

sunethwarna:
Thanks @AlejandroCornejo, Yes that's true, but that's what I want to avoid. I don't want to change existing elements. I want to be able to use existing elements throughout so it will enable me to use sensitivities with any element without special modification to element. :/

loumalouomega:
The DataValueContainer is in every element, you don't need to modify your
element one by one.

sunethwarna:
I am not sure I am getting your point @loumalouomega, @AlejandroCornejo. Let me summarize what I understood.

You are saying to store some `SENSITIVITY` variable in the data value contaienr of `Element`, and then use it within the `Element` when `YOUNGS_MODULUS` is calculated. But that means, I have to change the existing `Element` formulations which I don't want to do. This is highly error prone and hard to detect since if someone accidentally uses (highly likely because it will be hard to change all the elements and maintain them).

If we somehow find a way to mimick `GetValue` and `operater[]` to wrap original `Properties` such that, the `WrapperProperties` gives `Element`  `YOUNG_MODULUS` when required, otherwise it will get the original `Properties` values. At the moment, this is achieved by assigning different `Properties` object for each element. I want to avoid that. Having different `Properties` object for each element is redundant because `Properties` object has some material properties which are same for the all the elements in the model part, hence copying them is redundant (such as `CONSTITUTIVE_LAW`, etc...) Thats why I was thinking about making the `Properties` interface to support virtual method so, I can create `PropertiesWrapper` which gives `Element` value for `YOUNGS_MODULUS`, for ther variables it gives original `Properties` object value. Then I can have a process to assign these `PropertiesWrapper` by default for optimization problems.

The advantage of this is,
1. Less error prone, because all the exiting elements are compatible, no change at all required in `Element` formulations
2. No change in anywhere is required in the solving procedure (the whole process) except for adding one `json` process to replace existing `Properties` with `PropertiesWrapper`
3. No maintenance is required at all for elements, solution strategies, etc.

sunethwarna:
Or the second option is to use **Non-Virtual Interface** Idiom as in the following example.

```c++
class Properties
{
public:
    template<class TVar>
    void GetValue(TVar variable) { return GetDataContainer(variable).GetValue(variable);} 
private:
    virtual inline DataContainer& GetDataContainer(TVar Variable) { return mData; }
};

class PropertiesWrapper: public Properties
{
public:
   PropertiesWrapper(int id, Properties::Pointer pPointer, Element::Pointer pElement)
      :  Properties(id),
         mpElement(pElement),
         mpProperties(pProperties)
       {}

private:
   inline DataContainer& GetDataContainer(TVar variable) override {
      if (variable == YOUNG_MODULUES) { // this check can be done to any number of selected variables
         mpElement->mData;
      } else {
         mpProperties->mData;
      }
};
```

matekelemen:
> Or the second option is to use **Non-Virtual Interface** Idiom as in the following example.
> 
> ```c++
> class Properties
> {
> public:
>     template<class TVar>
>     void GetValue(TVar variable) { return GetDataContainer(variable).GetValue(variable);} 
> private:
>     virtual inline DataContainer& GetDataContainer(TVar Variable) { return mData; }
> };
> 
> class PropertiesWrapper: public Properties
> {
> public:
>    PropertiesWrapper(int id, Properties::Pointer pPointer, Element::Pointer pElement)
>       :  Properties(id),
>          mpElement(pElement),
>          mpProperties(pProperties)
>        {}
> 
> private:
>    inline DataContainer& GetDataContainer(TVar variable) override {
>       if (variable == YOUNG_MODULUES) { // this check can be done to any number of selected variables
>          return mpElement->mData;
>       } else {
>          return mpProperties->mData;
>       }
> };
> ```

?
This always calls `GetDataContainer`, which **is** a virtual function. By the way, I'm not a fan of making only the `double` overloads of `GetValue` virtual. I know that's what you need right now but it'll be super confusing for everyone else to see that some overloads are virtual while others are not.

As for performance, I'm gonna be a broken record and reiterate that a benchmark should be run with and without the change, and see just how much of an impact it has, all things considered. If only there was such a system at hand that could help us with that ...

sunethwarna:
Ok I did a small test to check the `virtual` cost. And the output is shown below. As far as I see, there is no need to be alarmed about the cost of using `virtual` since, there is no visual change in the execution time (they are almost same for all the problems). Hence use of it doesn't have a visible effect for higher number of elelemnts.

The reasons may be, the properties are used usually outside the gauss point loop, and gauss loop is relatively large. So the properties calls may be having cache missses a lot. I did the same test without calling the `CalculateLocalSystem`, and calling directly virtual `Properties::SetValue`, `Properties::GetValue` and `Properties::Has` for many number of times (1e+9). In that you can clearly see an advantage of not using virtual as @loumalouomega  pointed out. But in Kratos, we never call `Properties::GetValue`, etc million times consecutively in a row to get the cache miss advantage for non-virtual functions. Hence, I think having virtual does not have any impact for the performance.

@loumalouomega @AlejandroCornejo @pooyan-dadvand 

Following is the test program
```python
def Calculate(fname):
    n_list = [1e+2, 1e+3, 1e+4, 1e+5, 1e+6, 1e+7]

    results = []
    for n in n_list:
        n = int(n)
        start_time = timer.time()

        model = Kratos.Model()
        model_part = model.CreateModelPart("test")
        model_part.AddNodalSolutionStepVariable(Kratos.DISPLACEMENT)

        ## nodes_block - start
        nodes_start_time = timer.time()
        # create nodes
        pos = [
            [0.0, 0.0, 0.0],
            [1.0, 0.0, 0.0],
            [1.0, 1.0, 0.0]]
        for id in range(1, n+1):
            model_part.CreateNewNode(id, pos[id % 3][0], pos[id % 3][1], pos[id % 3][2])
        #nodes_block - end

        ## properties_block - start
        properties_start_time = timer.time()
        properties = model_part.CreateNewProperties(1)
        ## properties_block - end


        ## element_creation_block - start
        elements_start_time = timer.time()
        for id in range(1, n+1):
            model_part.CreateNewElement("SmallDisplacementElement2D3N", id, [((id) % n) + 1, ((id + 1 ) % n) + 1, ((id + 2) % n) + 1], properties)
        ## element_creation_block - end

        ## materials_block - start
        materials_start_time = timer.time()
        Kratos.ReadMaterialsUtility(Kratos.Parameters("""{"Parameters":{"materials_filename": "materials_json.json"}}"""), model)

        m = Kratos.Matrix()
        v = Kratos.Vector()

        ## materials_block - end

        ## initialize_block - start
        initialize_start_time = timer.time()
        for element in model_part.Elements:
            element.Initialize(model_part.ProcessInfo)
        ## initialize_block - end

        ## calculate_local_system_block - start
        calculate_local_system = timer.time()
        for element in model_part.Elements:
            element.CalculateLocalSystem(m, v, model_part.ProcessInfo)
        ## calculate_local_system_block - end

        end_time = timer.time()

        data = [
            n,
            end_time - start_time,
            end_time - calculate_local_system,
            calculate_local_system - initialize_start_time,
            initialize_start_time - materials_start_time,
            materials_start_time - elements_start_time,
            elements_start_time - properties_start_time,
            properties_start_time - nodes_start_time,
            nodes_start_time - start_time
        ]
        results.append(data)

    results = np.array(results)
    np.savetxt(fname, results, delimiter=",")
```

And the results are depicted for each start end block with the same names in following figures.
initialize_block
materials_block
Total_time
calculate_local_system_block
element_creation_block

loumalouomega:
> Ok I did a small test to check the `virtual` cost. And the output is shown below (each figure relates to a block in teh shown code). As far as I see, there is no need to be alarmed about the cost of using `virtual` since, there is no visual change in the execution time (they are almost same for all the problems). Hence use of it doesn't have a visible effect for any number of elelemnts.
> 
> The reasons may be, the properties are used usually outside the gauss point loop, and gauss loop is relatively large. So the properties calls may be having cache missses a lot. I did the same test without calling the `CalculateLocalSystem`, and calling directly virtual `Properties::SetValue`, `Properties::GetValue` and `Properties::Has` for many number of times (1e+9). In that you can clearly see an advantage of not using virtual as @loumalouomega pointed out. But in Kratos, we never call `Properties::GetValue`, etc million times consecutively in a row to get the cache miss advantage for non-virtual functions (because, we will have many `Node::FastGetSolutionStepValue` callsin between `Properties::GetValue` calls). Hence, I think having virtual does not have any impact for the performance.
> 
> @loumalouomega @AlejandroCornejo @pooyan-dadvand
> 
> Personally, I like the NVI method. I will push the NVI changes to the `properties.h`
> 
> Following is the test program
> 
> ```python
> def Calculate(fname):
>     n_list = [1e+2, 1e+3, 1e+4, 1e+5, 1e+6, 1e+7]
> 
>     results = []
>     for n in n_list:
>         n = int(n)
>         start_time = timer.time()
> 
>         model = Kratos.Model()
>         model_part = model.CreateModelPart("test")
>         model_part.AddNodalSolutionStepVariable(Kratos.DISPLACEMENT)
> 
>         ## nodes_block - start
>         nodes_start_time = timer.time()
>         # create nodes
>         pos = [
>             [0.0, 0.0, 0.0],
>             [1.0, 0.0, 0.0],
>             [1.0, 1.0, 0.0]]
>         for id in range(1, n+1):
>             model_part.CreateNewNode(id, pos[id % 3][0], pos[id % 3][1], pos[id % 3][2])
>         #nodes_block - end
> 
>         ## properties_block - start
>         properties_start_time = timer.time()
>         properties = model_part.CreateNewProperties(1)
>         ## properties_block - end
> 
> 
>         ## element_creation_block - start
>         elements_start_time = timer.time()
>         for id in range(1, n+1):
>             model_part.CreateNewElement("SmallDisplacementElement2D3N", id, [((id) % n) + 1, ((id + 1 ) % n) + 1, ((id + 2) % n) + 1], properties)
>         ## element_creation_block - end
> 
>         ## materials_block - start
>         materials_start_time = timer.time()
>         Kratos.ReadMaterialsUtility(Kratos.Parameters("""{"Parameters":{"materials_filename": "materials_json.json"}}"""), model)
> 
>         m = Kratos.Matrix()
>         v = Kratos.Vector()
> 
>         ## materials_block - end
> 
>         ## initialize_block - start
>         initialize_start_time = timer.time()
>         for element in model_part.Elements:
>             element.Initialize(model_part.ProcessInfo)
>         ## initialize_block - end
> 
>         ## calculate_local_system_block - start
>         calculate_local_system = timer.time()
>         for element in model_part.Elements:
>             element.CalculateLocalSystem(m, v, model_part.ProcessInfo)
>         ## calculate_local_system_block - end
> 
>         end_time = timer.time()
> 
>         data = [
>             n,
>             end_time - start_time,
>             end_time - calculate_local_system,
>             calculate_local_system - initialize_start_time,
>             initialize_start_time - materials_start_time,
>             materials_start_time - elements_start_time,
>             elements_start_time - properties_start_time,
>             properties_start_time - nodes_start_time,
>             nodes_start_time - start_time
>         ]
>         results.append(data)
> 
>     results = np.array(results)
>     np.savetxt(fname, results, delimiter=",")
> ```
> 
> And the results are depicted for each start end block with the same names in following figures (log-log) initialize_block materials_block Total_time calculate_local_system_block element_creation_block
> 
> And the results are depicted for each start end block with the same names in following figures (semilogx) Total_time calculate_local_system_block element_creation_block initialize_block materials_block

The problem is that "usually", some implementations, specially CL implementations, call the method internally in each GP. We do this in @KratosMultiphysics/altair , also some implementations of @AlejandroCornejo are like that. 

For your particular case I cannot think in any  solution. I will think about it. Probably metaprogramming could be a solution.

loumalouomega:
Assigning @pooyan-dadvand, this is relevant

RiccardoRossi (changes requested):
this is performance critical, so we cannot change it lighthartedly

PR #10629, opened 2023-01-09
[Core] Add process to execute operations

sunethwarna:
**📝 Description**
This PR adds a process to execute given operation at chosen execution points.

May be it is better to remove the `Prototype` version of the `Process` and `Operation` types since, this leads to if blocks as in the test of this PR. Just a thought :D 

**🆕 Changelog**
- Adds `ExecuteOperationProcess`
- Adds tests


sunethwarna:
Closing this since, it is no longer required.

PR #10632, opened 2023-01-09
[Core][ParallelUtils] adding nodiscard

philbucher:
to avoid bugs when not using the values, mainly from reductions

see also #10631


RiccardoRossi (approved):
this looks good to me!

PR #10633, opened 2023-01-09
[Core] turn FilesystemExtensions into class

philbucher:
since we prefer to use classes with static methods over namespaces I thought I could clean this up

### DIFF
--- a/kratos/tests/cpp_tests/sources/test_filesystem.cpp
+++ b/kratos/tests/cpp_tests/sources/test_filesystem.cpp
@@ -13,6 +13,7 @@
 
 // System includes
 #include <fstream>
+#include <tuple>
 
 // External includes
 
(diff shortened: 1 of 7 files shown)


kratos/tests/cpp_tests/sources/test_filesystem.cpp
```diff
@@ -13,6 +13,7 @@
 
 // System includes
 #include <fstream>
+#include <tuple>
```
matekelemen, line 16: tuple?

matekelemen (commented):
lgtm ®

kratos/tests/cpp_tests/sources/test_filesystem.cpp
philbucher: for std::ignore

rubenzorrilla (approved):
Thanks!

PR #10638, opened 2023-01-11
[Core] Model Parameters Factory - adding option to create a single item

ddiezrod:
**📝 Description**
I added an option to create a single item from model parameters factory instead of always passing a list. The code is exactly the same so in order to avoid duplications, ConstructListOfItems now calls ConstructItem for each element of the list.


loumalouomega (commented):
Okay

ddiezrod:
ping @rubenzorrilla

rubenzorrilla (approved):

PR #10639, opened 2023-01-11
[Core] Clean up legacy meshid from processes

loumalouomega:
**📝 Description**

Removing legacy  `mesh_id`. Draft until see that all tests pass.

**🆕 Changelog**

- [Clean up legacy `mesh_id` from processes](https://github.com/KratosMultiphysics/Kratos/commit/2fb4edd6cce0607bff4cb4098fc7be1788fd9076)

### DIFF
--- a/kratos/python_scripts/assign_scalar_variable_process.py
+++ b/kratos/python_scripts/assign_scalar_variable_process.py
@@ -72,7 +72,6 @@ class AssignScalarVariableProcess(KratosMultiphysics.Process):
             raise Exception(msg)
 
         self.model_part = Model[settings["model_part_name"].GetString()]
-        self.mesh = self.model_part.GetMesh()
         self.is_fixed = settings["constrained"].GetBool()
 
         self.value_is_numeric = False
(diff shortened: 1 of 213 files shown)


kratos/python_scripts/assign_scalar_variable_process.py
```diff
@@ -74,7 +72,7 @@ def __init__(self, Model, settings ):
             raise Exception(msg)
 
         self.model_part = Model[settings["model_part_name"].GetString()]
-        self.mesh = self.model_part.GetMesh(settings["mesh_id"].GetInt())
```
rubenzorrilla, line 77: I think we can even remove `self.mesh` if we pass the nodes container as `self.model_part.Nodes` below.

loumalouomega:
`mesh_id` is everywhere

philbucher:
Please add backward compatibility, otherwise this will break lots of cases

loumalouomega:
> Please add backward compatibility, otherwise this will break lots of cases

Done

loumalouomega:
@roigcarlo I think there is a bug, CI is not executing at all (maybe exceptions are not well defined for doc)

roigcarlo:
Mm which CI is not running?

loumalouomega:
> Mm which CI is not running?

Okay, last one actually worked

loumalouomega:
👋

philbucher:
As much as it is annoying, I recommend reverting some of the updates in the apps to make sure that the deprecation works

This is a necessary yet sensitive change that could break many existing cases

rubenzorrilla:
😿

PR #10643, opened 2023-01-12
[Structural] making IsElementRotated virtual

ddiezrod:
This is failing in one of our elements in altair because the constitutive law vector is not created at this point. As our element is quite specific we can just skip this step, so that is why I want to make this function virtual and re-implement it in our element.


loumalouomega (approved):
I agree, wink wink

AlejandroCornejo:
srry that I could not comment in here before. Maybe we could do this check only when `mConstitutiveLawVector.size() > 0`

loumalouomega:
> srry that I could not comment in here before. Maybe we could do this check only when `mConstitutiveLawVector.size() > 0`

Could be

PR #10645, opened 2023-01-13
[Core] Hexahedra3D8 - Adding average edge length

ddiezrod:
**📝 Description**
Adding average edge length for hexahedra3D8 with test.

### DIFF
--- a/kratos/geometries/hexahedra_3d_8.h
+++ b/kratos/geometries/hexahedra_3d_8.h
@@ -587,18 +587,26 @@ public:
     */
     double AverageEdgeLength() const override 
     {
-        return (MathUtils<double>::Norm3(this->GetPoint(0)-this->GetPoint(1)) +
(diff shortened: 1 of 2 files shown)


kratos/geometries/hexahedra_3d_8.h
```diff
@@ -580,6 +580,27 @@ template<class TPointType> class Hexahedra3D8 : public Geometry<TPointType>
...
     }
 
+    /** This method calculates and returns the average edge length of the geometry
+    *
+    * @return double value with the average edge length
+    *
+    */
+    double AverageEdgeLength() const override 
+    {
+        return (MathUtils<double>::Norm3(this->GetPoint(0)-this->GetPoint(1)) +
```
loumalouomega: Can you please get the coordinates first and then compute, I think it will be more efficient

ddiezrod: done!

loumalouomega (approved):

PR #10648, opened 2023-01-18
[GeoMechanicsApplication] Fix for removed core Register call in GeoMechanics

### LINKED ISSUE #10644: [GeoMechanicsApplication] C++ workflow for GeoFlow does not call KratosApplication::Register/RegisterVariables
In PR https://github.com/KratosMultiphysics/Kratos/pull/10421 the Register() method of the core Kratos application was removed. This has caused pure C++ workflows to not have these variables registered properly.

Currently, tests ran through the python testing scripts will work because the test code implicitly imports the Kratos core code and registers all variables.

carloslubbers:
Closes #10644


mcgicjn2 (approved):
These changes are correct if all test benches pass and it can be merged.

PR #10650, opened 2023-01-18
[Core] Return by value in Parameter iterators

### LINKED ISSUE #10649: [Core][Bug] Mutating Value in Parameter Iterators
## Description
Iterating through a `Parameters` object mutates previous values of the iterator.

## Scope
- KratosCore

## To Reproduce
```py
params = KratosMultiphysics.Parameters("[0, 1]")
lst = []

for item in params:
    lst.append(item)
    print(item.GetInt())

for item in lst:
    print(item.GetInt())
```
Output:
```
0
1
1
1
```

## Expected Behaviour
```
0
1
0
1
```

## Details
`Parameters::iterator_adaptor` and `Parameters::const_iterator_adaptor` both hold an instance to `Parameters` for the sole purpose of providing references and pointers instead of instances when dereferenced. The stored instance gets overwritten each time the iterator is dereferenced, which means that that previous values of the iterator get updated too, and thus cannot be relied on. This isn't a huge issue on the C++ side because anyone who wants to store the value behind an iterator can just copy construct a `Parameters` from the returned reference, but this is not an option on the python side. As a result, python users have only 2 options:
- store a reference to the value behind the iterator (this is unreliable as the reference gets updated every time the iterator is dereferenced)
- clone the `Parameters` object that the iterator references in each iteration. This isn't a solution either since changes to the cloned instance don't propagate back to the parent object

On the flip side, `Parameters::operator[]` and `Parameters::GetArrayItem` does what you'd expect and returns a `Parameters` instance instead of a reference, so the following runs just fine:
```py
params = KratosMultiphysics.Parameters("[0, 1]")
lst = []

for i in range(params.size()):
    lst.append(params[i])
    print(params[i].GetInt())

for item in lst:
    print(item.GetInt())
```
Output:
```
0
1
0
1
```

By the way, storing an instance that always gets updated, and returning references to it is not even thread-safe.

matekelemen:
Closes #10649 . See the issue for details.

The member `mpParameters` is removed from `Parameters::iterator_adaptor` and `Parameters::const_iterator_adaptor`. Newly constructed `Parameters` instances are returned instead, which means that `operator->` cannot be used with these iterator adaptors anymore.

### DIFF
--- a/kratos/includes/kratos_parameters.h
+++ b/kratos/includes/kratos_parameters.h
@@ -95,7 +95,7 @@ private:
          * @brief Default constructor (just iterator)
          * @param itValue The iterator to adapt
          */
-        iterator_adaptor(const iterator_adaptor& itValue);
+        iterator_adaptor(const iterator_adaptor& itValue) = default;
 
         ///@}
         ///@name Operators
@@ -135,14 +135,7 @@ private:
          * @details This operator returns the pointer of a given iterator
          * @return The Pointer of the given iterator
          */
-        Parameters& operator*() const;
-
-        /**
-         * @brief operator ->
-         * @details This operator acces to the pointer of the Parameter
-         * @return The pointer of the parameter
-         */
-        Parameters* operator->() const;
+        value_type operator*() const;
 
         ///@}
         ///@name Operations
@@ -168,7 +161,7 @@ private:
 
         std::size_t mDistance = 0;                       /// The iterator distance
         nlohmann::json& mrValue;                         /// The original container
-        std::unique_ptr<Parameters> mpParameters;        /// The unique pointer to the base Parameter
+        Kratos::shared_ptr<nlohmann::json> mpRoot;
 
         ///@}
     };
@@ -209,7 +202,7 @@ private:
          * @param itValue The iterator to adapt
          * @todo Use copy constructor in the following method
          */
-        const_iterator_adaptor(const const_iterator_adaptor& itValue);
+        const_iterator_adaptor(const const_iterator_adaptor& itValue) = default;
 
         ///@}
         ///@name Operators
@@ -249,14 +242,7 @@ private:
          * @details This operator returns the pointer of a given iterator
          * @return The Pointer of the given iterator
          */
-        const Parameters& operator*() const;
-
-        /**
-         * @brief operator ->
-         * @details This operator acces to the pointer of the Parameter
-         * @return The pointer of the parameter
-         */
-        const Parameters* operator->() const;
+        value_type operator*() const;
 
         ///@}
         ///@name Operations
@@ -281,7 +267,7 @@ private:
 
         std::size_t mDistance = 0;                       /// The iterator distance
         nlohmann::json& mrValue;                         /// The original container
-        std::unique_ptr<Parameters> mpParameters;        /// The unique pointer to the base Parameter
+        Kratos::shared_ptr<nlohmann::json> mpRoot;
 
         ///@}
     };
--- a/kratos/tests/cpp_tests/sources/test_parameters.cpp
+++ b/kratos/tests/cpp_tests/sources/test_parameters.cpp
@@ -41,12 +41,12 @@ namespace Kratos {
             KRATOS_CHECK((*i_parameter).IsBool());
             KRATOS_CHECK_EQUAL((*i_parameter).GetBool(), true);
             ++i_parameter;
-            KRATOS_CHECK(i_parameter->IsDouble());
-            KRATOS_CHECK_EQUAL(i_parameter->GetDouble(), 2.0);
+            KRATOS_CHECK((*i_parameter).IsDouble());
+            KRATOS_CHECK_EQUAL((*i_parameter).GetDouble(), 2.0);
             ++i_parameter;
             KRATOS_CHECK_EQUAL(i_parameter.name(), std::string("int_value"));
-            KRATOS_CHECK(i_parameter->IsInt());
-            KRATOS_CHECK_EQUAL(i_parameter->GetInt(), 10);
+            KRATOS_CHECK((*i_parameter).IsInt());
+            KRATOS_CHECK_EQUAL((*i_parameter).GetInt(), 10);
 
             unsigned int size = 0;
 
--- a/applications/ChimeraApplication/custom_processes/apply_chimera_process.cpp
+++ b/applications/ChimeraApplication/custom_processes/apply_chimera_process.cpp
@@ -152,10 +152,10 @@ void ApplyChimera<TDim>::DoChimeraLoop()
     BuiltinTimer do_chimera_loop_time;
 
     int i_current_level = 0;
-    for (auto& current_level : mParameters) {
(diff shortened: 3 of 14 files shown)


kratos/includes/kratos_parameters.h
```diff
@@ -168,7 +161,7 @@ class KRATOS_API(KRATOS_CORE) Parameters
 
         std::size_t mDistance = 0;                       /// The iterator distance
         nlohmann::json& mrValue;                         /// The original container
-        std::unique_ptr<Parameters> mpParameters;        /// The unique pointer to the base Parameter
+        Kratos::shared_ptr<nlohmann::json> mpRoot;
```
loumalouomega, line 164: huuuum, @RiccardoRossi

kratos/tests/cpp_tests/sources/test_parameters.cpp
```diff
@@ -41,12 +41,12 @@ namespace Kratos {
             KRATOS_CHECK((*i_parameter).IsBool());
             KRATOS_CHECK_EQUAL((*i_parameter).GetBool(), true);
             ++i_parameter;
-            KRATOS_CHECK(i_parameter->IsDouble());
-            KRATOS_CHECK_EQUAL(i_parameter->GetDouble(), 2.0);
+            KRATOS_CHECK((*i_parameter).IsDouble());
```
loumalouomega, line 44: Is it not possible to just override the -> operator?, this is a very bizarre way to use an iterator

loumalouomega (changes requested):
Blocking Just In Case

matekelemen:
> Blocking Just In Case

This PR is clearly not ready yet and I'm not sure what you're trying to achieve with blocking a PR that would need the approval of the technical committee anyway.

kratos/tests/cpp_tests/sources/test_parameters.cpp
matekelemen: `operator->` needs to return a pointer to `Parameters`, I don't know of any way around that.

The problem is that `Parameters` is not a container and doesn't really store any items, thus cannot provide stable references/pointers to "its" items. Our two options are:
1) create a much more complex `Parameters` and iterator adaptor class that is capable of providing stable references/pointers
2) drop the requirement of providing references, and remove the few cases where `operator->` is used (this is what this PR does). It's not a great solution because the standard clearly requires that iterators return references when dereferenced, so the iterator adaptors should not be used in stl algorithms in this form. However, the same can be said of the iterator adaptors currently in master too, with the added drawback that the related bug cannot be caught at compile time. At least this PR fixes that.

I won't spend my time implementing the first option because it would just get blocked and ignored anyway, but leaving this bug unaddressed is irresponsible so I went with the second one.

kratos/includes/kratos_parameters.h
matekelemen: the original stores a `unique_ptr` to `Parameters`, which in turn [stores a `shared_ptr` to `nlohmann_json`](https://github.com/KratosMultiphysics/Kratos/blob/3316a7b94c5f58d340d708e39ecc5e9b18197467/kratos/includes/kratos_parameters.h#L942) so it's **one less** heap allocations, if you're worrying about the heap.

matekelemen:
marking as ready because the CI won't launch for a draft; I'll comment when it's actually ready for review

RiccardoRossi (changes requested):
this is a big change...

kratos/tests/cpp_tests/sources/test_parameters.cpp
RiccardoRossi: i am a bit lost, what is the bug we are fixing by this?

i should say upfront that the Parameters object is very weird, but we had as a design constraint to be able to replace in glight the underlying json implementation and i could not find  a better way of doing it

matekelemen:
> this is a big change...

When you request changes, please actually state the changes you wish to be made. Otherwise I'll never find out what you're actually requesting.

RiccardoRossi:
> > this is a big change...
>       
> When you request changes, please actually state the changes you wish to be made (eg.: specific parts you're concerned about, or you don't like any of it and think the PR should just be closed). Otherwise I'll never find out what you're actually requesting.

hi @matekelemen i am "requesting changes" to block the PR until hearing the opinion of the  technical  committee,but i have no specific request to you. You can see some comments in the #10649 
to be honest I also do not like the way the Parameters are implemented, but we had no other solution with the constraints provided (of being able to swap the underlying library)

RiccardoRossi:
> > > this is a big change...
> > 
> > 
> > When you request changes, please actually state the changes you wish to be made (eg.: specific parts you're concerned about, or you don't like any of it and think the PR should just be closed). Otherwise I'll never find out what you're actually requesting.
> 
> hi @matekelemen i am "requesting changes" to block the PR until hearing the opinion of the technical committee,but i have no specific request to you. You can see some comments in the #10649 to be honest I also do not like the way the Parameters are implemented, but we had no other solution with the constraints provided (of being able to swap the underlying library)

the bigger (and open) question is: if we were to redesign the Parameters, would we be able to achieve smthg better while still hiding the underlying library to the end user?
Which also opens one for the technical committee: the nlohmann lib is very mature, what if we "marry" it (meaning we do not try to hide it any longer)?

philbucher:
there is a long standing proposal to add nlohman to the standard lib

I don't have a strong opinion in either direction 
What would be the benefits of exposing it?

matekelemen:
> there is a long standing proposal to add nlohman to the standard lib

That'd be awesome, but I doubt it'll ever happen.

> What would be the benefits of exposing it?

It's a *mostly* standard-conformant implementation, so it ties in neatly with the language and could *mostly* be used like any other standard object (const-correctness, clearly defined lifetimes, conforming iterators, exception guarantees, etc.). On the other hand, I understand the benefits of a wrapper like `Parameters` that lets us extend the features (eg: `Parameters::ValidateAndAssignDefaults`) within the object and also lets us keep our naming conventions.

I'm also torn between the two options, but I'm slightly more in favor of keeping the wrapper. However, it needs a **closed** subset of `nlohmann::json`'s features and must implement them well. Otherwise we'll keep running into similar issues that we're facing now.

kratos/tests/cpp_tests/sources/test_parameters.cpp
matekelemen: > Is it not possible to just override the -> operator?

another option is to provide a pointer-like interface to `Parameters`.
```cpp
class Parameters
{
    ...
    
    Parameters& operator*() {return *this;}

    const Parameters& operator*() const {return *this;}

    Parameters* operator->() {return this;}

    const Parameters* operator->() const {return this;}

    ...
};
```
but this is too esoteric for my taste.

loumalouomega: Still not big fan of new interface, the behaviour it is not the standard of an iterator

pooyan-dadvand:
@KratosMultiphysics/technical-committee should check the implementation to see what is the best way to do it. (In our Todo)

PR #10653, opened 2023-01-19
[HDF5] Add missing params instead of validation 

matekelemen:
## Description
Call `Parameters::AddMissingParameters` instead of `Parameters::ValidateAndAssignDefaults` on the C++ level. The reason is that objects in the HDF5 application are **always** nested, which means that the `Parameters` they get constructed by always go through multiple layers, possibly requiring a set of settings that gets narrower with each level.

Consider the following setup:
```py
class Inner
  def __init__(self, params):
    # Expects {
    #    "io_settings" : {}
    # }
    pass

class Outer
  def __init__(self, params):
    # Expects {
    #    "model_part_name" : "",
    #    "io_settings" : {}
    # }
    ...
    self.inner = Inner(...)
```

If `Parameters::ValidateAndAssignDefaults` is called on the passed parameters, the `params` that `Outer` gets constructed with needs to be cleaned of extra entries first. This PR encourages objects to get initialized using only a subset of the settings they get passed, instead of rejecting extra settings they don't recognize.

The benefit is that modifications to specific settings propagate back to the original user-defined `Parameters` object, and there's less need to nest `Parameters` into each other. The drawback is that typos go unnoticed, and default settings get applied instead.

### DIFF
--- a/applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
+++ b/applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
@@ -374,7 +374,7 @@ ContainerComponentIO<TContainerType, TContainerItemType, TComponents...>::Contai
             "list_of_variables": []
         })");
 
-    Settings.ValidateAndAssignDefaults(default_params);
+    Settings.AddMissingParameters(default_params);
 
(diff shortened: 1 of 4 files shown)


sunethwarna (approved):
Thanks @matekelemen

PR #10655, opened 2023-01-19
Fixed a CMake syntax error on Windows during installation

### LINKED ISSUE #10652: CMake warning (syntax error) when installing Kratos on Windows
**Description**
When I build and install Kratos through the command prompt on Windows using the `standard_configure.bat` file from the `scripts` directory (where I have changed the generator to Visual Studio 17 2022), I get the following output:
```
CMake Warning (dev) at build/Release/scripts/post_install/cmake_install.cmake:40 (execute_process):
  Syntax error in cmake code at

    C:/Users/graaf/dev/KratosProjects/pristine/build/Release/scripts/post_install/cmake_install.cmake:40

  when parsing string

    C:\Windows\py.exe

  Invalid escape sequence \W

  Policy CMP0010 is not set: Bad variable reference syntax is an error.  Run
  "cmake --help-policy CMP0010" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
Call Stack (most recent call first):
  build/Release/cmake_install.cmake:134 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

My understanding is that this string directly comes from the environment variable `PYTHON_EXECUTABLE`, which defaults to `C:\Windows\py.exe` if not set (see `scripts\standard_configure.bat`). The problem here is that the directory separator on Windows (`\`) is interpreted as an escape character when CMake processes the (path) string.

When I manually change the generated path string to `C:/Windows/py.exe` (in `build/Release/scripts/post_install/cmake_install.cmake`) the problem no longer occurs.

**Scope**
Installation (on Windows)

**To Reproduce**
Open a command prompt on Windows with the compiler etc. set up as described in `INSTALL.md`. Then navigate to the `scripts` directory and run the (modified) script `standard_configure.bat` (which also builds and installs Kratos). During the installation stage the above error is reported.

**Expected behavior**
I expect the (path) string to be such that it won't result in a syntax error.

**Environment**
- OS: Windows 10 Version 20H2
- Branch: master
- Python 3.11
- CMake 3.25
- Boost 1.81

avdg81:
**📝 Description**
The Windows path to the Python executable was used 'as-is' when defining a CMake string for the command to execute. As a result, a syntax error was reported by CMake (the directory separator `\` was interpreted as an escape character). It has been fixed by converting the path to the Python executable to a CMake path first.

**🆕 Changelog**
Fixes #10652.

### DIFF
--- a/scripts/post_install/CMakeLists.txt
+++ b/scripts/post_install/CMakeLists.txt
@@ -2,5 +2,6 @@ set(KRATOS_GENERATE_PYTHON_STUBS OFF CACHE BOOL "Generate hints for binary pytho
 
 if (${KRATOS_GENERATE_PYTHON_STUBS})
     install(CODE "message(STATUS \"Trying to install python stub files to enable python hinting support for cpp libraries in IDEs...\")")
-    install(CODE "execute_process(COMMAND \"${PYTHON_EXECUTABLE}\" \"${CMAKE_CURRENT_SOURCE_DIR}/stub_generation.py\" \"${CMAKE_INSTALL_PREFIX}\" --quiet)")
+    cmake_path(CONVERT ${PYTHON_EXECUTABLE} TO_CMAKE_PATH_LIST CONVERTED_PY_EXEC)
+    install(CODE "execute_process(COMMAND \"${CONVERTED_PY_EXEC}\" \"${CMAKE_CURRENT_SOURCE_DIR}/stub_generation.py\" \"${CMAKE_INSTALL_PREFIX}\" --quiet)")
 endif()


loumalouomega (approved):
Okay from my side

ddiezrod:
I am getting an error in this file after this change:
```

CMake Error at scripts/post_install/CMakeLists.txt:5 (cmake_path):
  Unknown CMake command "cmake_path".
```

I am using cmake 3.18 an it seems this is only included after cmake 3.20 (see https://cmake.org/cmake/help/latest/command/cmake_path.html) Should we updat the cmake_minimum_version? Maybe its a too recent version... @roigcarlo

scripts/post_install/CMakeLists.txt
```diff
@@ -2,5 +2,6 @@ set(KRATOS_GENERATE_PYTHON_STUBS OFF CACHE BOOL "Generate hints for binary pytho
 
 if (${KRATOS_GENERATE_PYTHON_STUBS})
     install(CODE "message(STATUS \"Trying to install python stub files to enable python hinting support for cpp libraries in IDEs...\")")
-    install(CODE "execute_process(COMMAND \"${PYTHON_EXECUTABLE}\" \"${CMAKE_CURRENT_SOURCE_DIR}/stub_generation.py\" \"${CMAKE_INSTALL_PREFIX}\" --quiet)")
+    cmake_path(CONVERT ${PYTHON_EXECUTABLE} TO_CMAKE_PATH_LIST CONVERTED_PY_EXEC)
+    install(CODE "execute_process(COMMAND \"${CONVERTED_PY_EXEC}\" \"${CMAKE_CURRENT_SOURCE_DIR}/stub_generation.py\" \"${CMAKE_INSTALL_PREFIX}\" --quiet)")
```
roigcarlo, line 6:
```CMake
\"${CMAKE_CURRENT_SOURCE_DIR}/stub_generation.py\"
```

Most likely the problem is the `/` in the 2nd argument of the install (The one I quoted). I assume the same solution that @avdg81 proposed will work for the other arguments. 

In the meantime @ddiezrod you can just set `KRATOS_GENERATE_PYTHON_STUBS=OFF` I think. It is not a critical feature....

roigcarlo:
Yep seems that it needs 3.20. 

We are reluctant to update the min version because some systems give problems. I would be more comfortable leaving it to 3.14 (note that now is 2.8.6, which means that essentially we are not using that feature), but since the change has already been merged... go ahead.

avdg81:
First of all, my apologies that I had not carefully checked the CMake
version in which `cmake_path` was introduced (3.20, so it appeared). I
agree that we should not require all users to upgrade their CMake
installations just because of this small enhancement. Unfortunately, its
predecessor (`file(TO_NATIVE_PATH ...)`) also seems to be relatively new
(requiring CMake 3.19 if I understand it correctly):
https://cmake.org/cmake/help/latest/command/file.html#to-cmake-path.

Another option I can think of is to write a small custom CMake function
that replaces Windows-style directory separators by CMake-style path
separators and use that (which is in essence identical to what my original
solution does, except that it should also work using older CMake versions).
Please let me know if you'd like me to implement such a change. Thanks.

Op ma 23 jan. 2023 om 13:09 schreef Carlos Roig ***@***.***>:

roigcarlo:
Don't worry, let's leave it like this (doc is already updated) if people complains we can revert back.

PR #10657, opened 2023-01-20
[Meshing] Convert linear tets to quadratic  - use root model part

ddiezrod:
**📝 Description**
I found a bug while using this utility to convert linear tests to quadratic ones. The replacement is not being done in the root model part, and therefore this root model part still has the original linear elements.

This is even worse if you call later the replace element and conditions process as this process aslways uses the elements from the root model part, so you will end up with linear elements again.


loumalouomega (approved):
Nice catch!

PR #10659, opened 2023-01-21
[HDF5] Pipes

matekelemen:
# Overview

Add a modular system that simplifies the definition, construction, configuration, and validation of model predicates from #10658.

*I don't recommend jumping into the code right away without reading the description. There's a lot of template machinery behind the scenes, so if you want to get a practical overview of what's happening here, I recommend reading the description and then following up with [piped_model_predicate.h](https://github.com/KratosMultiphysics/Kratos/blob/b875581ecf13e5bf70318d83946a4d4971ac1411/applications/HDF5Application/custom_utilities/piped_model_predicate.h) and [basic_pipes.h](https://github.com/KratosMultiphysics/Kratos/blob/b875581ecf13e5bf70318d83946a4d4971ac1411/applications/HDF5Application/custom_utilities/basic_pipes.h) if you like. Implementation headers (`*_impl.h`) contain template implementation details you may find difficult to understand if you haven't seen the higher level constructs of this PR.*

Requires #10658 and the updated interval utility #10301.

# Description

The essence of this PR is to define a **composable** system of building blocks (pipe segments) that can be mixed and matched together at compile time to form blocks capable of increasingly more complex tasks.

For example, imagine checking whether a ramp-up is active: it involves finding the related `ModelPart`, getting its `ProcessInfo`, querying its `TIME` and then feeding that into an `IntervalUtility`. As a sequence of steps, this process looks like so:
```
Model => ModelPart => ProcessInfo => Time => IntervalUtility::IsInInterval
``` 
Each `=>` can be considered an operator that takes its left hand side as input, and produces its right hand side as output that then gets fed into the next operator. In this PR, these operators are called pipes or segments, and chained pipes are called pipelines. Note however, that these terms can be interchangeable because individual segments are just as composable as entire pipelines. Any pipeline can be considered a segment of a longer one, and this relationship is reflected in the implementation.

It'll be easier if I explain the features from top to bottom:

### `PipedModelPredicate`

A `ModelPredicate` from #10658 (scans a `Model` to produce a `bool`) that uses a pipe to do the work. The pipe it stores as a member variable is typically a `CompoundPipe` that consists of multiple segments.

### `CompoundPipe`

A `CompoundPipe` has an input- and an output pipe segment, both of which **can** be other `CompoundPipe`s. One or more levels of nested `CompoundPipe`s form a pipeline that can either be default-constructed, or configured via an array of `Parameters`. Each segment takes the output argument of its predecessor as input, computes an output and forwards it to the next pipe segment. **Pipe segments should be stateless and must not mutate their inputs**, though there's no way of checking or enforcing that.

### Pipe

Despite mentioning pipes all the time, there is no `Pipe` class. Rather, it's a set of requirements a class must satisfy in order to be considered a pipe. Here's the related part of the docs:
```cpp
/**
 *  @details A pipe is a modular sequence of operations that takes an immutable input and produces an output.
 *           Modularity is provided via static polymorphism (templates) to avoid the overhead incurred by virtual
 *           functions. Therefore, there is no base class for a @a pipe; rather, a class must satisfy a set of
 *           requirements to be considered a @a Pipe. These requirements are:
 *              - has member type @a InputType
 *              - has member type @a OutputType
 *              - has member function operator() with the following signature:
 *                @code OutputType operator()(@a InputType) const @endcode
 *              - must not have side-effects (not checked)
 *              - must not mutate its input (not checked)
 *              - must be default, move, and copy constructible (not checked)
 *              - [optional]: should be constructible from @ref Parameters (required for @ref CompoundPipe).
 */
```

# Usage

Let's say we wanted to implement a new model predicate that checks whether a `ModelPart` is MPI distributed. Here's an overview of the required segments:
```
Model => ModelPart => ModelPart::IsDistributed
```
The `Model => ModelPart` pipe is already implemented in [ModelPartFromModel](https://github.com/KratosMultiphysics/Kratos/blob/b875581ecf13e5bf70318d83946a4d4971ac1411/applications/HDF5Application/custom_utilities/basic_pipes.h#L37) so we only need `ModelPart => ModelPart::IsDistributed`, which is pretty starightforward:

- Inherit from `Pipes::Traits` to get the type aliases necessary for pipes. The input type is `const ModelPart&` and the output type is `bool`, so more specifically we need to inherit from `Pipes::Traits<const ModelPart&, bool>`
- Implement `operator() const` that returns `IsDistributed` on the input model part: `bool operator()(const ModelPart& rModelPart) const {return rModelPart.IsDistributed()}`
- Define a default constructor.
- Define a constructor from `Parameters` even though we don't need anything configured in this case (necessary for composability).
- Define a move constructor.
- Define a copy constructor.

```cpp
namespace Pipes {
class IsModelPartDistributed : public Traits<const ModelPart&, bool>
{
public:
    IsModelPartDistributed() = default;

    IsModelPartDistributed(const Parameters& rParameters) {}

    IsModelPartDistributed(IsModelPartDistributed&& rOther) noexcept = default;

    IsModelPartDistributed(const IsModelPartDistributed& rOther) = default;

    bool operator()(const ModelPart& rModelPart) const
    {return rModelPart.IsDistributed()}
};
} // namespace Pipes
```

Now we have all the segments necessary to define our pipeline predicate:
 ```cpp
using IsDistributedPredicate = PipedModelPredicate<Pipes::Pipeline<
    Pipes::ModelPartFromModel,
    Pipes::IsModelPartDistributed
>>;
```

We're all set to use the new `IsDistributedPredicate` in C++, but we still have to expose it with pybind if we wish to use it in python:
```cpp
pybind11::class_<IsDistributedPredicate,
                 IsDistributedPredicate::Pointer,
                 ModelPredicate>(rModule, IsDistributedPredicate)
    .def(pybind11::init<>())
    .def(pybind11::init<const Parameters&>())
    .def("__call__", IsDistributedPredicate::operator());
```

And we're done. Now the predicate can be used in python as well:
```py
import KratosMultiphysics
model = KratosMultiphysics.Model()
model.CreateModelPart("example_model_part")
predicate = KratosMultiphysics.IsDistributedPredicate(
    KratosMultiphysics.Parameters("""[
        {"model_part_name" : "example_model_part"}, // <== model part from model
        {}                                          // <== is model part distributed
    ]""")
)
predicate(model) # <== False
```


RiccardoRossi (commented):
hi @matekelemen i looked into this diagonally and i think that the mechanism that you are using is very interesting (i learnt from reading this!).

However i have a bit of concern with the proliferation of templates. Wouldn't this take forever to compile??
This comment takes me the following followup: I am under the impression (but admittedly i can be wrong) that the kind of operations you need is not "time critical" in the sense that you do not really need maximum performance in such operations. If that is so, wouldn't it make sense to implement the same mechanism in python,  taking advantage of the relaxed type checking of python?

note that this is NOT a review, they are just open questions

Issue #9422, opened 2021-12-03
Edge/face ordering of high order geometries seem inconsistent with corresponding 1D ordering


Vahid-Galavi:
**Introduction**
If we consider a high order 2D geometry, for example quadrilateral2D8 or quadirlateral2D9, the nodal ordering is as follows:

~~~html
Quadrilateral2D8/       Quadrilateral2D9/
Quadrilateral3D8:       Quadrilateral3D9:

3-----6-----2           3-----6-----2 
|           |           |           | 
|           |           |           | 
7           5           7     8     5 
|           |           |           | 
|           |           |           | 
0-----4-----1           0-----4-----1 
~~~

If we consider a high order 1D geometry, i.e. Line2D3 or Line3D3, the ordering is as follows:

Line2D3/   
Line3D3:   
                                                
`0-----2----1`

So, (in all geometries) we get first the corner nodes and then the middle nodes which are consistent. 

**Problem**
Now if we look at `GenerateEdges()` functions which return edge or face node numbers, I found that the node numbers that I get in these elements: `Triangle2D6/Triangle3D6`, `Quadrilateral2D8/Quadrilateral3D8` and `Quadrilateral2D9/Quadrilateral3D9` are inconsistent with node numbering of `Line2D3/Line3D3`. These functions return the first, the middle and the last nodes instead of the first, the last and the middle. This implies that in order to check if a line belongs to a face, we need to do a redundant sorting. 

This also the case in `GenerateEdges` of `Hexahedron3D20` `Hexahedron3D27`, `Prism3D15`.

**Questions**
- Why do we use a different node numbering for edges?
- Can we modify the node numbering of edges to make them consistent with 1D geometries?

philbucher:
This is an important observation, I agree that this is a problem!
However I don't think it is possible to change the ordering, as this is sort of imposed by the meshers we use
Would changing the `GenerateEdges` function work? I think this might be possible as it is used less

Vahid-Galavi:
In my opinion, we only need to change `GenerateEdges`.

philbucher:
that makes sense to me
did you also check if the same problem exists with `GenerateFaces`?

Vahid-Galavi:
> that makes sense to me did you also check if the same problem exists with `GenerateFaces`?

@philbucher I am checking now.

Vahid-Galavi:
@philbucher  What is the ordering number for surface loads in 3D? Is there a documentation?

[...]

PR #10662, opened 2023-01-23
[Core] Adding average edge length to tet10

ddiezrod:
**📝 Description**
Adding average edge length to tet10.


kratos/geometries/tetrahedra_3d_10.h
```diff
@@ -536,28 +536,28 @@ template<class TPointType> class Tetrahedra3D10 : public Geometry<TPointType>
         typedef typename Geometry<TPointType>::Pointer EdgePointerType;
         edges.push_back( EdgePointerType( new EdgeType(
                                               this->pGetPoint( 0 ),
-                                              this->pGetPoint( 4 ),
-                                              this->pGetPoint( 1 ) ) ) );
+                                              this->pGetPoint( 1 ),
+                                              this->pGetPoint( 4 ) ) ) );
```
loumalouomega: Super bug, nice catch!

kratos/geometries/tetrahedra_3d_10.h
```diff
@@ -621,6 +621,20 @@ template<class TPointType> class Tetrahedra3D10 : public Geometry<TPointType>
...
+    /** This method calculates and returns the average edge length of the geometry
+     *
+     * @return double value with the average edge length
+     *
+     */
+    double AverageEdgeLength() const override {
+        double average_edge_length = 0.0;
+        GeometriesArrayType edges = this->GenerateEdges();
+        for (auto& edge : edges) {
+            average_edge_length += edge.Length();
```
loumalouomega:
```suggestion
        for (auto& r_edge : edges) {
            average_edge_length += r_edge .Length();
```

kratos/geometries/tetrahedra_3d_10.h
```diff
@@ -621,6 +621,20 @@ template<class TPointType> class Tetrahedra3D10 : public Geometry<TPointType>
...
+        }
+        return average_edge_length/6.0;
```
loumalouomega: Technically it will by faster to multiply by 0.16666666666666666666666

matekelemen:
```suggestion
        const GeometriesArrayType edges = this->GenerateEdges();
        return std::accumulate(
            edges.begin(),
            edges.end(),
            0.0,
            [](double sum, const auto& rEdge) {return sum + rEdge.Length();}
        ) / 6.0;
```

loumalouomega (dismissed):

kratos/geometries/tetrahedra_3d_10.h
matekelemen: the compiler converts divisions with literals to multiplications if it's safe to do so (it's not in this case, because 1/6 is not exactly representable with floating point nums)

loumalouomega (dismissed):

kratos/geometries/tetrahedra_3d_10.h
ddiezrod: done!

ddiezrod: done, i dont see clearly the advantage though, is it just it looks better?

matekelemen: In this case yes, it probably makes no practical difference because they're so few items in `edges`, but it's more straightforward to read what's happening.

In the general case though, it pays off to use standard algorithms. They're implemented with exception guarantees and in a manner that makes compiler optimizations possible. Also, they usually switch between algorithms depending on the input size to provide better performance / stable numerics.

matekelemen: I guess you need to `#include <numeric>`

btw I haven't compiled and tested the suggestion, I meant it as a comment.

ddiezrod: yep, I will add it now

loumalouomega (approved):

ddiezrod:
I see that same mistake is in a lot of other geometries.

https://github.com/KratosMultiphysics/Kratos/blob/5dfe7140ac935f20953a0c26bf7b97debee3b09c/kratos/geometries/triangle_3d_6.h#L654

https://github.com/KratosMultiphysics/Kratos/blob/5dfe7140ac935f20953a0c26bf7b97debee3b09c/kratos/geometries/triangle_2d_6.h#L677

just an example, I will create another PR to fix just that.

ping @KratosMultiphysics/geomechanics it seems this is breaking some of your tests. I think error comes from this trick https://github.com/KratosMultiphysics/Kratos/blob/5dfe7140ac935f20953a0c26bf7b97debee3b09c/applications/GeoMechanicsApplication/custom_processes/find_neighbour_elements_of_conditions_process.cpp#L50

loumalouomega:
> I see that same mistake is in a lot of other geometries.
> 
> https://github.com/KratosMultiphysics/Kratos/blob/5dfe7140ac935f20953a0c26bf7b97debee3b09c/kratos/geometries/triangle_3d_6.h#L654
> 
> https://github.com/KratosMultiphysics/Kratos/blob/5dfe7140ac935f20953a0c26bf7b97debee3b09c/kratos/geometries/triangle_2d_6.h#L677
> 
> just an example, I will create another PR to fix just that.
> 
> ping @KratosMultiphysics/geomechanics it seems this is breaking some of your tests. I think error comes from this trick
> 
> https://github.com/KratosMultiphysics/Kratos/blob/5dfe7140ac935f20953a0c26bf7b97debee3b09c/applications/GeoMechanicsApplication/custom_processes/find_neighbour_elements_of_conditions_process.cpp#L50

Huuum, copy and paste evilness

Vahid-Galavi:
Yes, this was known and reported here #9422

mcgicjn2:
@loumalouomega Am I ok to fix our tests in this issue ? 
@Vahid-Galavi I assume this will also close your issue #9422 ?

ddiezrod:
I divided this PR into two parts, the once concerning only the fixes in geometries is in https://github.com/KratosMultiphysics/Kratos/pull/10668

loumalouomega:
> @loumalouomega Am I ok to fix our GeoMechanicsApplication tests in this issue ? @Vahid-Galavi I assume this will also close your issue #9422 ?

I think if not in this PR, in an independent PR which merges with this branch (instead of the master), so the review in this branch will be done after that second PR is merged

Vahid-Galavi:
@mcgicjn2 I haven't checked this PR but I think this will close that issue.

ddiezrod:
This is ready for review after merging https://github.com/KratosMultiphysics/Kratos/pull/10668. It only computes average edge length now. I changed description accordingly.

loumalouomega (approved):

PR #10668, opened 2023-01-23
[Core] Fix generate edges in geometries

ddiezrod:
I found that in most geometries edges are not created with the correct numbering.

For example in the line with 3 points, the order should be: 0 ------ 2 ------ 1, but in most cases  they are being created with 0----1----2, which leads to several problems (for example, length is not correctly computed)

### DIFF
--- a/kratos/geometries/triangle_2d_10.h
+++ b/kratos/geometries/triangle_2d_10.h
@@ -693,9 +693,9 @@ public:
     GeometriesArrayType GenerateEdges() const override
     {
         GeometriesArrayType edges = GeometriesArrayType();
-        edges.push_back(Kratos::make_shared<EdgeType>(this->pGetPoint(0), this->pGetPoint(3), this->pGetPoint(4), this->pGetPoint(1)));
-        edges.push_back(Kratos::make_shared<EdgeType>(this->pGetPoint(1), this->pGetPoint(5), this->pGetPoint(6), this->pGetPoint(2)));
-        edges.push_back(Kratos::make_shared<EdgeType>(this->pGetPoint(2), this->pGetPoint(7), this->pGetPoint(8), this->pGetPoint(0)));
+        edges.push_back(Kratos::make_shared<EdgeType>(this->pGetPoint(0), this->pGetPoint(1), this->pGetPoint(3), this->pGetPoint(4)));
+        edges.push_back(Kratos::make_shared<EdgeType>(this->pGetPoint(1), this->pGetPoint(2), this->pGetPoint(5), this->pGetPoint(6)));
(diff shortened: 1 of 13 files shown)


kratos/geometries/triangle_2d_10.h
```diff
@@ -693,9 +693,9 @@ class Triangle2D10 : public Geometry<TPointType>
     GeometriesArrayType GenerateEdges() const override
     {
         GeometriesArrayType edges = GeometriesArrayType();
-        edges.push_back(Kratos::make_shared<EdgeType>(this->pGetPoint(0), this->pGetPoint(3), this->pGetPoint(4), this->pGetPoint(1)));
```
loumalouomega, line 696: These implementations are new @mnabideltares, is it correct?

loumalouomega (dismissed):
Wait for @mnabideltares, a priori the new elements should be right

kratos/geometries/triangle_2d_10.h
ddiezrod: They had the same mistake as the other ones I think.

ddiezrod:
I did not know there was already an issue for this. 

https://github.com/KratosMultiphysics/Kratos/issues/9422

I understand that what I am doing here goes in the same line of what was discussed there. @Vahid-Galavi  thanks for pointing me to the issue.

Vahid-Galavi:
You are welcome @ddiezrod

kratos/geometries/triangle_2d_10.h
mnabideltares: I followed the same reciepe of other high order elements. So, it needs to be investigted too. However, during implementation of these elements, my test cases all worked as expected with this definition of points

mnabideltares: However, my test cases all worked correctly with this definition of points

mnabideltares: @loumalouomega  and @ddiezrod 
Do I modify the the nodes order for 10N and 15N elements?

ddiezrod: I think the order is correct now with the changes in this branch, isn't it?

mnabideltares: Yes, it seems to be correct. Thank you

ddiezrod:
Tests are passing now with the change in @KratosMultiphysics/geomechanics , is this ok for everyone?

loumalouomega (approved):

loumalouomega:
From my side yes

mcgicjn2:
Yes these changes are fine from our side in GeoMechanicsApplication - Thanks for doing it

PR #10674, opened 2023-01-24
[Core] Tets10 - adding HasIntersection for planar geoms

ddiezrod:
**📝 Description**
This PR implements the method HasIntersections for tets10 if they have planar faces. When we are in this case, we can simply use the implementation from tet4.

The reason for this PR is that we need to have this function to be able to use the bin locator in our cases in Altair with quadratical tets. In our problems, the quadratic mesh is always generated from linear tets so that is why we only need to cover this case.

Of course, the ideal case would be to implement this function for any tet10, but there is no analytical solution for quadratic geometries and the implementation will be quite difficult. I hope we can this merged so we can use it for our problems.

### DIFF
--- a/kratos/geometries/tetrahedra_3d_10.h
+++ b/kratos/geometries/tetrahedra_3d_10.h
@@ -1065,10 +1065,10 @@ private:
     bool FacesArePlanar() const
     {
         constexpr double tol = 1e-6;
-        for (auto edge : this->GenerateEdges()) {
-            const double a = MathUtils<double>::Norm3(edge.GetPoint(0)-edge.GetPoint(1));
-            const double b = MathUtils<double>::Norm3(edge.GetPoint(1)-edge.GetPoint(2));
-            const double c = MathUtils<double>::Norm3(edge.GetPoint(2)-edge.GetPoint(0));
+        for (auto& r_edge : this->GenerateEdges()) {
+            const double a = MathUtils<double>::Norm3(r_edge.GetPoint(0)-r_edge.GetPoint(1));
+            const double b = MathUtils<double>::Norm3(r_edge.GetPoint(1)-r_edge.GetPoint(2));
+            const double c = MathUtils<double>::Norm3(r_edge.GetPoint(2)-r_edge.GetPoint(0));
             if (b + c > a + c*tol) {
                 return false;
             }
--- a/kratos/geometries/tetrahedra_3d_10.h
+++ b/kratos/geometries/tetrahedra_3d_10.h
@@ -1069,7 +1069,7 @@ private:
             const double a = MathUtils<double>::Norm3(edge.GetPoint(0)-edge.GetPoint(1));
             const double b = MathUtils<double>::Norm3(edge.GetPoint(1)-edge.GetPoint(2));
             const double c = MathUtils<double>::Norm3(edge.GetPoint(2)-edge.GetPoint(0));
-            if (b + c > a + c*tol) {
+            if (b + c > a*(1.0+tol) ) {
                 return false;
             }
         }
--- a/kratos/geometries/tetrahedra_3d_10.h
+++ b/kratos/geometries/tetrahedra_3d_10.h
@@ -23,6 +23,7 @@
 
 // Project includes
 #include "geometries/triangle_3d_6.h"
+#include "geometries/tetrahedra_3d_4.h"
 #include "utilities/integration_utilities.h"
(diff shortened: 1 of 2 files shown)


kratos/geometries/tetrahedra_3d_10.h
```diff
@@ -1031,6 +1055,27 @@ template<class TPointType> class Tetrahedra3D10 : public Geometry<TPointType>
...
+     *
+     */
+    bool FacesArePlanar() const
+    {
+        constexpr double tol = 1e-6;
+        for (auto edge : this->GenerateEdges()) {
+            const double a = MathUtils<double>::Norm3(edge.GetPoint(0)-edge.GetPoint(1));
+            const double b = MathUtils<double>::Norm3(edge.GetPoint(1)-edge.GetPoint(2));
+            const double c = MathUtils<double>::Norm3(edge.GetPoint(2)-edge.GetPoint(0));
+            if (b + c > a + c*tol) {
```
pablobecker:
```suggestion
            if (b + c > a*(1.0+tol) ) {
```

pablobecker: the tolerance should be based on the full lenght rather than c, which could be close to 0

pablobecker (changes requested):

kratos/geometries/tetrahedra_3d_10.h
ddiezrod: makes sense, done!

loumalouomega:
Just to comment that at the moment the mesh deforms is not planar anymore...

kratos/geometries/tetrahedra_3d_10.h
```diff
@@ -1031,6 +1055,27 @@ template<class TPointType> class Tetrahedra3D10 : public Geometry<TPointType>
...
+     * Checks if faces are planar. We iterate for all edges and check
+     * that the sum of 0-2 and 2-1 segments is no bigger than 0-1.
+     *
+     * @return bool faces are planar or not
+     *
+     */
+    bool FacesArePlanar() const
+    {
+        constexpr double tol = 1e-6;
+        for (auto edge : this->GenerateEdges()) {
```
loumalouomega:
```suggestion
        for (auto& edge : this->GenerateEdges()) {
```
I think otherwise you are doing a copy of a temporary object

loumalouomega (commented):
One comment (well, two, I just had an idea)

kratos/geometries/tetrahedra_3d_10.h
```diff
@@ -769,6 +770,29 @@ template<class TPointType> class Tetrahedra3D10 : public Geometry<TPointType>
...
+    bool HasIntersection(const Point& rLowPoint, const Point& rHighPoint) const override
+    {
+        if (this->FacesArePlanar()) {
+            return Tetrahedra3D4<TPointType>(
+                this->pGetPoint(0),
+                this->pGetPoint(1),
+                this->pGetPoint(2),
+                this->pGetPoint(3)).HasIntersection(rLowPoint, rHighPoint);
+        } else {
+             KRATOS_ERROR << "\"HasIntersection\" is not implemented for non-planar 10 noded tetrahedra.";
```
loumalouomega, line 791: Maybe instead of an error, we can decompose in 4 tet4 and throw a warning telling the approximation. This will be an intermediate solution, but I am OK with current implementation

qiukailu: @pooyan-dadvand suggested a similar approximation by subdivision of the triangular faces into 4 smaller ones. I think the "debate" really is whether we should provide a correct solution for a subset of all tet10s, or an approximation solution that covers all tet10s.

kratos/geometries/tetrahedra_3d_10.h
ddiezrod: you are right

kratos/geometries/tetrahedra_3d_10.h
ddiezrod: I honestly prefer to stick to the current solution, rather than giving an approximation that can be very wrong in many cases.

ddiezrod:
> Just to comment that at the moment the mesh deforms is not planar anymore...

Yes, then if you call this function again, it will throw an error.

ddiezrod:
So... can this be approved if its ok for everyone?

qiukailu (approved):

pablobecker (approved):

PR #10679, opened 2023-01-25
[Meshing] Convert linear tets to quad -Using root model part to do replacement in conditions

ddiezrod:
**📝 Description**
This is the same mistake that we found in https://github.com/KratosMultiphysics/Kratos/pull/10657 but for conditions.


pablobecker:
hopefully this is the last one !

pablobecker (approved):

PR #10681, opened 2023-01-26
Adding symmetrize option perturbation

AlejandroCornejo:
**📝 Description**
In this PR I am adding the capability of symmetrising the tangent constitutive operator obtained via perturbation method. This can be done by adding `"SYMMETRIZE_TANGENT_OPERATOR" : true` in the `Materials.json`

This allows the use of linear solvers for symmetric matrices, with its subsequent reduction of cost and memory.

@RiccardoRossi @SergioJimenezReyes


loumalouomega (approved):
Sounds good

PR #10687, opened 2023-01-26
[Core] Adding CheckSameModelPartUsingSkinDistanceProcess

loumalouomega:
**📝 Description**

This PR adds the `CheckSameModelPartUsingSkinDistanceProcess`, which as the name says, checks that two model parts are the same checking that the (discontinous) distance error norm is below a certain threshold. The idea is from @pooyan-dadvand. It uses `CalculateDiscontinuousDistanceToSkinProcess`. It assumes that the provided modelparts are a skin (conditions), then in computed the BoundingBox of the first skin model part and generates a background mesh to compute the distance with `StructuredMeshGeneratorProcess`. For the moment no compatible with MPI. 

Use:

 ~~~c++
Parameters parameters = Parameters(R"(
    {
        "skin_model_part_1_name"              : "PLEASE_SPECIFY_SKIN_MODEL_PART_2_NAME",
        "skin_model_part_2_name"              : "PLEASE_SPECIFY_SKIN_MODEL_PART_2_NAME",
        "tolerance"                           : 1.0e-3,
        "number_of_divisions_background_mesh" : 30,
        "bounding_box_scale_factor"           : 1.5,
        "discontinuous_distance_settings": {
        }
    })" );

CheckSameModelPartUsingSkinDistanceProcess<3>(model, parameters).Execute();
~~~

In case it is not the same throws an error, otherwise a mere info of the distance error.

**🆕 Changelog**

- [Initial commit](https://github.com/KratosMultiphysics/Kratos/commit/7ee0b95b33cea01531ec71173b3eb2560fb5bdaf)
- [Refactor with background mesh](https://github.com/KratosMultiphysics/Kratos/commit/d98cbda85106f5ae7449b692dc7992696be1fc3e)
- [Initial add test](https://github.com/KratosMultiphysics/Kratos/commit/1c17210500d291a5ec3b511450c20c947f09046c)
- [Adding failing case](https://github.com/KratosMultiphysics/Kratos/commit/6c664d645bdda91f69209c82ba296f76dbfe4fb0)


kratos/processes/check_same_modelpart_using_skin_distance_process.cpp
```diff
@@ -0,0 +1,179 @@
...
+        return rNode.Y();
+    });
+    const double min_y = bounding_box_scale_factor * block_for_each<MinReduction<double>>(r_skin_model_part_1.Nodes(), [&](NodeType& rNode) {
+        return rNode.Y();
+    });
+    const double max_z = bounding_box_scale_factor * block_for_each<MaxReduction<double>>(r_skin_model_part_1.Nodes(), [&](NodeType& rNode) {
+        return rNode.Z();
+    });
+    const double min_z = bounding_box_scale_factor * block_for_each<MinReduction<double>>(r_skin_model_part_1.Nodes(), [&](NodeType& rNode) {
+        return rNode.Z();
```
philbucher: use a combined reduction, then you only need to loop once over the nodes

Also why not use the existing BB?

loumalouomega: First thing, true. Second thing, the structured mesher expects a hexahedra or quadrilateral

kratos/processes/check_same_modelpart_using_skin_distance_process.h
```diff
@@ -0,0 +1,225 @@
...
+    ~CheckSameModelPartUsingSkinDistanceProcess() override = default;
+
+    ///@}
+    ///@name Operators
+    ///@{
+
+    void operator()()
+    {
+        Execute();
+    }
```
rubenzorrilla:
```suggestion
```
I don't think this is a good idea. I'd stick to the `Execute` only to make clear what is being done.

kratos/processes/check_same_modelpart_using_skin_distance_process.cpp
```diff
@@ -0,0 +1,172 @@
...
+        "number_of_divisions_background_mesh" : 30,
+        "discontinuous_distance_settings": {
+            "elemental_distances_variable"                          : "ELEMENTAL_DISTANCES",
+            "elemental_edge_distances_variable"                     : "ELEMENTAL_EDGE_DISTANCES",
+            "elemental_edge_distances_extrapolated_variable"        : "ELEMENTAL_EDGE_DISTANCES_EXTRAPOLATED",
+            "embedded_velocity_variable"                            : "EMBEDDED_VELOCITY",
+            "calculate_elemental_edge_distances"                    : false,
+            "calculate_elemental_edge_distances_extrapolated"       : false,
+            "use_positive_epsilon_for_zero_values"                  : true
+        }
```
rubenzorrilla: If we eventually change the discontinuous distance settings we'll require to update this manually. To avoid so, I'd call the discontinuous distance `GetDefaultParameters` to get these and place them in the `discontinuous_distance_settings`.
```suggestion
        "discontinuous_distance_settings": {
        }
```

kratos/processes/check_same_modelpart_using_skin_distance_process.cpp
```diff
@@ -0,0 +1,172 @@
...
+/***********************************************************************************/
+
+template<std::size_t TDim>
+const Parameters CheckSameModelPartUsingSkinDistanceProcess<TDim>::GetDefaultParameters() const
+{
+    KRATOS_TRY
+
+    const Parameters default_parameters = Parameters(R"(
+    {
+        //"model_part_1_name"                   : "PLEASE_SPECIFY_MODEL_PART_1_NAME", // The background mesh is generated automatically
```
rubenzorrilla: Why adding commented fields to the defaults?

kratos/processes/check_same_modelpart_using_skin_distance_process.cpp
```diff
@@ -0,0 +1,172 @@
...
+    });
+    average_length /= static_cast<double>(total_number_conditions);
+
+    // Now we check that the difference is near a tolerance
+    const double tolerance = average_length * mThisParameters["tolerance"].GetDouble();
+
+    // Interate over the elements
+    auto it_elem_begin_1 = r_model_part_1.ElementsBegin();
+    auto it_elem_begin_2 = r_model_part_2.ElementsBegin();
+    const Variable<Vector>& r_elem_dist_var = KratosComponents<Variable<Vector>>::Get(distance_parameters["elemental_distances_variable"].GetString());
```
rubenzorrilla:
```suggestion
    const auto& r_elem_dist_var = KratosComponents<Variable<Vector>>::Get(distance_parameters["elemental_distances_variable"].GetString());
```

rubenzorrilla (changes requested):
Shouldn't you check at some point that the condition number is non-zero?

I think you should also check that you have line conditions in 2D and surface-type ones in 3D. As in the case of point-type conditions you cannot compute any distance. I think that checking that the first condition of each of the skin parts fulfils this requirement is enough.

kratos/processes/check_same_modelpart_using_skin_distance_process.cpp
loumalouomega: Leftover

loumalouomega:
> Shouldn't you check at some point that the condition number is non-zero?
> 
> I think you should also check that you have line conditions in 2D and surface-type ones in 3D. As in the case of point-type conditions you cannot compute any distance. I think that checking that the first condition of each of the skin parts fulfils this requirement is enough.

Done

kratos/processes/check_same_modelpart_using_skin_distance_process.cpp
```diff
@@ -0,0 +1,230 @@
...
+        "create_skin_sub_model_part" : false
+    })");
+    mesher_parameters["number_of_divisions"].SetInt(mThisParameters["number_of_divisions_background_mesh"].GetInt());
+    ModelPart& r_model_part_1 = mrModel.CreateModelPart("BACKGROUND_MESH_1");
+    const double bounding_box_scale_factor = mThisParameters["bounding_box_scale_factor"].GetDouble();
+    if constexpr (TDim == 2) { // 2D
+        mesher_parameters["element_name"].SetString("Element2D3N");
+        using BBReduction = CombinedReduction<MaxReduction<double>, MinReduction<double>, MaxReduction<double>, MinReduction<double>>;
+        double max_x, min_x, max_y, min_y;
+        std::tie(max_x, min_x, max_y, min_y) = block_for_each<BBReduction>(r_skin_model_part_1.Nodes(), [&](NodeType& rNode) {
```
philbucher, line 69: can you please try to use structured bindings here?

```suggestion
        auto [max_x, min_x, max_y, min_y] = block_for_each<BBReduction>(r_skin_model_part_1.Nodes(), [&](NodeType& rNode) {
```

kratos/processes/check_same_modelpart_using_skin_distance_process.cpp
```diff
@@ -0,0 +1,230 @@
...
+        auto p_point_1 = Kratos::make_intrusive<NodeType>(1, min_x, min_y, 0.0);
+        auto p_point_2 = Kratos::make_intrusive<NodeType>(2, max_x, min_y, 0.0);
+        auto p_point_3 = Kratos::make_intrusive<NodeType>(3, max_x, max_y, 0.0);
+        auto p_point_4 = Kratos::make_intrusive<NodeType>(4, min_x, max_y, 0.0);
+        Quadrilateral2D4<NodeType> geometry(p_point_1, p_point_2, p_point_3, p_point_4);
+        StructuredMeshGeneratorProcess(geometry, r_model_part_1, mesher_parameters).Execute();
+    } else { // 3D
+        using BBReduction = CombinedReduction<MaxReduction<double>, MinReduction<double>, MaxReduction<double>, MinReduction<double>, MaxReduction<double>, MinReduction<double>>;
+        double max_x, min_x, max_y, min_y, max_z, min_z;
+        std::tie(max_x, min_x, max_y, min_y, max_z, min_z) = block_for_each<BBReduction>(r_skin_model_part_1.Nodes(), [&](NodeType& rNode) {
```
philbucher, line 87: same

kratos/processes/check_same_modelpart_using_skin_distance_process.cpp
loumalouomega: Doesn't work

kratos/processes/check_same_modelpart_using_skin_distance_process.cpp
loumalouomega: Same

kratos/processes/check_same_modelpart_using_skin_distance_process.cpp
philbucher: Hm thought this would work 🤔 
I am curious, can you please post the error?

rubenzorrilla (approved):

kratos/processes/check_same_modelpart_using_skin_distance_process.cpp
loumalouomega: Now look like it works, I will do a minor PR, sorry for the noise

PR #10688, opened 2023-01-27
[GeoMechanicsApplication] Fixed several cases where some piece of code was unreachable

avdg81:
**📝 Description**
Fixed the following bugs reported by SonarQube (regarding branch `master`):
- code will never be executed
- `return` will never be executed
- `break` will never be executed

**🆕 Changelog**
- Removed some redundant code and variables
- Rewrote some conditional inclusion (`#ifdef [...] #elif [...] #else [...] #endif`)


carloslubbers (approved):
Looks good

PR #10689, opened 2023-01-27
[TrinilinosApplication] Cleaning up Trilinos includes

loumalouomega:
**📝 Description**

 `#include "trilinos_space.h"` already include most of the Trilinos dependencies, so in this PR I am cleaning the duplicates on includes due to this inclusion. 

**🆕 Changelog**

- [Cleaning up Trilinos includes](https://github.com/KratosMultiphysics/Kratos/commit/9e3e7ac5f867d01f93bb4c0d41fede32c9863f87)
- Some clean up

### DIFF
--- a/applications/TrilinosApplication/external_includes/amgcl_mpi_solver_impl.cpp
+++ b/applications/TrilinosApplication/external_includes/amgcl_mpi_solver_impl.cpp
@@ -1,3 +1,18 @@
+//  KRATOS  _____     _ _ _
+//         |_   _| __(_) (_)_ __   ___  ___
+//           | || '__| | | | '_ \ / _ \/ __|
+//           | || |  | | | | | | | (_) \__
+//           |_||_|  |_|_|_|_| |_|\___/|___/ APPLICATION
+//
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
+//
+//  Main authors:    
+//
+
+// System includes
+
+// External includes
 #include <boost/range/iterator_range.hpp>
 #include <boost/property_tree/ptree.hpp>
 
@@ -19,8 +34,7 @@
 #  include <amgcl/backend/vexcl_static_matrix.hpp>
 #endif
 
-#include "Epetra_FECrsMatrix.h"
-#include "Epetra_FEVector.h"
+// Project includes
 #include "trilinos_space.h"
 #include "custom_utilities/trilinos_solver_utilities.h"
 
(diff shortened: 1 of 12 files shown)


applications/TrilinosApplication/external_includes/amgcl_mpi_solver_impl.cpp
```diff
@@ -1,3 +1,18 @@
+//  KRATOS  _____     _ _ _
+//         |_   _| __(_) (_)_ __   ___  ___
+//           | || '__| | | | '_ \ / _ \/ __|
+//           | || |  | | | | | | | (_) \__
+//           |_||_|  |_|_|_|_| |_|\___/|___/ APPLICATION
+//
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
+//
+//  Main authors:
```
rubenzorrilla: I'd take the chance to add the authors.

loumalouomega: Who?

loumalouomega: I assume these ones

philbucher (changes requested):
I do not like these changes
What is your motivation?

I am pretty hesitant changing something that (see the failing CI) can easily break with older versions of trilinos, which is often the case on the cluster

Usually I am all up for cleaning, but in this particular case I would not change a working system

loumalouomega:
> I do not like these changes What is your motivation?

Cleaning up, looks like I removed more includes than supposed, some classes have similar name, and maybe I confused some names

> I am pretty hesitant changing something that (see the failing CI) can easily break with older versions of trilinos, which is often the case on the cluster
> 
> Usually I am all up for cleaning, but in this particular case I would not change a working system

loumalouomega:
> I do not like these changes What is your motivation?
> 
> I am pretty hesitant changing something that (see the failing CI) can easily break with older versions of trilinos, which is often the case on the cluster
> 
> Usually I am all up for cleaning, but in this particular case I would not change a working system

Funny thing is that in this PR I discovered we are using inconsistently `Epetra_FEVector` and `Epetra_Vector`, the first derives from the second but is not the same. This could be conflicting...

philbucher:
> Funny thing is that in this PR I discovered we are using inconsistently `Epetra_FEVector` and `Epetra_Vector`, the first derives from the second but is not the same. This could be conflicting...

The FE versions have the assembly functionalities (same for the FECrsMatrix)

loumalouomega:
> > Funny thing is that in this PR I discovered we are using inconsistently `Epetra_FEVector` and `Epetra_Vector`, the first derives from the second but is not the same. This could be conflicting...
> 
> The FE versions have the assembly functionalities (same for the FECrsMatrix)

Yes i know, but we are mixing them, I don't know if in purpose

philbucher (approved):
Now is good

PR #10691, opened 2023-01-27
[TrilinosApplication] Extend TrilinosSpace in order to add some methods missing from UblasSpace

loumalouomega:
**📝 Description**

**NOTE: THIS PR DOES NOT MODIFY TRILINOS BLOCK BUILDER AND SOLVER, BUT BRINGS A METHOD FROM UBLAS SPACE TO FUTURE UNIFICATION OF BEHAVIOUR BETWEEN SERIAL AND TRILINOS B&S**

Brings:

- `CheckAndCorrectZeroDiagonalValues`
- `GetScaleNorm`
- `GetDiagonalNorm`
- `GetAveragevalueDiagonal`
- `GetMaxDiagonal`
- `GetMinDiagonal`

Also adds the following methods required by the previous ones and interesting in general:

- `Max` (vector)
- `Min` (vector)

Also adds corresponding tests and the infrastructure (apparently there weren't any C++ tests in TrilinosApplication)

**🆕 Changelog**

- [Extend TrilinosSpace](https://github.com/KratosMultiphysics/Kratos/commit/d65a9b710d8a924b652dc2f43d4061ac710f2d09)
- [Addings tests for Trilinos spaces](https://github.com/KratosMultiphysics/Kratos/commit/b8ba2b91d497ec10bc6ac7cab603755b19228a76)
- [Adding tests script](https://github.com/KratosMultiphysics/Kratos/commit/779f3e166d38cdd2d679302d7ed8f2a3800caae5)
- [Update tests](https://github.com/KratosMultiphysics/Kratos/commit/376f5631cc6323f8d8c5a749133a1c09b752b058)
- [Improve CMake](https://github.com/KratosMultiphysics/Kratos/commit/ecc3181b4a0befec7a8f77001a36c7cc0c34540e)
- Adding more documentation

PR #10698, opened 2023-01-30
[TrilinosApplication] Adding tests for Builder and Solver

loumalouomega:
**📝 Description**

**NOTE:** @RiccardoRossi the elimination builder and solver does not work with more than 1 core. 

This adds the tests for the Builder and Solvers in TrilinosApplication, based in the implementations of the same tests for the serial version. This is an intermediate step before the next changes coming, and I need coverage of tests before start to do some changes.

Common parts with https://github.com/KratosMultiphysics/Kratos/pull/10691 .

**🆕 Changelog**

- [Adding B&S tests to TrilinosApplication](https://github.com/KratosMultiphysics/Kratos/commit/532df74a2d604d51fab94f05ca2698061be83838)
- [Adding tests script](https://github.com/KratosMultiphysics/Kratos/commit/779f3e166d38cdd2d679302d7ed8f2a3800caae5)
- [Update tests](https://github.com/KratosMultiphysics/Kratos/commit/376f5631cc6323f8d8c5a749133a1c09b752b058)
- [Improve CMake](https://github.com/KratosMultiphysics/Kratos/commit/ecc3181b4a0befec7a8f77001a36c7cc0c34540e)


applications/TrilinosApplication/tests/cpp_tests/test_builder_and_solver.cpp
```diff
@@ -0,0 +1,893 @@
...
+            r_node.Fix(DISPLACEMENT_Y);
+            r_node.Fix(DISPLACEMENT_Z);
+        }
+        // Fix X in first node
+        if (rModelPart.HasNode(1)) {
+            auto pnode = rModelPart.pGetNode(1);
+            pnode->Fix(DISPLACEMENT_X);
+        }
+
+        // TODO: Fix this once MPC work!
```
philbucher, line 174: Instead of commenting I would skip the test

Then it is more visible, and you are sure that the code will compile

Same for the other tests

applications/TrilinosApplication/tests/cpp_tests/test_builder_and_solver.cpp
```diff
@@ -0,0 +1,893 @@
...
+                NodeType::Pointer pnode10 = rModelPart.CreateNewNode(10, 2.0, 0.0, 0.0);
+
+                /// Add PARTITION_INDEX
+                for (auto& r_node : rModelPart.Nodes()) {
+                    r_node.FastGetSolutionStepValue(PARTITION_INDEX) = rank;
+                }
+                pnode9->FastGetSolutionStepValue(PARTITION_INDEX) = 1;
+                pnode10->FastGetSolutionStepValue(PARTITION_INDEX) = 1;
+
+                GeometryType::Pointer pgeom2 = Kratos::make_shared<Line2D2<NodeType>>(PointerVector<NodeType>{std::vector<NodeType::Pointer>({pnode10, pnode8})});
```
philbucher, line 407: Seems like you could use a loop 😅

philbucher (commented):
Minor, rest looks good to me

applications/TrilinosApplication/tests/cpp_tests/test_builder_and_solver.cpp
loumalouomega: I will add the in a near future, this is an intermediate PR. I prefer to do it with a comment

applications/TrilinosApplication/tests/cpp_tests/test_builder_and_solver.cpp
loumalouomega: Because both nodes are in the interface of the partition, I don't see any advance in doing a loop to save 1 line of code when in general may not be the case that they are continuous.

philbucher (approved):

PR #10711, opened 2023-02-01
Fixed a few slicing issues reported by SonarQube

avdg81:
**📝 Description**
Due to copying of polymorphic objects slicing occurred. To remedy the problem I have replaced some copies by references. Also, I have used explicit type names when needed, and type deduction otherwise.

**🆕 Changelog**
- Replaced some values by references to objects
- Replaced some explicit type names by `auto`
- Only for `Vx`, `Vy`, and `Vz` I wanted to be explicit about their types, to avoid problems when using these objects

### DIFF
--- a/applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
+++ b/applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
@@ -190,26 +190,22 @@ void UPwFaceLoadInterfaceCondition<3,4>::CheckJointWidth(double& rJointWidth, bo
                                                             const double& MinimumJointWidth, const Element::GeometryType& Geom)
 {
     //Quadrilateral_interface_3d_4
-    array_1d<double, 3> pmid0;
-    array_1d<double, 3> pmid1;
-    array_1d<double,3> P2 = Geom.GetPoint( 2 );
-    noalias(pmid0) = 0.5 * (Geom.GetPoint( 0 ) + Geom.GetPoint( 3 ));
-    noalias(pmid1) = 0.5 * (Geom.GetPoint( 1 ) + P2);
+    const auto& rP2 = Geom.GetPoint( 2 );
+    const auto pmid0 = 0.5 * (Geom.GetPoint( 0 ) + Geom.GetPoint( 3 ));
+    const auto pmid1 = 0.5 * (Geom.GetPoint( 1 ) + rP2);
     
     //Unitary vector in local x direction
-    array_1d<double,3> Vx;
-    noalias(Vx) = pmid1 - pmid0;
-    double inv_norm_x = 1.0/norm_2(Vx);
+    array_1d<double,3> Vx = pmid1 - pmid0;
+    const double inv_norm_x = 1.0/norm_2(Vx);
     Vx[0] *= inv_norm_x;
     Vx[1] *= inv_norm_x;
     Vx[2] *= inv_norm_x;
     
     //Unitary vector in local z direction
-    array_1d<double,3> Vy;
-    noalias(Vy) = P2 - pmid0;
+    array_1d<double, 3> Vy = rP2 - pmid0;
     array_1d<double,3> Vz;
-    MathUtils<double>::CrossProduct(Vz,Vx,Vy);
-    double norm_z = norm_2(Vz);
+    MathUtils<double>::CrossProduct(Vz, Vx, Vy);
+    const double norm_z = norm_2(Vz);
     if(norm_z > 1.0e-8)
     {
         Vz[0] *= 1.0/norm_z;
--- a/applications/GeoMechanicsApplication/custom_conditions/U_Pw_normal_lysmer_absorbing_condition.cpp
+++ b/applications/GeoMechanicsApplication/custom_conditions/U_Pw_normal_lysmer_absorbing_condition.cpp
@@ -618,23 +618,19 @@ template< >
 void UPwLysmerAbsorbingCondition<3, 4>::CalculateRotationMatrix( BoundedMatrix<double, 3, 3>& rRotationMatrix, const Element::GeometryType& rGeom)
 {
     //Quadrilateral_3d_4
-    array_1d<double, 3> pmid0;
-    array_1d<double, 3> pmid1;
-    const array_1d<double, 3> P2 = array_1d<double, 3>(rGeom.GetPoint(2));
-    noalias(pmid0) = 0.5 * (rGeom.GetPoint(0) + rGeom.GetPoint(3));
-    noalias(pmid1) = 0.5 * (rGeom.GetPoint(1) + P2);
+    const auto& rP2 = rGeom.GetPoint(2);
+    const auto pmid0 = 0.5 * (rGeom.GetPoint(0) + rGeom.GetPoint(3));
+    const auto pmid1 = 0.5 * (rGeom.GetPoint(1) + rP2);
 
     //Unitary vector in local x direction
-    array_1d<double, 3> Vx;
-    noalias(Vx) = pmid1 - pmid0;
+    array_1d<double, 3> Vx = pmid1 - pmid0;
     const double inv_norm_x = 1.0 / norm_2(Vx);
     Vx[0] *= inv_norm_x;
     Vx[1] *= inv_norm_x;
     Vx[2] *= inv_norm_x;
 
     //Unitary vector in local z direction
-    array_1d<double, 3> Vy;
-    noalias(Vy) = P2 - pmid0;
+    array_1d<double, 3> Vy = rP2 - pmid0;
     array_1d<double, 3> Vz;
     MathUtils<double>::CrossProduct(Vz, Vx, Vy);
     const double norm_z = norm_2(Vz);


applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
```diff
@@ -190,26 +190,24 @@ void UPwFaceLoadInterfaceCondition<3,4>::CheckJointWidth(double& rJointWidth, bo
...
 {
     //Quadrilateral_interface_3d_4
-    array_1d<double, 3> pmid0;
-    array_1d<double, 3> pmid1;
-    array_1d<double,3> P2 = Geom.GetPoint( 2 );
-    noalias(pmid0) = 0.5 * (Geom.GetPoint( 0 ) + Geom.GetPoint( 3 ));
-    noalias(pmid1) = 0.5 * (Geom.GetPoint( 1 ) + P2);
+    const auto& P2 = Geom.GetPoint( 2 );
+    const auto& pmid0 = 0.5 * (Geom.GetPoint( 0 ) + Geom.GetPoint( 3 ));
+    const auto& pmid1 = 0.5 * (Geom.GetPoint( 1 ) + P2);
```
rubenzorrilla: Very minor comment, but we try to name reference variables with an `r` in front. We also use `snake_case` for local variables. You can find some guidelines in our [style guide](https://github.com/KratosMultiphysics/Kratos/wiki/Style-Guide).

applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
```diff
@@ -190,26 +190,24 @@ void UPwFaceLoadInterfaceCondition<3,4>::CheckJointWidth(double& rJointWidth, bo
...
     
     //Unitary vector in local x direction
-    array_1d<double,3> Vx;
-    noalias(Vx) = pmid1 - pmid0;
-    double inv_norm_x = 1.0/norm_2(Vx);
+    const auto& tmpVx = pmid1 - pmid0;
+    array_1d<double,3> Vx{tmpVx[0], tmpVx[1], tmpVx[2]};
+    const auto inv_norm_x = 1.0/norm_2(Vx);
```
rubenzorrilla:
```suggestion
    const double inv_norm_x = 1.0/norm_2(Vx);
```
We avoid using `auto` for basic types. Maybe it is also a good idea to check that `Vx` is non-zero here.

applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
```diff
@@ -190,26 +190,24 @@ void UPwFaceLoadInterfaceCondition<3,4>::CheckJointWidth(double& rJointWidth, bo
...
+    const auto& P2 = Geom.GetPoint( 2 );
+    const auto& pmid0 = 0.5 * (Geom.GetPoint( 0 ) + Geom.GetPoint( 3 ));
+    const auto& pmid1 = 0.5 * (Geom.GetPoint( 1 ) + P2);
     
     //Unitary vector in local x direction
-    array_1d<double,3> Vx;
-    noalias(Vx) = pmid1 - pmid0;
-    double inv_norm_x = 1.0/norm_2(Vx);
+    const auto& tmpVx = pmid1 - pmid0;
+    array_1d<double,3> Vx{tmpVx[0], tmpVx[1], tmpVx[2]};
```
rubenzorrilla:
```suggestion
    const array_1d<double,3> Vx = pmid1 - pmid0;
```
Wouldn't it be better to keep it like this? I don't see the point on having the temporary.

applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
```diff
@@ -190,26 +190,24 @@ void UPwFaceLoadInterfaceCondition<3,4>::CheckJointWidth(double& rJointWidth, bo
...
     //Unitary vector in local z direction
-    array_1d<double,3> Vy;
-    noalias(Vy) = P2 - pmid0;
+    const auto& tmpVy = P2 - pmid0;
+    array_1d<double,3> Vy{tmpVy[0], tmpVy[1], tmpVy[2]};
     array_1d<double,3> Vz;
-    MathUtils<double>::CrossProduct(Vz,Vx,Vy);
-    double norm_z = norm_2(Vz);
+    MathUtils<double>::CrossProduct(Vz, Vx, Vy);
+    const auto norm_z = norm_2(Vz);
```
rubenzorrilla: (same comments above)

applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
matekelemen:
```cpp
const auto& pmid0 = 0.5 * (Geom.GetPoint( 0 ) + Geom.GetPoint( 3 ));
```

I was getting ready to bash on about `pmid0` immediately becoming a dangling ref, but TIL that this is actually legal. Thanks!

[Here](https://herbsutter.com/2008/01/01/gotw-88-a-candidate-for-the-most-important-const/) are some details if anyone's wondering what I'm going on about.

So regarding the naming @rubenzorrilla, despite what the declaration would imply, that variable acts like a value would so I'd keep the name.

huhaas: Even though it is legal I prefer the variation without &. It is easier to understand and reason about. Computational wise they are the same.
The explanation in the link is correct but it is an exception in the language (see: https://en.cppreference.com/w/cpp/language/lifetime )
However, not everybody knows those exceptions.

matekelemen: > Even though it is legal I prefer the variation without &

same here

applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
huhaas: What about something like this:
`auto Vx = static_cast<array_1d<double,3>>(pmid1 - pmid0);`
to avoid the slicing warnings.

applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
rubenzorrilla: Agree.

applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
rubenzorrilla: Not 100% sure, but I've the feeling that you get the warning because of what is discussed above. I hope that following @huhaas suggestion of removing the `&` would remove the slicing warning too.

applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
avdg81: > Very minor comment, but we try to name reference variables with an `r` in front. We also use `snake_case` for local variables. You can find some guidelines in our [style guide](https://github.com/KratosMultiphysics/Kratos/wiki/Style-Guide).

Thank you for the comment. I have renamed the variable according to your suggestion. I would like to add that we may consider to use a clang format file to aid contributors in following Kratos' Style Guide. If you feel that is a good suggestion I'm willing to help set that up.

matekelemen: That's a good idea!

avdg81: > ```c++
> const auto& pmid0 = 0.5 * (Geom.GetPoint( 0 ) + Geom.GetPoint( 3 ));
> ```
> 
> I was getting ready to bash on about `pmid0` immediately becoming a dangling ref, but TIL that this is actually legal. Thanks!
> 
> [Here](https://herbsutter.com/2008/01/01/gotw-88-a-candidate-for-the-most-important-const/) are some details if anyone's wondering what I'm going on about.
> 
> So regarding the naming @rubenzorrilla, despite what the declaration would imply, that variable acts like a value would so I'd keep the name.

Thanks for not bashing, but following a more constructive approach. I appreciate that. Bashing is not beneficiary in any way. Constructive criticism is.

avdg81: > Even though it is legal I prefer the variation without &. It is easier to understand and reason about. Computational wise they are the same. The explanation in the link is correct but it is an exception in the language (see: https://en.cppreference.com/w/cpp/language/lifetime ) However, not everybody knows those exceptions.

I was not aware that the C++ Core Guidelines state that binding a reference-to-const to a temporary should be an exception. Thanks for pointing that out. After our internal discussion I feel that the modified code is better. Thanks for your contribution.

applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
avdg81: Can you please explain what is the rationale behind not using `auto` for basic types (I presume that with "basic types" you refer to built-in types like `int`, `bool`, etc). When it comes to `auto` I tend to follow the [AAA-style](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/) for the reasons set forward by Herb Sutter.

I do agree that it is a good idea to check that `Vx` is non-zero here. I will take care of that in a separate commit. Thanks.

applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
avdg81: I have modified this after an internal discussion with @huhaas

applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
avdg81: Done.

applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
matekelemen: I think some people got burnt by using `auto` with expression templates (`auto` would bind to an expression instead of forcing an evaluation, potentially causing performance issues later down the line due to unwanted lazy evaluation - a nasty bug to find indeed) and decided never to use it again.

As for built-ins, I'm not sure. I can imagine a scenario where one would unintentionally mix signed and unsigned indices/IDs that could potentially cause over/underflow, but
1) that's not the case here
2) the compiler warns about these cases so we'd be better off treating these warnings as errors.

huhaas (approved):

applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
rubenzorrilla: Besides @matekelemen observation, at some point we took into account more "practical" reasons such as avoid accidentally using a "wrong" type (float / int divisions and things of this sort) and code readability.

Issue #83, opened 2017-02-27
compilation of a testing exe


RiccardoRossi:
@roigcarlo i think it would be better to create a testing library (complete with a python interface) AND an exe file for the tests. This way the KratosCore would not depend on the tests, but the tests might depend on the kratoscore.

i think this would imply doing something like the following in the CMakeLists 

if(${KRATOS_BUILD_TESTING} MATCHES ON)
	file(GLOB_RECURSE KRATOS_TESTING_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/tests/*.cpp)
        add_library(KratosTests ${KRATOS_CORE_TESTING_SOURCES})
        add_executable(KratosTesting  ${CMAKE_CURRENT_SOURCE_DIR}/testing_main.cpp)
        target_link_libraries(KratosTesting KratosCore)
endif(${KRATOS_BUILD_TESTING} MATCHES ON)

roigcarlo:
I like the idea of decoupling the tests from the core :+1: 

About the executable file, no problem as well but I have some concerns:
* I recall @pooyan-dadvand telling me he doesn't want C++ executable files ( or I am making it up? sorry my memory... )
* How should this behave? If we are going to make it something "official" we must speak about  some topics such as: 
    - Is it going to be an interactive command line to run the tests?
    - Run all the tests without asking the user? 
    - Should we be able to pass arguments?

pooyan-dadvand:
- actually I like to have a direct executable for cpp tests because it makes the debugging and especially profiling much easier

- some command line ese which accepts arguments to define tests or suits would be enough

pooyan-dadvand:
Technical committee: The idea is to create a CTest executable for core and each application

loumalouomega:
Will be compatible with CDash?

https://cmake.org/related-software/

pooyan-dadvand:
It should

loumalouomega:
Can we consider this as solved?, or not until we add the CTest compatibilty?

roigcarlo:
It is not solved, tests are still compiled as part of the owner application.

I tried to split it in the past abd it gave me some problems in win, but I am pretty confident in being able to do it now. I can spend some time to it next week.

philbucher:
if the tests are in a separate exe then we also won't have to link the entire Kratos again after changing a test in the core right?

roigcarlo:
As a side effect. My idea is not only create a separate binary but also a separate lib with the tests. so te core will have:
[...]

PR #10718, opened 2023-02-02
[DRAFT] Experimenting with testing libs.

roigcarlo:
**📝 Description**
This is a draft comes from a discussion in #10693 and wants to be a preview on how this can be ported to Kratos as a whole. This has been a long overdo from our side (~10.5K items after #83 ...)

The aim is to provide a mechanism to generate stand-alone unitary test for the C++ layer of Kratos. Both for the kernel and for the apps, without having to perform any major refactor.

Disclaimer:
- **Why gtest and not doctest/xxxxtest/....**: So far, because I need mocking to make the kernel fixture and some very handy matching patterns for the asserts that are not present in doctest. This does not mean we will use this backed. Its just that it comes in handy to prototype something fast without spending much time.

Summary:
- Checks and Asserts/Expects are now separate classes. This means that you can still keep using the old `KRATOS_CHECK_XXXX` if you want to throw exceptions or check assertions in debug mode etc... but also have the same family of macros to produce results for asserts in checks.
- Tests are now compiled as individual executable. Each *.cpp test file defines its own suite. This is done automatically through CMake but is not the ideal solution. I would change this to better organize tests and reuse some helper classes, but as for now it let us use the current tests without changes.
- I am aware that are tests failing. To my knowledge:
  - The test testing the internal tester class (for obvious reasons)
  - Tests that check exception throwing for out of bounds accesses or other undefined behaviors, as are not captured (and shouldn't be)
  - Some few tests that are reporting memleaks. I will investigate those individually, but I suspect that we are detecting hidden errors derived from the fact of using a new Kernel in each test.


loumalouomega (changes requested):
Blocking for the moment

loumalouomega:
There are plenty of conflicts @roigcarlo

roigcarlo:
I know. I created the branch with a perfect timing, as always.

loumalouomega:
> I know. I created the branch with a perfect timing, as always.

Todo ha salido deacuerdo con el flan

huhaas:
I like the redirect macros to have an easier transition.

Creating a dependency on GoogleTest can be done with this instruction:
https://google.github.io/googletest/quickstart-cmake.html#set-up-a-project
Locally it works and looks quite elegant. CMake will take care of the building and include logic.
Later, updating will be as easy as updating a single hash.
The downside is that the build servers need to reach the zip.

I'm not in favor of creating individual executables per test file. The list of 'possible build targets' will explode.
Each executable will create its own test report file and has an execution overhead (I'm guessing this will be noticeable).
Combining test reports is a pain.

I would expect 1 executable for the core and 1 for each application.
The built-in argument mechanism should be sufficient to run a single or subset of tests.

PR #10721, opened 2023-02-02
Moved functionality to element utilities

### LINKED ISSUE #10719: [GeoMechanicsApplication] Add a custom element utilities
Add a custom element utilities, this is to reduce the duplication of codes.

mcgicjn2:
**📝 Description**
Moved CalculateHydraulicHeadsFromWaterPressures from element level to element utility, then can be used in multiple locations.

In order to reduce duplication

**🆕 Changelog**
- Moved functionality to utilities
- Updated 2 location - element level calculation and in DGeoFlow custom workflow

### DIFF
--- a/applications/GeoMechanicsApplication/custom_utilities/element_utilities.hpp
+++ b/applications/GeoMechanicsApplication/custom_utilities/element_utilities.hpp
@@ -669,32 +669,32 @@ public:
     //----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 
     template< unsigned int TNumNodes >
-    static inline array_1d<double, TNumNodes> CalculateNodalHydraulicHeadFromWaterPressures(const GeometryType& rGeom, const Properties& rProp)
+    static array_1d<double, TNumNodes> CalculateNodalHydraulicHeadFromWaterPressures(const GeometryType& rGeom, const Properties& rProp)
     {
-        const double NumericalLimit = std::numeric_limits<double>::epsilon();
+        const auto NumericalLimit = std::numeric_limits<double>::epsilon();
     	//Defining necessary variables
-        array_1d<double, TNumNodes> NodalHydraulicHead;
+        array_1d<double, TNumNodes> nodal_hydraulic_heads;
         for (unsigned int node = 0; node < TNumNodes; ++node) {
-            array_1d<double, 3> NodeVolumeAcceleration;
-            noalias(NodeVolumeAcceleration) = rGeom[node].FastGetSolutionStepValue(VOLUME_ACCELERATION, 0);
-            const double g = norm_2(NodeVolumeAcceleration);
+            array_1d<double, 3> node_volume_acceleration;
+            noalias(node_volume_acceleration) = rGeom[node].FastGetSolutionStepValue(VOLUME_ACCELERATION, 0);
+            const double g = norm_2(node_volume_acceleration);
             if (g > NumericalLimit) {
                 const double FluidWeight = g * rProp[DENSITY_WATER];
 
                 array_1d<double, 3> NodeCoordinates;
                 noalias(NodeCoordinates) = rGeom[node].Coordinates();
                 array_1d<double, 3> NodeVolumeAccelerationUnitVector;
-                noalias(NodeVolumeAccelerationUnitVector) = NodeVolumeAcceleration / g;
+                noalias(NodeVolumeAccelerationUnitVector) = node_volume_acceleration / g;
 
                 const double WaterPressure = rGeom[node].FastGetSolutionStepValue(WATER_PRESSURE);
-                NodalHydraulicHead[node] = -inner_prod(NodeCoordinates, NodeVolumeAccelerationUnitVector)
+                nodal_hydraulic_heads[node] = -inner_prod(NodeCoordinates, NodeVolumeAccelerationUnitVector)
                     - PORE_PRESSURE_SIGN_FACTOR * WaterPressure / FluidWeight;
             }
             else {
-                NodalHydraulicHead[node] = 0.0;
+                nodal_hydraulic_heads[node] = 0.0;
             }
         }
-        return NodalHydraulicHead;
+        return nodal_hydraulic_heads;
     }
 
 }; /* Class GeoElementUtilities*/
--- a/applications/GeoMechanicsApplication/custom_utilities/element_utilities.hpp
+++ b/applications/GeoMechanicsApplication/custom_utilities/element_utilities.hpp
@@ -665,6 +665,38 @@ public:
         }
      }
 
+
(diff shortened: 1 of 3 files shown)


applications/GeoMechanicsApplication/custom_utilities/element_utilities.hpp
```diff
@@ -665,6 +665,38 @@ typedef std::size_t IndexType;
         }
      }
 
+
+    //----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+
+    template< unsigned int TNumNodes >
+    static inline array_1d<double, TNumNodes> CalculateNodalHydraulicHeadFromWaterPressures(const GeometryType& rGeom, const Properties& rProp)
```
avdg81: Since this function is not really small and time-critical, I wouldn't `inline` it. See the [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#f5-if-a-function-is-very-small-and-time-critical-declare-it-inline) for details about when to `inline` functions.

applications/GeoMechanicsApplication/custom_utilities/element_utilities.hpp
```diff
@@ -665,6 +665,38 @@ typedef std::size_t IndexType;
...
+    {
+        const double NumericalLimit = std::numeric_limits<double>::epsilon();
```
avdg81: A minor thing: note that the type (`double`) is repeated here. Since you are explicit about it at the right hand side, I wouldn't repeat it at the left hand side. I think you'd better use `const auto` here.

applications/GeoMechanicsApplication/custom_utilities/element_utilities.hpp
```diff
@@ -665,6 +665,38 @@ typedef std::size_t IndexType;
...
+    	//Defining necessary variables
+        array_1d<double, TNumNodes> NodalHydraulicHead;
+        for (unsigned int node = 0; node < TNumNodes; ++node) {
+            array_1d<double, 3> NodeVolumeAcceleration;
```
avdg81: Again a minor thing: note that Kratos' Style Guide prescribes to [use snake case for local variable names](https://github.com/KratosMultiphysics/Kratos/wiki/Style-Guide#local-variables). So `NodeVolumeAcceleration` should actually be `node_volume_acceleration` if we want to comply with the Style Guide.

applications/GeoMechanicsApplication/custom_utilities/element_utilities.hpp
```diff
@@ -665,6 +665,38 @@ typedef std::size_t IndexType;
...
 
+
+    //----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+
+    template< unsigned int TNumNodes >
+    static inline array_1d<double, TNumNodes> CalculateNodalHydraulicHeadFromWaterPressures(const GeometryType& rGeom, const Properties& rProp)
+    {
+        const double NumericalLimit = std::numeric_limits<double>::epsilon();
+    	//Defining necessary variables
+        array_1d<double, TNumNodes> NodalHydraulicHead;
```
avdg81: A small renaming suggestion: to me `NodalHydraulicHead` suggests it holds a single value, but in fact it keeps a list of values, one for each node of the element. This may be expressed more clearly by using the plural form, i.e. `nodal_hydraulic_heads`. (Note that I have also converted from UpperCamelCase to snake_case to comply with Kratos' Style Guide. See also my previous comment.)

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -569,32 +569,15 @@ namespace Kratos
             for (Element element : model_part.Elements())
             {
                 auto rGeom = element.GetGeometry();
-                auto rProp = element.GetProperties();
+                const auto rProp = element.GetProperties();
```
avdg81: The names (`rGeom` and `rProp`) suggest that these are references, but `auto` doesn't deduce references. In this case a copy is being made, which is probably not what you intended. So, if you meant to have references, please append `&` to `auto`/`const auto`.

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -569,32 +569,15 @@ namespace Kratos
...
-                        const double WaterPressure = rGeom[node].FastGetSolutionStepValue(WATER_PRESSURE);
-                        rGeom[node].SetValue(*element_var, -inner_prod(NodeCoordinates, NodeVolumeAccelerationUnitVector) - PORE_PRESSURE_SIGN_FACTOR * WaterPressure / FluidWeight);
-                    }
-                    else
-                    {
-                        rGeom[node].SetValue(*element_var, 0.0);
-                    }
+            	for (unsigned int node = 0; node < 3; ++node)
+                {
+                    rGeom[node].SetValue(*element_var, NodalHydraulicHead[node]);
```
avdg81: A minor thing: if you convert `element_var` from a pointer to a reference-to-const, you no longer need to dereference it in the remainder of the function. Since `KratosComponents<Variable<double>>::Get("HYDRAULIC_HEAD")` returns `const Kratos::Variable<double>&`, you may simply write
```
const auto& element_var = KratosComponents<Variable<double>>::Get("HYDRAULIC_HEAD");
```
And where you use `element_var`, you'd remove the dereference operator (`*`).

mcgicjn2: Done

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
mcgicjn2: Done

applications/GeoMechanicsApplication/custom_utilities/element_utilities.hpp
mcgicjn2: I guess we need to do this over the whole code. Perhaps we should discuss this

applications/GeoMechanicsApplication/custom_utilities/element_utilities.hpp
mcgicjn2: Done. See comment below

applications/GeoMechanicsApplication/custom_utilities/element_utilities.hpp
mcgicjn2: Done

applications/GeoMechanicsApplication/custom_utilities/element_utilities.hpp
mcgicjn2: Done. Thanks for pointing it out

avdg81 (approved):
Looks fine to me.

PR #10722, opened 2023-02-03
[Core] STL IO WriteModelPart

qiukailu:
**📝 Description**
Add implementation of WriteModelPart for StlIO class.


kratos/input_output/stl_io.cpp
```diff
@@ -36,6 +37,33 @@ StlIO::StlIO(std::filesystem::path const& Filename)
...
+{
+    std::fstream* pFile = new std::fstream();
+    std::fstream::openmode OpenMode = std::fstream::in;
+
+    // Set the mode
+    if (mOptions.Is(IO::READ)) {
+        OpenMode = std::fstream::in;
+    } else if (mOptions.Is(IO::APPEND)) {
+        OpenMode = std::fstream::in | std::fstream::app;
+    } else if (mOptions.Is(IO::WRITE)) {
```
huhaas, line 37: I don't think writing functionality belongs in this class. All other member functions are read-oriented and none are protected from running in the other mode. E.g opening a file with the WRITE flag and calling read functions.

So it triggers a design question. Should this be added to the class and made safe or added to a separate class for writing only?

If this is the way to go:
The two constructors could be merged into one by using a default argument
`StlIO(std::filesystem::path const& Filename, const Flags Options = IO::READ)`
(if allowed by the code guidelines)
The `else` can be improved by throwing an error when it gets something unexpected.

kratos/input_output/stl_io.cpp
```diff
@@ -52,6 +80,33 @@ void StlIO::ReadModelPart(ModelPart & rThisModelPart)
...
+        const auto & rUnitNormal = rGeom.UnitNormal(rGeom.Center());
+        (*mpInputStream) << "    facet normal " << rUnitNormal[0] << " " << rUnitNormal[1] << " " << rUnitNormal[2] << "\n";
+    
+        (*mpInputStream) << "        outer loop\n";
+
+        for(std::size_t i = 0; i < rGeom.size(); i++ )
+        {
+            auto & rNode = rGeom.GetPoint(i);
+            (*mpInputStream) << "           vertex " << rNode.X() << " " << rNode.Y() << " " << rNode.Z() << "\n";
+        }
```
matekelemen: I don't know much about the STL format so I might be wrong, but I'm under the impression that it defines a **triangulation**, so you'd either have to restrict the input `ModelPart`s to ones that only contain triangle elements, or perform the triangulation yourself for any element.

kratos/input_output/stl_io.cpp
```diff
@@ -52,6 +80,33 @@ void StlIO::ReadModelPart(ModelPart & rThisModelPart)
         ReadSolid(rThisModelPart);
 }
 
+void StlIO::WriteModelPart(ModelPart & rThisModelPart)
+{    
+    (*mpInputStream) << "solid compensated_skin_part\n";
```
matekelemen: please don't hard code the name; you could pass it as a default string for example

kratos/input_output/stl_io.cpp
ddiezrod: This class derives from IO, that allows writing and reading, so I think it is perfectly normal that this class can write too (its implicit in the name also). The other member methods are private so they can be called witihin the class, so I think this is safe enough as you are only leaving this decission to the class itself. What do you think as the original designers @pooyan-dadvand  @loumalouomega ?

loumalouomega: I would do as ModelPartIO

kratos/input_output/stl_io.cpp
```diff
@@ -52,6 +80,33 @@ void StlIO::ReadModelPart(ModelPart & rThisModelPart)
         ReadSolid(rThisModelPart);
 }
 
+void StlIO::WriteModelPart(ModelPart & rThisModelPart)
```
philbucher:
```suggestion
void StlIO::WriteModelPart(const ModelPart & rThisModelPart)
```

the other version of this fct is deprecated

kratos/input_output/stl_io.cpp
```diff
@@ -52,6 +80,33 @@ void StlIO::ReadModelPart(ModelPart & rThisModelPart)
...
+{    
+    (*mpInputStream) << "solid compensated_skin_part\n";
+
+
+    for (auto & rElem : rThisModelPart.Elements()) {
```
philbucher: why elements?

what about geometries and conditions?

kratos/input_output/stl_io.cpp
ddiezrod: I agree with your last point of using a default argument @huhaas

kratos/input_output/stl_io.cpp
loumalouomega: +1

kratos/input_output/stl_io.cpp
```diff
@@ -17,17 +17,28 @@
...
 namespace Kratos
 {
 
-/// Constructor with  filenames.
-StlIO::StlIO(std::filesystem::path const& Filename)
-    : IO()
+StlIO::StlIO(std::filesystem::path const& Filename, const Flags Options)
+    : mOptions(Options)
 {
     std::fstream* pFile = new std::fstream();
```
matekelemen: I just noticed that this class does manual memory management :'(
It doesn't have to be in this PR, but someone must absolutely replace `mpInputStream` with a `unique_ptr` .

kratos/tests/cpp_tests/input_output/test_stl_io.cpp
```diff
@@ -90,4 +92,47 @@ endsolid 3 triangles
...
+    r_model_part.CreateNewNode(1, 0.1, -2.56114e-08, 0.1);
+    r_model_part.CreateNewNode(2, 0.1, -0.499156, -0.0352136);
+    r_model_part.CreateNewNode(3, 0.1, -0.473406, -0.0446259);
+
+    // create a triangle element
+    Properties::Pointer p_properties(new Properties(0));
+    r_model_part.CreateNewElement("Element3D3N", 101, {1, 2, 3}, p_properties);
+
+    // write stl
+    std::string filename = "test_stl_write.stl";
```
philbucher: Please use std::filesystem::path

kratos/tests/cpp_tests/input_output/test_stl_io.cpp
```diff
@@ -90,4 +92,47 @@ endsolid 3 triangles
...
+    StlIO * stl_write = new StlIO(filename, IO::WRITE);
```
philbucher: Instead of the manual memory allocation you should out those lines into a scope, which will have the same effect

kratos/tests/cpp_tests/input_output/test_stl_io.cpp
```diff
@@ -90,4 +92,47 @@ endsolid 3 triangles
...
+    KRATOS_CHECK_EQUAL(r_output_model_part.NumberOfNodes(), 3);
+    KRATOS_CHECK_EQUAL(r_output_model_part.NumberOfElements(), 1);
+    KRATOS_CHECK_EQUAL(r_output_model_part.NumberOfConditions(), 0);
+
+    KRATOS_CHECK_EQUAL(r_output_model_part.GetSubModelPart("Main").NumberOfNodes(), 3);
+    KRATOS_CHECK_EQUAL(r_output_model_part.GetSubModelPart("Main").NumberOfElements(), 1);
+    KRATOS_CHECK_EQUAL(r_output_model_part.GetSubModelPart("Main").NumberOfConditions(), 0);
+    
+    // remove the generated files
+    if (remove(filename.c_str()) != 0) {
```
philbucher: Please use std::filesystem::remove

kratos/tests/cpp_tests/input_output/test_stl_io.cpp
qiukailu: thanks for the suggestion!

kratos/input_output/stl_io.cpp
ddiezrod: Added

ddiezrod:
I think this is ready for final review.

kratos/input_output/stl_io.cpp
```diff
@@ -17,16 +17,29 @@
 // Project includes
 #include "includes/define.h"
 #include "input_output/stl_io.h"
+#include "utilities/parallel_utilities.h"
```
philbucher:
```suggestion
```

unused

philbucher (commented):
minor, otherwise LGTM

kratos/input_output/stl_io.cpp
ddiezrod: done!

philbucher (approved):

ddiezrod:
Thanks @philbucher !

PR #10726, opened 2023-02-03
[Core] Improve ReplaceElementsAndConditionsProcess

loumalouomega:
**📝 Description**

**NOTE: The current change is 100% backward compatible**

The main change is that in addition to consider just one type of element/condition it would be possible to specify a list of entities or a templated name (P.E: `Element#D#N`), in case we are considering a hybrid mesh, or we want to generalize our settings: The structure looks like this:

~~~cpp
// Compute process
Parameters settings( R"(
{
    "element_name"   : "Element2D3N;Element2D4N",
    "condition_name" : "LineCondition2D2N"
}  )" );

ReplaceElementsAndConditionsProcess process(r_model_part, settings);
process.Execute();
~~~

or :

~~~cpp
// Compute process
Parameters settings( R"(
{
    "element_name"   : "Element#D#N",
    "condition_name" : "LineCondition2D2N"
}  )" );

ReplaceElementsAndConditionsProcess process(r_model_part, settings);
process.Execute();
~~~

In addition to that, the code and tests have been cleaned up and updated to latests standards. (and yes a priori we should move this to a modeller, but in @KratosMultiphysics/altair  we have a heavy dependency in this process and for the moment we need to update this)

**🆕 Changelog**

- [#pragma once](https://github.com/KratosMultiphysics/Kratos/commit/03a8fcaa3af3f1c07a4d526b44b467d3ecde46dc)
- [Adding geometrytype check](https://github.com/KratosMultiphysics/Kratos/commit/b24641a98f319be69d10d7b0e8b242e471943d5f)
- [Making everything more generic](https://github.com/KratosMultiphysics/Kratos/commit/f545b8de03b623c99b236b0f6b21ec8cc5bdee61)
- [Clean up current test](https://github.com/KratosMultiphysics/Kratos/commit/54c917a321ff4df24b2a103fbaf8343b0d668103)
- [New test](https://github.com/KratosMultiphysics/Kratos/commit/f3920448bfabc29650f00186886cf3692db81ef6)
- [Refactor code](https://github.com/KratosMultiphysics/Kratos/pull/10726/commits/7a53e33ba1b748d5881e58ba14423e3ed7571fcb)
- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11376


rubenzorrilla (changes requested):
To be discussed as this is a sensitive one.

kratos/utilities/cpp_tests_utilities.cpp
```diff
@@ -69,8 +71,48 @@ void Create2DGeometry(
 /***********************************************************************************/
 /***********************************************************************************/
 
+void CreatePure2DGeometry(ModelPart& rModelPart)
```
rubenzorrilla: If you aim to add these to the common test utilities, I'd try to find a more self-descriptive name.

kratos/utilities/cpp_tests_utilities.cpp
```diff
@@ -159,6 +201,93 @@ void Create3DGeometry(
 /***********************************************************************************/
 /***********************************************************************************/
 
+void CreatePure3DGeometry(ModelPart& rModelPart)
```
rubenzorrilla: Same comment about the naming.

rubenzorrilla:
Wouldn't the problem be solved by saving each geometry type in a diferent submodelpart and calling current implementation for each of these?

philbucher:
isnt this the same as https://github.com/KratosMultiphysics/Kratos/blob/master/applications/StructuralMechanicsApplication/custom_response_functions/adjoint_processes/replace_multiple_elements_and_conditions_process.h?

loumalouomega:
> isnt this the same as [`master`/applications/StructuralMechanicsApplication/custom_response_functions/adjoint_processes/replace_multiple_elements_and_conditions_process.h](https://github.com/KratosMultiphysics/Kratos/blob/master/applications/StructuralMechanicsApplication/custom_response_functions/adjoint_processes/replace_multiple_elements_and_conditions_process.h?rgh-link-date=2023-02-04T16%3A56%3A35Z)?

It is not exactly the same, but we can unify them!

loumalouomega:
> Wouldn't the problem be solved by saving each geometry type in a diferent submodelpart and calling current implementation for each of these?

That's overcomplicated compared with this solution. We use intensively this process in @KratosMultiphysics/altair , and the problem is that:
 
1. We want to be 100% retrocompatible...
2. Yet, we want to be able to set the changes using the defaults in our implementations, without huge changes from our side.

What you suggest implies to do many changes in our side, we could simply solve this issue doing that changes only in our side, but then the community will lose this contribution which I consider beneficial for everyone. I would like now to check how to unify behaviour with suggestion by @philbucher .

kratos/utilities/cpp_tests_utilities.cpp
loumalouomega: Suggestion?, this was the best I could think.

rubenzorrilla:
> > Wouldn't the problem be solved by saving each geometry type in a diferent submodelpart and calling current implementation for each of these?
> 
> That's overcomplicated compared with this solution. We use intensively this process in @KratosMultiphysics/altair , and the problem is that:
> 
>     1. We want to be 100% retrocompatible...
> 
>     2. Yet, we want to be able to set the changes using the defaults in our implementations, without huge changes from our side.
> 
> 
> What you suggest implies to do many changes in our side, we could simply solve this issue doing that changes only in our side, but then the community will lose this contribution which I consider beneficial for everyone. I would like now to check how to unify behaviour with suggestion by @philbucher .

I don't think it's that complicated... I think you can do everything at Python level with a simple `ClassifyEntitiesByGeometryModeler`. OFC I'm not suggesting to do this at the GUI level.

loumalouomega:
I refactored the code following suggestions from @KratosMultiphysics/altair

kratos/processes/replace_elements_and_condition_process.cpp
```diff
@@ -20,22 +20,126 @@
...
 namespace Kratos {
 namespace {
 
-/// Replace entities in a given container if the entity id is present in a list of ids.
-    /*  @param rReferenceEntity New type of entity that will replace old one
-     *  @param rEntityContainer Container of elements susceptible to be replaces
-     *  @param rSetOfIds Set of entities ids we want to replace
-     */
-template <class TEntity, class TEntityContainer>
+std::string GetElementName(const GeometryData::KratosGeometryType ElementType)
```
philbucher: This is quite handy, can you please move this somewhere else? (Preferably separate PR for fat approval 😃)

We should have some utilities file where this would fit no?

rubenzorrilla: If so, I'd call it `GetGeometryName`.

loumalouomega:
@rubenzorrilla I applied the changes requested weeks ago

loumalouomega:
> @rubenzorrilla I applied the changes requested weeks ago

Now +1 month

loumalouomega:
> > @rubenzorrilla I applied the changes requested weeks ago
> 
> Now +1 month

Ping @rubenzorrilla

kratos/utilities/entities_utilities.h
```diff
@@ -32,6 +33,148 @@ namespace Kratos
...
+
+        ///@}
+    private:
+        ///@name Member Variables
+        ///@{
+
+        typename TEntity::Pointer mpPrototypeEntity = nullptr;                    /// The prototype entity
+        DefinitionType mDefinitionType = DefinitionType::Single;                  /// The type of definition
+        std::string mName;                                                        /// The name of the entity
+        std::unordered_map<GeometryData::KratosGeometryType, std::string> mTypes; /// The settings of the entities
```
RiccardoRossi: do you need to store the name of the entity? wouldn't it be better to store a pointer to the reference element/condition with that type?

in any case, let's imagine we want to apply two elements (or one element and a condition to the same geometry type). What would we do in that case?

loumalouomega: If you see the implementation you will see it is required for the templated case

RiccardoRossi (approved):
thank you for the patience in attending all of my suggestions @loumalouomega !

rubenzorrilla (approved):
Removible my block. Thanks for the patience.

PR #10706, opened 2023-02-01
[TrilinosApplication] Unify behavior between the serial/MPI BlockBuilderandSolver

loumalouomega:
**📝 Description**

Unify behaviour between the serial/MPI BlockBuilderAndSolver. Update the diagonal so it is not zero when the rest of the row is empty, as it was done for the `ResidualBasedEliminationBuilderAndSolver` in https://github.com/KratosMultiphysics/Kratos/pull/10105

**NOTE:** This will improve as well the condition number in MPI, but by default a 1 will be set i the diagonal as in serial.

**🆕 Changelog**

- [Cleaning](https://github.com/KratosMultiphysics/Kratos/commit/1e8f1176971e8c3eaf94f94522579e63a3c63a1d)
- [Improve constructors](https://github.com/KratosMultiphysics/Kratos/commit/e29398906b7776514831b475b0bde5be3c7609b9)
- [Adding flags](https://github.com/KratosMultiphysics/Kratos/commit/0dc40b5b990a70a6e6363969808559c0411be5c4)
- [Mising info](https://github.com/KratosMultiphysics/Kratos/commit/19b26d5f65ab36414d974e5cb9cfe7e79cdfed89)
- [Improve member variables](https://github.com/KratosMultiphysics/Kratos/commit/0f97a087ba612647e099407533ec99b75bd26a87)
- [Expose to python](https://github.com/KratosMultiphysics/Kratos/pull/10706/commits/b85798a05ac5106772f5fe1654c80e3dce750494)
- [Missing methods](https://github.com/KratosMultiphysics/Kratos/commit/b74d5b328a5d5d8367dd362da3f323e6d9ab46a0)
- [Fix diagonal](https://github.com/KratosMultiphysics/Kratos/commit/8f2c3abdc4b193e42de7f1f668b3df5f56289a79)
- [Update first test](https://github.com/KratosMultiphysics/Kratos/commit/90633c9a3ea302ecca3d74df90d30cadaccc0d3c)
- [More test result update](https://github.com/KratosMultiphysics/Kratos/commit/77ff7f42d58ca28345171ec3bb9b5e913eba1a54)

PR #10738, opened 2023-02-08
[Core] Consistent diagonal setting for constraints in BlockBuilderAndSolver

loumalouomega:
**📝 Description**

I realize that when setting constraints we always set the maximum value of the diagonal instead of the `mScaleFactor` as in the rest of the code. 

In addition, 2 new access methods are introduced:
- `GetScaleFactor()`: Returns the current value of the scale factor.
- `SetScaleFactor(double ScaleFactor)`: Sets a new value for the scale factor.

**🆕 Changelog**

- [Consistent diagonal setting for constraints](https://github.com/KratosMultiphysics/Kratos/commit/6d7c8c4874fb382126199c93d6e5057c27e0f1ed)
- [Better implementation with avoid unnecessary duplicated computation of the scale factor](https://github.com/KratosMultiphysics/Kratos/pull/10738/commits/cd0c57aac6fafcf2ef785780cedf8f78bead7bfc)
- [Giving up and calculate always](https://github.com/KratosMultiphysics/Kratos/pull/10738/commits/14f50dd02085b90cbe2d086f0455591b52fea9cb)


loumalouomega:
The builder and solver test fails as expected (I need to update the values), but the eigen tests I don't know why fails, @philbucher do you know these tests?

philbucher:
> The builder and solver test fails as expected (I need to update the values), but the eigen tests I don't know why fails, @philbucher do you know these tests?

no idea
but thats why I am scared of any changes to the B&S such as [this](https://github.com/KratosMultiphysics/Kratos/pull/10706/files#r1096068017)

loumalouomega:
> > The builder and solver test fails as expected (I need to update the values), but the eigen tests I don't know why fails, @philbucher do you know these tests?
> 
> no idea but thats why I am scared of any changes to the B&S such as [this](https://github.com/KratosMultiphysics/Kratos/pull/10706/files#r1096068017)

In that particular case the results obtained are the same as in serial. My question is why the eigen gives a so different value when there are 1s in the diagonal. I will investigate...

loumalouomega:
With the update of the default diagonal values should work now

loumalouomega:
> With the update of the default diagonal values should work now

I notice is not necessarily initialized, so I ensured it is initialized, this way we can avoid recompute the scale factor each time.

kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver.h
```diff
@@ -1200,6 +1221,7 @@ class ResidualBasedBlockBuilderAndSolver
     std::vector<IndexType> mMasterIds;                /// The equation ids of the master
     std::unordered_set<IndexType> mInactiveSlaveDofs; /// The set containing the inactive slave dofs
     double mScaleFactor = 1.0;                        /// The manually set scale factor
+    bool mScaleFactorInitialized = false;             /// If the value has been initialized
```
loumalouomega:
```suggestion

```

roigcarlo:
This also looks good to me. ping rest of the @KratosMultiphysics/technical-committee

rubenzorrilla (approved):
I agree @roigcarlo. Changes also look correct to me, including the changes in the test values.

loumalouomega:
Now I am going to make consistent the Trilinos one

PR #10740, opened 2023-02-10
[Core] Fix error when properties exists in read STL io

jrubiogonzalez:
**📝 Description**
This PR adds a minor fix when the properties exist but are not present in the destination modelpart, useful for combining mdpa's


ddiezrod (approved):
Ok to me

PR #10741, opened 2023-02-10
Compilation error in stl_io

AlejandroCornejo:
I had to add the `#include <cctype>` in the stl_io.h in order to compile in my Windows machine (VisualStudio 2017). In another windows machine is working without it :S

Any ideas?


loumalouomega:
@AlejandroCornejo I like your new profile picture

loumalouomega (dismissed):

matekelemen:
> aparently this is what's happening:
> 
> https://stackoverflow.com/questions/11059308/isspace-macro-conflicting-with-isspace-function-in-locale-facets-h

does that mean that we're including `ctype.h` somewhere? C includes should be replaced by their C++ counterparts.

philbucher:
what is the error?

AlejandroCornejo:
Apparently the error was the following: ""isspace" passed 2 arguments, but takes just 1".

Upgrading to Visual Studio 2019 solves the issue...

philbucher:
If it is necessary only in this case then I would protect this including with a ward, since other compilers don't need it and this header is included everywhere

PR #10747, opened 2023-02-15
[Core] Adding `GetGeometryName` to `GeometryUtils`

loumalouomega:
**📝 Description**

Adding `GetGeometryName` to `GeometryUtils`

**🆕 Changelog**

- [Adding `GetGeometryName` to `GeometryUtils`](https://github.com/KratosMultiphysics/Kratos/commit/b6ca184fc1f2b2afae4b6cd5ab56f73f65edaa4a)

### DIFF
--- a/kratos/utilities/geometry_utilities.cpp
+++ b/kratos/utilities/geometry_utilities.cpp
@@ -5,7 +5,7 @@
 //                   Multi-Physics
 //
 //  License:         BSD License
-//                     Kratos default license: kratos/license.txt
+//                   Kratos default license: kratos/license.txt
 //
 //  Main authors:    Riccardo Rossi
 //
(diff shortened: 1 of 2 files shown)


philbucher (dismissed):
:+1:

loumalouomega:
I made a mistake, I need reapproval

roigcarlo (approved):

philbucher (approved):

kratos/utilities/geometry_utilities.cpp
```diff
@@ -16,13 +16,112 @@
...
 
 namespace Kratos
 {
+
+std::string GetGeometryName(const GeometryData::KratosGeometryType TypeOfGeometry)
+{
+    KRATOS_TRY;
+
+    // Using switch over map as the compiler warns if some enum values are not handled in the switch
+    switch(TypeOfGeometry) {
```
huhaas, line 30: Would it be possible to reuse the `string_geometry_map` in `specifications_utilities.h`? Iterate over the values of the map and return the key. 
Fewer sources of the same list are easier to maintain.

loumalouomega: a switch statement is much more efficient than a map

philbucher: I would refactor the specification utils after this PR

huhaas: Well yes. But I assume this function is not part of the program’s critical path due to its string-based logic. I’m curious about the speed vs maintainability tradeoff in this case. Or am I wrong in my assumptions?

I do not intend to re-open any PR, but merely asking questions and trying to understand.

philbucher: That is a good question, I cannot answer that without proper profiling

This is usually where @matekelemen steps in to amaze us with his micro-benchmarking skills 😄

matekelemen: I'm with @huhaas on this one; correctness and maintainability is much more important here than performance. By the way, the more pressing question that pops into my mind is why getting the name of an object is not part of the object's (`Geometry`) or  the enum's (`KratosGeometryType`) features, but that of some obscure utility in a completely different place?

It doesn't make much sense, but I'd hate to disappoint @philbucher, so here goes:
The fact that we're using `std::string` instead of `std::string_view` for string literals way overshadows any performance benefit we may get from a `switch` statement instead of a linear search on a map:
- [switch <=> map comparison with std::string](https://quick-bench.com/q/xnVgNyvoUSJgCv3W53K6DD1seDc)
- [switch <=> map comparison with std::string_view](https://quick-bench.com/q/U2Warypfnsb62Zkjo7OWRI_otio)

huhaas: @matekelemen that website is very fun to use 😄 

Appended your example with two cases: [link](https://quick-bench.com/q/KESr40mqwSejkCi1QpG_XSsrO7o)
(the regular map can be either a bi-map or two maps (A->B and B->A) constructed from a single source)).
The problem size is too small for any specialized container. With N=20 the overhead is considerable.

Having the name be part of the object (OO style) will be a much bigger code rewrite. Then the enum construction will change and every usage of it, which are a lot.

PR #10751, opened 2023-02-15
[Core] Use all constraints to construct structure in BlockBuilderAndSolver

ddiezrod:
**📝 Description**

I am running a 3d print case where I need to start with all constraints turned off and activate them sequentially. I am getting an error because the matrix structure for the constraints is being created in the first time step, only with the active ones. I dont know all the details of this implementation, but I understand that the structure should be created with all the constraints. Later, when we build the actual matrix, we will add only the active ones.

**🆕 Changelog**

- [Use all constraints to construct structure](https://github.com/KratosMultiphysics/Kratos/pull/10751/commits/b4e89fd80660c8459708b9293849e765e476514b)
- [Remove unused](https://github.com/KratosMultiphysics/Kratos/pull/10751/commits/c2570d3fba459d2dde41a5720ff8f80740d2e403)
- [Adding missing access methods](https://github.com/KratosMultiphysics/Kratos/pull/10751/commits/d06231ac1cca7786f2316c86a217b41aeef13bc9)
- [Expose mT and mConstantVector](https://github.com/KratosMultiphysics/Kratos/pull/10751/commits/4e02a91ab9f946d74ff06392ed9b61c930279db9)
- [Adding BasicDisplacementBlockBuilderAndSolverWithInactiveConstraints](https://github.com/KratosMultiphysics/Kratos/pull/10751/commits/b1cc7b65c1bfe3ba5126570685d69fccaf09b26f)

### DIFF
--- a/kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver.h
+++ b/kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver.h
@@ -1144,6 +1144,24 @@ public:
     ///@name Access
     ///@{
 
+    /**
+     * @brief This method returns constraint relation (T) matrix
+     * @return The constraint relation (T) matrix
+     */
+    TSystemMatrixType& GetConstraintRelationMatrix()
+    {
+        return mT;
+    }
+
+    /**
+     * @brief This method returns constraint constant vector
+     * @return The constraint constant vector
+     */
+    TSystemVectorType& GetConstraintConstantVector()
+    {
+        return mConstantVector;
+    }
+
     ///@}
     ///@name Inquiry
     ///@{
--- a/kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver.h
+++ b/kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver.h
@@ -255,7 +255,6 @@ public:
 
         KRATOS_INFO_IF("ResidualBasedBlockBuilderAndSolver", this->GetEchoLevel() >= 1) << "Build time: " << timer.ElapsedSeconds() << std::endl;
 
-
         KRATOS_INFO_IF("ResidualBasedBlockBuilderAndSolver", (this->GetEchoLevel() > 2 && rModelPart.GetCommunicator().MyPID() == 0)) << "Finished parallel building" << std::endl;
 
         KRATOS_CATCH("")
@@ -316,7 +315,6 @@ public:
                     // Assemble the elemental contribution
                     AssembleLHS(rA, lhs_contribution, equation_id);
                 }
-
             }
 
             #pragma omp for  schedule(guided, 512)
@@ -328,8 +326,7 @@ public:
                 if (it_cond->IsDefined(ACTIVE))
                     condition_is_active = it_cond->Is(ACTIVE);
 
-                if (condition_is_active)
-                {
+                if (condition_is_active) {
                     // Calculate elemental contribution
                     pScheme->CalculateLHSContribution(*it_cond, lhs_contribution, equation_id, r_current_process_info);
 
@@ -547,7 +544,7 @@ public:
         ) override
     {
         Timer::Start("Linearizing on Old iteration");
-             
+
         KRATOS_INFO_IF("BlockBuilderAndSolver", this->GetEchoLevel() > 0) << "Linearizing on Old iteration" << std::endl;
 
         KRATOS_ERROR_IF(rModelPart.GetBufferSize() == 1) << "BlockBuilderAndSolver: \n"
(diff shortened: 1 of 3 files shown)


loumalouomega:
In any case we should check that the inactive ones have been activated, and then tell the B&S to update the connectivity

loumalouomega:
We should discuss this

loumalouomega (changes requested):
We need to discuss

ddiezrod:
> In any case we should check that the inactive ones have been activated, and then tell the B&S to update the connectivity

I dont understand what you mean. This function only prepares the indices for mT and mConstantVector. If you use here only active constraints, you will get a segmentation fault later when you activate another constraint because the index is not there. I think it makes sense to use all constraints to prepare the structure and then use only the ones that are needed in the assembly.

loumalouomega:
> > In any case we should check that the inactive ones have been activated, and then tell the B&S to update the connectivity
> 
> I dont understand what you mean. This function only prepares the indices for mT and mConstantVector. If you use here only active constraints, you will get a segmentation fault later when you activate another constraint because the index is not there. I think it makes sense to use all constraints to prepare the structure and then use only the ones that are needed in the assembly.

Do we check it for the elements/conditions?, let me see

loumalouomega (dismissed):
Okay, in the conditions/elements is not checked

rubenzorrilla:
> > In any case we should check that the inactive ones have been activated, and then tell the B&S to update the connectivity
> 
> I dont understand what you mean. This function only prepares the indices for mT and mConstantVector. If you use here only active constraints, you will get a segmentation fault later when you activate another constraint because the index is not there. I think it makes sense to use all constraints to prepare the structure and then use only the ones that are needed in the assembly.

I think this makes sense. However, I'd wait for @RiccardoRossi's opinion just in case.

loumalouomega (dismissed):

kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver.h
```diff
@@ -1308,21 +1305,11 @@ class ResidualBasedBlockBuilderAndSolver
                 #pragma omp for schedule(guided, 512) nowait
                 for (int i_const = 0; i_const < static_cast<int>(rModelPart.MasterSlaveConstraints().size()); ++i_const) {
                     auto it_const = it_const_begin + i_const;
+                    it_const->EquationIdVector(slave_ids, master_ids, r_current_process_info);
```
RiccardoRossi, line 1311: why are u removing this???

RiccardoRossi (dismissed):

RiccardoRossi (commented):
well i took a look at this, and i am really not clear about this change.

the point is that the current implementation is very delicate, since we are attempting to reuse the graph of the matrix.

We shall instead first build a matrix K, and then build a matrix TtKt (with the corresponding residual). Much of our problems come from that.

having said this, and regarding the specific change, what happens to the corresponding line in matrix T in the case in which the corresponding constraint is elided? does it look like an identity?

kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver.h
ddiezrod: green means adding : P , what i am actually removing is the `if(constraint_is_active)` as I am adding the indices from all constraints, not only active ones.

ddiezrod:
> having said this, and regarding the specific change, what happens to the corresponding line in matrix T in the case in which the corresponding constraint is elided? does it look like an identity?

@RiccardoRossi Regarding this comment, I will take a deeper look to check how everything works exactly, but think it this way: before the proposed change in this MR, what happens if we begin a simulation with all constraints ACTIVE and then we turn off some of them after a number of time steps? We are in the same situation, its contribution is not added to mT and mConstantVector so it is not applied, but everything works smoothly.

The key is that when we create the mT and mConstantVector matrices, we need to add all the indices of all constraints that will be used at some point in the calculation, otherwise we will get a segmentation fault when we activate other constraints.

loumalouomega:
@RiccardoRossi I was the one who originally implemented the T structure build, what @ddiezrod  is right, even if inactive the structure must be built as it is done in elements and conditions, otherwise will crash if reactivated. Do not worry about zero values, because the current code already checks that, I added a test to show that when inactive constraints T is an identity matrix (I needed to add Access methods for T and C, sorry)

loumalouomega (dismissed):

ddiezrod:
Thanks a lot @loumalouomega ! A test is always more convincing than words :)

loumalouomega:
> Thanks a lot @loumalouomega ! A test is always more convincing than words :)

Are you convinced by this?

~~~c++
KRATOS_CHECK(true);
~~~

RiccardoRossi (approved):
ok, it looks reasonable to me too

ddiezrod:
Thanks @RiccardoRossi !

PR #10453, opened 2022-11-15
[Core] Distance to path process

loumalouomega:
**📝 Description**

This adds a process able to compute the distance to a modelpart constituted of lines:

image

Depends on #10413

MPI not working for now, but I have higher order priorities right now, and I will fix it later.

**🆕 Changelog**

- [Initial commit](https://github.com/KratosMultiphysics/Kratos/commit/cffa08615cfe5394510fbf4e973bc7f7a7855c21)
- [Expose to python](https://github.com/KratosMultiphysics/Kratos/commit/95f54f540a2d075effd1fb5099a5166ad92356cd)
- [Test](https://github.com/KratosMultiphysics/Kratos/commit/9e4c0a9df43b0f08a58890181e794076903bcda6)

PR #10753, opened 2023-02-15
[Core] Comments from @rubenzorrilla from #10453

loumalouomega:
**📝 Description**

Comments from @rubenzorrilla from #10453.

**🆕 Changelog**

- [Moving enum](https://github.com/KratosMultiphysics/Kratos/commit/3945750570675fff76cdf3fc129f0b04c0e8b159)
- [Remove comment](https://github.com/KratosMultiphysics/Kratos/commit/8e501bb4440ccd6032ba7e29bb0680f2b16f381b)


rubenzorrilla (commented):
There is still the issue with the `public static` method. If it can be `static` in my opinion it shoulb be moved to a separate utility to keep the standard process API.

loumalouomega:
> There is still the issue with the `public static` method. If it can be `static` in my opinion it shoulb be moved to a separate utility to keep the standard process API.

The thing is that in order to check the function in a cpp test it is easier if the method is static, otherwise I need to create a dummy process just to check it.

rubenzorrilla:
> > There is still the issue with the `public static` method. If it can be `static` in my opinion it shoulb be moved to a separate utility to keep the standard process API.
> 
> The thing is that in order to check the function in a cpp test it is easier if the method is static, otherwise I need to create a dummy process just to check it.

Honestly, I think this is not enough justification to extend the API...

loumalouomega:
> > > There is still the issue with the `public static` method. If it can be `static` in my opinion it shoulb be moved to a separate utility to keep the standard process API.
> > 
> > 
> > The thing is that in order to check the function in a cpp test it is easier if the method is static, otherwise I need to create a dummy process just to check it.
> 
> Honestly, I think this is not enough justification to extend the API...

Okay

rubenzorrilla:
> > > > There is still the issue with the `public static` method. If it can be `static` in my opinion it shoulb be moved to a separate utility to keep the standard process API.
> > > 
> > > 
> > > The thing is that in order to check the function in a cpp test it is easier if the method is static, otherwise I need to create a dummy process just to check it.
> > 
> > 
> > Honestly, I think this is not enough justification to extend the API...
> 
> Okay

It is still in there... I'm not concerned about the `static` I'm concerned about the fact that you're adding a method to the public API. I think that processes should stick to the one in `process.h` (it is admittedly true that we didn't follow this in the past).

Considering that you initially made it `static` it could be technically placed somewhere else. If you aim to use it only for the testing I'd put it in a unnamed namespace in the corresponding cpp test. Otherwise, we can put it in any auxiliary utility.

loumalouomega:
> > > > > There is still the issue with the `public static` method. If it can be `static` in my opinion it shoulb be moved to a separate utility to keep the standard process API.
> > > > 
> > > > 
> > > > The thing is that in order to check the function in a cpp test it is easier if the method is static, otherwise I need to create a dummy process just to check it.
> > > 
> > > 
> > > Honestly, I think this is not enough justification to extend the API...
> > 
> > 
> > Okay
> 
> It is still in there... I'm not concerned about the `static` I'm concerned about the fact that you're adding a method to the public API. I think that processes should stick to the one in `process.h` (it is admittedly true that we didn't follow this in the past).
> 
> Considering that you initially made it `static` it could be technically placed somewhere else. If you aim to use it only for the testing I'd put it in a unnamed namespace in the corresponding cpp test. Otherwise, we can put it in any auxiliary utility.

Yes, but the return type is an enum that you ask me to take out

rubenzorrilla:
> > > > > > There is still the issue with the `public static` method. If it can be `static` in my opinion it shoulb be moved to a separate utility to keep the standard process API.
> > > > > 
> > > > > 
> > > > > The thing is that in order to check the function in a cpp test it is easier if the method is static, otherwise I need to create a dummy process just to check it.
> > > > 
> > > > 
> > > > Honestly, I think this is not enough justification to extend the API...
> > > 
> > > 
> > > Okay
> > 
> > 
> > It is still in there... I'm not concerned about the `static` I'm concerned about the fact that you're adding a method to the public API. I think that processes should stick to the one in `process.h` (it is admittedly true that we didn't follow this in the past).
> > Considering that you initially made it `static` it could be technically placed somewhere else. If you aim to use it only for the testing I'd put it in a unnamed namespace in the corresponding cpp test. Otherwise, we can put it in any auxiliary utility.
> 
> Yes, but the return type is an enum that you ask me to take out

But can't you access to the enum if it is within the class namespace? (maybe I'm missing something)

loumalouomega:
> > > > > > > There is still the issue with the `public static` method. If it can be `static` in my opinion it shoulb be moved to a separate utility to keep the standard process API.
> > > > > > 
> > > > > > 
> > > > > > The thing is that in order to check the function in a cpp test it is easier if the method is static, otherwise I need to create a dummy process just to check it.
> > > > > 
> > > > > 
> > > > > Honestly, I think this is not enough justification to extend the API...
> > > > 
> > > > 
> > > > Okay
> > > 
> > > 
> > > It is still in there... I'm not concerned about the `static` I'm concerned about the fact that you're adding a method to the public API. I think that processes should stick to the one in `process.h` (it is admittedly true that we didn't follow this in the past).
> > > Considering that you initially made it `static` it could be technically placed somewhere else. If you aim to use it only for the testing I'd put it in a unnamed namespace in the corresponding cpp test. Otherwise, we can put it in any auxiliary utility.
> > 
> > 
> > Yes, but the return type is an enum that you ask me to take out
> 
> But can't you access to the enum if it is within the class namespace? (maybe I'm missing something)

I hope last refactor makes you happy

rubenzorrilla:
> > > > > > > > There is still the issue with the `public static` method. If it can be `static` in my opinion it shoulb be moved to a separate utility to keep the standard process API.
> > > > > > > 
> > > > > > > 
> > > > > > > The thing is that in order to check the function in a cpp test it is easier if the method is static, otherwise I need to create a dummy process just to check it.
> > > > > > 
> > > > > > 
> > > > > > Honestly, I think this is not enough justification to extend the API...
> > > > > 
> > > > > 
> > > > > Okay
> > > > 
> > > > 
> > > > It is still in there... I'm not concerned about the `static` I'm concerned about the fact that you're adding a method to the public API. I think that processes should stick to the one in `process.h` (it is admittedly true that we didn't follow this in the past).
> > > > Considering that you initially made it `static` it could be technically placed somewhere else. If you aim to use it only for the testing I'd put it in a unnamed namespace in the corresponding cpp test. Otherwise, we can put it in any auxiliary utility.
> > > 
> > > 
> > > Yes, but the return type is an enum that you ask me to take out
> > 
> > 
> > But can't you access to the enum if it is within the class namespace? (maybe I'm missing something)
> 
> I hope last refactor makes you happy

This is exactly what I meant since the beginning.

Also note that the point in here isn't what makes me happy but that we cannot freely extend the API of this sort of classes, specially if they belong to the `KratosCore`. This is in my opinion a bad practice that we shouldn't promote in new developments.

rubenzorrilla (dismissed):
Approving. Thanks for taking into account my considerations.

loumalouomega:
@roigcarlo windows compilation fails and I don't know why

roigcarlo:
Looking into it

roigcarlo:
Still testing, but most likely the error is because of MS headers redefining the "ERROR" keyword in the return types of the geometrical utilities when included by the CoSimApplication when compiled using unitary builds (some windows.h will probably gets included from behind).

I would just simply change that for something else.

rubenzorrilla (approved):

PR #10754, opened 2023-02-15
[Core] Adding KratosGlobals test

loumalouomega:
**📝 Description**

In here not adding anything new to `KratosGlobals`, just adding tests which were missing and some clean up

**🆕 Changelog**

- [Initial clean up](https://github.com/KratosMultiphysics/Kratos/commit/b968801108078fd9e8f88b79664bb0cf33d09d15)
- [Adding HasFlag](https://github.com/KratosMultiphysics/Kratos/commit/9f19f7fc546e0b84da1c1577c8fdee3020e32b81)
- [Adding test for KratosGlobals](https://github.com/KratosMultiphysics/Kratos/commit/5cebfbd5842a2242d07b2a430067cba546d5fb9a)
- [Improve test](https://github.com/KratosMultiphysics/Kratos/commit/069cbf5288c10c7c95edf21fed6e88ae900b4401)

### DIFF
--- a/kratos/tests/test_kratos_globals.py
+++ b/kratos/tests/test_kratos_globals.py
@@ -1,0 +1,38 @@
+import KratosMultiphysics
+import KratosMultiphysics.KratosUnittest as KratosUnittest
+
+class TestKratosGlobals(KratosUnittest.TestCase):
+
+    def test_has_variable(self):
+        var_name = "VELOCITY"
+        self.assertTrue(KratosMultiphysics.KratosGlobals.HasVariable(var_name))
+        var_name = "VELOCITY2"
+        self.assertFalse(KratosMultiphysics.KratosGlobals.HasVariable(var_name))
+
+    def test_get_variable_type(self):
+        var_name = "TEMPERATURE"
+        self.assertEqual(KratosMultiphysics.KratosGlobals.GetVariableType(var_name), "Double")
+        var_name = "VELOCITY"
+        self.assertEqual(KratosMultiphysics.KratosGlobals.GetVariableType(var_name), "Array")
+
+    def test_get_variable(self):
+        var_name = "TEMPERATURE"
+        self.assertEqual(KratosMultiphysics.KratosGlobals.GetVariable(var_name), KratosMultiphysics.TEMPERATURE)
+        var_name = "VELOCITY"
+        self.assertEqual(KratosMultiphysics.KratosGlobals.GetVariable(var_name), KratosMultiphysics.VELOCITY)
+        var_name = "VELOCITY2"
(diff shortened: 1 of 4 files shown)


kratos/tests/test_kratos_globals.py
```diff
@@ -0,0 +1,49 @@
...
+    def test_has_flag(self):
+        flag_name = "STRUCTURE"
+        self.assertTrue(KratosMultiphysics.KratosGlobals.HasFlag(flag_name))
+        flag_name = "STRUCTURE2"
+        self.assertFalse(KratosMultiphysics.KratosGlobals.HasFlag(flag_name))
+
+    def test_get_flag(self):
+        flag_name = "STRUCTURE"
+        KratosMultiphysics.KratosGlobals.GetFlag(flag_name)
+        # self.assertEqual(KratosMultiphysics.KratosGlobals.GetFlag(flag_name), KratosMultiphysics.STRUCTURE) # TODO: This should work, they are different objects
```
rubenzorrilla, line 35: Isn't this working?

rubenzorrilla (commented):
Fine with the changes. Only two details to be clarified.

kratos/tests/test_kratos_globals.py
```diff
@@ -0,0 +1,49 @@
...
+    def test_has_constitutive_law(self):
+        law_name = "ConstitutiveLaw"
+        self.assertTrue(KratosMultiphysics.KratosGlobals.HasConstitutiveLaw(law_name))
+        law_name = "ConstitutiveLaw2"
+        self.assertFalse(KratosMultiphysics.KratosGlobals.HasConstitutiveLaw(law_name))
+
+    def test_get_constitutive_law(self):
+        law_name = "ConstitutiveLaw"
+        KratosMultiphysics.KratosGlobals.GetConstitutiveLaw(law_name)
+
```
rubenzorrilla, line 46: Not 100% sure if we should check something in here as the `GetConstitutiveLaw` would throw an error if not found. Just pointing it out since I think that as it is right now would be OK too.

kratos/tests/test_kratos_globals.py
loumalouomega: Nope, because flags are not like variables, which are created statically and exposed statically

kratos/tests/test_kratos_globals.py
loumalouomega: Okay

kratos/tests/test_kratos_globals.py
rubenzorrilla: OK, in any case the `GetFlag` would throw an error if not found so I don't think there is any problem.

rubenzorrilla (approved):

PR #10755, opened 2023-02-16
[PfemFluid] updating of Herschel Bulkley law

AFranci:
This PR updated the HB law making it compatible with the GiD interface and with a structure analogous to the other constitutive laws of PFEMfluid.


AlejandroCornejo (approved):
simple PR

AlejandroCornejo (approved):
ok

Issue #10757, opened 2023-02-17
WINDOWS COMPILATION ERROR


AlejandroCornejo:
**Description**
Today @atfard and me noticed that the master branch is not compiling in Windows (StructuralMEchApp):

C:\Kratos\applications\StructuralMechanicsApplication\custom_conditions\small_displacement_surface_load_condition_3d.cpp(149,20): error C2065: 'PRESSURE': identificador no declarado [C:\Kr
atos\build\Release\applications\StructuralMechanicsApplication\KratosStructuralMechanicsCore.vcxproj]
C:\Kratos\applications\StructuralMechanicsApplication\custom_conditions\small_displacement_surface_load_condition_3d.cpp(150,50): error C2065: 'PRESSURE': identificador no declarado [C:\Kr
atos\build\Release\applications\StructuralMechanicsApplication\KratosStructuralMechanicsCore.vcxproj]
C:\Kratos\applications\StructuralMechanicsApplication\custom_conditions\small_displacement_surface_load_condition_3d.cpp(152,20): error C2065: 'NEGATIVE_FACE_PRESSURE': identificador no de
clarado [C:\Kratos\build\Release\applications\StructuralMechanicsApplication\KratosStructuralMechanicsCore.vcxproj]
C:\Kratos\applications\StructuralMechanicsApplication\custom_conditions\small_displacement_surface_load_condition_3d.cpp(153,50): error C2065: 'NEGATIVE_FACE_PRESSURE': identificador no de
clarado [C:\Kratos\build\Release\applications\StructuralMechanicsApplication\KratosStructuralMechanicsCore.vcxproj]
C:\Kratos\applications\StructuralMechanicsApplication\custom_conditions\small_displacement_surface_load_condition_3d.cpp(155,20): error C2065: 'POSITIVE_FACE_PRESSURE': identificador no de
clarado [C:\Kratos\build\Release\applications\StructuralMechanicsApplication\KratosStructuralMechanicsCore.vcxproj]
C:\Kratos\applications\StructuralMechanicsApplication\custom_conditions\small_displacement_surface_load_condition_3d.cpp(156,50): error C2065: 'POSITIVE_FACE_PRESSURE': identificador no de
clarado [C:\Kratos\build\Release\applications\StructuralMechanicsApplication\KratosStructuralMechanicsCore.vcxproj]
[...]

PR #10758, opened 2023-02-17
[StructuralMechApp] Adding include to make it to compile

AlejandroCornejo:
That's it

### DIFF
--- a/applications/StructuralMechanicsApplication/custom_conditions/base_load_condition.h
+++ b/applications/StructuralMechanicsApplication/custom_conditions/base_load_condition.h
@@ -19,7 +19,6 @@
 
 // Project includes
 #include "includes/condition.h"
-#include "includes/variables.h"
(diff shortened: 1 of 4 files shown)


AlejandroCornejo:
solving https://github.com/KratosMultiphysics/Kratos/issues/10757

loumalouomega (dismissed):

loumalouomega:
It would be less "time compilation consuming" if the include is added in the cpp that require it instead in the base class header

AlejandroCornejo:
> It would be less "time compilation consuming" if the include is added in the cpp that require it instead in the base class header

ok, I'll change it in a few hours

loumalouomega (approved):

atfard (approved):

ipouplana:
Thank you @AlejandroCornejo !

PR #10760, opened 2023-02-17
[Core] Add flag to skip inbuilt TRY/CATCH

roigcarlo:
**📝 Description**
Adds a flag to prevent the TRY CATCH blocks of Kratos from generating code.

I've been trying to push different flavors of this for some time and this is IMHO the cleanest way to accomplish this.

The options is disabled in all release flavors and the idea is that it does not need to be activated unless you need to find very specific things.

Essentially this allows someone in GDB to easily generate a correct backtrace if a error is produced inside a `try/catch` block (otherwise a backtrace of the system call handling the exception is produced, which is mostly useless).
While it is possible to do the same with the current code you need to insert a breakpoint in `__cxa_throw` which is not obvious (specially when the exceptions are being propagated) and can complicate things if you want to add other breakpoints or you are not sure where the code breaks exactly.

**🆕 Changelog**
- Adding flag to prevent try/catch

### DIFF
--- a/kratos/includes/define.h
+++ b/kratos/includes/define.h
@@ -95,15 +95,20 @@ catch(...) { Block KRATOS_THROW_ERROR(std::runtime_error, "Unknown error", MoreI
 #define KRATOS_CATCH_BLOCK_BEGIN class ExceptionBlock{public: void operator()(void){
 #define KRATOS_CATCH_BLOCK_END }} exception_block; exception_block();
 
-#ifndef __SUNPRO_CC
-#define KRATOS_TRY try {
-
-#define KRATOS_CATCH(MoreInfo) \
-  KRATOS_CATCH_WITH_BLOCK(MoreInfo,{})
+#ifndef KRATOS_NO_TRY_CATCH
+    #define KRATOS_TRY_IMPL try {
+    #define KRATOS_CATCH_IMPL(MoreInfo) KRATOS_CATCH_WITH_BLOCK(MoreInfo,{})
 #else
-#define KRATOS_TRY { };
+    #define KRATOS_TRY_IMPL {};
+    #define KRATOS_CATCH_IMPL(MoreInfo) {};
+#endif
 
-#define KRATOS_CATCH(MoreInfo) { };
+#ifndef __SUNPRO_CC
+    #define KRATOS_TRY KRATOS_TRY_IMPL
+    #define KRATOS_CATCH(MoreInfo) KRATOS_CATCH_IMPL(MoreInfo)
+#else
+    #define KRATOS_TRY {};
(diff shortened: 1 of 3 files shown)


kratos/includes/define.h
```diff
@@ -95,15 +95,20 @@ catch(...) { Block KRATOS_THROW_ERROR(std::runtime_error, "Unknown error", MoreI
...
+    #define KRATOS_TRY_IMPL try {
+    #define KRATOS_CATCH_IMPL(MoreInfo) KRATOS_CATCH_WITH_BLOCK(MoreInfo,{})
 #else
-#define KRATOS_TRY { };
+    #define KRATOS_TRY_IMPL {};
+    #define KRATOS_CATCH_IMPL(MoreInfo) {};
+#endif
 
-#define KRATOS_CATCH(MoreInfo) { };
+#ifndef __SUNPRO_CC
```
philbucher, line 106: how about using this opportunity to get rid of this one?

philbucher (commented):
LGTM

how about adding an `option` to CMake for this?

kratos/includes/define.h
roigcarlo: Wanted to, but I am not sure what it does, so I followed the safe approach of "If it works don't touch it".

I will dig more

roigcarlo:
> how about adding an option to CMake for this?

How that would be? Another build config? I mean you can just `-DKRATOS_NO_TRY_CATCH`

matekelemen:
> > how about adding an option to CMake for this?

>How that would be? Another build config? I mean you can just -DKRATOS_NO_TRY_CATCH

Doesn't that only define a CMake variable but not the preprocessor macro? Anyway, I'd collect these kinds of definitions in one place in the root CMakeLists so that adventurous developers know what's available to them.

philbucher:
> Anyway, I'd collect these kinds of definitions in one place in the root CMakeLists so that adventurous developers know what's available to them.

thats exatly what I mean

> How that would be? Another build config? I mean you can just -DKRATOS_NO_TRY_CATCH

yes ofc. But if this is hidden in the code somewhere then it is hard for others to find
=> Can you please add this also to the compiler flags in the `Install.md`? Preferably with a hint of how  it should/can be used => for GDB

roigcarlo:
Ok, I will add it as an option :+1:

roigcarlo:
Ping

matekelemen (commented):
lgtm ®
thanks

matekelemen (approved):

AlejandroCornejo:
Is there any performace gain when running?

roigcarlo:
Nope. Just for debugging.

PR #10761, opened 2023-02-17
[TrilinosApplication] Extend SetValue in TrilinosSpace

loumalouomega:
**📝 Description**

Extend `SetValue` in `TrilinosSpace`. Adding local values set option + matrix case + C++ test

**🆕 Changelog**

- [Extend SetValue in TrilinosSpace](https://github.com/KratosMultiphysics/Kratos/commit/b3a36a30d98f4b0fc1247d7c86c7ecd66617eeb6)
- [Adding tests for Setvalue in TrilinosSpace](https://github.com/KratosMultiphysics/Kratos/commit/00e32aa8059a8d463791d0a445b8bad259329b27)


philbucher:
Overall 👍, but are the templated versions necessary? To me they are cluttering without real benefit

loumalouomega:
It is to avoid adding one more argument (I don't like functions with 100
arguments), and with the constexpr you avoid a check (if this method is
used recursively it can be significant)

philbucher:
I wouldnt be to worried about a bool comparison when you call `GlobalAssemble`

In any case, if you use this multiple times then I would say the algorithm is wrong / has a very bad performance.
Thinking about it some more I would be more inclined to remove this method altogether because of this

What is your usecase?

loumalouomega:
> I wouldnt be to worried about a bool comparison when you call `GlobalAssemble`
> 
> In any case, if you use this multiple times then I would say the algorithm is wrong / has a very bad performance. Thinking about it some more I would be more inclined to remove this method altogether because of this

Huuum, see my next answer

> What is your usecase?

Assembling the diagonal terms of the T matrix for MPC. I would do the GlobalAssemble at the end, but for the sake of consistency with the previous implementation I must keep by default the GlobalAssemble.

philbucher:
which previous implementation are you referring to?

In any case, if you want to update the diagonal, you should do it once for all values, and not one-by-one

loumalouomega:
> which previous implementation are you referring to?

There is already a SetValue implementation for Epetra Vector, just not tested

> In any case, if you want to update the diagonal, you should do it once for all values, and not one-by-one

I need to iterate over rows anyway, so it will not provide relevant performance increment in respect my proposal

philbucher:
> > which previous implementation are you referring to?
> 
> There is already a SetValue implementation for Epetra Vector, just not tested

This one seems unused, I would rather deprecate remove it given its performance issues

> > In any case, if you want to update the diagonal, you should do it once for all values, and not one-by-one
> 
> I need to iterate over rows anyway, so it will not provide relevant performance increment in respect my proposal

Hm not sure I follow. Hard to understand without seeing code

loumalouomega:
> > > which previous implementation are you referring to?
> > 
> > 
> > There is already a SetValue implementation for Epetra Vector, just not tested
> 
> This one seems unused, I would rather deprecate remove it given its performance issues

In any case I would change the default to not Assemble, those methods can be useful for simple implementations. Trilinos es not the most user friendly API in the world

> > > In any case, if you want to update the diagonal, you should do it once for all values, and not one-by-one
> > 
> > 
> > I need to iterate over rows anyway, so it will not provide relevant performance increment in respect my proposal
> 
> Hm not sure I follow. Hard to understand without seeing code

If you do it manually it would be essentially the same, but Okay, I can simplify it

philbucher:
> In any case I would change the default to not Assemble, those methods can be useful for simple implementations. Trilinos is not the most user friendly API in the world

that the thing, we are discussing about distributed linear algebra. This is IMO one of the most challenging topics in FEM. I.e. providing "simple" interfaces that have serious limitations can be dangerous.

Maybe you can instead open the PR with the code where you would actually use this? I have the feeling there are some gains to be made :)

loumalouomega:
> > In any case I would change the default to not Assemble, those methods can be useful for simple implementations. Trilinos is not the most user friendly API in the world
> 
> that the thing, we are discussing about distributed linear algebra. This is IMO one of the most challenging topics in FEM. I.e. providing "simple" interfaces that have serious limitations can be dangerous.
> 
> Maybe you can instead open the PR with the code where you would actually use this? I have the feeling there are some gains to be made :)

Well, I could, but not right now, once is finished. Meanwhile, I will be cherry-picking intermediate changes here

loumalouomega:
@philbucher

jcotela:
The new functions by themselves look fine to me, but I'm not convinced that TrilinosSpace is the right place for them... The *Space objects should isolate their caller from the details of the linear algebra interface (which may or may not be distributed). If the caller of the space needs to know about local and global indices, we are not really isolating them. SetValue for example was designed to work with global indices by design, because the non-MPI version of the same function has no concept of local indices

loumalouomega:
> The new functions by themselves look fine to me, but I'm not convinced that TrilinosSpace is the right place for them... The *Space objects should isolate their caller from the details of the linear algebra interface (which may or may not be distributed). If the caller of the space needs to know about local and global indices, we are not really isolating them. SetValue for example was designed to work with global indices by design, because the non-MPI version of the same function has no concept of local indices

Huuuum, in relation to `SetValue` I am extending what is already existing, and by design current implementation is local, but I can change it for global, but that would be a behaviour change.

loumalouomega:
I am going to cherry-pick some smaller changes to simplify this PR

philbucher:
fully agree with @jcotela (welcome back :tada: )

I suggest to do everything in the Trilinos B&S directly, and once it is working and it is necessary, then it can be refactored to the Space, if necessary

jcotela:
> > The new functions by themselves look fine to me, but I'm not convinced that TrilinosSpace is the right place for them... The *Space objects should isolate their caller from the details of the linear algebra interface (which may or may not be distributed). If the caller of the space needs to know about local and global indices, we are not really isolating them. SetValue for example was designed to work with global indices by design, because the non-MPI version of the same function has no concept of local indices
> 
> Huuuum, in relation to `SetValue` I am extending what is already existing, and by design current implementation is local, but I can change it for global, but that would be a behaviour change.

Then I probably misunderstood what you mean by local index... If I have a 10-row matrix distributed between two processes, I expect that the local row indices on each process are 0-4, while the global ones might be 0-4 on one process and 5-9 on the other. I would expect the input to the space functions to be in terms of the global 0-9 indices.

Following the example, in the proposed implementation of SetValue, I'd think a global (0-9) index is required if TGlobalIndex is true but a local (0-4) one is required when it is false. Am I misunderstanding the intent?

loumalouomega:
> > > The new functions by themselves look fine to me, but I'm not convinced that TrilinosSpace is the right place for them... The *Space objects should isolate their caller from the details of the linear algebra interface (which may or may not be distributed). If the caller of the space needs to know about local and global indices, we are not really isolating them. SetValue for example was designed to work with global indices by design, because the non-MPI version of the same function has no concept of local indices
> > 
> > 
> > Huuuum, in relation to `SetValue` I am extending what is already existing, and by design current implementation is local, but I can change it for global, but that would be a behaviour change.
> 
> Then I probably misunderstood what you mean by local index... If I have a 10-row matrix distributed between two processes, I expect that the local row indices on each process are 0-4, while the global ones might be 0-4 on one process and 5-9 on the other. I would expect the input to the space functions to be in terms of the global 0-9 indices.
> 
> Following the example, in the proposed implementation of SetValue, I'd think a global (0-9) index is required if TGlobalIndex is true but a local (0-4) one is required when it is false. Am I misunderstanding the intent?

Yes, that is what I mean

loumalouomega:
Ping @philbucher @jcotela

loumalouomega:
> Ping @philbucher @jcotela

Reping

loumalouomega:
@RiccardoRossi, can we discuss this with @jcotela  and @pooyan-dadvand this week.

loumalouomega:
Closing this one after discussing with @RiccardoRossi and @jcotela

loumalouomega:
(I will delete branch after i refactor the code of the other PR)

PR #10764, opened 2023-02-17
[OptApp] Restructure 1 - Adding Optimization Info

sunethwarna:
**📝 Description**
This PR adds `OptimizationInfo` class which is to be used inthe `OptimizationApplication`. It has following main functions.

1. It has a a buffer of type `dict`, which can be used to store data of any type in current, previous, etc depending on the specified minimum buffer size.
2. These data are used to write to optimization log files and can be transfered Vtk/HDF/Statis or any other `Kratos` output/input mechanism.
3. This also acts as a holder for objects of `Kratos::Process` which are constructed with `Kratos::Model`, `Kratos::Parameters`, and `OptimizationInfo`.
4. This is used to have inter-communication between objects in `OptimizationApplication`.

**🆕 Changelog**
- Adds `OptimizationInfo`
- Minor modifications to `ExecutionPolicy`


RezaNajian:
Hi Suneth,
Thanks for the improvements and extensions. In the following, I give my main/general comments.

1. As described, the main motivation/contribution of the PR is the OptimizationInfo class. However, a lot of not necessary dependencies and complexities are introduced. For example, OptimizationInfo is only used in the OptimizationProcessFactory and nowhere else. OptimizationInfo should ONLY belong to the optimization stuff and not it's building blocks like anaylsiswrappers. Please note that in the OptimiaztionApplication master-slave approach is followed meaning that the optimizer is the master and ["model_parts","analyses","responses","controls"] are slaves. Therefore, I would like to ask you to remove OptimizationInfo from the OptimizationProcessFactory.
2. Although this PR is about OptimiaztionInfo, the majority of changes happen in execution policies ?! Basically, there is no use of functionalities of OptimizationInfo ! So either PR or the implementation should be reconsidered.
3. About the ExecutionPolicyDecorator (and OptimizationProcessFactory in helper utilities) which is like a wrapper for execution policies, should be moved to the execution policies folders. It is not a simple utility. It does the most important calls of executions base on the analysis. Please also consider changing the names. Decorator ?!!! I would simply call it AnalysisExecutionWrapper, which is right to the point.

applications/OptimizationApplication/python_scripts/analyses_controller.py
```diff
@@ -10,7 +10,7 @@
 # ==============================================================================
 
 # additional imports
-from KratosMultiphysics.OptimizationApplication.execution_policies.execution_policy_wrapper import ExecutionPolicyWrapper
+from KratosMultiphysics.OptimizationApplication.utilities.execution_policy_decorator import ExecutionPolicyDecorator
```
RezaNajian: Please change the name of ExecutionPolicyDecorator such that it gets self-explaining. Something like AnalysisWrapper. Policy and Decorator do not sound good.

applications/OptimizationApplication/python_scripts/execution_policies/stepping_analysis_execution_policy.py
```diff
@@ -1,42 +1,33 @@
...
-#  KratosOptimizationApplication
-#
-#  License:         BSD License
-#                   license: OptimizationApplication/license.txt
-#
-#  Main authors:    Suneth Warnakulasuriya
-#
-# ==============================================================================
-
 import KratosMultiphysics as Kratos
```
RezaNajian, line 3: I suggest stepping_analysis_wrapper class name

applications/OptimizationApplication/python_scripts/optimization_info.py
```diff
@@ -0,0 +1,145 @@
+import KratosMultiphysics as Kratos
```
RezaNajian: Thanks a lot for introducing this class. This would be hell useful. Hoeve none of its functionality is used although this PR is about this class. Please consult my general comments in the description of the PR.
We must use its funcs at the base optimization algorithm class

applications/OptimizationApplication/python_scripts/utilities/execution_policy_decorator.py
```diff
@@ -0,0 +1,88 @@
...
+from KratosMultiphysics.OptimizationApplication.execution_policies.execution_policy import ExecutionPolicy
+from KratosMultiphysics.OptimizationApplication.utilities.logger_utilities import FileLogger
+from KratosMultiphysics.OptimizationApplication.utilities.helper_utilities import OptimizationProcessFactory
+
+class ExecutionPolicyDecorator(ExecutionPolicy):
+    def __init__(self, model: Kratos.Model, parameters: Kratos.Parameters):
+        super().__init__()
+
+        default_parameters = Kratos.Parameters("""{
+            "execution_policy_name"    : "",
```
RezaNajian: look, here I strongly suggest removing repeated execution_policy_ and also reducing the number of input settings. The default settings should be to use Kratos solvers like structural or fluid  solvers which can be identified from the analysis input file. Otherwise, we will come up with a long input file which can be easily avoided.

applications/OptimizationApplication/python_scripts/utilities/execution_policy_decorator.py
```diff
@@ -0,0 +1,88 @@
+import KratosMultiphysics as Kratos
```
RezaNajian: As I mentioned in my general comments, this the most important part of the analysiswrapper and it is not a simple utility. It must be moved to analysis_wrappers (execution_policies).

applications/OptimizationApplication/python_scripts/utilities/helper_utilities.py
```diff
@@ -0,0 +1,27 @@
+from importlib import import_module
```
RezaNajian, line 2: Please consult my general comments. The most important point is to remove optimization_info from the analysis stuff. optimization_info is about optimization things not analysis details.

RezaNajian (changes requested):
Again thanks a lot for introducing the OptimizationInfo, it is a building stone of the application. However, please try to address my points and reviews. Please note that this PR is about the OptimiaztionInfo but it is not used at all !!! rather we have a lot of changes in execution policies which I ask you to rename and reconsider. Such naming does not ring a bell.

sunethwarna:
Hi Reza, Thanks a lot for all your comments. Here are my reasonings/changes for your comments:


> 1. As described, the main motivation/contribution of the PR is the OptimizationInfo class. However, a lot of not necessary dependencies and complexities are introduced. For example, OptimizationInfo is only used in the OptimizationProcessFactory and nowhere else. OptimizationInfo should ONLY belong to the optimization stuff and not it's building blocks like analysis wrappers. Please note that in the OptimiaztionApplication master-slave approach is followed meaning that the optimizer is the master and ["model_parts","analyses","responses","controls"] are slaves. Therefore, I would like to ask you to remove OptimizationInfo from the OptimizationProcessFactory.

I still follow the master-slave approach of the optimization. The optimizor has the controll over what is being called. That controll is given by calling member functions of analysis, responses, etc. `OptimizationInfo` is merely a data container, which can be used to store and retrieve data at any point of the optimization process. Passing a data container to any point of optimization will not lose the master-slave constraint. (It is exactly like the `ProcessInfo` in `Kratos`). We won't be able to run `Fluid` optimization problems, if we don't have a common data container passed around in the optimization. Following is an example:

Fluid primal -> It is a standalone analysis wich does not know anything about adjoints
Fluid adjoint -> It is also a standalone analysis which only requires primal state. (We don't follow the structural approach since it is a too much memory consuming,  hazzle to debug, and restart in transient or steady fluid problems since we use pseudo time stepping and actual time stepping).

Fluid problems cannot be run in without creating a specific folder for each design iteration. So the primal solution will be written to a folder, and adjoint solution needs to know the solution. The `OptimizationInfo` can be used to pass around these information.

This problem will only grow when we goto coupled system optimizations, system identification, hence I would keep the `OptimizationInfo` passed everywhere possible. This was one of the biggest problems I had in `ShapeOptimizationApplication` before, so I used big hacks to get around it which is really not pleasant.

> 2. Although this PR is about OptimiaztionInfo, the majority of changes happen in execution policies ?! Basically, there is no use of functionalities of OptimizationInfo ! So either PR or the implementation should be reconsidered.

I purposely created a seperate PR like this, otherwise I have to merge few others in this pipe line (steps 2,3,4, ...) which will be a huge PR again and hard to review. The uses of this `OptimizationInfo` is in the upcoming PRs. I did not want to have a huge PR for this new introduction. Hence I would keep it as it is four your review convinience.

> 3. About the ExecutionPolicyDecorator (and OptimizationProcessFactory in helper utilities) which is like a wrapper for execution policies, should be moved to the execution policies folder. It is not a simple utility. It does the most important calls of executions based on the analysis. Please also consider changing the names. Decorator ?!!! I would simply call it AnalysisExecutionWrapper, which is right to the point.

These decorators, implementors come from the design patterns (Decorator design pattern, bridge design pattern). Wrapper (i.e. Adapter) design pattern gives a wrong indication about the design. Wrappers (i.e. Adapters) are used when you have non-conforming interfaces which is not the case. I put it there on the utilities so, they are not meant to be overriden by the developers. These are used to decouple response, execution policy interfaceimplementations with algorithm implementations so, it would be easier to extend any one of them. I have plans to extend `ExecutionPolicyDecorator` to support some othe functionalities from `Kratos::Core` which I can do without even touching `IndependentAnalysisExecutionPolicy` or `SteppingAnalysisExecutionPolicy` due to this design. Otherwise, I have to change codes in both.

> 4. Please also move optimization_info to the utilities folder. Only building blocks of the OptimizationApplication, i.e., [analyses_controller,controls_controller,model_parts_controller,optimizations_controller] and the folders should live there.

Done. :D 

> 5. Last but not least, the naming and unnecessary complexity of input variables at ExecutionPolicyDecorator. Please minimize the input settings of the analyses list using some default settings, pretty much like before the introduction of the execution policies. Please note that execution_policy_ gets repeated in four input vars. Please remove such redundancies.

Done. :D

applications/OptimizationApplication/python_scripts/analyses_controller.py
sunethwarna: addressed in my general answer

applications/OptimizationApplication/python_scripts/optimization_info.py
sunethwarna: addressed in my general answer

applications/OptimizationApplication/python_scripts/utilities/execution_policy_decorator.py
sunethwarna: done :D

applications/OptimizationApplication/python_scripts/utilities/execution_policy_decorator.py
sunethwarna: addressed in general answer

applications/OptimizationApplication/python_scripts/utilities/helper_utilities.py
sunethwarna: addressed in general answer

applications/OptimizationApplication/python_scripts/execution_policies/stepping_analysis_execution_policy.py
sunethwarna: addressed in general answer

RezaNajian:
> Decorator design pattern

Thanks Suneth for the clarifications. Now I understand more and the plans/intention behind this PR. So it is fine. Some question and requests for you:
1. I see that you moved the optimization info to the utils and I am wondering why you keep execution policy decorator in utilities. It does manage the executions and respective policies, for the sake of clarity and structure I highly recommend to move it to execution_policies.
2. I would also recommend to merge OptimizationProcessFactory into the ExecutionPolicyDecorator class. But its up you. I think we may come up with pasta code if we do not cluster and structure the application properly. In this for mit seems like ping ponging between python scripts, folders and classes.

applications/OptimizationApplication/python_scripts/utilities/helper_utilities.py
```diff
@@ -0,0 +1,27 @@
...
+
+    module = import_module(full_module_name)
+    if optimization_info is not None:
+        retrieved_object = getattr(module, class_name)(model, parameters, optimization_info)
+    else:
+        retrieved_object = getattr(module, class_name)(model, parameters)
+
+    # check retrieved object is of the required type
+    if not isinstance(retrieved_object, required_object_type):
+        raise RuntimeError(f"The retrieved object is of type \"{retrieved_object.__class__.__name__}\" which is not derrived from the \"{required_object_type.__name__}\".")
```
FMeister:
```suggestion
        raise RuntimeError(f"The retrieved object is of type \"{retrieved_object.__class__.__name__}\" which is not derived from the \"{required_object_type.__name__}\".")
```

Typo

applications/OptimizationApplication/python_scripts/analyses_controller.py
```diff
@@ -27,34 +27,34 @@ def __init__(self, execution_policies_settings, model, model_parts_controller):
 
         self.execution_policy_wrappers = {}
         for execution_policy_settings in self.execution_policies_settings:
-            execution_policy_wrapper = ExecutionPolicyWrapper(self.model, execution_policy_settings)
+            execution_policy_wrapper = ExecutionPolicyDecorator(self.model, execution_policy_settings)
             if not execution_policy_wrapper.GetExecutionPolicyName() in self.execution_policy_wrappers.keys():
                 self.execution_policy_wrappers[execution_policy_wrapper.GetExecutionPolicyName()] = execution_policy_wrapper
             else:
                 raise RuntimeError(f"Found already existing exeuction policy with the name \"{execution_policy_wrapper.GetExecutionPolicyName()}\". Please provide unique names.")
```
FMeister:
```suggestion
                raise RuntimeError(f"Found already existing execution policy with the name \"{execution_policy_wrapper.GetExecutionPolicyName()}\". Please provide unique names.")
```

Typo

applications/OptimizationApplication/python_scripts/analyses_controller.py
```diff
@@ -27,34 +27,34 @@ def __init__(self, execution_policies_settings, model, model_parts_controller):
 
         self.execution_policy_wrappers = {}
```
FMeister:
```suggestion
        self.execution_policy_decorators = {}
```

Rename wrapper -> decorator

applications/OptimizationApplication/python_scripts/execution_policies/execution_policy.py
```diff
@@ -1,27 +1,8 @@
...
+class ExecutionPolicy(Kratos.Process):
+    def __init__(self):
+        super().__init__()
 
-    @abstractmethod
-    def Execute(self, optimization_info: dict):
-        pass
\ No newline at end of file
+    def Execute(self):
+        raise NotImplementedError("Calling base class ExecutionPolicy::Execute method. This should be implemented in the derrived class.")
```
FMeister:
```suggestion
        raise NotImplementedError("Calling base class ExecutionPolicy::Execute method. This should be implemented in the derived class.")
```
Typo

applications/OptimizationApplication/python_scripts/utilities/optimization_info.py
```diff
@@ -0,0 +1,145 @@
...
+        self.__optimization_processes: 'dict[str, Kratos.Process]' = {}
+
+        # initializing the buffer variables
+        self.__iteration_data = []
+        self.__buffer_index = 0
+        self.__echo_level = echo_level
+
+    def AddOptimizationProcess(self, category: any, routine_name: str, routine: Kratos.Process):
+        if not isinstance(routine, Kratos.Process):
+            raise RuntimeError(f"Only allowed to add objects derrived from Kratos.Process. [ Type of added routine: {routine.__class__.__name__}]")
```
FMeister:
```suggestion
            raise RuntimeError(f"Only allowed to add objects derived from Kratos.Process. [ Type of added routine: {routine.__class__.__name__}]")
```

Typo

applications/OptimizationApplication/python_scripts/utilities/optimization_info.py
```diff
@@ -0,0 +1,145 @@
...
+
+        if not isinstance(routine, category):
+            raise RuntimeError(f"The provided routine with name \"{routine_name}\" is not derrived from the provided category type \"{category.__name__}\".")
```
FMeister:
```suggestion
            raise RuntimeError(f"The provided routine with name \"{routine_name}\" is not derived from the provided category type \"{category.__name__}\".")
```

Typo

applications/OptimizationApplication/tests/test_optimization_info.py
```diff
@@ -0,0 +1,96 @@
...
+        self.assertEqual(optimization_info["step"], 3)
+        self.assertEqual(optimization_info.GetSolutionStepData(1)["step"], 2)
+        self.assertEqual(optimization_info.GetSolutionStepData(2)["step"], 1)
+
+        optimization_info.AdvanceSolutionStep()
+        optimization_info["step"] = 4
+        self.assertEqual(optimization_info["step"], 4)
+        self.assertEqual(optimization_info.GetSolutionStepData(1)["step"], 3)
+        self.assertEqual(optimization_info.GetSolutionStepData(2)["step"], 2)
+        self.assertEqual(optimization_info.GetSolutionStepData(3)["step"], 4)
```
FMeister:
```suggestion
        self.assertEqual(optimization_info.GetSolutionStepData(3)["step"], 1)
```

Why do we expect a 4 here? Shouldn't it be a 1 to follow the pattern?

applications/OptimizationApplication/tests/test_optimization_info.py
```diff
@@ -0,0 +1,96 @@
...
+        self.assertEqual(optimization_info.GetSolutionStepData(2)["step"], 15)
+
+        optimization_info.AdvanceSolutionStep()
+        optimization_info["step"] = 4
+        self.assertEqual(optimization_info["step"], 4)
+        self.assertEqual(optimization_info.GetSolutionStepData(1)["step"], 3)
+        self.assertEqual(optimization_info.GetSolutionStepData(2)["step"], 2)
+        self.assertEqual(optimization_info.GetSolutionStepData(3)["step"], 4)
+
+    def test_OptimizationProcesss(self):
```
FMeister:
```suggestion
    def test_OptimizationProcess(self):
```

Typo

FMeister (changes requested):
Thanks for the PR Suneth. I found some Typos and have a Question about one of your tests.

Regarding the Rezas comments:
1. Placement of the ExecutionPolicyDecorator class: Since it is named "ExecutionPolicy..." and decorates the ExecutionPolicy class i would expect it to live inside the execution_policies folder. Even though it does not contain any sophisticated logic.
2. To me the file structure is much more clearer if we don't merge multiple classes into one file. I would also suggest renaming the file to the name of the class. How is this typically handled in Kratos?

applications/OptimizationApplication/tests/test_optimization_info.py
sunethwarna: This is to check cyclic nature of the buffer. It is initialized with buffer size 3 in line 15, so the indices 0, 1, 2 refers to different buffer positions. But the index 3 is same as index 0 because of the cyclic nature. Thats why we should get a "4" in this one, and not "1".

applications/OptimizationApplication/python_scripts/analyses_controller.py
sunethwarna: done :)

sunethwarna:
Hi @RezaNajian  and @FMeister, Thanks a lot for all the comments and I addressed your concerns. Can you take a look?

applications/OptimizationApplication/tests/test_optimization_info.py
sunethwarna: This is converted to an error. Now if solution step is requested which is out of the buffer size, an error is thrown.

RezaNajian (dismissed):
All look good. Thanks, @sunethwarna for the consideration of the reviews and comments.

sunethwarna:
Hi Reza, can you re-approve? I had some merge conflicts with your recent PR. Thanks.

applications/OptimizationApplication/python_scripts/utilities/helper_utilities.py
```diff
@@ -0,0 +1,48 @@
+from pathlib import Path
+from importlib import import_module
+
+import KratosMultiphysics as Kratos
+from KratosMultiphysics.OptimizationApplication.utilities.optimization_info import OptimizationInfo
+from KratosMultiphysics.kratos_utilities import GetListOfAvailableApplications
+from KratosMultiphysics.kratos_utilities import GetKratosMultiphysicsPath
+
+def GetClassMouleFromKratos(class_name: str) -> str:
```
FMeister:
```suggestion
def GetClassModuleFromKratos(class_name: str) -> str:
```
Typo

FMeister: Besides this little typo the PR is fine for me. Thanks

FMeister (approved):

PR #10765, opened 2023-02-17
[OptApp] Restructure 2 - Add mass response function

sunethwarna:
**📝 Description**
This PR adds `MassResonseFunction` to the optimization application.

**🆕 Changelog**
- Adds `MassResponseFunction`
- Adds `OptimizationUtils`
- Adds tests


applications/OptimizationApplication/custom_utilities/optimization_utils.cpp
```diff
@@ -0,0 +1,150 @@
...
+        }
+    } else {
+        return GeometryData::KratosGeometryType::Kratos_generic_type;
+    }
+
+    KRATOS_CATCH("")
+}
+
+template<class TContainerType, class TDataType>
+bool OptimizationUtils::IsVariableExistsInAllContainerProperties(
```
FMeister, line 55:
```suggestion
bool OptimizationUtils::DoesVariableExistInAllContainerProperties(
```
or "VariableExistInAllContainerProperties()"
I was a little bit confused by the is... in front and find it more readable this way.
Is there a convention in Kratos to only use Is... if the function returns a bool?

applications/OptimizationApplication/custom_utilities/optimization_utils.cpp
```diff
@@ -0,0 +1,150 @@
...
+        return rEntity.GetProperties().Has(rVariable);
+    });
+
+    return rDataCommunicator.AndReduceAll(local_value);
+
+    KRATOS_CATCH("");
+}
+
+template<class TContainerType, class TDataType>
+bool OptimizationUtils::IsVariableExistsInAtLeastOneContainerProperties(
```
FMeister, line 72:
```suggestion
bool OptimizationUtils::DoesVariableExistInAtLeastOneContainerProperty(
```
For consistency.

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
```diff
@@ -0,0 +1,271 @@
...
+    if (rModelPart.GetCommunicator().GlobalNumberOfElements() == 0) {
+        return 0.0;
+    }
+
+    KRATOS_ERROR_IF_NOT(HasVariableInProperties(rModelPart, DENSITY))
+        << "DENSITY is not found in element properties of " << rModelPart.FullName() << ".\n";
+
+    KRATOS_ERROR_IF(HasVariableInProperties(rModelPart, THICKNESS) && HasVariableInProperties(rModelPart, CROSS_AREA))
+        << rModelPart.FullName()
+        << " has elements with properties having both THICKNESS and CROSS_AREA. Please seperate the model part such that either one of them is present in elemental properties.\n";
```
FMeister:
```suggestion
        << " has elements with properties having both THICKNESS and CROSS_AREA. Please separate the model part such that either one of them is present in elemental properties.\n";
```
Typo

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
```diff
@@ -0,0 +1,271 @@
...
+    KRATOS_TRY
+
+    if (rModelPart.GetCommunicator().GlobalNumberOfElements() == 0) {
+        return 0.0;
+    }
+
+    KRATOS_ERROR_IF_NOT(HasVariableInProperties(rModelPart, DENSITY))
+        << "DENSITY is not found in element properties of " << rModelPart.FullName() << ".\n";
+
+    KRATOS_ERROR_IF(HasVariableInProperties(rModelPart, THICKNESS) && HasVariableInProperties(rModelPart, CROSS_AREA))
```
FMeister, line 104: Should we also check that the modelpart has at least on of both (THICKNESS or CROSS_AREA)?

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
```diff
@@ -0,0 +1,271 @@
...
+    if (rModelPart.NumberOfElements() == 0) {
+        return;
+    }
+
+    KRATOS_ERROR_IF_NOT(HasVariableInProperties(rModelPart, DENSITY))
+        << "DENSITY is not found in element properties of " << rModelPart.FullName() << ".\n";
+
+    KRATOS_ERROR_IF(HasVariableInProperties(rModelPart, THICKNESS) && HasVariableInProperties(rModelPart, CROSS_AREA))
+        << rModelPart.FullName()
+        << " has elements with properties having both THICKNESS and CROSS_AREA. Please seperate the model part such that either one of them is present in elemental properties.\n";
```
FMeister:
```suggestion
        << " has elements with properties having both THICKNESS and CROSS_AREA. Please separate the model part such that either one of them is present in elemental properties.\n";
```
Typo

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
```diff
@@ -0,0 +1,271 @@
...
+    VariableUtils().SetNonHistoricalVariableToZero(rOutputSensitivityVariable, rModelPart.Nodes());
+
+    if (rModelPart.NumberOfElements() == 0) {
+        return;
+    }
+
+    KRATOS_ERROR_IF_NOT(HasVariableInProperties(rModelPart, DENSITY))
+        << "DENSITY is not found in element properties of " << rModelPart.FullName() << ".\n";
+
+    KRATOS_ERROR_IF(HasVariableInProperties(rModelPart, THICKNESS) && HasVariableInProperties(rModelPart, CROSS_AREA))
```
FMeister, line 176: Same as before

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
```diff
@@ -0,0 +1,271 @@
...
+            };
+            break;
+        // Following is not consistent with the DomainSize calculation, hence commented out for the time being.
+        // case GeometryData::KratosGeometryType::Kratos_Hexahedra3D8:
+        //     volume_derivative_method = [](IndexType NodeIndex, IndexType DirectionIndex,  const GeometryType& rGeometry) {
+        //         return 3.0 * std::pow(ElementSizeCalculator<3, 8>::AverageElementSize(rGeometry), 2) * ElementSizeCalculator<3, 8>::AverageElementSizeDerivative(NodeIndex, DirectionIndex, rGeometry);
+        //     };
+        //     break;
+        default:
+            KRATOS_ERROR << "Non supported geometry type." << std::endl;
```
FMeister:
```suggestion
            KRATOS_ERROR << "Non supported geometry type for mass shape sensitivity calculation (CalculateMassShapeSensitivity())." << std::endl;
```
Make the error more specific

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
```diff
@@ -0,0 +1,271 @@
...
+    const Variable<double>& rOutputSensitivityVariable)
+{
+    KRATOS_TRY
+
+    KRATOS_ERROR_IF_NOT(HasVariableInProperties(rModelPart, DENSITY))
+        << "DENSITY is not found in element properties of " << rModelPart.FullName() << ".\n";
+
+    KRATOS_ERROR_IF(HasVariableInProperties(rModelPart, THICKNESS) && HasVariableInProperties(rModelPart, CROSS_AREA))
+        << rModelPart.FullName()
+        << " has elements with properties having both THICKNESS and CROSS_AREA. Please seperate the model part such that either one of them is present in elemental properties.\n";
```
FMeister:
```suggestion
        << " has elements with properties having both THICKNESS and CROSS_AREA. Please separate the model part such that either one of them is present in elemental properties.\n";
```
Typo

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
```diff
@@ -0,0 +1,271 @@
...
+void MassResponseUtils::CalculateMassDensitySensitivity(
+    ModelPart& rModelPart,
+    const Variable<double>& rOutputSensitivityVariable)
+{
+    KRATOS_TRY
+
+    KRATOS_ERROR_IF_NOT(HasVariableInProperties(rModelPart, DENSITY))
+        << "DENSITY is not found in element properties of " << rModelPart.FullName() << ".\n";
+
+    KRATOS_ERROR_IF(HasVariableInProperties(rModelPart, THICKNESS) && HasVariableInProperties(rModelPart, CROSS_AREA))
```
FMeister, line 287: Same here ...

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+double MassResponseUtils::CalculateMass(const ModelPart& rModelPart)
+{
+    KRATOS_TRY
+
+    if (rModelPart.GetCommunicator().GlobalNumberOfElements() == 0) {
+        return 0.0;
+    }
+
+    KRATOS_ERROR_IF_NOT(HasVariableInProperties(rModelPart, DENSITY))
```
FMeister, line 101: The section of getting the thickness and the cross_area are the same in all functions. Should we move that to a separate function then?

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
```diff
@@ -0,0 +1,271 @@
...
+    ModelPart& rModelPart,
+    const Variable<double>& rOutputSensitivityVariable)
+{
+    CalculateMassGeometricalPropertySensitivity(rModelPart, CROSS_AREA, THICKNESS, rOutputSensitivityVariable);
+}
+
+void MassResponseUtils::CalculateMassGeometricalPropertySensitivity(
+    ModelPart& rModelPart,
+    const Variable<double>& rGeometricalPropertySensitivityVariable,
+    const Variable<double>& rGeometricalCoflictingPropertySensitivityVariable,
```
FMeister:
```suggestion
    const Variable<double>& rGeometricalConflictingPropertySensitivityVariable,
```
Typo

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+    KRATOS_ERROR_IF_NOT(HasVariableInProperties(rModelPart, DENSITY))
+        << "DENSITY is not found in element properties of " << rModelPart.FullName() << ".\n";
+
+    KRATOS_ERROR_IF_NOT(HasVariableInProperties(rModelPart, rGeometricalPropertySensitivityVariable))
+        << rGeometricalPropertySensitivityVariable.Name() << " is not found in element properties of "
+        << rModelPart.FullName() << " which is required to compute sensitivities w.r.t. "
+        << rGeometricalPropertySensitivityVariable.Name() << ".\n";
+
+    KRATOS_ERROR_IF(HasVariableInProperties(rModelPart, rGeometricalCoflictingPropertySensitivityVariable))
```
FMeister:
```suggestion
    KRATOS_ERROR_IF(HasVariableInProperties(rModelPart, rGeometricalConflictingPropertySensitivityVariable))
```
Typo

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
```diff
@@ -0,0 +1,271 @@
...
+        << rModelPart.FullName() << " has elements with properties having both "
+        << rGeometricalPropertySensitivityVariable.Name() << " and " << rGeometricalCoflictingPropertySensitivityVariable.Name()
+        << ". Please seperate the model part such that either one of them is present in elemental properties.\n";
```
FMeister:
```suggestion
        << ". Please separate the model part such that either one of them is present in elemental properties.\n";
```
Typo

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
```diff
@@ -0,0 +1,55 @@
...
+    def CalculateSensitivity(self, sensitivity_variable: any, sensitivity_model_part: Kratos.ModelPart):
+        if sensitivity_variable == Kratos.SHAPE_SENSITIVITY:
+            KratosOA.ResponseUtils.MassResponseUtils.CalculateMassShapeSensitivity(sensitivity_model_part, sensitivity_variable)
+        elif sensitivity_variable == KratosOA.DENSITY_SENSITIVITY:
+            KratosOA.ResponseUtils.MassResponseUtils.CalculateMassDensitySensitivity(sensitivity_model_part, sensitivity_variable)
+        elif sensitivity_variable == KratosOA.THICKNESS_SENSITIVITY:
+            KratosOA.ResponseUtils.MassResponseUtils.CalculateMassThicknessSensitivity(sensitivity_model_part, sensitivity_variable)
+        elif sensitivity_variable == KratosOA.CROSS_AREA_SENSITIVITY:
+            KratosOA.ResponseUtils.MassResponseUtils.CalculateMassCrossAreaSensitivity(sensitivity_model_part, sensitivity_variable)
+        elif sensitivity_variable in [KratosOA.YOUNG_MODULUS_SENSITIVITY]:
```
FMeister: Is there a use case for this sensitivity (mass with respect to young's modulus)?

applications/OptimizationApplication/python_scripts/responses/response_function.py
```diff
@@ -0,0 +1,17 @@
+import KratosMultiphysics as Kratos
+
+class ResponseFunction(Kratos.Process):
+    def __init__(self):
+        super().__init__()
+
+    def Check(self):
+        raise NotImplementedError("Calling base class ResponseFunction::Check method. Please implement in the derrived class.")
```
FMeister:
```suggestion
        raise NotImplementedError("Calling base class ResponseFunction::Check method. Please implement in the derived class.")
```

applications/OptimizationApplication/python_scripts/responses/response_function.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def CalculateValue(self) -> float:
+        raise NotImplementedError("Calling base class ResponseFunction::CalculateValue method. Please implement in the derrived class.")
```
FMeister:
```suggestion
        raise NotImplementedError("Calling base class ResponseFunction::CalculateValue method. Please implement in the derived class.")
```

applications/OptimizationApplication/python_scripts/responses/response_function.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def CalculateSensitivity(self, sensitivity_variable: any, sensitivity_model_part: Kratos.ModelPart):
+        raise NotImplementedError("Calling base class ResponseFunction::CalculateSensitivity method. Please implement in the derrived class.")
```
FMeister:
```suggestion
        raise NotImplementedError("Calling base class ResponseFunction::CalculateSensitivity method. Please implement in the derived class.")
```

applications/OptimizationApplication/python_scripts/responses/response_function.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def GetModelPart(self) -> Kratos.ModelPart:
+        raise NotImplementedError("Calling base class ResponseFunction::GetModelPart method. Please implement in the derrived class.")
```
FMeister:
```suggestion
        raise NotImplementedError("Calling base class ResponseFunction::GetModelPart method. Please implement in the derived class.")
```

applications/OptimizationApplication/tests/test_mass_response_function.py
```diff
@@ -0,0 +1,254 @@
...
+        ref_value = response_function.CalculateValue()
+        for entity in entities:
+            v = sensitivity_method(entity)
+            update_method(entity, delta)
+            value = response_function.CalculateValue()
+            sensitivity = (value - ref_value)/delta
+            update_method(entity, -delta)
+            self.assertAlmostEqual(v, sensitivity, precision)
+
+    def _UpdateProperties(self, variable, entity, delta):
```
FMeister, line 34:
```suggestion
    def _IncreasePropertiesBy(self, variable, entity, delta):
```
For me "update" suggests that the value gets override.

applications/OptimizationApplication/tests/test_mass_response_function.py
```diff
@@ -0,0 +1,254 @@
...
+        entity.Properties[variable] += delta
+
+    def _UpdateNodalPositions(self, direction, entity, delta):
```
FMeister, line 37:
```suggestion
    def _IncreaseNodalPositionsBy(self, direction, entity, delta):
```
Same here.

applications/OptimizationApplication/tests/test_optimization_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        self.assertEqual(self.utils.GetVariableDimension(Kratos.VELOCITY, 2), 2)
+
+    def test_IsVariableExistsInAllContainerProperties(self):
+        self.assertTrue(self.utils.IsVariableExistsInAllContainerProperties(self.model_part.Conditions, Kratos.PRESSURE, self.model_part.GetCommunicator().GetDataCommunicator()))
+        for condition in self.model_part.Conditions:
+            if condition.Id != 4:
+                condition.Properties[Kratos.DISTANCE] = 1.0
+        self.assertFalse(self.utils.IsVariableExistsInAllContainerProperties(self.model_part.Conditions, Kratos.DISTANCE, self.model_part.GetCommunicator().GetDataCommunicator()))
+
+    def test_IsVariableExistsInAtLeastOneContainerProperties(self):
```
FMeister, line 53:
```suggestion
    def test_DoesVariableExistInAtLeastOneContainerProperty(self):
```
or "VariableExistInAtLeastOneContainerProperty(self)"
Same argument as before.

applications/OptimizationApplication/tests/test_optimization_utils.py
```diff
@@ -0,0 +1,61 @@
...
+            element.SetValue(Kratos.PRESSURE, id+5)
+            element.SetValue(Kratos.VELOCITY, Kratos.Array3([id+6, id+7, id+8]))
+
+        cls.utils = KratosOA.OptimizationUtils
+
+    def test_GetVariableDimension(self):
+        self.assertEqual(self.utils.GetVariableDimension(Kratos.PRESSURE, 1), 1)
+        self.assertEqual(self.utils.GetVariableDimension(Kratos.VELOCITY, 2), 2)
+
+    def test_IsVariableExistsInAllContainerProperties(self):
```
FMeister, line 46:
```suggestion
    def test_DoesVariableExistInAllContainerProperties(self):
```
or "VariableExistInAllContainerProperties(self)"
Same here.

FMeister (commented):
Overall the PR is ok for me.

RezaNajian:
> **📝 Description** This PR adds `MassResonseFunction` to the optimization application.
> 
> **🆕 Changelog**
> 
> * Adds `MassResponseFunction`
> * Adds `OptimizationUtils`
> * Adds tests

@sunethwarna Could you please mention changes and the reason/motivation behind? We already have implemented MassResponse function without utilities and this PR makes mass response use of only mass utilities at cpp level, which may not fit to the road map of optapp.

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
```diff
@@ -0,0 +1,271 @@
+//    |  /           |
```
RezaNajian, line 1: @sunethwarna I am not convinced to change the MassReponse to the utilities. We lose consistency if we do so. Mass is a response and its calculations should be under MassResponse class not use utilities to do so. Anyway, we should talk about it. Maybe thereare  reasons.

RezaNajian (changes requested):
@sunethwarna as I mentioned before, this PR changes the structures and flow of optimizationapplication. Let us meet and discuss. There are some scenarios that are overlooked.

applications/OptimizationApplication/custom_utilities/optimization_utils.cpp
sunethwarna: `Is` is used in Kratos to let the developer know that it is returining a bool.

applications/OptimizationApplication/custom_utilities/optimization_utils.cpp
sunethwarna: same comment as above

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
sunethwarna: I would not, because solid elements will not have both. Only shell elements (or plate) will have `THICKNESS`, beam (or truss) elements will have `CROSS_AREA`.

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
sunethwarna: see previous answer.

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
sunethwarna: see previous answer

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
sunethwarna: In the `SIMP` control it is required.

applications/OptimizationApplication/tests/test_mass_response_function.py
sunethwarna: The value is actually getting overriden by the update. `IncreasePropertiesBy` will be kind of misleading, because the properties can increase or decrease depending on the optimization outcome.

applications/OptimizationApplication/tests/test_mass_response_function.py
sunethwarna: same as before

applications/OptimizationApplication/tests/test_optimization_utils.py
sunethwarna: same answer as before

applications/OptimizationApplication/tests/test_optimization_utils.py
sunethwarna: same answer as before

applications/OptimizationApplication/custom_python/add_custom_response_utilities_to_python.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Kratos {
+namespace Python {
+
+void  AddCustomResponseUtilitiesToPython(pybind11::module& m)
+{
+    namespace py = pybind11;
+
+    py::class_<MassResponseUtils >(m, "MassResponseUtils")
+        .def_static("CalculateMass", &MassResponseUtils::CalculateMass)
```
RezaNajian: Please create a single function that can calculate sensitivities for different types of sensitivity as input. The function should be similar to elements that use 'CalculateOnIntegrationPoints' to ensure consistency and structure. By doing so, we can reduce unnecessary Python exposure and streamline our processes.

applications/OptimizationApplication/python_scripts/responses/response_function.py
```diff
@@ -0,0 +1,17 @@
+import KratosMultiphysics as Kratos
+
+class ResponseFunction(Kratos.Process):
```
RezaNajian: I was wondering if you could help me understand why you derive Response from Process. I'm a bit confused and would appreciate it if you could explain the reasoning behind this choice. Thank you!

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
```diff
@@ -0,0 +1,55 @@
...
+        default_settings = Kratos.Parameters("""{
+            "evaluated_model_part_name": "PLEASE_PROVIDE_A_MODEL_PART_NAME"
+        }""")
+        parameters.ValidateAndAssignDefaults(default_settings)
+        self.model_part = model[parameters["evaluated_model_part_name"].GetString()]
+
+    def GetModelPart(self) -> Kratos.ModelPart:
+        return self.model_part
+
+    def Check(self):
```
RezaNajian: The checks currently in place seem to be more relevant to mass response utilities rather than optimization utilities, and may benefit from being moved to mass response or mass utilities.

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
```diff
@@ -0,0 +1,55 @@
...
+        data_communicator = self.model_part.GetCommunicator().GetDataCommunicator()
+        if not KratosOA.OptimizationUtils.IsVariableExistsInAllContainerProperties(self.model_part.Elements, Kratos.DENSITY, data_communicator):
+            raise RuntimeError(f"Some elements' properties in {self.model_part.FullName()} does not have DENSITY variable.")
+
+        if KratosOA.OptimizationUtils.IsVariableExistsInAtLeastOneContainerProperties(self.model_part.Elements, Kratos.THICKNESS, data_communicator) and \
```
RezaNajian: I was wondering if it would be possible to minimize the naming of functions and their use, as it can be quite long. Would you mind considering the possibility of making them more concise?

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
```diff
@@ -0,0 +1,55 @@
...
+           KratosOA.OptimizationUtils.IsVariableExistsInAtLeastOneContainerProperties(self.model_part.Elements, KratosOA.CROSS_AREA, data_communicator):
```
RezaNajian: IsVariableExistsInAtLeastOneContainerProperties is toooo long we need to find a better naming and explain the function in its implementation.

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
```diff
@@ -0,0 +1,55 @@
...
+           raise RuntimeError(f"{self.model_part.FullName()} has elements consisting THICKNESS and CROSS_AREA. Please break down this response to SumResponseFunction where each sub response function only has elements with either THICKNESS or CROSS_AREA.")
+
+        if not KratosOA.OptimizationUtils.AreAllEntitiesOfSameGeometryType(self.model_part.Elements, data_communicator):
+            raise RuntimeError(f"{self.model_part.FullName()} has elements with different geometry types. Please break down this response to SumResponseFunction where each sub response function only has elements with one geometry type.")
+
+    def CalculateValue(self) -> float:
+        return KratosOA.ResponseUtils.MassResponseUtils.CalculateMass(self.model_part)
+
+    def CalculateSensitivity(self, sensitivity_variable: any, sensitivity_model_part: Kratos.ModelPart):
+        if sensitivity_variable == Kratos.SHAPE_SENSITIVITY:
```
RezaNajian: Please move these checks and calls to c++ and make a single function for CalculateSensitivity to be exposed to python.

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
RezaNajian: Here is exactly the point that makes the implementation wrong or confusing. In fact, there is no dependency of mass on young modulus and you may get it wrong from the SIMP technique. We definitely need to discuss this matter and resolve it before finalizing.

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
```diff
@@ -0,0 +1,55 @@
...
+            zero_sensitivity.SetDataForContainerVariableToZero(KratosOA.YOUNG_MODULUS_SENSITIVITY)
+            zero_sensitivity.AssignDataToContainerVariable(KratosOA.YOUNG_MODULUS_SENSITIVITY)
+        else:
+            msg = f"Unsupported sensitivity w.r.t. {sensitivity_variable.Name()} requested for {sensitivity_model_part.FullName()}. "
+            msg += "Followings are supported variables:"
+            msg += "\n\tSHAPE_SENSITIVITY"
+            msg += "\n\tDENSITY_SENSITIVITY"
+            msg += "\n\tTHICKNESS_SENSITIVITY"
+            msg += "\n\tCROSS_AREA_SENSITIVITY"
+            msg += "\n\tYOUNG_MODULUS_SENSITIVITY"
```
RezaNajian: This does not exist !!!

applications/OptimizationApplication/optimization_application.cpp
```diff
@@ -35,100 +35,100 @@
 namespace Kratos
 {
     KratosOptimizationApplication::KratosOptimizationApplication() :
-        KratosApplication("OptimizationApplication"),  
+        KratosApplication("OptimizationApplication"),
         /* ELEMENTS */
         mHelmholtzSurfShape3D3N(0, Element::GeometryType::Pointer(new Triangle3D3<NodeType >(Element::GeometryType::PointsArrayType(3)))),
```
RezaNajian, line 40: Thanks for the adjustments and cleanings here.

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.h
```diff
@@ -0,0 +1,81 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
```
RezaNajian, line 2: @sunethwarna Thank you for your work on the implementation. I was wondering if you could explain the major benefits of having response utilities instead of response function in C++. Is it primarily about parallelization? When I review the implementation, it seems that everything is the same as the implementation we currently have under custom responses, with only namings changed. I am having difficulty understanding the reason for this change. Would you be willing to discuss this with me further?

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
```diff
@@ -0,0 +1,271 @@
...
+                                    : [](const ModelPart::ElementType& rElement) { return 1.0; };
+
+    block_for_each(rModelPart.Elements(), [&](auto& rElement) {
+        rElement.GetProperties().SetValue(rOutputSensitivityVariable, rElement.GetGeometry().DomainSize() * get_thickness(rElement) * get_cross_area(rElement));
+    });
+
+    KRATOS_CATCH("")
+}
+
+void MassResponseUtils::CalculateMassThicknessSensitivity(
```
RezaNajian, line 310: Definitely, we can simplify the implementation by creating a single function for sensitivity analysis. Based on the type, we can then call different sensitive functions. Furthermore, we need to be able to call a single function from Python to compute the sensitivity of a response with respect to different types. However, it's important to note that this implementation does not support asking for sensitivities with respect to multiple properties in the case of concurrent optimization. We will need to address this issue in our implementation. Thank you.

RezaNajian (changes requested):
Dear Suneth, 

Thank you very much for your hard work on the implementations. I appreciate your efforts. However, I would like to discuss some major open points and issues that I have identified during my review. In light of these issues, I think it would be beneficial for us to meet again and discuss the best path forward.

Also, would you mind adding a brief description of the major changes or benefits of this PR in the description? This will help ensure that other team members are aware of the key improvements. Thank you very much for your time and consideration.

Cheers,
Reza

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
sunethwarna: I will remove it in this PR, and we can discuss it in the SIMP control PR.

applications/OptimizationApplication/python_scripts/responses/response_function.py
sunethwarna: I will create a base class with `Initialize`, `InitializeSolutionStep`, `FinalizeSolutionStep`, `Finalize`, `Check`

applications/OptimizationApplication/custom_python/add_custom_response_utilities_to_python.cpp
sunethwarna: I will move the if blocks to cpp side

sunethwarna: And will remove the old implementation, and re-use the new implementation.

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
sunethwarna: This will be moved to cpp, and I will remove `YOUNGS_MODULUES` from this PR, we can discuss it in SIMP control.

applications/OptimizationApplication/python_scripts/responses/response_function.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
sunethwarna: removed. done :)

applications/OptimizationApplication/custom_python/add_custom_response_utilities_to_python.cpp
sunethwarna: Ok, I am sorry, but the exisitng implementation cannot be removed because of the following reasons
1. The existing implementation uses projected values to compute mass, and its derivatives. This is not done in the new deisgn. `Responses` does not need to know whether their `THICKNESS`, `DENSITY`, `CROSS_AREA` are projected or not. They should be able to calculate mass sensitivity w.r.t. each `THICKNESS`, `DENSITY`, `CROSS_AREA`, etc without ever knowing whether they are independent/projected or not. The control mechanism decides whether they are independent or dependent, based on that control should have modification within it self. Otherwise, we will need seperate mass response for different control mechanisms such as SIMP, raw mass response for SI, raw mass response for constraints which is unmanagable.
2. If I am to use the proposed `MassResponseUtils` inside exsiting responses, the senstivitiy values of existing optimization factory test cases will be wrong, hence the tests will be useless.

Therefore, we have to live with both implementations for the time being untill the new control mechanism which works with this proposed new `Response` design is implemented to properly port these tests. I found this same reason when I was looking at this the first time, that was why I could not replace existing reposnes with the new ones. But I forgot the reasons. I wrote it here for my reference as well.

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
sunethwarna: Done :D

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.cpp
sunethwarna: As we discussed, I moved the if blocks to the cpp level. Hence closing this comment.

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.h
sunethwarna: As we discussed, main reasoning to have python response function and supporting utilities in cpp is to following reasons:
1. Having python response function allows developers to use both python level (i.e. `AnalysisStage`, etc) and `CPP` level developments without any problems.
2. Utilities are not meant to have object with life, therefore, they are made `static`. Hence, all the inputs and outputs are passed throught the function signature (no hidden tasks). This will be easier for a developer to follow than following two objects with life in python and c++.

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
sunethwarna: I am also open to any concise suggestions. :D

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
sunethwarna: I am also open to any concise suggestions. :D

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
sunethwarna: done :D

applications/OptimizationApplication/python_scripts/responses/response_function.py
```diff
@@ -0,0 +1,31 @@
...
+
+    @abstractmethod
+    def Check(self) -> None:
+        pass
+
+    @abstractmethod
+    def CalculateValue(self) -> float:
+        pass
+
+    @abstractmethod
```
RezaNajian, line 25: @sunethwarna This implementation of CalculateSensitivity is not general enough. For example, what if I have a response like Drag, which is defined on a submodel part like fluid.airfoil, but I want to change the shape of the fluid.inlet and the topology of a region inside the fluid domain, say fluid.topology? The whole idea behind the current implementation is to have maximum freedom with respect to where the function is evaluated, which can be a list of submodel parts (the "evaluated_objects" in the current implementation), a list of the controlled model parts (the "controlled_objects" in the current implementation), and a list of the control types ("control_types" in the current implementation). However, these have all been removed and simplified without reasoning. I think we need to discuss this further because you may have overlooked and missed covering such scenarios.

sunethwarna: Actually, both of them works in this implementation. I can explain it in the discussion. This is general enoguh as far as I see. The current implementation mixes `controls` responsibilities and `response` responsibilities which is hard to extend or manage.

sunethwarna: The control is suppose to know where it is controlling, not the `Response`. `Response` should not care where the controls are, it should just give sensitivities for the requested `sensitivity_model_part` w.r.t. requested `sensitivity_variable`, in there I am allowing multiple model parts. So control passes the `senstivity_model_part` which can be `fluid.inlet` and/or `fluid.topology`. The `sensitivity_model_part` can be different from the `evaluated_model_part`. The `evaluated_model_part` and `sensitivity_model_part` does not need to even have a the common root model part. Because sometimes (like in RANS and RANS sensitivities), we create model parts where the root model parts are different, but nodes, connectivities are the same with different elements.

And the example you gave is what exactly we are doing currently with `ShapeOptimizationApplication` as well in fluids where `evaluated_model_part` and `sensitivity_model_part` are different.

applications/OptimizationApplication/custom_utilities/response/mass_response_utils.h
```diff
@@ -0,0 +1,88 @@
...
+    ///@}
+    ///@name Static operations
+    ///@{
+
+    static void Check(const ModelPart& rModelPart);
+
+    static double CalculateValue(const ModelPart& rModelPart);
+
+    template<class TDataType>
+    static void CalculateSensitivity(
```
RezaNajian, line 54: The input should be a list of modelparts and a list of SensitivityVariable, due to the following facts:
1. since this function resets the rSensitivityVariable to zero at every call, and if the modelsparts have overlaps or common interfaces we will not get the full sensitivities on the interfaces. 
2. for adjoint-based sensitivities, we need a  list of rSensitivityVariable because there should be a single adjoint problem for multi controls like shape, thickness, .... If this function is called separately for every control variable then the adjoint problem should be solved at every call. This point is not applied to mass but in general the inputs should be lists for the sale of consistency and effieicny. 

The above points hold for the python implementation as well.

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
```diff
@@ -0,0 +1,38 @@
...
+        for model_part in self.model_parts:
+            KratosOA.ResponseUtils.MassResponseUtils.Check(model_part)
+
+    def CalculateValue(self) -> float:
+        value = 0.0
+        for model_part in self.model_parts:
+            value += KratosOA.ResponseUtils.MassResponseUtils.CalculateValue(model_part)
+        return value
+
+    def CalculateSensitivity(self, sensitivity_variable: SupportedSensitivityFieldVariableTypes, sensitivity_model_part: Kratos.ModelPart) -> None:
```
RezaNajian: The same points as mentioned in the mass_utils. To cover all the scenarios the inputs of the function should be lists.

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
```diff
@@ -0,0 +1,38 @@
+import KratosMultiphysics as Kratos
```
RezaNajian, line 1: @sunethwarna Thank you very much for the updates and effort. From my point of view, I am a little concerned about the consistency in calling the functions of Massutil. Specifically, the function CalculateSensitivity has loops in C++, but the other functions like CalculateValue and Check have loops in Python, and the rest are in C++. Would it make more sense and be more consistent to have these loops in C++ as well? It's up to you; I can live with the current implementation as well.

RezaNajian (approved):
@sunethwarna Thank you very much for all the effort you put into this PR to make it as general as possible. I highly appreciate it and look forward to the next PRs. Very good job and I highly enjoy this collaboration.

sunethwarna:
Thanks @RezaNajian @FMeister :D....

PR #10766, opened 2023-02-17
[OptApp] Restructure 3 - Add linear strain energy response function

sunethwarna:
**📝 Description**
This adds linear strain energy response function

**🆕 Changelog**
- Adds `LinearStrainEnergyResponseFunction`
- Adds finite difference tests


applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
```diff
@@ -0,0 +1,311 @@
...
+    const auto& r_process_info = rModelPart.GetProcessInfo();
+
+    block_for_each(rModelPart.Elements(), tls_type(), [&](auto& rElement, tls_type& rTLS) {
+        if ((!rElement.IsDefined(ACTIVE) || (rElement.Is(ACTIVE))) && rElement.Is(SELECTED)) {
+            Vector& r_u = std::get<0>(rTLS);
+            Vector& r_sensitivity = std::get<1>(rTLS);
+
+            rElement.GetValuesVector(r_u);
+
+            // now calculate perturbed
```
RezaNajian: remove this comment since you are not doing FD

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
```diff
@@ -0,0 +1,311 @@
...
+            rElement.GetProperties().SetValue(rOutputSensitivityVariable, 0.5 * inner_prod(r_u, r_sensitivity));
+        } else {
+            rElement.GetProperties().SetValue(rOutputSensitivityVariable, 0.0);
+        }
+    });
+
+    KRATOS_CATCH("");
+}
+
+void LinearStrainEnergyResponseUtils::CalculateStrainEnergyNonLinearSensitivity(
```
RezaNajian: Nonlinear term is confusing what is actually done is CalculateStrainEnergyFDPeropertySensitivity or ?

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
```diff
@@ -0,0 +1,311 @@
...
+    const Variable<double>& rPrimalVariable,
+    const Variable<double>& rOutputSensitivityVariable)
+{
+    KRATOS_TRY
+
+    using tls_type = std::tuple<Vector, Vector, Vector>;
+
+    const auto& r_process_info = rModelPart.GetProcessInfo();
+
+    block_for_each(rModelPart.Elements(), tls_type(), [&](auto& rElement, tls_type& rTLS) {
```
RezaNajian, line 349: I have a concern here. We have openmp pragma here if all elements share the same property then threads definitely interfere here and the results most probably will not be correct or ?!
I suggest to the same thing we did for shape, i.e., to create a tmp_property, copy the current property into it, and then do FD. what do you say ?

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
```diff
@@ -0,0 +1,311 @@
...
+            const auto domain_size = r_geometry.WorkingSpaceDimension();
+
+            rElement.GetValuesVector(r_u);
+
+            // calculate the reference value
+            rElement.CalculateRightHandSide(r_ref_rhs, r_process_info);
+
+            // initialize dummy element for parallelized perturbation based sensitivity calculation
+            // in each thread seperately
+            if (!p_element) {
```
RezaNajian: Why do we create a new model part to be deleted later? Could you please explain it ?
Cant we simply create a tmp_element which after computation and function call gets deleted ?!

RezaNajian: In this implementation, we are basically copying the model part mesh and deleting it later, which is a big memory overhead. We could create local points and elements to perform the calculation without the need for creating model parts and respectively nodes, etc. Or am I missing something?

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.h
```diff
@@ -0,0 +1,83 @@
+//    |  /           |
```
RezaNajian, line 1: There is a scenario that is not covered in this implementation and in the implementation on the master. What if we are interested in the strain energy of a submodel part of the analysis model part? In that case, we would need to solve for adjoints and so on. Should we cover this now, or will it be implemented later if needed?"

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
```diff
@@ -0,0 +1,47 @@
...
+from KratosMultiphysics.OptimizationApplication.utilities.optimization_info import OptimizationInfo
+from KratosMultiphysics.OptimizationApplication.execution_policies.execution_policy_decorator import ExecutionPolicyDecorator
+from KratosMultiphysics.OptimizationApplication.responses.response_function import ResponseFunction
+from KratosMultiphysics.OptimizationApplication.responses.response_function import SupportedSensitivityFieldVariableTypes
+
+
+class LinearStrainEnergyResponseFunction(ResponseFunction):
+    @classmethod
+    def GetSensitivityFieldVariables(cls) -> 'list[SupportedSensitivityFieldVariableTypes]':
+        return [Kratos.SHAPE_SENSITIVITY, KratosOA.YOUNG_MODULUS_SENSITIVITY, KratosOA.POISSON_RATIO_SENSITIVITY]
```
RezaNajian: what about THICKNESS, we have its implementation on Master, could you please include and cover it ?!

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
```diff
@@ -0,0 +1,47 @@
...
+        default_settings = Kratos.Parameters("""{
+            "evaluated_model_part_names": [
+                "PLEASE_PROVIDE_A_MODEL_PART_NAME"
+            ],
+            "primal_analysis_name"      : "",
+            "perturbation_size"         : 1e-8
+        }""")
+        parameters.ValidateAndAssignDefaults(default_settings)
+
+        self.primal_analysis_execution_policy_wrapper: ExecutionPolicyDecorator = optimization_info.GetOptimizationProcess(ExecutionPolicyDecorator, parameters["primal_analysis_name"].GetString())
```
RezaNajian: "The response should not own the analysis; this is exactly one of the issues or shortcomings of the ShapeOptApp. Analyses are called by the optimizer, and then responses and their sensitivities are calculated. Several responses may share the same analysis."

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
```diff
@@ -0,0 +1,47 @@
...
+
+        if len(self.model_parts) == 0:
+            raise RuntimeError("No model parts were provided for LinearStrainEnergyResponseFunction.")
+
+    def Check(self) -> None:
+        pass
+
+    def CalculateValue(self) -> float:
+        # execute the primal analysis
+        self.primal_analysis_execution_policy_wrapper.Execute()
```
RezaNajian: This should be removed since analysis should not be called and owned by response.

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
```diff
@@ -0,0 +1,47 @@
+import KratosMultiphysics as Kratos
+import KratosMultiphysics.OptimizationApplication as KratosOA
+from KratosMultiphysics.OptimizationApplication.utilities.optimization_info import OptimizationInfo
+from KratosMultiphysics.OptimizationApplication.execution_policies.execution_policy_decorator import ExecutionPolicyDecorator
+from KratosMultiphysics.OptimizationApplication.responses.response_function import ResponseFunction
+from KratosMultiphysics.OptimizationApplication.responses.response_function import SupportedSensitivityFieldVariableTypes
+
```
RezaNajian, line 5: can's we simply import LinearStrainEnergyResponseUtils without using KratosOA.ResponseUtils to save repeating things and space ?!

applications/OptimizationApplication/tests/test_linear_strain_energy_response_function.py
```diff
@@ -0,0 +1,157 @@
+from math import isclose
```
RezaNajian, line 1: "Please rename the 'linear_element_test' folder to something like 'linear_strain_energy_test', and move the test script there to make it cleaner and more structured."

applications/OptimizationApplication/tests/linear_element_test/primal_parameters.json
```diff
@@ -0,0 +1,106 @@
+{
```
RezaNajian: rename the folder to sth like linear_strain_energy_test

RezaNajian (changes requested):
@sunethwarna  "Thank you again for the effort and work. In general, it is fine; however, there are critical issues that need to be resolved and if needed discussed.

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
sunethwarna: I would go with having different properties for different elements (because that is what we are doing with these properties optimizations.) I have created a `OptimizationUtils::CreateEntitySpecificPropertiesForContainer` to do that. So if we do this, there won't be any problems with `OpenMP` parallelization. 

The one issue with temp_properties is, creation and removal includes heave performance hit since we have to do it for all the entities for all the responses and all the different controls. If we create specific properties for specific entities, then all of them can be done easily. 
(This was heavily discussed with Kratos core guys as well).

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
sunethwarna: Actually,  I am using Thread Local Storage (TLS).
```c++
            if (!p_element) {

                std::stringstream name;
                const int thread_id = OpenMPUtils::ThisThread();
                name << rModelPart.Name() << "_temp_" << thread_id;
                GeometryType::PointsArrayType nodes;
                {
                    KRATOS_CRITICAL_SECTION

                    model_part_names.push_back(name.str());
                    auto& tls_model_part = rModelPart.CreateSubModelPart(name.str());

                    for (IndexType i = 0; i < r_geometry.size(); ++i) {
                        nodes.push_back(tls_model_part.CreateNewNode((thread_id + 1) * max_id + i + 1, r_geometry[i]));
                    }
                }

                p_element = rElement.Create(1, nodes, rElement.pGetProperties());
                p_element->Initialize(r_process_info);
                p_element->InitializeSolutionStep(r_process_info);
            }
```
The above code segment is only executed **once per thread, and it only adds 4 nodes for the model part** hence, it is not expensive. I create the model part because, in kratos it is not advised to create `Nodes` on the fly because, the `Node` does not hold the `SolutionStepVariablesList`, it owned by the model part. So if you create `Nodes` without model part, then you may encounter seg-faults which are hard to trace.

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
sunethwarna: In my opinion, optimization should not care about analysis. Assume the following scenario.

What if a response needs to run a single analysis several times (adjoint analysis). This is typical in fluid optimizations because we have adjoint stabilization parameters which needs to be run with different values. And all of them are simulations in Kratos. Then we cannot do that anymore.

How should we enable this if optimizer runs the analysis, and it does not know which parameter to change because, it is only known by response?

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
sunethwarna: I will add..:)

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.h
sunethwarna: Let's implement it when it is required?

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
sunethwarna: Please consider the above comment.

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
sunethwarna: I don't think we can. It's a limitation by python side. I will check :D

applications/OptimizationApplication/tests/test_linear_strain_energy_response_function.py
sunethwarna: Will do :D

applications/OptimizationApplication/tests/linear_element_test/primal_parameters.json
sunethwarna: Will do :D

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
sunethwarna: As agreed, I will remove primal analysis calls from responses. `Analysis` list in optimizer should only have primal analyses. Adjoint analyses should be integrated in to respones. I will remove analysis call frfom response.

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
sunethwarna: answererd and agreed.

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
sunethwarna: removed :)

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
sunethwarna: renamed :D

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
sunethwarna: Added :D

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+    const auto& r_process_info = rModelPart.GetProcessInfo();
+
+    const int max_id = block_for_each<MaxReduction<int>>(rModelPart.GetRootModelPart().Nodes(), [](const auto& rNode) {
+        return rNode.Id();
+    });
+
+    std::vector<std::string> model_part_names;
+
+    block_for_each(rModelPart.Elements(), tls_type(), [&](auto& rElement, tls_type& rTLS) {
```
RezaNajian: Here there are two assumptions, the first is that the primal analysis model part only has only elements. I.e., only element residuals are considered.  However, the primal analysis model part may have conditions that are a function of the sensitivity variable like shape.
The other point is that if the sensitivty model parts are not 100 % covering the analysis model part then the interface elements will not get the full sensitivity. I mean if the analysis model part has elements that are not in the sensitivity model parts and have interface to them, then we do not receive the full sensitivity. Are you aware of a mechanism to identify these element and add their sensitivities, if not and we are not willing to implement it then we must do sensitivity analysis on the full analysis model part.

RezaNajian: Furtheremore, the primal analysis model part should be passed to the analysis-based CalculateSensitivity functions like strain energy.

applications/OptimizationApplication/tests/test_linear_strain_energy_response_function.py
sunethwarna: done :)

applications/OptimizationApplication/tests/linear_element_test/primal_parameters.json
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
sunethwarna: done :)

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
sunethwarna: Ok, I fixed the first point. Now shape sensitivity consideres the condition contributions.

sunethwarna: The second point only happens in the `SHAPE_SENSITIVITY` or, nodes based sensitivity computations ryt? The example is following.

```python
---------- TotalMP -----------
              ---- SenMP -----
1------2------3------7------8
|  a   |  b   |  c   |  d   |
4------5------6------9------10
```

In the above example, `TotalMP` is the full evaluated part. `SenMP` is the sensitivity model part. So the interface nodes will be `3` and `6`. So if we compute the sensitivities on `SenMP`, then we will get contributions from sensitivities of `a` element for them. If the evaluated model part is not `SenMP` and it is `TotalMP`, then this is wrong which is what you are explaining. I wrote it here with mesh so we can come back to this point at a later time. At the moment I don't have a nice solution. But I think also this is important aspect to concider. I will also keep thinking for a nice solution. It should be possible.

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
```diff
@@ -0,0 +1,376 @@
...
+    using tls_type = std::tuple<Matrix, Vector, Vector>;
+
+    const double local_element_value = block_for_each<SumReduction<double>>(rModelPart.Elements(), tls_type(), [&](auto& rElement, tls_type& rTLS) {
+        Matrix& r_lhs = std::get<0>(rTLS);
+        Vector& r_rhs = std::get<1>(rTLS);
+        Vector& r_u = std::get<2>(rTLS);
+        return CalculateEntityStrainEnergy(rElement, r_lhs, r_rhs, r_u, rModelPart.GetProcessInfo());
+    });
+
+    const double local_condition_value = block_for_each<SumReduction<double>>(rModelPart.Conditions(), tls_type(), [&](auto& rCondition, tls_type& rTLS) {
```
RezaNajian, line 78: I am not sure whether conditions should be included in the value evaluation! should not we only consider elements ?!

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
```diff
@@ -0,0 +1,376 @@
...
+        for (auto& r_variable : it.second) {
+            std::visit([&](auto&& r_variable) {
+                if (*r_variable == YOUNG_MODULUS_SENSITIVITY) {
+                    CalculateStrainEnergyLinearlyDependentPropertySensitivity(r_sensitivity_model_part, YOUNG_MODULUS, YOUNG_MODULUS_SENSITIVITY);
+                } else if (*r_variable == THICKNESS_SENSITIVITY) {
+                    CalculateStrainEnergyLinearlyDependentPropertySensitivity(r_sensitivity_model_part, THICKNESS, THICKNESS_SENSITIVITY);
+                } else if (*r_variable == POISSON_RATIO_SENSITIVITY) {
+                    CalculateStrainEnergyFiniteDifferencePropertySensitivity(r_sensitivity_model_part, PerturbationSize, POISSON_RATIO, POISSON_RATIO_SENSITIVITY);
+                } else if (*r_variable == SHAPE_SENSITIVITY) {
+                    CalculateStrainEnergyShapeSensitivity(r_sensitivity_model_part, PerturbationSize, SHAPE_SENSITIVITY);
```
RezaNajian: For the sake of consistency and clearity I would name the function CalculateStrainEnergyFiniteDifferenceShapeSensitivity. Just to be consistent with the property.

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
```diff
@@ -0,0 +1,376 @@
...
+    using tls_condition_type = std::tuple<Vector, Vector, Vector, Condition::Pointer>;
+
+    const int max_id = block_for_each<MaxReduction<int>>(rModelPart.GetRootModelPart().Nodes(), [](const auto& rNode) {
+        return rNode.Id();
+    });
+
+    std::vector<std::string> model_part_names;
+
+    block_for_each(rModelPart.Elements(), tls_element_type(), [&](auto& rElement, tls_element_type& rTLS) {
+        CalculateStrainEnergyEntityShapeSensitivity(
```
RezaNajian: This function computes entities residual shape sensitivities, therefore, I would call it CalculateEntityResidualShapeSensitivty

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
```diff
@@ -0,0 +1,376 @@
...
+        Vector& r_u = std::get<2>(rTLS);
+        return CalculateEntityStrainEnergy(rCondition, r_lhs, r_rhs, r_u, rModelPart.GetProcessInfo());
+    });
+
+    return rModelPart.GetCommunicator().GetDataCommunicator().SumAll(local_element_value + local_condition_value);
+
+    KRATOS_CATCH("");
+}
+
+void LinearStrainEnergyResponseUtils::CalculateSensitivity(
```
RezaNajian, line 90: For the record, this function should also have the analysis model part to compute the full sensitivity of the residuals w.r.t sensitivity model parts. It should be noted that for the partial derivatives, we need to find the overlapping elements of the evaluated and sensitivity model parts, and for the adjoint/residual/physics sensitivity we need to have overlapping and interface elements of the analysis and sensitivity model parts.

applications/OptimizationApplication/tests/test_linear_strain_energy_response_function.py
```diff
@@ -0,0 +1,158 @@
...
+
```
RezaNajian, line 2: Wow, you are really master, thanks Suneth for these validations test !

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
```diff
@@ -0,0 +1,376 @@
...
+                }
+            }, r_variable);
+        }
+    }
+
+    KRATOS_CATCH("");
+}
+
+template<class TEntityType>
+void LinearStrainEnergyResponseUtils::CalculateStrainEnergyEntityShapeSensitivity(
```
RezaNajian: To make it more elaborate, I suggest renaming it to CalculateEntityResidualShapeSensitivty

RezaNajian (dismissed):
Thanks a lot again for the effort. I am sure it will pay off in the long term. It would be great if you could address the minor issues and then merge. Although these are only suggestions.

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
sunethwarna: I changed it to have the semi analytics, since it is not only FD, it is semi analytics. Would you be ok with `CalculateStrainEnergySemiAnalyticShapeSensitivity`?

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
sunethwarna: changed to `CalculateStrainEnergyEntitySemiAnalyticShapeSensitivity`

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
sunethwarna: I will fix it in the new PR which I am working on now.

applications/OptimizationApplication/tests/test_linear_strain_energy_response_function.py
sunethwarna: Thanks :D

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
sunethwarna: changed to `CalculateStrainEnergySemiAnalyticShapeSensitivity`

applications/OptimizationApplication/custom_utilities/response/linear_strain_energy_response_utils.cpp
RezaNajian: @sunethwarna Do we need to evaluate conditions as well? Could you please explain ?

RezaNajian (approved):

sunethwarna:
@RezaNajian In the case, if there are contributions from `Condition` to the `K` matrix, then the matrix used in the linear strain energy computation becomes `K` = `K_elem` + `K_condition`, then `J = u^T(K_elem + K_cond)u` , Hence it is required ryt? I was thinking like this. Is it not usual? :|

RezaNajian:
@sunethwarna Yes, it is true. Then good to merge. Tnx.

PR #10767, opened 2023-02-17
[OptApp] Restructure 4 - Add model part controller

sunethwarna:
**📝 Description**
This PR adds a model part controller for optimization analysis. A basic reading (OpenMP) mechanism is added.

**🆕 Changelog**
- Adds model part controller
- Adds tests


applications/OptimizationApplication/python_scripts/model_part_controllers/mdpa_model_part_controller.py
```diff
@@ -0,0 +1,41 @@
+import KratosMultiphysics as Kratos
+from KratosMultiphysics.OptimizationApplication.model_part_controllers.model_part_controller import ModelPartController
+from KratosMultiphysics.OptimizationApplication.utilities.optimization_info import OptimizationInfo
+
+class MdpaModelPartController(ModelPartController):
+    def __init__(self, model: Kratos.Model, parameters: Kratos.Parameters, optimization_info: OptimizationInfo):
+        default_settings = Kratos.Parameters("""{
+            "model_part_name": "",
+            "input_filename" : "",
+            "domain_size"    : -1,
```
Igarizza, line 9: why default is -1?

applications/OptimizationApplication/tests/test_model_part_controllers.py
```diff
@@ -0,0 +1,31 @@
...
+from KratosMultiphysics.OptimizationApplication.model_part_controllers.mdpa_model_part_controller import MdpaModelPartController
+
+@kratos_unittest.skipIfApplicationsNotAvailable("StructuralMechanicsApplication")
+class TestMdpaModelPartController(kratos_unittest.TestCase):
+    def test_MdpaModelPartController(self):
+        with kratos_unittest.WorkFolderScope("linear_element_test", __file__):
+            model_part_controller_settings = Kratos.Parameters("""{
+                "model_part_name": "Structure",
+                "input_filename" : "Structure",
+                "domain_size"    : 3
```
Igarizza, line 42: Can you add a test to check both scenarios: read_data: true and false?

Igarizza (commented):
In general all looks good from my side.

applications/OptimizationApplication/python_scripts/model_part_controllers/mdpa_model_part_controller.py
FMeister: How is it handled else where? Would make sense to handle it the same way.

sunethwarna: In this case, if the user forgets to give the `domain_size`, it will throw an error since default value of -1 is not accepted. It is to avoid not having the `domain_size` from the user. 

@FMeister This is handled in the same manner in other places as well. (Not all... :D)

RezaNajian (changes requested):
thanks for the works. Please make use of the implemeneted modelpart controller class in the model_parts_controller.py. It should not harm anything or ?!

applications/OptimizationApplication/tests/test_model_part_controllers.py
sunethwarna: done :)

sunethwarna:
@RezaNajian @Igarizza This is done. Can you check?

RezaNajian (approved):
Thanks, Suneth. Please make sure that tests are passing before merging. Tnx.

sunethwarna:
Thanks :D...  I checked, and they are passing @RezaNajian .

PR #10768, opened 2023-02-17
[OptApp] Restructure 5 - Add container data

sunethwarna:
**📝 Description**
This adds a class named `ContainerVariableDataHolder`. It is added due to following reasons.

1. `OptimizationApplication` needs to have sensitivities computed on 6 containers namely: `historical nodes`, `non-historical nodes`, `conditions`, `condition properties`, `elements` and `element properties`. This container enables retrieving/assigning data from/to each respective data container easily.
2. `OptimizationApplication` requires data transfer between `nodal historical <-> nodal non-historical`, `element <-> element properties`, `condition <-> condition properties`. This is made easier by using  this `ContainerVariableDataHolder` class
3. `OptimizationApplication` requires data transfer between `nodes <-> elements`, `nodes <-> conditions` which is enabled by the use of `ContainverVariableDataHolderUtils`
4. `OptimizationApplication` requires sensitivities for each response function stored in all 6 different containers. That means, if there are `m` response functions and `n` derivative variables, then it will require `mxn` variables such as `MASS_SHAPE_SENSITIVITY`, `MASS_DENSITY_SENSITIVITY`, `STRAIN_ENERGY_SENSITIVITY`, `STRAIN_ENERGY_DENSITY_SENSITIVITY`, etc so not to have one sensitivity computation overwriting another sensitivity computation. This is hard to manage. Therefore, these introdudced containers allows to use same variable for every response function. Example is, the same `DENSITY_SENSITIVITY` variable can be used to store sensitivities for any number of responses without one response  overwriting other responses sensitivities.
5. All of these introduced contaienrs are `OpenMP` and `MPI` parallelized. Which makes the use of them in python super simple.
6. They have most of the operators defined.
7. Since it copies the data from its respective data containers variables to a flattened vector, it does not use additional space.
8. These containers can be used to store and post process data using vtk/hdf5/StatApp easily. It is introduced in a later PR.

**🆕 Changelog**
- Adds `ContainerVariableDataHolder`
- Adds `ContainerVariableDataHolderUtils`
- Adds tests


applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder_base.h
```diff
@@ -0,0 +1,194 @@
...
+    static void SetValue(
+        ModelPart::NodeType& rNode,
+        const Variable<TDataType>& rVariable,
+        const TDataType& rValue)
+    {
+        rNode.FastGetSolutionStepValue(rVariable) = rValue;
+    }
+
+    ///@}
+};
```
matekelemen: Getting and setting data from/to different sources/destinations like this is always painful, precisely because the names are different. Here, you provide a uniform naming for the functions (`SetValue`, `GetValue`) but still give different names to the classes, so the same problem is propagated.

I'd suggest putting a stop to naming historical and nonhistorical data operations, and make the distinction in the template arguments instead, by introducing tags for the different data types. For example:

```cpp
namespace Tags {
    struct Historical {};
    struct NonHistorical {};
} // namespace Tags

// Dummy generic class to be specialized later
template <class TTag>
struct ContainerDataIO {};

template <>
struct ContainerDataIO<Tags::Historical>
{
    template <class TData>
    static const TData& GetValue(const ModelPart::NodeType& rNode,
                                 const Variable<TData>& rVariable)
    {
        return rNode.FastGetSolutionStepValue(rVariable);
    }

    ...
}; // struct ContainerDataIO<Tags::Historical>

template <>
struct ContainerDataIO<Tags::NonHistorical>
{
    template<class TData, class TEntity>
    static const TData& GetValue(
        const TEntity& rEntity,
        const Variable<TData>& rVariable)
    {
        return rEntity.GetValue(rVariable);
    }

    ...
}; // struct ContainerDataIO<Tags::NonHistorical>
```

It'll be more template-friendly later on if you don't have to worry about different names.

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder_base.h
```diff
@@ -0,0 +1,194 @@
...
+
+class NonHistoricalContainerDataIO
+{
+public:
+    ///@name Public operations
+    ///@{
+
+    template<class TEntityType, class TDataType>
```
matekelemen: I'd change the order of the template parameters to make it more consistent with the historical IO.
```diff
- template<class TEntityType, class TDataType>
+ template<class TDataType, class TEntityType>
```

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder_base.h
```diff
@@ -0,0 +1,194 @@
...
+    ///@}
+};
+
+class PropertiesContainerDataIO
+{
+public:
+    ///@name Public operations
+    ///@{
+
+    template<class TEntityType, class TDataType>
```
matekelemen: template param order

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder_base.h
```diff
@@ -0,0 +1,194 @@
...
+
+class NonHistoricalContainerDataIO
+{
+public:
+    ///@name Public operations
+    ///@{
+
+    template<class TEntityType, class TDataType>
+    static TDataType& GetValue(
+        TEntityType& rEntity,
```
matekelemen: why is the getter non-const? If there's a separate `SetValue` member, I'd expect that reading can be performed on an immutable object, and wouldn't expect the object to return a mutable reference from a getter. To me it's counterintuitive and makes `SetValue` pointless.

If you need to get mutable references, I'd get rid of `SetValue` and add an immutable `GetValue`.

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder_base.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+// Include base h
+#include "container_variable_data_holder_base.h"
+
+namespace Kratos {
+
+template <class TContainerType>
+ContainerVariableDataHolderBase<TContainerType>::ContainerVariableDataHolderBase(ModelPart& rModelPart)
+    : mDataDimension(0),
+      mpModelPart(&rModelPart)
```
matekelemen:
```suggestion
    : mData(),
      mDataDimension(0),
      mpModelPart(&rModelPart)
```

if you're going for direct initializations, you need to explicitly initialize **all** members in the **exact order** they were defined in the class.

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder_base.cpp
```diff
@@ -0,0 +1,180 @@
...
+{
+}
+
+template <class TContainerType>
+ContainerVariableDataHolderBase<TContainerType>::ContainerVariableDataHolderBase(
+    const ContainerVariableDataHolderBase& rOther)
+    : mDataDimension(rOther.mDataDimension),
+      mpModelPart(rOther.mpModelPart)
```
matekelemen: initialization order

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.h
```diff
@@ -0,0 +1,157 @@
...
+    ContainerVariableDataHolder(ModelPart& rModelPart)
+        : BaseType(rModelPart)
+    {
+    }
+
+    /// Copy constructor
+    ContainerVariableDataHolder(const ContainerVariableDataHolder& rOther)
+        : BaseType(rOther)
+    {
+    }
```
matekelemen: This is unnecessary. The other constructor taking an instance of the baseclass does the same thing but is more general (+ it's safe in this case because the derived class doesn't have any extra member variables).

Also, I'd add a `noexcept` move constructor.

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.h
```diff
@@ -0,0 +1,157 @@
...
+
+    /// Copy constructor with base class used to transfer data between compatible data containers
+    ContainerVariableDataHolder(const BaseType& rOther)
+        : BaseType(rOther)
+    {
+    }
+
+    ~ContainerVariableDataHolder() override = default;
```
matekelemen: unnecessary

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.h
```diff
@@ -0,0 +1,157 @@
...
+
+    ContainerVariableDataHolder operator^(const ContainerVariableDataHolder& rOther) const;
+
+    ContainerVariableDataHolder& operator^=(const ContainerVariableDataHolder& rOther);
+
+    ContainerVariableDataHolder operator^(const double Value) const;
+
+    ContainerVariableDataHolder& operator^=(const double Value);
+
+    ContainerVariableDataHolder& operator=(const ContainerVariableDataHolder& rOther);
```
matekelemen: I'd put this somewhere near the constructors instead of the operators.

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.cpp
```diff
@@ -0,0 +1,657 @@
...
+IndexType GetLocalSize<array_1d<double, 3>>(const IndexType DomainSize)
+{
+    return DomainSize;
+}
+
+void AssignValueToVector(
+    Vector& rOutput,
+    const IndexType StartIndex,
+    const IndexType ComponentIndex,
+    const double& rValue)
```
matekelemen:
```suggestion
    double rValue)
```

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.cpp
```diff
@@ -0,0 +1,657 @@
...
+}
+
+void AssignValueFromVector(
+    Array3D& rOutput,
+    const IndexType ValueComponentIndex,
+    const IndexType VectoStartingIndex,
+    const Vector& rInput)
+{
+    rOutput[ValueComponentIndex] = rInput[VectoStartingIndex + ValueComponentIndex];
+}
```
matekelemen:
```suggestion
void AssignValueFromVector(
    Array3D& rOutput,
    const IndexType ValueComponentIndex,
    const IndexType VectorStartingIndex,
    const Vector& rInput)
{
    rOutput[ValueComponentIndex] = rInput[VectorStartingIndex + ValueComponentIndex];
}
```

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder_base.h
```diff
@@ -0,0 +1,194 @@
...
+        const TDataType& rValue)
+    {
+        rEntity.GetProperties().SetValue(rVariable, rValue);
+    }
+
+    ///@}
+};
+
+template <class TContainerType>
+class KRATOS_API(OPTIMIZATION_APPLICATION) ContainerVariableDataHolderBase {
```
matekelemen: missing docs

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.cpp
```diff
@@ -0,0 +1,657 @@
...
+
+    const IndexType number_of_entities = this->GetContainer().size();
+    this->mDataDimension = ContainerVariableDataHolderHelperUtilities::GetLocalSize<TDataType>(this->GetModelPart().GetProcessInfo()[DOMAIN_SIZE]);
+
+    auto& r_data = this->mData;
+    auto& r_container = this->GetContainer();
+
+    r_data.resize(number_of_entities * this->mDataDimension);
+
+    IndexPartition<IndexType>(number_of_entities).for_each([&](const IndexType Index){
```
matekelemen: it's usually good practice to explicitly state what you're capturing in a lambda instead of capturing everything by ref with `[&]`.
(for example, `number_of_entities` should be captured by value)

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.cpp
```diff
@@ -0,0 +1,657 @@
...
+        });
+    } else {
+        // no ghost settings required for elements or conditions nor historical values, hence no special treatment required.
+        block_for_each(r_container, [&](auto& rEntity) {
+            TContainerDataIO::SetValue(rEntity, rVariable, rVariable.Zero());
+        });
+    }
+
+    IndexPartition<IndexType>(number_of_entities).for_each([&](const IndexType Index){
+        auto& values = TContainerDataIO::GetValue(*(r_container.begin() + Index), rVariable);
```
matekelemen: naming

there's `r_values` and there's `values`, both of which are references and aren't too descriptive. How about `r_source_values` and `r_target_values`?

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.cpp
```diff
@@ -0,0 +1,657 @@
...
+    KRATOS_ERROR_IF_NOT(this->GetModelPart().GetProcessInfo().Has(DOMAIN_SIZE))
+        << "DOMAIN_SIZE variable is not found in process info of model part "
+           "with name "
+        << this->GetModelPart().FullName() << ".\n";
+
+    this->mDataDimension = ContainerVariableDataHolderHelperUtilities::GetLocalSize<TDataType>(this->GetModelPart().GetProcessInfo()[DOMAIN_SIZE]);
+
+    const IndexType number_of_entities = this->GetContainer().size();
+
+    if (this->mData.size() != this->mDataDimension * number_of_entities) {
```
matekelemen: I'm pretty sure resizing already checks this internally and doesn't do anything if the size is already correct.

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.cpp
```diff
@@ -0,0 +1,657 @@
...
+    KRATOS_ERROR_IF_NOT(this->mData.size() == rOther.mData.size())
+        << "Data size mismatch in operands for +.\n"
+        << "      Left operand data : " << *this << "\n"
+        << "      Right operand data: " << rOther << "\n";
+
+    ContainerVariableDataHolder<TContainerType, TContainerDataIO> result(*(this->mpModelPart));
+    result.mDataDimension = this->mDataDimension;
+    result.mData.resize(this->mData.size(), false);
+    IndexPartition<IndexType>(this->mData.size()).for_each([&](const IndexType Index) {
+        result.mData[Index] = this->mData[Index] + rOther.mData[Index];
```
matekelemen: Performancewise, I don't think you should implement operators on your own. For one, if you need multiple operations done on your array, you'll force multiple passes through the entire array. Linalg libs usually put a great deal of effort into making sure that these kinds of operations are efficient and make the best use of the available hardware.

I don't know about boost's ublas, but for example, with Eigen you could construct a wrapper (`Eigen::Map`) around your array that would enable you to use Eigen's linalg functions.

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.cpp
```diff
@@ -0,0 +1,657 @@
...
+    KRATOS_ERROR_IF_NOT(this->mData.size() == rOther.mData.size())
+        << "Data size mismatch in operands for -.\n"
+        << "      Left operand data : " << *this << "\n"
+        << "      Right operand data: " << rOther << "\n";
+
+    ContainerVariableDataHolder<TContainerType, TContainerDataIO> result(*(this->mpModelPart));
+    result.mDataDimension = this->mDataDimension;
+    result.mData.resize(this->mData.size(), false);
+    IndexPartition<IndexType>(this->mData.size()).for_each([&](const IndexType Index) {
+        result.mData[Index] = std::pow(this->mData[Index], rOther.mData[Index]);
```
matekelemen: oh hell no :D
`operator^` indicates binary XOR in C++

If you need this in python, implement it in the python bindings so that it's not visible in C++.

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.cpp
```diff
@@ -0,0 +1,657 @@
...
+    KRATOS_ERROR_IF(std::abs(Value) < std::numeric_limits<double>::epsilon())
+        << "Division by zero.\n"
+        << "      Left operand data : " << *this << "\n"
+        << "      Divisor           : " << Value << "\n";
+
+    ContainerVariableDataHolder<TContainerType, TContainerDataIO> result(*(this->mpModelPart));
+    result.mDataDimension = this->mDataDimension;
+    result.mData.resize(this->mData.size(), false);
+    IndexPartition<IndexType>(this->mData.size()).for_each([&](const IndexType Index) {
+        result.mData[Index] = this->mData[Index] / Value;
```
matekelemen: I'll reiterate that I don't think you should implement these operations yourself.

As an extra remark, division is several times slower than multiplication, and the compiler won't help you here because `Value` is not a literal. I'd compute the inverse of `Value` outside the loop, and multiply with that inverse inside.

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.cpp
```diff
@@ -0,0 +1,657 @@
...
+        msg << "Properties";
+    }
+
+    msg << BaseType::Info();
+
+    return msg.str();
+}
+
+//template instantiations
+#define INSTANTIATIE_CONTAINER_DATA_METHODS(ContainerType, ContainerDataIOType, DataType)                                                                   \
```
matekelemen: I know these macro names are already long, but I still recommend slapping `KRATOS_` in the front to avoid any surprising macro name clashes.

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
```diff
@@ -14,22 +14,89 @@
...
+        .def(py::self *= float())
+        .def(py::self /  py::self)
+        .def(py::self /= py::self)
+        .def(py::self /  float())
+        .def(py::self /= float())
+        .def("__pow__", py::overload_cast<const container_type&>(&container_type::operator^, py::const_))
+        .def("__ipow__", py::overload_cast<const container_type&>(&container_type::operator^=))
+        .def("__pow__", py::overload_cast<const double>(&container_type::operator^, py::const_))
+        .def("__ipow__", py::overload_cast<const double>(&container_type::operator^=))
+        .def("__neg__", [](container_type& rSelf) { return rSelf.operator*(-1.0); })
```
matekelemen: I don't like exposing these primitive operations on enormous arrays to python. It'll have terrible performance if you'll try chaining operations because there's no expression template system in python and you'll need to pass through the entire array for each operation.

I'd rather go for some system that represents the sequence of operations, and can evaluate them all at once (like expression templates in C++).

applications/OptimizationApplication/custom_utilities/container_variable_data_holder_utils.cpp
```diff
@@ -0,0 +1,569 @@
...
+        for (IndexType d = 0; d < data_dimension; ++d) {
+            auto& r_value = r_output_data[i * data_dimension + d];
+            for (IndexType j = 0; j < rMatrix.size2(); ++j) {
+                r_value += rMatrix(i, j) * r_input_data[j * data_dimension + d];
+            }
+        }
+    });
+}
+
+void ContainerVariableDataHolderUtils::Transpose(
```
matekelemen: you'll need to check the pointers to prevent aliasing.

applications/OptimizationApplication/tests/test_OptimizationApplication.py
```diff
@@ -23,7 +23,14 @@
 from test_mass_response_function import TestMassResponseFunctionShells
 from test_mass_response_function import TestMassResponseFunctionSolids
 from test_linear_strain_energy_response_function import TestLinearStrainEnergyResponseFunction
+from test_container_variable_data_holder import TestHistoricalContainerVariableDataHolder
```
RezaNajian: could you please make a single import and test call in the main test_OptimizationApplication.py for container_variables_data_holder, then inside it expand it to element,node, ... tests. Here we gonna add tests of major components and then inside each ciomponent we can test sub components and functions

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder_base.h
sunethwarna: done :)

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
sunethwarna: done :)

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.cpp
sunethwarna: done :)

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.cpp
sunethwarna: done :)

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.cpp
sunethwarna: done with vtables

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder_base.h
sunethwarna: moved to a different file

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder_base.h
sunethwarna: moved to a different file

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.h
sunethwarna: removed

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.h
sunethwarna: removed

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.h
sunethwarna: moved

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.cpp
sunethwarna: done :)

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.cpp
sunethwarna: done :) moved to `Pow` function

applications/OptimizationApplication/custom_utilities/container_variable_data_holder_utils.cpp
sunethwarna: done :)

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder_base.h
sunethwarna: done :)

philbucher:
FYI conceptually this seems to be very similar to the [CouplingInterfaceData](https://github.com/KratosMultiphysics/Kratos/blob/master/applications/CoSimulationApplication/python_scripts/coupling_interface_data.py)

FYI @matekelemen, maybe this could be combined in the future by moving this to the Core

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
```diff
@@ -79,22 +151,64 @@ void  AddCustomUtilitiesToPython(pybind11::module& m)
...
-        .def_static("AreAllEntitiesOfSameGeometryType", [](ModelPart::ElementsContainerType& rContainer, const DataCommunicator& rDataCommunicator) { return OptimizationUtils::GetContainerEntityGeometryType(rContainer, rDataCommunicator) != GeometryData::KratosGeometryType::Kratos_generic_type; } )
-        .def_static("CreateEntitySpecificPropertiesForContainer", &OptimizationUtils::CreateEntitySpecificPropertiesForContainer<ModelPart::ConditionsContainerType>)
-        .def_static("CreateEntitySpecificPropertiesForContainer", &OptimizationUtils::CreateEntitySpecificPropertiesForContainer<ModelPart::ElementsContainerType>)
-        .def_static("GetVariableDimension", &OptimizationUtils::GetVariableDimension<double>)
-        .def_static("GetVariableDimension", &OptimizationUtils::GetVariableDimension<array_1d<double, 3>>)
-        .def_static("CopySolutionStepVariablesList", &OptimizationUtils::CopySolutionStepVariablesList)
+    m.def_submodule("OptimizationUtils")
+        .def("IsVariableExistsInAllContainerProperties", &OptimizationUtils::IsVariableExistsInAllContainerProperties<ModelPart::ConditionsContainerType, double>)
```
matekelemen, line 95: less verbose names

```suggestion
        .def("IsVariableInAllContainerProperties", &OptimizationUtils::IsVariableInAllContainerProperties<ModelPart::ConditionsContainerType, double>)
```

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
```diff
@@ -79,22 +151,64 @@ void  AddCustomUtilitiesToPython(pybind11::module& m)
...
+        .def("IsVariableExistsInAtLeastOneContainerProperties", &OptimizationUtils::IsVariableExistsInAtLeastOneContainerProperties<ModelPart::ConditionsContainerType, array_1d<double, 3>>)
+        .def("IsVariableExistsInAtLeastOneContainerProperties", &OptimizationUtils::IsVariableExistsInAtLeastOneContainerProperties<ModelPart::ElementsContainerType,array_1d<double, 3>>)
+        .def("AreAllEntitiesOfSameGeometryType", [](ModelPart::ConditionsContainerType& rContainer, const DataCommunicator& rDataCommunicator) { return OptimizationUtils::GetContainerEntityGeometryType(rContainer, rDataCommunicator) != GeometryData::KratosGeometryType::Kratos_generic_type; } )
+        .def("AreAllEntitiesOfSameGeometryType", [](ModelPart::ElementsContainerType& rContainer, const DataCommunicator& rDataCommunicator) { return OptimizationUtils::GetContainerEntityGeometryType(rContainer, rDataCommunicator) != GeometryData::KratosGeometryType::Kratos_generic_type; } )
+        .def("CreateEntitySpecificPropertiesForContainer", &OptimizationUtils::CreateEntitySpecificPropertiesForContainer<ModelPart::ConditionsContainerType>)
```
matekelemen, line 105: less verbose names

```suggestion
        .def("CreateEntityProperties", &OptimizationUtils::CreateEntityProperties<ModelPart::ConditionsContainerType>)
```

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
```diff
@@ -15,24 +15,96 @@
...
+        .def(py::init<const container_type&>(), py::arg("other_container_data_to_copy_from"), py::doc("Creates a new same type container data object by copying data from other_container_data_to_copy_from."))
+        .def(py::init<const typename container_type::BaseType&>(), py::arg("other_container_data_to_copy_from"), py::doc("Creates a new destination type container data object by copying data from compatible other_container_data_to_copy_from."))
+        .def("AssignDataToContainerVariable", &container_type::template AssignDataToContainerVariable<double>, py::arg("scalar_variable"))
+        .def("AssignDataToContainerVariable", &container_type::template AssignDataToContainerVariable<array_1d<double, 3>>, py::arg("Array3_variable"))
+        .def("ReadDataFromContainerVariable", &container_type::template ReadDataFromContainerVariable<double>, py::arg("scalar_variable"))
+        .def("ReadDataFromContainerVariable", &container_type::template ReadDataFromContainerVariable<array_1d<double, 3>>, py::arg("Array3_variable"))
+        .def("SetDataToValue", &container_type::template SetDataToValue<double>, py::arg("scalar_value"))
```
matekelemen: less verbose names

```suggestion
        .def("SetData", &container_type::template SetData<double>, py::arg("scalar_value"))
```

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
```diff
@@ -15,24 +15,96 @@
...
+        .def("SetDataToValue", &container_type::template SetDataToValue<array_1d<double, 3>>, py::arg("Array3_value"))
+        .def("SetDataToVariableZeroValue", &container_type::template SetDataToVariableZeroValue<double>, py::arg("scalar_variable"))
```
matekelemen: less verbose names

```suggestion
        .def("SetZero", &container_type::template SetZero<double>, py::arg("scalar_variable"))
```

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
```diff
@@ -15,24 +15,96 @@
...
+    py::class_<container_type, typename container_type::Pointer, ContainerVariableData<TContainerType>>(m, rName.c_str())
+        .def(py::init<ModelPart&>(), py::arg("model_part"), py::doc("Creates a new container data object with model_part."))
+        .def(py::init<const container_type&>(), py::arg("other_container_data_to_copy_from"), py::doc("Creates a new same type container data object by copying data from other_container_data_to_copy_from."))
+        .def(py::init<const typename container_type::BaseType&>(), py::arg("other_container_data_to_copy_from"), py::doc("Creates a new destination type container data object by copying data from compatible other_container_data_to_copy_from."))
+        .def("AssignDataToContainerVariable", &container_type::template AssignDataToContainerVariable<double>, py::arg("scalar_variable"))
+        .def("AssignDataToContainerVariable", &container_type::template AssignDataToContainerVariable<array_1d<double, 3>>, py::arg("Array3_variable"))
+        .def("ReadDataFromContainerVariable", &container_type::template ReadDataFromContainerVariable<double>, py::arg("scalar_variable"))
```
matekelemen: less verbose names

```suggestion
        .def("Read", &container_type::template Read<double>, py::arg("scalar_variable"))
```

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
```diff
@@ -15,24 +15,96 @@
...
+void AddSpecializedContainerVariableDataToPython(pybind11::module& m, const std::string& rName)
+{
+    namespace py = pybind11;
+
+    using container_type = SpecializedContainerVariableData<TContainerType, ContainerDataIO<TContainerDataIOTag>>;
+    py::class_<container_type, typename container_type::Pointer, ContainerVariableData<TContainerType>>(m, rName.c_str())
+        .def(py::init<ModelPart&>(), py::arg("model_part"), py::doc("Creates a new container data object with model_part."))
+        .def(py::init<const container_type&>(), py::arg("other_container_data_to_copy_from"), py::doc("Creates a new same type container data object by copying data from other_container_data_to_copy_from."))
+        .def(py::init<const typename container_type::BaseType&>(), py::arg("other_container_data_to_copy_from"), py::doc("Creates a new destination type container data object by copying data from compatible other_container_data_to_copy_from."))
+        .def("AssignDataToContainerVariable", &container_type::template AssignDataToContainerVariable<double>, py::arg("scalar_variable"))
```
matekelemen: less verbose names

```suggestion
        .def("Assign", &container_type::template Assign<double>, py::arg("scalar_variable"))
```

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
```diff
@@ -79,22 +151,64 @@ void  AddCustomUtilitiesToPython(pybind11::module& m)
...
+        .def("CopySolutionStepVariablesList", &OptimizationUtils::CopySolutionStepVariablesList)
+        ;
+
+    AddContainerVariableDataToPython<ModelPart::NodesContainerType>(m, "NodalVariableData");
+    AddContainerVariableDataToPython<ModelPart::ConditionsContainerType>(m, "ConditionVariableData");
+    AddContainerVariableDataToPython<ModelPart::ElementsContainerType>(m, "ElementVariableData");
+
+    AddSpecializedContainerVariableDataToPython<ModelPart::NodesContainerType, ContainerDataIOTags::Historical>(m, "HistoricalVariableData");
+    AddSpecializedContainerVariableDataToPython<ModelPart::NodesContainerType, ContainerDataIOTags::NonHistorical>(m, "NodalNonHistoricalVariableData");
+    AddSpecializedContainerVariableDataToPython<ModelPart::ConditionsContainerType, ContainerDataIOTags::NonHistorical>(m, "ConditionNonHistoricalVariableData");
```
matekelemen: less verbose names

`Element`s and `Condition`s don't have historical data, so the distinction is unnecessary

```suggestion
    AddSpecializedContainerVariableDataToPython<ModelPart::ConditionsContainerType, ContainerDataIOTags::NonHistorical>(m, "ConditionVariableData");
```

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
```diff
@@ -79,22 +151,64 @@ void  AddCustomUtilitiesToPython(pybind11::module& m)
...
+        .def("Transpose", py::overload_cast<SparseMatrixType&,const SparseMatrixType&>(&ContainerVariableDataUtils::Transpose), py::arg("output_matrix"), py::arg("input_matrix"))
+        .def("Transpose", py::overload_cast<Matrix&,const Matrix&>(&ContainerVariableDataUtils::Transpose), py::arg("output_matrix"), py::arg("input_matrix"))
+        .def("ComputeNumberOfNeighbourConditions", &ContainerVariableDataUtils::ComputeNumberOfNeighbourEntities<ModelPart::ConditionsContainerType>, py::arg("output_nodal_container_data"))
+        .def("ComputeNumberOfNeighbourElements", &ContainerVariableDataUtils::ComputeNumberOfNeighbourEntities<ModelPart::ElementsContainerType>, py::arg("output_nodal_container_data"))
+        .def("MapContainerVariableDataToNodalVariableData", &ContainerVariableDataUtils::MapContainerVariableDataToNodalVariableData<ModelPart::ConditionsContainerType>, py::arg("output_nodal_container_data"), py::arg("input_container_data_to_map"), py::arg("nodal_container_data_with_neighbour_entities_for_input_container_data"))
```
matekelemen: `nodal_container_data_with_neighbour_entities_for_input_container_data`

Honestly, I'm lost on this name.

matekelemen: also, I think you can omit `Data` from most of the names. It should only be there when you **need** to make a clear distinction between `Varaible` and its underlying `Data` for example.

```diff
- MapContainerVariableDataToNodalVariableData
+ MapContainerVariableToNodalVariable
```

applications/OptimizationApplication/custom_utilities/container_variable_data/container_data_io.h
```diff
@@ -0,0 +1,103 @@
...
+
+    template <class TDataType>
+    static void SetValue(
+        ModelPart::NodeType& rNode,
+        const Variable<TDataType>& rVariable,
+        const TDataType& rValue)
+    {
+        rNode.FastGetSolutionStepValue(rVariable) = rValue;
+    }
+};
```
matekelemen: kool!

applications/OptimizationApplication/custom_utilities/container_variable_data/container_data_io.h
```diff
@@ -0,0 +1,103 @@
...
+
+template <>
+struct ContainerDataIO<ContainerDataIOTags::Properties>
+{
+    template<class TDataType, class TEntityType>
+    static const TDataType& GetValue(
+        const TEntityType& rEntity,
+        const Variable<TDataType>& rVariable)
+    {
+        static_assert(!(std::is_same_v<TEntityType, ModelPart::NodeType>), "Properties retrieval is only supported for element and conditions.");
```
matekelemen: nice

applications/OptimizationApplication/custom_utilities/container_variable_data/container_variable_data.h
```diff
@@ -0,0 +1,241 @@
...
+
+namespace Kratos {
+
+///@name Kratos Classes
+///@{
+
+/**
+ * @brief Container variable data holder
+ *
+ * Objects of this class is used to hold any type of data in a container
```
matekelemen:
```suggestion
 * Instances of this class are used to hold any type of data in a container
```

applications/OptimizationApplication/custom_utilities/container_variable_data/container_variable_data.h
```diff
@@ -0,0 +1,241 @@
...
+ * of TContainerType. TContainerType can be nodes, conditions or elements.
```
matekelemen:
```suggestion
 * of TContainerType. TContainerType can be a container of nodes, conditions or elements.
```

Also, template params are usually described in `@tparam` annotation blocks. For example:
```cpp
@tparam A container of nodes, conditions or elements.
```

applications/OptimizationApplication/custom_utilities/container_variable_data/container_variable_data.h
```diff
@@ -0,0 +1,241 @@
...
+     * @brief Copies the data from another same type container variable data.
+     *
+     * This method is used to copy data from another variable data. The model
+     * parts should be matching to successfully copy data. This does not
+     * copy the data, it only copies the expression pointer, hence this
+     * operation is also a light weight operation.
+     *
+     * @param rOther        Other container variable data
+     */
+    void CopyDataFrom(const ContainerVariableData<TContainerType>& rOther);
```
matekelemen: sooo it's a copy assignment operator? Also, you can omit the template parameter within the class definition.
```cpp
ContainerVariableData& operator=(const ContainerVariableData& rOther);
```

applications/OptimizationApplication/custom_utilities/container_variable_data/container_variable_data.cpp
```diff
@@ -0,0 +1,166 @@
...
+template <class TContainerType>
+void ContainerVariableData<TContainerType>::CopyDataFrom(
+    const ContainerVariableData<TContainerType>& rOther)
+{
+    KRATOS_ERROR_IF(this->mpModelPart != &rOther.GetModelPart())
+        << "Model part mismatch. [ Destination model part name: "
+        << this->mpModelPart->FullName()
+        << ", origin model part name: " << rOther.GetModelPart().FullName() << " ].\n";
+
+    mpExpression = rOther.mpExpression;
```
matekelemen: What happens when the expression behind `rOther.mpExpression` changes? Is it correct behaviour that `this->mpExpression` also changes, or should a deep copy be performed here instead?

applications/OptimizationApplication/custom_utilities/container_variable_data/container_variable_data.cpp
```diff
@@ -0,0 +1,166 @@
...
+}
+
+template <class TContainerType>
+const Expression& ContainerVariableData<TContainerType>::GetExpression() const
+{
+    if (mpExpression) {
+        return *mpExpression;
+    } else {
+        KRATOS_ERROR << "Uninitialized expression.\n";
+    }
```
matekelemen: this is exactly what `std::optional` is for. Consider changing the type of `mpExpression` to `std::optional<Expression*>`.

applications/OptimizationApplication/custom_utilities/container_variable_data/expressions.h
```diff
@@ -0,0 +1,438 @@
...
+
+// Application includes
+
+namespace Kratos {
+
+///@name Kratos Classes
+///@{
+
+/**
+ * @brief Base class or all the expression types.
```
matekelemen:
```suggestion
 * @brief Base class for all the expression types.
```

applications/OptimizationApplication/custom_utilities/container_variable_data/expressions.h
```diff
@@ -0,0 +1,438 @@
...
+    virtual IndexType GetDimension() const = 0;
+
+    ///@}
+
+private:
+    ///@name Private operations
+    ///@{
+
+    //*********************************************
+    // this block is needed for refcounting in th eintrusive ptr
```
matekelemen:
```suggestion
    // this block is needed for refcounting in @ref intrusive_ptr
```

applications/OptimizationApplication/custom_utilities/container_variable_data/expressions.h
```diff
@@ -0,0 +1,438 @@
...
+    //*********************************************
+
+    ///@}
+};
+
+/**
+ * @brief Expression to hold a double value
+ *
+ */
+class LiteralDoubleExpression : public Expression {
```
matekelemen: How about templating it on the value type?
```cpp
template <class TValue>
class LiteralExpression : public Expression {
```
Then `LiteralDoubleExpression` would be `LiteralExpression<double>`, ...

applications/OptimizationApplication/custom_utilities/container_variable_data/expressions.h
```diff
@@ -0,0 +1,438 @@
...
+    const IndexType mDimension;
+
+    ///@}
+};
+
+/**
+ * @brief Expression to add two given expressions.
+ *
+ */
+class BinaryAddExpression : public Expression {
```
matekelemen: Binary ops could have a common base class that takes care of construction and storing the member vars.
```cpp
class BinaryExpression : public Expression
{
public:
    BinaryExpresion(Expression::Pointer pLeft,
                    Expression::Pointer pRight)
        : mpLeft(std::move(pLeft)),
          mpRight(std::move(pRight))
    {}

protected:
    Expression::Pointer mpLeft;

    Expression::Pointer mpRight;
};
```

applications/OptimizationApplication/custom_utilities/container_variable_data/expressions.h
```diff
@@ -0,0 +1,438 @@
...
+    ///@name Life cycle
+    ///@{
+
+    BinaryAddExpression(
+        Expression::Pointer pLeft,
+        Expression::Pointer pRight);
+
+    BinaryAddExpression(const BinaryAddExpression& rOther) = delete;
+
+    ~BinaryAddExpression() override = default;
```
matekelemen: unnecessary

applications/OptimizationApplication/custom_utilities/container_variable_data/expressions.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+std::size_t LiteralArray3Expression::GetDimension() const
+{
+    return mDimension;
+}
+
+LiteralVectorExpression::LiteralVectorExpression(
+    Kratos::shared_ptr<const Vector> pValue,
+    const IndexType Dimension)
+    : mpValue(pValue),
```
matekelemen: You can avoid fiddling with the refcount unnecessarily, but I guess construction won't happen in performance critical parts.


```suggestion
    : mpValue(std::move(pValue)),
```

applications/OptimizationApplication/custom_utilities/container_variable_data/specialized_container_variable_data.h
```diff
@@ -0,0 +1,217 @@
...
+// Application includes
+#include "container_variable_data.h"
+
+namespace Kratos {
+
+///@name Kratos Classes
+///@{
+
+/**
+ * @brief Construct a new specialized container variable data
```
matekelemen:
```suggestion
 * @brief Construct a new specialized @ref ContainerVariableData.
```

applications/OptimizationApplication/custom_utilities/container_variable_data/specialized_container_variable_data.h
```diff
@@ -0,0 +1,217 @@
...
+ * These expressions are evaluated only if AssignDataToContainerVariable. This also does not create additional
+ * vectors to hold the resultant value of the expression. It uses the model parts respective containers entity input/output
+ * method specified to write evaluated entity resultant values to model part entities. Hence, these
+ * SpecializedContainerVariableData can be easily visualized using common variable data visualization methods.
+ *
+ * Copy constructor is introduced with the base class type becaues, this allows copying data
+ * between compatible SpecializedContainerVariableData containers. This copy is also light weight since
+ * this only copies the pointers, not the data itself.
+ *
+ * This class is optimized and compatible with OpenMP and MPI.
```
matekelemen:
```suggestion
 * This class can take advantage of OpenMP and MPI.
```

applications/OptimizationApplication/custom_utilities/container_variable_data/specialized_container_variable_data.h
```diff
@@ -0,0 +1,217 @@
...
+namespace Kratos {
+
+///@name Kratos Classes
+///@{
+
+/**
+ * @brief Construct a new specialized container variable data
+ *
+ * This class is used to represent objects which are the specializations
+ * of ContainerVariableData. The base class ifof type ContainerVariableData<TContainerType>.
```
matekelemen:
```suggestion
 * of ContainerVariableData. The base class is @ref ContainerVariableData<TContainerType>.
```

applications/OptimizationApplication/custom_utilities/container_variable_data/specialized_container_variable_data.h
```diff
@@ -0,0 +1,217 @@
...
+ * Hence this class can be specialized to read and write data to different entity containers
+ * for each container type.
+ *
+ * This class does not have any member variables, hence it uses expressions to represent the
```
matekelemen:
```suggestion
 * This class does not have any member variables, instead it uses expressions to represent the
```

applications/OptimizationApplication/custom_utilities/container_variable_data/specialized_container_variable_data.h
```diff
@@ -0,0 +1,217 @@
...
+ * data. Therefore, all the operations such as "+", "-", "*", etc are light weight because
+ * they do not compute the value in place, they just create an expression which keeps track fo the expressions.
```
matekelemen:
```suggestion
 * data. Therefore, all the operations such as "+", "-", "*", etc are light weight because
 * they do not compute the value in place, they just create an expression which keeps track of the expressions (lazy evaluation).
```

applications/OptimizationApplication/custom_utilities/container_variable_data/specialized_container_variable_data.h
```diff
@@ -0,0 +1,217 @@
...
+ * These expressions are evaluated only if AssignDataToContainerVariable. This also does not create additional
```
matekelemen:
```suggestion
 * These expressions are evaluated only if @ref AssignDataToContainerVariable is invoked. This also does not create additional
```

applications/OptimizationApplication/custom_utilities/container_variable_data/specialized_container_variable_data.h
```diff
@@ -0,0 +1,217 @@
...
+ * vectors to hold the resultant value of the expression. It uses the model parts respective containers entity input/output
```
matekelemen:
```suggestion
 * vectors to hold the resulting value of the expression. It uses the model part's respective containers entity input/output
```

applications/OptimizationApplication/custom_utilities/container_variable_data/specialized_container_variable_data.h
```diff
@@ -0,0 +1,217 @@
...
+ * method specified to write evaluated entity resultant values to model part entities. Hence, these
```
matekelemen: This part could be made more clear. Also, `resultant` => `resulting`.

applications/OptimizationApplication/custom_utilities/container_variable_data/specialized_container_variable_data.h
```diff
@@ -0,0 +1,217 @@
...
+ * SpecializedContainerVariableData can be easily visualized using common variable data visualization methods.
+ *
+ * Copy constructor is introduced with the base class type becaues, this allows copying data
```
matekelemen:
```suggestion
 * Copy constructor is introduced with the base class type because, this allows copying data
```

applications/OptimizationApplication/custom_utilities/container_variable_data/specialized_container_variable_data.h
```diff
@@ -0,0 +1,217 @@
...
+ * between compatible SpecializedContainerVariableData containers. This copy is also light weight since
```
matekelemen:
```suggestion
 * between compatible @ref SpecializedContainerVariableData containers. This copy is also light weight since
```

applications/OptimizationApplication/custom_utilities/container_variable_data/specialized_container_variable_data.cpp
```diff
@@ -0,0 +1,440 @@
...
+
+template <class TContainerType, class TContainerDataIO>
+SpecializedContainerVariableData<TContainerType, TContainerDataIO>& SpecializedContainerVariableData<TContainerType, TContainerDataIO>::operator=(const SpecializedContainerVariableData& rOther)
+{
+    KRATOS_ERROR_IF(&this->GetModelPart() != &rOther.GetModelPart())
+        << "Mismatching model parts found in assignment.\n"
+        << "      Assignee data: " << *this << "\n"
+        << "      Assignor data: " << rOther << "\n";
+
+    this->mpExpression = rOther.mpExpression;
```
matekelemen: shallow vs deep copy
I don't know which one is correct for this case, just asking whether you gave it some thought.

applications/OptimizationApplication/tests/test_OptimizationApplication.py
```diff
@@ -55,12 +62,20 @@ def AssembleTestSuites():
...
     smallSuite.addTests(KratosUnittest.TestLoader().loadTestsFromTestCases([TestOptimizationUtils]))
     smallSuite.addTests(KratosUnittest.TestLoader().loadTestsFromTestCases([TestModelPartUtils]))
+    smallSuite.addTests(KratosUnittest.TestLoader().loadTestsFromTestCases([TestContainerVariableDataUtils]))
 
     smallSuite.addTests(KratosUnittest.TestLoader().loadTestsFromTestCases([TestMassResponseFunctionBeams]))
     smallSuite.addTests(KratosUnittest.TestLoader().loadTestsFromTestCases([TestMassResponseFunctionShells]))
     smallSuite.addTests(KratosUnittest.TestLoader().loadTestsFromTestCases([TestMassResponseFunctionSolids]))
     smallSuite.addTests(KratosUnittest.TestLoader().loadTestsFromTestCases([TestLinearStrainEnergyResponseFunction]))
 
+    smallSuite.addTests(KratosUnittest.TestLoader().loadTestsFromTestCases([TestHistoricalContainerVariableDataHolder]))
```
RezaNajian: I just noticed that we have 7 tests in the test_container_variable_data_holder.py and six of them are used for testing. Please also add TestContainerVariableData.

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
sunethwarna: done :)

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
sunethwarna: done :)

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
sunethwarna: done :)

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
sunethwarna: done :)

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
sunethwarna: done :)

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
sunethwarna: done :)

sunethwarna: done :)

applications/OptimizationApplication/custom_utilities/container_variable_data/container_variable_data.cpp
sunethwarna: `Expressions` only have const interfaces, so they cannot be changed after instantiated.

applications/OptimizationApplication/custom_utilities/container_variable_data/container_variable_data.cpp
sunethwarna: done :)

applications/OptimizationApplication/custom_utilities/container_variable_data_holder/container_variable_data_holder.cpp
sunethwarna: I am not sure, I have never checked it. This was like this from the begining, and I was just using my old musclememory ;)

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
sunethwarna: These names will be changed in a different PR

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
sunethwarna: These names will be changed in a different PR

applications/OptimizationApplication/custom_utilities/container_variable_data/container_variable_data.h
sunethwarna: I would like to keep `CopyDataFrom` to indicate it is copying, the assignemnt operator may become suspicious if they think I am assigning the memory locations.

applications/OptimizationApplication/custom_utilities/container_variable_data/expressions.h
sunethwarna: done :) and rearranged it nicely :D

applications/OptimizationApplication/custom_utilities/container_variable_data/expressions.h
sunethwarna: Done and rearanged nicely :D

applications/OptimizationApplication/custom_utilities/container_variable_data/expressions.cpp
sunethwarna: Removed and rearanged nicely :D

applications/OptimizationApplication/custom_utilities/container_variable_data/expressions.h
sunethwarna: removed :)

applications/OptimizationApplication/custom_utilities/container_variable_data/specialized_container_variable_data.cpp
sunethwarna: This is working because, expressions only have `const` interfaces, hence the values within an expression cannot be changed once instantiated.

applications/OptimizationApplication/custom_utilities/container_variable_data/specialized_container_variable_data.h
sunethwarna: Done :)

sunethwarna:
Hi @matekelemen. I have changed the structure to be more flexible and more `KraotsCore` portable implementation. Sorry for making the (somewhat not so small) changes to the backend and tagging you again for review. Could you have a look when you have time?

RezaNajian (dismissed):

applications/OptimizationApplication/tests/test_OptimizationApplication.py
sunethwarna: done :)

applications/OptimizationApplication/tests/test_OptimizationApplication.py
sunethwarna: `TestContainerVariableData` is an abstract class, and other 6 classes are the derrived classes. Hence, we only have valid tests in the derrived classes.

applications/OptimizationApplication/custom_utilities/container_properties_data_io.h
```diff
@@ -0,0 +1,58 @@
...
+///@{
+
+namespace ContainerDataIOTags {
+    struct Properties    {};
+} // namespace Tags
+
+template <>
+struct ContainerDataIO<ContainerDataIOTags::Properties>
+{
+    static constexpr std::string_view mInfo = "Properties";
```
matekelemen, line 36: Props for using `std::string_view`!

applications/OptimizationApplication/tests/model_part_utils_test/quads.mdpa
```diff
@@ -34,42 +34,42 @@ Begin Nodes
...
-	8	1	8	10	15	14	
-	9	1	12	11	16	17	
-	10	1	11	13	18	16	
-	11	1	13	14	19	18	
-	12	1	14	15	20	19	
-	13	1	17	16	21	22	
-	14	1	16	18	23	21	
-	15	1	18	19	24	23	
-	16	1	19	20	25	24	
+Begin Elements	HelmholtzBulkTopology3D4N
```
matekelemen, line 37: wut? why?

applications/OptimizationApplication/tests/test_container_expression.py
```diff
@@ -0,0 +1,364 @@
+
+from abc import ABC
+from abc import abstractmethod
```
matekelemen, line 3: FYI you can do this too.
Don't accept it tho, it's not worth rerunning the CI for this.

```suggestion
from abc import ABC, abstractmethod
```

matekelemen (approved):

applications/OptimizationApplication/tests/model_part_utils_test/quads.mdpa
sunethwarna: just because I need an element which gives me a non-zero matrices for testing. This is the only element I found in the app which is compatible :D....

applications/OptimizationApplication/custom_utilities/container_properties_data_io.h
sunethwarna: Thats the only one which is a `constexpr` for strings as well :D

sunethwarna:
Thanks a lot @RezaNajian @matekelemen @Igarizza @FMeister :D....

PR #10769, opened 2023-02-17
[Docs] Add Support for Python Docs with Doxygen

matekelemen:
## Changelog
- update doxygen config version
- add project name `KratosMultiphysics`
- disable `PYTHON_DOCSTRING` (treat python docstrings as doxygen annotations)
- new alias `@classname` => `@class` to avoid a [doxygen bug](https://github.com/doxygen/doxygen/issues/9859)
- exclude tests from the docs
- enable support for latex math in doxygen annotations
- add annotations to `rans_formulation.py` as an example for using doxygen in python

## Notes
I don't have access to `pages`, so someone with those privileges will have to take care of deployment (looking at you @sunethwarna). What I usually do is designate a branch (`ci/docs` for example) that doesn't keep history and is regularly (say, nightly) pushed to by the doc generating workflow. It'd be nice if we could integrate the jekyll docs as well, but I don't have much experience with that.

It seems like doxygen isn't the most stable software and breaks easily when there's a mistake in the annotations (and sometimes even when there aren't any mistakes) and it just fails silently, skipping doc generation for some parts of the repo. At this point I'm out of my depth because 1) I can't fix everyone's annotations, and 2) I can't fix doxygen. That said, it's definitely better than nothing.

As a reference, running doxygen from scratch took ~18 mins on my machine to generate a (partially) complete documentation.


sunethwarna:
I am on it 👍

applications/RANSApplication/python_scripts/formulations/rans_formulation.py
```diff
@@ -1,119 +1,107 @@
...
+    def GetDomainSize(self) -> int:
+        """ @brief Returns domain size."""
         return self.__base_computing_model_part.ProcessInfo[Kratos.DOMAIN_SIZE]
 
-    def AddRansFormulation(self, formulation):
-        """Adds another RansFormulation to current formulation creating a list of formulations.
-
-        Args:
-            formulation (RansFormulation): Formulation to be added
+    def AddRansFormulation(self, formulation: "RansFormulation") -> None:
```
philbucher:
```suggestion
    def AddRansFormulation(self, formulation: RansFormulation) -> None:
```

I recommend to use it directly

applications/RANSApplication/python_scripts/formulations/rans_formulation.py
```diff
@@ -193,175 +171,139 @@ def IsConverged(self):
...
+            @param settings: Constants settings.
         """
         self.__ExecuteRansFormulationMethods("SetConstants", [settings])
 
-    def GetRansFormulationsList(self):
-        """Returns list of formulations in this formulation
-
-        Returns:
-            List(RansFormulation): List of formulations in this formulation
+    def GetRansFormulationsList(self) -> "list["RansFormulation"]":
```
philbucher:
```suggestion
    def GetRansFormulationsList(self) -> list[RansFormulation]:
```

sunethwarna: I think this is failing in python3.6... :/

matekelemen: it would fail on python 3.8 as well (which is the version we currently support I think)
I just messed up and forgot about the inner `""`; I should probably stop making commits on Sunday nights :(

applications/RANSApplication/python_scripts/formulations/rans_formulation.py
matekelemen: the python interpreter reads member signatures when the class isn't fully defined yet, so I can't use the class in there directly, only in the function body.

matekelemen:
Any ideas why checks in the CI seem duplicated?
@roigcarlo

roigcarlo:
Yep. We split the CI for doc / sources, and doc needed a mock so that it triggers the validation check. They should be exclusive, but it seems that if your touch doc and sources files in different commits inside the same PR, both of them trigger.

applications/RANSApplication/python_scripts/formulations/rans_formulation.py
philbucher: > I think this is failing in python3.6... :/

that is no longer supported :tada: 

> it would fail on python 3.8 as well (which is the version we currently support I think)

yes you are right this is introduced with 3.9
you can either do `from typing import List`, or what you are doing

rubenzorrilla:
@KratosMultiphysics/technical-committee agrees to put this html generation to nightly and export it to the examples repo, then connecting this with the exiting documentation. @sunethwarna will take care of it.

sunethwarna (dismissed):

matekelemen:
@sunethwarna Can you please add a doc entry on the deployment process once it's stable?

sunethwarna:
@matekelemen No CI action will be there in Kratos repo for doxygen. Hence It not possible to add a depoloyment entry in this PR. Doxygen action is in the Documentation repo. I can add an entry there.

matekelemen:
> No CI action will be there in Kratos repo for doxygen.
Yes, that's exactly why we need some description of how it's actually getting generated. How about putting it in the doc repo's readme?

sunethwarna (approved):

PR #10771, opened 2023-02-17
[OptApp] Restructure 7 - Added response function implementor

sunethwarna:
**📝 Description**
This PR adds `ResponseFunctionImplementor` based on the bridge pattern. It allows seamless calculation of standardized sensitivities for objectives and constraints. followings are supported.

1. Custom scaling of objectives, constraints.
2. Standardization of objective values and sensitivities for "minimization" and "maximization" problems.
3. Standardization of constraint values and sensitivities for "=", "<", "<=", ">=" equality or in-equality problems.

Todo:
- [x]  Add tests

**🆕 Changelog**
- Added `ResponseFunctionImplementor`


applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
```diff
@@ -0,0 +1,194 @@
...
+    def GetName(self) -> str:
+        return self.__name
+
+    def GetValue(self, solution_step_index: int = 0) -> float:
+        key = f"{self._key_prefix}/value"
+        if solution_step_index == 0 and not self._optimization_info.HasValue(key):
+            self._optimization_info.SetValue(key, self.__response_function.CalculateValue())
+        return self._optimization_info.GetValue(key, solution_step_index)
+
+    def GetSensitivity(self, sensitivity_variable, sensitivity_data_container: ContainerVariableDataHolderUnion, solution_step_index: int = 0):
```
FMeister: I find it confusion that the method is called "Get.." (just as the one above, "GetValue") but it is not returning anything. What the method does is coping data to the optimization info. I think that should be reflected in the naming. 
For example: def CopySensitivitiesToOptimizationInfo

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
```diff
@@ -0,0 +1,194 @@
...
+            self.__initial_response_value = self.GetValue()
+
+        return self.__initial_response_value
+
+    @abstractmethod
+    def GetStandardizedValue(self, solution_step_index: int = 0) -> float:
+        pass
+
+    @abstractmethod
+    def GetStandardizedSensitivity(self, sensitivity_variable, sensitivity_data_container: ContainerVariableDataHolderUnion, solution_step_index: int = 0):
```
FMeister: The same applies here

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
```diff
@@ -0,0 +1,194 @@
...
+        absolute_change = (self.GetValue() / self._GetInitialResponseValue() - 1.0)
+
+        msg = "\tObjective info:"
+        msg += f"\n\t\t name          : {self.GetName()}"
+        msg += f"\n\t\t type          : {self.__objective_type}"
+        msg += f"\n\t\t value         : {self.GetValue()}"
+        msg += f"\n\t\t rel_change [%]: {relative_change * 100.0}"
+        msg += f"\n\t\t abs_change [%]: {absolute_change * 100.0}"
+
+        self._optimization_info.SetValue(f"{self._key_prefix}/relative_change", relative_change)
```
FMeister: I find it counter intuitive that this function is responsible for returning a message and for adding data to the optimization info.
A better name (like: GetResponseInfoMsgAndCopyInfoToOptInfo) would fix this in my opinion.

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
```diff
@@ -0,0 +1,194 @@
...
+            self.__scaling *= -1.0
+        else:
+            raise RuntimeError(f"Requesting unsupported type {self.__objective_type} for objective response function. Supported types are: \n\tminimization\n\tmaximization")
+
+        super().__init__(parameters["response_name"].GetString(), optimization_info)
+
+    def GetStandardizedValue(self, solution_step_index: int = 0) -> float:
+        return self.GetValue(solution_step_index) * self.__scaling
+
+    def GetStandardizedSensitivity(self, sensitivity_variable, sensitivity_data_container: ContainerVariableDataHolderUnion, solution_step_index: int = 0):
```
FMeister: The same applies here

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
```diff
@@ -0,0 +1,194 @@
...
+        msg += f"\n\t\t name          : {self.GetName()}"
+        msg += f"\n\t\t value         : {self.GetValue()}"
+        msg += f"\n\t\t type          : {self.__constraint_type}"
+        msg += f"\n\t\t ref_value     : {self.GetReferenceValue()}"
+        msg += f"\n\t\t is_active     : {self.IsActive()}"
+        msg += f"\n\t\t rel_change [%]: {relative_change * 100.0}"
+        msg += f"\n\t\t abs_change [%]: {absolute_change * 100.0}"
+        msg += f"\n\t\t violation  [%]: {violation * 100.0}"
+
+        self._optimization_info.SetValue(f"{self._key_prefix}/relative_change", relative_change)
```
FMeister: The same applies here

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
```diff
@@ -0,0 +1,194 @@
...
+    def GetReferenceValue(self):
+        if self.__reference_value is None:
+            self.__reference_value = self.GetValue()
+
+        return self.__reference_value
+
+    def GetStandardizedValue(self, solution_step_index: int = 0) -> float:
+        return self.__standardization_value * (self.GetValue(solution_step_index) * self.__scaling - self.GetReferenceValue())
+
+    def GetStandardizedSensitivity(self, sensitivity_variable, sensitivity_data_container: ContainerVariableDataHolderUnion, solution_step_index: int = 0):
```
FMeister: The same applies here

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def GetValue(self, solution_step_index: int = 0) -> float:
+        key = f"{self._key_prefix}/value"
+        if solution_step_index == 0 and not self._optimization_info.HasValue(key):
+            self._optimization_info.SetValue(key, self.__response_function.CalculateValue())
+        return self._optimization_info.GetValue(key, solution_step_index)
+
+    def GetSensitivity(self, sensitivity_variable, sensitivity_data_container: ContainerVariableDataHolderUnion, solution_step_index: int = 0):
+        key = f"problem_data/response_data/{sensitivity_data_container.GetModelPart().FullName()}/{self.GetName()}/sensitivities/{sensitivity_variable.Name()}/raw"
+        if solution_step_index == 0 and not self._optimization_info.HasValue(key):
```
Igarizza: What would happen if the given key is wrong due to the wrong FullName, modelPart and etc. ?

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
Igarizza: I agree the Get Function should return the object. I don't know if the logic for solution_step_index == 0 should be inside this function. With less hidden logic, it would be easier to use and extend in the future. 

Maybe you can use GetSensitivity to simply Get and through an error if the data doesn't exist? What do you think? Of course, it leads to changes to move add CalculateSensitivity to some external place.

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
Igarizza: It should be called, StandardizeSensitivity()

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
```diff
@@ -0,0 +1,194 @@
...
+        self._optimization_info.SetValue(f"{self._key_prefix}/absolute_change", absolute_change)
+        self._optimization_info.SetValue(f"{self._key_prefix}/type", self.__objective_type)
+        return msg
+
+class ConstraintResponseFunctionImplementor(ResponseFunctionImplementor):
+    def __init__(self, parameters: Kratos.Parameters, optimization_info: OptimizationInfo):
+        default_parameters = Kratos.Parameters("""{
+            "response_name": "",
+            "type"         : "",
+            "scaling"      : 1.0,
```
Igarizza: This scaling is redundant. We typically don't weight constraints. Or am I missing smth ...

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def GetResponseInfo(self) -> str:
+        if self._optimization_info["step"] > 1:
+            relative_change = (self.GetValue() / self.GetValue(1) - 1.0)
+        else:
+            relative_change = 0.0
+        absolute_change = (self.GetValue() / self._GetInitialResponseValue() - 1.0)
+
+        if self.IsActive():
+            violation = self.GetStandardizedValue() / self.GetReferenceValue()
```
Igarizza: You need to check if self.GetReferenceValue() is not zero.

Igarizza (commented):
Looks good in general, have added few discussion points. They might be wrong as I am haven't seen the rest part of codes and how this class is used.

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
```diff
@@ -0,0 +1,194 @@
...
+
+    @abstractmethod
+    def GetResponseType(self) -> str:
+        pass
+
+    @abstractmethod
+    def GetResponseInfo(self) -> str:
+        pass
+
+class ObjectiveResponseFunctionImplementor(ResponseFunctionImplementor):
```
sunethwarna:
```suggestion
class Objective(ResponseFunctionImplementor):
```

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
```diff
@@ -0,0 +1,194 @@
...
+        msg += f"\n\t\t value         : {self.GetValue()}"
+        msg += f"\n\t\t rel_change [%]: {relative_change * 100.0}"
+        msg += f"\n\t\t abs_change [%]: {absolute_change * 100.0}"
+
+        self._optimization_info.SetValue(f"{self._key_prefix}/relative_change", relative_change)
+        self._optimization_info.SetValue(f"{self._key_prefix}/absolute_change", absolute_change)
+        self._optimization_info.SetValue(f"{self._key_prefix}/type", self.__objective_type)
+        return msg
+
+class ConstraintResponseFunctionImplementor(ResponseFunctionImplementor):
```
sunethwarna:
```suggestion
class Constraint(ResponseFunctionImplementor):
```

Move this to a different file.

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: Move this to a different file.

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
```diff
@@ -0,0 +1,194 @@
+from abc import ABC
+from abc import abstractmethod
+
+import KratosMultiphysics as Kratos
+from KratosMultiphysics.OptimizationApplication.optimization_info import OptimizationInfo
+from KratosMultiphysics.OptimizationApplication.responses.response_function import ResponseFunction
+from KratosMultiphysics.OptimizationApplication.utilities.helper_utilities import ContainerVariableDataHolderUnion
+
+class ResponseFunctionImplementor(ABC):
```
sunethwarna: Remove this, and make the common parts of code with `kwargs` in python in utils.

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: Lets change the name to `CalculateSensitivity`

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: will change as above

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: `CalculateStandardizedSensitivity`

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: I will change this :D....

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: This is just only for reference value scaling from the user side.

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: will change as above.

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: Agree, will do....

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: will change..

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
Igarizza: Do you mean the class is going to be in Constraint.py or so?

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: the key is an auto generated string, the model part name is taken from the corresponding container. So the model part name cannot be wrong. This is now called through the communicators so that, string manipulations are kept minimum to avoid developer mistakes.

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: changed :)

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: yes, and done :)

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: done :)

sunethwarna:
@RezaNajian @Igarizza I did all the changes requested by both. Could you take a look?

I introduced `Communicator` classes to minimize string manipulations required to store data in `OptimizationInfo` since it is flly a nested map with `(str, Any)` pairs. (to minimize developer mistakes).

applications/OptimizationApplication/python_scripts/utilities/response_function_implementor.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/responses/standardized_constraint.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def GetReferenceValue(self):
+        if self.__reference_value is None:
+            self.__reference_value = self.__communicator.GetScaledValue()
+        return self.__reference_value
+
+    def GetStandardizedValue(self, step_index: int = 0) -> float:
+        return self.__communicator.GetScaledValue(step_index, self.__standardization_value) - self.__standardization_value * self.GetReferenceValue()
+
+    def IsActive(self, step_index: int = 0) -> bool:
```
Igarizza, line 66: This is optimization algorithm-specific. There are algorithms where the constraints are always active, or they use different logic. I think you should move this function into the algorithm.

applications/OptimizationApplication/python_scripts/responses/standardized_constraint.py
```diff
@@ -0,0 +1,94 @@
...
+        return (self.__constraint_type == "=") or self.GetStandardizedValue(step_index) >= 0.0
+
+    def GetViolationRatio(self, step_index: int = 0) -> float:
+        return (self.GetStandardizedValue(step_index) / self.GetReferenceValue() if abs(self.GetReferenceValue()) > 1e-12 else self.GetStandardizedValue(step_index)) * self.IsActive()
```
Igarizza, line 70: if abs(self.GetReferenceValue()) > 1e-12 and if self.IsActive()==0 -> will it return 0.0?

applications/OptimizationApplication/python_scripts/utilities/optimization_info.py
```diff
@@ -1,145 +1,411 @@
...
+        self.__parent = None
+        self.__buffer_index = 0
+        self.__buffered_data: 'list[dict[str, Any]]' = []
+        self.__sub_items: 'dict[str, OptimizationInfo]' = {}
 
-    def HasOptimizationProcess(self, category: any, routine_name: str) -> bool:
-        if not self.HasOptimizationProcessType(category):
-            raise RuntimeError(f"No objects of type \"{category.__name__}\" [ requested routine_name = \"{routine_name}\" ]. Following are the available type options: \n\t" + "\n\t".join(self.__optimization_processes.keys()))
+        # sets the buffer
+        self.__SetBufferSize(buffer_size)
```
Igarizza: what is it, buffer_size?

Igarizza (commented):
Few more small comments

applications/OptimizationApplication/python_scripts/responses/standardized_objective.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def GetResponseFunctionName(self) -> str:
+        return self.__communicator.GetName()
+
+    def GetInitialValue(self) -> float:
+        if self.__initial_response_value is None:
+            self.__initial_response_value = self.__communicator.GetScaledValue()
+        return self.__initial_response_value
+
+    def GetResponseType(self) -> str:
```
RezaNajian: I would be consistent and either call the function GetObjectiveType() or GetType() simply

applications/OptimizationApplication/python_scripts/responses/standardized_objective.py
```diff
@@ -0,0 +1,65 @@
...
+        return self.__objective_type
+
+    def GetStandardizedValue(self, step_index: int = 0) -> float:
```
RezaNajian, line 46: here the same, since you are scaling I would call the method GetScaledValue, the name does not reflect what it does unless one check the implementation

applications/OptimizationApplication/python_scripts/responses/standardized_objective.py
```diff
@@ -0,0 +1,65 @@
+import KratosMultiphysics as Kratos
+import KratosMultiphysics.OptimizationApplication as KratosOA
+from KratosMultiphysics.OptimizationApplication.utilities.optimization_info import OptimizationInfo
+from KratosMultiphysics.OptimizationApplication.utilities.union_utilities import SupportedSensitivityFieldVariableTypes
+from KratosMultiphysics.OptimizationApplication.utilities.communicators.response_function_communicator import ResponseFunctionCommunicator
```
RezaNajian: I am not fully convinced about this concept/implementation of Communicators like ResponseFunctionCommunicator and OptimizationComponentCommunicator. It is getting like pasta and everyone needs the others. Basically, you are forwarding the implementation and work from this class to another class. Actually, the functions in these Communicators are only utilities and I would highly appreciate if you make them not class rather Python utilities if you do not like repeat function implementation for name, value and sensitivites

applications/OptimizationApplication/python_scripts/utilities/communicators/optimization_component_communicator.py
```diff
@@ -0,0 +1,159 @@
+from typing import Any
```
RezaNajian: Again, I believe we should not make the handling of optimization info more complicated than it already is. With this implementation, I see that you have decentralized some of the data and respective operations from OptimizationInfo to a new class called OptimizationComponentCommunicator. This adds another layer of complexity without significant benefits. I suggest unifying this class with OptimizationInfo because ultimately, what is stored and done here is related to Optimization info. No one will derive from classes like OptimizationComponentCommunicator and ResponseFunctionCommunicator, and they can be unified into either OptimizationInfo and Response classes or treated as utilities rather than new classes. Additionally, if they are treated as utilities, data like __components and __problem_data should be transferred to OptimizationInfo.

applications/OptimizationApplication/python_scripts/utilities/communicators/response_function_communicator.py
```diff
@@ -0,0 +1,199 @@
+import KratosMultiphysics as Kratos
```
RezaNajian: please check my point in optimization_component_communicator. We can discuss further

RezaNajian (changes requested):
@sunethwarna, thank you very much for creating the Objective and Constraint classes. However, I am not fully convinced about the strategy used to avoid double implementations. Please take a look at my comments and let me know if we need to discuss this matter further.

applications/OptimizationApplication/python_scripts/responses/standardized_objective.py
Igarizza: GetType() then better, as it can be used for both

sunethwarna: changed to `GetType`

applications/OptimizationApplication/python_scripts/responses/standardized_objective.py
sunethwarna: I would keep the name as it is, because this makes sense from the perspective of `Algorithms`. This does scaling as well as the negation for `maximization` objectives. I added more elaborate documentation for the method.

applications/OptimizationApplication/python_scripts/responses/standardized_objective.py
sunethwarna: Moved all the common methods to `StandardizationUtilities`, and removed all the `Communicator` classes.

applications/OptimizationApplication/python_scripts/utilities/communicators/optimization_component_communicator.py
sunethwarna: Moved all the common methods to `StandardizationUtilities`, and removed all the `Communicator` classes.

applications/OptimizationApplication/python_scripts/utilities/communicators/response_function_communicator.py
sunethwarna: Moved all the common methods to `StandardizationUtilities`, and removed all the `Communicator` classes.

applications/OptimizationApplication/python_scripts/responses/standardized_constraint.py
sunethwarna: These standardizations are only for the algorithm which are developed within `Kratos`. If someone wants then he/she can have their own mechanisms. This is only to simplify for the upcoming `Kratos` specific algorithms.

applications/OptimizationApplication/python_scripts/responses/standardized_constraint.py
sunethwarna: yes.

applications/OptimizationApplication/python_scripts/utilities/optimization_info.py
sunethwarna: `OptimizationInfo` is a buffered data container where the buffer sizes can vary within different `OptimizationInfo` items. The buffer is used to obtain previous iteration data (This is implemented with a cyclic buffer).

applications/OptimizationApplication/tests/test_standardized_responses.py
```diff
@@ -0,0 +1,217 @@
...
+
+class TestStandardizedComponent(kratos_unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.model = Kratos.Model()
+        cls.model_part = cls.model.CreateModelPart("test")
+        cls.optimization_info = OptimizationInfo()
+
+        cls.response_function = MassResponseFunction(cls.model, Kratos.Parameters("""{"evaluated_model_part_names": ["test"]}"""), cls.optimization_info)
+        cls.optimization_info.AddComponent("mass", cls.response_function)
```
RezaNajian: i would say we need to be more precise and use AddResponse rather than AddComponent

applications/OptimizationApplication/python_scripts/responses/standardized_objective.py
```diff
@@ -0,0 +1,92 @@
...
+            self.__scaling = parameters["scaling"].GetDouble()
+        elif self.__objective_type == "maximization":
+            self.__scaling = -parameters["scaling"].GetDouble()
+        else:
+            raise RuntimeError(f"Requesting unsupported type {self.__objective_type} for objective response function. Supported types are: \n\tminimization\n\tmaximization")
+
+        self.__utility = StandardizationUtilities(parameters["response_name"].GetString(), optimization_info)
+        self.__initial_response_value = None
+
+    def GetResponseFunctionName(self) -> str:
```
RezaNajian: I would rather call it GetName simply

applications/OptimizationApplication/python_scripts/responses/standardized_constraint.py
```diff
@@ -0,0 +1,94 @@
...
+        if self.__constraint_type in ["<", "<=", "="]:
+            self.__standardization_value = 1.0
+        elif self.__constraint_type in [">", ">="]:
+            self.__standardization_value = -1.0
+        else:
+            raise RuntimeError(f"Provided \"type\" = {self.__constraint_type} is not supported in constraint response functions. Followings are supported options: \n\t=\n\t<=\n\t<\n\t>=\n\t>")
+
+        self.__utility = StandardizationUtilities(parameters["response_name"].GetString(), optimization_info)
+
+    def GetResponseFunctionName(self) -> str:
```
RezaNajian: I would call it GetName

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, model: Kratos.Model, parameters: Kratos.Parameters, optimizat
         parameters.ValidateAndAssignDefaults(default_settings)
 
         self.perturbation_size = parameters["perturbation_size"].GetDouble()
-        self.primal_analysis_execution_policy_wrapper: ExecutionPolicyDecorator = optimization_info.GetOptimizationProcess(ExecutionPolicyDecorator, parameters["primal_analysis_name"].GetString())
+        self.primal_analysis_execution_policy_decorator: ExecutionPolicyDecorator = optimization_info.GetComponent(parameters["primal_analysis_name"].GetString())
```
RezaNajian: why not to be more precise, and make maps or list for every component based on the type for example here i would call it GetAnalysis to be crystally clear

RezaNajian: i also noticed that you are not calling the init of the base class or ?!

applications/OptimizationApplication/python_scripts/utilities/optimization_info.py
```diff
@@ -1,145 +1,594 @@
-import KratosMultiphysics as Kratos
+from __future__ import annotations
```
RezaNajian, line 1: i can make my self live with OptimizationInfo and OptimizationData but hoenestly it is very confusing, even the name of the file is optimization_info, why do not you merge them into single class ?! what are the reasons ?

applications/OptimizationApplication/tests/test_optimization_info.py
```diff
@@ -3,96 +3,253 @@
...
-    class TestRoutine(Kratos.Process):
-        def __init__(self):
-            super().__init__()
-
-    def test_AdvanceSolutionStepGet(self):
-        optimization_info = OptimizationInfo()
-        optimization_info.SetBufferSize(3)
+class TestOptimizationData(kratos_unittest.TestCase):
+    def test_AdvanceStep(self):
+        optimization_info = OptimizationData(3)
```
RezaNajian: optimization_data

RezaNajian: Here is an example of confusion where the terms OptimizationInfo and OptimizationData are used interchangeably. It is suggested to merge them into a single class, unless there is a specific reason not to do so.

RezaNajian (commented):
in general it is fine please either address or answer the reviews and merge asap :-)

applications/OptimizationApplication/python_scripts/utilities/optimization_info.py
sunethwarna: It cannot be merged into one class. `OptimizationData` is a recursive data holder which can hold itself as well. So it always needs to be a separate class. `OptimizationInfo` is holding an `OptimizationData` which is passed around every where.

applications/OptimizationApplication/tests/test_optimization_info.py
sunethwarna: This is a reminiscence from the past design. I will change the it to `optimization_data`. As explained in the previous PR, these classes needs to stay separate.

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
sunethwarna: It is not inherited, It is a subclass of the `OptimizationInfo` with private methods (friend like methods in python between `OptimizationInfo` and `OptimizationData`)

applications/OptimizationApplication/python_scripts/responses/standardized_constraint.py
sunethwarna: will change :)

applications/OptimizationApplication/python_scripts/responses/standardized_objective.py
sunethwarna: will change :)

sunethwarna: done :)

applications/OptimizationApplication/python_scripts/responses/standardized_constraint.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/responses/standardized_constraint.py
Igarizza: I agree, as currently, you have only the Gradient projection method. Still, the implemented logic is **only** for Gradient projection; hence, other Shape Optimization App methods can't be used. As a result, we will have an implementation inside response for the GP method and custom in any different algorithm. I find it inconsistent in the near future.

Igarizza (dismissed):
My only complaint regarding IsActive as I think it should be implemented inside opt algorithm. It can be changed in the near future when we implement more algorithms.

applications/OptimizationApplication/tests/test_standardized_responses.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
sunethwarna: moved to a different file to avoid confusion.

applications/OptimizationApplication/python_scripts/utilities/optimization_info.py
sunethwarna: Moved to a different file to avoid confusion.

applications/OptimizationApplication/tests/test_optimization_info.py
sunethwarna: moved to a different file to avoid confusion.

sunethwarna:
@RezaNajian I addressed your concerns :) Now it is better for me too with your suggestions :D.

Could you have a look?

applications/OptimizationApplication/python_scripts/utilities/buffered_dict.py
```diff
@@ -0,0 +1,406 @@
...
+    """
+    def __init__(self, buffer_size: int = 1):
+        """Creates an instance of BufferedDict with specified buffer size
+
+        This creates an instance of buffered data with given buffer size.
+
+        Args:
+            buffer_size (int, optional): Cyclic buffer size. Defaults to 1.
+        """
+        self.__parent = None
```
RezaNajian, line 30: @sunethwarna I understand the benefits of and needs for this class however, I am sure it can be simplified. It is too complicated and not friendly to be used. I explain it in the following. 
Basically, this class is containing historical data in step fashion, so there is a list of dicts, called __buffered_data which has a fixed size of buffer_size. Since the size is fixed, the __buffer_index and functions like  AdvanceStep become completely redundant. If the buffered data is ordered according the step index, i.e.,  _buffered_data = [step_0_dict,step_1_dict,step_2_dict,..], one can access everything without calling  AdvanceStep and using __buffer_index. In the current implementation since step_0 which is the latest state of the data, refers to the last item of the  _buffered_data one needs to use __buffer_index. However if we order the data like  _buffered_data = [step_0_dict,step_1_dict,step_2_dict,..], we are fully consistent and it is super handy to use. So I am sure we need to simplify the class as explained.

applications/OptimizationApplication/python_scripts/utilities/buffered_dict.py
```diff
@@ -0,0 +1,406 @@
...
+            info += f"\n{tabbing}\t--- Sub items ---"
+
+        next_tabbing = f"{tabbing}\t"
+        for sub_key, sub_value in self.__sub_items.items():
+            info += f"\n{tabbing}\t\"{sub_key}\": {sub_value.__Info(step_index, next_tabbing)}"
+
+        info += f"\n{tabbing}" + "}"
+        return info
+
+    def __getitem__(self, key: str) -> Any:
```
RezaNajian, line 364: again here we can simplify. Since you have implemented magic methods or dunder methods. there is absolutely no need for ClearStep function, one can do like buffered_data["step", 1].clear(). 
Using AdvanceStep and ClearStep functions is making complications without benefits.

RezaNajian (changes requested):

sunethwarna:
@RezaNajian I changed the brackets to `SetValue` and `GetValue` in tests and moved the contents of `ClearStep` to `AdvanceStep` and removed `ClearStep` method. could you take a look? Thanks :D

RezaNajian (approved):
Thanks Suneth for resolving the concerns and having discussions. It is fully fine to merge.

PR #10772, opened 2023-02-17
[OptApp] Restructure 8 - Add basic controls

sunethwarna:
**📝 Description**
This adds following baseic controls and their respective tests

1. Shape control to have shape optimization
2. Material properties control can be used for damage detection, topologiy optimization, system identification

**🆕 Changelog**
- Added `ShapeControl`
- Added `MaterialPropertiesControl`
- Added tests


applications/OptimizationApplication/python_scripts/controls/control.py
```diff
@@ -0,0 +1,21 @@
...
+        self.__name = name
+
+    def GetName(self) -> str:
+        return self.__name
+
+    def CalculateSensitivity(self, response_function: ResponseFunctionImplementor, output_sensitivities: KratosOA.CollectiveVariableDataHolder):
+        raise NotImplementedError("Calling base class Control::CalculateSensitivity method. Please implement it in the derrived class.")
+
+    def UpdateControl(self, update: KratosOA.CollectiveVariableDataHolder):
+        raise NotImplementedError("Calling base class Control::UpdateControl method. Please implement it in the derrived class.")
```
FMeister:
```suggestion
        raise NotImplementedError("Calling base class Control::UpdateControl method. Please implement it in the derived class.")
```
Typo

applications/OptimizationApplication/python_scripts/controls/control.py
```diff
@@ -0,0 +1,21 @@
...
+        self.__name = None
+
+    def SetName(self, name: str):
+        self.__name = name
+
+    def GetName(self) -> str:
+        return self.__name
+
+    def CalculateSensitivity(self, response_function: ResponseFunctionImplementor, output_sensitivities: KratosOA.CollectiveVariableDataHolder):
+        raise NotImplementedError("Calling base class Control::CalculateSensitivity method. Please implement it in the derrived class.")
```
FMeister:
```suggestion
        raise NotImplementedError("Calling base class Control::CalculateSensitivity method. Please implement it in the derived class.")
```
Typo

applications/OptimizationApplication/python_scripts/controls/material_properties_control.py
```diff
@@ -0,0 +1,65 @@
...
+        # clear the container
+        output_sensitivities.ClearVariableDataHolders()
+
+        for model_part in self.model_parts:
+            d_j_d_phi = KratosOA.ElementPropertiesContainerVariableDataHolder(model_part)
+            response_function.GetStandardizedSensitivity(self.sensitivity_variable, d_j_d_phi)
+
+            output_sensitivities.AddVariableDataHolder(d_j_d_phi)
+
+    def UpdateControl(self, update: KratosOA.CollectiveVariableDataHolder):
```
FMeister: I am not really getting what is happening here.
The name suggests we update the control (so our self).
But in the last line, we assign some data to some variable. But whose variable is it?

FMeister (commented):
Besides my comments, it's fine for me

applications/OptimizationApplication/python_scripts/controls/control.py
```diff
@@ -0,0 +1,21 @@
...
+    def SetName(self, name: str):
+        self.__name = name
+
+    def GetName(self) -> str:
+        return self.__name
+
+    def CalculateSensitivity(self, response_function: ResponseFunctionImplementor, output_sensitivities: KratosOA.CollectiveVariableDataHolder):
+        raise NotImplementedError("Calling base class Control::CalculateSensitivity method. Please implement it in the derived class.")
+
+    def UpdateControl(self, update: KratosOA.CollectiveVariableDataHolder):
```
sunethwarna:
```suggestion
    def Update(self, update: KratosOA.CollectiveVariableDataHolder):
```

applications/OptimizationApplication/python_scripts/controls/control.py
```diff
@@ -0,0 +1,21 @@
...
+        super().__init__()
+        self.__name = None
+
+    def SetName(self, name: str):
+        self.__name = name
+
+    def GetName(self) -> str:
+        return self.__name
+
+    def CalculateSensitivity(self, response_function: ResponseFunctionImplementor, output_sensitivities: KratosOA.CollectiveVariableDataHolder):
```
sunethwarna:
```suggestion
    def MapFirstDerivative(self, response_function: ResponseFunctionImplementor, output_sensitivities: KratosOA.CollectiveVariableDataHolder):
```

applications/OptimizationApplication/python_scripts/controls/material_properties_control.py
```diff
@@ -0,0 +1,65 @@
...
+        sensitivity_variable_name = control_variable_name + "_SENSITIVITY"
+        if not Kratos.KratosGlobals.HasVariable(sensitivity_variable_name):
+            raise RuntimeError(f"Sensitivity variable {sensitivity_variable_name} for control variable {control_variable_name} not found.")
+        elif Kratos.KratosGlobals.GetVariableType(sensitivity_variable_name) != "Double":
+            raise RuntimeError(f"Sensitivity variable {sensitivity_variable_name} type is not double.")
+
+        self.control_variable = Kratos.KratosGlobals.GetVariable(control_variable_name)
+        self.sensitivity_variable = Kratos.KratosGlobals.GetVariable(sensitivity_variable_name)
+
+    def ExecuteInitialize(self):
```
sunethwarna:
```suggestion
    def Initialize(self):
```

applications/OptimizationApplication/python_scripts/controls/control.py
sunethwarna: Add `Initialize` and `Finalize` as `ABC` to base class

applications/OptimizationApplication/python_scripts/controls/material_properties_control.py
```diff
@@ -0,0 +1,65 @@
+import KratosMultiphysics as Kratos
+import KratosMultiphysics.OptimizationApplication as KratosOA
+from KratosMultiphysics.OptimizationApplication.controls.control import Control
+from KratosMultiphysics.OptimizationApplication.optimization_info import OptimizationInfo
+from KratosMultiphysics.OptimizationApplication.utilities.response_function_implementor import ResponseFunctionImplementor
+
+class MaterialPropertiesControl(Control):
```
sunethwarna: this goes under `material` folder

applications/OptimizationApplication/python_scripts/controls/shape_control.py
```diff
@@ -0,0 +1,46 @@
...
+from KratosMultiphysics.OptimizationApplication.utilities.execution_policy_decorator import ExecutionPolicyDecorator
+from KratosMultiphysics.OptimizationApplication.utilities.response_function_implementor import ResponseFunctionImplementor
+
+class ShapeControl(Control):
+    def __init__(self, model: Kratos.Model, parameters: Kratos.Parameters, optimization_info: OptimizationInfo):
+        super().__init__()
+
+        default_settings = Kratos.Parameters("""{
+            "model_part_names"            : [""],
+            "mesh_moving_analysis_name"   : ""
```
sunethwarna: This needs to put under `shape` folder...

applications/OptimizationApplication/python_scripts/controls/control.py
```diff
@@ -0,0 +1,21 @@
...
+        raise NotImplementedError("Calling base class Control::CalculateSensitivity method. Please implement it in the derived class.")
+
+    def UpdateControl(self, update: KratosOA.CollectiveVariableDataHolder):
+        raise NotImplementedError("Calling base class Control::UpdateControl method. Please implement it in the derived class.")
```
Igarizza: Check the method name in the message.

applications/OptimizationApplication/python_scripts/controls/material_properties_control.py
```diff
@@ -0,0 +1,65 @@
...
+        # creates element specific properties
+        if "model_parts_with_element_specific_properties" not in self.optimization_info.GetSolutionStepData(0).keys():
+            self.optimization_info["model_parts_with_element_specific_properties"] = []
+
+        for model_part in self.model_parts:
+            if not f"{model_part.FullName()}.Elements" in self.optimization_info["model_parts_with_element_specific_properties"]:
+                KratosOA.OptimizationUtils.CreateEntitySpecificPropertiesForContainer(model_part, model_part.Elements)
+                self.optimization_info["model_parts_with_element_specific_properties"].append(f"{model_part.FullName()}.Elements")
+
+    def CalculateSensitivity(self, response_function: ResponseFunctionImplementor, output_sensitivities: KratosOA.CollectiveVariableDataHolder):
```
Igarizza: Why does it call Calculate if you just get it from DataHolder and assign to output_sensitivities. Should we call it GetSensitivities? Where does the calculation happen?

applications/OptimizationApplication/python_scripts/controls/material_properties_control.py
Igarizza: I think the function should be called: UpdateControlVariables.

applications/OptimizationApplication/python_scripts/controls/shape_control.py
```diff
@@ -0,0 +1,46 @@
...
+        }""")
+        parameters.ValidateAndAssignDefaults(default_settings)
+
+        self.model_parts = [model[model_part_name] for model_part_name in parameters["model_part_names"].GetStringArray()]
+        self.mesh_moving_execution_policy_wrapper: ExecutionPolicyDecorator = optimization_info.GetOptimizationProcess(ExecutionPolicyDecorator, parameters["mesh_moving_analysis_name"].GetString())
+
+    def CalculateSensitivity(self, response_function: ResponseFunctionImplementor, output_sensitivities: KratosOA.CollectiveVariableDataHolder):
```
Igarizza: the same here as in Material.

applications/OptimizationApplication/python_scripts/controls/shape_control.py
```diff
@@ -0,0 +1,46 @@
...
+        # clear the container
+        output_sensitivities.ClearVariableDataHolders()
+
+        for model_part in self.model_parts:
+            d_j_d_x = KratosOA.NodalContainerVariableDataHolder(model_part)
+            response_function.GetStandardizedSensitivity(Kratos.SHAPE_SENSITIVITY, d_j_d_x)
+
+            output_sensitivities.AddVariableDataHolder(d_j_d_x)
+
+    def UpdateControl(self, update: KratosOA.CollectiveVariableDataHolder):
```
Igarizza: UpdateControlVariables

Igarizza (commented):
Few small comments. We may need to discuss the place for mesh motion execution to simplify the UpdateControl method and be more flexible.

applications/OptimizationApplication/python_scripts/controls/material_properties_control.py
sunethwarna: changed to `Update`

applications/OptimizationApplication/python_scripts/controls/control.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/controls/control.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/controls/material_properties_control.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/controls/material_properties_control.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/controls/shape_control.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/controls/control.py
sunethwarna: Changed to `Update`

applications/OptimizationApplication/python_scripts/controls/material_properties_control.py
sunethwarna: This is only a mapping, so name is changed to `MapFirstDerivative`

applications/OptimizationApplication/python_scripts/controls/shape_control.py
sunethwarna: done and changed.

applications/OptimizationApplication/python_scripts/controls/shape_control.py
sunethwarna: changed to `Update`

sunethwarna: changed to `Update`

sunethwarna:
I have changed according to suggestions from @RezaNajian @Igarizza @FMeister . Please have a look.... Thanks :D

Igarizza (approved):
good from my side

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
```diff
@@ -0,0 +1,89 @@
...
+    """Material properties control
+
+    This is a generic material properties control which creates a control
+    for the specified control variable. This does not do any filtering.
+
+    TODO: Extend with filtering techniques when they are implemented.
+
+    """
+    def __init__(self, model: Kratos.Model, parameters: Kratos.Parameters, optimization_info: OptimizationInfo):
+        super().__init__()
```
RezaNajian, line 17: There are couple of things that have not been considered in this implementation. Firstly, if there is filtering involved, we need to distinguish clearly between the phyiscal controlled variable like DENSITY or YOUNG_MODULUS and the real controls which in case of SIMP is sth like C_DENSITY (control density). So "control_variable_name" should be updated to "controlled_variable" which can be any physical propertity of material.

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
```diff
@@ -0,0 +1,89 @@
...
+
+        self.optimization_info = optimization_info
+        self.model_parts = [model[model_part_name] for model_part_name in parameters["model_part_names"].GetStringArray()]
+
+        control_variable_name = parameters["control_variable_name"].GetString()
+        control_variable_type = Kratos.KratosGlobals.GetVariableType(control_variable_name)
+        if control_variable_type != "Double":
+            raise RuntimeError(f"{control_variable_name} with {control_variable_type} type is not supported. Only supports double variables")
+
+        sensitivity_variable_name = control_variable_name + "_SENSITIVITY"
```
RezaNajian, line 33: control does not have anything to do with sensitivties and if filtering is involved it only maps them. So self.sensitivity_variable should be removed !

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
```diff
@@ -0,0 +1,89 @@
...
+        if not problem_data_container.HasValue("model_parts_with_element_specific_properties"):
+            problem_data_container["model_parts_with_element_specific_properties"] = []
+
+        for model_part in self.model_parts:
+            # check if element specific properties are created for the model part by a different process.
+            if not f"{model_part.FullName()}.Elements" in problem_data_container["model_parts_with_element_specific_properties"]:
+                KratosOA.OptimizationUtils.CreateEntitySpecificPropertiesForContainer(model_part, model_part.Elements)
+                problem_data_container["model_parts_with_element_specific_properties"].append(f"{model_part.FullName()}.Elements")
+
+    def GetRequiredSensitivityFieldVariables(self) -> 'list[SupportedSensitivityFieldVariableTypes]':
```
RezaNajian, line 54: control does not control anything about sensitivties, it only maps the sensitivties, for the sake of consistency sensitivty stuff like this functions should be removed

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
```diff
@@ -0,0 +1,89 @@
...
+        return [self.sensitivity_variable]
+
+    def GetCollectiveExpression(self) -> KratosOA.ContainerExpression.CollectiveExpressions:
+        control_collective_expression = KratosOA.ContainerExpression.CollectiveExpressions()
+        for model_part in self.model_parts:
+            control_collective_expression.Add(KratosOA.ContainerExpression.ElementPropertiesExpression(model_part))
+
+        return control_collective_expression
+
+    def MapFirstDerivative(self, sensitivity_variable_first_derivative_map: 'dict[SupportedSensitivityFieldVariableTypes, KratosOA.ContainerExpression.CollectiveExpressions]', mapped_first_derivative: KratosOA.ContainerExpression.CollectiveExpressions):
```
RezaNajian, line 64: we need to simplify the inputs. The input to this function should only be the name of the sensitivty field and the rest like  CollectiveExpressions should be retrived from OptimizationInfo and data there. Furthermore, it is control's responsibility to decide about the name of the mapped field. If filtering is used the naming convention in this application is sensitivty_field_name_C_DENSITY for example. 
Please also note that SHAPE_SENSITIVITY, DENSITY_SENSITIVITY, ... are only means to manage or skip registraion of new fields in kratos. So the input for this function is not at all SHAPE_SENSITIVITY, ... . It should be rather exactly the name of the sensity field like D_STRAIN_ENERGY_D_SHAPE and it can be retreived from optimization info and data using experesions and so on ..

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
```diff
@@ -0,0 +1,89 @@
...
+        mapped_first_derivative.Clear()
+
+        for provided_container_expression in sensitivity_variable_first_derivative_map[self.sensitivity_variable].GetContainerExpressions():
+            # we are cloning in here because, it is not sure what the algorithm will
+            # do with the sensitivity container expressions. This is just a move operation
+            # of the pointers the expression is holding, hence not at all expensive operation.
+            # it does not also clone the memory space.
+            mapped_first_derivative.Add(provided_container_expression.Clone())
+
+    def Update(self, update: KratosOA.ContainerExpression.CollectiveExpressions) -> None:
```
RezaNajian, line 78: there is no need for any input here because every control knows how to uodate himself. it knows which field should be picked up for update and how to update. So the input should be removed !

applications/OptimizationApplication/python_scripts/controls/shape/shape_control_new.py
```diff
@@ -0,0 +1,88 @@
+import KratosMultiphysics as Kratos
+import KratosMultiphysics.OptimizationApplication as KratosOA
```
RezaNajian, line 2: Here  I would put the same comments i had about material control. Plus the follwoing:
this class should serve as the base class of shape control and not necessary mesh motion is involved. We might have shells even and mesh motion is only used for explicit surface filtering. So please make this class as general as possible.

applications/OptimizationApplication/tests/test_OptimizationApplication.py
```diff
@@ -26,6 +26,8 @@
 import test_collective_expressions
 import test_buffered_dict
 import test_standardized_responses
+import test_material_properties_control
```
RezaNajian, line 29: please make a seprate folder and sub folders for control and shape and material controls for the sake of maintaibaility and strcutrure.

applications/OptimizationApplication/tests/test_material_properties_control.py
```diff
@@ -0,0 +1,77 @@
+
+import KratosMultiphysics as Kratos
```
RezaNajian, line 2: please move it to folder like test_controls

RezaNajian (changes requested):
Thank you, @sunethwarna, for your work. However, this pull request is one of the most important changes to the OptimizationApplication, and it should be thoroughly examined. I believe we need to adjust the current controls to the suggested changes now, otherwise it could get out of control. We can use placeholders for now that can be removed later, but the controls should be adjusted and integrated as much as possible.

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
sunethwarna: This is purely without filtering. As we discussed, filtering is part of the control, hence they can implement their own control with different filtering mechanisms. They don't necessarily have to inherit from this.

They only have to be inherited from the base abstract `Control` class.

Thats what I had in mind. This is a really primitive example for control. (Even I cannot still use it without any kind of filtering.)

As far as I see, filtering techniques are different for different controls, hence there is no point making a general base class for material controls. We can have different material control classes which are inherited from base `Control` class which is abstract.

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
sunethwarna: I understand this but I don't have a solution for the following question.... How can the algorithm know which control requires which sensitivities to be taken from response function?

Current situation is:
We have `Response`, `Control` and `Algorithm`. Algorithm needs to know which sensitivities to be requested from response which is need to be passed to control. How can it know? Response only knows all the types of variables which it has dependence on. Control knows about the updating and its mechanisms, and what it requires to update. Thats why in current design I put required sensitivities in `Control`.

Following are the alternatives which came to my mind, (none of them are perfect).

My solutions:
1. Having a map of control variables and sensitivities (a global map) -> I don't like maintaining maps (error prone, hard to debug)
2. Requesting from control what sensitivities are required from their side. -> Current design
3. Passing the response as an input to the control so control can calculate what ever is required from the sensitivities. -> What I have proposed previously. -> I don't like the idea of passing resposne to the control. (Interface becomes entangled).

I am really open for suggestions since none of my solutions are the ideal.

sunethwarna: We need to discuss this as explained earlier. I don't like this either, but this was the least bad solution which I thought.

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
sunethwarna: As mentioned above.

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
sunethwarna: As far as I see, the control update is an outcome of the `Algorithm` ryt? In that case, only algorithm knows how much is the update ryt? So we still need this ryt?

applications/OptimizationApplication/python_scripts/controls/shape/shape_control_new.py
sunethwarna: I was thinking, this won't be a base a base class for `ShapeControl`. Because, as you mentioned, filtering techniques and whatnot will be different for different controls. So we don't need a base class, We can implement different shape controls with the abstract `Control` base class. If we see code duplication, we can move the duplicated code to a central place in shape control folder.

applications/OptimizationApplication/tests/test_OptimizationApplication.py
sunethwarna: will do :)

applications/OptimizationApplication/tests/test_material_properties_control.py
sunethwarna: will do :)

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
sunethwarna: First, there is no need to have variables like `sensitivty_field_name_C_DENSITY` due to `CollectiveExpressions` and `ContainerExpression`. We can put any arbitrary name now to any field (and with @FMeister impementations, we can output them to VTK as well, I will take care of the HDF5 output of `ContainerExpression`).

Second, If I am understanding this correctly, I don't like some arbitrary method/class accessing `OptimizationInfo` data which was not written by that method/class. This is a nightmare to debug, and hard to follow who writes what. Thats why we need to always communicate between objects using interfaces not string messages (`OptimizationInfo` is a string based data container).

Thirdly, My intention was not to make a base class for `MaterialControl`. As I stated in other comments, controls can have varying stuff, so they can have their own classes inherited from balnk abstract class `Control`. If we see code duplication, then we can move the duplicated code to a central place and start re-using. Thats what I had in mind when I introduced these. In this case, once the proper filtering techniques are implemented, we have to implement controls with different filtering techniques which will have their own classes.

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
Igarizza: You are both right in some way. @sunethwarna, the Algorithm computes the updates; that is 100% sure. I guess @RezaNajian proposes that Control can get the updates from Algorithm on its own, as it knows which field/results it needs. My question is if it is a good design to allow Control access to the Algorithm data or if an Algorithm can directly push the computed results to different Controls, that they expect.

sunethwarna:
This was completed with the new seperated PR. Hence closing.

PR #10778, opened 2023-02-17
[OptApp] Restructure 14 - Output mechanisms for OptimizationApplication

sunethwarna:
**📝 Description**
This adds output mechanisms to optimization application.

1. It introduces a process (`OptimizationInfoAsciiWriter`) which can write any ascii value in the `OptimizationInfo` to a log file. This is used to update the optimization summary file for each design iteration.
2. It introduces a process (`ContainerVariableDataHolderDecorator`) using the decorator pattern so, it can convert any `ContainerVariableDataHolder` to its corresponding nodal, condition, elemental data container variable values. Hence this allows all the `ContainerVariableDataHolders` to be used with any exisiting `Kratos::Process` or `Kratos::OutputProcess`. One example is, using this to output sensitivities to VTK/HDF5, or using `StatisticsApplication` to calculate spatial/iteration statistics for different containers and variables.

Todo:
- [ ] Add tests

**🆕 Changelog**
- Added `OptimizationInfoAsciiWriter`
- Added `ContainerVariableDataHolderDecorator`


applications/OptimizationApplication/python_scripts/output/container_variable_data_holder_output_decorator.py
```diff
@@ -0,0 +1,228 @@
...
+
+class ContainerVariableDataHolderInfo(ABC):
+    @classmethod
+    @abstractmethod
+    def GetDefaultParameters(cls) -> Kratos.Parameters:
+        pass
+
+    @classmethod
+    @abstractmethod
+    def GetPattern(cls) -> str:
```
matekelemen, line 23: some docs would be very useful, especially on the base class methods

applications/OptimizationApplication/python_scripts/output/output_utilities.py
```diff
@@ -0,0 +1,60 @@
...
+
+    results: 'list[str]' = []
+    for matched_groups in matched_groups_list:
+        current_result = ", ".join([f"{place_holder[1:-1]} = {value}" for place_holder, value in zip(place_holders, matched_groups.groups())])
+        results.append(current_result)
+
+    return sorted(list(set(results)))
+
+def GetPrefixWithoutDataName(pattern: str, settings: Kratos.Parameters) -> str:
+    place_holders = GetPlaceHolders(pattern)
```
matekelemen, line 52: [placeholder](https://www.dictionary.com/browse/placeholder) is a single word

applications/OptimizationApplication/python_scripts/output/output_utilities.py
```diff
@@ -0,0 +1,60 @@
...
+            elif required_type is None or isinstance(v, required_type):
+                valid_keys.append(f"{current_prefix}/{k}")
+
+    valid_keys = []
+    __get_valid_keys(valid_keys, "", optimization_info.GetSolutionStepData(0))
+    valid_keys = [valid_key[1:] for valid_key in valid_keys]
+    return valid_keys
+
+def GetPlaceHolders(pattern: str) -> 'list[str]':
+    return ["<" + place_holder_start[:place_holder_start.find(">")+1] for place_holder_start in pattern.split("<")][1:]
```
matekelemen, line 20: this will fail on unmatched `<>` pairs and on nested ones too. You're already using regex in a couple of places so you might as well do so here as well, with something like `<(\w*)>`.

Also, this placeholder business seems to pop up fairly frequently across applications so you may as well implement a common solution in core (in another PR ofc). Here's what I came up with a while ago if you're interested:
- [interface](https://github.com/matekelemen/WRApp/blob/68a8b70d06b9e65c669fbbf7faa7e473adc545f8/packages/wrapp/utils/inc/PatternUtility.hpp)
- [impl](https://github.com/matekelemen/WRApp/blob/68a8b70d06b9e65c669fbbf7faa7e473adc545f8/packages/wrapp/utils/src/PatternUtility.cpp)

applications/OptimizationApplication/python_scripts/output/container_variable_data_holder_output_decorator.py
```diff
@@ -0,0 +1,228 @@
+from abc import ABC
+from abc import abstractmethod
```
matekelemen, line 2:
```suggestion
from abc import ABC, abstractmethod
```
same for other multiple imports

applications/OptimizationApplication/python_scripts/output/container_variable_data_holder_output_decorator.py
```diff
@@ -0,0 +1,228 @@
...
+from KratosMultiphysics.process_factory import KratosProcessFactory
+from KratosMultiphysics.OptimizationApplication.optimization_info import OptimizationInfo
+from KratosMultiphysics.OptimizationApplication.utilities.helper_utilities import CallOnAll
+from KratosMultiphysics.OptimizationApplication.utilities.helper_utilities import ContainerVariableDataHolderUnion
+from KratosMultiphysics.OptimizationApplication.output.output_utilities import GetOptimizationInfoAvailableKeysForType
+from KratosMultiphysics.OptimizationApplication.output.output_utilities import GetPatternMatchedOptimizationKeys
+from KratosMultiphysics.OptimizationApplication.output.output_utilities import GetPrefixWithoutDataName
+from KratosMultiphysics.OptimizationApplication.output.output_utilities import GetPlaceHolderWithValues
+
+class ContainerVariableDataHolderInfo(ABC):
```
matekelemen, line 15: As a general remark, docstrings that provide an overview/list of features of what a class/function is responsible for would be welcome. This is even more important for base classes like this and its virtual members. 

Reviewing is difficult, but understanding/refactoring/extending the entire system later on is nearly impossible without them.

applications/OptimizationApplication/python_scripts/output/container_variable_data_holder_output_decorator.py
```diff
@@ -0,0 +1,228 @@
...
+    def GetDefaultParameters(cls) -> Kratos.Parameters:
+        return Kratos.Parameters("""{
+            "type"           : "control_data",
+            "model_part_name": "PLEASE_PROVIDE_MODEL_PART_NAME",
+            "container_type" : "PLEASE_PROVIDE_CONTAINER_TYPE",
+            "control_name"   : "PLEASE_PROVIDE_CONTROL_NAME",
+            "data_list"      : [
+                {
+                    "data_name"     : "PLEASE_PROVIDE_DATA_NAME",
+                    "variable_name" : "PLEASE_PROVIDE_VARIABLE_NAME"
```
matekelemen, line 145: this isn't so important, but I prefer empty defaults to invalid ones. If a default value is non-empty, it should be valid.

Also, giving an example of what kind of items you're expecting in `"data_list"` is neat, but it should be stated in the docstrings rather than the defaults. (Even better would be a system that generates valid defaults given a model, but that's out of scope for this PR).

applications/OptimizationApplication/python_scripts/output/container_variable_data_holder_output_decorator.py
```diff
@@ -0,0 +1,228 @@
...
+class ContainerVariableDataHolderOutputDecorator(Kratos.OutputProcess):
+    def __init__(self, model: Kratos.Model, parameters: Kratos.Parameters, optmization_info: OptimizationInfo):
+        super().__init__()
+
+        default_parameters = Kratos.Parameters("""{
+            "module"                           : "KratosMultiphysics.OptimizationApplication.utilities",
+            "type"                             : "ContainerVariableDataHolderOutputDecorator",
+            "output_processes"                 : [],
+            "optimization_info_output_settings": []
+        }""")
```
matekelemen, line 163: I think `Process` has `GetDefaultParameters` so I think it's better to override it rather than defining it in local scope.

applications/OptimizationApplication/python_scripts/output/container_variable_data_holder_output_decorator.py
```diff
@@ -0,0 +1,228 @@
...
+    def ExecuteFinalizeSolutionStep(self):
+        CallOnAll(self.__output_processes_list, Kratos.OutputProcess.ExecuteFinalizeSolutionStep)
+
+    def ExecuteFinalize(self):
+        CallOnAll(self.__output_processes_list, Kratos.OutputProcess.ExecuteFinalize)
+
+    def IsOutputStep(self) -> bool:
+        for output_process in self.__output_processes_list:
+            if output_process.IsOutputStep():
+                return True
```
matekelemen, line 210:
```suggestion
        if any(output_process.IsOutputStep() for output_process in self.__output_processes_list):
            return True
```

also, this triggers the output for all processes in `self.__output_processes_list` even if only one them returns true from `IsOutputStep`. Is that intended?

applications/OptimizationApplication/python_scripts/output/container_variable_data_holder_output_decorator.py
```diff
@@ -0,0 +1,228 @@
...
+        return True
+
+    def PrintOutput(self):
+        # write containers to respective model parts
+        for container_variable_data_holder_info in self.__list_of_data_sets:
+            container_variable_data_holder_info.ApplyDataConainer()
```
matekelemen, line 216:
```suggestion
            container_variable_data_holder_info.ApplyDataContainer()
```
typo

applications/OptimizationApplication/python_scripts/output/output_utilities.py
```diff
@@ -0,0 +1,60 @@
...
+        place_holders.append("<data_name>")
+
+    # replace problamatic "/"
+    regex_pattern = regex_pattern.replace("/", "\/")
+
+    # get the matching list
+    groups_list = [re.match(regex_pattern, key) for key in available_keys]
+    matched_groups_list = [matched_group for matched_group in groups_list if not matched_group is None]
+
+    results: 'list[str]' = []
```
matekelemen, line 44: you eventually want to return a sorted list of unique items, so it would make sense to collect `results` into a `set` instead of a `list` right away.

applications/OptimizationApplication/python_scripts/output/output_utilities.py
```diff
@@ -0,0 +1,60 @@
...
+    for matched_groups in matched_groups_list:
+        current_result = ", ".join([f"{place_holder[1:-1]} = {value}" for place_holder, value in zip(place_holders, matched_groups.groups())])
+        results.append(current_result)
```
matekelemen, line 47: if a `place_holder` is duplicated in the input pattern and its values don't match, the output list will have unpredictable items.

Example: 
`<oopsie>_what_<daisy>_ever` => `<data_name>_what_<data_name>_ever`

applications/OptimizationApplication/python_scripts/output/optimization_info_ascii_writer.py
```diff
@@ -0,0 +1,300 @@
...
+
+import KratosMultiphysics as Kratos
+from KratosMultiphysics.OptimizationApplication.optimization_info import OptimizationInfo
+from KratosMultiphysics.time_based_ascii_file_writer_utility import TimeBasedAsciiFileWriterUtility
+from KratosMultiphysics.OptimizationApplication.output.output_utilities import GetOptimizationInfoAvailableKeysForType
+from KratosMultiphysics.OptimizationApplication.output.output_utilities import GetPatternMatchedOptimizationKeys
+from KratosMultiphysics.OptimizationApplication.output.output_utilities import GetPrefixWithoutDataName
+from KratosMultiphysics.OptimizationApplication.output.output_utilities import GetPlaceHolderWithValues
+
+class OptimizationAsciiInfo(ABC):
```
sunethwarna, line 14: Restrictions:

There should be three categories to read from `ResponseData`, `ControlData`, `AlgorithmData`. The defaults list of strings should be provided for each category.


```json
            "optimization_info_output_settings": [
                "response_data_names": ["this_name"],
                "control_data_names": [],
                "algorithm_data_names": [],
               "precision": "0.4e"
            ]
```

applications/OptimizationApplication/python_scripts/output/container_variable_data_holder_output_decorator.py
```diff
@@ -0,0 +1,228 @@
+from abc import ABC
+from abc import abstractmethod
+
+import KratosMultiphysics as Kratos
```
sunethwarna, line 4: Need to have a different VTKOutput mechanism for `ContainerVariableData`. Need to add the same for `HDF5`.

PR #10779, opened 2023-02-20
No longer create a symlink to the compilation database

avdg81:
**📝 Description**
Creating a symlink on Windows requires elevated rights. As a result, the following error was raised by CMake when a regular user attempted to build Kratos on a Windows platform:
```
CMake Error: failed to create symbolic link
'C:/path/to/kratos/source/compile_commands.json': A required privilege
is not held by the client.
```
It was triggered by an attempt to create a symlink in the source directory that points to file `compile_commands.json` in the build directory.

**🆕 Changelog**
In consultation with the original author of the CMake code involved (@matekelemen) we have decided to remove the command that attempts to create the symlink, since it is no longer needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,10 +71,6 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${KRATOS_SOURCE_DIR}/cmake_modules")
 
 # Generate and copy configuration for language servers
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
-EXECUTE_PROCESS(
-    COMMAND ${CMAKE_COMMAND} -E create_symlink
-    "${CMAKE_BINARY_DIR}/compile_commands.json"
-    "${CMAKE_SOURCE_DIR}/compile_commands.json")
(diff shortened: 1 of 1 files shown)


matekelemen (approved):
Yep, that's my mistake. At the time I didn't know how to configure VSCode's [clangd extension](https://marketplace.visualstudio.com/items?itemName=llvm-vs-code-extensions.vscode-clangd) to point to the build directory, where `compile_commands.json` is written during the build process. Even worse, I had no idea that symlinking required elevated privileges on Windows (which is something I'm still baffled about).

Now I'm a bit wiser and learned [how to configure clangd](https://stackoverflow.com/a/64886320/12350793), so this is no longer necessary.

roigcarlo:
:+1: Good to me.

For the record, we have the same problem if we try to create symlinks for *.py files for fast prototyping so we just disabled that by default in win and added a warning:

```ps2
rem Warning: In windows this option only works if you run through a terminal with admin privileges
rem set KRATOS_INSTALL_PYTHON_USING_LINKS=ON
```

PR #10785, opened 2023-02-21
[TrilinosApplication] Consistent local conditions with local elements in B&S

loumalouomega:
**📝 Description**

In `SetUpDofSet` in Trilinos block B&S the local elements are used, but not local conditions. This PR makes that consistent.

**🆕 Changelog**

- [Consistent local conditions with local elements](https://github.com/KratosMultiphysics/Kratos/commit/afbe2aca7a5c3c680827d3a9c8eb46a08c72c9bc)


rubenzorrilla (approved):
If I'm not wrong, there shouldn't be any difference considering our MPI implementation.

loumalouomega:
> If I'm not wrong, there shouldn't be any difference considering our MPI implementation.

Yes, but it is inconsistent :)

PR #10792, opened 2023-02-22
[GeoMechanicsApplication] Geo/create parameter field

### LINKED ISSUE #10651: [GeomechanicsApplication] Make it possible to initialize a State parameter field in the Geomechanics application
Make it possible to initialize a Material or State parameter field in the Geomechanics application. 

- Implement a method which makes it possible to initialize material state parameters per gauss point (not only per material). 
- Example applications are random fields and depth dependent Overconsolidation Ratio

aronnoordam:
Added possibility to create a field of parameters where the value of the parameter is different on each element within a modelpart. This parameter field is set during the initialisation phase (when application is restarted)

also added possibility to show parameter values on the integration points, as required for the GID output

3 possibilities to generate a parameter field are give:

  | option 1, func_type = 'input': with this option, a parameter field is generated based on a function which is
  directly written in the projectparameters.json at the 'function' parameter. This function can depend on
  the x, y and z coordinate.

  | option 2, func_type = 'python': with this option, a parameter field is generated, using a user defined python
  script. This python script has to be inherited from the 'ParameterFieldBase' class. Which is located at:
  'GeoMechanicsApplication->python_scripts->user_defined_scripts->user_defined_parameter_field_base.py'
  the name of the script (without '.py') should be filled in at the 'function' parameter in the projectparameters.json

  | option 3, func_type = 'json': with this option, a parameter field can be directly read from a json dictionary.
  This dictionary has to contain the 'values' key, which is a 1D list of all the field values. The list has to have
  the same size as the elements within the model part, and need to be accordingly sorted. The filename should be
  filled in at the 'dataset' parameter, within the projectparameters.json


applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_element.cpp
```diff
@@ -692,6 +693,16 @@ void UPwSmallStrainElement<TDim,TNumNodes>::
...
     }
+    else if (r_prop.Has(rVariable))
+    {
+        // map initial material property to gauss points, as required for the output 
+        if (rOutput.size() != mConstitutiveLawVector.size())
+            rOutput.resize(mConstitutiveLawVector.size());
+
+        for (unsigned int i = 0; i < rOutput.size(); ++i) {
+            rOutput[i] = r_prop.GetValue(rVariable);
+        }
```
avdg81: I think we might be able to express this in a more direct and slightly clearer way:
```suggestion
        rOutput.clear();
        std::fill_n(std::back_inserter(rOutput), mConstitutiveLawVector.size(), r_prop.GetValue(rVariable));
```
This says we first discard all values present in `rOutput` and then fill it with the same value (i.e. `r_prop.GetValue(rVariable)`) a certain number of times (i.e. `mConstitutiveLawVector.size()`).

Alternatively, you could resize the vector like you did and then assign the same value to all the elements in the vector:
```suggestion
        rOutput.resize(mConstitutiveLawVector.size());
        std::fill(rOutput.begin(), rOutput.end(), r_prop.GetValue(rVariable));
```
Note that there is no need to explicitly check for `rOutput`'s size upfront. The `resize` member function will do that for you. See [this page on cppreference.com](https://en.cppreference.com/w/cpp/container/vector/resize).

The take-away here is that you are encouraged to avoid "raw" `for`-loops and replace them by the intended STL algorithms. `std::fill` (but also `std::fill_n`) is one of them. If you'd like to learn more about why this is good practice search for "Sean Parent no raw loops".

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,143 @@
...
+#include "geo_mechanics_application_variables.h"
+
+
+namespace Kratos
+{
+
+    SetParameterFieldProcess::SetParameterFieldProcess(ModelPart& rModelPart,
+                                                            Parameters Settings)
+                                                            : mrModelPart(rModelPart),
+                                                            mParameters(Settings)
```
avdg81: Since class `SetParameterField` is derived from class `Process` we have to name the base class constructor explicitly:
```suggestion
    SetParameterFieldProcess::SetParameterFieldProcess(ModelPart& rModelPart,
                                                            Parameters Settings)
                                                            : Process(),
                                                              mrModelPart(rModelPart),
                                                              mParameters(Settings)
```

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,143 @@
...
+        {
+            "help"            : "This process applies a moving load condition belonging to a modelpart. The load moves over line elements.",
+            "model_part_name" : "please_specify_model_part_name",
+            "variable_name"   : "CUSTOM",
+            "func_type"       : "input",               
+            "function"        : "0",
+            "dataset"         : "dummy"
+        }  )"
+    );
+    Parameters mParameters;
```
avdg81: Note that this new object has the same name as one of your data members. As a result, you are _shadowing_ the data member. That is probably not what you intended. In that case, please remove this line as suggested below. Otherwise, I would strongly recommend you to pick some other name to avoid the shadowing. By the way, Clang-Tidy warns you about this kind of problem.
```suggestion
```

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,143 @@
...
+    mParameters.RecursivelyValidateAndAssignDefaults(default_parameters);
+
+}
+
+
+void SetParameterFieldProcess::SetValueAtElement(Element& rElement, const Variable<double>& rVar, double Value)
+{
+
+
+    Properties& r_prop = rElement.GetProperties();
```
avdg81, line 56: Since you don't intend to use this reference for modifying the referenced object, I would prefer to make this a reference-to-const. In addition, I would also use `auto` in this case since it is better to program against interfaces then against concrete types:
```suggestion
    const auto& r_prop = rElement.GetProperties();
```

In fact, I would suggest to follow the so-called AAA Style (Almost Always Auto). Herb Sutter explains this style [here](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/) and he also provides some strong arguments in favor of it. [Herb Sutter is one of the members of the C++ Technical Committee.]

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,143 @@
...
+}
+
+
+void SetParameterFieldProcess::ExecuteInitialize()
+{
+    KRATOS_TRY
+
+    if (!this->mrModelPart.GetProcessInfo()[IS_RESTARTED]){
+
+        const Variable<double>& r_var = KratosComponents< Variable<double> >::Get(mParameters["variable_name"].GetString());
```
avdg81: This is a nice example that demonstrates why `auto` is useful. As the code is now, you are forced to repeat the object type at the left hand side, whereas the type information is also provided on the right hand side (`KratosComponents<Variable<double>>`). By using `auto` you eliminate this duplication.

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+        // set parameter field from input function
+        if (mParameters["func_type"].GetString().compare("input") == 0)
```
avdg81: Is there any particular reason why we should use `compare` rather than `operator==`? I feel that the latter is a little easier  to read:
```suggestion
        if (mParameters["func_type"].GetString() == "input")
```

This comment also applies to the other two string comparisons further down.

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+                auto& r_geom = r_element.GetGeometry();
+
+                // calculate parameter value at current element
+                const double val = parameter_function.CallFunction(r_geom.Center().X(), r_geom.Center().Y(), r_geom.Center().Z(), current_time, 0, 0, 0);
+                
+
+                this->SetValueAtElement(r_element, r_var, val);
+
+            }
```
avdg81: You may consider to move this code to a separate (private) member function (e.g. `SetParameterFieldUsingInputFunction`). This would have a positive impact in two ways: (1) the current function body becomes much shorter (and it gets more to the point), and (2) you no longer need the comment above this block since the function name is self-explanatory. If you agree, please apply this also to the other two blocks of this selection statement.

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,143 @@
...
+        }
+        // set parameter field with a python function
+        else if (mParameters["func_type"].GetString().compare("python") == 0)
+        {
+
+            // get new data from the data set
+            const std::string dataset = mParameters["dataset"].GetString();
+            Parameters new_data = Parameters(dataset);
```
avdg81: No need to use the copy constructor here. The simplest way to achieve what you want is probably this:
```suggestion
            Parameters new_data{dataset};
```

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,143 @@
...
+                this->SetValueAtElement(r_element, r_var, val);
+
+            }
+        }
+        // set parameter field with a python function
+        else if (mParameters["func_type"].GetString().compare("python") == 0)
+        {
+
+            // get new data from the data set
+            const std::string dataset = mParameters["dataset"].GetString();
```
avdg81: By using `auto` you would avoid repeating the type (which is already clear from the call to `GetString()`). You might be able to improve this even further by letting the variable name reflect what the string represents, e.g. `dataset_filename` when it is a filename.
```suggestion
            const auto dataset = mParameters["dataset"].GetString();
```

This comment also applies to other lines of code further down. I won't repeat it for each and every occurrence. You know what to look for :smile:

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.hpp
```diff
@@ -0,0 +1,124 @@
...
+#include "utilities/mortar_utilities.h"
+namespace Kratos {
+
+///@name Kratos Globals
+///@{
+
+///@name Kratos Classes
+///@{
+
+
```
avdg81, line 35: These comments and blank lines seem redundant to me. Note that there are more such redundant comments. By removing them it becomes easier to grasp the essence of the code.

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.hpp
```diff
@@ -0,0 +1,124 @@
...
+public:
+    ///@name Type Definitions
+    ///@{
+
+    ///@}
+    ///@name Pointer Definitions
+    /// Pointer definition of SetMovingLoadProcess
+    KRATOS_CLASS_POINTER_DEFINITION(SetParameterFieldProcess);
+
+    typedef std::size_t SizeType;
```
avdg81: Please follow "T.43: Prefer `using` over `typedef` for defining aliases" from the [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md). It improves readability.
```suggestion
    using SizeType = std::size_t;
```

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.hpp
```diff
@@ -0,0 +1,124 @@
...
+     */
+    void ExecuteInitialize() override;
+
+    
+    ///@}
+    ///@name Input and output
+    ///@{
+
+    /// Turn back information as a string.
+    virtual std::string Info() const override {
```
avdg81: No need to repeat `virtual` when you use `override`.
```suggestion
    std::string Info() const override {
```

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.hpp
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /// Print information about this object.
+    void PrintInfo(std::ostream& rOStream) const override {
+        rOStream << "SetParameterFieldProcess";
+    }
+
+    /// Print object's data.
+    void PrintData(std::ostream& rOStream) const override {
+    }
```
avdg81: This one is identical to the base class' implementation (`Process::PrintData`). Therefore, there is no need to override it here.

avdg81 (commented):
I have primarily looked at the C++ code from a software engineering point of view. As such I have not verified the physics that you are trying to simulate. Nor have I verified whether the Python scripts are correct.

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
Vahid-Galavi: Process() is the default constructor of the base class. The default constructor is implicitly is called in C++ when the constructor of the derived class is called. So, in my opinion, there is no need to add Process() here but adding it does not hurt.

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+#include "utilities/interval_utility.h"
+#include "geo_mechanics_application_variables.h"
+
+
+namespace Kratos
+{
+
+    SetParameterFieldProcess::SetParameterFieldProcess(ModelPart& rModelPart,
+                                                       const Parameters Settings)
```
Vahid-Galavi: A copy is made here. I think you mean `const Parameters& Settings`

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,155 @@
...
+    const std::string dataset = mParameters["dataset"].GetString();
+    const Parameters new_data{ dataset };
+    const Vector data_vector = new_data["values"].GetVector();
+
+    // set new data on the elements
+    IndexType i = 0;
+    for (Element& r_element : mrModelPart.Elements())
+    {
+        SetValueAtElement(r_element, rVar, data_vector[i]);
+        i++;
```
Vahid-Galavi: You can compact these  two lines:
```suggestion
        SetValueAtElement(r_element, rVar, data_vector[i++]);
```

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,155 @@
...
+        SetValueAtElement(r_element, rVar, val);
+
+    }
+}
+
+
+void SetParameterFieldProcess::SetParameterFieldUsingPythonFunction(const Variable<double>& rVar)
+{
+    // get new data from the data set
+    const std::string dataset = mParameters["dataset"].GetString();
```
Vahid-Galavi:
```suggestion
    const std::string& dataset = mParameters["dataset"].GetString();
```

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,155 @@
...
+    const Parameters new_data{ dataset };
+    const Vector data_vector = new_data["values"].GetVector();
```
Vahid-Galavi:
```suggestion
    const Vector& data_vector = new_data["values"].GetVector();
```

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,155 @@
...
+void SetParameterFieldProcess::SetParameterFieldUsingInputJson(const Variable<double>& rVar)
+{
+    // Read json string in field parameters file, create Parameters
+    const std::string& field_file_name = mParameters["dataset"].GetString();
+
+    KRATOS_ERROR_IF_NOT(std::filesystem::exists(field_file_name)) << "The parameter field file specified with name \"" << field_file_name << "\" does not exist!" << std::endl;
+
+    std::ifstream ifs(field_file_name);
+    Parameters new_data(ifs);
+    Vector data_vector = new_data["values"].GetVector();
```
Vahid-Galavi:
```suggestion
    const Vector& data_vector = new_data["values"].GetVector();
```

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+    KRATOS_ERROR_IF_NOT(data_vector.size() == mrModelPart.Elements().size()) << "The parameter field: \""
+        << field_file_name << "\" does not have the same size as the amount of elements within the model part!" << std::endl;
+
+    IndexType i = 0;
+    for (Element& r_element : mrModelPart.Elements())
+    {
+        SetValueAtElement(r_element, rVar, data_vector[i]);
+        i++;
```
Vahid-Galavi:
```suggestion
        SetValueAtElement(r_element, rVar, data_vector[i++]);
```

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.hpp
```diff
@@ -0,0 +1,139 @@
...
+    ///@name Pointer Definitions
+    /// Pointer definition of SetMovingLoadProcess
+    KRATOS_CLASS_POINTER_DEFINITION(SetParameterFieldProcess);
+
+    ///@}
+    ///@name Life Cycle
+    ///@{
+
+    SetParameterFieldProcess(ModelPart& rModelPart,
+                                  Parameters Parameters);
```
Vahid-Galavi:
```suggestion
                                  const Parameters& Parameters);
```

Vahid-Galavi (commented):
Great job!
I just have some minor remarks

Vahid-Galavi (dismissed):

aronnoordam:
@avdg81 @Vahid-Galavi , Thanks for the review both! After internal discussion i removed the inline increment again for readability purposes

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,155 @@
...
+                                                       const Parameters& rSettings)
+                                                            : Process(), mrModelPart(rModelPart),
+                                                            mParameters(rSettings)
+{
+    // function type: python, cpp, input
+    const Parameters default_parameters(R"(
+        {
+            "help"            : "This process applies a moving load condition belonging to a modelpart. The load moves over line elements.",
+            "model_part_name" : "please_specify_model_part_name",
+            "variable_name"   : "CUSTOM",
```
mcgicjn2: Is this correct ? It appears to refer to your moving load implementation

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,155 @@
...
+        // calculate parameter value at current element
+        const double val = parameter_function.CallFunction(r_geom.Center().X(), r_geom.Center().Y(), r_geom.Center().Z(), current_time, 0, 0, 0);
+
+        SetValueAtElement(r_element, rVar, val);
+
+    }
+}
+
+
+void SetParameterFieldProcess::SetParameterFieldUsingPythonFunction(const Variable<double>& rVar)
```
mcgicjn2: To Kratos is this a string of pre-calculated values ? Thus the term PythonFunction should not be in the name, as the mParameters can be generated by any code base

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,155 @@
...
+    IndexType i = 0;
+    for (Element& r_element : mrModelPart.Elements())
+    {
+        SetValueAtElement(r_element, rVar, r_data_vector[i]);
+        ++i;
+    }
+}
+
+
+void SetParameterFieldProcess::SetParameterFieldUsingInputJson(const Variable<double>& rVar)
```
mcgicjn2: Could you combine these this method and the one above in a better way, i.e. a parsing \ an assigning mechanism, thus reducing the duplications

applications/GeoMechanicsApplication/python_scripts/set_parameter_field_process.py
```diff
@@ -0,0 +1,111 @@
...
+
+            # initialise input and output
+            input_dict = {}
+            return_dict = {}
+
+            values = []
+            all_coordinates = []
+
+            # Get variable object
+            imported_modules = [KratosGeo, KratosMultiphysics]
```
mcgicjn2, line 71: Perhaps you need to check if the variable is in any imported module. I can imagine the same functionality would work if we want to change the properties along say a beam element from the structural mechanics application.

applications/GeoMechanicsApplication/python_scripts/set_parameter_field_process.py
```diff
@@ -0,0 +1,111 @@
...
+            # raise an error if the variable is not imported
+            if variable is None:
+                raise AttributeError(f'The variable: {self.params["variable_name"].GetString()} is not present within the '
+                                     f'imported modules')
+
+            for element in self.model_part.Elements:
+
+                # calculate center coordinates of the element
+                center = element.GetGeometry().Center()
+                coords = [center[0], center[1], center[2]]
```
mcgicjn2, line 88: What is the difference between center and coords, they both appear to be lists of the x,y,z ???

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
mcgicjn2: I would argue that to C++, this is SetParameterFieldUsingJsonString (So confused right now by style guide so it could be set_parameter_field_using_json_string)

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
mcgicjn2: thus if above is SetParameterFieldUsingJsonString, this is ParseParameterFieldFromJsonFile

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,155 @@
...
+{
+    // Read json string in field parameters file, create Parameters
+    const std::string& field_file_name = mParameters["dataset"].GetString();
```
mcgicjn2: Can we perhaps make a seperate parameter to clearly have "dataset_json_file" and "dataset". This is just for clarity in the input data. It is always better to know what to expected after the keyword in a json string so we can check for it.

mcgicjn2 (commented):
Generally speaking great implementation, just a small rearrangement on the c++ side, as c++ shouldn't have knowledge that it was called from python. A json string can also come from C# for instance

applications/GeoMechanicsApplication/python_scripts/set_parameter_field_process.py
aronnoordam: center is a kratos:Point object, while coords, is a pyhton list. I'll take a look if it can be used in the same way

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -0,0 +1,158 @@
...
+}
+
+
+
+
+void SetParameterFieldProcess::ExecuteInitialize()
+{
+    KRATOS_TRY
+
+    if (!this->mrModelPart.GetProcessInfo()[IS_RESTARTED]){
```
avdg81: Since this function only has an effect if something is not restarted, you may consider to use an early exit to be explicit out this:
```
if (mrModelPart.GetProcessInfo()[IS_RESTARTED]) {
    return;
}

// Here goes the rest of your code (including the try-catch-construct)
```
This will reduce the indentation level of the function's body by one. Consequently, if we would measure the code's complexity it would score better since you'd have fewer nested `if` statements, which is good.

avdg81 (dismissed):
Looks fine to me. I had only one minor remark.

avdg81 (approved):

PR #10793, opened 2023-02-22
[Core] PlaceholderPattern and ModelPartPattern

matekelemen:
## Description

- add a utility class `PlaceholderPattern` for string and filesystem related tasks involving basic string patterns
- add a convenience class `ModelPartPattern` that uses predefined placeholders `<model_part_name>`, `<step>`, `<time>` and `<rank>` (+ an interface for fetching the related values from a `ModelPart`)

Requested from #10778 by @sunethwarna.

## Features

- check whether an input string matches the pattern
- extract placeholders' values from an input string (even if the pattern contains duplicate placeholders but the input string has different values for them)
- substitute placeholders for given values
- collect all directories/files matching the pattern

## Example

A simple example with `ModelPartPattern`:
```py
import KratosMultiphysics

pattern = KratosMultiphysics.ModelPartPattern("<model_part_name>_<step>/output_<step>_<rank>.log")
pattern.IsAMatch("a_1/output_1_0.log") # <== true
pattern.IsAMatch("a_1/output_1_zero.log") # <== false
pattern.Match("a_1/output_2_0.log") # <== {"<model_part_name>" : ["a"], "<rank>" : ["0"], "<step>" : ["1", "2"]}
pattern.Glob() # <== depends on your filesystem, but most likely []

model_part = KratosMultiphysics.Model().CreateModelPart("whatever")
model_part.ProcessInfo[KratosMultiphysics.STEP] = 5
pattern.Apply(model_part) # <== "whatever_5/output_5_0.log"
```

`PlaceholderPattern` is a generalization of `ModelPartPattern`. You need to define it via a placeholder-regex map that specifies what expression to associate with each placeholder. Otherwise it works identically to `ModelPartPattern`.

### DIFF
--- a/kratos/python/add_other_utilities_to_python.cpp
+++ b/kratos/python/add_other_utilities_to_python.cpp
@@ -186,6 +186,13 @@ void AddSubModelPartEntitiesBooleanOperationToPython(pybind11::module &m, std::s
         ;
 }
 
+/// @brief Collect globbed paths to an array of strings.
+std::vector<std::filesystem::path> Glob (const PlaceholderPattern& rInstance) {
+    std::vector<std::filesystem::path> output;
+    rInstance.Glob(std::back_inserter(output));
+    return output;
+}
+
 void AddOtherUtilitiesToPython(pybind11::module &m)
 {
 
@@ -803,7 +810,37 @@ void AddOtherUtilitiesToPython(pybind11::module &m)
              "snake_case to CamelCase conversion")
         ;
 
+    pybind11::class_<PlaceholderPattern, PlaceholderPattern::Pointer>(m, "PlaceholderPattern")
+        .def(pybind11::init<const std::string&,const PlaceholderPattern::PlaceholderMap&>())
+        .def("IsAMatch",
+             &PlaceholderPattern::IsAMatch,
+             "Check whether a string satisfies the pattern")
+        .def("Match",
+             &PlaceholderPattern::Match,
+             "Find all placeholders' values in the input string.")
+        .def("Apply",
+             &PlaceholderPattern::Apply,
+             "Substitute values from the input map into the stored pattern.")
+        .def("Glob",
+             &Glob,
+             "Collect all file/directory paths that match the pattern.")
+        .def("GetRegexString",
+             &PlaceholderPattern::GetRegexString,
+             "Get the string representation of the regex.")
+        .def("IsConst",
+             &PlaceholderPattern::IsConst,
+             "Return true if the input pattern contains no placeholders.")
+        ;
 
+    pybind11::class_<ModelPartPattern, ModelPartPattern::Pointer, PlaceholderPattern>(m, "ModelPartPattern")
+        .def(pybind11::init<const std::string&>())
+        .def("Apply",
+             static_cast<std::string(ModelPartPattern::*)(const ModelPartPattern::PlaceholderMap&)const>(&ModelPartPattern::Apply),
+             "Substitute values from the input map into the stored pattern.")
+        .def("Apply",
+             static_cast<std::string(ModelPartPattern::*)(const ModelPart&)const>(&ModelPartPattern::Apply),
+             "Substitute values from the model part into the stored pattern.")
+        ;
 }
 
 } // namespace Python.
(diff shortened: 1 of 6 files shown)


roigcarlo:
Some toughs:

- I like the idea of the `RegexUtility` class. I would separate that into an individual file and expose it to python.
- `PlaceholderPattern` / `ModelPartPattern`:
    - Seems overly complicated to my taste. While I understand that the idea is to be able to automatically generate regex that are able to match given pattern from files it would be much simpler for the user to manually construct the matching regex directly with the `RegexUtility`. 
    - the format of the "fields" is tied to the class they belong. For example. I will never be able to match a `time_<time>` of the form `time_0.01a` unless I derive another different class with different substitution rules.
    - The interface seems to cover functions that are pretty much already present in c++ through `std::regex` or python's `re`
        - `IsAMatch` / `Match` : std::regex_match` / `re.match` or `re.search`
        - `Apply` : `std::regex_replace` / `re.sub`, simple format functions or the future `OutputController` from Kratos

Example of what I mean, I could directly write this:
```Python
import re
import pathlib

def get_step_re():
    """ aux fnc, that would come from RegexUtility """
    return r"(0|(?:-?[1-9]+[0-9]*))"
    
def get_time_re():
    """ aux fnc, that would come from RegexUtility """
    return r"(-?(?:(?:(?:[1-9][0-9]*)(?:\.[0-9]*)?)|(?:0(?:\.[0-9]*)?))(?:[eE][\+-]?[0-9]+)?)"

test_root = pathlib.Path().cwd()

paths = {
    test_root / "mdpa_name_step_1_time_0.0.h5",
    test_root / "mdpa_name_step_1_time_1.0.h5",
    test_root / "mdpa_name_step_1_time_1.1.h5",
    test_root / "mdpa name_step_3_time_03.h5",
}

RE_VALID_PATH = "_step_" + get_step_re() + "_time_" + get_time_re() + ".h5"

valid_match = [(p,m) for p in paths if (m:=re.findall(RE_VALID_PATH, p.parts[-1])) and len(m)]

for x in valid_match:
    print(x)
```

Which would correctly detect the valid paths:

```
(PosixPath('/home/roigcarlo/Problems/Tests/mdpa_name_step_1_time_1.0.h5'), [('1', '1.0')])
(PosixPath('/home/roigcarlo/Problems/Tests/mdpa_name_step_1_time_0.0.h5'), [('1', '0.0')])
(PosixPath('/home/roigcarlo/Problems/Tests/mdpa_name_step_1_time_1.1.h5'), [('1', '1.1')])
```

matekelemen:
Yep, I agree with most of what you wrote; that said, these classes are supposed to:
1) provide a uniform interface for both python **and** C++, so we don't have resort to exposing everything to the python level when we need these features
2) provide a layer completely free from regex (most people I know just never touch anything related to regex) while also giving the option for other users to define their own regexes if they so desire

If we only plan to do these kinds of tasks in python, or we expect users not to shy away from regex, I agree that this PR is unnecessary.

matekelemen:
By the way, any idea what MSVC is angry about?

roigcarlo:
> By the way, any idea of what MSVC is angry about?

Actually I enter the PR because of this. So far the only suspicious thing I've found is maybe a problem here:

```c++
PathType pattern(this->GetRegexString().substr(1, this->GetRegexString().size()-2));
```

It is possible that MSVC is not adding the trailing `\0` to their implementation of `std::filesystem::path` which would cause the last parenthesis of the most external capturing group to be deleted.

> If we only plan to do these kinds of tasks in python and we expect users not to shy away from regex, I agree that this PR is unnecessary.

I would like to discuss this first with the @KratosMultiphysics/technical-committee , because as I said, we plan to add a `OutputControler` which will effectively split the write process from the logic, and most of the token substitution can be probably integrated there directly.

That being said, I have no problem of merging this in the meantime.

matekelemen:
> > By the way, any idea of what MSVC is angry about?
>
> Actually I enter the PR because of this. So far the only suspicious thing I've found is maybe a problem here:
>
> ```cpp
> PathType pattern(this->GetRegexString().substr(1, this->GetRegexString().size()-2));
> ```
>
> It is possible that MSVC is not adding the trailing \0 to their implementation of std::filesystem::path which would cause > > the last parenthesis of the most external capturing group to be deleted.

Aaaa good point! It's a pretty ugly solution so I should probably tidy it up.

> That being said, I have no problem of merging this in the meantime.

Take your time; I'm not in any urgent need of this right now.

PR #10794, opened 2023-02-23
[FluidBio] WSS calculation minor fixes

rubenzorrilla:
**📝 Description**
This fixes some minor errors in the calculation of the WSS. Besides I removed a check that is no longer necessary. Test values have been updated accordingly.


agiuliodori (approved):
👍

PR #10798, opened 2023-02-23
[Core] Write STL - ignore triangles with area equal to zero

ddiezrod:
**📝 Description**
For some reason, sometimes stls contain triangles with area = 0.0 ( lines ) when we generate them from Inspire. We read these stls and after some modifications we write them in new files. The problem is that when we compute the normals for these degenerated triangles, the geometry throws an error because area is zero. The fix that I am adding here is to skip triangles whose area is = 0.0.


kratos/input_output/stl_io.cpp
```diff
@@ -95,12 +89,7 @@ void StlIO::WriteEntityBlock(const TContainerType& rThisEntities)
...
 {
     for (auto & r_geometry : rThisGeometries) {
-        // restrict to triangles only for now
-        const bool is_triangle = (
-            r_geometry.GetGeometryType() == GeometryData::KratosGeometryType::Kratos_Triangle3D3 ||
-            r_geometry.GetGeometryType() == GeometryData::KratosGeometryType::Kratos_Triangle3D6);
-
-        if (is_triangle) {
+        if (IsValidGeometry(r_geometry)) {
             WriteFacet(r_geometry);
```
rubenzorrilla, line 98: Maybe we can throw a warning if this happens. At least in debug mode.

loumalouomega (dismissed):

matekelemen:
I don't think this class should filter what it's reading/writing; it would mix in a responsibility that isn't enforced by the format. As a user, I'd expect `StlIO` to read/write whatever gets fed into it (provided it's in the right format).

This is actually a good example of why it's better to expose an interface that uses iterators instead of containers. You could easily write your own iterator class that filters degenerate triangles and feed that into `StlIO`, getting the functionality you need without changing the class.

ddiezrod:
> I don't think this class should filter what it's reading/writing; it would mix in a responsibility that isn't enforced by the format. As a user, I'd expect `StlIO` to read/write whatever gets fed into it (provided it's in the right format).
> 
> This is actually a good example of why it's better to expose an interface that uses iterators instead of containers. You could easily write your own iterator class that filters degenerate triangles and feed that into `StlIO`, getting the functionality you need without changing the class.

I get your idea, but the thing is that if we do not skip this element it will just throw an error, you cannot compute the normal of a line. Without just this small change you fix the problem without overengineering the whole code.

rubenzorrilla:
> 

I refer to my previous comment. Something in between would be to throw a warning. Maybe we should always throw it after @matekelemen 's observation.

matekelemen:
> you cannot compute the normal of a line

it's not unique, but it does exist and is computable. I just don't want a can of worms planted here that will get opened when someone eventually relies on having an identical number of elements/nodes as the source.

ddiezrod:
> > you cannot compute the normal of a line
> 
> it's not unique, but it does exist and is computable. I just don't want a can of worms planted here that will get opened when someone eventually relies on having an identical number of elements/nodes as the source.

We are now throwing an error in the geometry for these degenerated geometries https://github.com/KratosMultiphysics/Kratos/blob/672011bad421dbdb1757da4ad89b9fb87e4950c7/kratos/geometries/geometry.h#L1658 so, at least now in Kratos standards it id not computable. Maybe this can be discussed, although the current solution seems logic to me.

kratos/input_output/stl_io.cpp
ddiezrod: Warning added!

kratos/input_output/stl_io.cpp
```diff
@@ -138,6 +127,19 @@ void StlIO::PrintData(std::ostream& rOStream) const{
...
 
+
+bool StlIO::IsValidGeometry(const Geometry<Node<3>>& rGeometry) const {
+    // restrict to triangles only for now
+    const bool is_triangle = (
+        rGeometry.GetGeometryType() == GeometryData::KratosGeometryType::Kratos_Triangle3D3 ||
+        rGeometry.GetGeometryType() == GeometryData::KratosGeometryType::Kratos_Triangle3D6);
+    const bool area_greater_than_zero = rGeometry.Area() > std::numeric_limits<double>::epsilon();
+    if (!area_greater_than_zero) {
+        KRATOS_WARNING_ONCE("STLIO") << "Ignoring geometries with area = 0.0\n";
```
philbucher: how about counting how many were ignored and log that?

philbucher (dismissed):

kratos/input_output/stl_io.cpp
```diff
@@ -78,32 +78,29 @@ void StlIO::WriteModelPart(const ModelPart & rThisModelPart)
...
-        const bool is_triangle = (
-            r_geometry.GetGeometryType() == GeometryData::KratosGeometryType::Kratos_Triangle3D3 ||
-            r_geometry.GetGeometryType() == GeometryData::KratosGeometryType::Kratos_Triangle3D6);
-
-        if (is_triangle) {
+        if (IsValidGeometry(r_geometry, num_degenerate_geometries)) {
             WriteFacet(r_geometry);
         }
     }
+    KRATOS_WARNING_IF("STL-IO", num_degenerate_geometries > 0)
```
philbucher, line 88: I think using std::endl for warnings is a fair usage of flush

philbucher (dismissed):

kratos/input_output/stl_io.cpp
ddiezrod: Ok, done!

philbucher (approved):

ddiezrod:
Thanks @philbucher !

PR #10802, opened 2023-02-24
[Meshing] Fix error in father nodes in fulldebug

ddiezrod:
This was throwing an error in FullDebug, because if FATHER_NODES were not assigned before calling this method, you are doing a GetValue in a parallel section, allocating new memory.


loumalouomega (approved):
Question, do we have this as variable utils?

PR #10806, opened 2023-02-27
Release 9.3

roigcarlo:
**📝 Description**
Moving changes from Release to Master.

**🆕 Changelog**
- Updating versions
- Added python 3.11 as target
- Removed python 3.6 and 3.7 as targets

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,8 +27,8 @@ endif(POLICY CMP0057)
 
 # Set here the version number **** only update upon tagging a release!
 set (KratosMultiphysics_MAJOR_VERSION 9)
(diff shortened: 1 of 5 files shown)


rubenzorrilla (approved):

sunethwarna:
Can we wait until we check whether the python stubs can be integrated to the wheels?

sunethwarna:
@roigcarlo allryt... :+1: im totally fine with this then...

roigcarlo:
@sunethwarna Yep I am sorry but the limitation is from PyPI, if it were from me I would add it in 9.3, but there is no way to change when the number has been assigned :(

PR #10808, opened 2023-02-27
Update install instructions

roigcarlo:
**📝 Description**
Moved MinGW to a separate category.
Changes GNU/Linux to Linux/WSL, which is more useful for the users.

@loumalouomega I know that is a subsection of windows, but we are having lots of users which are mixing MSVC with MINGW, so I prefer to split it clearly

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -6,11 +6,11 @@
     - [Specific Application Dependencies](#specific-application-dependencies)
   - [Basic Configuration](#basic-configuration)
   - [Configuration scripts examples](#configuration-scripts-examples)
-    - [GNU/Linux](#gnulinux)
+    - [Linux/WSL](#gnulinux)
     - [Windows](#windows)
       - [Visual Studio](#visual-studio)
         - [Windows Visual Studio compilation configuration](#windows-visual-studio-compilation-configuration)
-      - [MinGW](#mingw)
+    - [MinGW](#mingw)
     - [MacOS](#macos)
   - [Adding Applications](#adding-applications)
   - [Post Compilation](#post-compilation)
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -6,11 +6,11 @@
     - [Specific Application Dependencies](#specific-application-dependencies)
   - [Basic Configuration](#basic-configuration)
   - [Configuration scripts examples](#configuration-scripts-examples)
-    - [GNU/Linux](#gnulinux)
+    - [Linux/WSL](#gnulinux)
```
loumalouomega, line 9: But its GNU/Linux, so I need to mention Stallman?

loumalouomega (approved):
As long as it doesn't remove it...

roigcarlo:
To be honest I've never seen the real use of mingw builds, even less now that we have WSL... but I have no problem leaving it as long as it is clearly separated from the Win section

loumalouomega:
> To be honest I've never seen the real use of mingw builds, even less now that we have WSL... but I have no problem leaving it as long as it is clearly separated from the Win section

It is far more performant than WSL..., and from scratch is easier to configure than VS, take far less space and doesn't require admin rights.

roigcarlo:
Well ease to use is arguable, and I don't have performance metrics of WSL/WSL2/CYGWIN/MINGW to compare (It is a an interesting metric, but so far I have seen no results comparing this).

The only points I could se with mingw is:
- To generate portable native builds of Kratos under windows. Anyone interested in those probably uses linux.
- To generate binaries that depend on windows components (WIN32's GUI mostly). Not for kratos.

PR #10809, opened 2023-02-28
[ConstitutiveLawsApp] Adding perturbation to coordinates of nodes

AlejandroCornejo:
**📝 Description**
In this PR I am adding a new process for the CLApp in which we perturbate the initial coordinates of a submodelpart by a normal distribution of probability.

The coordinate perturbation is done according to a numpy.normal_distribution in the direction of the nodal normal. 

**Required:** the submodelpart passed to the process must include a set of conditions to be able to compute the nodal normal field.

Finally, it prints the used coordinates of the perturbed nodes since this information will be lost after running.

Input to the json:

~~~json
{
            "python_module" : "add_texture_normal_distribution_process",
            "kratos_module" : "KratosMultiphysics.ConstitutiveLawsApplication",
            "process_name"  : "AddTextureNormalDistributionProcess",
            "Parameters"    : {
                "model_part_name" : "Structure.SurfacePressure3D_Pressure_on_surfaces_Auto1",
                "normal_mu"                  : 0.0,
                "normal_sigma"               : 0.05,
                "print_modified_coordinates" : true,
                "echo_level"                 : 1
            }
    }
~~~

Example of the result:

Without perturbation:
imagen

With perturbation:

imagen

### DIFF
--- a/applications/ConstitutiveLawsApplication/python_scripts/add_texture_normal_distribution_process.py
+++ b/applications/ConstitutiveLawsApplication/python_scripts/add_texture_normal_distribution_process.py
@@ -49,7 +49,7 @@ class AddTextureNormalDistributionProcess(KM.Process):
         self.echo = settings["echo_level"].GetInt()
 
 
-    def ExecuteInitializeSolutionStep(self):
+    def ExecuteInitialize(self):
         """This method is executed in order to initialize the current step
 
         Keyword arguments:
@@ -58,5 +58,24 @@ class AddTextureNormalDistributionProcess(KM.Process):
         # Here we compute the normal field
         normal_calculation_utils = KM.NormalCalculationUtils()
         normal_calculation_utils.CalculateUnitNormalsNonHistorical(self.model_part, 3)
+        number_nodes = len(self.model_part.Nodes)
+        normal = KM.Array3([0,0,0])
+
+        # We generate a random normal distribution
+        norm_distr = np.random.normal(self.mu, self.sigma, number_nodes)
+        counter = 0
+
+        for node in self.model_part.Nodes:
+            normal = node.GetValue(KM.NORMAL)
+            node.X0 += norm_distr[counter] * normal[0]
+            node.Y0 += norm_distr[counter] * normal[1]
+            node.Z0 += norm_distr[counter] * normal[2]
+            node.X  += norm_distr[counter] * normal[0]
+            node.Y  += norm_distr[counter] * normal[1]
+            node.Z  += norm_distr[counter] * normal[2]
+            counter += 1
+        
+        # Now we print a backup of the nodes coordinates used in the simulation
+        
 
 
--- a/applications/ConstitutiveLawsApplication/python_scripts/add_texture_normal_distribution_process.py
+++ b/applications/ConstitutiveLawsApplication/python_scripts/add_texture_normal_distribution_process.py
@@ -93,6 +93,8 @@ class AddTextureNormalDistributionProcess(KM.Process):
             counter += 1
             if self.print_coords:
                 # Now we print a backup of the nodes coordinates used in the simulation
+                if counter == 1:
+                    self.ascii_writer.write("The applied distribution has a mean of " + "{0:.4e}".format(np.mean(norm_distr)).rjust(11) + " and a standard deviation of " + "{0:.4e}".format(np.std(norm_distr)).rjust(11) + "\n\n")
                 self.ascii_writer.write(str(node.Id)+ "\t\t")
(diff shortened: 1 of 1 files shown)


AlejandroCornejo:
Example of application:
[test_case_1.gid.zip](https://github.com/KratosMultiphysics/Kratos/files/10847485/test_case_1.gid.zip)

rubenzorrilla:
I think this should be an operation rather than a process.

AlejandroCornejo:
> I think this should be an operation rather than a process.

What do you mean?

rubenzorrilla (commented):
Besides, the `Operation` comment I think:
- you should throw an error if there are no conditions in the provided model part to warn the user.
- not sure if you always requires the nodal coordinates output, if not I'd make it optional
- consider implementing this in c++ (I think you can use `std::normal_distribution`)
- testing is missing

rubenzorrilla:
> > I think this should be an operation rather than a process.
> 
> What do you mean?

That it should be derived from `operation.h`. Note that you're only calling the `ExecuteInitialize`, that is to say, you only need to execute this once.

AlejandroCornejo:
> 

Is there any example of this `Operation `in python? I've been looking for it and I cannot find any :S

rubenzorrilla:
> > 
> 
> Is there any example of this `Operation `in python? I've been looking for it and I cannot find any :S
> 
> How to include it in the json? It exists something like "liss_of_operations" ?

You can see `add_operations_to_python.cpp`. About how to use them you can see `multistage_analysis.py`.  Note that this would require reading the model part using the `import_mdpa_modeler.py`, which is indeed what it should be and what not what we're doing right now.

AlejandroCornejo:
I've added the error if no conditions are provided. The test will be added before merging. The printing of the coordinates was already optional via the `"print_modified_coordinates"`.

Checking what you mentioned, I understand how to use it (similar to processes) but as far as I know this is not yet included in the classical `analysis_stage` right? 

My idea was only to add a block in the `ProjectParameters.json` to do the thing.

manuelmessmer (commented):
Just as a comment. I am not 100% sure if this is what you do, but there is perturb_geometry_process in Kratos, which uses random fields. [See](https://github.com/KratosMultiphysics/Kratos/blob/master/applications/StructuralMechanicsApplication/python_scripts/perturb_geometry_subgrid_utility.py). 
Do you just want to perturb the nodes?

AlejandroCornejo:
> Just as a comment. I am not 100% sure if this is what you do, but there is perturb_geometry_process in Kratos, which uses random fields. [See](https://github.com/KratosMultiphysics/Kratos/blob/master/applications/StructuralMechanicsApplication/python_scripts/perturb_geometry_subgrid_utility.py). Do you just want to perturb the nodes?

Hi! I know what you mean but this is by far simpler... Just modify the coordinates of **some nodes** (not all of them) according to a certain probabilistic distribution :)

loumalouomega:
I mostly agree with @rubenzorrilla. But also I consider that if not submodelpart of conditions is provided, you can use SkinDetectionProcess

applications/ConstitutiveLawsApplication/python_scripts/add_texture_normal_distribution_process.py
```diff
@@ -0,0 +1,104 @@
...
+        normal = KM.Array3([0,0,0])
+
+        # We generate a random normal distribution
+        norm_distr = np.random.normal(self.mu, self.sigma, number_nodes)
+        counter = 0
+
+        # We modify the initial coordinates of the nodes to generate the new texture
+        for node in self.model_part.Nodes:
+            normal = node.GetValue(KM.NORMAL)
+            node.X0 += norm_distr[counter] * normal[0]
```
loumalouomega, line 87: Do you want to move also the initial coordinate?

AlejandroCornejo:
> I mostly agree with @rubenzorrilla. But also I consider that if not submodelpart of conditions is provided, you can use SkinDetectionProcess

I thought exactly the same. My feeling is that it gives more generality but you lose control over the domain to be accounted.

applications/ConstitutiveLawsApplication/python_scripts/add_texture_normal_distribution_process.py
AlejandroCornejo: Maybe is not necessary but I wanted to be sure of not affecting the displacements :)

AlejandroCornejo:
In the end, since this process is really particular for our problem, I will close the PR and maintain the developments in a branch.

@lagoncalvesjr @SergioJimenezReyes

PR #6747, opened 2020-04-08
[Core][Communnicator] adding GlobalNumberOf... functions

philbucher:
I find this quite helpful

I think the `Communicator` is a good place for this

Tests are added

PR #10811, opened 2023-02-28
[Core][MPI] Adding GlobalNumberOfMasterSlaveConstraints to communicator

loumalouomega:
**📝 Description**

Adding `GlobalNumberOfMasterSlaveConstraints` to communicator. This was missing and it was inconsistent. In addition, it was not possible to consider MPC in MPI without this. Updating the [`ParallelFillCommunicator`](https://github.com/KratosMultiphysics/Kratos/commit/f267277900fe2882d80df4ef4c31d76e8cea77b7) correspondingly.

**🆕 Changelog**

- [Adding GlobalNumberOfMasterSlaveConstraints to communicator](https://github.com/KratosMultiphysics/Kratos/commit/21052f40bb00affab951b765fdd07f2c2f4f6087)
- [Update `ParallelFillCommunicator`](https://github.com/KratosMultiphysics/Kratos/commit/f267277900fe2882d80df4ef4c31d76e8cea77b7)
- [Adding corresponding test](https://github.com/KratosMultiphysics/Kratos/commit/90a0f9c367b83d381c5b0521055afb4c60bce79d)


philbucher:
There is a reason I didn't add those: https://github.com/KratosMultiphysics/Kratos/pull/6747#discussion_r405885108

Can you please comment on this issue?

loumalouomega:
> There is a reason I didn't add those: [#6747 (comment)](https://github.com/KratosMultiphysics/Kratos/pull/6747#discussion_r405885108)
> 
> Can you please comment on this issue?

From Aditya:

> "I am not sure if this will work. Constraints are not treated in the ParallelFillComm or in partitioning ..
> so if constraints are added through ModelPart they are not in mpLocalMesh."

In this PR I am adding it to the ParallelFillComm.

loumalouomega:
Anyone, @philbucher ?

rubenzorrilla (commented):
Changes make sense for me so I'm approving after @philbucher thumbs up.

My unique concern is that we need to be aware of the case in which we create/destroy constraints (the communicator constraint data would be no longer valid). Though this is be the same for nodes, elements, etc. constraints are more prone to be used like this.  I'm just pointing the issue to keep it in mind.

loumalouomega:
> Changes make sense for me so I'm approving after @philbucher thumbs up.
> 
> My unique concern is that we need to be aware of the case in which we create/destroy constraints (the communicator constraint data would be no longer valid). Though this is be the same for nodes, elements, etc. constraints are more prone to be used like this. I'm just pointing the issue to keep it in mind.

I agree with your point, and I am aware of the issue. In fact, in my current task I am planning to touch that issue, as we need to do additional work in the `ParallelFillCommunicator` as the constraints usually relate DoFs non-existing in the current partition and a solution for this problem must be done (if you try to relate DoFs non-existing in the same partition you will get an error).

rubenzorrilla:
> > Changes make sense for me so I'm approving after @philbucher thumbs up.
> > My unique concern is that we need to be aware of the case in which we create/destroy constraints (the communicator constraint data would be no longer valid). Though this is be the same for nodes, elements, etc. constraints are more prone to be used like this. I'm just pointing the issue to keep it in mind.
> 
> I agree with your point, and I am aware of the issue. In fact, in my current task I am planning to touch that issue, as we need to do additional work in the `ParallelFillCommunicator` as the constraints usually relate DoFs non-existing in the current partition and a solution for this problem must be done (if you try to relate DoFs non-existing in the same partition you will get an error).

Yes... this will be also a weird one...

rubenzorrilla (approved):

PR #10815, opened 2023-03-01
[CLApp] Adding option to use custom shear odulus in Generic Anisot law

AlejandroCornejo:
**📝 Description**
Just adding the option to use custom Shear modulus for the Generic anisotropic law.


SergioJimenezReyes (approved):
Let's wait for the tests, but this is a small change not affecting the original implementation. Looks fine from my point of view.

PR #10823, opened 2023-03-03
[Core][Parameters] Trying to remove the python iterator for `Parameters`

RiccardoRossi:
**📝 Description**

I want to see how many errors we get, in order to try addressing #10649

**🆕 Changelog**

- [Trying to remove the python iterator for parameters](https://github.com/KratosMultiphysics/Kratos/pull/10823/commits/4780f91461d9512b5dc9ae223d192e2055eb8969)


loumalouomega:
CI is not passing

sunethwarna:
We may need to deprecate the existing and promote using `items`, `values` and `keys`...

philbucher:
While I approve of removing this method, I strongly agree with @sunethwarna , this must be deprecated.

Imagine all the code that is outside the repo (custom processes etc!) that would be broken if you remove it directly!

RiccardoRossi:
the problem is that you cannot deprecate this since it is in the python interface...

philbucher:
You can add a deprecation warning in the py interface

We have done this dozens of times

philbucher (changes requested):
> the problem is that you cannot deprecate this since it is in the python interface...

[it is](https://github.com/KratosMultiphysics/Kratos/pull/10835) ;)

RiccardoRossi:
adding a comment for myself: the following function allows returning the correct type in python from parameters without having to call the GetXXX method.

~~~cpp
.def("find", [](Parameters& rParam, std::string name) -> py::object {
        auto my_obj = rParam[name];
        if(my_obj.IsDouble())
            return py::cast(my_obj.GetDouble());
        else if(my_obj.IsInt())
            return py::cast(my_obj.GetInt());
        else if(my_obj.IsBool())
            return py::cast(my_obj.GetBool());
        else if(my_obj.IsString())
            return py::cast(my_obj.GetString());
        else if(my_obj.IsVector())
            return py::cast(my_obj.GetVector());
        else if(my_obj.IsVector())
            return py::cast(my_obj.GetVector());
        else if(my_obj.IsMatrix())
            return py::cast(my_obj.GetMatrix());
        else if(my_obj.IsNull())
            return py::none();
        else
            return py::cast(my_obj);
                })
~~~

if we replaced the current implementation of __getitem__ and __setitem__ by this, the Parameters would work as a normal dictionary. too bad that we would break everything...

matekelemen:
> adding a comment for myself: the following function allows returning the correct type in python from parameters without having to call the GetXXX method.
> 
> ```c++
> .def("find", [](Parameters& rParam, std::string name) -> py::object {
>         auto my_obj = rParam[name];
>         if(my_obj.IsDouble())
>             return py::cast(my_obj.GetDouble());
>         else if(my_obj.IsInt())
>             return py::cast(my_obj.GetInt());
>         else if(my_obj.IsBool())
>             return py::cast(my_obj.GetBool());
>         else if(my_obj.IsString())
>             return py::cast(my_obj.GetString());
>         else if(my_obj.IsVector())
>             return py::cast(my_obj.GetVector());
>         else if(my_obj.IsVector())
>             return py::cast(my_obj.GetVector());
>         else if(my_obj.IsMatrix())
>             return py::cast(my_obj.GetMatrix());
>         else if(my_obj.IsNull())
>             return py::none();
>         else
>             return py::cast(my_obj);
>                 })
> ```
> 
> if we replaced the current implementation of **getitem** and **setitem** by this, the Parameters would work as a normal dictionary. too bad that we would break everything...

we have [ParametersWrapper](https://github.com/KratosMultiphysics/Kratos/blob/6dcf133f24441f8cf356a42a31a9f983f9781228/applications/HDF5Application/python_scripts/core/utils.py#L16) in the HDF5Application that does exactly this. It could be nice but in my experience, unless **all** `Parameters` are replaced by this in python, it's a colossal pain to support both versions.

philbucher:
Sure, but can you please remove the accidentally committed files in the DEMApp?

kratos/tests/test_kratos_parameters.py
```diff
@@ -490,10 +490,10 @@ def test_iterators(self):
         kp = Parameters(json_string)
 
         #iteration by range
-        nitems = 0
-        for iterator in kp:
-            nitems = nitems + 1
-        self.assertEqual(nitems, 5)
+        #nitems = 0
```
philbucher: And please remove this

RiccardoRossi (commented):
@philbucher done!

applications/SwimmingDEMApplication/python_scripts/parameters_tools.py
```diff
@@ -54,7 +54,7 @@ def Condition(p, k):
 
     elif CanBeIterated(parameters): # there could be lists
 
-        for i, _ in enumerate(parameters):
+        for i, _ in enumerate(parameters.values()):
```
philbucher, line 57: That one could be redone better I believe 🤔
FYI @sunethwarna

philbucher (approved):
Great!

RiccardoRossi:
@matekelemen this, together with #10835 should solve for good the problem with parameters iterators in python

PR #10825, opened 2023-03-03
[TrilinosApplication] More consistent and rigorous checks

loumalouomega:
**📝 Description**

I noticed that checks in sparse matrices could give false positive (my own implementation, my fault). So I added a more rigorous check, checking that all values provided are checked.

**🆕 Changelog**

- [Using epetra communicator as suggested by](https://github.com/KratosMultiphysics/Kratos/pull/10825/commits/d8dd0e9ac6d7eba4406753e37548e68f930c91ba) @philbucher


applications/TrilinosApplication/custom_utilities/trilinos_cpp_test_utilities.cpp
```diff
@@ -270,13 +276,18 @@ void TrilinosCPPTestUtilities::CheckSparseMatrix(
...
                         break;
                     }
                 }
                 break;
             }
         }
     }
+
+    // Checking that all the values has been validated
+    const std::size_t global_validated_values = rDataCommunicator.SumAll(local_validated_values);
```
philbucher: you can do that also with the epetra comm (basically the same concept as our data comm)

loumalouomega: Ours is easier to use :)

philbucher: this way you could avoid passing of the additional argument
same as you did with other functions

philbucher: it also has the [SumAll](https://docs.trilinos.org/dev/packages/epetra/doc/html/classEpetra__Comm.html#af81a51a85a8756371c6005de62378dbb) function

No strong opinion as this is only for testing

philbucher (dismissed):
I didn't notice this before, can you please move `trilinos_cpp_test_utilities.cpp` to `tests/cpp_tests`?

applications/TrilinosApplication/custom_utilities/trilinos_cpp_test_utilities.cpp
```diff
@@ -233,25 +234,30 @@ void TrilinosCPPTestUtilities::CheckSparseMatrixFromLocalMatrix(
     values.reserve(total_size);
     for (std::size_t i = 0; i < rB.size1(); ++i) {
         for (std::size_t j = 0; j < rB.size2(); ++j) {
-            row_indexes.push_back(i);
```
RiccardoRossi: why are u doing this?? the zeros should be there ,...

applications/TrilinosApplication/custom_utilities/trilinos_cpp_test_utilities.cpp
```diff
@@ -233,25 +234,30 @@ void TrilinosCPPTestUtilities::CheckSparseMatrixFromLocalMatrix(
...
 void TrilinosCPPTestUtilities::CheckSparseMatrix(
     const TrilinosSparseMatrixType& rA,
+    const DataCommunicator& rDataCommunicator,
     const std::vector<int>& rRowIndexes,
     const std::vector<int>& rColumnIndexes,
     const std::vector<double>& rValues,
     const double Tolerance
     )
 {
+    int local_validated_values = 0;
```
RiccardoRossi: why don't you use the epetra communicator that comes with rA?

applications/TrilinosApplication/custom_utilities/trilinos_cpp_test_utilities.cpp
RiccardoRossi: i agree with @philbucher on this

applications/TrilinosApplication/custom_utilities/trilinos_cpp_test_utilities.h
```diff
@@ -140,24 +140,28 @@ class TrilinosCPPTestUtilities
     * @brief This method checks the values of a sparse matrix with the given serial matrix
     * @param rA The matrix to check
     * @param rB The reference matrix
+    * @param rDataCommunicator The data communicator considered
     * @param Tolerance The tolerance considered
     */
     static void CheckSparseMatrixFromLocalMatrix(
         const TrilinosSparseMatrixType& rA,
         const TrilinosLocalMatrixType& rB,
+        const DataCommunicator& rDataCommunicator,
```
RiccardoRossi: same

RiccardoRossi (dismissed):

applications/TrilinosApplication/custom_utilities/trilinos_cpp_test_utilities.cpp
loumalouomega: This is a function to check a sparse matrix from a local matrix, not all values are defined, otherwise will fail the check

loumalouomega:
Change done @RiccardoRossi

philbucher:
Can you please also move the file?

loumalouomega:
Where?

loumalouomega:
The cpp utilities in the core are in the utilities folder.

philbucher:
But they shouldn't be ;)

Those things should not be used in production

loumalouomega:
@RiccardoRossi this is still blocked

loumalouomega:
Please @RiccardoRossi check this

applications/TrilinosApplication/tests/cpp_tests/trilinos_cpp_test_utilities.cpp
```diff
@@ -233,9 +233,12 @@ void TrilinosCPPTestUtilities::CheckSparseMatrixFromLocalMatrix(
     values.reserve(total_size);
     for (std::size_t i = 0; i < rB.size1(); ++i) {
         for (std::size_t j = 0; j < rB.size2(); ++j) {
-            row_indexes.push_back(i);
-            column_indexes.push_back(j);
-            values.push_back(rB(i, j));
+            const double value = rB(i, j);
```
RiccardoRossi, line 237: why not qdding if It is. a zero?

loumalouomega: Usually what defines a sparse matrix is the number of non-zero terms, that's why

loumalouomega: Any case, see my previous answer too:

> This is a function to check a sparse matrix from a local matrix, not all values are defined, otherwise will fail the check

RiccardoRossi: I am not convinced by this answer... a matrix should have all the values that may be potentially non zero...so i don't think it is a good idea to make the check you are doing here. Furthermore that check is not really robust...

loumalouomega: I know that a sparse matrix can have non zero terms which value is actually is zero. The implementation is to check the actual non zero values, otherwise the graph must be provided. I added an additional method where you provide the indices but the values are check from a local matrix, this will work for you?

loumalouomega: @RiccardoRossi do you agree?

loumalouomega:
Ping @RiccardoRossi

applications/TrilinosApplication/tests/cpp_tests/trilinos_cpp_test_utilities.cpp
loumalouomega: Hello @RiccardoRossi

loumalouomega:
Please @RiccardoRossi check this, you are blocking

loumalouomega:
@RiccardoRossi reping

applications/TrilinosApplication/tests/cpp_tests/trilinos_cpp_test_utilities.cpp
RiccardoRossi: mmm, the local matrix type is dense? 

at least give a more descriptive name instead of tolerance. something like "neglibible_value_threshold" or smthg of the sort

applications/TrilinosApplication/tests/cpp_tests/trilinos_cpp_test_utilities.cpp
```diff
@@ -234,14 +234,40 @@ void TrilinosCPPTestUtilities::CheckSparseMatrixFromLocalMatrix(
...
 }
 
+/***********************************************************************************/
+/***********************************************************************************/
+
+void TrilinosCPPTestUtilities::CheckSparseMatrixFromLocalMatrix(
+    const TrilinosSparseMatrixType& rA,
+    const std::vector<int>& rRowIndexes,
+    const std::vector<int>& rColumnIndexes,
+    const TrilinosLocalMatrixType& rB,
```
RiccardoRossi, line 255: btw, in my tests i was normally using a map to store B.
why do you need it to be a dense matrix?

loumalouomega: Because I use B to check results. The multiplication between denses matrices is well-defined, and  we can be sure will provide a reliable result. This is purely for testing, not used in production code.

RiccardoRossi (approved):

PR #10830, opened 2023-03-06
Fix warnings about initialization of member variables.

### LINKED ISSUE #10857: [GeoMechanicsApplication] Fix warnings about initialization of member variables

huhaas:
**📝 Description**
Fix warnings and simplify retention law parameter logic.

**🆕 Changelog**
- Fix warnings and simplify retention law parameter logic.

### DIFF
--- a/applications/GeoMechanicsApplication/custom_retention/retention_law.h
+++ b/applications/GeoMechanicsApplication/custom_retention/retention_law.h
@@ -68,80 +68,59 @@ public:
         * @param mpConstitutiveMatrix pointer to the material tangent matrix (*OUTPUT with COMPUTE_CONSTITUTIVE_TENSOR flag)
 
         * GEOMETRIC PARAMETERS:
-        * @param mpElementGeometry pointer to the element's geometry (input data)
+        * @param mrElementGeometry reference to the element's geometry (input data)
 
         * MATERIAL PROPERTIES:
-        * @param mpMaterialProperties pointer to the material's Properties object (input data)
+        * @param mrMaterialProperties reference to the material's Properties object (input data)
 
         * PROCESS PROPERTIES:
-        * @param mpCurrentProcessInfo pointer to current ProcessInfo instance (input data)
+        * @param mrCurrentProcessInfo reference to current ProcessInfo instance (input data)
 
         */
 
     private:
-        /*** NOTE: Member Pointers are used only to point to a certain variable, 
-         * no "new" or "malloc" can be used for this Parameters ***/
 
-        double mFluidPressure;
-        double mMeanStress;
-        double mTemperature;
-        double mVolumetricStrain;
+        double mFluidPressure = 0.0;
+        double mMeanStress = 0.0;
+        double mTemperature = 0.0;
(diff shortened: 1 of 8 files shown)


applications/GeoMechanicsApplication/custom_retention/retention_law.h
```diff
@@ -68,80 +68,61 @@ class KRATOS_API(GEO_MECHANICS_APPLICATION) RetentionLaw
...
 
         * PROCESS PROPERTIES:
-        * @param mpCurrentProcessInfo pointer to current ProcessInfo instance (input data)
+        * @param mrCurrentProcessInfo reference to current ProcessInfo instance (input data)
 
         */
 
     private:
         /*** NOTE: Member Pointers are used only to point to a certain variable, 
          * no "new" or "malloc" can be used for this Parameters ***/
```
avdg81: I believe this comment no longer applies here, so please remove it. Thanks.

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
```diff
@@ -519,15 +520,15 @@ class NewmarkQuasistaticUPwScheme : public Scheme<TSparseSpace,TDenseSpace>
...
-    double mDeltaTime;
+    double mDeltaTime = 0.0;
 
 //----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
-    NewmarkQuasistaticUPwScheme() : Scheme<TSparseSpace,TDenseSpace>()
+    NewmarkQuasistaticUPwScheme()
+        : Scheme<TSparseSpace,TDenseSpace>()
+        , mBeta(0.25)
+        , mGamma(0.5)
+        , mTheta(0.5)
```
avdg81: What about moving these default values to the declaration of the class (just like you did for `mDeltaTime`)? In that way we ensure that the data members are always initialized properly (including any future constructors). As a bonus, it also puts all default values in a single location, which makes looking them up easy. If you do that, you probably also don't need to explicitly define the default constructor (the compiler can generate it for you).

avdg81 (dismissed):
Nice cleanup Hugo. Just two minor remarks from my side. Please merge when you're done.

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
```diff
@@ -514,21 +512,10 @@ class NewmarkQuasistaticUPwScheme : public Scheme<TSparseSpace,TDenseSpace>
 
 protected:
 
-    /// Member Variables
-
     double mBeta;
     double mGamma;
     double mTheta;
```
avdg81: I would prefer to give these default values (based on what used to be set in the default constructor), just so we can't forget to initialize them (in particular when new constructors are added).

avdg81 (dismissed):

avdg81 (approved):

PR #10832, opened 2023-03-06
[Core] Adding KRATOS_WATCH_MPI

loumalouomega:
**📝 Description**

Adding KRATOS_WATCH_MPI. Basically the same as KRATOS_WATCH, but you also pass the DataCommunicator, so you can see from which partition it comes each message.

Example output:

~~~cpp
const auto& r_comm = rModelPart.GetCommunicator().GetDataCommunicator();
KRATOS_WATCH_MPI(number_of_local_dofs, r_comm)
~~~

~~~sh
RANK 1/2      number_of_local_dofs : 4
RANK 0/2      number_of_local_dofs : 2
~~~

**🆕 Changelog**

- [Adding KRATOS_WATCH_MPI](https://github.com/KratosMultiphysics/Kratos/commit/2917b7429b897ca83e663d0e47710de79e344e33)

### DIFF
--- a/kratos/includes/define.h
+++ b/kratos/includes/define.h
@@ -803,7 +803,7 @@ namespace Kratos
 //Print Trace if defined
 #define KRATOS_WATCH(variable) std::cout << #variable << " : " << variable << std::endl;
 #define KRATOS_WATCH_CERR(variable) std::cerr << #variable << " : " << variable << std::endl;
-#define KRATOS_WATCH_MPI(variable, mpi_data_comm) std::cout << "RANK " << mpi_data_comm.Rank() << "/" << mpi_data_comm.Size() - 1 << "\t"; KRATOS_WATCH(variable);
+#define KRATOS_WATCH_MPI(variable, mpi_data_comm) std::cout << "RANK " << mpi_data_comm.Rank() << "/" << mpi_data_comm.Size()  << "    "; KRATOS_WATCH(variable);
 
 }  /* namespace Kratos.*/
 
--- a/kratos/includes/define.h
+++ b/kratos/includes/define.h
@@ -803,6 +803,7 @@ namespace Kratos
 //Print Trace if defined
 #define KRATOS_WATCH(variable) std::cout << #variable << " : " << variable << std::endl;
 #define KRATOS_WATCH_CERR(variable) std::cerr << #variable << " : " << variable << std::endl;
+#define KRATOS_WATCH_MPI(variable, mpi_data_comm) std::cout << "RANK " << mpi_data_comm.Rank() + 1 << "/" << mpi_data_comm.Size() << "\t" << #variable << " : " << variable << std::endl;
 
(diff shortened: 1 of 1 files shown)


kratos/includes/define.h
```diff
@@ -803,6 +803,7 @@ namespace Kratos
 //Print Trace if defined
 #define KRATOS_WATCH(variable) std::cout << #variable << " : " << variable << std::endl;
 #define KRATOS_WATCH_CERR(variable) std::cerr << #variable << " : " << variable << std::endl;
+#define KRATOS_WATCH_MPI(variable, mpi_data_comm) std::cout << "RANK " << mpi_data_comm.Rank() + 1 << "/" << mpi_data_comm.Size() << "\t" << #variable << " : " << variable << std::endl;
```
philbucher: Please remove the +1, this is not the convention for MPI ranks

Also please call KRATOS_WATCH inside this macro to avoid code duplication

philbucher (changes requested):

kratos/includes/define.h
loumalouomega: I can remove the +1

loumalouomega: Done

kratos/includes/define.h
```diff
@@ -803,6 +803,7 @@ namespace Kratos
 //Print Trace if defined
 #define KRATOS_WATCH(variable) std::cout << #variable << " : " << variable << std::endl;
 #define KRATOS_WATCH_CERR(variable) std::cerr << #variable << " : " << variable << std::endl;
+#define KRATOS_WATCH_MPI(variable, mpi_data_comm) std::cout << "RANK " << mpi_data_comm.Rank() << "/" << mpi_data_comm.Size() - 1 << "\t"; KRATOS_WATCH(variable);
```
philbucher:
```suggestion
#define KRATOS_WATCH_MPI(variable, mpi_data_comm) std::cout << "RANK " << mpi_data_comm.Rank() << "/" << mpi_data_comm.Size()  << "    "; KRATOS_WATCH(variable);
```

please follow the standard conventions of MPI

otherwise if you run with one rank it will print 0 :)

also I would not use tab as it can be quite long on some terminals

philbucher (approved):
:+1:

matekelemen:
Btw you can spawn a terminal emulator (eg: `konsole`) for each MPI process and track the outputs separately to reduce clutter
```
konsole -e mpirun ... python ...
```

loumalouomega:
> Btw you can spawn a terminal emulator (eg: `konsole`) for each MPI process and track the outputs separately to reduce clutter
> 
> ```
> mpirun ... konsole -e python ...
> ```

Huum, didn't know that. Anyway I am debugging with ssh, so I cannot not do that

philbucher:
I also occasionally used to spawn multiple terminals, but it is super messy beyond 4 processes

Nice though for running gdb in each rank

PR #10834, opened 2023-03-06
[DEMApplication] Some minor modifications to the parallel bond model

ChengshunShang1996:
**📝 Description**
Some minor modifications to the parallel bond model in DEMApplication.

**🆕 Changelog**
- Using different BOND_ROTATIONAL_COEFFICIENT for normal and tengential direction.
- Added feature to calculate the normal bonded force after bond break.
- Added a parallel_bond_model based membrane model. 
- Added a variable GROUP_ID for visualization.
- Added a funtion to calculate the local porosity.


maceligueta:
Hi @ChengshunShang1996, some questions:
- BOND_ROTATIONAL_COEFFICIENT_NORMAL means torsional and BOND_ROTATIONAL_COEFFICIENT_TANGENTIAL means flexion?
- Concerning the membrane contact model, we used to use https://github.com/KratosMultiphysics/Kratos/blob/master/applications/DEMApplication/custom_constitutive/DEM_KDEM_fabric_CL.cpp . There's a 3D version and a 2D version. If your element is better, we should remove the old one. Please take a look and comment with @salvalatorre . Let's avoid duplicates for the same capability.

ChengshunShang1996:
> 

Hi Miguel Angel, thanks for your commnets.

- The BOND_ROTATIONAL_COEFFICIENT_NORMAL and the BOND_ROTATIONAL_COEFFICIENT_TANGENTIAL correspond to the bond moment contributions coefficient to normal and shear contact stresses. The BOND_ROTATIONAL_COEFFICIENT_NORMAL relates to flexion and BOND_ROTATIONAL_COEFFICIENT_TANGENTIAL relates to torsional. It can be found in Fig.2 of [Ding, 2014](https://doi.org/10.1016/j.ijrmms.2014.03.008).

- For the membrane contact model, I have discussed it with @salvalatorre. Actually, both versions of membrane models are simple derived models from their father models. Their function is to reduce the value of the particle rotational moment or set the rotational moment as 0. BUT, they are different in some aspects because they have different father models, although they may achieve similar simulations. Maybe we should keep both versions.

AlejandroCornejo:
@rlrangel I think this does not affect you but just in case

maceligueta (dismissed):

ChengshunShang1996:
Hi @maceligueta, I missed your last approval. Please help me review it again.

maceligueta (approved):

PR #10835, opened 2023-03-06
[Core] deprecate __iter__of parameters

philbucher:
This deprecates the `__iter__` method of  `Kratos::Parameters`

a deprecation warning is issued ONCE per call location, including file name and line number, e.g.
`Deprecated method called in test_kratos_parameters.py in line 495: Iterating through "Parameters" object is deprecated, please use the "values" method instead`

This way it is fully backward compatible

### DIFF
--- a/kratos/python_interface/__init__.py
+++ b/kratos/python_interface/__init__.py
@@ -141,3 +141,24 @@ def _ImportApplication(application, application_name):
 
 def IsDistributedRun():
     return KratosGlobals.Kernel.IsDistributedRun()
+
+
+# iterating through the parameters is deprecated
+# the following wraps the __iter__ method to issue a deprecation warning
+list_deprecation_warnings = []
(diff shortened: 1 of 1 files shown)


RiccardoRossi:
pretty cool!

philbucher:
> pretty cool!

:sunglasses:

matekelemen:
[](https://youtube.com/shorts/BESOAAiYhBs?feature=share)

While we're on topic: do we have some standard mechanism to remind us when some **T** time passed since deprecation, and we can remove the feature?

philbucher:
> [](https://youtube.com/shorts/BESOAAiYhBs?feature=share)
> 
> While we're on topic: do we have some standard mechanism to remind us when some **T** time passed since deprecation, and we can remove the feature?

not really, but there probably should be one

RiccardoRossi (approved):

PR #10840, opened 2023-03-07
[CoSim] Minor readme updates

matekelemen:
- fix image links
- replace relative filesystem links with github links
- replace `~~~` code blocks with backtick blocks
- replace markdown anchors with html anchors (required by the doxygen docs)

### DIFF
--- a/applications/CoSimulationApplication/README.md
+++ b/applications/CoSimulationApplication/README.md
@@ -2,35 +2,48 @@
 
 The CoSimulation Application contains the core developments in coupling black-box solvers and other software-tools within Kratos Multiphysics.
 
+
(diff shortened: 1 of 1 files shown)


sunethwarna (approved):

philbucher (commented):
Just for my understanding, why the full links?

matekelemen:
> Just for my understanding, why the full links?

The readme is now part of the docs, which is deployed on some static site that has no concept of the kratos directory layout. By providing the full path to the corresponding files on github, the links work on both a cloned kratos repo on your local machine and the deployed docs on the web (at the expense of opening a browser even if you're skimming through the readme on a local kratos clone).

PR #10841, opened 2023-03-07
Class name change: FluidDynamicsAnalysisRVE ==> FluidDynamicsAnalysisRve

Rbravo555:
**📝 Description**
While attempting a Reduced Order Model of the fluid RVE using the automatic processes implemented in the RomApplication, Kratos complaint that the RVE stage name is not in Pascal case. 

**🆕 Changelog**
Changing the name of the class in the two places where it appeared:
- the python script: fluid_dynamics_analysis_rve.py
- the test: fluid_rve_test.py

### DIFF
--- a/applications/FluidDynamicsApplication/python_scripts/fluid_dynamics_analysis_rve.py
+++ b/applications/FluidDynamicsApplication/python_scripts/fluid_dynamics_analysis_rve.py
@@ -5,7 +5,7 @@ from KratosMultiphysics.FluidDynamicsApplication.fluid_dynamics_analysis import 
 # Importing other libraries
 import math
 
-class FluidDynamicsAnalysisRVE(FluidDynamicsAnalysis):
+class FluidDynamicsAnalysisRve(FluidDynamicsAnalysis):
     def __init__(self, model, project_parameters):
         # Validate RVE settings
         input_rve_settings = project_parameters["rve_settings"]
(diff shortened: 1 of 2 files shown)


applications/FluidDynamicsApplication/python_scripts/fluid_dynamics_analysis_rve.py
```diff
@@ -5,7 +5,7 @@
 # Importing other libraries
 import math
```
rubenzorrilla, line 7:
```suggestion

class FluidDynamicsAnalysisRVE(FluidDynamicsAnalysisRve):
    def __init__(self, model, project_parameters):
        KratosMultiphysics.Logger.PrintWarning("'FluidDynamicsAnalysisRVE' is deprecated. Use 'FluidDynamicsAnalysisRve' instead.")
        super().__init__(model, project_parameters)
```
Something of this sort. Note that you'll need to place it at the bottom of the file in order to have ' FluidDynamicsAnalysisRve' defined.

rubenzorrilla (commented):
My bad. My unique concern is retrocompatibility. To address this I'd add a fake derived class with the old name throwing a warning at compile time (see my suggestion). Then we can remove this one after a while.

Besides, we must update the cases in the Examples repository accordingly.

rubenzorrilla (approved):

PR #10842, opened 2023-03-07
[Core] Fix potential bug in B&S with MPC

loumalouomega:
**📝 Description**

Previous implementation was wrong as indexes were not consistent. This is already discussed with @KratosMultiphysics/altair team. A priori, no test is broken.

**🆕 Changelog**

- [Fix potential bug B&S](https://github.com/KratosMultiphysics/Kratos/commit/6f3b701a77707b9dbce30bf8170ad990240e3f1e)
- [Minor code duplication reduction](https://github.com/KratosMultiphysics/Kratos/pull/10842/commits/afc98776c44008d83ec9e3fc2c71cafab48648dd)

### DIFF
--- a/kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver.h
+++ b/kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver.h
@@ -1409,6 +1409,7 @@ protected:
             #pragma omp for schedule(guided, 512)
             for (int i_const = 0; i_const < number_of_constraints; ++i_const) {
                 auto it_const = rModelPart.MasterSlaveConstraints().begin() + i_const;
+                it_const->EquationIdVector(slave_equation_ids, master_equation_ids, r_current_process_info);
 
                 // Detect if the constraint is active or not. If the user did not make any choice the constraint
                 // It is active by default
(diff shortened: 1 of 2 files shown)


kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver.h
```diff
@@ -1327,10 +1327,10 @@ class ResidualBasedBlockBuilderAndSolver
...
 
                 // Merging all the temporal indexes
-                for (int i = 0; i < static_cast<int>(temp_indices.size()); ++i) {
-                    lock_array[i].lock();
-                    indices[i].insert(temp_indices[i].begin(), temp_indices[i].end());
-                    lock_array[i].unlock();
+                for (auto& pair_temp_indices : temp_indices) {
+                    lock_array[pair_temp_indices.first].lock();
+                    indices[pair_temp_indices.first].insert(pair_temp_indices.second.begin(), pair_temp_indices.second.end());
+                    lock_array[pair_temp_indices.first].unlock();
```
matekelemen, line 1333: ~~I don't get it. Why is this in a naked parallel region in the first place?~~ nevermind, got it

matekelemen: actually, no I don't get it. For some reason it seems to me that we're computing the exact same thing on all threads, and the only reason why we're not getting duplicate results is because `set`s are being used.

loumalouomega: In fact, it is because we are using the locks

matekelemen: aight, @sunethwarna explained what's happening here

jcotela (approved):
👍 fixes a bug.

PR #10844, opened 2023-03-08
[Core] Add MapReducer to core

sunethwarna:
**📝 Description**
This adds `MapReducer` to core at `ReductionUtilities`.

**🆕 Changelog**
- Adds `MapReducer`
- Adds test

### DIFF
--- a/kratos/utilities/reduction_utilities.h
+++ b/kratos/utilities/reduction_utilities.h
@@ -302,9 +302,7 @@ public:
     void ThreadSafeReduce(const MapReduction<MapType>& rOther)
     {
         KRATOS_CRITICAL_SECTION
-        for (const auto& it : rOther.mValue) {
-            mValue.emplace(it);
-        }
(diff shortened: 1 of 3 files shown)


philbucher (commented):
I wrote this already somewhere, please wait with merging until I found it

kratos/utilities/reduction_utilities.h
```diff
@@ -278,6 +278,36 @@ class AccumReduction
     }
 };
 
+template<class MapType>
+class MapReduction
```
philbucher, line 282: Please check with what I added in #10339, I would think that `merge` is faster

In any case, can you then please also update the custom solution?

sunethwarna: done :D

philbucher (dismissed):
Thx

philbucher (approved):

PR #10846, opened 2023-03-08
[Core] adding base class controller

ddiezrod:
**📝 Description**
Adding a controller as discussed in https://github.com/KratosMultiphysics/Kratos/pull/10617#issuecomment-1443349441
This new class gives an interface for all objects of the form  input(Model, Parameters) -> bool.
I basically copied whatever is done for operations.

### DIFF
--- a/kratos/controllers/controller.h
+++ b/kratos/controllers/controller.h
@@ -58,7 +58,7 @@ public:
 
     /// Copy constructor.
     //TODO: Check. It is required by the registry
-    Controller(Controller const& rOther) {}
+    Controller(Controller const& rOther) = default;
 
     /// Move constructor
     Controller(Controller&& rOther) noexcept = default;
@@ -83,11 +83,7 @@ public:
      */
     virtual Controller::Pointer Create(
         Model& rModel,
-        Parameters ThisParameters) const
-    {
-        KRATOS_ERROR << "Calling base class Create. Please override this method in the corresonding Controller" << std::endl;
-        return nullptr;
-    }
+        Parameters ThisParameters) const = 0;
 
     /**
      * @brief Checks that input conditions are correct. 
@@ -100,31 +96,15 @@ public:
     /**
      * @brief Using input data, returns bool.
      */
-    virtual bool Evaluate()
-    {
-        KRATOS_ERROR << "Calling base class Evaluate. Please override this method in the corresonding Controller" << std::endl;
-    }
+    virtual bool Evaluate() = 0;
 
     /**
      * @brief This method provides the defaults parameters to avoid conflicts between the different constructors
      */
     virtual Parameters GetDefaultParameters() const
     {
-        KRATOS_ERROR << "Calling the base Controller class GetDefaultParameters. Please implement the GetDefaultParameters in your derived process class." << std::endl;
-        const Parameters default_parameters = Parameters(R"({})" );
-
-        return default_parameters;
+        return Parameters(R"({})" );
     }
-
-    ///@}
-    ///@name Access
-    ///@{
-
-
-    ///@}
-    ///@name Inquiry
-    ///@{
-
 
     ///@}
     ///@name Input and output
@@ -150,11 +130,7 @@ public:
     ///@}
 }; // Class Controller
 
-///@name Type Definitions
-///@{
 
-
-///@}
 ///@name Input and output
 ///@{
 
--- a/kratos/controllers/controller.h
+++ b/kratos/controllers/controller.h
@@ -78,7 +78,7 @@ public:
      * @brief This method creates an pointer of the controller
      * @details We consider as input, a Model and a set of Parameters for the sake of generality
      * @warning Must be overrided in controller implementation
-     * @param rModel The model to be consider
+     * @param rModel The model to be considered
      * @param ThisParameters The configuration parameters
      */
     virtual Controller::Pointer Create(
--- a/kratos/controllers/controller.h
+++ b/kratos/controllers/controller.h
@@ -18,9 +18,7 @@
 
 // Project includes
 #include "containers/model.h"
-#include "includes/kratos_flags.h"
 #include "includes/kratos_parameters.h"
-#include "includes/registry.h"
 
 namespace Kratos
 {
--- a/kratos/controllers/controller.h
+++ b/kratos/controllers/controller.h
@@ -51,7 +51,7 @@ public:
     ///@{
 
     /// Default constructor.
-    explicit Controller() = default;
+    Controller() noexcept = default;
 
     /// Destructor.
     virtual ~Controller() {}
--- a/kratos/controllers/controller.h
+++ b/kratos/controllers/controller.h
@@ -61,7 +61,7 @@ public:
     Controller(Controller const& rOther) {}
 
     /// Move constructor
-    Controller(Controller&& rOther) = default;
+    Controller(Controller&& rOther) noexcept = default;
 
     ///@}
     ///@name Operators
--- a/kratos/controllers/controller.h
+++ b/kratos/controllers/controller.h
@@ -1,0 +1,175 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
+//
+//  Main authors:    Daniel Diez
+//
+
+#pragma once
+
+// System includes
+
+// External includes
+
+// Project includes
+#include "containers/model.h"
+#include "includes/kratos_flags.h"
+#include "includes/kratos_parameters.h"
+#include "includes/registry.h"
+
+namespace Kratos
+{
+
+///@name Kratos Classes
+///@{
+
+/**
+ * @class Controller
+ * @ingroup KratosCore
+ * @brief The base  class for all Controllers in Kratos.
+ * @details Controller is the base class for all controllers and defines the interface for them.
+ * Each class must define their own Evaluate method.
+ * Controller parameters must be passed at construction time.
+  @author Daniel Diez
+*/
+class Controller
+{
+public:
+    ///@name Type Definitions
+    ///@{
+
+    /// Pointer definition of Controller
+    KRATOS_CLASS_POINTER_DEFINITION(Controller);
+
+    ///@}
+    ///@name Life Cycle
+    ///@{
+
+    /// Default constructor.
+    explicit Controller() = default;
+
+    /// Destructor.
+    virtual ~Controller() {}
+
+    /// Copy constructor.
+    //TODO: Check. It is required by the registry
+    Controller(Controller const& rOther) {}
+
+    /// Move constructor
+    Controller(Controller&& rOther) = default;
+
+    ///@}
+    ///@name Operators
+    ///@{
+
+    /// Assignment operator.
+    Controller& operator=(Controller const& rOther) = delete;
+
+    ///@}
+    ///@name Operations
+    ///@{
+
+    /**
+     * @brief This method creates an pointer of the controlle
+     * @details We consider as input a Model and a set of Parameters for the sake of generality
+     * @warning Must be overrided in controller implementation
+     * @param rModel The model to be consider
+     * @param ThisParameters The configuration parameters
+     */
+    virtual Controller::Pointer Create(
+        Model& rModel,
+        Parameters ThisParameters) const
+    {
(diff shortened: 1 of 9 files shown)


kratos/controllers/controller.h
```diff
@@ -0,0 +1,175 @@
...
+     */
+    virtual bool Evaluate()
+    {
+        KRATOS_ERROR << "Calling base class Evaluate. Please override this method in the corresonding Controller" << std::endl;
+    }
+
+    /**
+     * @brief This method provides the defaults parameters to avoid conflicts between the different constructors
+     */
+    virtual const Parameters GetDefaultParameters() const
```
matekelemen:
```suggestion
    virtual Parameters GetDefaultParameters() const
```

kratos/controllers/controller.h
```diff
@@ -0,0 +1,175 @@
...
+
+    /// Pointer definition of Controller
+    KRATOS_CLASS_POINTER_DEFINITION(Controller);
+
+    ///@}
+    ///@name Life Cycle
+    ///@{
+
+    /// Default constructor.
+    explicit Controller() = default;
```
matekelemen:
```suggestion
    explicit Controller() noexcept = default;
```

also, why `explicit`?

kratos/controllers/controller.h
```diff
@@ -0,0 +1,175 @@
...
+
+    /// Destructor.
+    virtual ~Controller() {}
+
+    /// Copy constructor.
+    //TODO: Check. It is required by the registry
+    Controller(Controller const& rOther) {}
+
+    /// Move constructor
+    Controller(Controller&& rOther) = default;
```
matekelemen:
```suggestion
    Controller(Controller&& rOther) noexcept = default;
```

kratos/controllers/controller.h
```diff
@@ -0,0 +1,175 @@
...
+     * @details We consider as input a Model and a set of Parameters for the sake of generality
+     * @warning Must be overrided in controller implementation
+     * @param rModel The model to be consider
+     * @param ThisParameters The configuration parameters
+     */
+    virtual Controller::Pointer Create(
+        Model& rModel,
+        Parameters ThisParameters) const
+    {
+        KRATOS_ERROR << "Calling base class Create. Please override this method in the corresonding Controller" << std::endl;
```
matekelemen: why not make these functions pure virtual?

kratos/controllers/controller.h
```diff
@@ -0,0 +1,175 @@
...
+///@name Type Definitions
+///@{
+
+
+///@}
+///@name Input and output
+///@{
+
+/// input stream function
+inline std::istream& operator >> (std::istream& rIStream, Controller& rThis);
```
matekelemen:
```suggestion
std::istream& operator >> (std::istream& rIStream, Controller& rThis);
```

kratos/controllers/controller.h
```diff
@@ -0,0 +1,175 @@
...
+
+/// output stream function
+inline std::ostream& operator << (std::ostream& rOStream, const Controller& rThis)
```
matekelemen:
```suggestion
std::ostream& operator << (std::ostream& rOStream, const Controller& rThis)
```

kratos/controllers/controller.h
ddiezrod: This is like this in other places, process.h and operation.h for example. It makes sense to me that you cannot modify the default parameters.

kratos/controllers/controller.h
ddiezrod: no reason, it was like that in operation.h but I think its confusing here.

kratos/controllers/controller.h
ddiezrod: Its also like this in process and operation, but I think its better to remove it as you suggest. There is clearly an overuse of "inline" in kratos code.

kratos/controllers/controller.h
matekelemen: The qualifiers of the variable that gets constructed has nothing to do with the return type of the function. The following is perfectly valid C++:
```cpp
struct S {};

const S immutableReturn()
{
    return S();
}

int main()
{
    S mutableVariable = immutableReturn();
}
```

Live example on [godbolt](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:12,endLineNumber:5,positionColumn:12,positionLineNumber:5,selectionStartColumn:12,selectionStartLineNumber:5,startColumn:12,startLineNumber:5),source:'struct+S+%7B%7D%3B%0A%0Aconst+S+immutableReturn()%0A%7B%0A++++return+S()%3B%0A%7D%0A%0Aint+main()%0A%7B%0A++++S+mutableVariable+%3D+immutableReturn()%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),header:(),k:50,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:executor,i:(argsPanelShown:'1',compilationPanelShown:'0',compiler:g122,compilerName:'',compilerOutShown:'0',execArgs:'',execStdin:'',fontScale:14,fontUsePx:'0',j:2,lang:c%2B%2B,libs:!(),options:'-std%3Dc%2B%2B17',source:1,stdinPanelShown:'1',tree:'1',wrap:'1'),l:'5',n:'0',o:'Executor+x86-64+gcc+12.2+(C%2B%2B,+Editor+%231)',t:'0')),header:(),k:50,l:'4',m:100,n:'0',o:'',s:0,t:'0')),l:'2',m:100,n:'0',o:'',t:'0')),version:4). Also #9922.

ddiezrod: Tricky! Thanks for the clarification.

kratos/controllers/controller.h
```diff
@@ -0,0 +1,175 @@
...
+        Parameters ThisParameters) const
+    {
+        KRATOS_ERROR << "Calling base class Create. Please override this method in the corresonding Controller" << std::endl;
+        return nullptr;
+    }
+
+    /**
+     * @brief Checks that input conditions are correct. 
+     */
+    virtual int Check()
```
philbucher:
```suggestion
    virtual int Check() const
```

kratos/controllers/controller.h
```diff
@@ -0,0 +1,175 @@
...
+
+    ///@}
+    ///@name Life Cycle
+    ///@{
+
+    /// Default constructor.
+    Controller() noexcept = default;
+
+    /// Destructor.
+    virtual ~Controller() {}
```
philbucher:
```suggestion
    virtual ~Controller() = default;
```

kratos/controllers/controller.h
```diff
@@ -0,0 +1,175 @@
...
+    {
+        KRATOS_ERROR << "Calling base class Evaluate. Please override this method in the corresonding Controller" << std::endl;
+    }
+
+    /**
+     * @brief This method provides the defaults parameters to avoid conflicts between the different constructors
+     */
+    virtual Parameters GetDefaultParameters() const
+    {
+        KRATOS_ERROR << "Calling the base Controller class GetDefaultParameters. Please implement the GetDefaultParameters in your derived process class." << std::endl;
```
rubenzorrilla: Not sure if we should throw an error in here. Note that we might want to add some common fields in here in the future (e.g. echo level).

kratos/controllers/register_controllers.cpp
```diff
@@ -0,0 +1,34 @@
+//    |  /           |
```
rubenzorrilla, line 1: @roigcarlo we should find the time to advance in the cpp registry stuff as this is starting to be populated...

(@ddiezrod please ignore this comment)

kratos/sources/kratos_application.cpp
```diff
@@ -129,6 +129,8 @@ void KratosApplication::RegisterKratosCore() {
     // Registering all the standard (model - parameters constructible) processes
     KratosApplication::RegisterProcesses();
 
+    KratosApplication::RegisterControllers();
```
rubenzorrilla, line 133:
```suggestion
    // Registering all the standard (model - parameters constructible) controllers
    KratosApplication::RegisterControllers();
```

rubenzorrilla (commented):
Aside of the minor comments this looks good to me. Basically it complies with all the requirements that we posted in the controller-based output PR.

Thanks for the implementation @ddiezrod.

kratos/controllers/controller.h
```diff
@@ -0,0 +1,175 @@
...
+
+    /// Assignment operator.
+    Controller& operator=(Controller const& rOther) = delete;
+
+    ///@}
+    ///@name Operations
+    ///@{
+
+    /**
+     * @brief This method creates an pointer of the controlle
```
sunethwarna:
```suggestion
     * @brief This method creates an pointer of the controller
```

kratos/controllers/controller.h
```diff
@@ -0,0 +1,175 @@
...
+     * @details We consider as input a Model and a set of Parameters for the sake of generality
```
sunethwarna:
```suggestion
     * @details We consider as input, a Model and a set of Parameters for the sake of generality
```

kratos/controllers/controller.h
```diff
@@ -0,0 +1,175 @@
...
+     * @warning Must be overrided in controller implementation
+     * @param rModel The model to be consider
```
sunethwarna:
```suggestion
     * @param rModel The model to be considered
```

kratos/controllers/controller.h
```diff
@@ -0,0 +1,175 @@
...
+    ///@name Access
+    ///@{
+
+
+    ///@}
+    ///@name Inquiry
+    ///@{
+
+
+    ///@}
```
sunethwarna: Do we need these empty groups?

kratos/controllers/controller.h
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    ///@}
+}; // Class Controller
+
+///@name Type Definitions
+///@{
+
+
+///@}
```
sunethwarna: Do we need these empty groups?

sunethwarna (commented):
Minor comments on documentation since, we are now going in the direction of creating nicer documentations for end-user and developers :dagger: 

Small question, Why isn't this exposed to python?

kratos/controllers/controller.h
ddiezrod: Well, I think that if I make this class abstract, we cannot register it or export it to python. If you are ok with this I will make these functions pure virtual @matekelemen @philbucher @rubenzorrilla @sunethwarna

matekelemen: > if I make this class abstract, we cannot register it or export it to python.

You'll need to define a trampoline class that handles the pure virtual members:
- [pybind docs](https://pybind11.readthedocs.io/en/stable/advanced/classes.html)
- [example in kratos](https://github.com/KratosMultiphysics/Kratos/blob/b04aa01d2f368c51db283cd501aafe7e9b42384b/applications/HDF5Application/custom_python/add_custom_io_to_python.cpp#L54)

I think it's worth putting in the effort because this is supposed to be a widely used base class.

ddiezrod: I'm learning a lot in this MR :D, I'll give it a try

loumalouomega: Technically, this is a PR

kratos/controllers/controller.h
ddiezrod: done!

kratos/controllers/controller.h
ddiezrod: done!

kratos/controllers/controller.h
ddiezrod: hmm ok, returning an empty parameters

kratos/controllers/controller.h
ddiezrod: works like a charm!

matekelemen: thx!

kratos/controllers/controller.h
```diff
@@ -0,0 +1,151 @@
...
+
+// System includes
+
+// External includes
+
+// Project includes
+#include "containers/model.h"
+#include "includes/kratos_flags.h"
+#include "includes/kratos_parameters.h"
+#include "includes/registry.h"
```
philbucher:
```suggestion
#include "containers/model.h"
#include "includes/kratos_parameters.h"
```

unused

rubenzorrilla (approved):
Very nice. Approving on behalf of the @KratosMultiphysics/technical-committee.

Thanks for the implementation @ddiezrod.

ddiezrod:
Thanks everybody for the review!

PR #10847, opened 2023-03-08
[CoSim] Restrict internal interfaces

matekelemen:
## Description
Restrict arguments passed to the constructors of `ConvergenceAcceleratorWrapper` and `ConvergenceCriteriaWrapper` to the objects they actually require, instead of passing entire `CoSimulationSolverWrapper`s.

This PR is an experiment to see how much opposition I'm gonna run into if I touch the `CoSimulationApplication`. 

## Motivation
I'm working on multiscale time integration, which would have a very similar abstraction system to the one implemented by `CoSimulationApplication`. The only issue is that solvers and the coupling logic share a common interface (`CoSimulationSolverWrapper`), which would need to be separated in my case. These commits represent the shy first steps in that direction.

## Changelog
- pass `CouplingInterfaceData` to `ConvergenceAcceleratorWrapper` and `ConvergenceCriteriaWrapper` instead of `CoSimulationSolverWrapper`
- factory signatures aren't changed, but they extract the necessary objects to pass on as arguments to the corresponding constructors
- update tests accordingly

### DIFF
--- a/applications/CoSimulationApplication/python_scripts/base_classes/co_simulation_solver_wrapper.py
+++ b/applications/CoSimulationApplication/python_scripts/base_classes/co_simulation_solver_wrapper.py
@@ -118,7 +118,7 @@ class CoSimulationSolverWrapper:
     def IsDefinedOnThisRank(self):
         return self.data_communicator.IsDefinedOnThisRank()
 
-    def GetInterfaceData(self, data_name):
+    def GetInterfaceData(self, data_name) -> CouplingInterfaceData:
         try:
             return self.data_dict[data_name]
         except KeyError:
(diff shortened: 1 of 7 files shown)


philbucher:
No real objections, those wrappers were only necessary for MPI

Can give a bit more context in the next days

philbucher (approved):
makes sense to me

Actually the two wrappers are only necessary for MPI, for cases when the underlying classes do not support MPI

Maybe we could remove them...

matekelemen:
> Actually the two wrappers are only necessary for MPI, for cases when the underlying classes do not support MPI

Can you please point me to what's being used if the classes **do** support MPI?

philbucher:
If they **do** support MPI then the gather (before) and scatter (after) is omitted, see e.g. the constant relaxation accelerator.

I don't think the effort of having the other classes support MPI would be too large, I guess it was more of a hack back in the day (we then had lots more restrictions, but thats a long story for another day)

Ideally we should use the accelerators from the FSI-App (moving them to the core first, which @rubenzorrilla and I already prepared)

The conv crits should be refactored, don't think it would be complicated.

PR #10859, opened 2023-03-13
[Core] Adding use_initial_stiffness option

AlejandroCornejo:
**📝 Description**

**This is a Draft!** Adding the option to compute the LHS only once at the begining of the simulation via:

`"use_initial_stiffness"    : true,` in the `ProjectParameters.json`

@aronnoordam @RiccardoRossi @rubenzorrilla

### DIFF
--- a/kratos/solving_strategies/strategies/residualbased_newton_raphson_strategy.h
+++ b/kratos/solving_strategies/strategies/residualbased_newton_raphson_strategy.h
@@ -901,23 +901,33 @@ class ResidualBasedNewtonRaphsonStrategy
         mpConvergenceCriteria->InitializeNonLinearIteration(r_model_part, r_dof_set, rA, rDx, rb);
         bool is_converged = mpConvergenceCriteria->PreCriteria(r_model_part, r_dof_set, rA, rDx, rb);
 
-        // Function to perform the building and the solving phase.
-        if (BaseType::mRebuildLevel > 0 || BaseType::mStiffnessMatrixIsBuilt == false) {
-            TSparseSpace::SetToZero(rA);
-            TSparseSpace::SetToZero(rDx);
-            TSparseSpace::SetToZero(rb);
+        const bool initial_stiffness = true;
 
-            if (mUseOldStiffnessInFirstIteration){
-                p_builder_and_solver->BuildAndSolveLinearizedOnPreviousIteration(p_scheme, r_model_part, rA, rDx, rb,BaseType::MoveMeshFlag());
-            } else {
-                p_builder_and_solver->BuildAndSolve(p_scheme, r_model_part, rA, rDx, rb);
-            }
-        } else {
+        // Function to perform the building and the solving phase.
+        if ((initial_stiffness && r_model_part.GetProcessInfo()[STEP] == 1)) {
+            p_builder_and_solver->BuildAndSolve(p_scheme, r_model_part, rA, rDx, rb);
+        } else if (initial_stiffness) {
             TSparseSpace::SetToZero(rDx);  // Dx = 0.00;
             TSparseSpace::SetToZero(rb);
-
             p_builder_and_solver->BuildRHSAndSolve(p_scheme, r_model_part, rA, rDx, rb);
+        } else {
+            if (BaseType::mRebuildLevel > 0 || BaseType::mStiffnessMatrixIsBuilt == false) {
+                TSparseSpace::SetToZero(rA);
+                TSparseSpace::SetToZero(rDx);
+                TSparseSpace::SetToZero(rb);
+
+                if (mUseOldStiffnessInFirstIteration){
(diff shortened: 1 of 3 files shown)


loumalouomega:
I agree with this, but this is delicate, please @RiccardoRossi take a look

AlejandroCornejo:
@lagoncalvesjr this could reduce dramatically the computational cost of HCF (we iterate more when damage increases but with low cost)

kratos/solving_strategies/strategies/residualbased_newton_raphson_strategy.h
```diff
@@ -901,24 +919,33 @@ class ResidualBasedNewtonRaphsonStrategy
...
-            TSparseSpace::SetToZero(rA);
-            TSparseSpace::SetToZero(rDx);
+        if ((GetUseInitialStiffnessFlag() && r_model_part.GetProcessInfo()[STEP] == 1)) {
+            p_builder_and_solver->BuildAndSolve(p_scheme, r_model_part, rA, rDx, rb);
+        } else if (GetUseInitialStiffnessFlag()) {
+            TSparseSpace::SetToZero(rDx);  // Dx = 0.00;
             TSparseSpace::SetToZero(rb);
+            p_builder_and_solver->BuildRHSAndSolve(p_scheme, r_model_part, rA, rDx, rb);
+        } else {
+            if (BaseType::mRebuildLevel > 0 || BaseType::mStiffnessMatrixIsBuilt == false) {
```
RiccardoRossi, line 932: you should use the value of mStiffnessMatrixISBuilt to build it for the first time

kratos/solving_strategies/strategies/residualbased_newton_raphson_strategy.h
```diff
@@ -901,24 +919,33 @@ class ResidualBasedNewtonRaphsonStrategy
         mpConvergenceCriteria->InitializeNonLinearIteration(r_model_part, r_dof_set, rA, rDx, rb);
         bool is_converged = mpConvergenceCriteria->PreCriteria(r_model_part, r_dof_set, rA, rDx, rb);
 
+
         // Function to perform the building and the solving phase.
-        if (BaseType::mRebuildLevel > 0 || BaseType::mStiffnessMatrixIsBuilt == false) {
-            TSparseSpace::SetToZero(rA);
-            TSparseSpace::SetToZero(rDx);
+        if ((GetUseInitialStiffnessFlag() && r_model_part.GetProcessInfo()[STEP] == 1)) {
```
RiccardoRossi: the checko on STEP is fragile ... it is better to rely on mStiffnessMatrixIsBuilt.

AlejandroCornejo:
> hi,
> 
> i am ok in principle with this change, however if you want it to be effective it is incomplete, since also preconditioners should only be mounted once

How can we do this? Who is in charge of that the BaS?

aronnoordam:
if  the "rebuild_level" is set on 0, doesnt this have the same effect?

AlejandroCornejo:
I tried to add `this->SetRebuildLevel(0);` at the beginning of the `SolveSolutionStep()` of the NR and it computes the LHS at each iteration anyways

RiccardoRossi:
> I tried to add `this->SetRebuildLevel(0);` at the beginning of the `SolveSolutionStep()` of the NR and it computes the LHS at each iteration anyways

we may have some long standing mistake here...

RiccardoRossi:
> > hi,
> > i am ok in principle with this change, however if you want it to be effective it is incomplete, since also preconditioners should only be mounted once
> 
> How can we do this? Who is in charge of that the BaS?

unfortunately this interacts with the linear solver (and preconditioners) interface. As i recall the interface has a Initialize function but it is not used in any of the solvers (or preconditioners). This is to say that there is not a easy solution to this... but we should tackle if we want to harvest all the possible advantages

AlejandroCornejo:
Without any action, when solving solid mech problems with pardiso it works well by activating the new feature. 

Do you mean that if I use AMGCL with preconditioning (changes LHS) the following steps will apply a duplication of the precond? If we build the LHS only once, we call the `BuildRHSAndSolve()`

RiccardoRossi:
i mean that if the system matrix is constant, then you could (should) factorize it once and then only solve the corresponding triangular system.

Similarly you could mount the preconditioner only once and use it many times (thing of amg, but also ilu, etc)

aronnoordam:
> I tried to add `this->SetRebuildLevel(0);` at the beginning of the `SolveSolutionStep()` of the NR and it computes the LHS at each iteration anyways

I also did ".SetStiffnessMatrixIsBuilt(True)" after the stiffnessmatrix was build for the first time. In this case, for me it only builds the LHS in the first iteration

AlejandroCornejo:
> > I tried to add `this->SetRebuildLevel(0);` at the beginning of the `SolveSolutionStep()` of the NR and it computes the LHS at each iteration anyways
> 
> I also did ".SetStiffnessMatrixIsBuilt(True)" after the stiffnessmatrix was build for the first time. In this case, for me it only builds the LHS in the first iteration

Nice! I suppose that the same problems arise with this option @RiccardoRossi

PR #10867, opened 2023-03-14
Prepare release 9.3.1

### LINKED ISSUE #10865: Move version string to interface layer
The version string for printing the python minor in the greeting message is creating a (very dark, hidden and nasty) compile time dependency on the kernel, which is preventing the re-usability of the kernel while targeting different interface version.

Mostly the problem is in `kratosCore.dir/flags.make`:
```make
flags.make:CXX_DEFINES = -DBOOST_UBLAS_MOVE_SEMANTICS -DKRATOS_CORE=IMPORT,API -DKRATOS_MAJOR_VERSION=9 -DKRATOS_MINOR_VERSION=3 -DKRATOS_PYTHON -DKRATOS_SMP_OPENMP -DKRATOS_USING_MPI -DKratosCore_EXPORTS -DPYTHON_VERSION_MAJOR=3 -DPYTHON_VERSION_MINOR=11 -DPYTHON_VERSION_PATCH=0 -DUSE_TRIANGLE_NONFREE_TPL
```

In order to prevent this we should initialize the python version AFTER the kernel has been loaded and otherwise assign a default in the kernel.

This is important as it reduces the current wheel release build time to 1/4.

roigcarlo:
**📝 Description**
@jginternational reported a problem with the registry and while fixing it I've encounter some major problems with the wheels, the kernel dependency on the interface, and various inconsistencies in the versioning numbers.

This aims to fix it all. @sunethwarna still looking into the stub problems. I will forward the fix here when its done.

This fixes #10865

**🆕 Changelog**
- Deleted copy of the __inti__.py file used by the wheelbuilder.
- FindPython.cmake no longer adds version definition to the whole project.
- Python version related `#defines` in `kratos_version` are now moved to the interface.
- Python version is now stored in the kernel.
- Python version is now registered in the kernel while loaded by python.
- `kernel` and `logger` greeting messages now use the `kernel` info rather than directly accessing `kratos_version` functions
- Update Wheel info with correct supported versions etc...


sunethwarna:
Hi @roigcarlo . Would it be possible to add `StatisticsApplication` as well for the release wheels?

roigcarlo:

Current release is broken, I need this ASAP

rubenzorrilla (approved):
Important fix.

PR #10876, opened 2023-03-14
[FEMDEM] App maintenance

AlejandroCornejo:
**📝 Description**
In this PR I am:

- Removing unused MEshing support within the FEMDEM
- Separating the app into CORE and Python interface
- Adding some missing nodal variables from the PFEM side


roigcarlo (commented):
:+1: From my side. I will leave @AFranci for approval since I am not sure about the changes in the scripts etc...

AFranci (approved):

PR #10878, opened 2023-03-14
[TrilinosApplication] More memory management improvement when doing BtDb and BDBt sparse multiplications

loumalouomega:
**📝 Description**

More memory management improvement when doing BtDb and BDBt sparse multiplications. This is less troublesome that current implementation.

**🆕 Changelog**

- [More memory management improvement when doing BtDb and BDBt sparse multiplications](https://github.com/KratosMultiphysics/Kratos/commit/be2ee3ddde8591d694be4d97f87849f5e652ee4d)
- [Update implementation to combine graphs](https://github.com/KratosMultiphysics/Kratos/pull/10878/commits/f151f92a0241d8736ebf6f67e4fce95eac8915c5)
- [Ensuring zero initialization](https://github.com/KratosMultiphysics/Kratos/pull/10878/commits/76e658ca1ecd5ab0555c30d2fefd8a5707c1802f)
- [Extend for cases of no same column map](https://github.com/KratosMultiphysics/Kratos/pull/10878/commits/284b3dc9c73793a7edd2a5f864d757cc98751c51)
- [Adding test for CombineMatricesGraphs](https://github.com/KratosMultiphysics/Kratos/pull/10878/commits/36abe27c4921b0d07178865534e14d3c355956ac)


loumalouomega:
Anyone can take a look at this?

jcotela (approved):

applications/TrilinosApplication/trilinos_space.h
```diff
@@ -69,7 +69,6 @@ namespace Kratos
  * @author Riccardo Rossi
  * @tparam TMatrixType The matrix type considered
  * @tparam TVectorType the vector type considered
- * @tparam TGraph The graph considered
```
loumalouomega: True, haha

applications/TrilinosApplication/trilinos_space.h
```diff
@@ -1004,6 +1016,91 @@ class TrilinosSpace
         KRATOS_CATCH("");
     }
 
+    /**
+     * @brief Generates a graph combining the graphs of two matrices
+     * @param rA The first matrix
+     * @param rB The second matrix
+     */
+    static Epetra_CrsGraph CombineMatricesGraphs(
```
jcotela, line 1024: If you need to use this function from the outside (not only from within this file) I would move this function to the trilinos utility class instead of the space.

If it is only an internal functionality, I would suggest making it private to make it clear that it is not part of the Space interface.

applications/TrilinosApplication/trilinos_space.h
```diff
@@ -418,26 +420,31 @@ class TrilinosSpace
...
 
-            // First multiplication
-            TransposeMult(rB, rD, aux, {true, false}, CallFillCompleteOnResult, true);
+            // Create an Epetra_Matrix
+            MatrixType* aux_3 =  new MatrixType(::Copy, CombineMatricesGraphs(rA, aux_2));
 
-            // Empty the solution Epetra_Matrix
-            SetToZero(rA);
+            // Copy values
+            CopyMatrixValues(*aux_3, aux_2);
```
RiccardoRossi, line 440: i am lost. you copy 3->2 and then swap A and 3? why? cannot you swap A and 2?

loumalouomega: The non-zeros are not the same, so if you try to add a value to the matrix you may get an error. If you define in the multiplication to keep the hard-non-zeros it will raise an error due to an error in Epetra, that the developers (I contacted them) don't plan to fix as Epetra is going to be gone in 2 years, and they recommend switching to Tpetra. This is the fix I could find after several days of debugging and testing. Everything related with sparse matrix multiplication in Epetra is very buggy (that are not going to be fixed), and I assure you this is the best I can do.

applications/TrilinosApplication/trilinos_space.h
loumalouomega: I need it to be public in order to add a test

PR #10881, opened 2023-03-15
[Structural] Eigenvalue postprocess avoid old results removal

rubenzorrilla:
**📝 Description**
As it is right now, the eigenvalue postprocess process in the StructuralMechanicsApplication always removes the existing results. I think this is very inconvenient, specially when trying to do mesh convergence analysis or running the case with different settings. Besides, this is done without warning the user, which is even worse in my opinion. Also note that the standard in Kratos is to overwrite existing results if these exist, not to wipe everything in the destination folder.

Having said this, I suggest removing the piece of code that automatically removes any existing result folder.


philbucher:
I see that it can be annoying 

However it also has a valid usecase IMO

Proposal: leave the removal as optional, but disable it by default

loumalouomega:
> I see that it can be annoying
> 
> However it also has a valid usecase IMO
> 
> Proposal: leave the removal as optional, but disable it by default

I support the proposal

rubenzorrilla:
> > I see that it can be annoying
> > However it also has a valid usecase IMO
> > Proposal: leave the removal as optional, but disable it by default
> 
> I support the proposal

Me too. I'll make the changes.

rubenzorrilla:
@loumalouomega @philbucher I made the changes according to the discussion. However, I did them with the idea of removing the "useless" Python factory module once we have the cpp registry finished. Long story short, this means to use the `std::filesystem::remove_all` to delete the folder. Note that this is now optional (as we agreed) according to the value of the new `wipe_results_folder`.

I also took the chance to do the following refactoring:
- Implement the defaults in the `GetDefaultParameters`
- Add a Model-Parameters constructor with the aim of using the new factories. Note that keeping current behavior (that is to take `Structure` as default `model_part_name` instead of assuming that this will be always provided prevents delegating constructors. In this regard, I also required to switch from a model part member reference var to a pointer one.
- The folder defaults in the Python factory layer were already present in the cpp one so there is no need to add them as the json validation system will do so later on.

applications/StructuralMechanicsApplication/custom_processes/postprocess_eigenvalues_process.cpp
```diff
@@ -145,28 +145,33 @@ struct VtkEigenOutputWrapper : public BaseEigenOutputWrapper
...
+    if (!std::filesystem::exists(folder_name)) {
+        std::filesystem::create_directories(folder_name);
+    }
+}
 
+PostprocessEigenvaluesProcess::PostprocessEigenvaluesProcess(
+    ModelPart& rModelPart,
+    Parameters OutputParameters)
+    : mpModelPart(&rModelPart)
+    , mOutputParameters(OutputParameters)
```
philbucher: why is this constructor still necessary?

I thought thats what you refactored? I.e. can it be removed?

rubenzorrilla: I left it in there because of backward compatibility. Adding to my comment that the position of the @KratosMultiphysics/technical-committee in this regard is to promote Model-parameters ones and allow alternative ones in case there is a technical need. 

Having said this, if you consider that we can remove it without any harm I'd remove it.

philbucher: as long as it works from python via project parameters I think we are ok

If someone complains we can still do sth abt it :)

rubenzorrilla: Very good. I'll remove it then.

rubenzorrilla: Done.

applications/StructuralMechanicsApplication/custom_processes/postprocess_eigenvalues_process.cpp
```diff
@@ -145,28 +145,17 @@ struct VtkEigenOutputWrapper : public BaseEigenOutputWrapper
...
-
+    mOutputParameters.RecursivelyValidateAndAssignDefaults(GetDefaultParameters());
+    mpModelPart = &(rModel.GetModelPart(mOutputParameters["model_part_name"].GetString()));
     const std::string folder_name = mOutputParameters["folder_name"].GetString();
+    if (mOutputParameters["wipe_results_folder"].GetBool()) {
+        std::filesystem::remove_all(folder_name);
+    }
     if (!std::filesystem::exists(folder_name)) {
         std::filesystem::create_directories(folder_name);
     }
```
philbucher:
```suggestion
    if (OutputParameters["save_output_files_in_folder"].GetBool()) {
        if (mOutputParameters["wipe_results_folder"].GetBool()) {
            std::filesystem::remove_all(folder_name);
        }
        if (!std::filesystem::exists(folder_name)) {
            std::filesystem::create_directories(folder_name);
        }
    }
```

final comment I promise :D

rubenzorrilla: Makes perfect sense :+1:

philbucher (approved):
nice

PR #10852, opened 2023-03-09
[OptApp] Enabling sensitivity computation on sub model parts using Adjont/Direct approach.

sunethwarna:
**📝 Description**
This PR enables computation of sensitivities on sub model parts using adjoint or direct approach.

The method `ModelPartUtils::GetModelPartsWithCommonReferenceEntitiesBetweenReferenceListAndExaminedList` creates sub model parts in each reference model part with common entities between reference model parts and the examined model parts. (Always reference model part entities are used to populate the new sub model part). If the parents are also required, then they are also added accordingly. This is tested in `OpenMP` and `MPI` as well.

Following changes are done to `ExecutionPolicy`. `ResponseFunction`, `LinearStrainEnergyResponseFunction` and `MassResponseFunction` to use above mentioned methodology.
1. `ExecutionPolicy::GetAnalysisModelPart` method is introduced to `IndependentAnalysisExecutionPolicy` and `SteppingAnalysisExecutionPolicy` to retrieve the `analysis model part`.
2. `ResponseFunction::CalculateSensitivity` signature is changed to have dict of `(sensitivity_variable: [ModelPart])` pairs.
3. `MassResponseFunction` was changed accordingly.
4. `LinearStrainEnergyResponseFunction` response function is based on adjoint method, hence it is required to pass the `analysis model part`. It is retrieved from the `primal` analysis by using the `ExecutionPolicy`. Because of that, I have to get the primal analysis from the `OptimizationInfo`.

All tests are passing with this mechanism as well.

**🆕 Changelog**
- Adds `ResponseUtils::GetModelPartsWithCommonReferenceEntitiesBetweenReferenceListAndExaminedList` method
- Adds `ResponseUtils::RemoveModelPartsWithCommonReferenceEntitiesBetweenReferenceListAndExaminedList` method
- Adds tests for above methods for `OpenMP` and `MPI`
- Added `ExecutionPolicy::GetAnalysisModelPart` method to `ExecutionPolicy`, `IndependentAnalysisExecutionPolicy`  and `SteppingAnalysisExecutionPolicy`.
- Update `ResponseFunction`, `LinearStrainEnergyResponseFunction` and `MassResponseFunction` to use above mentioned implementations.
- Cleaned `OptimizationUtils`

PR #10883, opened 2023-03-16
[Core] Template BlockPartition on Iterators Instead of Containers

matekelemen:
## Description

https://github.com/KratosMultiphysics/Kratos/blob/02d27883c115c5f5508e2349cd1cea162468c45b/kratos/utilities/parallel_utilities.h#L140-L145

`BlockPartition` is templated on the container type even though it doesn't use it anywhere within the class. Furthermore, it restricts the types of iterators it acts on to the one returned by `TContainerType::begin` by default. This PR makes `BlockPartition` template directly on the iterator type.

## Motivation

1) Kratos has some containers that  have multiple iterator types (eg.: pointer containers that have iterators returning dereferenced pointers, or ones returning the pointers themselves), and choosing between them requires explicitly specifying them in `BlockPartition`'s template argument list, which results in a lot more and unreadable code than a simple `block_for_each` call.
2) There's no way of creating a `BlockPartition` on a subset of a container's items, because it automatically creates a partitioning of **all** items within the container (hard-coded to `TContainerType::begin` and `TContainerType::end` in `block_for_each`).

Removing the container type from the template argument list solves both problems. Now there shouldn't be any reason to use `BlockPartition` directly (unless `MaxThreads` needs to be set manually); `block_for_each` can handle all cases.

## Changelog
- Remove `TContainerType` from `BlockPartition`'s template argument list.
- Remove the default type of `BlockPartition::TIteratorType`.
- add overloads to `block_for_each` taking iterator pairs.
- reroute `block_for_each` calls taking containers to ones taking iterator pairs.
- update tests directly using `BlockPartition`.
- replace direct uses of `BlockPartition` with `block_for_each`

### DIFF
--- a/kratos/tests/cpp_tests/utilities/test_parallel_utilities.cpp
+++ b/kratos/tests/cpp_tests/utilities/test_parallel_utilities.cpp
@@ -56,8 +56,9 @@ KRATOS_TEST_CASE_IN_SUITE(BlockPartitioner, KratosCoreFastSuite)
     std::vector<double> data_vector(nsize, 5.0);
 
     //here we raise every entry of a vector to the power 0.1
-    BlockPartition<std::vector<double>>(data_vector).for_each(
-                                         [](double& item)
+    BlockPartition<std::vector<double>::iterator>(data_vector.begin(),
+                                                  data_vector.end()).for_each(
+        [](double& item)
     {
         item = std::pow(item, 0.1);
     });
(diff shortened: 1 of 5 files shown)


loumalouomega (commented):
Okay from my side, if @KratosMultiphysics/technical-committee  agrees this can go

sunethwarna (commented):
Nice... Thanks @matekelemen . This is gonna be really nice on the PR #10852 :+1: . Lets keep it for some time, and then lets approve :D

sunethwarna (approved):
Thanks @matekelemen .

Issue #10880, opened 2023-03-15
[Core] Calling Clear model part "breaks" model part in FullDebug in Windows


ddiezrod:
This code throws an error in Windows in FullDebug (I have not checked on other platforms) when calling the second AddElement.

```
import KratosMultiphysics as KM


model = KM.Model()
model_part = model.CreateModelPart("main")
model_part.CreateNewNode(1,0,0,0)
model_part.CreateNewNode(2,0,0,0)
model_part.CreateNewNode(3,0,0,0)
model_part.CreateNewNode(4,0,0,0)
prop = model_part.CreateNewProperties(0)
new_elem = model_part.CreateNewElement("Element3D4N",1,[1,2,3,4],prop)
model_part.AddElement(new_elem)
model_part.Clear()
model_part.AddElement(new_elem)
print(model_part)
```

![image](https://user-images.githubusercontent.com/25667299/225359418-e966b28b-1b22-45cd-a858-99c68883fe32.png)

I think the reason is that when we call the Clear we are removing all mMeshes in model part 
https://github.com/KratosMultiphysics/Kratos/blob/94dfc0a01c20aae6e9949bb63934bc261fc28b49/kratos/sources/model_part.cpp#L84

but model part should always have at least one mesh (id 0) 
https://github.com/KratosMultiphysics/Kratos/blob/94dfc0a01c20aae6e9949bb63934bc261fc28b49/kratos/sources/model_part.cpp#L56

Is it really necessary to clear the meshes? Would we break anything if we remove that line?

ping @KratosMultiphysics/implementation-committee  @KratosMultiphysics/technical-committee @philbucher

matekelemen:
`clear` should leave an object in a default constructed state. `ModelPart` doesn't have a default constructor so I'd go with the constructor you pointed out => instead of removing the line, I'd follow it up with what the constructor does:
```cpp
mMeshes.clear();
MeshType mesh;
mMeshes.push_back(Kratos::make_shared<MeshType>(mesh.Clone()));
``` 

P.S.: maybe this is a good time to reflect on why we need to allocate memory 3 times and construct a `MeshType` twice, just to append an array with an empty object. How about this instead:
```diff
- MeshType mesh;
- mMeshes.push_back(Kratos::make_shared<MeshType>(mesh.Clone()));
+ mMeshes.emplace_back(Kratos::make_shared<MeshType>());
```

ddiezrod:
[...]

PR #10885, opened 2023-03-16
[Core] Pointer Vector - adding emplace back and improvig push back

ddiezrod:
Improving pointer vector as discussed in https://github.com/KratosMultiphysics/Kratos/issues/10880

Im  not sure if the implementation is correct, @matekelemen please take a look when you can.


loumalouomega (dismissed):

loumalouomega:
I would ask for a test...

loumalouomega (dismissed):

matekelemen:
Looks like you'll need to change to original `push_back` to take a const ref instead of a value to make the definitions unambiguous.

Also if you're adding tests, I'd add one where the number of moves/copies are tracked and tested. For example, something like [this](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIM6SuADJ4DJgAcj4ARpjE/gAcpAAOqAqETgwe3r7%2ByanpAiFhkSwxcWaJdpgOGUIETMQEWT5%2BAVU1AnUNBEUR0bEJtvWNzTltwz2hfaUDFQCUtqhexMjsHCYaAILmAMyhyN5YANQmO25OCgTEmKyn2BvbZnsMB17Hp24AbtVExHcPu32h0wJzOogUSka/y2D0uxC8DiOQgBAHYrFsjpikRA5ic0dZrCwACKYKhMLy0JoCOEIxzU05WFFEh5Y7HIswANnMHNxDFQmFUqySBDxVkshIAsqhvh4GDSOnKGSYmSysUIIGg5SL2TzRQTLCwPEkAJ6y%2BV0xU7RnMjFq7lHVBJWJMX6nIkQHXc3n8wWYYV68UGqXfTYQvDABhsQQKBlHa4EZYMI4AKgICDwMatyptm1ZOodTuILpIbo11O1XoDYoshsdxtDaQjUYImYsccwCeISdT6db2YBtsxlxdeGQR1CIuJpPJlLNV1pGT7g6Ow8cY4nRxYwcwc/hCqXuaxq9H48Em6NpvLe4tB9Zx/XZ630swDfDkcYLaVy/vp8nF9fTYfge/ZZjCWwbkIIAgFOZIUlSWrXouoJEkcGhfpsEFQU%2BMpXgu1LIah6GYdBF67nhcoEWhoEYWekHQduAHvtGlFEbRWH/mGgHMW6hHUQ8G4sEwoQ4qi6KHkOBDoFB3wOCWZzIjs2BHA0fx8cuKkAHRJF4CgIAA%2BlETDIAA1h6OJzOhrJMBCsQEB6WEkrBs64fuyE8VwFnUVZNmNPZ9HPmRrluu5nlid5kJ2XRtYmoFN5uTsKEaKFqqYtZEV%2BdhL6cUxLbxYlyXqT5kXsXWjHNgoeWoQV2zLkIK5UWFWKadpukGUZpkKElllNUVGWOTO8HmkhwUJUcZjVeFtkZdusXDQlPHjd1qW9VFpEuXFI0oR5S3KStWEMdl5WVV1Xk9elq2lYdQHHdVKV6sueYrlwO1WcQxBaTp%2BmGSZECXFJ0HPr9HkTayZ1TVF/VwbN%2BGbUcOwg6Du3nftAXrXNzKjQArAjoNpeDJUxWjMPzaN22nYjeO%2BVFB2NjlFWwydjUU3tJGXbTR0Mzj/YwiqWwcAstCcJjvB%2BBwWikKgnBuPqbYKEsKwgrsPCkAQmj8wsCA3FgcQ4qQxkgJjGj6Jwki8CwBtG6L4uSxwvAKCARuq2L/OkHAsBIAK1ReL85CUA0wAKMohiYLQQgIKgADuovK2gLBJHQI4CEHYSh%2BHUdq6Qsfx/QcTAFwZgBFndCxOErBrLwRc5wA8t7YeR1bvCe8gmzEAHnCN6o1R1Pgou8PwggiGI7BSDIgiKCo6jO6QuhcPohjGDL%2Bh4FE9uQAsjo3u3izLKseh/aEych3X6fcLwEdFkknA8ALQsixnNvYJ3yDeyQRyqPEHIALQcpIRzAMgY584aTMEcCA0tAw2COLgQgr8lZzF4E7LQcwNZawGLrfWhtjYcFNqQBuEst720dmrZBWCzB3ynjbBBxCFjfGIGkZwkggA%3D%3D%3D) but of course tweaked for `PointerVector`.

matekelemen:
But then you need to remove the `std::move` :D

kratos/containers/pointer_vector.h
```diff
@@ -267,11 +267,21 @@ class PointerVector final
         mData.swap(rOther.mData);
     }
 
-    void push_back(TPointerType x)
+    void push_back(const TPointerType& x)
     {
-        mData.push_back(x);
+        mData.push_back(std::move(x));
```
matekelemen:
```suggestion
        mData.push_back(x);
```

ddiezrod: makes sense XD

kratos/containers/pointer_vector.h
```diff
@@ -267,11 +267,21 @@ class PointerVector final
         mData.swap(rOther.mData);
     }
 
-    void push_back(TPointerType x)
+    void push_back(const TPointerType& x)
```
loumalouomega, line 270: Don't tell me that current push_back was doing copies?, that explains a lot... I think this is not a minor change anymore...

loumalouomega: A Ok, it is a pointer, in that case is not that bad

matekelemen: it did 2 copies:
- 1 copy when passing the pointer from userland to `PointerVector`
- 1 copy when passing the pointer from `PointerVector` to `std::vector`

Now it's one copy from `PointerVector` to `std::vector`

To completely bypass copies, you can pass a temporary to `emplace_back` (provided the pointer type has a move constructor).

And by the way, this is where `noexcept` move constructors come in. Try removing `noexcept` from the move constructor of the godbolt example from my previous example and see how the counters change.

ddiezrod:
> > It seems this does not work anymore
> 
> Since `push_back` is now overloaded, you need to specify which overload you want to bind. It's the python bindings so I suggest the const ref overload. [Here](https://pybind11.readthedocs.io/en/stable/classes.html#overloaded-methods) are the pybind docs on how to bind overloaded functions.

got it,  I think this should compile now

kratos/containers/pointer_vector.h
pooyan-dadvand: We should not use emplace_back here because we don't  want to store the original (smart) pointer here without adding to the reference counter. And if it is a normal pointer then it make no difference passing by const reference or value

matekelemen: this is `push_back`; `emplace_back` is another member function.

For now, let's forget about raw pointers because the compiler will optimize the differences between copy/move operations away in that case. When it comes to smart pointers, there are 2 situations in userland:
1) The user creates a new managed object from scratch for the sole purpose of inserting it into `PointerVector`.
2) The user inserts an existing smart pointer into `PointerVector`.

Currently, there's only one function to do both jobs:
https://github.com/KratosMultiphysics/Kratos/blob/b228896684903e6f2aefcae73e508f483d6f33fc/kratos/containers/pointer_vector.h#L270-L273

## Current Behaviour

### 1
```cpp
// userland
PointerVector<SmartPointer> pointer_vector;
SmartPointer pointer; // <== a new manged object is created
pointer_vector.push_back(pointer); // <== first copy happens here: 'pointer' is passed by value (moving would make no difference because push_back takes 'pointer' by value anyway)

// pointer_vector.h
void push_back(TPointerType x) 
{ 
    mData.push_back(x); // <== second copy happens here: 'x' is passed by reference and then copied at least once within std::vector
} 
```

*Ignoring the internals of `mData` that is an `std::vector`, here's the tally: there was 1 default construction and 2 copy constructions, resulting in incrementing the reference count a whopping 3 times (at least, because we're ignoring the internals of `std::vector`) and decrementing it twice.*

The user can help the compiler by constructing the managed object in place:
```cpp
// userland
pointer_vector.push_back(SmartPointer()); // <== non-mandatory copy elision
```
*This effectively eliminates the first copy, because the compiler is **allowed** to elide it and construct the managed object directly in the argument of `push_back`. However, the copy within `push_back` cannot be elided and so the tally is: 1 default construction, 1 copy construction ==> 2 reference count increments and 1 decrement.*

### 2
```cpp
// userland
PointerVector<SmartPointer> pointer_vector;
SmartPointer pointer; // <== a new manged object is created
pointer_vector.push_back(pointer); // <== first copy

// pointer_vector.h
void push_back(TPointerType x) 
{ 
    mData.push_back(x); // <== second copy within std::vector
} 
```
*Same thing happens as the case without copy elision. If we assume that `pointer` was already constructed and will survive the container's lifetime, we had 2 refcount increments and 1 decrement.*

## New Behaviour

### 1
```cpp
// userland
PointerVector<SmartPointer> pointer_vector;
SmartPointer pointer; // a new managed object is created
pointer_vector.push_back(pointer); // <== 'pointer' is now passed by reference so the refcount is untouched

// pointer_vector.h
void push_back(const TPointerType& x) 
{ 
    mData.push_back(x); // <== 'x' is passed by reference again, but is then copied within std::vector
} 
```

*Tally: 1 default construction  and 1 copy construction => 2 refcount increments and 1 decrement.*

Constructing in place has now completely different behaviour, because there's a move overload for `push_back`:
```cpp
// userland
PointerVector<SmartPointer> pointer_vector;
pointer_vector.push_back(SmartPointer()); <== a new managed object is created and the move constructor is elided

// pointer_vector.h
void push_back(TPointerType&& x) 
{ 
    mData.push_back(std::move(x)); // <== 'x' is moved, and then move constructed within `std::vector`
} 
```
*Tally: 1 Default construction, 3 moves => **1 refcount increment**.*

Even better, we can now use `emplace_back`:
```cpp
// userland
PointerVector<SmartPointer> pointer_vector;
pointer_vector.emplace_back(); // <== nothing happens here

// pointer_vector.h
template <class ...TArgs>
void emplace_back(TArgs&&... rArgs) 
{ 
    mData.emplace_back(std::forward<TArgs>(rArgs)...); // <== nothing happens here either (rArgs is empty), but the managed object is constructed somewhere down the line
} 
```

*Tally: 1 default construction => **1 refcount increment**.*

### 2
```cpp
// userland
PointerVector<SmartPointer> pointer_vector;
SmartPointer pointer; // a new managed object is created
pointer_vector.push_back(pointer); <== 'pointer' is now passed by reference so the refcount is untouched

// pointer_vector.h
void push_back(const TPointerType& x) 
{ 
    mData.push_back(x); // <== 'x' is passed by reference again, but is then copied within std::vector
} 
```
*If we assume that `pointer` was already constructed and will survive the container's lifetime, we had **1 refcount increment**.*

applications/HDF5Application/custom_python/add_custom_utilities_to_python.cpp
```diff
@@ -81,7 +81,9 @@ void AddCustomUtilitiesToPython(pybind11::module& rModule)
 
     pybind11::class_<HDF5::Detail::VertexContainerType, HDF5::Detail::VertexContainerType::Pointer>(rModule, "VertexContainer")
         .def(pybind11::init<>())
-        .def("push_back", &HDF5::Detail::VertexContainerType::push_back)
+        .def("push_back", static_cast<void (HDF5::Detail::VertexContainerType::*)
+            (HDF5::Detail::VertexContainerType::const_pointer&)>
+            (&HDF5::Detail::VertexContainerType::push_back))
```
matekelemen:
```suggestion
        .def("push_back", pybind11::overload_cast<const HDF5::Detail::VertexContainerType::pointer&>(&HDF5::Detail::VertexContainerType::push_back))
```

we're on C++17, so I think this should also work but admittedly, it's not much nicer in this case so I'll leave it up to you

matekelemen (dismissed):

kratos/containers/pointer_vector.h
pooyan-dadvand: Thanks @matekelemen  for your comprehensive answer!!

Some comments:
1. At first I missed understood you and thought you want to use emplace back inside the push back.
2. This code was written far before move semantic exist in C++
3. In modern C++, to my understanding, if you want to take ownership of the shared pointer you would take the argument by value (to explicitly announce that you are sharing ownership) and user should move it if he wants. (Of course this is arguable)

Saying that I am open to the changes in this branch

loumalouomega:
Okay now from my side

rubenzorrilla:
Ok also from my side. Changes are for good and current implementation is optimal.

ddiezrod:
So can this be approved? @sunethwarna @roigcarlo @rubenzorrilla @matekelemen ?

roigcarlo (commented):
To me it can go in

ddiezrod:
Merging master, just in case. It seems everybody agrees, so can someone approve it?

rubenzorrilla (approved):
Approving to be merged after CI checks.

ddiezrod:
Thanks @rubenzorrilla !

PR #10886, opened 2023-03-16
[TrilinosApplication] Adding support for MPC in `TrilinosBlockBuilderAndSolver`

loumalouomega:
**📝 Description**

With this PR the `TrilinosBlockBuilderAndSolver` will support **MPC**, giving the same results as in serial (for the sake of consistency between MPI/serial). Includes the preceptive tests replicating the one in serial.

This is a big PR, it includes the following PR changes that must be merged before than this one:

1. [[TrilinosApplication] Unify behavior between the serial/MPI BlockBuilderandSolver](https://github.com/KratosMultiphysics/Kratos/pull/10706). This is already an important change, and the first step to unify behavioir serial/MPI.
2. [[TrilinosApplication] More consistent and rigorous checks](https://github.com/KratosMultiphysics/Kratos/pull/10825) 
3. [[TrilinosApplication] More memory management improvement when doing BtDb and BDBt sparse multiplications](https://github.com/KratosMultiphysics/Kratos/pull/10878)
4. [[TrilinosApplication] Adding SetValue methods and corresponding tests for TrilinosAssemblingUtilities](https://github.com/KratosMultiphysics/Kratos/pull/11060)

I may create another intermediate PR, as I refactored a bit of the `TrilinosBlockBuilderAndSolver` in order to behave the closest possible to the serial one. **Update: https://github.com/KratosMultiphysics/Kratos/pull/10887**

**🆕 Changelog**

- Changes from #10706
- Changes from #10825
- Changes from #10878
- Changes from #11060
- Changes from #11090
- Changes from #11273
- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11385
- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11342
- Adding support for **MPC** in `TrilinosBlockBuilderAndSolver`
- Extend C++ B&S tests for **MPC**


loumalouomega:
@RiccardoRossi this is ready for review

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
```diff
@@ -509,7 +525,7 @@ class TrilinosBlockBuilderAndSolver
         using DofsVectorType = Element::DofsVectorType;
 
         // Gets the array of elements from the modeler
-        DofsVectorType dof_list;
+        DofsVectorType dof_list, second_dof_list;
```
RiccardoRossi, line 528: what is the "second_dof_list"?

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
```diff
@@ -610,6 +637,24 @@ class TrilinosBlockBuilderAndSolver
         // by Riccardo ... it may be wrong!
         mFirstMyId = free_offset - mLocalSystemSize;
 
+        // For MPC we fill mFirstMyIds
+        if (rModelPart.GetCommunicator().GlobalNumberOfMasterSlaveConstraints() > 0) {
+            mFirstMyIds.resize(world_size);
+            mFirstMyIds[current_rank] = mFirstMyId;
+            const int tag_id_send = 0;
+            for (int i_rank = 0; i_rank < world_size; ++i_rank) {
```
RiccardoRossi: this is not a "broadcast"? 

https://mpitutorial.com/tutorials/mpi-broadcast-and-collective-communication/

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
```diff
@@ -817,12 +864,108 @@ class TrilinosBlockBuilderAndSolver
...
+
+        if (rModelPart.GetCommunicator().GlobalNumberOfMasterSlaveConstraints() > 0) {
+            BuildMasterSlaveConstraints(rModelPart);
+
+            // Reference to T
+            const TSystemMatrixType& r_T = *mpT;
+
+            // Compute T' b
+            const TSystemVectorType copy_b(rb);
+            TSparseSpace::TransposeMult(r_T, copy_b, rb);
```
RiccardoRossi, line 878: this operation may fail, as T^t@b may be of different size than b

also one of the problems is that if we assign and after this rb changed its size, it is not any longer apt for use in assembly

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
```diff
@@ -817,12 +864,108 @@ class TrilinosBlockBuilderAndSolver
...
+        KRATOS_TRY
+
+        if (rModelPart.GetCommunicator().GlobalNumberOfMasterSlaveConstraints() > 0) {
+            BuildMasterSlaveConstraints(rModelPart);
+
+            // Reference to T
+            const TSystemMatrixType& r_T = *mpT;
+
+            // Compute T' A T
+            const TSystemMatrixType copy_A(rA);
```
RiccardoRossi, line 918: careful! the sparsity structure of rA after this may be changed, so it may well not be possible to assemble on the top of it

(this is also a problem of the current "Serial" implementation)

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
```diff
@@ -817,12 +864,108 @@ class TrilinosBlockBuilderAndSolver
...
+        TSystemVectorType& rb
+        ) override
+    {
+        KRATOS_TRY
+
+        if (rModelPart.GetCommunicator().GlobalNumberOfMasterSlaveConstraints() > 0) {
+            BuildMasterSlaveConstraints(rModelPart);
+
+            // Reference to T
+            const TSystemMatrixType& r_T = *mpT;
```
RiccardoRossi: shouldn't you use the function

GetConstraintRelationMatrix

here?

applications/TrilinosApplication/tests/cpp_tests/test_builder_and_solver.cpp
```diff
@@ -551,6 +578,281 @@ namespace Kratos::Testing
         TrilinosCPPTestUtilities::CheckSparseMatrix(rA_scale, row_indexes, column_indexes, values);
     }
 
+    /**
+    * Checks if the block builder and solver with constraints performs correctly the assemble of the system
+    */
+    KRATOS_DISTRIBUTED_TEST_CASE_IN_SUITE(TrilinosBasicDisplacementBlockBuilderAndSolverWithConstraints, KratosTrilinosApplicationMPITestSuite)
```
RiccardoRossi, line 720: i would advise to check the MPC case in 2 cases that could be complicated:

1 - reduce to one single master_dof, and then try to do again the assembly (if i understood right the code, that will fail)
2 - try one case in which the master dof is one that was NOT originally included in the list by which you assembled the matrix. (imagine you want to add as a DOF a translation of the entire system)

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
loumalouomega: The same thing for the serial B&S, for the constraints (master/slave dofs)

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
loumalouomega: T is squared as in serial, we are considering all dofs

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
loumalouomega: In fact if you look at the details of the implementation you will see that there are some tricks in order to avoid problems with Trilinos, this is because I found a bug in Epetra that a priori Trilinos is not going to fix as Epetra is already legacy.

loumalouomega: I assure it works, I did many checks

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
loumalouomega: Could be

applications/TrilinosApplication/tests/cpp_tests/test_builder_and_solver.cpp
loumalouomega: Okay

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
RiccardoRossi: naming is not really fortunate. not a blocker though

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
RiccardoRossi: it could still be larger. Imagine you have a master dof that is not any of the original dofs of the problem

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
loumalouomega: In the serial implementation is like that: https://github.com/KratosMultiphysics/Kratos/blob/88350e25f3c1fdedf341e7ec657ca51bfcf3fd76/kratos/solving_strategies/builder_and_solvers/residualbased_block_builder_and_solver.h#L723 (also mine :P)

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
loumalouomega: That dof is added in the global system as well (see serial implementation)

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
loumalouomega: Indeed I think can be simplified with a  broadcast

loumalouomega: Even better, AllGather

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
loumalouomega: If you have any better suggestion I will change it

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
loumalouomega: If you have questions about details on implementation, maybe it would be better to speak directly, but I will consider this conversation as resolved.

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
loumalouomega: I will resolve the conversation, if you have more questions we can speak directly.

applications/TrilinosApplication/tests/cpp_tests/test_builder_and_solver.cpp
loumalouomega: First case will not fail, I will show you. Neither the second one, I will add it to prove you (also in serial).

loumalouomega: I added serial version in https://github.com/KratosMultiphysics/Kratos/pull/11385, I will add MPI version now. The behaviour is the same, so the same result is expected.

loumalouomega: Well, the assembling per se work with one core, but there is a communication fail with more partitions, so at least has served to find a bug...

loumalouomega: Okay, I found the bug, if a constraint is defined in more than one partition it fails. This happens when the master slave nodes are in more than one partition (like in the case you suggested asking to set everything to one master dof, which implies to duplicate the constraint with nodes shared between partitions). Thinking in a solution.

loumalouomega: Okay, I think I have the solution, which luckily only implies to properly set the test. One sec.

loumalouomega: Anyway, I will add a check in debug

loumalouomega: Huuum, thanks to this test I found a bug, but not in what @RiccardoRossi expected, I am doing something weird in this corner case, and I still don't know exactly what

loumalouomega: Okay I found the issue, I was creating nodes in wrong partitions...

loumalouomega: Solved

loumalouomega:
@RiccardoRossi check again. Ask me if any question.

loumalouomega:
> @RiccardoRossi check again. Ask me if any question, may be interesting to solve remaining questions in a online meet.

Please, @RiccardoRossi take a look again at this, I applied you commentaries

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
RiccardoRossi: i guess that mps_dofs_list or something of the sort would be more suitable...but let's leave it for a different PR

RiccardoRossi (approved):
ok, it looks ok to me.

crossing my fingers that there are no hidden destructive changes...:-D

PR #10890, opened 2023-03-17
[Docs] Update Annotations in ResidualBasedBossakScheme

matekelemen:
As the title suggests, doc updates only.


loumalouomega (approved):

matekelemen:
yeah I realized that halfway through editing it. I was looking for `ResidualbasedBossakDisplacementScheme` and probably misclicked :(

anyway, thanks

PR #10900, opened 2023-03-21
[Core] Adding property accessor class and properties operations

AlejandroCornejo:
**📝 Description**
This is a work in progress PR following the discussion of https://github.com/KratosMultiphysics/Kratos/pull/9053. In this PR I have added:

- A new class called `Accessor` that will serve as a basis for commonly used accesors (i.e. retrieve from a table) (to be completed in this PR!)

Inside the  `properties.h` I have added 

- An empty `std::unordered_map<KeyType, AccessorType> mAccessors`
- `AddAccessor`method
- A new typedef: `typedef std::function<double(const KeyType, const Properties *, const Geometry<Node<3>>&, const Vector&, const ProcessInfo&)> AccessorType;`
- A new custom `GetValue(...)` to be used together with accessors. This methods checks wether a certain `Variable.Key()` is associated with an `Accessor` and retrieves its custom implementation.

Minimal test:

~~~cpp
Model current_model;
auto &r_model_part = current_model.CreateModelPart("ModelPart",1);
auto p_prop = r_model_part.CreateNewProperties(0);
p_prop->SetValue(YOUNG_MODULUS, 2.1e11);

const double E = ((*p_prop)[YOUNG_MODULUS]);
KRATOS_WATCH(E) // E=2.1e11

// custom accessor that returns 2.0
class CornejoAccessor {
	// Types
	public:
		// we define here the custom funtion accessor
	Properties::AccessorType mAccessor = [](const Properties::KeyType VarId, const Properties *pProps, const Geometry<Node<3>> &rGeom, const Vector &rN, const ProcessInfo &rProcessInfo) -> double
	{
		return 2.0;
	};
	private:
};
// custom accessro tat return 3.0
class CornejoAccessor2 {
	// Types
	public:
	// we define here the custom funtion accessor
	Properties::AccessorType mAccessor = [](const Properties::KeyType VarId, const Properties *pProps, const Geometry<Node<3>> &rGeom, const Vector &rN, const ProcessInfo &rProcessInfo) -> double
	{
		return 3.0;
	};

	private:
};

// Aux stuff...
auto& r_process_info = r_model_part.GetProcessInfo();
r_model_part.CreateNewNode(1 , 0.0 , 1.0 , 1.0);
r_model_part.CreateNewNode(2 , 0.0 , 1.0 , 0.0);
r_model_part.CreateNewNode(3 , 0.0 , 0.0 , 1.0);
r_model_part.CreateNewNode(4 , 1.0 , 1.0 , 1.0);
r_model_part.CreateNewNode(5 , 0.0 , 0.0 , 0.0);
r_model_part.CreateNewNode(6 , 1.0 , 1.0 , 0.0);
r_model_part.CreateNewNode(7 , 1.0 , 0.0 , 1.0);
r_model_part.CreateNewNode(8 , 1.0 , 0.0 , 0.0);
// Now we create the elements
r_model_part.CreateNewElement("SmallDisplacementElement3D8N", 1, {{5,8,6,2,3,7,4,1}}, p_prop);
Vector N;
auto geom = r_model_part.GetElement(1).GetGeometry();

auto cornejo_accessor = CornejoAccessor();
p_prop->AddAccessor(YOUNG_MODULUS, std::move(cornejo_accessor.mAccessor));

double Emod = p_prop->GetValue(YOUNG_MODULUS, geom, N, r_process_info);
KRATOS_WATCH(Emod) // expected 2.0

auto cornejo_accessor_2 = CornejoAccessor2();
p_prop->AddAccessor(YOUNG_MODULUS, std::move(cornejo_accessor_2.mAccessor));

Emod = p_prop->GetValue(YOUNG_MODULUS, geom, N, r_process_info);
KRATOS_WATCH(Emod) // expected 3.0
~~~

@RiccardoRossi @roigcarlo @rubenzorrilla @loumalouomega 

Questions:

- What if the variable is a `Vector `? Currently the `std::function` always returns a `double`...
- Who and where creates this custom_accessors?

### DIFF
--- a/kratos/includes/properties.h
+++ b/kratos/includes/properties.h
@@ -11,9 +11,7 @@
 //                   Riccardo Rossi
 //
 
-
-#if !defined(KRATOS_PROPERTIES_H_INCLUDED )
-#define  KRATOS_PROPERTIES_H_INCLUDED
+# pragma once
 
 // System includes
 #include <string>
@@ -645,5 +643,3 @@ inline std::ostream& operator << (std::ostream& rOStream,
 
 
 }  // namespace Kratos.
-
-#endif // KRATOS_PROPERTIES_H_INCLUDED  defined
--- a/kratos/includes/properties.h
+++ b/kratos/includes/properties.h
@@ -90,7 +90,7 @@ public:
     typedef IndexType KeyType;
 
     // Accessor type: std::function (rVariableKey, Properties, Geometry, ShapeFunctions, ProcessInfo)
-    typedef std::function<double(const KeyType VariableKey, const Properties *, const Geometry<Node<3>>&, const Vector&, const ProcessInfo&)> AccessorType;
+    typedef std::function<double(const KeyType, const Properties *, const Geometry<Node<3>>&, const Vector&, const ProcessInfo&)> AccessorType;
 
     typedef std::unordered_map<std::size_t, TableType> TablesContainerType; // This is a provisional implementation and should be changed to hash. Pooyan.
 
--- a/kratos/includes/properties.h
+++ b/kratos/includes/properties.h
@@ -87,8 +87,10 @@ public:
 
     typedef Table<double> TableType;
 
+    typedef IndexType KeyType;
+
     // Accessor type: std::function (rVariableKey, Properties, Geometry, ShapeFunctions, ProcessInfo)
-    typedef std::function<double(const IndexType VariableKey, const Properties *, const Geometry<Node<3>>&, const Vector&, const ProcessInfo&)> AccessorType;
+    typedef std::function<double(const KeyType VariableKey, const Properties *, const Geometry<Node<3>>&, const Vector&, const ProcessInfo&)> AccessorType;
 
     typedef std::unordered_map<std::size_t, TableType> TablesContainerType; // This is a provisional implementation and should be changed to hash. Pooyan.
 
@@ -570,6 +572,8 @@ private:
     TablesContainerType mTables;                /// The tables contained on the properties
 
     SubPropertiesContainerType mSubPropertiesList; /// The vector containing the list of subproperties
+
+    std::unordered_map<KeyType, AccessorType> mAccessors  = {};
 
     ///@}
     ///@name Private Operators
--- a/kratos/includes/properties.h
+++ b/kratos/includes/properties.h
@@ -24,6 +24,8 @@
 // Project includes
 #include "includes/define.h"
 #include "includes/node.h"
+#include "geometries/geometry.h"
+#include "includes/variables.h"
 #include "includes/indexed_object.h"
 #include "containers/data_value_container.h"
 #include "includes/process_info.h"
@@ -84,6 +86,9 @@ public:
     typedef NodeType::IndexType IndexType;
 
     typedef Table<double> TableType;
+
+    // Accessor type: std::function (rVariableKey, Properties, Geometry, ShapeFunctions, ProcessInfo)
+    typedef std::function<double(const IndexType VariableKey, const Properties *, const Geometry<Node<3>>&, const Vector&, const ProcessInfo&)> AccessorType;
 
     typedef std::unordered_map<std::size_t, TableType> TablesContainerType; // This is a provisional implementation and should be changed to hash. Pooyan.
 
--- a/kratos/includes/properties.h
+++ b/kratos/includes/properties.h
@@ -258,6 +258,27 @@ public:
         return rThisNode.GetValue(rV, SolutionStepIndex);
     }
 
+    /*
+    Custom GetValue in which we check the Accessor
+    */
+    template<class TVariableType, typename ...TArgs>
+    double GetValue(const TVariableType& rVariable, TArgs... Args) {
+        const KeyType variable_id = rVariable.Key();
+        auto value = mAccessors.find(variable_id);
+        if (value != mAccessors.end()) {
+            return value->second(variable_id, this, Args...);
+        } else {
+            return GetValue(TVariableType);
+        }
+    }
+
+    /* 
+    Method to add Accessors to properties
+    */
+    void AddAccessor(KeyType VariableKey, AccessorType && Accessor) {
+        mAccessors[VariableKey] = Accessor;
+    }
+
     template<class TVariableType>
     void SetValue(TVariableType const& rV, typename TVariableType::Type const& rValue)
     {
--- a/kratos/includes/properties.h
+++ b/kratos/includes/properties.h
@@ -268,15 +268,17 @@ public:
         if (value != mAccessors.end()) {
             return value->second(variable_id, this, Args...);
         } else {
-            return GetValue(TVariableType);
+            return GetValue(rVariable);
         }
     }
 
     /* 
     Method to add Accessors to properties
     */
-    void AddAccessor(KeyType VariableKey, AccessorType && Accessor) {
-        mAccessors[VariableKey] = Accessor;
+    template<class TVariableType>
+    void AddAccessor(const TVariableType& rVariable, AccessorType && Accessor) {
+        const IndexType variable_key = rVariable.Key();
+        mAccessors[variable_key] = Accessor;
     }
 
     template<class TVariableType>
--- a/kratos/includes/accessor.h
+++ b/kratos/includes/accessor.h
@@ -1,0 +1,107 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
+//
+//  Main authors:    Alejandro Cornejo
+//                   Riccardo Rossi
+//                   Carlos Roig
+//
+//
+
+# pragma once
+
+// External includes
+
+// Project includes
+#include "includes/define.h"
+#include "includes/variables.h"
+
+namespace Kratos
+{
+///@addtogroup KratosCore
+///@{
+
+///@name Kratos Classes
+///@{
+
+/**
+ * @class Accessor
+ * @ingroup StructuralMechanicsApplication
+ * @brief This class defines the way a certain property is accessed to
+ * @author Alejandro Cornejo, Riccardo Rossi and Carlos Roig
+ */
+class KRATOS_API(KRATOS_CORE) Accessor
+{
+    public:
+        ///@name Type Definitions
+        ///@{
+
+
+        /// Pointer definition of NodeSearchUtility
+        KRATOS_CLASS_INTRUSIVE_POINTER_DEFINITION(Accessor);
+
+        ///@}
+        ///@name Life Cycle
+        ///@{
+
+        /// Default constructor.
+        Accessor()
+        {}
+
+        /// Destructor.
+        virtual ~Accessor() {}
+
+        ///@}
+        ///@name Operations
+        ///@{
+
+
+        ///@}
+        ///@name Input and output
+        ///@{
+
+        /// Turn back information as a string.
+        virtual std::string Info() const
+        {
+            std::stringstream buffer;
+            buffer << "Accessor" ;
+
+            return buffer.str();
+        }
+
+        /// Print information about this object.
+        virtual void PrintInfo(std::ostream& rOStream) const  {rOStream << "Accessor";}
+
+        /// Print object's data.
+        virtual void PrintData(std::ostream& rOStream) const {}
+
+        ///@}
+
+    private:
+
+        ///@name Member Variables
+        ///@{
+
+
+        friend class Serializer;
+
+        void save(Serializer& rSerializer) const
+        {
+        }
+
+        void load(Serializer& rSerializer)
+        {
+        }
+
+
+}; // class
+///@}
+
+///@} addtogroup block
+
+} // namespace Kratos
--- a/kratos/includes/accessor.h
+++ b/kratos/includes/accessor.h
@@ -31,7 +31,7 @@ namespace Kratos
 
 /**
  * @class Accessor
- * @ingroup StructuralMechanicsApplication
+ * @ingroup Kratos Core
  * @brief This class defines the way a certain property is accessed to
  * @author Alejandro Cornejo, Riccardo Rossi and Carlos Roig
  */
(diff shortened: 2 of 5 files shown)


loumalouomega:
> **📝 Description** This is a work in progress PR following the discussion of #9053. In this PR I have added:
> 
>     * A new class called `Accessor` that will serve as a basis for commonly used accesors (i.e. retrieve from a table) (to be completed in this PR!)
> 
> 
> Inside the `properties.h` I have added
> 
>     * An empty `std::unordered_map<KeyType, AccessorType> mAccessors`
> 
>     * `AddAccessor`method
> 
>     * A new typedef: `typedef std::function<double(const KeyType, const Properties *, const Geometry<Node<3>>&, const Vector&, const ProcessInfo&)> AccessorType;`
> 
>     * A new custom `GetValue(...)` to be used together with accessors. This methods checks wether a certain `Variable.Key()` is associated with an `Accessor` and retrieves its custom implementation.
> 
> 
> Minimal test:
> 
> ```c++
> Model current_model;
> auto &r_model_part = current_model.CreateModelPart("ModelPart",1);
> auto p_prop = r_model_part.CreateNewProperties(0);
> p_prop->SetValue(YOUNG_MODULUS, 2.1e11);
> 
> const double E = ((*p_prop)[YOUNG_MODULUS]);
> KRATOS_WATCH(E) // E=2.1e11
> 
> // custom accessor that returns 2.0
> class CornejoAccessor {
> 	// Types
> 	public:
> 		// we define here the custom funtion accessor
> 	Properties::AccessorType mAccessor = [](const Properties::KeyType VarId, const Properties *pProps, const Geometry<Node<3>> &rGeom, const Vector &rN, const ProcessInfo &rProcessInfo) -> double
> 	{
> 		return 2.0;
> 	};
> 	private:
> };
> // custom accessro tat return 3.0
> class CornejoAccessor2 {
> 	// Types
> 	public:
> 	// we define here the custom funtion accessor
> 	Properties::AccessorType mAccessor = [](const Properties::KeyType VarId, const Properties *pProps, const Geometry<Node<3>> &rGeom, const Vector &rN, const ProcessInfo &rProcessInfo) -> double
> 	{
> 		return 3.0;
> 	};
> 
> 	private:
> };
> 
> // Aux stuff...
> auto& r_process_info = r_model_part.GetProcessInfo();
> r_model_part.CreateNewNode(1 , 0.0 , 1.0 , 1.0);
> r_model_part.CreateNewNode(2 , 0.0 , 1.0 , 0.0);
> r_model_part.CreateNewNode(3 , 0.0 , 0.0 , 1.0);
> r_model_part.CreateNewNode(4 , 1.0 , 1.0 , 1.0);
> r_model_part.CreateNewNode(5 , 0.0 , 0.0 , 0.0);
> r_model_part.CreateNewNode(6 , 1.0 , 1.0 , 0.0);
> r_model_part.CreateNewNode(7 , 1.0 , 0.0 , 1.0);
> r_model_part.CreateNewNode(8 , 1.0 , 0.0 , 0.0);
> // Now we create the elements
> r_model_part.CreateNewElement("SmallDisplacementElement3D8N", 1, {{5,8,6,2,3,7,4,1}}, p_prop);
> Vector N;
> auto geom = r_model_part.GetElement(1).GetGeometry();
> 
> auto cornejo_accessor = CornejoAccessor();
> p_prop->AddAccessor(YOUNG_MODULUS, std::move(cornejo_accessor.mAccessor));
> 
> double Emod = p_prop->GetValue(YOUNG_MODULUS, geom, N, r_process_info);
> KRATOS_WATCH(Emod) // expected 2.0
> 
> auto cornejo_accessor_2 = CornejoAccessor2();
> p_prop->AddAccessor(YOUNG_MODULUS, std::move(cornejo_accessor_2.mAccessor));
> 
> Emod = p_prop->GetValue(YOUNG_MODULUS, geom, N, r_process_info);
> KRATOS_WATCH(Emod) // expected 3.0
> ```
> 
> @RiccardoRossi @roigcarlo @rubenzorrilla @loumalouomega
> 
> Questions:
> 
>     * What if the variable is a `Vector `? Currently the `std::function` always returns a `double`...
> 
>     * Who and where creates this custom_accessors?

I read ConejoAccesor and I imagined Bugs Bunny looking at a table

kratos/includes/properties.h
```diff
@@ -253,6 +258,29 @@ class Properties : public IndexedObject
         return rThisNode.GetValue(rV, SolutionStepIndex);
     }
 
+    /*
+    Custom GetValue in which we check the Accessor
+    */
+    template<class TVariableType, typename ...TArgs>
+    double GetValue(const TVariableType& rVariable, TArgs... Args) {
```
loumalouomega, line 265: This means that old GetValue still exists?, avoiding the penalty?

kratos/includes/properties.h
```diff
@@ -568,6 +596,8 @@ class Properties : public IndexedObject
 
     SubPropertiesContainerType mSubPropertiesList; /// The vector containing the list of subproperties
 
+    std::unordered_map<KeyType, AccessorType> mAccessors  = {};
```
loumalouomega, line 599: Huuum, maybe is a good time to introduce the flat maps from boost (C++23...), because it is faster when reading (not for writing)

kratos/includes/properties.h
```diff
@@ -253,6 +258,29 @@ class Properties : public IndexedObject
         return rThisNode.GetValue(rV, SolutionStepIndex);
     }
 
+    /*
+    Custom GetValue in which we check the Accessor
+    */
+    template<class TVariableType, typename ...TArgs>
+    double GetValue(const TVariableType& rVariable, TArgs... Args) {
+        const KeyType variable_id = rVariable.Key();
+        auto value = mAccessors.find(variable_id);
```
loumalouomega, line 267: Technically it_value. I would like to check how to do it to minimize the overhead

kratos/includes/properties.h
AlejandroCornejo: exacty!

loumalouomega: In that case we should study using this: https://www.boost.org/doc/libs/1_81_0/libs/unordered/doc/html/unordered.html#compliance_open_addressing_containers_unordered_flat_set_unordered_flat_map

sunethwarna: The variadic templates are unnecessary in here if we are fixing the `AccessorType` signature.

AlejandroCornejo:
> This is really cool, I also like the general idea. I am just putting my opinions here for the record.
> 
>     1. If we go for this custom `GetValue` signature, then if this is to be used within existing elements, then the elements needs to be updated to this custom one. Then there is no point in maintaining the old `GetValue`. Then there will be an overhead cost for the elements which does not need this feature.
> 
>     2. If we are allowing both, and some elements use the old `GetValue` and some elements use the new `GetValue`, then it will be hard for debugging. Having two mechanisms for the same thing is dangerous in my opinion. (deprecating is fine for the time being if we go for the new one). As an example, if some one sees these new properties defined in the `materials.json` and they copy it to their case which uses elements with the old `GetValue`, at least there should be a way to put a warning so the debugging time can be saved.
> 
> 
> > What if the variable is a `Vector `? Currently the `std::function` always returns a `double`...
> 
> May be we can have the `AccessorType` to be dependent on the variable type, and use `std::variant` in the storage map of the `AccessorType` lambdas. AFAIK, in this case, the memory usage in `std::variant` will be almost the same as storing the lambda function in the map you proposed.

I'd say that this new `GetValue()` will be used only for those situations in which is needed... I think that in this way the old ones (standard GetValue) is not affected neither in performance or implementation whereas the more complex cases have this new  feature available.

kratos/includes/properties.h
roigcarlo: Agree. The Varargs were suggested in case we wanted to be flesible with the signature (see std::palceholder). If we stick to the accessor signature you can just remove that and add the defaults in the calls.

kratos/includes/properties.h
roigcarlo: Not sure flatmap can be used inside the properties. We need to at least assume that we know beforehand the number of registered Accessors, which is not something available at compile time as every one (and every different instance) could hook / have hooked different ones.

matekelemen:
The system for getting `Element` properties definitely needs an overhaul, but I'm not sure this is the right way to do it (I don't have a catch-em-all solution in mind either).

First and foremost, using `std::function` in the hottest region we have in Kratos will be insanely expensive. On top of that, the function to be called needs to be looked up from a map, which again happens in the hot region. Readability and debugging is also an issue when it comes to invoking an arbitrary function that was looked up from a map that can change anywhere at runtime. If you're **absolutely certain** you need dynamic polymorphism here, I suggest defining a `virtual` overload of `Properties::GetValue` and get your properties from derived classes. This would replace the `map` + `std::function` combo with a single virtual lookup.

However, I strongly suggest finding a completely compile-time solution, since every cycle counts here and dynamic polymorphism has a pretty hefty price when compared to simple material property calculations. How about defining a `Property::GetValue` that is templated on a function that computes the desired variable?
```cpp
double CornejoAccessor(const Properties::KeyType VarId, const Properties *pProps, const Geometry<Node<3>> &rGeom, const Vector &rN, const ProcessInfo &rProcessInfo)
{
    return 2.0;
}

template <class TAccessor, class TArgs...>
auto Property::GetValue(TAccessor&& rAccessor, const TArgs&...)
{
    return rAccessor(TArgs...);
}


auto p_props = r_model_part.CreateNewProperties(0);
const auto E = p_properties->GetValue(CornejoAccessor, YOUNG_MODULUS, geom, N, r_process_info);
```

Or something along these lines. You could store the function pointer somewhere (eg `Element` or some object related to it) and overwrite it with whatever you need if you must set it at runtime.

```cpp
class WhateverElement : public Element
{
    decltype(CornejoAccessor) mAccessor;

    auto ComputeWhatever() 
    {
        return mpProperties->GetValue(mAccessor, you_get_the_idea...);
    }
};
```

There's still a bit of potential performance hit, because the compiler cannot `inline` the accessor since it's hidden behind a pointer, so the best solution is still to implement the property querying logic within the framework of your specific `Element` types.

loumalouomega:
> The system for getting `Element` properties definitely needs an overhaul, but I'm not sure this is the right way to do it (I don't have a catch-em-all solution in mind either).
> 
> First and foremost, using `std::function` in the hottest region we have in Kratos will be insanely expensive. On top of that, the function to be called needs to be looked up from a map, which again happens in the hot region. Readability and debugging is also an issue when it comes to invoking an arbitrary function that was looked up from a map that can change anywhere at runtime. If you're **absolutely certain** you need dynamic polymorphism here, I suggest defining a `virtual` overload of `Properties::GetValue` and get your properties from derived classes. This would replace the `map` + `std::function` combo with a single virtual lookup.
> 
> However, I strongly suggest finding a completely compile-time solution, since every cycle counts here and dynamic polymorphism has a pretty hefty price when compared to simple material property calculations. How about defining a `Property::GetValue` that is templated on a function that computes the desired variable?
> 
> ```c++
> double CornejoAccessor(const Properties::KeyType VarId, const Properties *pProps, const Geometry<Node<3>> &rGeom, const Vector &rN, const ProcessInfo &rProcessInfo)
> {
>     return 2.0;
> }
> 
> template <class TAccessor, class TArgs...>
> auto Property::GetValue(TAccessor&& rAccessor, const TArgs&...)
> {
>     return rAccessor(TArgs...);
> }
> 
> 
> auto p_props = r_model_part.CreateNewProperties(0);
> const auto E = p_properties->GetValue(CornejoAccessor, YOUNG_MODULUS, geom, N, r_process_info);
> ```
> 
> Or something along these lines. You could store the function pointer somewhere (eg `Element` or some object related to it) and overwrite it with whatever you need if you must set it at runtime.
> 
> ```c++
> class WhateverElement : public Element
> {
>     decltype(CornejoAccessor) mAccessor;
> 
>     auto ComputeWhatever() 
>     {
>         return mpProperties->GetValue(mAccessor, you_get_the_idea...);
>     }
> };
> ```
> 
> There's still a bit of potential performance hit, because the compiler cannot `inline` the accessor since it's hidden behind a pointer, so the best solution is still to implement the property querying logic within the framework of your specific `Element` types.

I basically agree with everything you said. We need to add that benchmark library if we want to actually measure performance impact.

sunethwarna:
Adding onto this, a small test on virtual call is done for properties in #10622 already.

matekelemen:
> If I understand your solution correctly, the problem is that the element will need to know all about the accessors. I think the whole point of @AlejandroCornejo 's implementation is that you only need to change the call to GetValue in your element, and you will be able to use all kratos elements with minimum changes, as the accessors can be added from anywhere to the properties...

That's correct, and that's the feature I want you to double check you actually need. If you **do** need it, I still believe adding a `virtual` `GetValue` is a cleaner and less pricey solution. The only feature you'd lose in that case is the ability to **define** the accessors at runtime because they'd have to be defined in derived classes.

RiccardoRossi:
@matekelemen we cannot use compile time polymorphyism as the idea is to be able to do this at runtime.

indeed my original idea is that a property would have a single accessor, however in that case you would be moving the map from the properties into that accessor in case you want to access two different things (imagine you need to access YOUNG_MODULUS and POISSON_RATIO as a function of the temperature. How would u do that?)

actually the solution of using std::functions is based on the assumption (to be tested) that std::function is cheaper than a virtual call. I would expect that this is the case, exactly as in the case of using a virtual function you will to check in the virtual table for the correct call.
no doubts i can be very wrong on this.

RiccardoRossi:
just to tell that this post 

https://stackoverflow.com/questions/62627564/overhead-of-stdfunction-vs-virtual-function-call-for-type-erasure

seems to suggest that @matekelemen is right and virtual functions will be faster. Cool news as it is also a cleaner solution.

btw: the idea of having a different function is exactly that it should be used parsimoniously as it implies an overhead

AlejandroCornejo:
I will create another branch and PR with this approach

sunethwarna:
@AlejandroCornejo, If you want this capability to **only** function in the new elements you design, then wouldn't it make sense to create a variable like `ADVANCED_MATERIAL` with class type `AdvancedMaterial` and design this base class to have some generic virtual methods so everyone can design their own advanced materials rather than narrowing down scope to one `virtual GetValue` method. (The design needs to be finalized). Then you can use this variable within your new elements, and do advanced material stuff. This gives more flexibility than having another `virtual GetValue` with a differen signature because, in future, someone else might need to have some more information passed to `virtual GetValue`, and then again we have to start this discussion again. This way, `Properties` stays the same... @RiccardoRossi @rubenzorrilla @roigcarlo

Issue #10898, opened 2023-03-21
[GeoMechanicsApplication] Slicing issue reported by SonarCloud


avdg81:
**Description**
SonarCloud reports the following slicing issue in `applications/GeoMechanicsApplication/custom_conditions/U_Pw_normal_lysmer_absorbing_condition.cpp`:

> Use pointer or reference to avoid slicing from "Node<3>" to "array_1d<double, 3>".

A previous attempt to fix this issue resulted in a failure of a long-running test on our build server. It appeared that failure was triggered by using some uninitialized variable(s). In previous commits, some uninitialized variables have been resolved. Therefore, I'd like to try again to fix this slicing issue.

**Scope**
The slicing issue is reported by [SonarCloud](https://sonarcloud.io/project/overview?id=KratosGeomechanics).

Issue #7414, opened 2020-09-03
Problem with using auto for array_1d with GCC


philbucher:
**Description**
@sunethwarna found that the following code fails when using GCC:

~~~cpp
array_1d<double, 3> rNormal; // also properly initialized ...
const auto result = rNormal / norm_2(rNormal);
~~~

with Clang the type of `result` is correctly determined as `array_1d<double, 3>`
With GCC however the resulting type is `const boost::numeric::ublas::vector_binary_scalar2<Kratos::array_1d<double, 3>, const double, boost::numeric::ublas::scalar_divides<double, double> >` which gives problems later on

**Scope**
- KratosCore

**To Reproduce**
use `auto`, see above

**Expected behavior**
Gcc should produce the same as clang

**Environment**
Tested on multiple machines with different versions of Clang and Gcc. Always works with Clang and always fails with Gcc
E.g.:
- OS: Linux (different ones)
- Branch: master
- Gcc 5.4, Gcc 9.3, Gcc 10

any idea why this is happening?
@KratosMultiphysics/technical-committee

roigcarlo:
Mm no idea to be honest, I will investigate

roigcarlo:
@philbucher @sunethwarna , which version of boost are you using?

philbucher:
I think 1.73

sunethwarna:
I am using 1.72

pooyan-dadvand:
Seems that you are creating an expression of division instead of the result array. So if the expression is used in another place after the original normal array is gone of scope. It will be a memory error and it will mysteriously crashed. 

As a rule of thumb, you should avoid auto for the expressions unless you are sure about the life time of the terms involved in the expression.

pooyan-dadvand:
Also performance is an issue here. Because if you use result 3 times, you are evaluating the division 3 times.

pooyan-dadvand:
So in any case I would suggest to store the values in a array_1d instead of creating auto expression:

```c++
const array_1d<double,3> result = rNormal / norm_2(rNormal);
```

or directly

```c++
rNormal /= norm_2(rNormal);
```

philbucher:
FYI @armingeiser

armingeiser:
Thanks for the explanation! I solved a similar problem like this:
```cpp
auto new_array = rNormal;
new_array /= norm_2(rNormal);
```

Is there any chance such a not recommended use of auto for expressions is detected by the compiler?

[...]

PR #10903, opened 2023-03-21
[GeoMechanicsApplication] Fixed a slicing issue

avdg81:
**📝 Description**
Avoid constructing a (temporary) object, which slices off a part of the original object. Use a reference-to-const instead.

This fixes issue #10898.

### DIFF
--- a/applications/GeoMechanicsApplication/custom_conditions/U_Pw_normal_lysmer_absorbing_condition.cpp
+++ b/applications/GeoMechanicsApplication/custom_conditions/U_Pw_normal_lysmer_absorbing_condition.cpp
@@ -639,7 +639,7 @@ void UPwLysmerAbsorbingCondition<3, 4>::CalculateRotationMatrix( BoundedMatrix<d
     //Quadrilateral_3d_4
     array_1d<double, 3> p_mid_0;
     array_1d<double, 3> p_mid_1;
-    const auto& r_p_2 = array_1d<double, 3>(rGeom.GetPoint(2));
+    const array_1d<double, 3>& r_p_2 = rGeom.GetPoint(2);
     noalias(p_mid_0) = 0.5 * (rGeom.GetPoint(0) + rGeom.GetPoint(3));
     noalias(p_mid_1) = 0.5 * (rGeom.GetPoint(1) + r_p_2);
 


mnabideltares (dismissed):

applications/GeoMechanicsApplication/custom_conditions/U_Pw_normal_lysmer_absorbing_condition.cpp
```diff
@@ -639,7 +639,7 @@ void UPwLysmerAbsorbingCondition<3, 4>::CalculateRotationMatrix( BoundedMatrix<d
     //Quadrilateral_3d_4
     array_1d<double, 3> p_mid_0;
     array_1d<double, 3> p_mid_1;
-    const auto& r_p_2 = array_1d<double, 3>(rGeom.GetPoint(2));
+    const auto& r_p_2 = rGeom.GetPoint(2);
```
philbucher:
```suggestion
    const array_1d<double, 3>& r_p_2 = rGeom.GetPoint(2);
```

I recommend not to use `auto` with `array_1d<double, 3>`, as in the past GCC had issues deriving the correct types, leading to very strange behaviors

avdg81: Hi @philbucher, thanks for pointing out this issue, I was not aware of that. Do you recall which version of GCC that was? Also, what kind of strange behavior are you referring to? Could it be that the issue has been resolved by now, since I have no failing builds (including GCC builds)?

In general, I'm not in favor of explicitly repeating type names, since that makes the code less flexible with respect to future changes and extensions. I will discuss the point that you've raised with one of my colleagues tomorrow to see what is the best way to proceed. Thanks.

philbucher: See #7414

I think it was with GCC8/9/10, but do not remember

I think [this](https://github.com/KratosMultiphysics/Kratos/issues/7414#issuecomment-704121996) is a good guideline, so in your case it would most probably be ok

avdg81: I have discussed your comment with a colleague and I feel that we need to have a deeper understanding here. To us it seems that the problem that you had raised in the past (#7414) may be related to lazy evaluation of the expression. We have the impression that by being explicit about the type the evaluation may be triggered (in particular when a type conversion is required like here). Again, we need to dig deeper first to gain a better understanding. For now I will go with your suggestion, but I may come back to it later. Thanks anyway for your comments.

mnabideltares (approved):

PR #10909, opened 2023-03-22
Adding an option to evaluate the stabilization tau at nodes for the level set convection process. 

uxuech:
**📝 Description**

With this PR, it is now possible for the SUPG element in the level set convection process to evaluate the stabilization coefficient tau at nodes. By evaluating the stabilization coefficient at nodes the  jump of its value between elements decreases.  The purpose of this is to avoid artificial bumbles for the two fluid simulations .

**🆕 Changelog**

 The modified files are: 

- _test_levelset_convection.py._ 
         A test has been added where the nodal tau option is activated. 
- _levelset_convection_process.h_
         The nodal tau is calculated for each node. 
- _levelset_convection_element_simplex.h_ 
         For each gauss point the calculated nodal tau is interpolated.

### DIFF
--- a/kratos/elements/levelset_convection_element_simplex.h
+++ b/kratos/elements/levelset_convection_element_simplex.h
@@ -110,6 +110,21 @@ public:
         KRATOS_CATCH("");
     }
 
+    void Initialize(const ProcessInfo& rProcessInfo) override
+    {
+        std::size_t count = 0;
+        for (const auto& r_node : GetGeometry()) {
+            if(r_node.Has(TAU)){
+                count++;
+            }
+        }
+        if (count == TNumNodes) {
+            mElementTauNodal = true;
+        } else {
+            mElementTauNodal = false;
+        }
+    }
+    
     void CalculateLocalSystem(MatrixType& rLeftHandSideMatrix, VectorType& rRightHandSideVector, const ProcessInfo& rCurrentProcessInfo) override
     {
         KRATOS_TRY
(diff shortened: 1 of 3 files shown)


kratos/elements/levelset_convection_element_simplex.h
```diff
@@ -110,6 +110,21 @@ class LevelSetConvectionElementSimplex
...
+        for (const auto& r_node : GetGeometry()) {
+            if(r_node.Has(TAU)){
+                count++;
+            }
+        }
+        if (count == TNumNodes) {
+            mElementTauNodal = true;
+        } else {
+            mElementTauNodal = false;
+        }
```
matekelemen: Assuming `TNumNodes` is equal to the number of nodes in `GetGeometry()`:

```suggestion
        const auto& r_geometry = GetGeometry();
        mElementTauNodal = std::all_of(r_geometry.begin(),
                                       r_geometry.end(),
                                       [](const auto& rNode) {
                                           rNode.Has(TAU)
                                       });
```

If not, you can use [std::count_if](https://en.cppreference.com/w/cpp/algorithm/count) instead.

matekelemen: P.S. the failures could indicate that `mElementTauNodal` is not computed correctly, which might mean that `TNumNodes < GetGeometry().size()` in some cases. Can you please add a debug check in the constructor that makes sure that `TNumNodes == GetGeometry.size()` if they're supposed to be equal?

rubenzorrilla: I think we can safely assume that `TNumNodes` equals the geometry size. Note that this is specified at the specialization level and that in the case of a wrong specialization sneaking in, this would result in a segfault when calculating the local system contributions of the element.

Having said this, I'd directly add your suggestion as it is. BTW, I wasn't aware of this `std::all_of`, pretty nice for this kind of tests (which we do a lot in the standard and unfitted fluid). Thanks for the hint!

kratos/processes/levelset_convection_process.h
```diff
@@ -371,6 +373,8 @@ class LevelSetConvectionProcess
 
     bool mElementTauNodal; 
 
+    bool mHIsCalculated = true;
```
rubenzorrilla:
```suggestion
    bool mCalculateNodalH = true;
```
`mHIsCalculated` is way misleading IMO (what comes to my mind is exactly the contrary to current behavior).

kratos/processes/levelset_convection_process.h
```diff
@@ -176,14 +177,27 @@ class LevelSetConvectionProcess
...
             const auto it_node = mpDistanceModelPart->NodesBegin() + i_node;
             mVelocity[i_node] = it_node->FastGetSolutionStepValue(*mpConvectVar);
             mVelocityOld[i_node] = it_node->FastGetSolutionStepValue(*mpConvectVar,1);
             mOldDistance[i_node] = it_node->FastGetSolutionStepValue(*mpLevelSetVar,1);
+            double velocity_norm = norm_2(mVelocity[i_node]);
+            double nodal_h = it_node-> GetValue(NODAL_H);
+
+            if (mElementTauNodal)
+            {
+                double tau = 1.0 / (dynamic_tau / previous_delta_time + velocity_norm / std::pow(nodal_h,2));
```
rubenzorrilla, line 198: Shouldn't we use current delta time in here instead of the previous one?

rubenzorrilla (approved):
I think that this is ready now.

kratos/processes/levelset_convection_process.h
uxuech: It is the currrent delta time but its name is previous_delta_time.

PR #10911, opened 2023-03-23
[Core] Automatic prototypes registry

rubenzorrilla:
First, this has been done together with @roigcarlo.

This PR adds a new `KRATOS_REGISTRY_ADD_PROTOTYPE` to a new `define_registry.h`. This macro is intended to be used for the automatic registry of any class prototype. As an example of the usage, current PR adds it to the KratosCore base processes `Process` and `OutputProcess`.

Important comments/thoughts:
1) Everything is based on firstly registering a lambda function instead of the prototype. The reasoning behind this is
- In order to force the class to execute the registering code at initialize time we need "something static"
- If we define the variables outside the class, either in a cpp (which not all the classes have) or in the bottom part of the header (which would be creating one static variable per compilation unit)
- Hence, we want this to be inside the definition of the class.
- If we want it to be inside the definition of the class, we need the class to be completed in order to have its prototype :arrow_forward:  **BOOM. This is impossible**.
- In order to be able to do this we delegate the creation of the prototype to a lambda function that will be called once the first time we take the value from the corresponding registry entry.

2) To trigger the execution of such lambda automatically, the register macro creates a static bool within the class that is being registered. 

3) Note that in order to materialize this, we required to introduce some changes in the current `RegistryItem`. These are:
- `mpCallable` holds the reference to the prototype. This means that the `mpValue` holds the reference to the lambda function.
- The corresponding `AddCallable` function (equivalent to the previous `AddItem`) has been created to be called by the registry macros.
- `AddItem` function has been change to check if the item to be added is a `RegistryItem` (i.e. another dictionary level) or a value item, which will be wrapped inside a new lambda function that is created in there.
- Constructors have changed in order to add the possibility to create a `RegistryItem` with a function as value. The previous constructors have been also changed in order to initialize `mpValue` to a function that returns the provided value.
- Finally we have substituted the `mGetValueStringMethod` by a `const std::string mpValueName` to be able to get the string from a lambda which is potentially not possible.


kratos/includes/define_registry.h
```diff
@@ -0,0 +1,53 @@
...
+#pragma once
+
+/* System includes */
+#include <stdexcept>
+#include <sstream>
+
+
+/* External includes */
+
+/* Project includes */
```
loumalouomega:
```suggestion
// System includes
#include <stdexcept>
#include <sstream>

// External includes

// Project includes
```

loumalouomega: TOC

rubenzorrilla: @roigcarlo and I would appreciate if we can leave style comments until we come up with a final solution to avoid polluting the discussion. Thanks.

kratos/includes/registry_item.h
```diff
@@ -181,23 +199,70 @@ class KRATOS_API(KRATOS_CORE) RegistryItem
...
+    RegistryItem &AddItem(
+        std::string const &ItemName,
+        TArgumentsList &&...Arguments)
     {
         KRATOS_ERROR_IF(this->HasItem(ItemName))
             << "The RegistryItem '" << this->Name() << "' already has an item with name "
             << ItemName << "." << std::endl;
 
-        using ValueType = typename std::conditional<std::is_same<TItemType, RegistryItem>::value, SubRegistryItemFunctor, SubValueItemFunctor<TItemType>>::type;
+        if constexpr (std::is_same<TItemType, RegistryItem>::value == true) {
```
sunethwarna:
```suggestion
        if constexpr (std::is_same_v<TItemType, RegistryItem>) {
```

pooyan-dadvand:
@roigcarlo seems that Centos don't like the registry!

kratos/includes/registry_item.h
```diff
@@ -228,11 +292,39 @@ class KRATOS_API(KRATOS_CORE) RegistryItem
...
+        using TFunctionReturnType = std::shared_ptr<TDataType>;
+        using TFunctionType = std::function<TFunctionReturnType()>;
+
+        // Assign callable value
+        TFunctionType func = std::any_cast<TFunctionType>(mpValue);
+        mpCallable = func();
+
+        // Set value name
+        std::stringstream buffer;
+        buffer << *(std::any_cast<TFunctionReturnType>(mpCallable));
```
matekelemen: the naming is **very** confusing. To my understanding, `mpCallable` is not a callable, but a pointer to the stored type. On the other hand, `mpValue` **is** actually a callable `std::function` instance and not a pointer, let alone a pointer to the stored type.

kratos/includes/registry_item.h
```diff
@@ -228,11 +292,39 @@ class KRATOS_API(KRATOS_CORE) RegistryItem
...
+    TDataType const &GetValue()
+    {
+        KRATOS_TRY
+
+        // This is executed the first time we access the GetValue for this item
+        using TFunctionReturnType = std::shared_ptr<TDataType>;
+
+        if (std::any_cast<TFunctionReturnType>(&mpCallable) == nullptr) {
+            this->EvaluateValue<TDataType>();
+        }
```
matekelemen: [`std::optional`](https://en.cppreference.com/w/cpp/utility/optional)

kratos/includes/registry_item.h
```diff
@@ -181,23 +199,69 @@ class KRATOS_API(KRATOS_CORE) RegistryItem
...
+            KRATOS_ERROR_IF_NOT(insert_result.second)
+                << "Error in inserting '" << ItemName
+                << "' in registry item with name '" << this->Name() << "'." << std::endl;
+
+            return *insert_result.first->second;
+        }
+        else
+        {
+            std::function<std::shared_ptr<TItemType>()> aux_lambda = [=]() -> std::shared_ptr<TItemType> {
+                return std::make_shared<TItemType>((Arguments)...);
```
matekelemen:
```suggestion
                return std::make_shared<TItemType>(Arguments...);
```

ah yes, C(lojure)++, my favorite language :D

kratos/includes/registry_item.h
```diff
@@ -144,28 +145,45 @@ class KRATOS_API(KRATOS_CORE) RegistryItem
...
 
     /// Constructor with the name and shared ptr
     template<class TItemType>
     RegistryItem(
         const std::string&  rName,
         const shared_ptr<TItemType>& pValue)
         : mName(rName),
-          mpValue(pValue),
-          mGetValueStringMethod(&RegistryItem::GetItemString<TItemType>) {}
+          mpValue((std::function<std::shared_ptr<TItemType>()>)[=](){return pValue;})
```
matekelemen:
```suggestion
          mpValue((std::function<std::shared_ptr<TItemType>()>)[pValue](){return pValue;})
```

kratos/includes/registry_item.h
```diff
@@ -144,28 +145,45 @@ class KRATOS_API(KRATOS_CORE) RegistryItem
...
 
     /// Constructor with the name and value
     template<class TItemType>
     RegistryItem(
         const std::string&  rName,
         const TItemType& rValue)
         : mName(rName),
-          mpValue(Kratos::make_shared<TItemType>(rValue)),
-          mGetValueStringMethod(&RegistryItem::GetItemString<TItemType>) {}
+          mpValue((std::function<std::shared_ptr<TItemType>()>)[=](){return Kratos::make_shared<TItemType>(rValue);})
```
matekelemen:
```suggestion
          mpValue((std::function<std::shared_ptr<TItemType>()>)[rValue](){return Kratos::make_shared<TItemType>(rValue);})
```

kratos/includes/define_registry.h
```diff
@@ -0,0 +1,54 @@
...
+ * of the lambda function to something that is not an std::function.
+ */
+#define KRATOS_REGISTRY_ADD_PROTOTYPE(NAME, X, ...)                                       \
+    static inline bool KRATOS_REGISTRY_NAME(_is_registered_, __LINE__) = []() -> bool {   \
+        using TFunctionType = std::function<std::shared_ptr<X>()>;                        \
+        std::string key_name = NAME + std::string(".") + std::string(#X);                 \
+        if (!Registry::HasItem(key_name))                                                 \
+        {                                                                                 \
+            auto &r_item = Registry::AddItem<RegistryItem>(key_name);                     \
+            TFunctionType dispatcher = [](){return std::make_shared<X>(__VA_ARGS__);};    \
```
matekelemen: I'm a bit confused about what goes into `__VA_ARGS__` here. Are they always literals? If they aren't, they'll need to be included in the lambda's capture list, but then it won't work with literals.

kratos/includes/registry_item.h
```diff
@@ -181,23 +199,69 @@ class KRATOS_API(KRATOS_CORE) RegistryItem
...
+
+            KRATOS_ERROR_IF_NOT(insert_result.second)
+                << "Error in inserting '" << ItemName
+                << "' in registry item with name '" << this->Name() << "'." << std::endl;
+
+            return *insert_result.first->second;
+        }
+        else
+        {
+            std::function<std::shared_ptr<TItemType>()> aux_lambda = [=]() -> std::shared_ptr<TItemType> {
```
matekelemen:
```suggestion
            std::function<std::shared_ptr<TItemType>()> aux_lambda = [Arguments...]() -> std::shared_ptr<TItemType> {
```

kratos/tests/test_python_registry.py
```diff
@@ -171,31 +171,31 @@ def testKeys(self):
...
+    #                 if not KratosMultiphysics.Registry.HasValue(f"{item}.{subitem}"):
+    #                     for subsubitem in KratosMultiphysics.Registry[f"{item}.{subitem}"]:
+    #                         self.assertTrue(subsubitem in sub_sub_items_keys)
+    #                     if not KratosMultiphysics.Registry.HasValue(f"{item}.{subitem}.{subsubitem}"):
+    #                         for subsubsubitem in KratosMultiphysics.Registry[f"{item}.{subitem}.{subsubitem}"]:
+    #                             self.assertTrue(subsubsubitem in sub_sub_sub_items_keys)
+
+    #     # Remove the auxiliary testing entries from the Python registry
+    #     KratosMultiphysics.Registry.RemoveItem("Processes")
+    #     KratosMultiphysics.Registry.RemoveItem("PythonRootItem")
```
matekelemen, line 198: what's up with the python tests?

kratos/includes/registry_item.h
rubenzorrilla: At the very beginning I though the same... then we got used and we leave it like this (`mpCallable` is the thing to be always called once we've stored the prototype and `mpValue` is the  very first function that creates the prototype when first called).

I agree that it would be good to change it before merging it, but I'd leave this as the very last aesthetic change. Makes sense for you?

kratos/includes/registry_item.h
rubenzorrilla: For the record, I told @roigcarlo that I like to capture only what is used :smile:.

kratos/includes/define_registry.h
rubenzorrilla: You're right. Note that so far we tried with default `Process` and `Controller`  prototypes so `__VA_ARGS__` is empty. I agree this will crash with non-default prototypes.

kratos/tests/test_python_registry.py
rubenzorrilla: This was a shitty test I did to test the iteration of the registry (see my code comment, I was expecting this to fail as soon as we add more stuff to the registry). I think that we should modify this to only iterate, without checking the values, otherwise it would soon become a complete mess to maintain.

kratos/includes/define_registry.h
roigcarlo: No need to be literals, but they way this works (it is evaluated in the initialization of a static value) will force all the arguments to be static as well, otherwise it would throw a compiler error when constructing the class. If I am not wrong that means that there is no need to capture anything here.

kratos/includes/registry_item.h
rubenzorrilla: I've never used this `std::optional`. From the standard documentation I guess you mean using
```c++
if (!std::optional(&mpCallable)) {
    this->EvaluateValue<TDataType>();
}
```
Am I wrong?

kratos/includes/define_registry.h
rubenzorrilla: Considering @roigcarlo 's observation, which I agree, this means that effectively, we would never be able to register prototypes of non-default constructible classes (please correct me if I'm wrong). Note that in here I'm assuming no use-case of static arguments in Kratos. Having said this, can we directly remove ` __VA_ARGS__`?.

kratos/includes/registry_item.h
roigcarlo: Not sure how is the best way to apply it in this case. `EvaluateValue` do more things aside from assigning the return of the lambda (mainly initialize the value of the variable so the Registry can be serialized)

If we were to apply an `std::optional` here that would mean that we have to either:
1 - Split the function in `EvalVariable` and `EvalVariableValue` and only use the `optional`in the first half (because we don't want the code that assigns the name to be executed more than once)
2 - Make the `EvaluateVariable` return something like `std::optional<std::reference_wrapper<std::pair<TFunctionRetyrnType,std::string>>>` . Also I am not sure if this prevents the code of  `EvaluateVariable` from being executed or not.

Not an expert on the use of `std::optional`. If I am misusing this glad to see how its done :)

matekelemen: `std::optional` is a wrapper for objects that may or may not have a value.

For example, if you default construct a smart pointer, it's value will be in an invalid state (`nullptr`) until you replace it with something it can point to. Unfortunately, someone can still dereference the pointer while it's in an invalid state, leading to a segfault. `std::optional` stores a flag indicating whether the objects has a valid value, and will throw an exception if someone tries to access it in its invalid state. The point is that it effectively forces people to think about what happens if the object doesn't have a value, and provides a default behaviour (throw an exception) when they forget to do so.

Example
```cpp
struct S
{
    std::optional<std::unique_ptr<int>> mpWhatever;
};

int main()
{
    S s; // <== default constructor, mpWhatever has no value
    std::cout << *s.mpWhatever.value(); // <== exception, mpWhatever has no value

    s.mpWhatever = std::make_unique(1); // <== s.mpWhatever now has a valid value
    std::cout << *s.mpWhatever.value(); // <== 1
}
```

So, I suggest wrapping `mpCallable` with `std::optional` whenever it's assigned to.

```cpp
mpCallable = std::make_optional(func());
```

Then the check would become uglier, but we wouldn't run into segfaults if we forgot to check the pointer elsewhere.
```cpp
using TOptionalFunctionReturnType = std::optional<TFunctionReturnType>;
if (!std::any_cast<TOptionalFunctionReturnType>(&mpCallable).has_value()) {
    this->EvaluateValue<TDataType>();
}
```

kratos/includes/define_registry.h
matekelemen: @roigcarlo Alright, got it.
@rubenzorrilla No need to remove it, it could still be used with literals or static objects, though I think it's generally a bad idea to depend on statics too much.

kratos/includes/registry_item.h
roigcarlo: Oh!, ok :+1:

kratos/includes/define_registry.h
roigcarlo: No, what I am saying is:

Valid: (no params)
```C++

class foo {
   ...
   KRATOS_REGISTRY_ADD_PROTOTYPE(foo)
}
```

Valid: (liters)
```C++

class foo {
   ...
   KRATOS_REGISTRY_ADD_PROTOTYPE(foo, 1, 2, "potatoe")
}
```


Valid: (statics)
```C++

class boo {
    static Parameters BooParameters;
}

boo::BooParameters = Parameters::Initialize();

class foo {
   ...
   KRATOS_REGISTRY_ADD_PROTOTYPE(foo, Parameters::BooParameters)
}
```

Valid (maybe?)
```C++

class boo {
    boo() mBooParameters(Parameters())

    Parameters mBooParameters;

    Parameters& GetParameters(return mBooParameters);
}

class foo {
   ...
   KRATOS_REGISTRY_ADD_PROTOTYPE(foo, boo().GetParameters()) // The object gets constructed INSIDE the call at execution time.
}
```

Invalid
```C++

class boo {
    boo() mBooParameters(Parameters())

    Parameters mBooParameters;

    Parameters& GetParameters(return mBooParameters);
}

global gBoo = boo();

class foo {
   ...
   KRATOS_REGISTRY_ADD_PROTOTYPE(foo, gBoo.GetParameters()) // We have no clue about the life-cycle of gBoo, may or may not exists and we would need to capture the value.
}
```

roigcarlo: Btw, I think this is more interesting towards the discussion of being able to directly register the `Create`. For protos should suffice with the default constructor.

kratos/includes/registry_item.h
roigcarlo: Solved, I will push with the rest of the changes :+1:

kratos/includes/define_registry.h
rubenzorrilla: That's exactly my point, I think that for our use case (i.e. registering Kratos prototypes) we want to always register with the default constructor.

About the create, I think it would be the same since, as far as I understand, we will be registering a pointer to the function that does the create from a prototype.

Long story short, for the particular case of the prototypes I'd get rid of the `__VA_ARGS__`.

roigcarlo: Oook

roigcarlo:
@matekelemen Still working on the optional. For some reason I can't figure out the casts correctly....

sunethwarna (approved):
Finally :)

roigcarlo:
As a note, with @rubenzorrilla we have discovered that there is a bug solved in clang 12.0 (https://github.com/llvm/llvm-project/issues/44558) which prevents template classes which are instantiated from using the register mechanism. 

As a consequence we will have to set the minimum version of clang to 12.

PR #10916, opened 2023-03-24
[CORE] Adding Accessor and Properties extension (virtual version)

AlejandroCornejo:
**📝 Description**
In this parallel PR of the https://github.com/KratosMultiphysics/Kratos/pull/10900 in which I am adding:

A new `Accessor `class with a virtual method `GetProperty(...)` method.

Some operations to the `Properties` class:
- `typedef Accessor::Pointer AccessorPointerType;`
- A new `GetValue(...)` method that calls the `Accessor` to retrieve the variable
- A `AddAccessor(...)` to a certain variable

Use case:
~~~cpp
            Model current_model;
            auto &this_model_part = current_model.CreateModelPart("ModelPart",1);
            auto p_prop = this_model_part.CreateNewProperties(0);
            p_prop->SetValue(YOUNG_MODULUS, 2.1e11);

            const double initial_E = ((*p_prop)[YOUNG_MODULUS]);
            KRATOS_CHECK_LESS_EQUAL(2.1e11 - initial_E, 1.0e-8);

            // custom accessor that returns 2.0
            class CustomAccessor
                : public  Accessor 
            {
                public:
                double const& GetProperty(
                    const Variable<double> &rVariable,
                    const Properties &rProperties,
                    const GeometryType &rGeometry,
                    const Vector &rShapeFunctionVector,
                    const ProcessInfo &rProcessInfo) const override
                {
                    return mValue * rProperties[rVariable];
                }

                Accessor::Pointer Clone() const override
                {
                    return Kratos::make_shared<CustomAccessor>(*this);
                }

                private:
                    const double mValue = 2.0;
            };

            auto& r_process_info = this_model_part.GetProcessInfo();
            TestConstitutiveLaw r_clone_cl = TestConstitutiveLaw();
            p_prop->SetValue(CONSTITUTIVE_LAW, r_clone_cl.Clone());

            auto p_node_1 = this_model_part.CreateNewNode(1, 0.0 , 0.0 , 0.0);
            auto p_node_2 = this_model_part.CreateNewNode(2, 1.0 , 0.0 , 0.0);
            auto p_node_3 = this_model_part.CreateNewNode(3, 1.0 , 1.0 , 0.0);
            auto p_node_4 = this_model_part.CreateNewNode(4, 0.0 , 1.0 , 0.0);

            std::vector<NodeType::Pointer> geom(4);
            geom[0] = p_node_1;
            geom[1] = p_node_2;
            geom[2] = p_node_3;
            geom[3] = p_node_4;
            auto pgeom = Kratos::make_shared<Quadrilateral2D4<NodeType>>(PointerVector<NodeType>{geom});

            auto p_elem = Kratos::make_intrusive<TestElement>(0, pgeom, p_prop, TestElement::ResidualType::LINEAR);
            p_elem->Initialize(r_process_info);
            this_model_part.AddElement(p_elem);

            CustomAccessor custom_accessor = CustomAccessor();
            p_prop->SetAccessor(YOUNG_MODULUS, custom_accessor.Clone());

            Vector N;
            const double modified_E = p_prop->GetValue(YOUNG_MODULUS, *pgeom, N, r_process_info);
            KRATOS_CHECK_LESS_EQUAL(2.1e11 * 2.0 - modified_E, 1.0e-8);
~~~


loumalouomega:
@KratosMultiphysics/altair

kratos/includes/properties.h
```diff
@@ -25,7 +23,9 @@
 
 // Project includes
 #include "includes/define.h"
+#include "includes/accessor.h"
 #include "includes/node.h"
+#include "includes/variables.h"
```
philbucher: This is a super heavy include, please remove it, it is unused

philbucher (commented):
Minor

AlejandroCornejo:
ok! dear all, this is the first version of the `Accessor `class and its associated methods inside the `Properties`.

Please @KratosMultiphysics/technical-committee review it so we can move forward :)

AlejandroCornejo:
@KratosMultiphysics/implementation-committee

kratos/sources/accessor.cpp
```diff
@@ -0,0 +1,166 @@
...
+        const Variable<double> &rVariable,
+        const Properties &rProperties,
+        const GeometryType &rGeometry,
+        const Vector &rShapeFunctionVector,
+        const ProcessInfo &rProcessInfo
+        ) const
+    {
+        KRATOS_ERROR << "You are calling the GetProperty of the virtual Accessor class..." << std::endl;
+        double aux;
+        return aux;
```
matekelemen: 1) `aux` is constructed
2) a reference to the locally constructed `aux` is returned
3) => dangling reference

this is true for every default implementation in this class. I'd just recommend turning them into pure virtual functions instead.

kratos/tests/cpp_tests/includes/test_property_accessor.cpp
```diff
@@ -0,0 +1,98 @@
...
+            {
+                public:
+                double const& GetProperty(
+                    const Variable<double> &rVariable,
+                    const Properties &rProperties,
+                    const GeometryType &rGeometry,
+                    const Vector &rShapeFunctionVector,
+                    const ProcessInfo &rProcessInfo) const override
+                {
+                    return mValue * rProperties[rVariable];
```
matekelemen: same here: a dangling reference is returned, which is causing the CI to fail

kratos/includes/initial_state.h
```diff
@@ -43,7 +42,16 @@ class KRATOS_API(KRATOS_CORE) InitialState
     public:
         ///@name Type Definitions
         ///@{
-        typedef std::size_t SizeType;
+
+        // Size type definition
+        using SizeType = std::size_t;
```
AlejandroCornejo, line 47: why is better `using `now? :)

loumalouomega: iS more modern, its what modern people use

kratos/includes/initial_state.h
```diff
@@ -44,7 +42,16 @@ class KRATOS_API(KRATOS_CORE) InitialState
...
+
+        /// Pointer definition of NodeSearchUtility
+        KRATOS_CLASS_INTRUSIVE_POINTER_DEFINITION(InitialState);
```
matekelemen, line 50: First of all, why is this file changed in this PR? I don't see it getting used anywhere. Also, copypasta:

```suggestion
        /// Pointer definition of InitialState
        KRATOS_CLASS_INTRUSIVE_POINTER_DEFINITION(InitialState);
```

kratos/includes/properties.h
```diff
@@ -206,51 +212,71 @@ class Properties : public IndexedObject
...
-    typename TVariableType::Type const& GetValue(const TVariableType& rV, NodeType const& rThisNode, IndexType SolutionStepIndex) const
+    typename TVariableType::Type GetValue(const TVariableType& rVariable, const GeometryType &rGeometry, const Vector& rShapeFunctionVector, const ProcessInfo &rProcessInfo) const
     {
-        if(mData.Has(rV))
-            return mData.GetValue(rV);
+        auto it_value = mAccessors.find(rVariable.Key());
+        if (it_value != mAccessors.end()) {
+            return (it_value->second)->GetProperty(rVariable, *this, rGeometry, rShapeFunctionVector, rProcessInfo);
+        } else {
+            return mData.GetValue(rVariable);
```
loumalouomega, line 296: Maybe we should add a Warning if not accesor found

kratos/includes/properties.h
```diff
@@ -206,51 +212,71 @@ class Properties : public IndexedObject
...
+        return rThisNode.GetValue(rVariable, SolutionStepIndex);
     }
 
+    /*
+    Custom GetValue in which we check the Accessor
+    */
     template<class TVariableType>
-    typename TVariableType::Type const& GetValue(const TVariableType& rV, NodeType const& rThisNode, IndexType SolutionStepIndex) const
+    typename TVariableType::Type GetValue(const TVariableType& rVariable, const GeometryType &rGeometry, const Vector& rShapeFunctionVector, const ProcessInfo &rProcessInfo) const
```
loumalouomega:
```suggestion
    typename TVariableType::Type GetValue(const TVariableType& rVariable, const GeometryType& rGeometry, const Vector& rShapeFunctionVector, const ProcessInfo& rProcessInfo) const
```

kratos/includes/properties.h
AlejandroCornejo: well... This is a decision we have to make

loumalouomega: https://www.boost.org/doc/libs/1_81_0/libs/unordered/doc/html/unordered.html#benchmarks_boostunordered_flat_map I still think we should also discuss about using this instead

kratos/includes/accessor.h
```diff
@@ -0,0 +1,248 @@
...
+    /**
+     * @brief Custom method to retrieve double type properties
+     * @param rVariable The variable considered (double type properties)
+     * @param rProperties The properties considered
+     * @param rGeometry The geometry considered
+     * @param rShapeFunctionVector The shape function of the GP considered
+     * @param rProcessInfo The process info considered
+     * @return The double type properties
+     */
+    virtual double GetProperty(
```
philbucher: I would call this `GetValue` to be consistent with the other containers

otherwise the access would be:
`GetProperties().GetAccessor().GetProperty()` which I believe is quite confusing

(or maybe `FastGetValue` to increase the speed :eyes: :runner: @RiccardoRossi  xD)

kratos/includes/accessor.h
```diff
@@ -0,0 +1,248 @@
...
+
+    /// Pointer definition of NodeSearchUtility
+    KRATOS_CLASS_POINTER_DEFINITION(Accessor);
+
+    ///@}
+    ///@name Life Cycle
+    ///@{
+
+    /// Default constructor.
+    Accessor(){};
```
philbucher:
```suggestion
```

You can leave this to the compiler :)

I would however probably delete the Copy-Ctor and the Assignment operator

kratos/includes/accessor.h
```diff
@@ -0,0 +1,248 @@
...
+
+    /// Destructor.
+    virtual ~Accessor() {}
```
philbucher:
```suggestion
    virtual ~Accessor() = default;
```

kratos/includes/properties.h
```diff
@@ -206,51 +212,71 @@ class Properties : public IndexedObject
...
+    Custom GetValue in which we check the Accessor
+    */
     template<class TVariableType>
-    typename TVariableType::Type const& GetValue(const TVariableType& rV, NodeType const& rThisNode, IndexType SolutionStepIndex) const
+    typename TVariableType::Type GetValue(const TVariableType& rVariable, const GeometryType& rGeometry, const Vector& rShapeFunctionVector, const ProcessInfo& rProcessInfo) const
     {
-        if(mData.Has(rV))
-            return mData.GetValue(rV);
+        auto it_value = mAccessors.find(rVariable.Key());
+        if (it_value != mAccessors.end()) {
```
philbucher, line 293: I think a more efficient version would be to have a "Dummy" accessor without implementation, which basically forwards the call to the `mData`

Otherwise you have to pay the cost of the virtual access AND the search in the map

philbucher (commented):
some comments otherwise cool!

kratos/includes/properties.h
```diff
@@ -206,51 +212,71 @@ class Properties : public IndexedObject
...
+    typename TVariableType::Type const& GetValue(const TVariableType& rVariable) const
     {
-        return mData.GetValue(rV);
+
+        return mData.GetValue(rVariable);
     }
 
     template<class TVariableType>
-    typename TVariableType::Type& GetValue(const TVariableType& rV, NodeType& rThisNode)
+    typename TVariableType::Type& GetValue(const TVariableType& rVariable, NodeType& rThisNode)
```
philbucher, line 255: BTW it would be great if you could merge the cosmetic changes in a separate PR, this one gets already confusing

kratos/includes/initial_state.h
ddiezrod: I think its mainly because of this https://www.nextptr.com/tutorial/ta1193988140/how-cplusplus-using-or-aliasdeclaration-is-better-than-typedef

kratos/includes/properties.h
AlejandroCornejo: ok for me, but then we always have to make sure that this default creation is done when reading the material right?

kratos/includes/accessor.h
AlejandroCornejo: I agree with that, I put that name without thinking much hehe

kratos/includes/properties.h
philbucher: I think you are right, this is actually required

sorry for the noise, I confused myself

ddiezrod:
We were discussing this in @KratosMultiphysics/implementation-committee . We think current implementation is ok (only minor changes wil be needed) , although we are a little hesitant that this will be really used in the future. Just out of curiosity I am going to do a small profiling to check how fast this is. I will post the results here, you can carry on with the merge without waiting for this.

AlejandroCornejo:
ping

loumalouomega:
> ping

Do I approve or we wait for @KratosMultiphysics/technical-committee ?

kratos/includes/accessor.h
```diff
@@ -0,0 +1,245 @@
...
+    /**
+     * @brief Custom method to retrieve array_1d<double, 6 > type properties
+     * @param rVariable The variable considered (array_1d<double, 6 > type properties)
+     * @param rProperties The properties considered
+     * @param rGeometry The geometry considered
+     * @param rShapeFunctionVector The shape function of the GP considered
+     * @param rProcessInfo The process info considered
+     * @return The array_1d<double, 6 > type properties
+     */
+    virtual array_1d<double, 6> GetValue(
```
sunethwarna, line 183: Just wondering, we have `array_1d<double, 3>`, `array_1d<double, 4>`, `array_1d<double, 6>`, `array_1d<double, 9>` interfaces in core. Is this intentional to limit it to these types only?

kratos/includes/properties.h
```diff
@@ -506,6 +532,12 @@ class Properties : public IndexedObject
                 r_subprop.PrintData(rOStream);
             }
         }
+        if (mAccessors.size() > 0) {
+            rOStream << "\nThis properties contains the following " << mAccessors.size() << " accessors" << std::endl;
+            for (auto entry : mAccessors) {
```
sunethwarna:
```suggestion
            for (auto& entry : mAccessors) {
```

kratos/sources/accessor.cpp
```diff
@@ -0,0 +1,154 @@
...
+    const Variable<Vector>& rVariable,
+    const Properties& rProperties,
+    const GeometryType& rGeometry,
+    const Vector& rShapeFunctionVector,
+    const ProcessInfo& rProcessInfo
+    ) const
+{
+    KRATOS_ERROR << "You are calling the GetValue of the virtual Accessor class..." << std::endl;
+    Vector aux(0);
+    return aux;
```
sunethwarna: This should also work since it is returning by copy, ryt?
```suggestion
    return Vector();
```

kratos/sources/accessor.cpp
```diff
@@ -0,0 +1,154 @@
...
+    const Variable<Matrix>& rVariable,
+    const Properties& rProperties,
+    const GeometryType& rGeometry,
+    const Vector& rShapeFunctionVector,
+    const ProcessInfo& rProcessInfo
+    ) const
+{
+    KRATOS_ERROR << "You are calling the GetValue of the virtual Accessor class..." << std::endl;
+    Matrix mat(0, 0);
+    return mat;
```
sunethwarna:
```suggestion
    return Matrix(0,0);
```

kratos/sources/accessor.cpp
```diff
@@ -0,0 +1,154 @@
...
+    const Variable<array_1d<double, 3>>& rVariable,
+    const Properties& rProperties,
+    const GeometryType& rGeometry,
+    const Vector& rShapeFunctionVector,
+    const ProcessInfo& rProcessInfo
+    ) const
+{
+    KRATOS_ERROR << "You are calling the GetValue of the virtual Accessor class..." << std::endl;
+    array_1d<double, 3> mat;
+    return mat;
```
sunethwarna:
```suggestion
    return array_1d<double, 3>();
```

kratos/sources/accessor.cpp
```diff
@@ -0,0 +1,154 @@
...
+    const Variable<array_1d<double, 6>>& rVariable,
+    const Properties& rProperties,
+    const GeometryType& rGeometry,
+    const Vector& rShapeFunctionVector,
+    const ProcessInfo& rProcessInfo
+    ) const
+{
+    KRATOS_ERROR << "You are calling the GetValue of the virtual Accessor class..." << std::endl;
+    array_1d<double, 6> mat;
+    return mat;
```
sunethwarna:
```suggestion
    return array_1d<double, 6>();
```

kratos/tests/cpp_tests/includes/test_property_accessor.cpp
```diff
@@ -0,0 +1,95 @@
...
+*/
+KRATOS_TEST_CASE_IN_SUITE(PropertyAccessorSimpleProperties, KratosCoreFastSuite)
+{
+    Model current_model;
+    auto &this_model_part = current_model.CreateModelPart("ModelPart",1);
+    auto p_prop = this_model_part.CreateNewProperties(0);
+    p_prop->SetValue(YOUNG_MODULUS, 2.1e11);
+
+    const double initial_E = ((*p_prop)[YOUNG_MODULUS]);
+    KRATOS_CHECK_LESS_EQUAL(2.1e11 - initial_E, 1.0e-8);
```
sunethwarna: Why `KRATOS_CHECK_LESS_EQUAL` and not `KRATOS_CHECK_NEAR`?
```suggestion
    KRATOS_CHECK_NEAR(2.1e11, initial_E, 1.0e-8);
```

kratos/tests/cpp_tests/includes/test_property_accessor.cpp
```diff
@@ -0,0 +1,95 @@
...
+    auto p_elem = Kratos::make_intrusive<TestElement>(0, pgeom, p_prop, TestElement::ResidualType::LINEAR);
+    p_elem->Initialize(r_process_info);
+    this_model_part.AddElement(p_elem);
+
+    CustomAccessor custom_accessor = CustomAccessor();
+    p_prop->SetAccessor(YOUNG_MODULUS, custom_accessor.Clone());
+
+    Vector N;
+    const double modified_E = p_prop->GetValue(YOUNG_MODULUS, *pgeom, N, r_process_info);
+    KRATOS_CHECK_LESS_EQUAL(2.1e11 * 2.0 - modified_E, 1.0e-8);
```
sunethwarna: Why `KRATOS_CHECK_LESS_EQUAL` and not `KRATOS_CHECK_NEAR`?
```suggestion
    KRATOS_CHECK_NEAR(2.1e11 * 2.0, modified_E, 1.0e-8);
```

kratos/tests/cpp_tests/includes/test_property_accessor.cpp
```diff
@@ -0,0 +1,95 @@
...
+            return Kratos::make_shared<CustomAccessor>(*this);
+        }
+
+        private:
+            const double mValue = 2.0;
+    };
+
+    auto& r_process_info = this_model_part.GetProcessInfo();
+    TestConstitutiveLaw r_clone_cl = TestConstitutiveLaw();
+    p_prop->SetValue(CONSTITUTIVE_LAW, r_clone_cl.Clone());
```
sunethwarna: Do we have to set the `ConstitutiveLaw` in here? Wouldn't the test be better without it?

sunethwarna (commented):
Thanks @AlejandroCornejo. Minor comments only.

One final observation:
Since we are passing the `rShapeFunctionVector`, that means we are allowing this to be called within the gauss point loops which are expensive. That means, part of the element gauss point formulation will live within the `Accessor`. Are we allowing that?

kratos/includes/accessor.h
AlejandroCornejo: I will add them! :)

sunethwarna:
@ddiezrod 

> We were discussing this in https://github.com/orgs/KratosMultiphysics/teams/implementation-committee . We think current implementation is ok (only minor changes wil be needed) , although we are a little hesitant that this will be really used in the future. Just out of curiosity I am going to do a small profiling to check how fast this is. I will post the results here, you can carry on with the merge without waiting for this.

I am curious. Do you happen to know any other way for this? :)

ddiezrod:
> @ddiezrod
> 
> > We were discussing this in https://github.com/orgs/KratosMultiphysics/teams/implementation-committee . We think current implementation is ok (only minor changes wil be needed) , although we are a little hesitant that this will be really used in the future. Just out of curiosity I am going to do a small profiling to check how fast this is. I will post the results here, you can carry on with the merge without waiting for this.
> 
> I am curious. Do you happen to know any other way for this? :)

Not really. In altair thermomechanical elements, when we want to retrieve e.g. YOUNG_MODULUS we check if table exists and if not, we check if the value exists in data value container. Of course this is not general but it is pretty much all you need. I understand that this is not general enough, but I wanted to check in terms of speed which one is slower.

AlejandroCornejo:
@sunethwarna comments addressed

kratos/sources/accessor.cpp
```diff
@@ -0,0 +1,180 @@
...
+    KRATOS_ERROR << "You are calling the GetValue of the virtual Accessor class..." << std::endl;
+    return "string";
+}
+
+/***********************************************************************************/
+/***********************************************************************************/
+
+Accessor::Pointer Accessor::Clone() const
+{
+    return Kratos::make_shared<Accessor>(*this);
```
sunethwarna:
```suggestion
    KRATOS_ERROR << "You are calling the Clone of the virtual Accessor class..." << std::endl;
    return Kratos::make_shared<Accessor>(*this);
```

kratos/includes/accessor.h
sunethwarna: Thanks :)

kratos/sources/accessor.cpp
sunethwarna: There is no point in cloning the base class ryt?

AlejandroCornejo:
@KratosMultiphysics/technical-committee  ping

kratos/sources/accessor.cpp
rubenzorrilla: Agree.

kratos/includes/accessor.h
```diff
@@ -0,0 +1,279 @@
...
+    virtual std::string GetValue(
+        const Variable<std::string>& rVariable,
+        const Properties& rProperties,
+        const GeometryType& rGeometry,
+        const Vector& rShapeFunctionVector,
+        const ProcessInfo& rProcessInfo
+        ) const;
+
+    // Getting a pointer to the class
+    virtual Accessor::Pointer Clone() const;
```
rubenzorrilla: Shouldn't we add a `Create` method as well?

rubenzorrilla: We've been discussing about this in the @KratosMultiphysics/technical-committee. We'd leave this for the future, as it is not clear at all that we'll be able to standarize the accessors to be factory-based.

AlejandroCornejo:
@rubenzorrilla @roigcarlo I tried to make it unique pointer but there are some compilation errors that IDK how to solve...

loumalouomega:
> @rubenzorrilla @roigcarlo I tried to make it unique pointer but there are some compilation errors that IDK how to solve...

It is related with serializer..., can you try to comment the serializer part and see if compiles?

rubenzorrilla:
> > @rubenzorrilla @roigcarlo I tried to make it unique pointer but there are some compilation errors that IDK how to solve...
> 
> It is related with serializer..., can you try to comment the serializer part and see if compiles?

Actually there is the copy constructor of the properties involved as well. My suggestion is to also copy (i.e. clone) the accessors. See my last commit, it should fix this. However, the serialization is not ready yet...

kratos/includes/accessor.h
rubenzorrilla: Just pointing that now we need it after the switch to a `unique_ptr`.

rubenzorrilla:
BTW, shouldn't we export this to Python?

AlejandroCornejo:
> BTW, shouldn't we export this to Python?

Yes, indeed. Can we do it in a following PR?

rubenzorrilla:
> > BTW, shouldn't we export this to Python?
> 
> Yes, indeed. Can we do it in a following PR?

I'd do so. We've already discussed a lot about this one.

AlejandroCornejo:
Then someone please approve! :)

sunethwarna (approved):
Finally :D. :100:

kratos/includes/properties.h
```diff
@@ -588,6 +649,13 @@ class Properties : public IndexedObject
         rSerializer.save("Data", mData);
         rSerializer.save("Tables", mTables);
         rSerializer.save("SubPropertiesList", mSubPropertiesList);
+        std::vector<std::pair<const KeyType, Accessor*>> aux_accessors_container;
```
RiccardoRossi, line 652: are u sure that this saving and loading mechanism works? is it tested?

PR #10919, opened 2023-03-24
[Fluid][FSI] Embedded velocity from RBF interpolation

rubenzorrilla:
This PR makes possible to calculate the `EMBEDDED_VELOCITY` from an RBF interpolation of the structure `VELOCITY` values. The RBF search radius can be user-defined or set in terms of the problem bounding box size. Current behavior is kept by default.

Note that I also took the chance to remove some legacy code and warnings.


applications/FluidDynamicsApplication/python_scripts/navier_stokes_embedded_solver.py
```diff
@@ -285,6 +289,9 @@ def GetDefaultParameters(cls):
             "fm_ale_settings": {
                 "fm_ale_step_frequency": 0,
                 "mesh_movement": "implicit",
+                "embedded_velocity_calculation": "from_fluid_mesh_velocity",
+                "rbf_interpolation_search_radius": 0.0,
```
RiccardoRossi, line 293: you could make this a multiplier to the max edge length, this way It could be almost fully automatic

rubenzorrilla: Better this than the bounding box :+1: . Thanks for the suggestion.

rubenzorrilla: Done.

rubenzorrilla:
@RiccardoRossi just pointing that I added a `FindMaximumEdgeLength` function to the `FluidAuxiliaryUtilities`. We can consider  moving it to the core if required for the upcoming edge-based stuff.

RiccardoRossi (approved):

PR #2567, opened 2018-07-25
[Core] Constraint builder solver

adityaghantasala:
This PR adds : 
- The builder and solver to use the master slave constraints. 
- The auxiliary data structure used only in the builder and solver. 
- Test for the linear master-slave constraints in the core (depends on StructuralMechanicsApplication)

PR #10926, opened 2023-03-27
[StructuralMechanicsApplication] Clarify MPC B&S creation

loumalouomega:
**📝 Description**

@pablobecker found this https://github.com/KratosMultiphysics/Kratos/blob/ed73cf379f8523e1581092381f89394f28615844/applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_solver.py#L339, and he asked me why doing this in a so inefficient manner. I did a research and a git blame and I found this was merged 5 years ago here: https://github.com/KratosMultiphysics/Kratos/pull/2567

I was explicitly asked why doing so at that time, and apparently I replied with a meme and not answering the question:

image

The thing is that this was done in order to avoid potential issues when using the elimination builder and solver, that in the case of using MPC is an independent B&S (not the same one as in the case of block builder and solver). This is because in the elimination the graph of the matrix is very dependent of the DoFs and its status. 

@philbucher and @RiccardoRossi sorry for not clarifying that at the time

**🆕 Changelog**

- [Clarify MPC B&S creation](https://github.com/KratosMultiphysics/Kratos/commit/2882f9856ea5cfc1f7d1b4628889ac5503410765)

### DIFF
--- a/applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_solver.py
+++ b/applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_solver.py
@@ -337,8 +337,7 @@ class MechanicalSolver(PythonSolver):
         if not hasattr(self, '_builder_and_solver'):
             self._builder_and_solver = self._CreateBuilderAndSolver()
         elif not self.use_block_builder: # Block builder and solver are unified with MPC and without. In the case of the elimination this could be a problem
-            if self.settings["multi_point_constraints_used"].GetBool() is False and
-                self.GetComputingModelPart().NumberOfMasterSlaveConstraints() > 0 and not self.mpc_block_builder_initialized:
+            if self.settings["multi_point_constraints_used"].GetBool() is False and self.GetComputingModelPart().NumberOfMasterSlaveConstraints() > 0 and not self.mpc_block_builder_initialized:
                 self.settings["multi_point_constraints_used"].SetBool(True)
                 self._builder_and_solver = self._CreateBuilderAndSolver()
                 self.mpc_block_builder_initialized = True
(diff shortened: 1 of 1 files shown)


applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_solver.py
```diff
@@ -327,20 +334,21 @@ def _GetLinearSolver(self):
...
     def _GetBuilderAndSolver(self):
-        if (self.settings["multi_point_constraints_used"].GetBool() is False and
-            self.GetComputingModelPart().NumberOfMasterSlaveConstraints() > 0):
-            self.settings["multi_point_constraints_used"].SetBool(True)
-            self._builder_and_solver = self._CreateBuilderAndSolver()
         if not hasattr(self, '_builder_and_solver'):
             self._builder_and_solver = self._CreateBuilderAndSolver()
+        elif not self.use_block_builder: # Block builder and solver are unified with MPC and without. In the case of the elimination this could be a problem
+            if self.settings["multi_point_constraints_used"].GetBool() is False and self.GetComputingModelPart().NumberOfMasterSlaveConstraints() > 0 and not self.mpc_block_builder_initialized:
+                self.settings["multi_point_constraints_used"].SetBool(True)
```
philbucher, line 341: we should probably remove this check and rely on `self.GetComputingModelPart().NumberOfMasterSlaveConstraints() > 0` (global ofc bcs well, MPI)

loumalouomega: Maybe

philbucher (approved):
I dont understand this code but ok

PR #10927, opened 2023-03-27
Prevent double deprecated warning

roigcarlo:
**📝 Description**
This is starting to drive me insane. Since nobody is getting ride of the `Dimension()` warning and we cannot remove the function I added a pragma that prevents the warning from triggering from every file that includes `geometry_data.h` (quite a few)

The problem is that `GeometryData::Dimension()` (deprecated) calls `GeometryDimension::Dimension()` (also deprecated). When compiling the body of the first, we always calling a deprecated code.

### DIFF
--- a/kratos/geometries/geometry_data.h
+++ b/kratos/geometries/geometry_data.h
@@ -374,7 +374,10 @@ public:
     KRATOS_DEPRECATED_MESSAGE("'Dimension' is deprecated. Use either 'WorkingSpaceDimension' or 'LocalSpaceDimension' instead.")
     SizeType Dimension() const
     {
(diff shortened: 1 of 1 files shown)


roigcarlo:
> Rather than doing this, something that will effectively turn into keeping that legacy code in there forever I'd remove the deprecated stuff...

Been asking for that for months and nobody does it. I would myself but I don't understand half of the elements formulation.

Btw. This does not suppress the warning, it only removes the ones caused because we include the double deprecation even if it has nothing to do with geometries (false positives)

roigcarlo:
Well seems that compiler does not like that

PR #10928, opened 2023-03-27
[Core] ContainerExpression

sunethwarna:
**📝 Description**
This adds a class named `ContainerExpression`. It is adding following capabilities.

1. This container enables retrieving/assigning data from/to each `historical nodes`, `non-historical nodes`, `conditions` and`elements` data container easily.
2. Data transfer between `nodal historical <-> nodal non-historical` and between potential other type (implemented in #10768)
3. All of these introduced contaienrs are `OpenMP` and `MPI` parallelized. Which makes the use of them in python super simple.
4. They have most of the operators defined.
5. These containers can be used to store and post process data using vtk/hdf5/StatApp easily. It is introduced in a later PR.
6. These uses `lazy-expressions` hence, all the arithmatic operations performed on the containers will only create expressions. Only `Evaluate` method will evaluate the final combined expression.
7. This includes a `c-interface` for data transfer between `ModelPart` container variable data and other compatible software as well. (In the case of `CoSim` or other libraries.)
8. It has numpy import and export as well. So now we can export variable data in any container to numpy, do calculations there and transform back the numpy arrays to variables in our model parts.
9. It has `MoveFrom` method which moves the data from another container. This is supported for numpy as well. Hence, no copies of data will be kept within the container.
10. These containers can be easily used to transfer variable data back and forward between third party libraries with ease. (numpy is an example).

It supports any of the following variable types:
-  `double`
-  `array_1d<double, 3>`
-  `array_1d<double, 4>`
-  `array_1d<double, 6>`
-  `array_1d<double, 9>`
-  `Vector`
-  `Matrix`

**Todo**
- [x]  Add test for c-interface
- [x]  Couple it with numpy arrays efficiently
- [x] Fix failing CI issues which are due to CPP version mismatching.

**🆕 Changelog**
- Adds `ContainerVariableData`
- Adds tests


kratos/python/add_container_variable_data_to_python.cpp
```diff
@@ -0,0 +1,243 @@
...
+// Include base h
+#include "add_container_variable_data_to_python.h"
+
+namespace Kratos::Python
+{
+
+template <class TDataType>
+using ContiguousNumpyArray = pybind11::array_t<
+    TDataType,
+    /*column-major*/ pybind11::array::c_style | /*cast instead of looking for a suitable overload*/ pybind11::array::forcecast
```
matekelemen: this is a leftover, just remove the forcecast tag because we don't actually want that.

kratos/python/add_container_variable_data_to_python.cpp
```diff
@@ -0,0 +1,243 @@
...
+    namespace py = pybind11;
+
+    using container_variable_data_holder_base = ContainerVariableData<TContainerType>;
+    py::class_<container_variable_data_holder_base, typename container_variable_data_holder_base::Pointer>(m, rName.c_str())
+        .def("CopyFrom", &container_variable_data_holder_base::CopyFrom, py::arg("origin_container_data"))
+        .def("MoveFrom", [](container_variable_data_holder_base& rSelf, py::array_t<double>& rData){
+            KRATOS_ERROR_IF(rData.ndim() == 0) << "Passed data is not compatible.\n";
+
+            std::vector<int> shape(rData.ndim() - 1);
+            std::copy(rData.shape() + 1, rData.shape() + rData.ndim(), shape.begin());
```
matekelemen: you should probably explain the difference between the shape of the numpy array, and the shape of the internal representation somewhere. It's not immediately clear here why we're omitting the first component.

kratos/containers/container_variable_data/container_variable_data.h
```diff
@@ -0,0 +1,327 @@
...
+ *      1. Nodal historical and non-historical.
+ *      2. Element and element properties.
+ *      3. Condition and condition properties.
+ *
+ * This class is optimized and compatible with OpenMP and MPI.
+ *
+ * @tparam TContainerType
+ */
+template <class TContainerType>
+class KRATOS_API(OPTIMIZATION_APPLICATION) ContainerVariableData {
```
matekelemen:
```suggestion
class KRATOS_API(KRATOS_CORE) ContainerVariableData {
```

kratos/containers/container_variable_data/container_variable_data.cpp
```diff
@@ -0,0 +1,279 @@
...
+// Include base h
+#include "container_variable_data.h"
+
+namespace Kratos {
+
+template <class TContainerType>
+ContainerVariableData<TContainerType>::ContainerVariableData(ModelPart& rModelPart)
+    : mpModelPart(&rModelPart)
+{
+}
```
matekelemen: Wrong initialization order!

```suggestion
template <class TContainerType>
ContainerVariableData<TContainerType>::ContainerVariableData(ModelPart& rModelPart)
    : mpExpression(),
      mpModelPart(&rModelPart)
{
}
```

kratos/containers/container_variable_data/container_variable_data.cpp
```diff
@@ -0,0 +1,279 @@
...
+    KRATOS_ERROR_IF_NOT(number_of_entities == this->GetContainer().size())
+        << "Number of entities does not match with the local container size. [ "
+           "NumberOfEntities = "
+        << NumberOfEntities
+        << ", local container size = " << this->GetContainer().size() << " ].\n";
+
+    std::vector<IndexType> shape(ShapeSize);
+    for (int i = 0; i < ShapeSize; ++i) {
+        shape[i] = *(pShapeBegin++);
+    }
```
matekelemen:
```suggestion
    std::copy(pShapeBegin,
              pShapeBegin + ShapeSize,
              shape.begin());
```

kratos/containers/container_variable_data/container_variable_data.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+    auto p_expression = LiteralFlatExpression::Create(number_of_entities, shape);
+    this->mpExpression = p_expression;
+
+    const IndexType local_size = this->GetExpression().GetFlattenedSize();
```
matekelemen:
```suggestion
    const IndexType flattened_size = this->GetExpression().GetFlattenedSize();
```

kratos/containers/container_variable_data/container_variable_data.cpp
```diff
@@ -0,0 +1,279 @@
...
+    KRATOS_ERROR_IF_NOT(number_of_entities == this->GetContainer().size())
+        << "Number of entities does not match with the local container size. [ "
+           "NumberOfEntities = "
+        << NumberOfEntities
+        << ", local container size = " << this->GetContainer().size() << " ].\n";
+
+    std::vector<IndexType> shape(ShapeSize);
+    for (int i = 0; i < ShapeSize; ++i) {
+        shape[i] = *(pShapeBegin++);
+    }
```
matekelemen: `std::copy`

kratos/containers/container_variable_data/container_variable_data.cpp
```diff
@@ -0,0 +1,279 @@
...
+        << ", available shape = " << r_expression.GetShape() << " ].\n";
+
+    const IndexType local_size = r_expression.GetFlattenedSize();
+
+    IndexPartition<IndexType>(number_of_entities).for_each([pBegin, local_size, &r_expression](const IndexType EntityIndex) {
+        const IndexType entity_data_begin_index = EntityIndex * local_size;
+        double* p_input_data_begin = pBegin + entity_data_begin_index;
+        for (IndexType i = 0; i < local_size; ++i) {
+            *(p_input_data_begin+i) = r_expression.Evaluate(EntityIndex, entity_data_begin_index, i);
+        }
```
matekelemen: `std::transform`?

kratos/containers/container_variable_data/container_variable_data.cpp
```diff
@@ -0,0 +1,279 @@
...
+template <class TContainerType>
+void ContainerVariableData<TContainerType>::SetExpression(Expression::Pointer pExpression)
+{
+    this->mpExpression = pExpression;
+}
+
+template <class TContainerType>
+const Expression& ContainerVariableData<TContainerType>::GetExpression() const
+{
+    return *(*mpExpression);
```
matekelemen: I'm not sure tho
```suggestion
    return **mpExpression;
```

kratos/containers/container_variable_data/expressions/expression.h
```diff
@@ -0,0 +1,113 @@
...
+
+namespace Kratos {
+
+///@name Kratos Classes
+///@{
+
+/**
+ * @brief Base class or all the expression types.
+ *
+ * This is used to represent an expression for arithmatic evaluation
```
matekelemen:
```suggestion
 * This is used to represent an expression for arithmetic evaluation
```

kratos/containers/container_variable_data/expressions/expression.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Kratos {
+
+std::size_t Expression::GetFlattenedSize() const
+{
+    IndexType result = 1;
+    for (const auto v : this->GetShape()) {
+        result *= v;
+    }
+    return result;
```
matekelemen:
```suggestion
    const auto& r_shape = this->GetShape();
    return std::accumulate(r_shape.begin(),
                           r_shape.end(),
                           1UL,
                           [] (auto left, auto right) {return left * right;});
```

you'll need to `#include <numeric>`

kratos/containers/container_variable_data/expressions/literal/literal_flat_expression.h
```diff
@@ -0,0 +1,179 @@
...
+        const std::vector<IndexType>& rShape);
+
+    LiteralFlatExpression(
+        double* pDataBegin,
+        const IndexType NumberOfEntities,
+        const std::vector<IndexType>& rShape);
+
+    LiteralFlatExpression(const LiteralFlatExpression& rOther) = delete;
+
+    ~LiteralFlatExpression() override = default;
```
matekelemen: unnecessary

kratos/containers/container_variable_data/expressions/literal/literal_flat_expression.h
```diff
@@ -0,0 +1,179 @@
...
+
+    /**
+     * @brief A specific create method is used in here to return a pointer to the LiteralFlatExpression
+     *
+     * This method is used to return the derrived class LiteralFlatExpression pointer
+     * so that additional SetData method can be accessed to initialize
+     * the vector data.
+     *
+     * @param NumberOfEntities                  Number of entities in the container.
+     * @param rShape                            Shape of the data in each entitiy.
```
matekelemen:
```suggestion
     * @param rShape                            Shape of the data in each entity.
```

kratos/containers/container_variable_data/expressions/literal/literal_flat_expression.h
```diff
@@ -0,0 +1,179 @@
...
+    ///@{
+
+    /**
+     * @brief A specific create method is used in here to return a pointer to the LiteralFlatExpression
+     *
+     * This method is used to return the derrived class LiteralFlatExpression pointer
+     * so that additional SetData method can be accessed to initialize
+     * the vector data.
+     *
+     * @param NumberOfEntities                  Number of entities in the container.
```
matekelemen:
```suggestion
     * @param NumberOfEntities                  Number of entities (eg. nodes, elements, etc.) in the container.
```

kratos/containers/container_variable_data/expressions/literal/literal_flat_expression.h
```diff
@@ -0,0 +1,179 @@
...
+    const std::vector<IndexType> GetShape() const override;
+
+    std::string Info() const override;
+
+    ///@}
+protected:
+    ///@name Protected classes
+    ///@{
+
+    class Data
```
matekelemen: you should probably add a short note on why this is necessary (it has to either represent an array whose lifetime it manages, or one whose lifetime is externally managed)

kratos/containers/container_variable_data/expressions/literal/literal_flat_expression.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+const std::vector<std::size_t> LiteralFlatExpression::GetShape() const
+{
+    return mShape;
+}
+
+std::string LiteralFlatExpression::Info() const
+{
+    std::stringstream msg;
+    msg << "V" << mShape;
```
matekelemen: [V](https://www.youtube.com/watch?v=wKn1R6fekk4)?

kratos/containers/container_variable_data/expressions/literal/literal_expression.h
```diff
@@ -0,0 +1,70 @@
...
+class LiteralExpression : public Expression {
+public:
+    ///@name Life cycle
+    ///@{
+
+    LiteralExpression(const TDataType& Value);
+
+    LiteralExpression(const LiteralExpression& rOther) = delete;
+
+    ~LiteralExpression() override = default;
```
matekelemen: unnecessary

kratos/containers/container_variable_data/expressions/binary/binary_expression.h
```diff
@@ -0,0 +1,87 @@
...
+///@{
+
+namespace BinaryOperations
+{
+    struct Addition       { static inline constexpr double Evaluate(const double V1, const double V2) { return V1 + V2; } };
+    struct Substraction   { static inline constexpr double Evaluate(const double V1, const double V2) { return V1 - V2; } };
+    struct Multiplication { static inline constexpr double Evaluate(const double V1, const double V2) { return V1 * V2; } };
+    struct Division       { static inline constexpr double Evaluate(const double V1, const double V2) { return V1 / V2; } };
+    struct Power          { static inline constexpr double Evaluate(const double V1, const double V2) { return std::pow(V1, V2); } };
+}
```
matekelemen: I'd put these structs into the `BinaryExpression` class def instead of the `Kratos::BinaryOperations` namespace. It'd be a bit more clear what they're supposed to be used for when someone is browsing this header but doesn't know anything about the system.

kratos/containers/container_variable_data/specialized_container_variable_data.h
```diff
@@ -0,0 +1,230 @@
...
+ * between compatible @ref SpecializedContainerVariableData containers. This copy is also light weight since
+ * this only copies the pointers, not the data itself.
+ *
+ * This class can take advantage of OpenMP and MPI.
+ *
+ * @tparam TContainerType           Container type
+ * @tparam TContainerDataIO         Container entity input/output type
+ */
+template <class TContainerType, class TContainerDataIO>
+class KRATOS_API(OPTIMIZATION_APPLICATION) SpecializedContainerVariableData : public ContainerVariableData<TContainerType> {
```
matekelemen:
```suggestion
class KRATOS_API(KRATOS_CORE) SpecializedContainerVariableData : public ContainerVariableData<TContainerType> {
```

kratos/containers/container_variable_data/variable_expression_data_io.cpp
```diff
@@ -0,0 +1,220 @@
...
+{
+    IndexType local_index = 0;
+
+    const IndexType entity_data_begin_index = EntityIndex * rExpression.GetFlattenedSize();
+
+    for (IndexType i = 0; i < Value.size1(); ++i) {
+        for (IndexType j = 0; j < Value.size2(); ++j) {
+            rExpression.SetData(entity_data_begin_index, local_index++, Value(i, j));
+        }
+    }
```
matekelemen: how about assigning to the flattened array directly instead of the nested loop?
(provided `Matrix` exposes direct access to its underlying data)

matekelemen (dismissed):
This is awesome! It's an extremely useful set of features (especially the interface with raw numpy arrays) and a very clean implementation. I think it'll be super helpful for interop with 3rd party software through numpy, as well as prototyping/scripting for newcomers.

Now my only gripe is with the python interface's names. I know this started out from `OptimizationApplication` and variables that can store containers, but by now it evolved into something more general, and the naming should reflect that.

Once the names are simplified, I think it'd be a good idea to write some example scripts that show how the system can be used from python.  For example:
- fetching variables from specific `ModelPart`s and plotting them
- transforming the fetched data through numpy and assigning them back to some other variable
- perform some multistep operation on variables without storing temporaries 
- ... 

All in all excellent work @sunethwarna, thanks!

matekelemen (commented):
apparently, templates cannot (and must not) be exported

sunethwarna:
Hi @matekelemen. You are correct to remove the `KRATOS_API` from the templated `SpecializedContainerExpression` class because, the headers only classes does not need the `KRATOS_API`, and this is headers only :D

kratos/containers/container_variable_data/expressions/binary/binary_expression.h
sunethwarna: I;m not sure this is possible since the `BinaryExpression` class is templated, and these template arguments are `BinaryOperations`. How is it possible to put the template arguments within the templated class and instantiate in this case? :D.... :/

kratos/containers/container_variable_data/expressions/literal/literal_flat_expression.cpp
sunethwarna: short form for `Vector`.  I did not use the long name because, the expresions can get messy. I am open for suggestions :D

kratos/containers/container_variable_data/expressions/binary/binary_expression.h
matekelemen: ah yes, I see now. Nevermind then, it's fine as it is.

kratos/containers/container_variable_data/expressions/literal/literal_flat_expression.cpp
matekelemen: I don't see a problem with writing `Vector`, but if you think that's too much, I'd go with `Vec` instead.

kratos/python/add_container_variable_data_to_python.cpp
sunethwarna: done :)

kratos/python/add_container_variable_data_to_python.cpp
sunethwarna: done :)

kratos/containers/container_variable_data/container_variable_data.h
sunethwarna: done :)

kratos/containers/container_variable_data/container_variable_data.cpp
sunethwarna: done :)

kratos/containers/container_variable_data/container_variable_data.cpp
sunethwarna: done :)

kratos/containers/container_variable_data/container_variable_data.cpp
sunethwarna: done :)

kratos/containers/container_variable_data/container_variable_data.cpp
sunethwarna: done :)

kratos/containers/container_variable_data/container_variable_data.cpp
sunethwarna: I am not sure how to use `std::transform`. We don't have two vectors in here. Is there a way?

kratos/containers/container_variable_data/container_variable_data.cpp
sunethwarna: I would keep the existing.... Since we are working in python and c++, this seems like a funny expression ;) ;)

kratos/containers/container_variable_data/expressions/expression.h
sunethwarna: I need to start using a spell corrector in vscode... :D

kratos/containers/container_variable_data/expressions/expression.cpp
sunethwarna: done :)

kratos/containers/container_variable_data/expressions/literal/literal_flat_expression.h
sunethwarna: done :)

kratos/containers/container_variable_data/expressions/literal/literal_flat_expression.h
sunethwarna: done :)

kratos/containers/container_variable_data/expressions/literal/literal_flat_expression.h
sunethwarna: done :)

kratos/containers/container_variable_data/expressions/literal/literal_flat_expression.cpp
sunethwarna: done ;)

kratos/containers/container_variable_data/expressions/literal/literal_expression.h
sunethwarna: removed :D

kratos/containers/container_variable_data/variable_expression_data_io.cpp
sunethwarna: it supports and done :)

kratos/containers/container_variable_data/expressions/literal/literal_flat_expression.h
sunethwarna: removed :)

sunethwarna:
@matekelemen I applied your suggestion. Could you check it? :D

kratos/containers/container_variable_data/container_variable_data.cpp
matekelemen: yeah right you need the index, my bad

matekelemen (approved):
👌

PR #10929, opened 2023-03-28
[Dam] New process to apply different Young Modulus along the structure using Random Fields

joaquinirazabal:
**📝 Description**
A new process has been created to apply different Young Modulus along the structure using Random Fields.

- Applications, Enhancement

**🆕 Changelog**
- Added new process


aronnoordam:
Hi @joaquinirazabal, we are working on the same problem. I was also using gstools, however I found out that gstools is licensed under LGPL, which is incompatible with Kratos BSD-4 license (it is compatible with BSD-3 though). I think the random fields have to be generated in another way (not using Gstools), or the Kratos license has to be altered to BSD-3. @rubenzorrilla, what is your opinion about this?

rubenzorrilla:
It is a bit delicate, so let me try to target the issue from two different perspectives.

From the implementation point of view, to do it safely and with meaningful information for the user is as easy as doing a try catch as follows
```py
try:
    import gstools
except ImportError:
    raise ImportError("This module requires 'gstools' or the information message you want to give to the user.")
```
The issues come with the licensing. If I'm not wrong, LGPL is not compatible with our BSD-4 as @aronnoordam pointed out. If so, I'm afraid you cannot include this in the current master branch. In any case, let me ping @roigcarlo who knows much better than me which are the license compatibilities.

joaquinirazabal:
Thank you @rubenzorrilla and @aronnoordam. Let's see what @roigcarlo say and is this is not possible I will try to find an alternative or I will keep those changes in another branch,

roigcarlo:
If the dependency is from a python package, in principle it is possible to leave it (legally speaking) as long a you don't distribute it along with Kratos.
- You can leave the import and make the Dam application depend on that library
- You cannot make a redistributable package that contains your application AND the gstools library together.

That being said, bear in mind that leaving it will essentially cause that anyone that wants to base their developments in the DAM ( either companies or particulars) will have to deal with a double licensing issue because of the incompatibility.

As @rubenzorrilla has suggested, the preferred solution here would be to leave it as an optional import, and state that if allow to use that library, some extra functionality is present in the app.

roigcarlo:
Also, if you want the pip package to work out of the box, you need to fill up the dependencies in https://github.com/KratosMultiphysics/Kratos/blob/master/applications/DamApplication/DamApplication.json Otherwise it will not be downloaded along with the KratosDamApplication package.

djvicente (approved):

PR #10931, opened 2023-03-28
[Core] Bugfix - STL reader - fix new entitity ids

ddiezrod:
This is failing when the model part ids do not start with one and are consecutively numbered. I am making sure now that new ids are not used by any other entity.

### DIFF
--- a/kratos/input_output/stl_io.cpp
+++ b/kratos/input_output/stl_io.cpp
@@ -17,6 +17,8 @@
 // Project includes
 #include "includes/define.h"
 #include "input_output/stl_io.h"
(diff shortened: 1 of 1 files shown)


rubenzorrilla (approved):

philbucher:
Shouldn't this create Geometries?

loumalouomega:
> Shouldn't this create Geometries?

I agree with you

ddiezrod:
I think that would be make more sense @philbucher , I will try to do the change in a new PR (I am not sure if everything will work on our end)

PR #10932, opened 2023-03-28
[GeoMechanicsApplciation] Geo/pre calculate mass and damping matrix in solver

### LINKED ISSUE #10804: [Builder and Solver] passing constant Damping and mass matrix
In a situation where my mass and damping matrix are constant throughout solution steps/ non-linear iterations. I would like to pass the damping and mass matrix through the function "CalculateRHSContribution" in a scheme.  Currently every non-linear iteration I call "CalculateRHSContribution", this function then requires to recalculate the mass and damping matrix, to add the mass and damping contribution to the RHS. 

Is there a way to pass existing mass and damping matrices to "CalculateRHSContribution", such that I don't have to recalculate?

aronnoordam:
added build and solver in which the global mass and damping matrices are build and saved, such that they are not rebuild when building the RHS


AlejandroCornejo:
just to note that this could be useful in the core @RiccardoRossi 

@aronnoordam In this implementation **M** and **D** are only computed once?

aronnoordam:
> just to note that this could be useful in the core @RiccardoRossi
> 
> @aronnoordam In this implementation **M** and **D** are only computed once?

Hi @AlejandroCornejo , the mass and damping matrix are only computed when the whole system is built, they are not rebuild when only the RHS is built. So it depends on your strategy when the M and D are built. Note that for this implementation, when the user wants to rebuild the LHS every iteration, the M and D matrix are build both in the strategy and in the builder, thus making this implementation slow.

applications/GeoMechanicsApplication/custom_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_mass_and_damping.h
```diff
@@ -0,0 +1,634 @@
...
+                }
+            }
+        }
+
+        TSystemVectorType first_derivative_vector;
+        TSystemVectorType second_derivative_vector;
+
+        GetFirstAndSecondDerivativeVector(first_derivative_vector, second_derivative_vector, rModelPart);
+
+        TSystemVectorType mass_contribution = prod(mMassMatrix, second_derivative_vector);
```
aronnoordam: @loumalouomega , is this the most efficient way of doing a dot product of a sparse matrix and vector?

AlejandroCornejo: nope, the `TSparseSpace::Mult(const MatrixType& A, const VectorType& in, VectorType& out)` method should be used I think

loumalouomega: Indeed, Ublas Spaces Mult must be used

aronnoordam: okay, thank you!

Vahid-Galavi:
@aronnoordam I only have a general remark. If you use Rayleigh damping, it is a function of both mass and stiffness matrices. In case, the stiffness matrix is updated per iteration, the damping matrix should also be recalculated. So, the damping matrix should be updated together with the stiffness matrix.

AlejandroCornejo:
> @aronnoordam I only have a general remark. If you use Rayleigh damping, it is a function of both mass and stiffness matrices. In case, the stiffness matrix is updated per iteration, the damping matrix should also be recalculated. So, the damping matrix should be updated together with the stiffness matrix.

That is a good point. I have read both approaches. Some authors use the initial stiffness to be accounted for the damping. Do you have insight on that? Using the updated stiffness can add a certain non-linearity that I do not know if is always convenient

Vahid-Galavi:
> > @aronnoordam I only have a general remark. If you use Rayleigh damping, it is a function of both mass and stiffness matrices. In case, the stiffness matrix is updated per iteration, the damping matrix should also be recalculated. So, the damping matrix should be updated together with the stiffness matrix.
> 
> That is a good point. I have read both approaches. Some authors use the initial stiffness to be accounted for the damping. Do you have insight on that? Using the updated stiffness can add a certain non-linearity that I do not know if it is always convenient

Yes, I think it will make the system highly non-linear and inconvenient in some cases. Maybe updating the damping matrix can be activated through an input parameter just to do some experiment.

aronnoordam:
> @aronnoordam I only have a general remark. If you use Rayleigh damping, it is a function of both mass and stiffness matrices. In case, the stiffness matrix is updated per iteration, the damping matrix should also be recalculated. So, the damping matrix should be updated together with the stiffness matrix.

Hi @Vahid-Galavi , this is what I'm doing, I only rebuild the mass and damping matrix when the whole system is build, so also the stiffness matrix. When i do build, I build the stiffness matrix -> mass matrix -> damping matrix, in this order. I re-use the mass and damping matrix when I'm building soleley the rhs

aronnoordam:
> > > @aronnoordam I only have a general remark. If you use Rayleigh damping, it is a function of both mass and stiffness matrices. In case, the stiffness matrix is updated per iteration, the damping matrix should also be recalculated. So, the damping matrix should be updated together with the stiffness matrix.
> > 
> > 
> > That is a good point. I have read both approaches. Some authors use the initial stiffness to be accounted for the damping. Do you have insight on that? Using the updated stiffness can add a certain non-linearity that I do not know if it is always convenient
> 
> Yes, I think it will make the system highly non-linear and inconvenient in some cases. Maybe updating the damping matrix can be activated through an input parameter just to do some experiment.

I can add an input parameter to keep the damping matrix constant at all times, even when the system is rebuilt

Vahid-Galavi:
> > > > @aronnoordam I only have a general remark. If you use Rayleigh damping, it is a function of both mass and stiffness matrices. In case, the stiffness matrix is updated per iteration, the damping matrix should also be recalculated. So, the damping matrix should be updated together with the stiffness matrix.
> > > 
> > > 
> > > That is a good point. I have read both approaches. Some authors use the initial stiffness to be accounted for the damping. Do you have insight on that? Using the updated stiffness can add a certain non-linearity that I do not know if it is always convenient
> > 
> > 
> > Yes, I think it will make the system highly non-linear and inconvenient in some cases. Maybe updating the damping matrix can be activated through an input parameter just to do some experiment.
> 
> I can add an input parameter to keep the damping matrix constant at all times, even when the system is rebuilt

No, I think it is not needed.

aronnoordam:
> > > > > @aronnoordam I only have a general remark. If you use Rayleigh damping, it is a function of both mass and stiffness matrices. In case, the stiffness matrix is updated per iteration, the damping matrix should also be recalculated. So, the damping matrix should be updated together with the stiffness matrix.
> > > > 
> > > > 
> > > > That is a good point. I have read both approaches. Some authors use the initial stiffness to be accounted for the damping. Do you have insight on that? Using the updated stiffness can add a certain non-linearity that I do not know if it is always convenient
> > > 
> > > 
> > > Yes, I think it will make the system highly non-linear and inconvenient in some cases. Maybe updating the damping matrix can be activated through an input parameter just to do some experiment.
> > 
> > 
> > I can add an input parameter to keep the damping matrix constant at all times, even when the system is rebuilt
> 
> No, I think it is not needed.

Okay

applications/GeoMechanicsApplication/custom_python/add_custom_strategies_to_python.cpp
```diff
@@ -106,6 +107,13 @@ void AddCustomStrategiesToPython(pybind11::module& m)
         BuilderAndSolverType::Pointer, Parameters&, int, bool, bool, bool >())
     .def("UpdateLoads",&GeoMechanicsRammArcLengthStrategyType::UpdateLoads);
 
+
+
+    typedef ResidualBasedBlockBuilderAndSolverWithMassAndDamping< SparseSpaceType, LocalSpaceType, LinearSolverType > ResidualBasedBlockBuilderAndSolverWithMassAndDampingType;
```
avdg81: Please use `using` rather than `typedef`.

applications/GeoMechanicsApplication/custom_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_mass_and_damping.h
```diff
@@ -0,0 +1,658 @@
...
+///@name  Functions
+///@{
+
+///@}
+///@name Kratos Classes
+///@{
+
+/**
+ * @class ResidualBasedBlockBuilderAndSolverWithMassAndDamping
+ * @ingroup KratosCore
```
avdg81:
```suggestion
 * @ingroup GeoMechanicsApplication
```

applications/GeoMechanicsApplication/custom_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_mass_and_damping.h
```diff
@@ -0,0 +1,658 @@
...
+ * @brief Current class provides an implementation for builder and solving operations, while the global
+ * mass and damping matrices are stored.
+ * @details When the LHS is build, the global mass and damping matrices are build seperately. When building the RHS,
```
avdg81: Just a small typo:
```suggestion
 * @details When the LHS is build, the global mass and damping matrices are build separately. When building the RHS,
```

applications/GeoMechanicsApplication/custom_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_mass_and_damping.h
```diff
@@ -0,0 +1,658 @@
...
+ * This class is intended to be used when the mass and damping matrices are constant throughout the iterations,
+ * using this class, when rebuilding the LHS every iteration is not efficient.
+ * @tparam TSparseSpace The sparse system considered
+ * @tparam TDenseSpace The dense system considered
+ * @tparam TLinearSolver The linear solver considered
+ * @author Aron Noordam
+ */
+template<class TSparseSpace,
+         class TDenseSpace, //= DenseSpace<double>,
+         class TLinearSolver //= LinearSolver<TSparseSpace,TDenseSpace>
```
avdg81: What is the intent of the comments here? Do you plan to set these as the defaults for the template parameters?

applications/GeoMechanicsApplication/custom_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_mass_and_damping.h
```diff
@@ -0,0 +1,658 @@
...
+    ///@}
+    ///@name Life Cycle
+    ///@{
+
+    /**
+     * @brief Default constructor
+     */
+    explicit ResidualBasedBlockBuilderAndSolverWithMassAndDamping() : BaseType()
+    {
+    }
```
avdg81: To be in line with [C.80: Use =default if you have to be explicit about using the default semantics](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c80-use-default-if-you-have-to-be-explicit-about-using-the-default-semantics) from the C++ Core Guidelines I would change this as follows:
```suggestion
    ResidualBasedBlockBuilderAndSolverWithMassAndDamping() = default;
```

applications/GeoMechanicsApplication/custom_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_mass_and_damping.h
```diff
@@ -0,0 +1,658 @@
...
+
+    /**
+     * @brief Default constructor. (with parameters)
```
avdg81: This constructor is not a default constructor, since it can't be called called without any argument.

applications/GeoMechanicsApplication/custom_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_mass_and_damping.h
```diff
@@ -0,0 +1,658 @@
...
+        Parameters ThisParameters
+        ) : BaseType(pNewLinearSystemSolver)
+    {
+        // Validate and assign defaults
+        ThisParameters = this->ValidateAndAssignParameters(ThisParameters, this->GetDefaultParameters());
+        this->AssignSettings(ThisParameters);
+    }
+
+    /**
+     * @brief Default constructor.
```
avdg81: Neither is this one.

applications/GeoMechanicsApplication/custom_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_mass_and_damping.h
```diff
@@ -0,0 +1,658 @@
...
+    explicit ResidualBasedBlockBuilderAndSolverWithMassAndDamping(typename TLinearSolver::Pointer pNewLinearSystemSolver)
+        : BaseType(pNewLinearSystemSolver)
+    {
+    }
+
+    /** Destructor.
+     */
+    ~ResidualBasedBlockBuilderAndSolverWithMassAndDamping() override
+    {
+    }
```
avdg81: Use `= default` to define a trivial destructor:
```suggestion
    ~ResidualBasedBlockBuilderAndSolverWithMassAndDamping() override = default;
```

applications/GeoMechanicsApplication/custom_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_mass_and_damping.h
```diff
@@ -0,0 +1,658 @@
...
+        ModelPart& rModelPart,
+        TSystemMatrixType& rA,
+        TSystemVectorType& rb) override
+    {
+        KRATOS_TRY
+
+        KRATOS_ERROR_IF(!pScheme) << "No scheme provided!" << std::endl;
+
+        // Getting the elements from the model
+        const int nelements = static_cast<int>(rModelPart.Elements().size());
```
avdg81: Avoid repeating the type (the `static_cast` is already explicit about it):
```suggestion
        const auto nelements = static_cast<int>(rModelPart.Elements().size());
```

applications/GeoMechanicsApplication/custom_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_mass_and_damping.h
```diff
@@ -0,0 +1,658 @@
...
+
+        // Getting the array of the conditions
+        const int nconditions = static_cast<int>(rModelPart.Conditions().size());
```
avdg81:
```suggestion
        const auto nconditions = static_cast<int>(rModelPart.Conditions().size());
```

applications/GeoMechanicsApplication/custom_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_mass_and_damping.h
```diff
@@ -0,0 +1,658 @@
...
+
+        const ProcessInfo& r_current_process_info = rModelPart.GetProcessInfo();
+        ModelPart::ElementsContainerType::iterator el_begin = rModelPart.ElementsBegin();
+        ModelPart::ConditionsContainerType::iterator cond_begin = rModelPart.ConditionsBegin();
+
+        //contributions to the system
+        LocalSystemMatrixType lhs_contribution = LocalSystemMatrixType(0, 0);
+        LocalSystemMatrixType mass_contribution = LocalSystemMatrixType(0, 0);
+        LocalSystemMatrixType damping_contribution = LocalSystemMatrixType(0, 0);
+        LocalSystemVectorType rhs_contribution = LocalSystemVectorType(0);
```
avdg81: Also here type information is being repeated. What about this:
```suggestion
        LocalSystemMatrixType lhs_contribution(0, 0);
        LocalSystemMatrixType mass_contribution(0, 0);
        LocalSystemMatrixType damping_contribution(0, 0);
        LocalSystemVectorType rhs_contribution(0);
```

applications/GeoMechanicsApplication/custom_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_mass_and_damping.h
```diff
@@ -0,0 +1,658 @@
...
+
+        mMassMatrix.resize(BaseType::mEquationSystemSize, BaseType::mEquationSystemSize, false);
+
+        BaseType::ConstructMatrixStructure(pScheme, mMassMatrix, rModelPart);
+
+        mDampingMatrix.resize(BaseType::mEquationSystemSize, BaseType::mEquationSystemSize, false);
+        BaseType::ConstructMatrixStructure(pScheme, mDampingMatrix, rModelPart);
+
+        TSparseSpace::SetToZero(mMassMatrix);
+        TSparseSpace::SetToZero(mDampingMatrix);
```
avdg81: I would suggest to extract the following member function:
```
void InitializeMatrix(TSystemMatrixType& rMatrix, unsigned int MatrixSize, typename TSchemeType::Pointer pScheme, ModelPart& rModelPart)
{
    rMatrix.resize(MatrixSize, MatrixSize, rModelPart);
    BaseType::ConstructMatrixStructure(pScheme, rMatrix, rModelPart);
    TSparseSpace::SetToZero(rMatrix);
}
```
And replace the inline code by two function calls:
```
    InitializeMatrix(mMassMatrix);
    InitializeMatrix(mDampingMatrix);
```
I haven't compiled and tested the above suggestions, so please check them carefully.

applications/GeoMechanicsApplication/custom_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_mass_and_damping.h
```diff
@@ -0,0 +1,658 @@
...
+    {
+        KRATOS_TRY
+
+    	Timer::Start("BuildRHS");
+
+        BuildRHSNoDirichlet(pScheme, rModelPart, rb);
+
+        //NOTE: dofs are assumed to be numbered consecutively in the BlockBuilderAndSolver
+        block_for_each(BaseType::mDofSet, [&](Dof<double>& r_dof) {
+            const std::size_t i = r_dof.EquationId();
```
avdg81: I would move this statement to the if-clause, since `i` is only used there (so limit the scope of `i` to where it is actually being used).

applications/GeoMechanicsApplication/custom_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_mass_and_damping.h
```diff
@@ -0,0 +1,658 @@
...
+
+        Timer::Stop("BuildRHS");
+
+        KRATOS_CATCH("")
+    }
+
+
+    /**
+     * @brief This method provides the defaults parameters to avoid conflicts between the different constructors
+     * @return The default parameters
```
avdg81, line 399: Note that virtual functions should not be called from constructors. See [C.50: Use a factory function if you need “virtual behavior” during initialization](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c50-use-a-factory-function-if-you-need-virtual-behavior-during-initialization) of the C++ Core Guidelines.

applications/GeoMechanicsApplication/custom_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_mass_and_damping.h
```diff
@@ -0,0 +1,658 @@
...
+protected:
+    ///@name Protected static Member Variables
+    ///@{
+
+    ///@}
+    ///@name Protected member Variables
+    ///@{
+
+    TSystemMatrixType mMassMatrix;
+    TSystemMatrixType mDampingMatrix;
```
avdg81: Please avoid protected data members as per [C.133: Avoid protected data](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c133-avoid-protected-data) of the C++ Core Guidelines. (SonarCloud will also complain about this.)

applications/GeoMechanicsApplication/custom_strategies/builder_and_solvers/residualbased_block_builder_and_solver_with_mass_and_damping.h
```diff
@@ -0,0 +1,658 @@
...
+
+    ///@}
+    ///@name Protected Operations
+    ///@{
+
+
+    void GetFirstAndSecondDerivativeVector(TSystemVectorType& rFirstDerivativeVector, TSystemVectorType& rSecondDerivativeVector, ModelPart& rModelPart)
+    {
+        NodesArrayType& r_nodes = rModelPart.Nodes();
+        const int n_nodes = static_cast<int>(r_nodes.size());
```
avdg81: Avoid repeating the type:
```suggestion
        const auto n_nodes = static_cast<int>(r_nodes.size());
```

aronnoordam:
@avdg81  as discussed, corrected suggestions except initialisation, as it is intertwined throughout the whole kratos code

avdg81 (approved):

PR #10936, opened 2023-03-30
[MPM] Element classes renamed

andinoboerst:
**📝 Description**
The Updated Lagrangian Element Class name causes confusion when running a simulation with the Co-Simulation Application. For this reason, the MPM Classes were renamed using a unique name to avoid wrong serialization.

Implemented in combination with KratosMultiphysics/GiDInterface#929 .

### DIFF
--- a/applications/CoSimulationApplication/tests/mpm_dem/beam_Body.mdpa
+++ b/applications/CoSimulationApplication/tests/mpm_dem/beam_Body.mdpa
@@ -93,7 +93,7 @@ Begin Nodes
 End Nodes
 
 
-Begin Elements UpdatedLagrangian2D3N// GUI group identifier: Solid Auto1
+Begin Elements MPMUpdatedLagrangian2D3N// GUI group identifier: Solid Auto1
         1          0    87    97    68 
         2          0   347   342   336 
         3          0    47    37    53 
--- a/applications/CoSimulationApplication/tests/mpm_fem_beam/particle_Body.mdpa
+++ b/applications/CoSimulationApplication/tests/mpm_fem_beam/particle_Body.mdpa
@@ -459,7 +459,7 @@ Begin Nodes
 End Nodes
 
 
-Begin Elements UpdatedLagrangian2D4N// GUI group identifier: particle
+Begin Elements MPMUpdatedLagrangian2D4N// GUI group identifier: particle
         1          0        428        429        440        439 
         2          0        417        418        429        428 
         3          0        406        407        418        417 
--- a/applications/ParticleMechanicsApplication/custom_elements/updated_lagrangian.cpp
+++ b/applications/ParticleMechanicsApplication/custom_elements/updated_lagrangian.cpp
@@ -34,15 +34,15 @@ namespace Kratos
 /**
  * Flags related to the element computation
  */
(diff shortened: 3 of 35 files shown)


VeronikaSinger (approved):

rubenzorrilla:
Note that in Kratos the general rule is to implement one class per file and match the class with the filename (one in CamelCase and the other in snake_case). After this PR ou're no longer following this convention. I'd do a PR to amend this. Note that by doing `git mv` you can rename the files without loosing their history.

andinoboerst:
@rubenzorrilla thanks for pointing that out. I renamed the files in PR #10954.

philbucher:
I am wondering how this was working before

Didn't this cause problems when importing both apps?
The KratosComponents do bot allow to register different objects with the same names

andinoboerst:
I don't know how this didn't cause any problems before, but it was possible to run the simulations initially.
I think @VeronikaSinger mentioned that she had a similar problem with the conditions before as well though.

The problem with the elements only started to occur when using the restart serialization. But I think other applications might have the same problem, as there are other applications using the same element names I believe.

ncrescenzio:
Hi @philbucher, just to understand what was happening, when you say that `KratosComponents` do not allow to register different objects with the same names, do you refer to the name of the class implementing the element (before this PR both the `ParticleMechanicsApplication` and the `StructuralMechanicsApplication` implemented a class called `UpdatedLagrangian`) or to the name between quotation marks used as first argument of the macro `KRATOS_REGISTER_ELEMENT` (in this case the names used in `particle_mechanics_application.cpp` were all different from the ones used in `structural_mechanics_application.cpp` even before this PR)?

As far as I have understood, no error occurred when importing both the `ParticleMechanicsApplication` and the `StructuralMechanicsApplication` because the names used as first argument of `KRATOS_REGISTER_ELEMENT` were not the same in the two applications (therefore no problems when calling `KratosComponents::Add` inside the macro `KRATOS_REGISTER_ELEMENT`).

However, an error occurs when using the restart in a coupled simulation that uses `CoSimulationApplication`, `ParticleMechanicsApplication` and `StructuralMechanicsApplication`. In particular, the error occurs when trying to load the `*.rest` files of the `ParticleMechanicsApplication`. Looking at these files, the name of the element that is written is wrong: I found `UpdatedLagrangianElement2D3N` (an element defined in the `StructuralMechanicsApplication`) instead of `UpdatedLagrangian` (the element used in the simulation for the mpm part and defined in the `ParticleMechanicsApplication`). I think that the wrong name is written because these two elements have the same `typeid(element).name()` but the first to be registered using `Serializer::Register` is the one from the `StructuralMechanicsApplication` (i.e., `UpdatedLagrangianElement2D3N`). Changing manually the name from `UpdatedLagrangianElement2D3N` to `UpdatedLagrangian` in the `*rest` file then the restart works.

andinoboerst:
@ncrescenzio, correct, that's why the class names had to be changed.
As a precaution, I changed the name in quotation marks as well to make sure that there are no other errors generated in the future in case another application would choose to use the same name in quotation marks. If the name in quotation marks is the same then the cosimulation doesn't work at all.

PR #10944, opened 2023-03-31
[Core] Removed dofs initialization instance in explicit_solving_strategy

Marco1410:
**📝 Description**
I removed the DOFs initialization instance in the explicit strategy to allow imposing initial conditions before initializing the solver.

Before, when the strategy started these conditions were set to zero.

This allows generalizing its use through multistage.

**🆕 Changelog**
- Removed InitializeDofSetValues


rubenzorrilla (approved):
For the record, I'd like to add that this was meaningless as the historical database automatically allocates and initialize the nodal values.

Issue #10950, opened 2023-03-31
[Core][Geometry] Line element edges are not consistent 


aronnoordam:
- The edgenumber in the line2d3,line2d4, line2d5, line3d3 element is 2 while the edgenumber in the lined2d2 is 1;

- GenerateEdges is not present in the  line2d3,line2d4, line2d5 line3d3 element, while it is in the line2d2 element

loumalouomega:
We should discuss how to unify behaviour. I will consider line2d2 as reference, as the one most used. So adding `GenerateEdges`  to all lines and set `EdgeNumber`  to 1 in every line geometry.

loumalouomega:
@KratosMultiphysics/technical-committee and @KratosMultiphysics/implementation-committee

avdg81:
We (@KratosMultiphysics/geomechanics) have prepared the following PR that aims at resolving the inconsistency; see #13916. We would appreciate it very much if we may bring this issue back to your attention, since we will need the proposed improvements in GeoMechanicsApplication in the near future. Thank you.

avdg81:
Closing this issue, since a fix was merged a while ago. See https://github.com/KratosMultiphysics/Kratos/pull/13916.

PR #10951, opened 2023-03-31
[GeoMechanicsApplication] Geo/10805 normal load

### LINKED ISSUE #10805: [GeoMechanicsApplication] The condition "Normal Load" does not work correctly in Geo applications
**Description**
The "Normal Load" is tested to set normal pressure on a beam. It leads to error related to linking to elements. 
It wrorks correctly in StructuralMechanics but not in GeoMechanics.

It seems this issue was also recognized by other coleagues in sheetpiling and settelment. 

**Scope**
The bug occurs during runtime with error message:

    self.find_neighbour_elements_of_conditions_process.Execute()
    RuntimeError: Error: Some conditions found without any corresponding element

**To Reproduce**
The bug, more probably, comes from find_neighbour_elements_of_conditions_process function.
This function needs to be reviewed and we need to find out the reason why it works with other conditions but not with the normal load.

**Expected behavior**
After fix, the cases with the normal load conditions (normal pressure) need to run succesfully.

**Environment**
Tested on Windows 10

[GeoBeam.zip](https://github.com/KratosMultiphysics/Kratos/files/10848547/GeoBeam.zip)

aronnoordam:
1d elements are now found in the search neighbour process, but only when the condition is not connected to 2d or 3d elements.

added test for:
- normal load on 2d2n beam 
- normal load on 2d2n beam attached to soil

For now, problems in higher order line conditions stay until #10950 is solved


applications/GeoMechanicsApplication/custom_processes/find_neighbour_elements_of_conditions_process.cpp
```diff
@@ -214,40 +176,103 @@ void FindNeighbourElementsOfConditionsProcess::Execute()
...
+    PointerVector<Element>::iterator pItElem)
+{
+
+	const std::pair<hashmap::iterator, hashmap::iterator> face_pair = rFacesMap.equal_range(rItFace->first);
+    for (hashmap::iterator it = face_pair.first; it != face_pair.second; ++it) {
+        std::vector<Condition::Pointer>& r_conditions = it->second;
+
+        GlobalPointersVector< Element > vector_of_neighbours;
+        vector_of_neighbours.resize(1);
+        vector_of_neighbours(0) = Element::WeakPointer(*pItElem.base());
```
mcgicjn2:
```suggestion
	const std::pair<hashmap::iterator, hashmap::iterator> face_pair = rFacesMap.equal_range(rItFace->first);
        for (hashmap::iterator it = face_pair.first; it != face_pair.second; ++it) {
            std::vector<Condition::Pointer>& r_conditions = it->second;

            GlobalPointersVector< Element > vector_of_neighbours;
            vector_of_neighbours.resize(1);
            vector_of_neighbours(0) = Element::WeakPointer(*pItElem.base());
```
Corrected the alignment

mcgicjn2 (changes requested):
Slight change required otherwise fine by me.

mnabideltares (dismissed):

applications/GeoMechanicsApplication/custom_processes/find_neighbour_elements_of_conditions_process.cpp
```diff
@@ -160,35 +141,16 @@ void FindNeighbourElementsOfConditionsProcess::Execute()
...
-
-                        GlobalPointersVector< Element > VectorOfNeighbours;
-                        VectorOfNeighbours.resize(1);
-                        VectorOfNeighbours(0) = Element::WeakPointer( *itElem.base() );
-
-                        for (Condition::Pointer pCondition : ListConditions) {
-                            pCondition->Set(VISITED,true);
-                            pCondition->SetValue(NEIGHBOUR_ELEMENTS, VectorOfNeighbours);
-                        }
-                    }
```
avdg81, line 176: Very nice that you have found and eliminated this duplicated piece of code 😄

applications/GeoMechanicsApplication/custom_processes/find_neighbour_elements_of_conditions_process.cpp
```diff
@@ -214,40 +176,103 @@ void FindNeighbourElementsOfConditionsProcess::Execute()
...
 }
 
+bool FindNeighbourElementsOfConditionsProcess::CheckIfAllConditionsAreVisited() const
+{
+    for (auto& r_cond : mrModelPart.Conditions()) {
+        if (r_cond.IsNot(VISITED)) {
+            return false;
+        }
+    }
+    return true;
```
avdg81: I believe this is an algorithm from the STL:
```suggestion
    const auto& rConditions = mrModelPart.Conditions();
    return std::all_of(rConditions.begin(), rConditions.end(),
                       [](const auto& r_cond){return r_cond.Is(VISITED);});
```
Using such an algorithm makes your intent very clear and easy to read. But please check (and test) this suggested change carefully before incorporating it in your work.

applications/GeoMechanicsApplication/custom_processes/find_neighbour_elements_of_conditions_process.cpp
```diff
@@ -214,40 +176,103 @@ void FindNeighbourElementsOfConditionsProcess::Execute()
...
+
+void FindNeighbourElementsOfConditionsProcess::CheckIf1DElementIsNeighbour(hashmap& rFacesMap)
+{
+    // Now loop over all elements and check if one of the faces is in the "FacesMap"
+    for (auto itElem = mrModelPart.ElementsBegin(); itElem != mrModelPart.ElementsEnd(); ++itElem) {
+        const auto& r_geometry_element = itElem->GetGeometry();
+
+	    // for 1D elements, the edge geometry is the same as the element geometry 
+	    if (r_geometry_element.LocalSpaceDimension() == 1)
+	    {
```
avdg81: I have found a few accidentals tabs here.

applications/GeoMechanicsApplication/custom_processes/find_neighbour_elements_of_conditions_process.cpp
```diff
@@ -214,40 +176,103 @@ void FindNeighbourElementsOfConditionsProcess::Execute()
...
+            const auto rBoundaryGeometries = PointerVector(r_geometry_element.GenerateEdges());
+
+            for (IndexType iFace = 0; iFace < rBoundaryGeometries.size(); ++iFace) {
+                DenseVector<int> FaceIds(rBoundaryGeometries[iFace].size());
+
+                // get edge ids
+                for (IndexType iNode = 0; iNode < FaceIds.size(); ++iNode) {
+                    FaceIds[iNode] = rBoundaryGeometries[iFace][iNode].Id();
+                }
```
avdg81: I believe this is a transformation:
```suggestion
                const auto& rNodes = rBoundaryGeometries[iFace];
                std::transform(rNodes.begin(), rNodes.end(), FaceIds.begin(),
                               [](const auto& rNode){ return rNode.Id(); });
```
And once again: please check this suggestion carefully before incorporating it. Thanks.

applications/GeoMechanicsApplication/custom_processes/find_neighbour_elements_of_conditions_process.cpp
```diff
@@ -214,40 +176,103 @@ void FindNeighbourElementsOfConditionsProcess::Execute()
...
+        }
+    }
+}
+
+
+void FindNeighbourElementsOfConditionsProcess::CheckForMultipleConditionsOnElement(hashmap& rFacesMap, hashmap::iterator& rItFace, 
+    PointerVector<Element>::iterator pItElem)
+{
+
+	const std::pair<hashmap::iterator, hashmap::iterator> face_pair = rFacesMap.equal_range(rItFace->first);
```
avdg81: I have found an accidental tab here. Also the following appear to have incorrect indentation levels, which makes it more difficult to read.

avdg81 (commented):
I have looked at the C++ code only. You have made some nice improvements there, so thanks for that.

mcgicjn2 (approved):

PR #10953, opened 2023-04-02
[OptApp] Move OptimizationInfo to CPP

sunethwarna:
**📝 Description**
This PR moves the current python implementation of `OptimizationInfo`  to cpp with following allowed data types.

- `ModelPart`
- `bool`
- `int`
- `double`
- `string`
- `ContainerExpressions`

Requires #10768 .

Todo:

- [ ] Extend to `PropertiesContainerExpressions`
- [ ] Remove existing python `OptimizationInfo`

**🆕 Changelog**
- Ported python `OptimizationInfo` to cpp
- Added tests


sunethwarna:
Since, `OptimizationApplication` needs to be driven by python even to support third party libs which does not have python interfacing, it does not make sense to port `OptimizationInfo` to cpp. Hence closing this while keeping the branch. If required, we can visit this later.

PR #10937, opened 2023-03-30
[Core] STL Reader - Adding new entities to geometry list

ddiezrod:
As suggested by @philbucher in https://github.com/KratosMultiphysics/Kratos/pull/10931 I am adding new entities to the geometry container instead of the element one.

PR #10961, opened 2023-04-03
[Core] STL Reader - pass new entity type by parameters

ddiezrod:
I tried to use stl reader with geometries after https://github.com/KratosMultiphysics/Kratos/pull/10937 but the problem is that in many places we cannot directly use this, as many processes/utilities do not work with geometry containers, only with elements/conditions. @pooyan-dadvand proposed to pass this as an option, so by default we still read the new entities as geometries, but we can also create elements or conditions.

This class already received a Flags object in the constructor. I thought it'd be more clear to put everything together under Parameters.

### DIFF
--- a/kratos/input_output/stl_io.cpp
+++ b/kratos/input_output/stl_io.cpp
@@ -23,47 +23,74 @@
 namespace Kratos
 {
 
-StlIO::StlIO(std::filesystem::path const& Filename, const Flags Options)
-    : mOptions(Options)
+StlIO::StlIO(std::filesystem::path const& Filename, Parameters ThisParameters):
+    mParameters(ThisParameters)
 {
-    Kratos::shared_ptr<std::fstream> pFile = Kratos::make_shared<std::fstream>();
+    mParameters.ValidateAndAssignDefaults(GetDefaultParameters());
+    Kratos::shared_ptr<std::fstream> p_file = Kratos::make_shared<std::fstream>();
 
     // set default mode to read
-    std::fstream::openmode OpenMode;
+    std::fstream::openmode open_mode;
 
+    const std::string open_mode_str = mParameters["open_mode"].GetString();
     // handle other mode options
-    if (mOptions.Is(IO::READ)) {
-        OpenMode = std::fstream::in;
-    } else if (mOptions.Is(IO::APPEND)) {
-        OpenMode = std::fstream::in | std::fstream::app;
-    } else if (mOptions.Is(IO::WRITE)) {
-        OpenMode = std::fstream::out;
+    if (open_mode_str == "read") {
+        open_mode = std::fstream::in;
+    } else if (open_mode_str == "append") {
+        open_mode = std::fstream::in | std::fstream::app;
+    } else if (open_mode_str == "write") {
+        open_mode = std::fstream::out;
     } else {
-        KRATOS_ERROR << "Unsupported IO options: " << Options << std::endl;
+        KRATOS_ERROR << "Unsupported open mode: " << open_mode_str << std::endl;
     }
 
-    pFile->open(Filename.c_str(), OpenMode);
+    p_file->open(Filename.c_str(), open_mode);
 
-    KRATOS_ERROR_IF_NOT(pFile->is_open()) << "Could not open the input file  : " << Filename << std::endl;
+    KRATOS_ERROR_IF_NOT(p_file->is_open()) << "Could not open the input file  : " << Filename << std::endl;
 
     // Store the pointer as a regular std::iostream
-    mpInputStream = pFile;
+    mpInputStream = p_file;
 }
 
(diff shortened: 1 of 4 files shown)


loumalouomega:
Indeed, the proper workflow would be to read geometries and then parse it into elements/conditions with `ReplaceElementsAndConditionsProcess`. Of course, as you say, the current expected behaviour is different and that change would break current code. Approving then.

kratos/input_output/stl_io.cpp
```diff
@@ -23,47 +23,74 @@
...
 
+    const std::string open_mode_str = mParameters["open_mode"].GetString();
     // handle other mode options
-    if (mOptions.Is(IO::READ)) {
-        OpenMode = std::fstream::in;
-    } else if (mOptions.Is(IO::APPEND)) {
-        OpenMode = std::fstream::in | std::fstream::app;
-    } else if (mOptions.Is(IO::WRITE)) {
-        OpenMode = std::fstream::out;
+    if (open_mode_str == "read") {
```
loumalouomega, line 37: Why not flag?

loumalouomega (changes requested):
Some comment

kratos/input_output/stl_io.cpp
philbucher: Legacy, I fully agree with parameters

loumalouomega: It is fully agreed with @KratosMultiphysics/technical-committee ?, I will remove my block

loumalouomega (dismissed):

philbucher:
> parse it into elements/conditions with ReplaceElementsAndConditionsProcess

I would not touch this one

Instead we need a proper `CreateElementsAndConditionsModeler`, which creates `Element`s an `Condition`s from `Geomety`s

kratos/input_output/stl_io.cpp
```diff
@@ -208,16 +234,23 @@ void StlIO::ReadLoop(ModelPart & rThisModelPart)
...
+
     Element::NodesArrayType temp_geom_nodes;
     while(word == "vertex"){
         Point coordinates = ReadPoint();
-        temp_geom_nodes.push_back(rThisModelPart.CreateNewNode(node_id++, coordinates[0], coordinates[1], coordinates[2] ));
+        temp_geom_nodes.push_back(rThisModelPart.CreateNewNode(mNextNodeId++, coordinates[0], coordinates[1], coordinates[2] ));
         *mpInputStream >> word; // Reading vertex or endloop
     }
-    rThisModelPart.CreateNewGeometry("Triangle3D3", temp_geom_nodes);
+    const std::string new_entity_type = mParameters["new_entity_type"].GetString();
```
philbucher, line 274: this is quite inefficient, since it is parsed for every triangle

I would instead create a lambda or std::function once, which you can pass

kratos/tests/cpp_tests/input_output/test_stl_io.cpp
```diff
@@ -51,6 +51,37 @@ KRATOS_TEST_CASE_IN_SUITE(ReadTriangleFromSTL, KratosCoreFastSuite)
...
+    Model current_model;
+    ModelPart& r_model_part = current_model.CreateModelPart("Main");
+
+    Parameters settings(R"({
+        "new_entity_type" : "element"
+    })");
+    StlIO stl_io(p_input,settings);
+    stl_io.ReadModelPart(r_model_part);
+
+    std::cout << r_model_part << std::endl;
```
philbucher:
```suggestion
```

kratos/input_output/stl_io.cpp
ddiezrod: Ok! done ;)

kratos/tests/cpp_tests/input_output/test_stl_io.cpp
ddiezrod: done!

ddiezrod:
Comments addressed! I like your idea about the new modeler. We should apply the changes for https://github.com/KratosMultiphysics/Kratos/pull/10726 directly there I think.

loumalouomega:
Ci failing: 

~~~sh
[ 12%] Building CXX object kratos/CMakeFiles/KratosCore.dir/Unity/unity_5_cxx.cxx.o
In file included from /__w/Kratos/Kratos/build/FullDebug/kratos/CMakeFiles/KratosCore.dir/Unity/unity_5_cxx.cxx:61:
/__w/Kratos/Kratos/kratos/input_output/stl_io.cpp:94:31: error: lambda capture 'this' is not used [-Werror,-Wunused-lambda-capture]
        create_entity_func = [this](
                              ^~~~
[ 12%] Building CXX object kratos/CMakeFiles/KratosCore.dir/Unity/unity_4_cxx.cxx.o
~~~

loumalouomega (approved):

PR #10968, opened 2023-04-04
[Core] pass settings to apply ray casting process

ddiezrod:
This PR modifies apply ray casting class so settings can be passed by Parameters object instead of passing each setting as an argument. The default behavior stays the same. 

(I recommend to hide whitespace changes as indentation was using tabs in these classes)

### DIFF
--- a/kratos/processes/apply_ray_casting_process.cpp
+++ b/kratos/processes/apply_ray_casting_process.cpp
@@ -86,7 +86,7 @@ namespace Kratos
         } else if (database == "nodal_non_historical") {
             node_distance_getter = [](NodeType& rNode, const Variable<double>& rDistanceVariable)->double&{return rNode.GetValue(rDistanceVariable);};
         } else {
-            KRATOS_ERROR << "Unrecognized database " <<  database << std::endl;
+            KRATOS_ERROR << "Provided 'distance_database' is '" << database << "'. Available options are 'nodal_historical' and 'nodal_non_historical'." <<  std::endl;
         }
 
         const Variable<double>* mpDistanceVariable = &KratosComponents<Variable<double>>::Get(mSettings["distance_variable"].GetString());
--- a/kratos/processes/apply_ray_casting_process.cpp
+++ b/kratos/processes/apply_ray_casting_process.cpp
@@ -24,102 +24,98 @@
 namespace Kratos
 {
 
-	template<std::size_t TDim>
-	ApplyRayCastingProcess<TDim>::ApplyRayCastingProcess(
-		ModelPart& rVolumePart,
-		ModelPart& rSkinPart)
-		: mpFindIntersectedObjectsProcess(new FindIntersectedGeometricalObjectsProcess(rVolumePart, rSkinPart)),
-		  mIsSearchStructureAllocated(true)
-	{
-	}
+    template<std::size_t TDim>
+    ApplyRayCastingProcess<TDim>::ApplyRayCastingProcess(
+        ModelPart& rVolumePart,
+        ModelPart& rSkinPart,
+        Parameters ThisParameters)
+        : mSettings(ThisParameters),
+          mpFindIntersectedObjectsProcess(new FindIntersectedGeometricalObjectsProcess(rVolumePart, rSkinPart)),
+          mIsSearchStructureAllocated(true)
+    {
+        mSettings.ValidateAndAssignDefaults(GetDefaultParameters());
+        mRelativeTolerance = mSettings["relative_tolerance"].GetDouble();
+    }
 
-	template <std::size_t TDim>
-	ApplyRayCastingProcess<TDim>::ApplyRayCastingProcess(
-		ModelPart &rVolumePart,
-		ModelPart &rSkinPart,
-		const double RelativeTolerance)
-		: mRelativeTolerance(RelativeTolerance),
-		  mpFindIntersectedObjectsProcess(new FindIntersectedGeometricalObjectsProcess(rVolumePart, rSkinPart)),
-		  mIsSearchStructureAllocated(true)
-	{
-	}
+    template <std::size_t TDim>
+    ApplyRayCastingProcess<TDim>::ApplyRayCastingProcess(
+        FindIntersectedGeometricalObjectsProcess &TheFindIntersectedObjectsProcess,
+        Parameters ThisParameters)
+        : mSettings(ThisParameters),
+          mpFindIntersectedObjectsProcess(&TheFindIntersectedObjectsProcess),
+          mIsSearchStructureAllocated(false)
+    {
+        mSettings.ValidateAndAssignDefaults(GetDefaultParameters());
+        mRelativeTolerance = mSettings["relative_tolerance"].GetDouble();
+    }
 
-	template <std::size_t TDim>
-	ApplyRayCastingProcess<TDim>::ApplyRayCastingProcess(
-		FindIntersectedGeometricalObjectsProcess &TheFindIntersectedObjectsProcess,
-		const double RelativeTolerance)
-		: mRelativeTolerance(RelativeTolerance),
-		  mpFindIntersectedObjectsProcess(&TheFindIntersectedObjectsProcess),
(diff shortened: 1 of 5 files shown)


loumalouomega (dismissed):

loumalouomega (dismissed):

kratos/processes/apply_ray_casting_process.cpp
```diff
@@ -24,102 +24,100 @@
...
+
+        // Set the getter function according to the database to be used
+        NodeScalarGetFunctionType node_distance_getter;
+        const std::string database = mSettings["distance_database"].GetString();
+        if (database == "nodal_historical") {
+            node_distance_getter = [](NodeType& rNode, const Variable<double>& rDistanceVariable)->double&{return rNode.FastGetSolutionStepValue(rDistanceVariable);};
+        } else if (database == "nodal_non_historical") {
+            node_distance_getter = [](NodeType& rNode, const Variable<double>& rDistanceVariable)->double&{return rNode.GetValue(rDistanceVariable);};
+        } else {
+            KRATOS_ERROR << "Unrecognized database " <<  database << std::endl;
```
rubenzorrilla:
```suggestion
            KRATOS_ERROR << "Provided unrecognized database '" <<  database << "'. Available options are 'nodal_historical' and 'nodal_non_historical'." << std::endl;
```

rubenzorrilla (changes requested):
You're removing the old constructors. Though I agree with the addition and promotion of the new `Parameters`-based one this would break current API. Instead of removing them, I'd suggest to deprecate those.

kratos/processes/apply_ray_casting_process.h
```diff
@@ -209,16 +179,14 @@ class KRATOS_API(KRATOS_CORE) ApplyRayCastingProcess : public Process
     ///@name Member Variables
     ///@{
 
+    Parameters mSettings;
     double mEpsilon = 1.0e-12;
     double mExtraRayOffset = 1.0e-8;
```
rubenzorrilla: I'd take the chance to make this user-definable by including them in the settings.

kratos/processes/apply_ray_casting_process.h
```diff
@@ -209,16 +179,14 @@ class KRATOS_API(KRATOS_CORE) ApplyRayCastingProcess : public Process
...
     double mRelativeTolerance = 1.0e-12;
     FindIntersectedGeometricalObjectsProcess* mpFindIntersectedObjectsProcess;
     bool mIsSearchStructureAllocated;
     double mCharacteristicLength = 1.0;
```
rubenzorrilla: ~Not 100% sure if we should expose this one too.~

EDIT: Forget about this, it should be computed later on based on the background mesh.

kratos/processes/calculate_distance_to_skin_process.cpp
```diff
@@ -24,278 +24,285 @@
...
-		mRayCastingRelativeTolerance = rParameters["ray_casting_relative_tolerance"].GetDouble();
-        CalculateDiscontinuousDistanceToSkinProcess<TDim>::mpElementalDistancesVariable = &KratosComponents<Variable<Vector>>::Get(rParameters["elemental_distances_variable"].GetString());;
-        mpDistanceVariable = &KratosComponents<Variable<double>>::Get(rParameters["distance_variable"].GetString());;
-		const std::string distance_database = rParameters["distance_database"].GetString();
-		if (distance_database == "nodal_historical") {
-			mDistanceDatabase = DistanceDatabase::NodeHistorical;
-		} else if (distance_database == "nodal_non_historical") {
-			mDistanceDatabase = DistanceDatabase::NodeNonHistorical;
-		} else {
-			KRATOS_ERROR << "Provided 'distance_database' is '" << distance_database << "'. Available options are 'nodal_historical' and 'nodal_non_historical'." <<  std::endl;
```
rubenzorrilla, line 62: Indeed, I'd use this same error message in the error thrown you're adding to the ray-casting process to keep consistency.

kratos/python/add_processes_to_python.cpp
```diff
@@ -462,12 +462,12 @@ void  AddProcessesToPython(pybind11::module& m)
     // Continuous distance computation methods
     py::class_<ApplyRayCastingProcess<2>, ApplyRayCastingProcess<2>::Pointer, Process>(m,"ApplyRayCastingProcess2D")
         .def(py::init<ModelPart&, ModelPart&>())
-        .def(py::init<ModelPart&, ModelPart&, double>())
```
rubenzorrilla, line 465: You're breaking current Python API.

kratos/python/add_processes_to_python.cpp
```diff
@@ -462,12 +462,12 @@ void  AddProcessesToPython(pybind11::module& m)
...
+        .def(py::init<ModelPart&, ModelPart&, Parameters>())
     ;
 
     py::class_<ApplyRayCastingProcess<3>, ApplyRayCastingProcess<3>::Pointer, Process>(m,"ApplyRayCastingProcess3D")
         .def(py::init<ModelPart&, ModelPart&>())
-        .def(py::init<ModelPart&, ModelPart&, double>())
```
rubenzorrilla, line 470: Same.

kratos/python/add_processes_to_python.cpp
ddiezrod: Yes, but it is not used anywhere...

kratos/processes/calculate_distance_to_skin_process.cpp
ddiezrod: I think I will do it the other way around as this message contains more info.

kratos/processes/apply_ray_casting_process.h
ddiezrod: This is being set here actually https://github.com/KratosMultiphysics/Kratos/blob/6aa01de1cfa6a5c56ddfff2ad16d422f1a3ffef0/kratos/processes/apply_ray_casting_process.cpp#L500

I will remove the default values as they are kind of confusing

rubenzorrilla: Agree.

kratos/python/add_processes_to_python.cpp
rubenzorrilla: It might be used in any fork.

kratos/processes/calculate_distance_to_skin_process.cpp
rubenzorrilla: Yes, this is what I meant. Using the one that contains more info.

kratos/python/add_processes_to_python.cpp
ddiezrod: ok!

kratos/processes/apply_ray_casting_process.h
```diff
@@ -99,16 +102,17 @@ class KRATOS_API(KRATOS_CORE) ApplyRayCastingProcess : public Process
...
-    /**
+	/**
      * @brief Construct a new Apply Ray Casting Process object using an already created search strucutre
      *
      * @param TheFindIntersectedObjectsProcess reference to the already created search structure
      * @param RelativeTolerance user-defined relative tolerance to be multiplied by the domain bounding box size
      * @param pDistanceVariable user-defined variabe to be used to read and store the distance to the skin
      * @param rDistanceDatabase enum value specifying the database from which the distance variable is retrieved (see DistanceDatabase)
      */
+    KRATOS_DEPRECATED_MESSAGE("Deprecated constructor, please use the one with Parameters.")
```
rubenzorrilla, line 115: OK but this will throw a warning at compile time, which is not useful at all to warn the users during the deprecation period. Hence,  I'd also throw the warning in the constructor body, so the user can see it when running the problem.

ddiezrod: ok, done

rubenzorrilla (approved):
No more concerns from my side. Thanks @ddiezrod, these are important steps towards exploiting the new registry + factories.

PR #10972, opened 2023-04-04
Kratos ROM with CoSimulation

SADPR:
### Description:

This pull request enables the use of CoSimulation with reduced order models in the Kratos Multiphysics repository. Specifically, it adds a `rom_wrapper` to facilitate this functionality.

Previously, working with two different `RomParameters.json` files was not possible, so I adapted the `rom_analysis.py` file to read the `rom_basis_output_name` from the `calculate_rom_basis_output_process`. If this parameter is not specified, `rom_analysis` now uses the default name `RomParameters.json`.

Additionally, to enable the transfer of Neumann boundary conditions in the coupling interface data, I built a `CalculateReactions` function inside the `ROM Builder And Solver`. Note that `HROM` has not yet been tested as it may require an extra step for coupling interface transmission.

To showcase and test the new functionality, I included a unit test for a `2D thermal-thermal` CoSimulation problem in the `rom application unit tests`.

### Changes Made:

- Added `rom_wrapper` to enable CoSimulation with reduced order models
- Adapted `rom_analysis.py` to read `rom_basis_output_name` from `calculate_rom_basis_output_process`
- Built `CalculateReactions` function inside `ROM Builder And Solver` to transfer Neumann boundary conditions
- Included unit test for a `2D thermal-thermal` CoSimulation problem in `rom application unit tests`.


loumalouomega:
This is awesome, many posibilities

philbucher (changes requested):
wow awesome!

Blocking as I want to review this in a bit more detail in the next days

applications/RomApplication/python_scripts/rom_analysis.py
```diff
@@ -19,9 +19,17 @@ def __init__(self,global_model, parameters):
 
         def _CreateSolver(self):
             """ Create the Solver (and create and import the ModelPart if it is not alread in the model) """
+            self.rom_basis_output_name = 'RomParameters.json' #Default
```
matekelemen, line 22: Is JSON you primary output format? If yes, I strongly recommend looking for something else.

rubenzorrilla: It used to be. Now there are other alternatives available.

applications/RomApplication/python_scripts/rom_analysis.py
```diff
@@ -19,9 +19,17 @@ def __init__(self,global_model, parameters):
...
+            if self.project_parameters.Has("output_processes"):
+                for name, _ in self.project_parameters["output_processes"].items():
+                    if name=="rom_output":
+                        rom_output_paramaters = self.project_parameters["output_processes"]["rom_output"]
+                        rom_basis_output_format = rom_output_paramaters[0]["Parameters"]["rom_basis_output_format"].GetString()
+                        rom_basis_output_name = rom_output_paramaters[0]["Parameters"]["rom_basis_output_name"].GetString()
+                        self.rom_basis_output_name = rom_basis_output_name + "." + rom_basis_output_format
```
rubenzorrilla, line 29: Why is this related to the `CoSimulationApplication` enabling? I think we already upgrade the output stuf in @Rbravo555 's last PR.

applications/RomApplication/custom_strategies/rom_builder_and_solver.h
```diff
@@ -708,6 +708,41 @@ class ROMBuilderAndSolver : public BuilderAndSolver<TSparseSpace, TDenseSpace, T
...
+                    if (dofs[i]->IsFixed()){
+                        rb[dofs[i]->EquationId()] += rTLS.rhs_elem[i]; // Building RHS.
+                        dofs[i]->GetSolutionStepReactionValue() = -rb[dofs[i]->EquationId()]; // Reactions are not Zero in the next step for ROMs. Reaction will be overwritten if dof is repeated in other elements, nontheless, it avoids to double loop.
+                    }
+                }
+
+            });
+        }   
+        KRATOS_CATCH("")
+    }
```
rubenzorrilla, line 745: This is a very delicate class. Please move changes in this file to a separate PR.

rubenzorrilla (changes requested):
Changes in the `RomApplication`  must be done in a separate PR (one for the B&S and another one for the `rom_analysis.py`.

Besides, I wonder if we should add the test to the `CoSimulationApplication` instead.

applications/RomApplication/custom_strategies/rom_builder_and_solver.h
```diff
@@ -708,6 +708,41 @@ class ROMBuilderAndSolver : public BuilderAndSolver<TSparseSpace, TDenseSpace, T
...
+        TSystemVectorType& rDx,
+        TSystemVectorType& rb) override
+    {
+        KRATOS_TRY
+        // Get ProcessInfo from main model part
+        const auto& r_current_process_info = rModelPart.GetProcessInfo();
+
+        struct my_tls {
+            Kratos::Vector rhs_elem;
+        };
```
rubenzorrilla, line 724: You don't need an auxiliary struct in here (see below).
```suggestion
```

applications/RomApplication/custom_strategies/rom_builder_and_solver.h
```diff
@@ -708,6 +708,41 @@ class ROMBuilderAndSolver : public BuilderAndSolver<TSparseSpace, TDenseSpace, T
...
+
+        auto& r_elements = rModelPart.Elements();
+        if(!r_elements.empty())
+        {
+            block_for_each(r_elements, my_tls(), [&](Element& r_element, my_tls& rTLS)
```
rubenzorrilla, line 729:
```suggestion
            block_for_each(r_elements, Vector(), [&](Element& r_element, Vector& rTLS)

```

applications/RomApplication/custom_strategies/rom_builder_and_solver.h
```diff
@@ -708,6 +708,41 @@ class ROMBuilderAndSolver : public BuilderAndSolver<TSparseSpace, TDenseSpace, T
...
+            {
+                Element::GeometryType& geom = r_element.GetGeometry();
```
rubenzorrilla, line 731:
```suggestion
```
This is not used.

applications/RomApplication/custom_strategies/rom_builder_and_solver.h
```diff
@@ -708,6 +708,41 @@ class ROMBuilderAndSolver : public BuilderAndSolver<TSparseSpace, TDenseSpace, T
...
+                DofsVectorType dofs = {};
+                r_element.CalculateRightHandSide(rTLS.rhs_elem, r_current_process_info);
```
rubenzorrilla, line 733:
```suggestion
                r_element.CalculateRightHandSide(rTLS, r_current_process_info);

```

applications/RomApplication/custom_strategies/rom_builder_and_solver.h
```diff
@@ -708,6 +708,41 @@ class ROMBuilderAndSolver : public BuilderAndSolver<TSparseSpace, TDenseSpace, T
...
+                r_element.GetDofList(dofs, r_current_process_info);
+                for (IndexType i = 0; i < dofs.size(); ++i){
+                    if (dofs[i]->IsFixed()){
+                        rb[dofs[i]->EquationId()] += rTLS.rhs_elem[i]; // Building RHS.
```
rubenzorrilla, line 737:
```suggestion
                        rb[dofs[i]->EquationId()] += rTLS[i]; // Building RHS.

```

rubenzorrilla: Besides, I think that this addition is not threadsafe.

applications/RomApplication/custom_strategies/rom_builder_and_solver.h
```diff
@@ -708,6 +708,41 @@ class ROMBuilderAndSolver : public BuilderAndSolver<TSparseSpace, TDenseSpace, T
...
+                        dofs[i]->GetSolutionStepReactionValue() = -rb[dofs[i]->EquationId()]; // Reactions are not Zero in the next step for ROMs. Reaction will be overwritten if dof is repeated in other elements, nontheless, it avoids to double loop.
```
rubenzorrilla, line 738: Most probably I'm missing something here, but doesn't this mean that you wouldn't be assembling all the element contributions but the last one to that node?

Besides, I also think this addition is not threadsafe.

applications/RomApplication/python_scripts/rom_analysis.py
SADPR: This code enables users to have two different `RomParameters.json` files, such as `RomParameters_simulation_1.json` and `RomParameters_simulation_2.json`. This feature can be useful when users need to simulate different scenarios that require different parameters. For example, in the context of potential flow, users may need to simulate both the primal and adjoint problems, which would require different sets of parameters.

applications/RomApplication/custom_strategies/rom_builder_and_solver.h
SADPR: I agree with your point. It would be necessary to merge the reactions calculation before merging this pull request, as the reactions are needed to properly couple the problem and ensure a valid test.

applications/RomApplication/custom_strategies/rom_builder_and_solver.h
SADPR: I will make sure this additions are threadsafe with an AtomicAdd.

applications/RomApplication/python_scripts/rom_analysis.py
rubenzorrilla: OK, but please create a dedicated PR for it.

applications/RomApplication/custom_strategies/rom_builder_and_solver.h
SADPR: I changed my approach to assigning reactions to DOFs separately. Here, I was using a trick to avoid building the RHS first, but I realized that following the traditional way in Kratos is safer and more straightforward. Now, the code should be easier to maintain, understand, and ensures it works as expected (uou will see in the new PR).

SADPR: The pull request for calculating the reactions in ROM is the following #10985.

SADPR:
This pull request will be split into three separate PRs:

1. Calculate ROM reactions - This PR adds the code to calculate ROM reactions #10985.

2. Running a ROM case with different name on the RomParameters.json - This PR adds support for running a ROM case with a different name in the RomParameters.json file.  In the case of a numpy we have an extra folder.

3. CoSimulation ROM wrapper with corresponding unit test - This PR adds a CoSimulation ROM wrapper and its corresponding unit test.

Each of these PRs will be created from a separate brancht. Please note that the third PR depends on the first PR being merged, as the unit test requires the ROM reactions to be calculated before it can be run.

Please let me know if you have any questions or concerns. Thanks!

roigcarlo:
@SADPR Is this PR still need or the separate PR's are enough?

PR #10982, opened 2023-04-10
[Core] Ray Casting process - add option to set nodes on interface to zero

ddiezrod:
When computing ray distances we check if the distance to the intersection is lower than mEpsilon here
https://github.com/KratosMultiphysics/Kratos/blob/7a7e56d2259ddf2ad524884ee8f0d7b3c5ed8eeb/kratos/processes/apply_ray_casting_process.cpp#L172

However, this is not considered later when we set the nodal distance, we only change the sign there. Here I am adding the option to set the distance to zero on nodes that are close to the interface (closer than mEpsilon). This would be useful for me to recognize objects that are completely intersect with the skin. Current behaviour is kept though by default.

### DIFF
--- a/kratos/processes/apply_ray_casting_process.cpp
+++ b/kratos/processes/apply_ray_casting_process.cpp
@@ -103,7 +103,8 @@ namespace Kratos
         return Parameters(R"({
             "distance_database" : "nodal_historical",
             "distance_variable" : "DISTANCE",
-            "relative_tolerance" : 1e-12
+            "relative_tolerance" : 1e-12,
+            "set_distance_on_interface_to_zero" : false
         })");
 
     }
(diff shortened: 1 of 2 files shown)


loumalouomega (approved):
Okay from my side

rubenzorrilla (changes requested):
Though current behavior is  kept, we need to carefully consider changes in this class.

First, is this a bugfix or introducing a new feature/behavior. If the first, I think it makes no sense to keep current behavior.

On the contrary, if it is the second I'm of the opinion that, instead of adding an option to circunvent current behavior to use this process for a valid but different purpose, a derived class with clear intended usage must be created. 

EDIT:
Besides, regardless if it is the first or the second option, a test is missing.

@loumalouomega all the levelset-related algorithms are extremely delicated classes, with and admittedly very low testing coverage considering the amount of corner cases. I'd appreciate you to be more careful with your approvals when these are involved.

ddiezrod:
I think this is more a design decision rather than a bug. I also considered if this should go to a new class or as an option and I do not have a strong opinion, I did it this way first because it was easier for the discussion. I'll do it the other way around.

rubenzorrilla:
> 

If so, I think it would be better to create a `DetectSkinObjectProcess` (or something of this sort) with clear settings and intended usage. Besides, I also think that this would give us complete freedom to give special treatment to the corner cases that can eventually arise.

ddiezrod:
> > 
> 
> If so, I think it would be better to create a `DetectSkinObjectProcess` (or something of this sort) with clear settings and intended usage. Besides, I also think that this would give us complete freedom to give special treatment to the corner cases that can eventually arise.

The thing is I need to identify all cases, if a node falls out, in, or on the interface. It'd be good if I could do that with only one process.

rubenzorrilla:
@ddiezrod as far as I understand we can close this one now. Am I wrong?

ddiezrod:
> @ddiezrod as far as I understand we can close this one now. Am I wrong?

Yep! closing

PR #10988, opened 2023-04-12
[Core] Ray Casting - create virtualfunction to set distances

ddiezrod:
As an alternative to https://github.com/KratosMultiphysics/Kratos/pull/10982, here I am just making private functions protected and using a virtual function to set the nodal distances.

The idea is that I will create a class on our side that does what we need: using ray casting, classify nodes as in, out or interface.

### DIFF
--- a/kratos/processes/apply_ray_casting_process.cpp
+++ b/kratos/processes/apply_ray_casting_process.cpp
@@ -129,14 +129,14 @@ namespace Kratos
             KRATOS_ERROR << "Provided 'distance_database' is '" << database << "'. Available options are 'nodal_historical' and 'nodal_non_historical'." <<  std::endl;
         }
 
-        const Variable<double>* mpDistanceVariable = &KratosComponents<Variable<double>>::Get(mSettings["distance_variable"].GetString());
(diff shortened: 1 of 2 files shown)


rubenzorrilla (commented):
Just pointing that I'm much more in favor of this solution. I think it much clean as it doesn't pollute current implementation and having a new class gives us more flexibility to deal with the eventual corner cases that may appear in your target application.

ddiezrod:
This is fine for me too, shall we merge this?

pooyan-dadvand (commented):
Also agree that this is a clean solution! 

But I would make the function more generic by passing the node and the ray_distance, so the derived class can use it to fix the interface nodes or any other operation with node based on its position.

ddiezrod:
I have tried my best do it generic (I think it makes a lot of sense that we only pass the node and the ray distance to a function that can be overrided by other classes) and also more or less efficient, so I had to store the distance_variable and the distance gettter in the class. @pooyan-dadvand @rubenzorrilla

rubenzorrilla (approved):
This makes perfect sense for me. Thanks @ddiezrod for taking my comments into consideration.

PR #10990, opened 2023-04-12
Kratos rom multiple rom parameters

SADPR:
### Description

This pull request enables the user to specify a custom name for the `RomParameters.json` file used in ROM and HROM simulations. Previously, all ROMs had to use a file named `RomParameters.json`. With this change, users can pass the name of their custom `RomParameters.json` file to the `calculate_rom_basis_output_process` and ensure consistency with the `rom_basis_output_name` given when building the basis.

Additionally, this change supports both the `json` and `numpy` formats for the `RomParameters` file. For the `numpy` format, a folder named `numpy_rom_data_{name of RomParameters}` is created that contains the matrices and vectors for the corresponding `RomParameters.json` file. The existing tests were updated to include the `numpy_rom_data_RomParameters` folder for each case without changing the `ExpectedOutput` of the tests. This flexibility in naming conventions and format can be particularly useful for working with CoSimulations, PotentialFlow (primal and adjoint models), and/or any simulations that involve two models.

The changes were made in the following files:
- `rom_analysis.py`
- `hrom_training_utility.py`
- `petrov_galerkin_training_utility`
- `calculate_rom_basis_output_process`
- `rom_manager`

### Changes Made
- Updated files to accept a custom name for the `RomParameters.json` file.
- Ensured that the `calculate_rom_basis_output_process` is consistent with the `rom_basis_output_name` used to build the basis.
- When choosing `numpy` as format, a new folder named `numpy_rom_data_{name of RomParameters}` will gather all the numpy binaries needed for the corresponding `RomParameters`.
- Updated tests (no change in the `ExpectedOutput` and bases). Only added the binaries to their corresponding folder.


applications/RomApplication/python_scripts/hrom_training_utility.py
```diff
@@ -97,7 +98,7 @@ def CreateHRomModelParts(self):
         if  self.hrom_output_format == "numpy":
             hrom_info = KratosMultiphysics.Parameters(json.JSONEncoder().encode(self.__CreateDictionaryWithRomElementsAndWeights()))
         elif self.hrom_output_format == "json":
-            with open('RomParameters.json','r') as f:
+            with open(self.rom_basis_output_name + ".json",'r') as f:
```
rubenzorrilla:
```suggestion
            with open("{self.rom_basis_output_name}.json",'r') as f:
```
Since we dropped Python2 support we can use f-strings.

applications/RomApplication/python_scripts/hrom_training_utility.py
```diff
@@ -212,16 +213,16 @@ def AppendHRomWeightsToRomParameters(self):
         if self.hrom_output_format=="numpy":
             element_indexes = np.where( indexes < number_of_elements )[0]
             condition_indexes = np.where( indexes >= number_of_elements )[0]
-            np.save('HROM_ElementWeights.npy',weights[element_indexes])
-            np.save('HROM_ConditionWeights.npy',weights[condition_indexes])
-            np.save('HROM_ElementIds.npy',indexes[element_indexes]) #FIXME fix the -1 in the indexes of numpy and ids of Kratos
-            np.save('HROM_ConditionIds.npy',indexes[condition_indexes]-number_of_elements) #FIXME fix the -1 in the indexes of numpy and ids of Kratos
+            np.save(f'numpy_rom_data_{self.rom_basis_output_name}/HROM_ElementWeights.npy',weights[element_indexes])
```
rubenzorrilla: When comparing with the json formatting above I realized that they behave slightly different. 

In the json case, we are saving the ROM basis in the current work directory. However, in the numpy case the files are always saved in a folder. On top, in the json case the `rom_basis_output_name` refers to the filename (which I think is what the user would expect). However, in the npy case this refers to the folder name (which is not expected by the user) but the filenames are hardcoded.

At this point, maybe the easiest solution is to always save the ROM basis in a folder according to a `rom_basis_output_folder` and prepend the filenames with the content of `rom_basis_output_name` in the case of HROM npy output.

applications/RomApplication/python_scripts/calculate_rom_basis_output_process.py
```diff
@@ -176,9 +177,18 @@ def _PrintRomBasis(self, snapshots_matrix):
                 i += n_nodal_unknowns
 
         elif self.rom_basis_output_format == "numpy":
+            # Set the name of the folder where the numpy basis are going to be stored (Same name as rom_basis_output_name).
+            folder_name = f"numpy_rom_data_{self.rom_basis_output_name}"
+
+            # Create the folder if it doesn't already exist
+            if not os.path.exists(folder_name):
+                os.makedirs(folder_name)
+                print("Folder created successfully!")
```
rubenzorrilla, line 186:
```suggestion
```

applications/RomApplication/python_scripts/calculate_rom_basis_output_process.py
```diff
@@ -176,9 +177,18 @@ def _PrintRomBasis(self, snapshots_matrix):
...
+            else:
+                print("Folder already exists.")
```
rubenzorrilla:
```suggestion
```

applications/RomApplication/python_scripts/petrov_galerkin_training_utility.py
```diff
@@ -126,25 +127,25 @@ def __AppendNewBasisToRomParameters(self, u):
 
         elif self.rom_format == "numpy":
             # Storing Petrov-Galerkin modes in Numpy format
-            np.save('LeftBasisMatrix.npy', u)
+            np.save(f'numpy_rom_data_{self.rom_basis_output_name}/LeftBasisMatrix.npy', u)
 
-        with open('RomParameters.json','r') as f:
+        with open(self.rom_basis_output_name + ".json",'r') as f:
```
rubenzorrilla:
```suggestion
        with open("{self.rom_basis_output_name}.json",'r') as f:
```

applications/RomApplication/python_scripts/petrov_galerkin_training_utility.py
```diff
@@ -126,25 +127,25 @@ def __AppendNewBasisToRomParameters(self, u):
...
             updated_rom_parameters = json.load(f)
             updated_rom_parameters["rom_settings"]["petrov_galerkin_number_of_rom_dofs"] = petrov_galerkin_number_of_rom_dofs
             updated_rom_parameters["petrov_galerkin_nodal_modes"] = petrov_galerkin_nodal_modes
 
-        with open('RomParameters.json','w') as f:
+        with open(self.rom_basis_output_name + ".json",'w') as f:
```
rubenzorrilla:
```suggestion
        with open("{self.rom_basis_output_name}.json",'w') as f:
```

applications/RomApplication/python_scripts/petrov_galerkin_training_utility.py
```diff
@@ -126,25 +127,25 @@ def __AppendNewBasisToRomParameters(self, u):
...
+        if self.echo_level > 0 : KratosMultiphysics.Logger.PrintInfo(f"PetrovGalerkinTrainingUtility","\'RomParameters.json\' file updated with HROM weights.")
 
     def __GetPrettyFloat(self, number):
         float_format = "{:.12f}"
         pretty_number = float(float_format.format(number))
         return pretty_number
 
     def __GetGalerkinBasis(self):
-        with open('RomParameters.json','r') as f:
+        with open(self.rom_basis_output_name + ".json",'r') as f:
```
rubenzorrilla:
```suggestion
        with open("{self.rom_basis_output_name}.json",'r') as f:
```

applications/RomApplication/python_scripts/rom_analysis.py
```diff
@@ -19,9 +19,16 @@ def __init__(self,global_model, parameters):
...
+            if self.project_parameters.Has("output_processes"):
+                for name, _ in self.project_parameters["output_processes"].items():
+                    if name=="rom_output":
+                        rom_output_paramaters = self.project_parameters["output_processes"]["rom_output"]
+                        rom_basis_output_format = rom_output_paramaters[0]["Parameters"]["rom_basis_output_format"].GetString()
+                        self.rom_basis_output_name = rom_output_paramaters[0]["Parameters"]["rom_basis_output_name"].GetString()
 
             # Get the ROM settings from the RomParameters.json input file
-            with open('RomParameters.json') as rom_parameters:
+            with open(self.rom_basis_output_name + ".json") as rom_parameters:
```
rubenzorrilla:
```suggestion
            with open("{self.rom_basis_output_name}.json") as rom_parameters:
```

applications/RomApplication/python_scripts/rom_analysis.py
```diff
@@ -19,9 +19,16 @@ def __init__(self,global_model, parameters):
 
         def _CreateSolver(self):
             """ Create the Solver (and create and import the ModelPart if it is not alread in the model) """
+            self.rom_basis_output_name = 'RomParameters' #Default
+            if self.project_parameters.Has("output_processes"):
+                for name, _ in self.project_parameters["output_processes"].items():
+                    if name=="rom_output":
+                        rom_output_paramaters = self.project_parameters["output_processes"]["rom_output"]
+                        rom_basis_output_format = rom_output_paramaters[0]["Parameters"]["rom_basis_output_format"].GetString()
+                        self.rom_basis_output_name = rom_output_paramaters[0]["Parameters"]["rom_basis_output_name"].GetString()
```
rubenzorrilla: This is a bit hacky. I'm starting to think that now that we have unified ROM solver and analysis plus the rom manager, maybe we can directly call the rom_output_process in the analysis stage rather than including it in the output processes.

We should at least consider and discuss this option.

applications/RomApplication/python_scripts/rom_analysis.py
```diff
@@ -156,10 +163,14 @@ def ModifyAfterSolverInitialize(self):
...
             # Check for HROM stages
             if self.train_hrom:
+                # Pass the name of the Rom Parameters file
+                self.rom_parameters.AddString("rom_basis_output_name", self.rom_basis_output_name)
                 # Create the training utility to calculate the HROM weights
                 self.__hrom_training_utility = HRomTrainingUtility(
                     self._GetSolver(),
                     self.rom_parameters)
+                # Remove the name of the Rom Parameters file
+                self.rom_parameters.RemoveValue("rom_basis_output_name")
```
rubenzorrilla: Why adding and removing?

applications/RomApplication/python_scripts/rom_analysis.py
```diff
@@ -183,9 +194,13 @@ def ModifyAfterSolverInitialize(self):
 
             # Check and Initialize Petrov Galerkin Training stage
             if self.train_petrov_galerkin:
+                # Pass the name of the Rom Parameters file
+                self.rom_parameters.AddString("rom_basis_output_name", self.rom_basis_output_name)
                 self.__petrov_galerkin_training_utility = PetrovGalerkinTrainingUtility(
                     self._GetSolver(),
                     self.rom_parameters)
+                # Remove the name of the Rom Parameters file
+                self.rom_parameters.RemoveValue("rom_basis_output_name")
```
rubenzorrilla: Same comment. Why adding and removing?

applications/RomApplication/python_scripts/rom_manager.py
```diff
@@ -230,6 +234,10 @@ def LaunchTrainHROM(self, mu_train):
         RedidualsSnapshotsMatrix = []
         for mu in mu_train:
             parameters = self.UpdateProjectParameters(parameters, mu)
+            parameters = self._AddBasisCreationToProjectParameters(parameters)
+            parameters["output_processes"]["rom_output"][0]["Parameters"]["snapshots_interval"].SetDouble(1e30)  # To avoid storing snapshots during training,
+            # we set the value to 1e30. This ensures that the snapshot storage mechanism is not triggered,
```
rubenzorrilla, line 239: Maybe a stupid question, but which is the reasoning behind adding the ROM output process to then deactivate the snapshot taking?

rubenzorrilla (changes requested):
I've some concerns about this. Blocking until them get clarified.

applications/RomApplication/python_scripts/hrom_training_utility.py
SADPR: Changing to:
```
f"{self.rom_basis_output_name}.json"
```

applications/RomApplication/python_scripts/hrom_training_utility.py
SADPR: To ensure consistency in the way we store ROM data, I suggest creating a default folder named `rom_data` for this purpose. Inside this folder, the user should be able to find the `RomParameters.json` file, and in case NumPy is used, all `.npy` files should also be stored there. This way, users can easily find all the relevant files in one place.

applications/RomApplication/python_scripts/rom_analysis.py
SADPR: I completely agree that we should consider calling rom_output_process directly from the analysis stage, rather than including it in the output processes. To achieve this, I suggest merging the relevant code and creating a separate pull request for the integration with the analysis stage, adding it to the list of TODOs. However, we should keep in mind that this method allows the user to manipulate the ROM output and folder names, so calling it from the analysis stage might require some extra options to allow the user to specify these names. Overall, I think this could be a good solution and I'm happy to work on it further.

applications/RomApplication/python_scripts/rom_manager.py
SADPR: The purpose of adding the ROM output process is to allow the 'calculate_rom_basis_output_process' to pass the ROM output name and folder name as strings.

applications/RomApplication/python_scripts/rom_analysis.py
rubenzorrilla: Agree. We'll need to expose some settings. In any case, lets do what you suggest, we merge this and then we discuss with the rest of the team how we approach this issue.

applications/RomApplication/python_scripts/hrom_training_utility.py
rubenzorrilla: Makes sense for me.

applications/RomApplication/python_scripts/rom_manager.py
rubenzorrilla: Again, way hacky solution. We need to figure out how to amend this.

applications/RomApplication/python_scripts/rom_analysis.py
SADPR: It is not used further, I want to be as extrusive as possible.

applications/RomApplication/python_scripts/rom_analysis.py
SADPR: Same comment as before

applications/RomApplication/python_scripts/rom_manager.py
SADPR: I will add it to the TODOs list.

SADPR:
Siwtching this PR to Draft because it will be adapted to the following PR. [#11063](https://github.com/KratosMultiphysics/Kratos/pull/11063)

roigcarlo:
If this was implemented in #11063 please close this.

PR #10991, opened 2023-04-12
Core/export unique dofs array type

rubenzorrilla:
**📝 Description**
This exports the `Unique` of the `DofsArrayType`. While adding the corresponding test I took the chance to amend a bit current one to make it (now them) more unittest-like.


kratos/python/add_dofs_to_python.cpp
```diff
@@ -59,6 +59,9 @@ void  AddDofsToPython(pybind11::module& m)
         .def("append", [](ModelPart::DofsArrayType &self, Dof<double> &value) {
             self.push_back(&value);
         })
+        .def("unique", [](ModelPart::DofsArrayType &self) {
```
loumalouomega, line 62: I have my reserves, your proposal is more pythonic, but then is not consistent with c++ methods and other exposed methods, in any case I agree

loumalouomega (dismissed):
Okay from my side

kratos/tests/test_dofs.py
```diff
@@ -35,6 +45,20 @@ def test_model_part_sub_model_parts(self):
         self.assertEqual(dx.GetReaction(), KratosMultiphysics.REACTION_X)
         self.assertEqual(dx.EquationId, 5)
 
+        self.assertTrue(p<p2)
+        self.assertTrue(dy>dx)
```
philbucher: There is assertGreater etc :)

rubenzorrilla: I left it like this in order to avoid any modification in current code. If this is a problem I've no inconvenience on changing it.

kratos/python/add_dofs_to_python.cpp
rubenzorrilla: True. In this case, I'd leave it like this to be consistent with the `append` export.

RiccardoRossi:
i am quite in favor of the current solution since it is a new addition...

kratos/tests/test_dofs.py
philbucher: It shows up as green thus thought its new

Wont hurt to change it, but up to you

Benefit of the other function is the better error message

rubenzorrilla: It is in green because I refactored a bit the previous one to be more unittest-like. In any case I'm changing it, thanks for the suggestion.

loumalouomega (approved):

PR #10993, opened 2023-04-13
[ThermalDEM] Small fixes for disk particles

rlrangel:
Fixing some heat conduction formulas to consider that 2D particles are cylinders with unitary length, which is consistent with the mechanical DEM analysis.


applications/ThermalDEMApplication/custom_constitutive/conduction/direct_conduction_pipe.cpp
```diff
@@ -31,7 +31,11 @@ namespace Kratos {
     const double kavg      = particle->ComputeAverageConductivity();
     const double temp_grad = particle->GetNeighborTemperature() - particle->GetParticleTemperature();
 
-    return kavg * (Globals::Pi * Rc * Rc) * temp_grad / d;
+    double area = 0.0;
+    if      (particle->mDimension == 2) area = 2.0 * Rc;
+    else if (particle->mDimension == 3) area = Globals::Pi * Rc * Rc;
```
AlejandroCornejo: it is cleaner to use `const double area = (particle->mDimension == 2) ? 2.0 * Rc : Globals::Pi * Rc * Rc;`

rlrangel: Changed

AlejandroCornejo: Ok! as a side comment, since in DEM any improvement counts, this file may be `templated `with the `dimension `so you could avoid this if with the dimension.

AlejandroCornejo (approved):

PR #10996, opened 2023-04-13
Update INSTALL.md

jtneto:
I have noticed that the installer may run out of memory on Linux. Was able to solve this by temporarily increasing swap size. I believe this is useful information to have in the instructions.

**📝 Description**
Added note to INSTALL.md

- Documentation

**🆕 Changelog**
- Added note to INSTALL.md

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -205,6 +205,8 @@ Then, these scripts can be launched through the system terminal.
 sh /path_to_kratos/scripts/configure.sh
 ```
 
+**NOTE**: In case the compiler runs out of memory, try increasing the swap size to at least 16 GB and re-starting the compilation process.
+
 *Windows*
 
 ```Shell


loumalouomega:
Okay, but are you compiling in which system?, a RPI4 with 4GB?

jtneto:
I'm using a HP Spectre Notebook, with 8th gen i7 processor and 8GB RAM, Ubuntu Linux 22.10.

loumalouomega:
Okay, 8 Gb could be problematic, but maybe you can solve the problem as well if you compile with just 1/2/3/4 threads, instead of -j$(nproc), just -j2

philbucher:
> Okay, 8 Gb could be problematic, but maybe you can solve the problem as well if you compile with just 1/2/3/4 threads, instead of -j$(nproc), just -j2

That seems like the better option than increasing swap size

roigcarlo:
Agree. Maybe we can keep the note suggesting the user to use less procs (or increasing the swap if there is no other way). Tbh it is pretty unclear what one should do if it runs out of ram.

roigcarlo (approved):

PR #11000, opened 2023-04-13
[Core] Add SpecializedSpatialSearch class and implement search methods

loumalouomega:
**📝 Description**

This PR introduces a new class called `SpecializedSpatialSearch`, which is a template class that takes a `SpatialContainer` as a template parameter. The class is used to search for elements, nodes, and conditions in a radius around given elements, nodes, or conditions. The PR adds the following methods to the class:

1. `SearchElementsInRadiusExclusive`: Searches for neighbours of elements, excluding the elements themselves, in the given radius.
2. `SearchNodesInRadiusExclusive`: Searches for neighbours of nodes, excluding the nodes themselves, in the given radius.
3. `SearchConditionsInRadiusExclusive`: Searches for neighbours of conditions, excluding the conditions themselves, in the given radius.

This commit specializes in the `SpecializedSpatialSearch` class for `KDTree`, `Octree`, `BinsStatic`, and `BinsDynamic` spatial container. 

In addition, the PR introduces checks to ensure that consistent sizes are provided for input containers and radius arrays, improving the robustness of the search methods. This helps prevent potential errors and issues related to mismatched container sizes.

A Python test is also added to verify the functionality of the `SpecializedSpatialSearch` class and its specialized implementations for the different spatial containers. The test ensures that the search methods work as expected, can work with any entity, but in here only the nodes are tested.

Additionally, it updates the header and source files, adds comments to explain the methods, and makes some minor code modifications for consistency.

This is a transition PR, before adding MPI support to the search (basically exposing these search strategies to python and using the testing utilities for serial/MPI in Kratos core simplifies a unified testing framework).

**🆕 Changelog**

- Added `SpecializedSpatialSearch` class for efficient searches in spatial containers.
- Specialized implementations for `KDTree`, `Octree`, `BinsStatic`, and `BinsDynamic` containers.
- [Introduced checks for consistent input container sizes and radius arrays.](https://github.com/KratosMultiphysics/Kratos/pull/11000/commits/8cb03ae54072240044b83acfb52c73a2726d993a), see #11001
- Implemented search methods for elements, nodes, and conditions with inclusive and exclusive options.
- Added a Python test to verify the functionality and accuracy of the search methods.


kratos/spatial_containers/specialized_spatial_search.h
```diff
@@ -0,0 +1,807 @@
...
+    {
+        UpdatePoint();
+    }
+
+    ///Copy constructor  (not really required)
+    PointObject(const PointObject& rRHS):
+        BaseType(rRHS),
+        mpObject(rRHS.mpObject)
+    {
+    }
```
matekelemen: unnecessary

kratos/spatial_containers/specialized_spatial_search.h
```diff
@@ -0,0 +1,807 @@
...
+
+    /// Destructor.
+    ~PointObject() override= default;
```
matekelemen: unnecessary

kratos/spatial_containers/specialized_spatial_search.h
```diff
@@ -0,0 +1,807 @@
...
+    /// Pointer definition of SpecializedSpatialSearch
+    KRATOS_CLASS_POINTER_DEFINITION(SpecializedSpatialSearch);
+
+    /// The base type
+    using BaseType = SpatialSearch;
+
+    /// Common Defines
+    using PointType = BaseType::PointType;
+
+    using ElementsContainerType = BaseType::ElementsContainerType;
```
matekelemen:
```suggestion
    using BaseType::ElementsContainerType;
```

valid for all other typedefs that redeclare types from the base class as well

kratos/spatial_containers/specialized_spatial_search.h
```diff
@@ -0,0 +1,807 @@
...
+
+    /**
+     * @brief This method prepares the search
+     * @param rStructure The structure to be searched
+     * @param rInput The input to be searched
+     * @param rResults The results
+     * @param rResultsDistance The results distance
+     * @tparam TContainer The container type
+     * @tparam TResultType The result type
+     */
```
matekelemen, line 369: these kinds of docstrings don't provide any more information than what can already be deduced from type and variable names. I suggest removing it entirely or expanding the explanations and writing proper docs.

kratos/spatial_containers/specialized_spatial_search.h
```diff
@@ -0,0 +1,807 @@
...
+    template<class TContainer, class TResultType>
+    std::vector<typename PointObject<typename TContainer::value_type>::Pointer> PrepareSearch(
```
matekelemen, line 371: why not just collect the points themselves in the vector? Storing the pointers basically reduces the efficiency to that of a linked list.

kratos/spatial_containers/specialized_spatial_search.cpp
```diff
@@ -0,0 +1,353 @@
...
+    using PointTypePointer = typename PointType::Pointer;
+    using PointVector = std::vector<PointTypePointer>;
+    using PointIterator = PointVector::iterator;
+    using DistanceIterator = std::vector<double>::iterator;
+
+    // Defining the PointVector
+    PointVector points = PrepareSearch(rStructureElements, rInputElements, rResults, rResultsDistance);
+
+    // Defining the search structure
+    if constexpr (TSearchBackend == SpatialContainer::KDTree) {
```
matekelemen, line 72: `KDTree` is misleading here because you only support octrees

kratos/spatial_containers/specialized_spatial_search.cpp
```diff
@@ -0,0 +1,353 @@
...
+
+        // Creating the bins
+        DynamicBins dynamic_bins(points.begin(), points.end());
+
+        // Performing search
+        ParallelSearch(rInputElements, rRadius, dynamic_bins, rResults, rResultsDistance);
+    // } else if constexpr (TSearchBackend == SpatialContainer::BinsStaticObjects) {
+    // } else if constexpr (TSearchBackend == SpatialContainer::BinsDynamicObjects) {
+    } else {
+        KRATOS_ERROR << "Unknown search backend" << std::endl;
```
matekelemen: the `else` branch here is a compile time error, so I'd use something like a `static_assert`

kratos/spatial_containers/specialized_spatial_search.h
```diff
@@ -0,0 +1,807 @@
...
+     * @param rRadius The radius array
+     * @param rSearch The spatial search
+     * @param rResults The results
+     * @param rResultsDistance The results distance
+     * @tparam TContainer The container type
+     * @tparam TSpatialContainer The spatial container type
+     * @tparam TResultType The result type
+     */
+    template<class TContainer, class TSpatialContainer, class TResultType>
+    void ParallelSearch(
```
matekelemen, line 418: both `PrepareSearch` and `ParallelSearch` are `private` and only used in a single translation unit, so they'd be better off in the source file to reduce the weight of the header.

kratos/spatial_containers/specialized_spatial_search.h
loumalouomega: Maybe in this PR but can be used later

kratos/spatial_containers/specialized_spatial_search.h
loumalouomega: Idem

kratos/spatial_containers/specialized_spatial_search.h
loumalouomega: Okay

kratos/spatial_containers/specialized_spatial_search.h
loumalouomega: yes, but it is what the spatial container expects

kratos/spatial_containers/specialized_spatial_search.h
loumalouomega: They use a lot of templated arguments

kratos/spatial_containers/specialized_spatial_search.cpp
loumalouomega: That's not true, I support all three used

kratos/spatial_containers/specialized_spatial_search.h
loumalouomega: It is the same doc as in the base class, I can remove it, but the base class is exactly the same

loumalouomega: Done

kratos/spatial_containers/specialized_spatial_search.cpp
loumalouomega: Done

kratos/spatial_containers/specialized_spatial_search.h
loumalouomega: I am going to try, but i don't promise anything

kratos/spatial_containers/specialized_spatial_search.h
loumalouomega: (therefore not my choise)

kratos/spatial_containers/specialized_spatial_search.h
loumalouomega: There are many template combinations, sorry

kratos/spatial_containers/specialized_spatial_search.h
matekelemen: I mean they're auto-generated by the compiler

kratos/spatial_containers/specialized_spatial_search.h
matekelemen: again, auto-generated

kratos/spatial_containers/specialized_spatial_search.h
loumalouomega: Yes, I realized, already removed.

kratos/spatial_containers/specialized_spatial_search.h
matekelemen: they can remain templates, just move them to the source file. Everything that's using them is in that source file.

loumalouomega: Yes, but I need to do the explicit specializations and it is complicated with many combinations

loumalouomega:
@roigcarlo

loumalouomega:
@roigcarlo ping

kratos/python/add_search_strategies_to_python.cpp
```diff
@@ -90,6 +91,8 @@ void AddSearchStrategiesToPython(pybind11::module& m)
         // Get the size of the radius array
         const std::size_t size_array = rListOfRadius.size();
 
+        KRATOS_ERROR_IF(size_array != rModelPart.NumberOfElements()) << "The size of the radius array must be equal to the size of the input elements array" << std::endl;
```
roigcarlo: `>=`  Would be better IMHO. I agree that passing less slots for the results than entities to search is incorrect, but as odd as it can be the other way around is perfectly valid.

In any case if you want to do this check as it is ( exactly same size ) I would add it as a debug assert.

kratos/spatial_containers/specialized_spatial_search.h
```diff
@@ -0,0 +1,707 @@
...
+    ///@}
+}; // Class SpecializedSpatialSearch
+
+/**
+* @class SpecializedSpatialSearchFactory
+* @ingroup KratosCore
+* @brief Factory for the specialized spatial search
+* @author Vicente Mataix Ferrandiz
+*/
+class SpecializedSpatialSearchFactory
```
roigcarlo: Please put this class in a different file.

kratos/spatial_containers/specialized_spatial_search.cpp
```diff
@@ -0,0 +1,353 @@
...
+    VectorResultElementsContainerType& rResults,
+    VectorDistanceType& rResultsDistance
+    )
+{
+    // Defining the point type for the search
+    using PointType = PointObject<Element>;
+    using PointTypePointer = typename PointType::Pointer;
+    using PointVector = std::vector<PointTypePointer>;
+    using PointIterator = PointVector::iterator;
+    using DistanceIterator = std::vector<double>::iterator;
```
roigcarlo: This block dos not depend on anything specific to this function and is used in the others, It should be defined in the class scope and be in the header.

roigcarlo (commented):
Its been a while since I last touched that spatial containers, so take my criticism as a grain of salt.

From what I see seems that the intent of the PR is provide with an unified that extends the current functionality of tree (posting it here for reference) with the functions you said in the OP:

```c++
// TreeNode - Base (abs)
void SearchNearestPoint(PointType const& ThisPoint, PointerType& rResult, CoordinateType& rResultDistance);
void SearchNearestPoint(PointType const& ThisPoint, PointerType& rResult, CoordinateType& rResultDistance, SearchStructureType& Auxiliar);

void SearchInRadius(PointType const& ThisPoint, CoordinateType const& Radius, CoordinateType const& Radius2, IteratorType& Results, DistanceIteratorType& ResultsDistances, SizeType& NumberOfResults, SizeType const& MaxNumberOfResults);
void SearchInRadius(PointType const& ThisPoint, CoordinateType const& Radius, CoordinateType const& Radius2, IteratorType& Results, DistanceIteratorType& ResultsDistances, SizeType& NumberOfResults, SizeType const& MaxNumberOfResults, SearchStructureType& Auxiliar);

void SearchInRadius(PointType const& ThisPoint, CoordinateType const& Radius, CoordinateType const& Radius2, IteratorType& Results, SizeType& NumberOfResults, SizeType const& MaxNumberOfResults);
void SearchInRadius(PointType const& ThisPoint, CoordinateType const& Radius, CoordinateType const& Radius2, IteratorType& Results, SizeType& NumberOfResults, SizeType const& MaxNumberOfResults, SearchStructureType& Auxiliar);

void SearchInBox(PointType const& SearchMinPoint, PointType const& SearchMaxPoint, IteratorType& Results, SizeType& NumberOfResults, SizeType const& MaxNumberOfResults );
```

Because of this there are some changes that I would made (apart from the things I pointed out in the code):

I understand that you want to exploit the use of `if constexpr` to avoid template specialization, but in my opinion this is not the correct approach here. As pointed above, it happens that all spatial search classes in Kratos derive from `TreeNode` (at least the ones used here). There is no need to use template or put the header in the source if you just create a pointer to a `TreeNode` holder to point to the specific search structure you want to use and you pass it to the constructor. With some lambda magic we can delegate the constructor of the specific structure until we have the data it needs to be constructed, which completely removes the need for templates and lets you move all the implementation to the source:

```C++
using BackendBuilder = std::function<TreeNode::Pointer(PointVector, PointVector, int)>;

SpecializedSpatialSearch(Parameters ThisParameters, BackendBuilder)
        : mParameters(ThisParameters), mBackendBuilder(SearchBackend)
    {
        ...
    }

private:

    BackendBuilder mBackendBuilder;

```

Then we delegate the construction of the search structure
```c++
using BackendBuilder = std::function<TreeNode::Pointer(PointVector, PointVector, int)>;

void SpecializedSpatialSearch::SearchConditionsInRadiusExclusive(
    const ConditionsContainerType& rStructureConditions,
    const ConditionsContainerType& rInputConditions,
    const RadiusArrayType& rRadius,
    VectorResultConditionsContainerType& rResults,
    VectorDistanceType& rResultsDistance,
    )
{
    // Defining the PointVector
    PointVector points = PrepareSearch(rStructureConditions, rInputConditions, rResults, rResultsDistance);

    // Retrieving parameters
    const int bucket_size = mParameters["bucket_size"].GetInt();

    // Construct the builder backend
    auto search_backend = mBackendBuilder(points.begin(), points.end(), bucket_size);

    // Performing search
    ParallelSearch(rInputConditions, rRadius, search_backend, rResults, rResultsDistance);
}
```

Then its super easy to expose to python, or use it directly in c++ (you can retrieve the external lambda using an aux function in pure C++ for example) exploiting the fact that we are returning a TreeNode::Pointer with an unified interface
```c++
   py::class_<SpecializedSpatialSearch, SpecializedSpatialSearch::Pointer, SpatialSearch>(m, "SpatialSearchKDTree")
    .def(py::init([](Parameters mParameters){
        return std::unique_ptr<SpecializedSpatialSearch>(mParameters, [](PointVector::iterator PointsBegin, PointVector::iterator PointsEnd, std::size_t BucketSize) {
            // KDtree definitions
            using BucketType = Bucket< 3ul, PointType, PointVector, PointTypePointer, PointIterator, DistanceIterator>;
            using KDTree = Tree<KDTreePartition<BucketType>>;

            return std:::shared_ptr<KDTree>(PointsBegin, PointsEnd, BucketSize);
        });
    }))
    .def(py::init<Parameters>())
    ;
```
_Disclaimer: Coded it by heart, may contain errors._

Note that this way we could even get rid of the `search_structure.h` and `.cpp`(sadly not my proudest contribution to Kratos)

kratos/python/add_search_strategies_to_python.cpp
loumalouomega: There is a PR with this change, I will do the change there.

loumalouomega: See https://github.com/KratosMultiphysics/Kratos/pull/11001

loumalouomega:
> Note that this way we could even get rid of the `search_structure.h` and .cpp(sadly not my proudest contribution to Kratos)

Huuum, dependencies are very deep, if I try to remove it breaks many things.

loumalouomega:
I tried to use your approach, but I don't see the advantage, the code is not actually reduced as you are now doing the changes in the python binding. To avoid the template argument?

loumalouomega:
> My point is that the code already takes superlong to compile, we should use templates where performance is key, and here its only used in the creation of the proper interface, not to speed up the execution.
> 
> We are sacrificing a little performance in the creation for compiling 1 class vs N, where (as you said) N can grow over time if we add more classes. (you are making the code of the class belong to the core as well, which is good)
> 
> Also, in my opinion if you move the complexity of creating the individual instance to the interface layer (even if it remains the same code length), the base clase remains much more clean and easy to modify. For example if it happens that you change the `TreeNode` interface you would need to propagate the changes to N places in the code.
> 
> That being said, the whole interface for spatial_containers is very messy, we propably should arrange a meeting specifically to discuss how we are going to proceed with it.

Understood

loumalouomega:
After speaking with @roigcarlo, he will try to do the suggested changes itself, due to implementations problems from my side

roigcarlo:
Yep, sadly it's impossible right now, but I would like to try a couple of things if you don't mibd and this is not urgent

loumalouomega:
> Yep, sadly it's impossible right now, but I would like to try a couple of things if you don't mibd and this is not urgent

Feel free to suggest changes

loumalouomega:
> > Yep, sadly it's impossible right now, but I would like to try a couple of things if you don't mibd and this is not urgent
> 
> Feel free to suggest changes

I already simplified the code a bit

roigcarlo (approved):
I've been going through it and basically, as it is right now, what you did is in fact the best possible solution. 

That being said, I still thing we have to ret-think the whole spatial search in general.

Thx for trying for trying my suggestions even if was not possible

PR #11001, opened 2023-04-13
[Core] Add missing radius array validation and creation to search strategies in Python bindings

loumalouomega:
**📝 Description**

It includes validation of the size of the radius array and ensures it is equal to the size of the input elements, nodes, or conditions array, depending on the context. KRATOS_ERROR_IF statements have been added to perform these checks, and if the sizes do not match, an error message is printed.

**🆕 Changelog**

- [Add missing radius array validation and creation to search strategies in Python bindings](https://github.com/KratosMultiphysics/Kratos/commit/c9bb63c1a238e15707dfdf35d5c67f697ecb1a13)


matekelemen:
why don't you perform the check directly in `CopyRadiusArrayToPython`?

loumalouomega:
> why don't you perform the check directly in `CopyRadiusArrayToPython`?

Because I did not realize that :P

loumalouomega:
> > why don't you perform the check directly in `CopyRadiusArrayToPython`?
> 
> Because I did not realize that :P

Okay, you can do it, because you check against the size of the vector of entities, so it would not work

loumalouomega:
Ping @matekelemen @roigcarlo

roigcarlo (approved):

PR #11014, opened 2023-04-17
[Core] Convection process usability fixes

rubenzorrilla:
**📝 Description**
@uxuech discovered a bug when trying to use two instances of the `LevelSetConvectionProcess` at the same time. The key aspect of these two instances is that the unknown and convection variables (potentially the unknown gradient one as well) are different and the fact that the two instances come from the same base model part. 

Considering that both instances come from the same base model part, the `ProcessInfo` container is the same in all `LevelSetConvectionProcess` instances, meaning that the pointer to the `CONVECTION_DIFFUSION_SETTINGS` is also the same. Accordingly, we need to update the unknown, unknown gradient and convection variables at each call of the `Execute` in order to reset the content of the `CONVECTION_DIFFUSION_SETTINGS`, which will be later on used by the element.

This was being done only for the unknown variable but not for the others. Besides, the way we used to do this sort of operation was quite messy which in my opinion resulted in the current bug. To avoid this, I've created a `private` auxiliary container called `ProcessInfoDataContainer` that handles this by saving the input data and restoring it after the convection. Besides, I renamed the private `GetFillProcessInfoFunction` to `GetFillProcessInfoFormulationDataFunction` to make clear that this is the element-related data and distinguish it from the new class.

@ddiezrod I think this shouldn't affect nothing in your side. Besides, the `ProcessInfoDataContainer` can be eventually customized in your side by using a `using` statement.
@uxuech I noted that in the nodal tau case we were using the input delta time instead of the substep one. I think this was by accident. I took the chance to change it to the convection substep one, which should be in principle less diffusive.

### DIFF
--- a/kratos/processes/levelset_convection_process.h
+++ b/kratos/processes/levelset_convection_process.h
@@ -69,6 +69,37 @@ template< unsigned int TDim, class TSparseSpace, class TDenseSpace, class TLinea
 class LevelSetConvectionProcess
     : public Process
 {
+private:
+    ///@name Type Definitions
+    ///@{
+
+    class ProcessInfoDataContainer
+    {
+    public:
+        ProcessInfoDataContainer(const ProcessInfo& rInputProcessInfo)
+        : DeltaTime(rInputProcessInfo.GetValue(DELTA_TIME))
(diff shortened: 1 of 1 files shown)


uxuech (approved):
Thanks!

ddiezrod (approved):
Makes sense to me, it should not break anything on our side.

PR #11025, opened 2023-04-17
Adding a new two fluid solver for hydraulic application

uxuech:
**📝 Description**
This PR introduces a new two-fluid solver with a hydraulic application, and includes several key differences from the previous "two_fluid_hydraulic_solver":

- The addition of an option to implement **_artificial viscosity_** based on the residual, which improves robustness and eliminates potential numerical noise.

- A correction is applied to the **_historical velocity variable_** before solving the N.S equations, achieved by solving a convection problem where the velocity variable from the previous step is convected (**eulerian fm ale problem**), thus preserving **energy**.

- The addition of an option to calculate the **_stabilization_** parameter **_tau_** for the convection element _**nodally,**_ which prevents sudden jumps in the  value of the tau between elements.

In addition, a new test has been added to verify all changes made in both 2D and 3D of the **_sloshing_** benchmark.


applications/FluidDynamicsApplication/python_scripts/python_solvers_wrapper_fluid.py
```diff
@@ -36,6 +36,9 @@ def CreateSolverByParameters(model, solver_settings, parallelism):
 
         elif solver_type == "two_fluids" or solver_type == "TwoFluids":
             solver_module_name = "navier_stokes_two_fluids_solver"
+        
+        elif solver_type == "two_fluids_hydraulic" or solver_type == "TwoFluidsHydraulic":
```
philbucher:
```suggestion
        elif solver_type == "two_fluids_hydraulic":
```

please follow the style guide regarding naming for new additions

ddiezrod: I think the correct name should be "two_fluid_hydraulic"

ddiezrod:
@rubenzorrilla  @uxuech Is there any plan to unify the developments here with the  navier stokes two fluid solver? I understand that separating both solvers will accelerate the implementation but I would like to have only 1 solver....

rubenzorrilla:
> @rubenzorrilla @uxuech Is there any plan to unify the developments here with the navier stokes two fluid solver? I understand that separating both solvers will accelerate the implementation but I would like to have only 1 solver at some point....

First, I'd like to note that this is what we tried first.

Answering to your question, not in the short term, an I foresee that not in the mid term either.  The point is that while developing we realized that @uxuech 's target application needs various specifics (mass conservation algorithm, history convection, artificial viscosity, etc.) that would result in an (even more) cumbersome implementation in current solver. 

Besides, I think that the defaults are to be rather different in both cases. Note that current 2-fluid solver is being used for quite different stuff (yours, microfluidics, etc.) while this one is intended to be applied large scale hydraulic applications.

ddiezrod:
Thanks for the answer @rubenzorrilla I am afraid that the whole structure might be a little misleading (for example I think my casting filling examples will have more in common with this hydraulic solver than with some of the more recent developments in the two fluid solver, mostly related to bubble formation etc). Besides the code duplication will be a little of a headache. This being said, I understand that we have to be flexible until everything is properly tested and stabilized.

rubenzorrilla:
> Thanks for the answer @rubenzorrilla I am afraid that the whole structure might be a little misleading (for example I think my casting filling examples will have more in common with this hydraulic solver than with some of the more recent developments in the two fluid solver, mostly related to bubble formation etc). Besides the code duplication will be a little of a headache. This being said, I understand that we have to be flexible until everything is properly tested and stabilized, so for me it is ok to add this new solver.

Thinking in the long run, the best solution could be to have a base "do nothing" two-fluid solver with all the ingredients and different derived solvers re-implementing the defaults and key functions.

applications/FluidDynamicsApplication/python_scripts/python_solvers_wrapper_fluid.py
uxuech: Done!

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+                "materials_filename": ""
+            },
+            "distance_reading_settings"    : {
+                "import_mode"         : "from_mdpa",
+                "distance_file_name"  : "no_distance_file"
+            },
+            "maximum_iterations": 7,
+            "echo_level": 0,
+            "time_order": 2,
+            "time_scheme": "NEEDS TO BE FIXED --> PROBLEMS CONVERGENCE CRITERIO",
```
rubenzorrilla:
```suggestion
            "time_scheme": "NEEDS TO BE FIXED --> PROBLEMS CONVERGENCE CRITERIO",
```
?

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+            "material_import_settings": {
+                "materials_filename": ""
+            },
+            "distance_reading_settings"    : {
+                "import_mode"         : "from_mdpa",
+                "distance_file_name"  : "no_distance_file"
+            },
+            "maximum_iterations": 7,
+            "echo_level": 0,
+            "time_order": 2,
```
rubenzorrilla: As far as I remember, we only support the alpha scheme in here. If so, I'd remove it.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+                "input_filename": "unknown_name",
+                "reorder": false
+            },
+            "material_import_settings": {
+                "materials_filename": ""
+            },
+            "distance_reading_settings"    : {
+                "import_mode"         : "from_mdpa",
+                "distance_file_name"  : "no_distance_file"
+            },
```
rubenzorrilla: Are you using this?

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+            "assign_neighbour_elements_to_conditions": true,
+            "no_skin_parts":[""],
+            "time_stepping"                : {
+                "automatic_time_step" : true,
+                "CFL_number"          : 1,
+                "minimum_delta_time"  : 1e-2,
+                "maximum_delta_time"  : 1.0,
+                "time_step"           : 0.0
+            },
+            "periodic": "periodic",
```
rubenzorrilla: What is this?

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+                "avoid_almost_empty_elements": false,
+                "deactivate_full_negative_elements": false
+            }
+        }""")
+
+        default_settings.AddMissingParameters(super(NavierStokesTwoFluidsHydraulicSolver, cls).GetDefaultParameters())
+        return default_settings
+
+    def __init__(self, model, custom_settings):
+        # TODO: DO SOMETHING IN HERE TO REMOVE THE "time_order" FROM THE DEFAULT SETTINGS BUT KEEPING THE BACKWARDS COMPATIBILITY
```
rubenzorrilla: Not sure if it applies but you don't need to keep backwards compatibility anymore.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+        KratosMultiphysics.Logger.PrintInfo(self.__class__.__name__, "Construction of NavierStokesTwoFluidsHydraulicSolver finished.")
+
+    def AddVariables(self):
+        self.main_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.DENSITY)
+        self.main_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.DYNAMIC_VISCOSITY)
+        self.main_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.PRESSURE)
+        self.main_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.VELOCITY)
+        self.main_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.ACCELERATION)
+        self.main_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.MESH_VELOCITY)
+        self.main_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.IS_STRUCTURE)
```
rubenzorrilla: Why do we need `IS_STRUCTURE`?

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+        self.main_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.BODY_FORCE)
+        self.main_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.NODAL_H)
+        self.main_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.NODAL_AREA)
+        self.main_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.REACTION)
+        self.main_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.REACTION_WATER_PRESSURE)
+        self.main_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.NORMAL)
+        self.main_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.EXTERNAL_PRESSURE)
+        self.main_model_part.AddNodalSolutionStepVariable(KratosMultiphysics.FLAG_VARIABLE)
```
rubenzorrilla: Same. Do we need this one?

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+        default_settings.AddMissingParameters(super(NavierStokesTwoFluidsHydraulicSolver, cls).GetDefaultParameters())
+        return default_settings
+
+    def __init__(self, model, custom_settings):
+        # TODO: DO SOMETHING IN HERE TO REMOVE THE "time_order" FROM THE DEFAULT SETTINGS BUT KEEPING THE BACKWARDS COMPATIBILITY
+
+        if custom_settings.Has("levelset_convection_settings"):
+            if custom_settings["levelset_convection_settings"].Has("levelset_splitting"):
+                custom_settings["levelset_convection_settings"].RemoveValue("levelset_splitting")
+                KratosMultiphysics.Logger.PrintWarning("NavierStokesTwoFluidsHydraulicSolver", "\'levelset_splitting\' has been temporarily deactivated. Using the standard levelset convection with no splitting.")
```
rubenzorrilla: If not supported I'd remove it.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+
+        #TODO: Remove this after the retrocompatibility period
+        if custom_settings.Has("bfecc_convection"):
+            KratosMultiphysics.Logger.PrintWarning("NavierStokesTwoFluidsHydraulicSolver", "the semi-Lagrangian \'bfecc_convection\' is no longer supported. Using the standard Eulerian levelset convection.")
+            custom_settings.RemoveValue("bfecc_convection")
+            if custom_settings.Has("bfecc_number_substeps"):
+                custom_settings.RemoveValue("bfecc_number_substeps")
```
rubenzorrilla, line 126: If not supported I'd remove it.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+
+        super(NavierStokesTwoFluidsHydraulicSolver,self).__init__(model,custom_settings)
+
```
rubenzorrilla:
```suggestion
        super().__init__(model,custom_settings)
```
No longer required as we dropped Python2 support.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+
+        self.element_name = "TwoFluidNavierStokesAlphaMethod"
+        self.rho_inf = 0.0
+        self.main_model_part.ProcessInfo.SetValue(KratosCFD.SPECTRAL_RADIUS_LIMIT, self.rho_inf)
+        self.condition_name = "TwoFluidNavierStokesWallCondition"
+        self.element_integrates_in_time = True
+        self.element_has_nodal_properties = True
+
+        self.min_buffer_size = 3
```
rubenzorrilla: Isn't 2 the buffer required for the alpha scheme?

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+            self.main_model_part.AddNodalSolutionStepVariable(KratosCFD.CONVECTION_VELOCITY)
+            # Auxiliary variable to store the gradient of the historical scalar to be convected
+            self.main_model_part.AddNodalSolutionStepVariable(KratosCFD.CONVECTION_SCALAR_GRADIENT)
+
+
+        KratosMultiphysics.Logger.PrintInfo(self.__class__.__name__, "Fluid solver variables added correctly.")
+
+    def PrepareModelPart(self):
+        # Call the base solver PrepareModelPart()
+        super().PrepareModelPart()
```
rubenzorrilla: If we are only calling the base one we can avoid implementing it.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+        self._SetNodalProperties()
+
+        # Initialize the distance correction process
+        self._GetDistanceModificationProcess().ExecuteInitialize()
+        self._GetDistanceModificationProcess().ExecuteInitializeSolutionStep()
+
+        # Instantiate the level set convection process
+        # Note that is is required to do this in here in order to validate the defaults and set the corresponding distance gradient flag
+        # Note that the nodal gradient of the distance is required either for the eulerian BFECC limiter or by the algebraic element antidiffusivity
+
```
rubenzorrilla:
```suggestion
```

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+        self._GetLevelSetConvectionProcess()
+
+        # Instantiate the eulerian historical data convection
+        if self.eulerian_fm_ale:
+            self._GetEulerianFmAleProcess()
+
+        self.mass_source = False
+        if self.settings["formulation"].Has("mass_source"):
```
rubenzorrilla: `mass_source` must be always present (if the user doesn't provide it the default settings will do).

rubenzorrilla: As far as I see it is not in the defaults.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+            self.mass_source = self.settings["formulation"]["mass_source"].GetBool()
+
+        # Non historical variable are initilized in order to avoid memory problems
+
```
rubenzorrilla:
```suggestion
```

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+        KratosMultiphysics.VariableUtils().SetNonHistoricalVariableToZero(KratosMultiphysics.ACCELERATION, self.main_model_part.Nodes)
+        if self.artificial_viscosity:
+            KratosMultiphysics.VariableUtils().SetNonHistoricalVariableToZero(KratosCFD.ARTIFICIAL_DYNAMIC_VISCOSITY, self.main_model_part.Elements)
+
+        #FIXME: Check that this is required. Note that right now the convection process is the one computing NODAL_H if nodal tau is active
+        find_nodal_h = KratosMultiphysics.FindNodalHNonHistoricalProcess(self.GetComputingModelPart())
+        find_nodal_h.Execute()
```
rubenzorrilla: Did we check this?

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+
+        KratosMultiphysics.Logger.PrintInfo(self.__class__.__name__, "Solver initialization finished.")
+
+
+    def InitializeSolutionStep(self):
+
+        # Inlet and outlet water discharge is calculated for current time step, first discharge and the considering the time step inlet and outlet volume is calculated
+        if self.mass_source:
+            self._ComputeInitialStepWaterVolume()
```
rubenzorrilla:
```suggestion
            self._ComputeStepInitialWaterVolume()
```
Very minor, but InitialStep would refer to the first step. You mean current step initial volume.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+        # Update the DENSITY and DYNAMIC_VISCOSITY values according to the new level-set
+        self._SetNodalProperties()
+
+        # Accumulative water volume error ratio due to level set. Adding source term
+        self._ComputeVolumeError()
+
+        # Perform the convection of the historical database (Eulerian FM-ALE)
+        if self.eulerian_fm_ale:
+            self._PerformEulerianFmAleVelocity()
+
```
rubenzorrilla: As this step might be computationally demanding as the levelset convection maybe we can put a `Logger.PrintInfo` message as we do in there.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+        # Finalize the solver current step
+        self._GetSolutionStrategy().FinalizeSolutionStep()
+
+        # Acceleration for generalised alpha time integration method.
+        self.__CalculateTimeIntegrationAcceleration()
+
+        # Artificial viscosity for shock capturing.
+        if self.artificial_viscosity:
+            self.__CalculateArtificialViscosity()
+
```
rubenzorrilla:
```suggestion
```

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+            KratosMultiphysics.Logger.PrintInfo(self.__class__.__name__, "Redistancing process is finished.")
+
+        # Prepare distance correction for next step
+        self._GetDistanceModificationProcess().ExecuteFinalizeSolutionStep()
+
+        # Finalize the solver current step
+        self._GetSolutionStrategy().FinalizeSolutionStep()
+
+        # Acceleration for generalised alpha time integration method.
+        self.__CalculateTimeIntegrationAcceleration()
```
rubenzorrilla: I might be wrong, but shouldn't we do this in the `FinalizeSolutionStep`? I think that in the no-predict case as we do now it makes no difference but wouldn't be the case if we add a prediction.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+        inlet_discharge = KratosCFD.FluidAuxiliaryUtilities.CalculateFlowRateNegativeSkin(self.GetComputingModelPart(), KratosMultiphysics.INLET)
+        current_dt = self.main_model_part.ProcessInfo[KratosMultiphysics.DELTA_TIME]
+        inlet_volume = -current_dt * inlet_discharge
+        outlet_volume = current_dt * outlet_discharge
+
+        # System water volume is calculated for current time step considering inlet and outlet discharge.
+        system_volume = inlet_volume + self.__initial_system_volume - outlet_volume
+        self.__initial_system_volume = system_volume
+
+
```
rubenzorrilla:
```suggestion
```

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+        if self.mass_source:
+            water_volume_after_transport = KratosCFD.FluidAuxiliaryUtilities.CalculateFluidNegativeVolume(self.GetComputingModelPart())
+
+            volume_error = (water_volume_after_transport -  self.__initial_system_volume) /  self.__initial_system_volume
+        else:
+            volume_error=0
+
+        self.main_model_part.ProcessInfo.SetValue(KratosCFD.VOLUME_ERROR, volume_error)
+
+
```
rubenzorrilla:
```suggestion
```

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+        alpha_f=1/(1+self.rho_inf)
+        gamma= 0.5+alpha_m-alpha_f
+        dt=self.main_model_part.ProcessInfo[KratosMultiphysics.DELTA_TIME]
+        for node in self.GetComputingModelPart().Nodes:
+            vn=node.GetSolutionStepValue(KratosMultiphysics.VELOCITY,1)
+            v = node.GetSolutionStepValue(
+                KratosMultiphysics.VELOCITY, 0)
+            acceleration_n=node.GetValue(KratosMultiphysics.ACCELERATION)
+            acceleration_alpha_method=(v-vn)/(gamma*dt)+((gamma-1)/gamma)*acceleration_n
+            node.SetValue(KratosMultiphysics.ACCELERATION,acceleration_alpha_method)
```
rubenzorrilla: Not for this PR, but we should consider implementing this in c++.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+
+        properties_1 = self.main_model_part.Properties[1]
+        water_dynamic_viscosity_max = self.artificial_limiter_coefficient * properties_1.GetValue(KratosMultiphysics.DYNAMIC_VISCOSITY)
+        for element in self.GetComputingModelPart().Elements:
+            artificial_viscosity_gauss = element.CalculateOnIntegrationPoints(KratosCFD.ARTIFICIAL_DYNAMIC_VISCOSITY, self.main_model_part.ProcessInfo)
+            item = 0
+            elem_artificial_viscosity= 0.0
+            for item in artificial_viscosity_gauss:
+                elem_artificial_viscosity += item
+                item +=1
```
rubenzorrilla: I think this is wrong, you're incrementing the iterator of the loop.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+            elem_artificial_viscosity /= item
+            if elem_artificial_viscosity > water_dynamic_viscosity_max:
+                elem_artificial_viscosity = water_dynamic_viscosity_max
+            element.SetValue(KratosCFD.ARTIFICIAL_DYNAMIC_VISCOSITY, elem_artificial_viscosity)
+
+    def _PerformLevelSetConvection(self):
```
rubenzorrilla:
```suggestion
    def __PerformLevelSetConvection(self):
```
I'd mangle this one.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+        # Solve the levelset convection problem
+        self._GetLevelSetConvectionProcess().Execute()
+
+    def _PerformEulerianFmAleVelocity(self):
```
rubenzorrilla:
```suggestion
    def __PerformEulerianFmAleVelocity(self):
```
I'd mangle this one.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+        # Solve the historical data convection problem
+
+        domain_size = self.main_model_part.ProcessInfo[KratosMultiphysics.DOMAIN_SIZE]
+        velocity_components = [KratosMultiphysics.VELOCITY_X,KratosMultiphysics.VELOCITY_Y,KratosMultiphysics.VELOCITY_Z]
+        mesh_var = [KratosMultiphysics.MESH_VELOCITY_X,KratosMultiphysics.MESH_VELOCITY_Y,KratosMultiphysics.MESH_VELOCITY_Z]
+
+        KratosMultiphysics.VariableUtils().CopyModelPartNodalVar(KratosMultiphysics.VELOCITY,self.eulerian_convection_var, self.main_model_part, self.main_model_part, 0, 0)
+        #TODO: These 2 are not needed as the CloneSolutionStep does it
+        KratosMultiphysics.VariableUtils().CopyModelPartNodalVar(KratosMultiphysics.VELOCITY,self.eulerian_convection_var, self.main_model_part, self.main_model_part, 1, 1)
+        KratosMultiphysics.VariableUtils().CopyModelPartNodalVar(KratosMultiphysics.VELOCITY,self.eulerian_convection_var, self.main_model_part, self.main_model_part, 2, 2)
```
rubenzorrilla: Maybe I'm saying something stupid, but do we really need to copy third position of the buffer?

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+        if not hasattr(self, '_levelset_linear_solver'):
+            self._levelset_linear_solver = self._CreateLinearSolver() # TODO: add customized configuration
+        return self._levelset_linear_solver
+
+    #TODO: Leave it like this to remove it when its working
+    def _GetEulerianFmAleLinearSolver(self):
+        # A linear solver configured specifically for the level-set convection process
+        if not hasattr(self, '_eulerian_fm_ale_linear_solver'):
+            self._eulerian_fm_ale_linear_solver = self._CreateLinearSolver() # TODO: add customized configuration
+        return self._eulerian_fm_ale_linear_solver
```
rubenzorrilla: I think we can now have a unique linear solver for the convections.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+
+    def _GetDistanceGradientProcess(self):
+        if not hasattr(self, '_distance_gradient_process'):
+            self._distance_gradient_process = self._CreateDistanceGradientProcess()
+        return self._distance_gradient_process
+
+    def _GetConsistentNodalPressureGradientProcess(self):
+        if not hasattr(self, '_consistent_nodal_pressure_gradient_process'):
+            self._consistent_nodal_pressure_gradient_process = self._CreateConsistentNodalPressureGradientProcess()
+        return self._consistent_nodal_pressure_gradient_process
```
rubenzorrilla: I think you're not using this one.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+
+    def _GetDistanceReinitializationProcess(self):
+        if not hasattr(self, '_distance_reinitialization_process'):
+            self._distance_reinitialization_process = self._CreateDistanceReinitializationProcess()
+        return self._distance_reinitialization_process
+
+    def _GetDistanceGradientProcess(self):
+        if not hasattr(self, '_distance_gradient_process'):
+            self._distance_gradient_process = self._CreateDistanceGradientProcess()
+        return self._distance_gradient_process
```
rubenzorrilla: I think you're not using this one.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+        return distance_reinitialization_process
+
+    def _CreateDistanceGradientProcess(self):
+        distance_gradient_process = KratosMultiphysics.ComputeNodalGradientProcess(
+                self.main_model_part,
+                self._levelset_variable,
+                self._levelset_gradient_variable,
+                KratosMultiphysics.NODAL_AREA)
+
+        return distance_gradient_process
```
rubenzorrilla: Same comment. I think you're not using this one.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
```diff
@@ -0,0 +1,692 @@
...
+
+    def _CreateConsistentNodalPressureGradientProcess(self):
+        consistent_nodal_pressure_gradient_process = KratosCFD.CalulateLevelsetConsistentNodalGradientProcess(
+                self.main_model_part)
+
+        return consistent_nodal_pressure_gradient_process
```
rubenzorrilla: I think you're not using this one.

applications/FluidDynamicsApplication/python_scripts/python_solvers_wrapper_fluid.py
```diff
@@ -37,6 +37,9 @@ def CreateSolverByParameters(model, solver_settings, parallelism):
         elif solver_type == "two_fluids" or solver_type == "TwoFluids":
             solver_module_name = "navier_stokes_two_fluids_solver"
 
+        elif solver_type == "two_fluid_hydraulic":
+            solver_module_name = "navier_stokes_two_fluids_hydraulic_solver"
```
rubenzorrilla, line 41:
```suggestion
            solver_module_name = "navier_stokes_two_fluid_hydraulic_solver"
```

applications/FluidDynamicsApplication/tests/two_fluid_hydraulic_test.py
```diff
@@ -0,0 +1,200 @@
...
+import KratosMultiphysics.KratosUnittest as UnitTest
+import KratosMultiphysics.kratos_utilities as KratosUtilities
+
+class FluidDynamicsAnalysisHydraulic(FluidDynamicsAnalysis):
+
+    def __init__(self,model,project_parameters,flush_frequency=10.0):
+        super().__init__(model,project_parameters)
+
+        self.flush_frequency = flush_frequency
+        self.last_flush = time.time()
```
rubenzorrilla, line 18: You don't need flushing for the test.
```suggestion
```

applications/FluidDynamicsApplication/tests/two_fluid_hydraulic_test.py
```diff
@@ -0,0 +1,200 @@
...
+                node.SetSolutionStepValue(KratosMultiphysics.DISTANCE,d0)
+                node.SetSolutionStepValue(KratosMultiphysics.DISTANCE,1,d0)
+
+    def FinalizeSolutionStep(self):
+        super().FinalizeSolutionStep()
+        if self.parallel_type == "OpenMP":
+            now = time.time()
+            if now - self.last_flush > self.flush_frequency:
+                sys.stdout.flush()
+                self.last_flush = now
```
rubenzorrilla: Same comment. You don't need flushing for the test.
```suggestion
```

applications/FluidDynamicsApplication/tests/two_fluid_hydraulic_test.py
```diff
@@ -0,0 +1,200 @@
+import KratosMultiphysics
+import KratosMultiphysics.FluidDynamicsApplication
+from KratosMultiphysics.FluidDynamicsApplication.fluid_dynamics_analysis import FluidDynamicsAnalysis
+
+import math
+import time
+import sys
```
rubenzorrilla, line 7: Not required.
```suggestion
```

applications/FluidDynamicsApplication/tests/two_fluid_hydraulic_test.py
```diff
@@ -0,0 +1,200 @@
...
+            now = time.time()
+            if now - self.last_flush > self.flush_frequency:
+                sys.stdout.flush()
+                self.last_flush = now
+
+# Class derived from the UnitTest (KratosMultiphysics.KratosUnittest) class
+class TwoFluidHydraulicSolverTest(UnitTest.TestCase):
+
+    def setUp(self):
+        self.work_folder = "TwoFluidMassHydraulicTest"
```
rubenzorrilla:
```suggestion
        self.work_folder = "two_fluid_hydraulic_solver_test"
```
Please follow filename convention.

rubenzorrilla: Besides, I think that you're testing more stuff than the mass.

rubenzorrilla (changes requested):
There are some concerns to be amended.

Besides, from my understanding I think you should have one results file per test . However there are only two. Regarding the mdpa(s), do we need them to be that fine? Couldn't we use a coarser mesh?

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
uxuech: No, you are right it is not necessary.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
uxuech: Yes, now it is removed

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
uxuech: Now it is not necessary to fix since we are only using alpha time integration scheme.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
uxuech: This is the option to turn on/off the artificicial viscosity in order to avoid numerical noise. In addition there is a limiter in case we want to limit the maximum amount of artificial viscosity

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
uxuech: You are right!removed

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
uxuech: It should not be there. Removed

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
uxuech: No, You are right

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
uxuech: Oups, Yes, now it is modified.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
uxuech: Ok, thanks

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
uxuech: Done

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
uxuech: It was not checked. Now it is and it is no neccesary.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
uxuech: Perfect. Thanks for the suggestion

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
uxuech: Yes, I agree.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
uxuech: Sure, Now it is solved.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
uxuech: No, you are right, It was done for a check during the implementation.

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
uxuech: After a discussion we decided to do it in the FinalizeSolutiolStep

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluids_hydraulic_solver.py
uxuech: You are right!

applications/FluidDynamicsApplication/python_scripts/navier_stokes_two_fluid_hydraulic_solver.py
```diff
@@ -0,0 +1,659 @@
...
+        if distance_modification_settings["deactivate_full_negative_elements"].GetBool():
+            distance_modification_settings["deactivate_full_negative_elements"].SetBool(False)
+            KratosMultiphysics.Logger.PrintWarning("Provided distance correction \'deactivate_full_negative_elements\' is \'True\'. Setting to \'False\' to avoid deactivating the negative volume (e.g. water).")
+
+        # Create and return the distance correction process
+        return KratosCFD.DistanceModificationProcess(
+            self.model,
+            distance_modification_settings)
+
+    #TODO: Temporary solution for fixed alpha scheme
```
uxuech, line 646: Answering to why the "time_scheme" is present in the default Parameters, it's here. This solver is derived from the fluid_solver, and therefore, we decided to implement this temporary solution to make it work. @rubenzorrilla , do you think we should modify this part of the code?

rubenzorrilla: Considering that this new solver is specifically tailored for your element, which hardcodes an alpha-scheme within the RHS and LHS implementation, I'd leave it as it is. In other words, this solver is only compatible with `ResidualBasedIncrementalUpdateStaticSchemeSlip`.

applications/FluidDynamicsApplication/tests/two_fluid_hydraulic_solver_test/hydraulic_solver_2d_eularian_fm_ale_results.json
```diff
@@ -0,0 +1,600 @@
+{
```
rubenzorrilla: There's a typo in the filename (change `eularian` by `eulerian`.

applications/FluidDynamicsApplication/tests/two_fluid_hydraulic_solver_test/hydraulic_solver_3d_eularian_fm_ale_results.json
```diff
@@ -0,0 +1,1720 @@
+{
```
rubenzorrilla: Same typo in the filename.

rubenzorrilla:
> Regarding the Python tests, I would like to know how much you suggest we should reduce the MDPA (Mesh Data and Parameter Archive). Thank you for your help @rubenzorrilla.

If it works, I'd do a mesh of 5 times 5 divisions (5 times 5 times 5 in 3D). At the end of the day you just want to check one of the very first steps.

rubenzorrilla (approved):
Ready 🚀

PR #11028, opened 2023-04-18
Fixing some tests that fail when mpi size is 1.

jcotela:
**📝 Description**
Some MPI tests are currently failing when run with `mpirun -np 1` and I understand that it is just because these tests are not designed to work that way. I tried to skip or adjust them as appropriate.

**🆕 Changelog**
- Fixed some tests MPI that were failing when run on a single process

### DIFF
--- a/kratos/mpi/tests/cpp_tests/sources/test_debug_utilities.cpp
+++ b/kratos/mpi/tests/cpp_tests/sources/test_debug_utilities.cpp
@@ -61,6 +61,7 @@ KRATOS_DISTRIBUTED_TEST_CASE_IN_SUITE(DebugToolsCheckSingleHistoricalVariableFix
 {
     const DataCommunicator& r_comm = Testing::GetDefaultDataCommunicator();
     const int world_size = r_comm.Size();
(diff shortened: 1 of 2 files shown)


roigcarlo (approved):
:+1: Yep I did not design those to work with np 1. (or the class for what it matters)

If you are interested I can modify them to run in serial as well.

philbucher:
Should this be added to the CI?

jcotela:
@roigcarlo if I understood correctly, most failing tests are for an utility that checks for consistency between processes, so for me it makes sense that it auto-succeeds or is skipped if there is only one process.

I found this while working on something else, I just wanted to have a clean test output for further development.

jcotela:
@philbucher I do not know about the current CI setup, but I suspect that this thing of running with -np 1 is a somewhat marginal use case...

I will go ahead and merge, let me know if I can help with an eventual CI modification.

roigcarlo:
@philbucher Na, no need to add. This is only for debug, it does not affect any functionality. As @jcotela said the main purpose of this is to ensure that variable values are in synch at different stages of the execution.

In any case, we should use this to make tests for processes, but not the other way around.

PR #11032, opened 2023-04-19
[Core] Add compiler flag visibility-inlines-hidden to reduce exposed symbols

huhaas:
**📝 Description**
Noticed a problem when building on Windows with msys2 + Clang. 
```
[build] ld.lld: error: too many exported symbols (got 138885, max 65535)
``` 
Adding the compiler flag -fvisibility-inlines-hidden solves the problem. See [link](https://gcc.gnu.org/wiki/Visibility) for explenation.
Because Kratos exposes many functions, it might positively impact compiling time, binary size, and sometimes load times.

**🆕 Changelog**
- Add compiler flag visibility-inlines-hidden to reduce exposed symbols


loumalouomega (approved):
Okay from my side

matekelemen:
as far as I know, we only explicitly export symbols on windows (because MSVC's default policy is not to export anything), so you'll have to define the relevant export macros for GCC and Clang as well:

https://github.com/KratosMultiphysics/Kratos/blob/81deb698f608f8f188cc4e9ea7dfd9464655336d/kratos/includes/kratos_export_api.h#L16-L29

huhaas:
> as far as I know, we only explicitly export symbols on windows (because MSVC's default policy is not to export anything), so you'll have to define the relevant export macros for GCC and Clang as well:
> 
> P.S.: nevermind it's there, but it looks like it isn't correct.

Thanks, will look into this. The PR needs a bit more work :)
If it does not improve anything I will remove it of course.

matekelemen:
> If it does not improve anything I will remove it of course.

I don´t think it´s a bad idea, at least it forces some consistency between windows and linux. Just keep in mind that it´s possible that not all applications are compiled on windows, and so those may have incorrect exporting rules. Not sure about it tho, so I´ll go ahead and invoke @roigcarlo.

huhaas:
Will close this PR for now. Cannot get it to work on all platforms, unfortunately.

PR #11038, opened 2023-04-20
[OptApp]  Extend CollectiveExpressions to support continuous double vectors

sunethwarna:
**📝 Description**
This PR extends the `CollectiveExpressions` to easily evaluate `lazy_expressions` within it to a flat continuous `double` array. 

These `CollectiveExpressions` can be used to hold **different variables** from **different containers** of **different model parts** and
1. In the `Kratos` environment they can utilize the `lazy_expressions` to do computations thinking them as a flat double vector. Once done, the values can be automatically assigned to respective containers. (No additional memory cost)
2. In the external libs, these `lazy_expressions` can be flattened to one continuous `double*` or one numpy array.

Following capabilities are added:

1. Reading from a single variable of type `double`, `array_1d<double, 3>`, `array_1d<double, 4>`, `array_1d<double, 6>`, `array_1d<double, 9>`, `Vector`, `Matrix`
3. Reading from a different variables of type `double`, `array_1d<double, 3>`, `array_1d<double, 4>`, `array_1d<double, 6>`, `array_1d<double, 9>`, `Vector`, `Matrix`
4. Reading from `c-interface` like `double*` for `c/c++`
5. Reading from `numpy` arrays for python (no copying of numpy arrays when converting to `cpython`)
6. Moving of a numpy array data to the `CollectiveExpression`. In this case, the `CollectiveExpression` will not copy any data from the numpy array. So, when the `CollectiveExpression` or its `ContainerExpression` objects are evaluated, they will use the original numpy array. This will seg fault if that original numpy array is destroyed.
7. Evaluates `CollectiveExpression` to given `c-interface` like `double*` for 'c/c++`
8. Evaluates `CollectiveExpression` to a numpy array for python
9. Evaluates `CollectiveExpression` to a single variable of type `double`, `array_1d<double, 3>`, `array_1d<double, 4>`, `array_1d<double, 6>`, `array_1d<double, 9>`, `Vector`, `Matrix`
10. Evaluates `CollectiveExpression` to different variable of type `double`, `array_1d<double, 3>`, `array_1d<double, 4>`, `array_1d<double, 6>`, `array_1d<double, 9>`, `Vector`, `Matrix`

All methods are compatible with OpenMP and MPI.

**🆕 Changelog**
- Added `Read`, `MoveFrom`, `Evaluate` methods to `CollectiveExpressions`
- Added numpy support
- Added tests

### DIFF
--- a/applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
+++ b/applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
@@ -147,7 +147,7 @@ void  AddCustomUtilitiesToPython(pybind11::module& m)
     AddSpecializedContainerExpressionToPython<ModelPart::ElementsContainerType, ContainerDataIOTags::Properties>(sub_module, "ElementPropertiesExpression");
 
     py::class_<CollectiveShapeInfo>(sub_module, "CollectiveShapeInfo")
-        .def(py::init<const pybind11::array_t<int>&, const pybind11::array_t<int>&, const pybind11::array_t<int>&>(), py::arg("flat_dimension_sizes_list"), py::arg("number_of_dimensions_for_each_container"), py::arg("number_of_entities_for_each_container"))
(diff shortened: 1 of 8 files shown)


RezaNajian (approved):
Impressive work! !, thanks @sunethwarna for adding this great feature.

PR #11041, opened 2023-04-20
[Core] Add `IsStringArray` to Kratos `Parameters`

ncrescenzio:
**📝 Description**
This PR adds the method `IsStringArray` in Kratos `Parameters` class. The method is used in `DeprecationManager`.

**🆕 Changelog**
- Added `IsStringArray` to Kratos `Parameters`
- Added python and c++ tests
- Added `IsStringArray` to `DeprecationManager`


loumalouomega (approved):

rubenzorrilla (changes requested):
To be discussed in the @KratosMultiphysics/technical-committee. I'm not sure about the need of extending an already large API. Note that the same can be achieved with the already existent `IsArray`. 

@ncrescenzio we would appreciate the description of your use case justifying the addition.

ncrescenzio:
I am modifying a python process and changing the name of an entry in the input parameters. In the same process the `DeprecationManager` is used to handle the changes in the names of the parameters. However, the `DeprecationManager` does not work when the value is a string array and that's why I decided to add also this case. I know that I can change the name of an entry in the input parameters with few lines of code without using the `DeprecationManager` but I just wanted to be consistent with the remaining part of the process that was using it.

But maybe should I avoid to use the `DeprecationManager`? I have seen that it is not used that much in Kratos, is there a specific reason?

Moreover, I have seen that there is the method `IsArray` but other methods involving string arrays are implemented in `Kratos.Parameters`and also this commented line was suggesting me that `IsStringArray` was missing but could be implemented, so I didn't think that such a change would have been a problem. Maybe I misunderstood the comment, sorry!

https://github.com/KratosMultiphysics/Kratos/blob/216b3d8b472a4210bd9088c51ee79d9676418ef6/kratos/sources/kratos_parameters.cpp#L750

Anyway, this change is not essential for me and I have no problem in closing this PR.

rubenzorrilla:
> I am modifying a python process and changing the name of an entry in the input parameters. In the same process the `DeprecationManager` is used to handle the changes in the names of the parameters. However, the `DeprecationManager` does not work when the value is a string array and that's why I decided to add also this case. I know that I can change the name of an entry in the input parameters with few lines of code without using the `DeprecationManager` but I just wanted to be consistent with the remaining part of the process that was using it.
> 
> But maybe should I avoid to use the `DeprecationManager`? I have seen that it is not used that much in Kratos, is there a specific reason?
> 
> Moreover, I have seen that there is the method `IsArray` but other methods involving string arrays are implemented in `Kratos.Parameters`and also this commented line was suggesting me that `IsStringArray` was missing but could be implemented, so I didn't think that such a change would have been a problem. Maybe I misunderstood the comment, sorry!
> 
> https://github.com/KratosMultiphysics/Kratos/blob/216b3d8b472a4210bd9088c51ee79d9676418ef6/kratos/sources/kratos_parameters.cpp#L750
> 
> Anyway, this change is not essential for me and I have no problem in closing this PR.

OK. Let's leave it like this to be briefly discussed in the KTC. What I'd like to avoid is populating even more the API if not fully needed.

rubenzorrilla (approved):
We've been briefly discussing about this in the @KratosMultiphysics/technical-committee. Considering that there is already a `GetStringArray` and `SetStringArray` it makes complete sense to add the corresponding `IsStringArray`.

My apologies for the delay @ncrescenzio.

kratos/sources/kratos_parameters.cpp
```diff
@@ -581,6 +581,24 @@ bool Parameters::IsArray() const
...
+    if (!mpValue->is_array()) {
+        return false;
+    } else {
+        const auto& r_array = (*mpValue);
+        for (IndexType i = 0; i < mpValue->size(); ++i) {
+            if (!r_array[i].is_string()) {
+                return false;
+            }
+        }
+        return true; // All entries are numbers or Vector is empty
```
rubenzorrilla, line 595:
```suggestion
        return true; // All entries are strings or Vector is empty
```

kratos/tests/cpp_tests/sources/test_kratos_parameters.cpp
```diff
@@ -590,6 +591,12 @@ KRATOS_TEST_CASE_IN_SUITE(KratosParametersIsMethods, KratosCoreFastSuite)
             KRATOS_CHECK_IS_FALSE(tmp[key].IsString());
         }
 
+        if (key.find("s_array") != std::string::npos) {
```
rubenzorrilla, line 594:
```suggestion
        if (key.find("string_array") != std::string::npos) {
```

kratos/tests/cpp_tests/sources/test_kratos_parameters.cpp
```diff
@@ -559,6 +559,7 @@ KRATOS_TEST_CASE_IN_SUITE(KratosParametersIsMethods, KratosCoreFastSuite)
         "double_value": 2.0, // This is comment too, but using another comment
         "bool_value" : true, // This is another comment being meta as realizing that all the possibilities are already check
         "string_value" : "hello",/* This is a nihilist comment about the futile existence of the previous comment as a metacomment */
+        "s_array_value" : ["hello", "world"],
```
rubenzorrilla, line 562:
```suggestion
        "string_array_value" : ["hello", "world"],
```

ncrescenzio: I have used `"s_array_value"` because with `"string_array_value"` the test was failing. Indeed, when `key` is `"string_array_value"`, these lines below are executed and the wrong method is called (thus the test is failing)

https://github.com/KratosMultiphysics/Kratos/blob/5278a666891eaee42f61ac20d2c02aea45ebb5d7/kratos/tests/cpp_tests/sources/test_kratos_parameters.cpp#L588-L589

kratos/sources/kratos_parameters.cpp
ncrescenzio: Sorry, I have seen that the PR was approved and I merged it before you added the comments... Should I open another PR to apply your suggestion?

rubenzorrilla: Yes, lets do it like this. You can even use this very same branch if you want.

ncrescenzio: I have changed the comment in #11200

kratos/tests/cpp_tests/sources/test_kratos_parameters.cpp
ncrescenzio: I have not included this suggestion in #11200 because of my comment above, ok?

PR #11042, opened 2023-04-21
RomApp Easily Storing Matrices as Numpy Arrays in the RomManager

Rbravo555:
**📝 Description**
This PR allows the user to easily store the snapshots matrices of a given training campain by passing to the Fit() method of the RomManager the following parameters:
- store_all_snapshots=False/True
- store_fom_snapshots=False/True
- store_rom_snapshots=False/True
- store_hrom_snapshots=False/True
- store_residuals_projected=False/True
The matrices are stored in a folder named "./SnapshotsMatrices/" in the current directory

**🆕 Changelog**
- Added the parameters with default value to the RomManager.Fit() method 
- Added the method RomManager._StoreSnapshotsMatrix(self,string_numpy_array_name, numpy_array)


SADPR (commented):
Looks good to me.

NicolasSR (commented):
Seems correct to me aswell

applications/RomApplication/python_scripts/rom_manager.py
```diff
@@ -517,5 +535,15 @@ def _GetDefaulHromTrainingParameters(self):
 
 
 
+    def _StoreSnapshotsMatrix(self, string_numpy_array_name, numpy_array):
+
+        # Define the directory and file path
+        directory = './SnapshotsMatrices'
```
rubenzorrilla, line 541: Maybe we should do this user definable?

applications/RomApplication/python_scripts/rom_manager.py
```diff
@@ -517,5 +535,15 @@ def _GetDefaulHromTrainingParameters(self):
...
+        file_path = os.path.join(directory, string_numpy_array_name + '.npy')
```
rubenzorrilla:
```suggestion
        file_path = os.path.join(directory, f'{string_numpy_array_name}.npy')
```

rubenzorrilla (commented):
Two minor details, aside of these changes look good to me.

applications/RomApplication/python_scripts/rom_manager.py
Rbravo555: We shall be careful not to hyper-parametrise the RomManager. That is one of the reasons I have opted not to pass an extra parameter to the RomManager itself, but pass an optional parameter to its Fit() method. 

Having said that, in an upcoming PR which is being prepared by @SADPR(#10990), multiple ProjectParameters.json will be usable, and therefore the path names will have to be provided. If we can leave it hard-coded (although quite evocative) for now, we will address the issue of allowing users to define their own folder names in that PR.

rubenzorrilla: Fair enough.

rubenzorrilla (approved):

PR #11047, opened 2023-04-24
[Core] Model part io add conditions to base model part

ddiezrod:
I was trying to use model part io to read a model part directly into a submodelpart (so for example instead of reading the mdpa insde main_model_part, I put it inside main_model_part/structure. I am getting an error while reading mdpa submodel part conditions saying that "condition does not exist in root model part".
I was checking the code and it seems we are doing different things when reading elements and conditions. For elements , we are calling AddElements function, which push elements to all parent model parts.  However, in the case of conditions we directly push conditions to the ConditionsContainer instead and therefore they are not being added to the parent model part. This usually does not matter because most of the times the input model part is not a submodel part.
I saw that the change for the elements was done a long time ago by @rubenzorrilla , but it was done only for elements, not for the conditions. 
https://github.com/KratosMultiphysics/Kratos/commit/c43752d0e301aa369e7ef3d91790da58d1e42756
In this PR I modify the function for conditions so it does the same in both cases.

### DIFF
--- a/kratos/sources/model_part_io.cpp
+++ b/kratos/sources/model_part_io.cpp
@@ -1959,7 +1959,62 @@ void ModelPartIO::ReadElementsBlock(NodesContainerType& rThisNodes, PropertiesCo
 
 void ModelPartIO::ReadConditionsBlock(ModelPart& rModelPart)
 {
-    ReadConditionsBlock(rModelPart.Nodes(), rModelPart.rProperties(), rModelPart.Conditions());
+    KRATOS_TRY
+
+    SizeType id;
+    SizeType properties_id;
+    SizeType node_id;
+    SizeType number_of_read_conditions = 0;
+
+
+    std::string word;
+    std::string condition_name;
+
+    ReadWord(condition_name);
+    KRATOS_INFO("ModelPartIO") << "  [Reading Conditions : ";
+
+    if(!KratosComponents<Condition>::Has(condition_name))
+    {
+        std::stringstream buffer;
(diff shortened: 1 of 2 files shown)


philbucher (commented):
Instead of copy pasting how about making the function general so that it works for both elements and conditions?

ddiezrod:
> Instead of copy pasting how about making the function general so that it works for both elements and conditions?

Yep, I was thinking about that, but I wanted to know first if there was any reason why this was being done differently for elements and conditions. There is also other two functions that are ver similar so I think we could unify them also after agreeing on this.

ddiezrod:
I found a way to do this without duplicating code, in fact it was pretty obvious. You can call the ReadConditionsBlock function with an auxiliar condition container and then add them to the model part. I did the same change for the elements to remove the duplicated code.

ddiezrod:
ping @philbucher @rubenzorrilla

loumalouomega:
> ping @philbucher @rubenzorrilla

This is good cleaning. We can apply same logic for other blocks (future PR)

philbucher (approved):

philbucher:
there is still a lot of duplicated code no?

rubenzorrilla (commented):
I don't think there was a specific reason for the elements-only. Indeed, most probably I came across with the same problem and I forgot doing it for the conditions as well. No concerns from my side.

ddiezrod:
Thanks @rubenzorrilla and @philbucher , indeed code might be further improved. I will give it a try in a future PR.

PR #11058, opened 2023-04-26
[MetisApplication] Fix bug in synchronous condition partitioning.

### LINKED ISSUE #10938: [Core] MetisDivideHeterogeneousInputProcess assigns conditions wrongly
**Description**

The `MetisDivideHeterogeneousInputProcess` fails to assign conditions correctly in partitioning.

Following are the details of the mdpa:

MDPA (Quads 2D4N, Line2D2N):
Partitioning: grey: rank 0, green: rank 1, red: rank 2
Node ids: In bold underlined pink near nodes
Condition ids: Strike through numbers
Element ids: numbers in the middle of elements

image

I digged through the bug, and found the following:

After metis does partitioning based on element connectivities, we do the following call:

https://github.com/KratosMultiphysics/Kratos/blob/68f678be94b7bac4521dd709c7b5970e0cc3fbfc/applications/MetisApplication/custom_processes/metis_divide_heterogeneous_input_process.cpp#L405-L409
Metis partitioner input for the above line:
```c++
rNodePartition = [0, 0, 2, 0, 0, 0, 0, 0, 2, 2, 2, 0, 2, 2, 2, 1, 1, 1, 2, 1, 1, 1, 1, 1, 2]
rElemPartition = [0, 0, 0, 0, 0, 2, 0, 2, 1, 2, 2, 2, 1, 1, 1, 2]
```

The issue is, if you check the element with id = 4 (index 3 in rElemPartition):
The connectivity node ids: [7, 9, 10, 8]
The connectivity sorted index ids in rNodePartition: [6, 7, 8, 9]
Ranks those node ids belongs to is: [0 ,0 ,2, 2]

Then we call `PartitionConditionsSynchronous`, in there after the following:
https://github.com/KratosMultiphysics/Kratos/blob/68f678be94b7bac4521dd709c7b5970e0cc3fbfc/applications/MetisApplication/custom_processes/metis_divide_heterogeneous_input_process.cpp#L424-L443

the cond_id = 10 (9th index in `rCondPartition`) is assigned to rank = 2 because, all the connectivity nodes in that condition are in rank 2 as shown below. [connectivity ids: [9, 10], connectivity node id indices in `rNodePartition` = [8, 9]]
```c++
rCondPartition = [0, 1, 0, 1, 0, 1, -1, -1, 0, 2, 0, 2, -1, -1, 1, -1]
```

then the following check will be skipped for cond_id = 10:
https://github.com/KratosMultiphysics/Kratos/blob/68f678be94b7bac4521dd709c7b5970e0cc3fbfc/applications/MetisApplication/custom_processes/metis_divide_heterogeneous_input_process.cpp#L450

hence this condition will not be assigned to the same rank where its parent element belongs to.

**Scope**
- Kratos Core

**To Reproduce**
[Archive.zip](https://github.com/KratosMultiphysics/Kratos/files/11109658/Archive.zip)
Extract the above archive zip and run the following command. It only happens with `-np 3`.
```bash
mpirun -np 3 python test_code.py  
```

**Expected behavior**
the cond_id=10 should be in rank 0. If I remove the `if *itPart == -1)` then it correctly assigns the condition to the partition.

**Environment**
- Linux (Manjaro) (I have only tested this in linux)
- Metis 5.1.0
- python 3.10.9
- gcc 12.2.1

jcotela:
**📝 Description**
If the condition should belong to the same partition as the element it is a face of, do not use node counts to assign it a partition.

Fixes #10938 

**🆕 Changelog**
- Bugfix: Ensure that conditions always end up in the same partition as the element they are a face of when partitioning with synchronize_conditions=True.


loumalouomega (dismissed):

philbucher:
can you please add a test?

Ideally the one from #10938

jcotela:
@philbucher good idea, done

applications/MetisApplication/tests/quad_partition_test.py
```diff
@@ -0,0 +1,46 @@
...
+                for condition in model_part.Conditions:
+                    parent = condition.GetValue(FRACTIONAL_STEP)
+                    self.assertTrue(parent in model_part.Elements)
+
+    def tearDown(self):
+        with KratosUnittest.WorkFolderScope(self.work_dir, ''):
+            try:
+                shutil.rmtree(f'{self.test_file}_partitioned')
+            except:
+                pass
```
philbucher: this can be done better:
https://github.com/KratosMultiphysics/Kratos/blob/5a9b687fb858fc9c2bcd8c86c0eadd8d799c5d1a/applications/CoSimulationApplication/tests/co_simulation_test_factory.py#L183

jcotela: Updated to use the cleanUp method, but be careful with the linked example: if you register the cleanUp operation after running the test, they will not run if the test ever throws an error.

applications/MetisApplication/tests/quad_partition_test.py
```diff
@@ -0,0 +1,42 @@
...
+class QuadPartitionTest(KratosUnittest.TestCase):
+
+    def testConditionPartitioning(self):
+        test_file = "quads"
+        partitions = 3
+        sync_conditions = True
+        domain_size = 2
+        verbosity = 0
+
+        work_dir = os.path.abspath(os.path.dirname(__file__))
```
philbucher: I recommend using the pathlib module
e.g. sth like this

```suggestion
        work_dir = Path(__file__).parent.absolute
```

applications/MetisApplication/tests/quad_partition_test.py
```diff
@@ -0,0 +1,42 @@
...
+        model = Model()
+        io_flags = ModelPartIO.READ|ModelPartIO.SKIP_TIMER|ModelPartIO.IGNORE_VARIABLES_ERROR
+
+        # Register clean up operation for files created by the partitioner
+        self.addCleanup(
+            DeleteDirectoryIfExisting,
+            os.path.join(work_dir, f'{test_file}_partitioned'))
```
philbucher: then you can also do
```suggestion
            work_dir /f'{test_file}_partitioned')
```

philbucher (dismissed):
added some comments

philbucher (approved):

PR #11059, opened 2023-04-26
[MPI][TrilinosApplication] Builder and solver hanging node support

### LINKED ISSUE #10994: MPI B&S does not work correctly with "hanging" local nodes
I have found that the global system in trilinos block builder and solver does not work correctly if we have a partition with a hanging local node, i.e. a local node that belongs to an element that is in a different partition. If the node does not belong to any element in its own partition then it is not considered in the global system here
https://github.com/KratosMultiphysics/Kratos/blob/b1958c394fe5d7b99d3c8e807fe1c05c799d4ff3/applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h#L569

In principle, metis makes sure that this never happens, if there is a hanging node it should be ghost, not local. The problem is that in my case the computing model part is a submodelpart from the main model part that is used in the partitioning, so this can happen.

I understand that the way to correct this would be to synchronize dofs so these hanging nodes are also added to mDofSet somewhere here
https://github.com/KratosMultiphysics/Kratos/blob/b1958c394fe5d7b99d3c8e807fe1c05c799d4ff3/applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h#L518

I am not sure how this should be done though, or if this could break other parts of the code or affect performance significantly.

Ping @KratosMultiphysics/technical-committee @KratosMultiphysics/implementation-committee @jcotela @loumalouomega  @philbucher

jcotela:
**📝 Description**
Add support for hanging nodes in TrilinosBlockBuilderAndSolver. For performance, it is disabled by default, enable it by setting `"synchronize_dof_set": true"` on the constructor's json parameters.

Add a new constructor for TrilinosBlockBuilderAndSolver using json parameters.

Fixes #10994 

**🆕 Changelog**
- Add syncrhonize_dof_set option to TrilinosBlockBuilderAndSolver to support hanging nodes.
- Add constructor accepting JSON parameters to TrilinosBlockBuilderAndSolver.


applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
```diff
@@ -128,7 +145,8 @@ class TrilinosBlockBuilderAndSolver
                                   typename TLinearSolver::Pointer pNewLinearSystemSolver)
         : BuilderAndSolver<TSparseSpace, TDenseSpace, TLinearSolver>(pNewLinearSystemSolver),
           mrComm(rComm),
-          mGuessRowSize(GuessRowSize)
+          mGuessRowSize(GuessRowSize),
+          mSynchronizeDofSet(false)
```
philbucher: please use the definition of the member to initialize it, otherwise it needs to be added to every CTor

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
```diff
@@ -120,6 +120,23 @@ class TrilinosBlockBuilderAndSolver
     ///@name Life Cycle
     ///@{
 
+    /**
+     * @brief Default constructor with Parameters
+     * @param pNewLinearSystemSolver The linear solver for the system of equations
+     * @param ThisParameters The configuration parameters
+     */
+    explicit TrilinosBlockBuilderAndSolver(
```
loumalouomega: I guess you got this from my branch? (avoiding conflicts)

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
```diff
@@ -155,6 +172,33 @@ class TrilinosBlockBuilderAndSolver
     ///@name Operations
     ///@{
 
+    /**
+     * @brief This method assigns settings to member variables
+     * @param ThisParameters Parameters that are assigned to the member variables
+     */
+    void AssignSettings(const Parameters ThisParameters) override
+    {
+        this->mEchoLevel = ThisParameters["echo_level"].GetInt();
```
loumalouomega: Same

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
```diff
@@ -155,6 +172,33 @@ class TrilinosBlockBuilderAndSolver
...
+     * @return The default parameters
+     */
+    Parameters GetDefaultParameters() const override
+    {
+        const Parameters default_parameters = Parameters(R"(
+        {
+            "name"                  : "trilinos_block_builder_and_solver",
+            "echo_level"            : 1,
+            "guess_row_size"        : 80,
+            "synchronize_dof_set"   : false
```
loumalouomega: 👍🏻

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
```diff
@@ -885,6 +934,7 @@ class TrilinosBlockBuilderAndSolver
 
     EpetraCommunicatorType& mrComm;
     int mGuessRowSize;
+    bool mSynchronizeDofSet = false;
```
loumalouomega: We can even use a local flag (in my branch I register them)

loumalouomega (commented):
Okay from my side, wait until all the other branches are discussed and @KratosMultiphysics/technical-committee  agrees

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
jcotela: Nope, I took it from the base class. I didn't notice a params constructor on your branch, otherwise I would have used that one...

Will try to merge, no sense in duplicating efforts.

ddiezrod: this needs to be exported to python by the way

loumalouomega: I think is exported on my PR, if not should be

loumalouomega: Yes, the commit: https://github.com/KratosMultiphysics/Kratos/commit/ba77c3ad5d20b81fdf7286567bda51f40b85f15d, you can cherry-pick

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
philbucher: I think a member is fine here, local flag seems an overkill

loumalouomega: In the other PR we are already adding flags, so I will have not additional cost

philbucher: I still don't see the advantage in this particular case, where a simple bool is enough :)

jcotela:
In pause until #10886 is merged (mentioning mostly as a note to self)

ddiezrod:
I took the freedom to merge master here myself as https://github.com/KratosMultiphysics/Kratos/pull/10886 is already done. This is ready for review @KratosMultiphysics/technical-committee

loumalouomega (dismissed):
Okay from my side

RiccardoRossi:
why this was not merged?

loumalouomega:
> why this was not merged?

Looks like conflict pending to fix, @jcotela

jcotela:
This PR is not finished (If I remember correctly, at least the python interface is missing). Unfortunately it fell by the wayside due to higher priority tasks. @ddiezrod should we go for this now?

ddiezrod:
> This PR is not finished (If I remember correctly, at least the python interface is missing). Unfortunately it fell by the wayside due to higher priority tasks. @ddiezrod should we go for this now?

If it does not take too much of your time, yes please. It's not critical because it will not affect any of our users for the moment, but we will need it for some upcoming features.

jcotela:
Looks like the work to integrate the new changes to the python interface was already done by @ddiezrod in his last merge from master, so I think this is ready now

ddiezrod:

applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
```diff
@@ -560,6 +560,11 @@ class TrilinosBlockBuilderAndSolver
         }
 
         temp_dofs_array.Unique();
+
+        if (mSynchronizeDofSet) {
```
RiccardoRossi, line 646: why do we need this here? in principle dofs should be already in sync when we are at this point (dofs should be always at sync after update in fact)

RiccardoRossi: i had a short offline conversation with @jcotela, but i am still very much convinced that dofs should be considered in sync when you are outside of the builder and solver. If they are not, they should be made so before entering there, otherwise you are fixing something in the wrong point

jcotela: Hi @RiccardoRossi I checked the code locally after the call and line 564 above (the entry point for the new functionality in the builder and solver) belongs to SetUpDofSet, not to SystemSolveWithPhysics as it would seem from the changelog. The extra cost is incurred once per system set-up not per system solve, which I agree with you that would be wrong.

The changes in mpi_communicator are needed because there is no method that can add dofs to the dofset of a different partition (the dofs are known to exist on that partition too, no actual transfer of dofs is happening).

Personally, I would prefer to leave the extra synchronization step as off by default, since using the standard partitioning strategy in Kratos should be enough to ensure that this situation never happens. Fixing the dofset here requires an extra communication step followed by an additional sort+unique of the dofs, which would happen even if no dofs were added.

jcotela:
This is ready for review again

kratos/mpi/includes/mpi_communicator.h
```diff
@@ -512,6 +519,49 @@ class DofIdAccess: public NodalContainerAccess
...
+        std::size_t node_id, last_id = 0;
+        std::size_t i = 0;
+
+        for (auto& r_dof: rDofSet) {
+            node_id = r_dof.Id();
+
+            if (last_id != node_id) {
+                mNodeMap[node_id] = {i, 1};
+            }
+            else {
```
loumalouomega:
```suggestion
            } else {
```

loumalouomega (commented):
Minor comment, Okay from my side

pooyan-dadvand:
@RiccardoRossi would you please take a look at it?

PR #11063, opened 2023-04-27
Kratos global galerkin

SADPR:
### Description

This pull request addresses the time-consuming ROMs in the application, which took longer as the number of modes used increased. We identified two main issues causing this: the use of Ublas instead of Eigen library, and building the reduced order model element by element. #11031

### Changes
To resolve this, we have made the following changes:

Instead of building the ROM element by element ($\sum_e \mathbf{\Phi}^{eT} \mathbf{J}^e \mathbf{\Phi}^{e}$), we now build it globally. This involves creating the Global Jacobian $\mathbf{J}$ and Global Basis $\mathbf{\Phi}$, performing a Sparse to Dense product $\mathbf{Aux}=\mathbf{J} \mathbf{\Phi}$, and then a Dense to Dense product $\mathbf{\Phi}^T\mathbf{Aux}$. This approach is more efficient, especially when building the ROM reduced system for a basis $\mathbf{\Phi}$ with a large number of modes.

We have also replaced Ublas with Eigen library to improve performance.

The following files have been modified as part of this change:

- `applications/RomApplication/CMakeLists.txt`: Importing eigen library to rom application, which introduces a dependency with the linear solvers application.
- `applications/RomApplication/custom_python/add_custom_strategies_to_python.cpp`: Added the global builder and solver for the Galerkin case.
- `applications/RomApplication/custom_strategies/global_rom_builder_and_solver.h`: Introduced the global builder and solver. Note that the elemental builder and solver still exists. This is just a new option.
- `applications/RomApplication/python_scripts/calculate_rom_basis_output_process.py`: Added the assembling strategy flag, setting the global as the default.
- `applications/RomApplication/python_scripts/new_python_solvers_wrapper_rom.py`: Removed the assembling strategy flag to create the solver.
- `applications/RomApplication/python_scripts/rom_analysis.py`: Set the default assembling strategy when not provided.
- `applications/RomApplication/python_scripts/rom_manager.py`: Added the assembling strategy to the manager. Also fixed a bug related to the creation of the HROM model part by adding a missing line.
- `applications/RomApplication/python_scripts/rom_solver.py`: Added the "assembling_strategy" prefix to the solving strategy for the Galerkin projection only. We plan to do this for all solving strategies soon.

The following tests were preserved with the elemental assembling:

- `applications/RomApplication/tests/compressible_potential_test_files/RomParameters.json`
- `applications/RomApplication/tests/thermal_dynamic_test_files/ROM/RomParameters.json`
- `applications/RomApplication/tests/thermal_static_test_files/ROM/RomParameters.json`

We have multiple pull requests in the pipeline, so let's merge them carefully but quickly as many projects are using the ROM application.

### DIFF
--- a/applications/RomApplication/CMakeLists.txt
+++ b/applications/RomApplication/CMakeLists.txt
@@ -5,7 +5,41 @@ message("**** configuring KratosRomApplication ****")
 ################### PYBIND11
 include(pybind11Tools)
 
-include_directories( ${KRATOS_SOURCE_DIR}/kratos )
+add_definitions( -DEIGEN_MPL2_ONLY)
+
+include_directories(
+    ${KRATOS_SOURCE_DIR}/kratos
+    SYSTEM ${KRATOS_SOURCE_DIR}/applications/LinearSolversApplication/external_libraries/eigen3
+    SYSTEM ${KRATOS_SOURCE_DIR}/applications/LinearSolversApplication/
+)
+
+if( USE_EIGEN_MKL MATCHES ON )
+    if( DEFINED ENV{MKLROOT} )
+        message( "-- MKLROOT = $ENV{MKLROOT}" )
+
+        include_directories( SYSTEM $ENV{MKLROOT}/include )
+
+        link_directories("$ENV{MKLROOT}/lib")
+
+        if( NOT MSVC )
+            set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -m64 -L$ENV{MKLROOT}/lib/intel64 -Wl,--no-as-needed -lpthread -lm -ldl" )
+        endif()
+    elseif( DEFINED ENV{CONDA_PREFIX} )
+        message("-- Found Conda environment: $ENV{CONDA_PREFIX}")
+
+        if( WIN32 )
+            include_directories(SYSTEM $ENV{CONDA_PREFIX}/Library/include)
+            link_directories("$ENV{CONDA_PREFIX}/Library/lib")
+        else()
+            include_directories(SYSTEM $ENV{CONDA_PREFIX}/include)
+            link_directories("$ENV{CONDA_PREFIX}/lib")
+        endif()
+    else()
+        message( FATAL_ERROR "MKLROOT not defined" )
+    endif()
+
+    add_definitions( -DUSE_EIGEN_MKL -DEIGEN_USE_MKL_ALL )
+endif()
 
 ## RomApplication Core sources
 file(GLOB_RECURSE KRATOS_ROM_APPLICATION_CORE
@@ -41,6 +75,25 @@ IF(CMAKE_UNITY_BUILD MATCHES ON)
     set_target_properties(KratosRomApplicationCore PROPERTIES UNITY_BUILD_BATCH_SIZE ${KRATOS_UNITY_BUILD_BATCH_SIZE})
     set_target_properties(KratosRomApplication PROPERTIES UNITY_BUILD_BATCH_SIZE ${KRATOS_UNITY_BUILD_BATCH_SIZE})
 ENDIF(CMAKE_UNITY_BUILD MATCHES ON)
+
+if( USE_EIGEN_MKL MATCHES ON )
+    if( MSVC )
+        find_library(MKL_RT_LIB mkl_rt)
+        if(NOT MKL_RT_LIB)
+            message( FATAL_ERROR "mkl_rt.lib not found")
+        else(NOT MKL_RT_LIB)
+            message( "mkl_rt.lib found at: ${MKL_RT_LIB}")
+            target_link_libraries( KratosLinearSolversCore PUBLIC ${MKL_RT_LIB} )
+        endif(NOT MKL_RT_LIB)
+    elseif( ${CMAKE_CXX_COMPILER_ID} MATCHES Clang )
+        message( FATAL_ERROR "Clang does not yet support MKL" )
+    else( MSVC )
+        target_link_libraries( KratosLinearSolversCore PUBLIC mkl_rt )
+        if( USE_EIGEN_FEAST MATCHES ON )
+            target_link_libraries(KratosLinearSolversCore PUBLIC feast4 gfortran m)
+        endif()
+    endif( MSVC )
+endif()
 
 # changing the .dll suffix to .pyd (Windows)
 if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
--- a/applications/RomApplication/CMakeLists.txt
+++ b/applications/RomApplication/CMakeLists.txt
@@ -5,41 +5,9 @@ message("**** configuring KratosRomApplication ****")
 ################### PYBIND11
 include(pybind11Tools)
 
-add_definitions( -DEIGEN_MPL2_ONLY)
+kratos_add_dependency(${KRATOS_SOURCE_DIR}/applications/LinearSolversApplication)
 
-include_directories(
-    ${KRATOS_SOURCE_DIR}/kratos
-    SYSTEM ${KRATOS_SOURCE_DIR}/applications/LinearSolversApplication/external_libraries/eigen3
-    SYSTEM ${KRATOS_SOURCE_DIR}/applications/LinearSolversApplication/
-)
-
-if( USE_EIGEN_MKL MATCHES ON )
-    if( DEFINED ENV{MKLROOT} )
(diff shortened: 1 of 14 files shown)


emiroglu:
What do you think about the RAM usage? In exchange for computational efficiency some big models cant be run with ROM on systems with less RAM

applications/RomApplication/CMakeLists.txt
```diff
@@ -5,7 +5,41 @@ message("**** configuring KratosRomApplication ****")
 ################### PYBIND11
 include(pybind11Tools)
 
-include_directories( ${KRATOS_SOURCE_DIR}/kratos )
+add_definitions( -DEIGEN_MPL2_ONLY)
```
loumalouomega: Huuum, strong dependency

applications/RomApplication/custom_strategies/global_rom_builder_and_solver.h
```diff
@@ -0,0 +1,800 @@
...
+        LocalSystemVectorType RHS_Contribution = LocalSystemVectorType(0);
+
+        //vector containing the localization in the system of the different
+        //terms
+        Element::EquationIdVectorType EquationId;
+
+        // Assemble all elements
+        const auto timer = BuiltinTimer();
+
+        #pragma omp parallel firstprivate(nelements,nconditions, LHS_Contribution, RHS_Contribution, EquationId )
```
loumalouomega, line 642: Try to use parallel utilities (even despite not used in B&S, will be updated at some point)

applications/RomApplication/custom_strategies/global_rom_builder_and_solver.h
```diff
@@ -0,0 +1,788 @@
...
+     * which will be asked and set by another strategy object
+     */
+
+    ///@name Type Definitions
+    ///@{
+
+    // Class pointer definition
+    KRATOS_CLASS_POINTER_DEFINITION(GlobalROMBuilderAndSolver);
+
+    // The size_t types
```
loumalouomega, line 82: Try to use `using` instead of `typedef`

applications/RomApplication/custom_strategies/global_rom_builder_and_solver.h
loumalouomega: The same for all OMP loops

applications/RomApplication/custom_strategies/global_rom_builder_and_solver.h
```diff
@@ -0,0 +1,788 @@
...
+
+    ///@}
+    ///@name Protected life cycle
+    ///@{
+private:
+    ///@name Private member variables 
+    ///@{
+        
+    SizeType mNumberOfRomModes;
+    EigenDynamicMatrix meigen_romA;
```
loumalouomega: Should be 
```suggestion
    EigenDynamicMatrix mEigenRomA;
```

applications/RomApplication/custom_strategies/global_rom_builder_and_solver.h
```diff
@@ -0,0 +1,788 @@
...
+    EigenDynamicVector meigen_romb;
```
loumalouomega: Should be:

```suggestion
    EigenDynamicVector mEigenRomB;
```

applications/RomApplication/CMakeLists.txt
loumalouomega: My main problem is that it is hardcoded, today we discussed about this with @RiccardoRossi . For the moment no consensus how to replace spaces

loumalouomega (changes requested):
Some comments

applications/RomApplication/custom_strategies/global_rom_builder_and_solver.h
SADPR: Done.

applications/RomApplication/custom_strategies/global_rom_builder_and_solver.h
SADPR: Done.

applications/RomApplication/custom_strategies/global_rom_builder_and_solver.h
SADPR: Done.

loumalouomega (dismissed):
Removing my blockage (do bot merge yet!)

applications/RomApplication/CMakeLists.txt
SADPR: I am not an expert in this, maybe if @RiccardoRossi and/or @roigcarlo can give me a hand with this. I would appreciate it. I know we are making a strong dependency here. If you have any ideas, please let me know.

RiccardoRossi:
@loumalouomega we decided that we can live with having the ROM application to depend on eigen.

at a later stage, if we get a c++native solution we will depend on that, but meanwhile this is the solution we go for (different thing would be if it was in the core)

emiroglu:
> What do you think about the RAM usage? In exchange for computational efficiency some big models cant be run with ROM on systems with less RAM 
> 
> 
> 
> Response:
> 
> Thank you for bringing up the issue of RAM usage. We are aware that performing the sparse to dense (FOM size) products and dense to dense products may cause the system to run out of memory. Currently, the user can choose to run the elemental approach if they encounter this issue. However, we plan to address this by implementing matrix product splitting to reduce memory usage. Specifically, we plan to perform the matrix multiplication in smaller batches or using tiling or blocking techniques to reduce the memory requirements. This will enable us to continue to benefit from the computational efficiency of the updated Builder and Solver, while avoiding running out of memory.

I dont know who edited my comment but thanks for the response :)

applications/RomApplication/CMakeLists.txt
```diff
@@ -5,7 +5,41 @@ message("**** configuring KratosRomApplication ****")
...
-include_directories( ${KRATOS_SOURCE_DIR}/kratos )
+add_definitions( -DEIGEN_MPL2_ONLY)
+
+include_directories(
+    ${KRATOS_SOURCE_DIR}/kratos
+    SYSTEM ${KRATOS_SOURCE_DIR}/applications/LinearSolversApplication/external_libraries/eigen3
+    SYSTEM ${KRATOS_SOURCE_DIR}/applications/LinearSolversApplication/
+)
+
+if( USE_EIGEN_MKL MATCHES ON )
```
philbucher: I strongly recommend to avoid this

if you need eigen, then you should make the romapp dependent on the LinearSolversApp with

`kratos_add_dependency(${KRATOS_SOURCE_DIR}/applications/LinearSolversApplication)`

philbucher (commented):
is this necessary if you include it properly?

roigcarlo:
No, no needed if we derive the full app form the linear solvers ones (I believe at least). If @SADPR agrees it could be done.

That being said, the dependency is not on the LinearSolvers app. The problem is that we want to use eigen (the matrices in particular).

IMO the good solution here is to move eigen as a core library and use it in any app that needs it. But that introduces MKL into the core and ricc does not want it.

roigcarlo:
> We can expose MKL only in the LinearSolverApp and put Eigen in external_libraries IMO (I was thinking about at the beggining)

Mm if you know how to do it go ahead (in another PR) and if it works we can merge it here.

Anyway seems that thw policy change does not work properly :S

philbucher:
Why is this case different from how we use Trilinos in other Apps?

I can use it straight forward in the mapping App, and conceptually this PR does the same what I do

SADPR:
Unfortunatelly, I can't help a lot in this discussion since I don't fully understand the issue. It seems now that it has problems importing the RomApplication, but I guess it is still related to this since it is compiling and working perfectly in my computer.

philbucher:
locally I can import it now with MKL compiled and without (which I think was an issue of your implementation)
Should be much cleaner now and straight forward to use

roigcarlo:
> Why is this case different from how we use Trilinos in other Apps?

Because we didnt want to derive from the LinearSolvers, we are not using anything from there, its just used as a way to access the eigen matrix lib.

That being said, if it works like that ok :+1:

philbucher:
I see
BTW the Ublas wrappers are used from there

I think it is better as is, otherwise by avoiding (but not really) a small dependency, a lot of complex code is duplicated

Up to you guys. I needed sth to procrastinate with :smiley:

SADPR:
For me it seems ready to merge now, but it needs one approve. Thanks to all.

roigcarlo (approved):

PR #11069, opened 2023-05-01
[CORE] Exporting Accessor related from properties to Python

AlejandroCornejo:
**📝 Description**
Work in progress PR in which we'll be adding the newly added methods from the `Properties` related to the new `Accessor` to Python.

I have several questions:

- Exporting the `SetAccessor(const TVariableType& rVariable, AccessorPointerType pAccessor)` to Python. Again, the `AccessorPointerType `is a c++ class, how this works when exporting?

Than you for your help,


kratos/python/add_properties_to_python.cpp
```diff
@@ -112,7 +113,17 @@ typename TVariableType::Type GetValueHelperFunction1( TContainerType& rContainer
...
+template< class TContainerType, class TVariableType >
+typename TVariableType::Type GetValueHelperFunctionAccessor( TContainerType& rContainer,
+        const TVariableType& rVar,
+        const GeometryType& rGeometry,
+        const Vector& rShapeFunctionVector,
+        const ProcessInfo& rProcessInfo )
+{
+    return rContainer.GetValue(rVar, rGeometry, rShapeFunctionVector, rProcessInfo);
+}
+
```
rubenzorrilla: I'd use a lambda instead.

rubenzorrilla:
> 
> * @rubenzorrilla was saying to export the `Accessor` class (as far as I understood) but, since it is a c++ class I have doubts about how to do that...

As far as I understand there shouldn't be any difference to any other class. Did you experience any problems when trying?

rubenzorrilla:
I've been checking this and I'm afraid that if we want to export the `SetValue` of the `Properties` to Python (something that we definitively want), we need to revert the `unique_ptr` change. This is a long-standing problem in Pybind. The point is that we cannot transfer the ownership of the Python accessor instance to the c++ side.

Long story short, we either switch to the `shared_ptr` (easy to go option) or we can explore the use of an intrusive ptr (suggested by @sunethwarna).

To be discussed in the @KratosMultiphysics/technical-committee .

RiccardoRossi:
about exporting the Accessor to python (the issue about using UniquePtr) we propose the folliwing:

in essence we should NOT export the constructor of the accessor to python. Instead we should do something like

~~~cpp
py::class_<Accessor>(m,"ModelPart")
     //do NOT export constructor!!!
    .def("Create", [](Properties& rProp, ...parameters we need in construction ...) -> Accessor&
    {
        auto pnew = Kratos::make_unique<Accessor>(...);
        rProp.SetAccessor(std::move(pnew));
        return rProp.GetAccessor();
    })
~~~

AlejandroCornejo:
> 

So this means that the `SetAccessor` method of the properties will not be exported either right?

RiccardoRossi:
correct

rubenzorrilla:
Note for myself: working but I'm doing an undesired copy somewhere.

rubenzorrilla (commented):
Now it should be ready. Important things to be commented:
- It is ensured that there are no copies when retrieving the accessors from Python.
- I deliberately didn't export the `GetValue`(s) methods to Python. By doing so we enforce to always retrieve the values through the corresponding properties' `GetValue`.
- I took the chance to add a `HasAccessor` method to the properties. By doing so we can throw an error if asking for a non-existing accessor.

AlejandroCornejo:
OK from my side, thank you @rubenzorrilla  for your help. I can approve if the @KratosMultiphysics/technical-committee  agrees

kratos/includes/properties.h
```diff
@@ -297,24 +297,76 @@ class Properties : public IndexedObject
...
+     * @tparam TVariableType The variable type
+     * @param rVariable Variable to which the accessor refers to
+     * @return AccessorPointerType& Pointer to the requested variable accessor
+     */
+    template <class TVariableType>
+    Accessor& GetAccessor(const TVariableType& rVariable)
     {
-        mData.SetValue(rV, rValue);
+        auto it_value = mAccessors.find(rVariable.Key());
+        KRATOS_ERROR_IF(it_value == mAccessors.end())
```
RiccardoRossi, line 322: i wonder if we shouldn't do this check only in debug mode for efficiency. 
This is the same as not verifying that a given [i] entry exists when we access to a vector, and it is something that we normally accept for performance reasons

kratos/includes/properties.h
```diff
@@ -297,24 +297,76 @@ class Properties : public IndexedObject
...
+     * @tparam TVariableType The variable type
+     * @param rVariable Variable to which we are checking if an accessor exists
+     * @return true If there is accessor for the requested variable
+     * @return false If there is no accessor for the requested variable
+     */
+    template <class TVariableType>
+    bool HasAccessor(const TVariableType& rVariable) const
+    {
+        return (mAccessors.find(rVariable.Key()) == mAccessors.end()) ? false : true;
     }
```
RiccardoRossi, line 374: finally didn't we say we would also export an extended "GetValue" method for the properties? (which under the hood calls the accessor)

kratos/includes/properties.h
rubenzorrilla: I agree with this observation. This is one of the things we should discuss.

rubenzorrilla: Nasty thing is that this would result in a meaningless error...

kratos/includes/properties.h
rubenzorrilla: Yes. This is part of this PR.

AlejandroCornejo:
What is missing in this PR?

rubenzorrilla:
We've been discussing in the @KratosMultiphysics/technical-committee about this. Though current design is valid, ideally we'd like to have a sort of `SetAccessor` in the `Properties` API rather than passing the properties in which we are to set the accessor (which is somehow twisted).

@roigcarlo and @rubenzorrilla will give it a try.

AlejandroCornejo:
Let me know if I can do anything or if there is any update about this :)

roigcarlo:
It passes :tada:

kratos/includes/properties.h
sunethwarna: On the other hand, the `Element::Check` method can do the checks to identify (like always) whether required acessors are there, so we can leave this check only in debug. I'm just suggesting only.

sunethwarna (approved):
Finally after the long discussions :D.

PR #11068, opened 2023-04-29
[Core] Extend AccumReduction to be used with std::set

sunethwarna:
**📝 Description**
This PR extends existing `AccumReduction` to be used with `std::set` as well.

**🆕 Changelog**
- Extends `AccumReduction` to be used with `std::set`

PR #11070, opened 2023-05-02
[Core] Add ModelPart Union/Substract/Intersect operators

sunethwarna:
**📝 Description**
This PR adds ability to do `union/substract/intersect` operations with different model parts. These operations are done based on the nodes and not on the node positions. 

This works with model parts with different root model parts where they share same nodes. (e.g.: Model parts made by `ConnectivityPreserveModeller`)

This requires #11068.

**🆕 Changelog**
- Added `ModelPartOperationUtilities` with `Union`, `Substract`, and `Intersect` methods.
- Added `HasIntersection` method.
- Added OpenMP tests.
- Added MPI tests.


loumalouomega (commented):
Okay from my side

loumalouomega:
> Okay from my side

But Ci is failing...

loumalouomega:
> > Okay from my side
> 
> But Ci is failing...

~~~
/__w/Kratos/Kratos/kratos/utilities/model_part_operation_utilities.cpp:476:81:   required from here
/__w/Kratos/Kratos/kratos/utilities/reduction_utilities.h:270:16: error: 'class std::set<std::vector<const Kratos::Node<3>*> >' has no member named 'push_back'
         mValue.push_back(value);
~~~

Replace the set for the vector or look for a different strategy

sunethwarna:
The CI is failing because it requires #11068 which adds `AccumReduction` for `std::set` :)

ddiezrod:
This looks quite interesting, it'd be great if this could be somehow used as a modeler in my opinion....

kratos/python/add_other_utilities_to_python.cpp
```diff
@@ -803,6 +804,13 @@ void AddOtherUtilitiesToPython(pybind11::module &m)
              "snake_case to CamelCase conversion")
         ;
 
+    m.def_submodule("ModelPartOperationUtilities", "Free-floating utility functions for model part operations.")
+        .def("CheckValidityOfModelPartsForOperations", &ModelPartOperationUtilities::CheckValidityOfModelPartsForOperations, py::arg("main_model_part"), py::arg("list_of_checking_model_parts"), py::arg("thow_error"))
+        .def("Merge", &ModelPartOperationUtilities::Merge, py::arg("output_model_part_name"), py::arg("main_model_part"), py::arg("model_parts_to_merge"), py::arg("add_neighbours"), py::return_value_policy::reference_internal)
```
philbucher: I would call it `Union` if it is a union

I would not take the python typing as blocker, this can always be ignored

sunethwarna: done :)

kratos/utilities/model_part_operation_utilities.cpp
```diff
@@ -0,0 +1,609 @@
...
+                is_valid = false;
+                break;
+            }
+        }
+
+        return is_valid;
+    }
+};
+
+} // namespace ModelPartOperationHelperUtilities
```
roigcarlo: Up to you, but I would split this into 2/3 files:

1 - The ModelPartUtilities + Helper Functions ( or 2 )
2 - For the operation structs.

Since you made the effort to make it generic so the condition is defined in a struct, we could split it so its expandable by someone else who wants to implement a different operation (Xor, !Difference, custom, ...) like we have with the ParallelUtils reduction for example.

roigcarlo (dismissed):
Super minor comments, other than that :+1:  from my side.

kratos/utilities/model_part_operation_utilities.cpp
sunethwarna: done :)

sunethwarna:
@roigcarlo I seperated it to three files. Could you re-approve?

roigcarlo (approved):

PR #11071, opened 2023-05-02
[CLApp] Adding more tools to be able to print layer-wise info

AlejandroCornejo:
**📝 Description**
In this PR I am adding new capabilities to the Parallel Rule of Mixtures to be able to print several values (`DAMAGE`, `PLASTIC_DISSIPATION`, `PLASTIC_STRAIN_VECTOR` and `CAUCHY_STRESS_VECTOR`) at a desired layer (global axes).

This can be done now by writting (example with layer 1, could be up to 5):

`PLASTIC_STRAIN_VECTOR_LAYER_1`
`DAMAGE_LAYER_1`
`UNIAXIAL_STRESS_LAYER_1`
`CAUCHY_STRESS_VECTOR_LAYER_1`

I think that this will be really useful for you @balcayde


applications/ConstitutiveLawsApplication/constitutive_laws_application.cpp
```diff
@@ -477,5 +479,32 @@ void KratosConstitutiveLawsApplication::Register()
     KRATOS_REGISTER_SYMMETRIC_3D_TENSOR_VARIABLE_WITH_COMPONENTS(ISOTROPIC_ANISOTROPIC_YIELD_RATIO);
     KRATOS_REGISTER_SYMMETRIC_3D_TENSOR_VARIABLE_WITH_COMPONENTS(ORTHOTROPIC_ELASTIC_CONSTANTS);
 
+    // For printing pursposes in Rule of Mixtures...
+    KRATOS_REGISTER_VARIABLE(CAUCHY_STRESS_VECTOR_LAYER_1)
```
loumalouomega, line 491: Huuuuuum, I don't like this... Huuuum, thinking alternatives

applications/ConstitutiveLawsApplication/custom_constitutive/composites/rule_of_mixtures_law.h
```diff
@@ -649,6 +649,17 @@ class KRATOS_API(CONSTITUTIVE_LAWS_APPLICATION) ParallelRuleOfMixturesLaw
...
 
+    /**
+     * @brief This method returns true if the variable is related to a layer
+     * @param rVariable The variable to be printed
+     */
+    template<class TVariableType>
+    const bool IsLayerVariable(const TVariableType& rVariable)
+    {
+        // returns true if LAYER is found
+        return (rVariable.Name().find("LAYER") != std::string::npos);
```
loumalouomega, line 660: String find is sloooow

loumalouomega (commented):
Sorry I need to think a cleaner way to do this. @roigcarlo any suggestion?

applications/ConstitutiveLawsApplication/custom_constitutive/composites/rule_of_mixtures_law.h
AlejandroCornejo: I could basically repeat a lot of if...else if with all the new variables but I though that in this way is easier to maintain

loumalouomega: Yes, I am thinking something else, like for example checking the variable Id if all variables of layer have a way to be filtered. Thinking. Luckily this is only post-process, but I need to think about it.

applications/ConstitutiveLawsApplication/constitutive_laws_application.cpp
loumalouomega: Maybe we can add a macro to add several variables at once, the declaration time will be the same but at least the code will be cleaner

AlejandroCornejo:
ping

applications/ConstitutiveLawsApplication/constitutive_laws_application.cpp
loumalouomega: I am still not happy with this...

loumalouomega: @pooyan-dadvand any idea how to solve it?

AlejandroCornejo:
is there any alternative?

PR #11073, opened 2023-05-03
[OptApp] Resructure 15 - Updated control interface.

sunethwarna:
**📝 Description**
This adds the base `Control` interface and one implementation of it named `MaterialPropertiesControl`.

**🆕 Changelog**
- Adds `Control`
- Adds `MaterialPropertiesControl`
- Adds tests


applications/OptimizationApplication/python_scripts/controls/control.py
```diff
@@ -0,0 +1,114 @@
+from abc import ABC, abstractmethod
+from KratosMultiphysics.OptimizationApplication.utilities.union_utilities import SupportedSensitivityFieldVariableTypes
+from KratosMultiphysics.OptimizationApplication.utilities.union_utilities import ContainerExpressionTypes
+
+class Control(ABC):
+    """Base abstract control class.
+
+    The base abstract control class has the following responsibilities.
+        1. Initalizes and finalizes the control.
+        2. Mapping physical space gradients to control space gradients.
```
RezaNajian: I would make it more elaborated by saying  "Mapping consistently physical space gradients if a mapping/filtering/parametrization is involved, otherwise pass "

applications/OptimizationApplication/python_scripts/controls/control.py
```diff
@@ -0,0 +1,114 @@
...
+        2. Updating the controlled parts of the model part with new design given in control space.
```
RezaNajian: Since it is the abstract base class let's be abstract and say simply "Updating the controlled parts of the model part with new design given" without any word about control space

applications/OptimizationApplication/python_scripts/controls/control.py
```diff
@@ -0,0 +1,114 @@
...
+    def Initialize(self) -> None:
+        """Initializes the control.
+
+        This method initializes the control. This is only called once in the whole optimization process.
+
+        """
+        pass
+
+    @abstractmethod
+    def Check(self) -> None:
```
RezaNajian, line 35: what is the need for this function, what kind of check is it. It does not ring a bell.

applications/OptimizationApplication/python_scripts/controls/control.py
```diff
@@ -0,0 +1,114 @@
...
+    def Finalize(self) -> None:
+        """Finalizes the control.
+
+        This method finalizes the control. This is only called once in the whole optimization process.
+
+        """
+        pass
+
+    @abstractmethod
+    def GetPhysicalKratosVariables(self) -> 'list[SupportedSensitivityFieldVariableTypes]':
```
RezaNajian, line 53: We need to update the name of SupportedSensitivityFieldVariableTypes because it is not really a Sensitivity type I would call it simply SupportedFieldVariableTypes, You need to update union_utils

applications/OptimizationApplication/python_scripts/controls/control.py
```diff
@@ -0,0 +1,114 @@
...
+
+        This method returns list of physical variable which are controlled by the control. In most of the cases, there is only one
+        control variable per control. But in some rare cases, there may be the requirement to control two physical variables. Anyhow,
+        the control space should only have one variable per control in the control space, but they can have multiple physical variables
+        in their design space.
+
+        If the case is that this control has one physical variable, then it should be same as the control variable.
+
+        Returns:
+            list[SupportedSensitivityFieldVariableTypes]: List of physical control variables.
```
RezaNajian, line 62: should be updated to list[SupportedControlledFieldVariableTypes] or list[SupportedPhysicalFieldVariableTypes]

applications/OptimizationApplication/python_scripts/controls/control.py
```diff
@@ -0,0 +1,114 @@
...
+    @abstractmethod
+    def GetPhysicalKratosVariables(self) -> 'list[SupportedSensitivityFieldVariableTypes]':
+        """Returns list of physical control variables controlled by this control.
+
+        This method returns list of physical variable which are controlled by the control. In most of the cases, there is only one
+        control variable per control. But in some rare cases, there may be the requirement to control two physical variables. Anyhow,
+        the control space should only have one variable per control in the control space, but they can have multiple physical variables
+        in their design space.
+
+        If the case is that this control has one physical variable, then it should be same as the control variable.
```
RezaNajian: This is not true, please remove it

applications/OptimizationApplication/python_scripts/controls/control.py
```diff
@@ -0,0 +1,114 @@
...
+
+        Returns:
+            list[SupportedSensitivityFieldVariableTypes]: List of physical control variables.
+        """
+        pass
+
+    @abstractmethod
+    def GetEmptyControlField(self) -> ContainerExpressionTypes:
```
RezaNajian: This is weird to get an empty ContainerExpression to get control's model part's. If only the control model part is needed I would call this function GetControlModelPart and return it. Please correct it.

RezaNajian: Also I would not make it abstract since it is shared between all controls. Also I would make self.control_model_part at the base class

applications/OptimizationApplication/python_scripts/controls/control.py
```diff
@@ -0,0 +1,114 @@
...
+        this model part is acting on. This has O(1) complexity, hence has the least cost because it does not read
+        any data from respective model part's container.
+
+        Returns:
+            ContainerExpressionTypes: Returns a new empty ContainerExpression corresponding to control's model part's respective container.
+        """
+        pass
+
+    @abstractmethod
+    def MapGradient(self, physical_gradient_variable_container_expression_map: 'dict[SupportedSensitivityFieldVariableTypes, ContainerExpressionTypes]') -> ContainerExpressionTypes:
```
RezaNajian, line 99: I would not make this abstract since not all controls have mapping operation/transformation

applications/OptimizationApplication/python_scripts/controls/control.py
```diff
@@ -0,0 +1,114 @@
...
+            ContainerExpressionTypes: Gradients mapped in to control space.
+        """
+        pass
+
+    @abstractmethod
+    def Update(self, control_field: ContainerExpressionTypes) -> bool:
+        """Modifies the current control with the given control field.
+
+        This method modifies the control with the given control field from @ref control_field. This
+        control_field should be in the control space, and it should not represent a change of design.
```
RezaNajian: This is extra info with the assumption that there is always two spaces, namely physical and control. Please remove

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
```diff
@@ -0,0 +1,102 @@
...
+    TODO: Extend with filtering techniques when they are implemented.
+
+    """
+    def __init__(self, model: Kratos.Model, parameters: Kratos.Parameters, optimization_problem: OptimizationProblem):
+        super().__init__()
+
+        default_settings = Kratos.Parameters("""{
+            "combined_output_model_part_name": "<CONTROL_NAME>_combined_no_neighbours_control",
+            "model_part_names"               : [""],
+            "control_variable_name"          : ""
```
RezaNajian: control_variable_name should be specified by the control and it should not be an input. So it should not be given as input because the control implementation and type decides about it. since we have list of model parts let's call the input "model_parts_names"

RezaNajian: I would not make room for "combined_output_model_part_name". The control model part and its output should be implemented explicitly by control.

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
```diff
@@ -0,0 +1,102 @@
...
+        }""")
+        parameters.ValidateAndAssignDefaults(default_settings)
+
+        self.output_model_part_name = parameters["combined_output_model_part_name"].GetString()
+        self.model_part_names = parameters["model_part_names"].GetStringArray()
+        self.optimization_problem = optimization_problem
```
RezaNajian: I would put self.optimization_problem = optimization_problem and self.model = model at the base class otherwise it gets repeated ?!

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
```diff
@@ -0,0 +1,102 @@
...
+        control_variable_name = parameters["control_variable_name"].GetString()
+        control_variable_type = Kratos.KratosGlobals.GetVariableType(control_variable_name)
+        if control_variable_type != "Double":
+            raise RuntimeError(f"{control_variable_name} with {control_variable_type} type is not supported. Only supports double variables")
+
+        self.control_variable = Kratos.KratosGlobals.GetVariable(control_variable_name)
+
+    def Initialize(self) -> None:
+        # get the model part name
+        self.control_name = self.optimization_problem.GetComponentName(self)
```
RezaNajian: The name of the control should be given at init not retrieved here from optimization problem or ?

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
```diff
@@ -0,0 +1,102 @@
...
+
+        # create the combined model part
+        if not root_model_part.HasSubModelPart(output_model_part_name):
+            self.model_part: Kratos.ModelPart = Kratos.ModelPartOperationUtilities.Merge(output_model_part_name, root_model_part, model_parts_list, False)
+            # now create entity specific properties for the merged model part which is used for the control.
+            KratosOA.OptimizationUtils.CreateEntitySpecificPropertiesForContainer(self.model_part, self.model_part.Elements)
+        else:
+            self.model_part = root_model_part.GetSubModelPart(output_model_part_name)
+
+    def Check(self) -> None:
```
RezaNajian, line 57: as i mentioned before this check looks useless better not to have it in base class

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
```diff
@@ -0,0 +1,102 @@
...
+        self.model_part_names = parameters["model_part_names"].GetStringArray()
+        self.optimization_problem = optimization_problem
+        self.model = model
+
+        control_variable_name = parameters["control_variable_name"].GetString()
+        control_variable_type = Kratos.KratosGlobals.GetVariableType(control_variable_name)
+        if control_variable_type != "Double":
+            raise RuntimeError(f"{control_variable_name} with {control_variable_type} type is not supported. Only supports double variables")
+
+        self.control_variable = Kratos.KratosGlobals.GetVariable(control_variable_name)
```
RezaNajian: this is controlled/physical_variable

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
```diff
@@ -0,0 +1,102 @@
...
+    def Check(self) -> None:
+        pass
+
+    def Finalize(self) -> None:
+        pass
+
+    def GetPhysicalKratosVariables(self) -> list[SupportedSensitivityFieldVariableTypes]:
+        return [self.control_variable]
+
+    def GetEmptyControlField(self) -> ContainerExpressionTypes:
```
RezaNajian: Please refer to my comment on this function in the base class

applications/OptimizationApplication/python_scripts/utilities/union_utilities.py
```diff
@@ -1,6 +1,15 @@
 from typing import Union
 
 import KratosMultiphysics as Kratos
+import KratosMultiphysics.OptimizationApplication as KratosOA
 
 # used union types for type hinting in python and for automated documentation generation
-SupportedSensitivityFieldVariableTypes = Union[Kratos.DoubleVariable, Kratos.Array1DVariable3]
\ No newline at end of file
+SupportedSensitivityFieldVariableTypes = Union[Kratos.DoubleVariable, Kratos.Array1DVariable3]
```
RezaNajian, line 7: The name is not good and self-explaining, would not it be better to use directly  Union[Kratos.DoubleVariable, Kratos.Array1DVariable3] rather than making SupportedSensitivityFieldVariableTypes ?! 
if not then I would prefer SupportedFieldVariableTypes

applications/OptimizationApplication/tests/test_material_properties_control.py
```diff
@@ -0,0 +1,111 @@
+
```
RezaNajian: put it under a folder dedicated to controls tests

RezaNajian (changes requested):
@sunethwarna Thanks for making this PR. In general it is good however please address or comment on the points.

applications/OptimizationApplication/python_scripts/controls/control.py
Igarizza: list[SupportedPhysicalFieldVariableTypes]

applications/OptimizationApplication/python_scripts/controls/control.py
Igarizza: "This is weird to get an empty ContainerExpression to get control's model parts." Please explain what you mean as get control's model part. I am not sure if I get you correctly.

applications/OptimizationApplication/python_scripts/controls/control.py
Igarizza: I guess, one should implement it in individual control implementations if you need to check smth.

applications/OptimizationApplication/python_scripts/controls/control.py
Igarizza: consistently is also not correct. One can filter only gradients and not updates if he wants. 

```suggestion
        2. Maps physical space gradients to control space gradients, if needed. Otherwise, it passes. 
```

applications/OptimizationApplication/python_scripts/controls/control.py
Igarizza: I think we can keep it because most controls do filtering. If not, the control can simply copy. We anyway need a place to copy the fields, even if the filtering doesn't happen.

applications/OptimizationApplication/python_scripts/controls/control.py
sunethwarna: I thought about it too and it makes sense. I will do this in a seperate PR. Renaming will change lots of files.

applications/OptimizationApplication/python_scripts/controls/control.py
sunethwarna: renaming will happen in a seperate PR once I am finished untill the standardized routines.

applications/OptimizationApplication/python_scripts/controls/control.py
sunethwarna: I would say it is not weird and it is required. Returning only a model part is not enough, because model parts has, nodes/conditions/elements and control needs to provide which container of model part it is controlling. This empty control field has all the information.

I would keep the base control class fully clean and abstract and not have any member variables so users can call through predefined member functions to retrieve data.

applications/OptimizationApplication/python_scripts/controls/control.py
sunethwarna: I agree with @Igarizza....

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
sunethwarna: The control variable name is a variable for `SystemIndetification` hence, i would keep it for this control. For specific controls, we can remove this `control_variable_name`. This is not for all the controls.

I need the output `combined_output_model_part_name`  which to create the intersected and merged model parts so if the user wants, then they can re-use model parts. (and outputing to VTK will be really nice with nice names).

The defaults are there in most cases, and user won't have to touch the defaults.

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
sunethwarna: I would keep the repition. I like to keep the interfaces fully clean. because, some controls may do everything in the `__init__` and then there is no point having them in the base class.

applications/OptimizationApplication/python_scripts/controls/control.py
sunethwarna: I would keep the repition. I like to keep the interfaces fully clean. because, some controls may do everything in the `__init__` and then there is no point having them in the base class.

applications/OptimizationApplication/python_scripts/controls/control.py
sunethwarna: Only use case for this is in the following situation, when a control has to do check once all the components (responses, execiton_policies, other controls) are properly initialized. Otherwise, we can do the check in Initialize it self. 

Whats your decision @RezaNajian @Igarizza ?

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
sunethwarna: as mentioned above.

applications/OptimizationApplication/python_scripts/controls/control.py
RezaNajian: As far as I understood this function has two usages. The first use is creating an empty data field over the control's containers to be used by responses and mapping but also in the MapGradient of the master control for cross-checking the containers. Based on the first use and for the sake of clarity, I would rename it to GetEmptyField

applications/OptimizationApplication/python_scripts/controls/control.py
sunethwarna: removed :)

applications/OptimizationApplication/python_scripts/controls/control.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
```diff
@@ -0,0 +1,105 @@
...
+    """Material properties control
+
+    This is a generic material properties control which creates a control
+    for the specified control variable. This does not do any filtering.
+
+    TODO: Extend with filtering techniques when they are implemented.
+
+    """
+    def __init__(self, model: Kratos.Model, parameters: Kratos.Parameters, optimization_problem: OptimizationProblem):
+        super().__init__()
```
RezaNajian: please remove super init

applications/OptimizationApplication/python_scripts/controls/control.py
sunethwarna: renamed to `GetEmptyField`

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
sunethwarna: changed to `controlled_physical_variable`

applications/OptimizationApplication/python_scripts/controls/control.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
sunethwarna: changed :)

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
sunethwarna: changed :)

applications/OptimizationApplication/python_scripts/utilities/union_utilities.py
sunethwarna: Will do the renaming in a clean PR only having the renamings..

applications/OptimizationApplication/tests/test_material_properties_control.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
sunethwarna: done :)

sunethwarna (commented):
@RezaNajian @Igarizza Can you check the answers?
I did all the changes as we spoke as well. Could you check?

Igarizza (dismissed):
Good to go!

applications/OptimizationApplication/python_scripts/controls/control.py
RezaNajian: It is fine for now.

RezaNajian (approved):
Tnx Suneth :-) Just sth crossed my mind after checking PR 16. We need to have a function at base class which returns true control variables

RezaNajian (changes requested):
Please implement GetControlVariables which retruns CollectiveExpressions of the variables

RezaNajian (approved):

PR #11074, opened 2023-05-03
[OptApp] Restructure 16 - Add master control

sunethwarna:
**📝 Description**
Adds the `MasterControl` and the relevant tests

**🆕 Changelog**
- Add `MasterControl`
- Add tests.


applications/OptimizationApplication/python_scripts/controls/master_control.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        if len(self.__list_of_controls) != len(update_collective_expressions.GetContainerExpressions()):
+            raise RuntimeError(f"Controls size and update size mismatch [ number of controls: {len(self.__list_of_controls)}, number of container expressions: {len(update_collective_expressions.GetContainerExpressions())} ].")
+
+        update_map: 'dict[Control, bool]' = {}
+        for control, container_expression in zip(self.__list_of_controls, update_collective_expressions.GetContainerExpressions()):
+            update_map[control] = control.Update(container_expression)
+
+        return update_map
+
```
Igarizza, line 181: I have found out that we need a new method, which is GetEmptyControlField(). The function should go through all controls and call GetEmptyControlField() for each one. The function should return collective expression.

For VM control GetEmptyControlField() should return zeros. For the rest ones, I don't know=), I guess in topology we take directly the Young's Modulus and densitiy?

Igarizza (commented):
All looks good for me, but I need GetInitialControlField() to get initial control field. For VM I can use GetEmptyControlField()

applications/OptimizationApplication/python_scripts/controls/master_control.py
sunethwarna: If you need the initial control field, what you can do is, call `GetEmptyControlField` of the `MasterControl` at the initialize, and save it somewhere you want. I would keep it inside the algorithm because, this requiremenet is algorithm specific. What would you say?

Igarizza: That is the point, that I have tried to explain. You can use GetEmptyControlField to Init the control field if we use Vertex Morphing. In the case of topology optimization, I think we need to get the Young Module of the Model as an Initial control field.

This comment is not a blocker. For now, I can code with GetEmptyControlField. If for Topology we need smth else, we can add a function and adjust the algorithm's Initial. 

So, basically, in PR 15, the functions should be added to each control (and the control knows which Initial values to provide, zeros or smth from the model), then Master Control can flatten and return collective expression.

Igarizza (dismissed):
Good for me

Igarizza (approved):

PR #11075, opened 2023-05-03
[OptApp] Restructure 17 - New response function interface

sunethwarna:
**📝 Description**
This updates the `ResponseFunction` interface to the new design. The `MassResponseFunction` and `LinearStrainEnergyResponseFunction` have been accordingly changed.

**🆕 Changelog**
- Updated `ResponseFunction` base interface.
- Updated `MassResponseFunction` and its utilities accordingly.
- Updated `LinearStrainEnergyResponseFunction` and its utilities accordingly.
- Updated tests


applications/OptimizationApplication/python_scripts/responses/response_function.py
```diff
@@ -1,27 +1,110 @@
...
 
 class ResponseFunction(ABC):
-    @classmethod
+    """Base response function.
+
+    This reponse function is the base response function. This is assumed to have following responsibilities.
+        1. CalculateValue for a fresh design. (@see CalculateValue)
+        2. CalculateSensitivity for a fresh design (@see CalculateSensitivity)
+
+    This control should only work on one model part. Hence, if multiple model parts required then,
```
RezaNajian: This response

applications/OptimizationApplication/python_scripts/responses/response_function.py
```diff
@@ -1,27 +1,110 @@
...
+        Please return all the dependent physical variables, eventhough the gradient computation is not yet implemented
+        for some to avoid future bugs.
+
+        Returns:
+            list[SupportedSensitivityFieldVariableTypes]: All dependent physical variables of the response.
+        """
         pass
 
     @abstractmethod
     def CalculateValue(self) -> float:
```
RezaNajian, line 67: As far as I remember, since the optimization algorithm will call every response for the given design, we agreed that design_vars as collective_expressions are input to this function, or I am wrong ?! That's why we have def Update(self, control_field: ContainerExpressionTypes) -> bool: in the master control

applications/OptimizationApplication/python_scripts/responses/response_function.py
```diff
@@ -1,27 +1,110 @@
...
+
+        This method should always calculate the response value assuming the domain has changed.
+
+        Returns:
+            float: Calculated response value.
+        """
+        pass
+
+    @abstractmethod
+    def CalculateGradient(self, physical_variable_collective_expressions: 'dict[SupportedSensitivityFieldVariableTypes, KratosOA.ContainerExpression.CollectiveExpressions]') -> None:
```
RezaNajian, line 78: To facilitate and simplify the implementation of optimization algorithms, I would make the design collective_expressions as input to this function. Similar to the dependency check in the MapGradient function of Master control, we check if there is a dependency and if yes then compute sens for the found sensitivity types

RezaNajian (changes requested):
@sunethwarna In the base response class, there are two major issues that will result in changes. I will do a comprehensive review after settling them down.

applications/OptimizationApplication/python_scripts/responses/response_function.py
sunethwarna: done :)

applications/OptimizationApplication/python_scripts/responses/response_function.py
sunethwarna: This will be done in the `ResponseRoutine`. This  `ResponseFunction` is the one which will be developed by most of the developers who does not care about the updates and all the stuff for optimization.

applications/OptimizationApplication/python_scripts/responses/response_function.py
sunethwarna: Same as above. This will be facilitated in the `ResponseRoutine`

applications/OptimizationApplication/python_scripts/responses/response_function.py
sunethwarna: see #11077

applications/OptimizationApplication/python_scripts/responses/response_function.py
sunethwarna: see #11077

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
```diff
@@ -6,11 +6,13 @@
...
+def Factory(name: str, model: Kratos.Model, parameters: Kratos.Parameters, optimization_problem: OptimizationProblem) -> ResponseFunction:
+    return LinearStrainEnergyResponseFunction(name, model, parameters, optimization_problem)
 
 class LinearStrainEnergyResponseFunction(ResponseFunction):
-    def __init__(self, model: Kratos.Model, parameters: Kratos.Parameters, optimization_problem: OptimizationProblem):
+    def __init__(self, name: str, model: Kratos.Model, parameters: Kratos.Parameters, optimization_problem: OptimizationProblem):
+        super().__init__(name)
+
         default_settings = Kratos.Parameters("""{
             "combined_output_model_part_name": "<RESPONSE_NAME>_combined_no_neighbours_response",
```
RezaNajian: This option is not self explaining, and I would not expose it to the user, what if there is no overlap between evaluated and sensitivity model part, from the implementation it looks like you combine the evaluated model parts.

RezaNajian (dismissed):

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
sunethwarna: makes sense, and removed.

applications/OptimizationApplication/custom_python/optimization_python_application.cpp
```diff
@@ -173,6 +173,7 @@ PYBIND11_MODULE(KratosOptimizationApplication, m)
     //adjoint RHS
     KRATOS_REGISTER_IN_PYTHON_3D_VARIABLE_WITH_COMPONENTS(m, ADJOINT_RHS);
 
+    KRATOS_REGISTER_IN_PYTHON_3D_VARIABLE_WITH_COMPONENTS(m, SHAPE);
```
Igarizza, line 180: COORDS ??? as a better name
```suggestion
    KRATOS_REGISTER_IN_PYTHON_3D_VARIABLE_WITH_COMPONENTS(m, COORDS);
```

applications/OptimizationApplication/python_scripts/responses/response_function.py
```diff
@@ -1,27 +1,115 @@
...
 import KratosMultiphysics as Kratos
+import KratosMultiphysics.OptimizationApplication as KratosOA
 from KratosMultiphysics.OptimizationApplication.utilities.union_utilities import SupportedSensitivityFieldVariableTypes
 
 class ResponseFunction(ABC):
-    @classmethod
+    """Base response function.
+
+    This reponse function is the base response function. This is assumed to have following responsibilities.
+        1. CalculateValue for a fresh design. (@see CalculateValue)
```
Igarizza:
```suggestion
        1. CalculateValue for a new design. (@see CalculateValue)
```

applications/OptimizationApplication/python_scripts/responses/response_function.py
```diff
@@ -1,27 +1,115 @@
...
+        2. CalculateSensitivity for a fresh design (@see CalculateSensitivity)
```
Igarizza:
```suggestion
        2. CalculateSensitivity for a new design (@see CalculateSensitivity)
```

applications/OptimizationApplication/python_scripts/responses/response_function.py
```diff
@@ -1,27 +1,115 @@
...
+        """Finalizes the response.
+
+        This method finalizes the response. This is only called once in the whole optimization process.
+
+        """
         pass
 
     @abstractmethod
-    def Check(self) -> None:
+    def GetDependentPhysicalKratosVariables(self) -> 'list[SupportedSensitivityFieldVariableTypes]':
```
Igarizza:
```suggestion
    def GetImplementedPhysicalKratosVariables(self) -> 'list[SupportedSensitivityFieldVariableTypes]':
```

applications/OptimizationApplication/python_scripts/responses/response_function.py
```diff
@@ -1,27 +1,115 @@
...
+        This method should return all the dependent physical variables of the response to make sure
+        that sensitivities w.r.t. those variables are requested from the response. If sensitivities
+        w.r.t. other variables are required from this response, they are ASSUMED TO BE ZERO.
+
+        Please return all the dependent physical variables, eventhough the gradient computation is not yet implemented
+        for some to avoid future bugs.
+
+        Returns:
+            list[SupportedSensitivityFieldVariableTypes]: All dependent physical variables of the response.
+        """
```
Igarizza, line 63: @RezaNajian Please think about whether we assume zero sensitivities for physical variables which are not implemented where the control have partial overlap, or should the developer give all the influencial physical variables of the response, so then requested physical gradients and allowed gradients can be cross checked return proper zero senstivities for non-influential physical variables.

```suggestion
        """Returns the implemented dependent physical kratos variables of the response.

        This method should return the implemented dependent physical variables of the response to make sure
        that sensitivities w.r.t. those variables are requested from the response. If sensitivities
        w.r.t. other variables are required from this response, they are ASSUMED TO BE ZERO.

        Please return all the dependent physical variables, eventhough the gradient computation is not yet implemented
        for some to avoid future bugs.

        Returns:
            list[SupportedSensitivityFieldVariableTypes]: All dependent physical variables of the response.
        """
```

Igarizza (approved):

applications/OptimizationApplication/python_scripts/responses/mass_response_function.py
```diff
@@ -2,33 +2,69 @@
...
-            self.model_parts.append(model[model_part_name])
+        self.model_part_names = parameters["evaluated_model_part_names"].GetStringArray()
+        self.model = model
+        self.model_part: Kratos.ModelPart = None
 
-        if len(self.model_parts) == 0:
+        if len(self.model_part_names) == 0:
             raise RuntimeError("No model parts were provided for MassResponseFunction.")
 
+    def GetDependentPhysicalKratosVariables(self) -> list[SupportedSensitivityFieldVariableTypes]:
```
Igarizza:
```suggestion
    def GetImplementedPhysicalKratosVariables(self) -> list[SupportedSensitivityFieldVariableTypes]:
```

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
```diff
@@ -4,42 +4,73 @@
...
-            self.model_parts.append(model[model_part_name])
+        self.model = model
+        self.model_part: Kratos.ModelPart = None
+        self.primal_analysis_execution_policy_decorator: ExecutionPolicyDecorator = optimization_problem.GetExecutionPolicy(parameters["primal_analysis_name"].GetString())
 
-        if len(self.model_parts) == 0:
+        if len(self.model_part_names) == 0:
             raise RuntimeError("No model parts were provided for LinearStrainEnergyResponseFunction.")
 
+    def GetDependentPhysicalKratosVariables(self) -> list[SupportedSensitivityFieldVariableTypes]:
```
Igarizza:
```suggestion
    def GetImplementedPhysicalKratosVariables(self) -> list[SupportedSensitivityFieldVariableTypes]:
```

Igarizza (changes requested):

applications/OptimizationApplication/custom_python/optimization_python_application.cpp
Igarizza: Let's keep the old name shape as we used to it and a lot of var in the code follow the name SHAPE. For other shape param, like NURBS, we can use NURBS_SHAPE or similar.

applications/OptimizationApplication/python_scripts/responses/response_function.py
Igarizza: After one more round of discussions we get to a decision:
1. We return here only the list of the des_var that are implemneted for Sens. Analysis
2. We add the Messages into log file, where we assume sens are zero
3. All raw sens can be outputted into the vtk vtu etc. for visual check
4. We go with suggested name above: GetImplementedPhysicalKratosVariables

RezaNajian (dismissed):

RezaNajian:
@sunethwarna Can we merge this?

sunethwarna:
Can you give me till tomorrow? I have to fix the merge conflicts, and did not get time for that yet.... :/

RezaNajian (dismissed):

sunethwarna:
The current PR was failing due to following reasons:
1. Linear strain energy response test's project files were missing.
2. In core `MergeModelParts` changed to `UnionModelParts`.
3. The `COORDS` did not have its counter sensitivity variable `COORDS_SENSITIVITY`.

Hi @RezaNajian , I fixed all of the above.  I would suggest to keep the variable `SHAPE`, because we already have `SHAPE_SENSITIVITY` in the core which will be beneficial when we make different responses in different applications. In the case of other types of shapes, we can call them with different prefixes to shape (e.g. `NURBS_SHAPE`)

Side note:
Each application can define their own variables, even the same variable can be redefined in all the applications. But if they have the same name, then they will be the same variable. So we can use this approach to develop responses in different applications by redefining the same variables in OptApp if they are not found in the core or in that application. `SHAPE` will be the mostly used control, hence, if we are to change it to `COORDS` then, each application needs to redefine `COORDS` and `COORDS_SENSITIVITY` which in my opinon not ideal. So if we stick with shape, then in future I can move `SHAPE` variable also to core, so for shape sensitivity, no one has to redefine any variables.

Igarizza:
> 

Regarding keeping "SHAPE": I fully agree to keep shape for two reasons: 
1. we have used to it after so many years
2. a lot of parts of the codes or comments use shape keywords a lot.

Igarizza (approved):
Thank you for your work!

sunethwarna:
Hi @RezaNajian we will merge this since we need this for others, if you feel the variable name need to be changed, lets do it in a seperate renaming PR. :)

PR #11083, opened 2023-05-05
[Core] Neighbor MPC Assigner: A process and utility for assigning MPCs to neighbouring nodes

SADPR:
### DESCRIPTION:

**Utility:**

This pull request adds a new class that provides a method to search for neighboring nodes of a given nodes structure within a specified radius and assign a multipoint constraint (MPC). The MPC weights are calculated based on a spatial function that employs radial basis functions.

**Process:**

This pull request enhances the functionality of the master-slave MPC algorithm by facilitating the discovery of neighboring nodes in a master model part within a designated radius for each node in the slave model part. The process involves the following steps:

1. For each node in the slave model part, the utility searches for neighboring nodes in the master model part within a specified radius.
image
image
image

3. The utility then establishes an MPC between the slave node and its neighboring master nodes, and calculates the MPC weights based on a spatial function that employs radial basis functions.
image
image
4. The process repeats for all nodes in the slave model part.

**Added files:**

- `assign_mpcs_to_neighbours_utility.cpp`: This file contains all the functions to look for neighboring nodes and apply multipoint constraints (MPCs) using radial basis functions.
- `assign_mpcs_to_neighbours_utility.h`: This header file defines the functions and data structures used in assign_mpcs_to_neighbours_utility.cpp`.
- `assign_mpcs_to_neighbours_process.py`: This Python script calls the functions in `assign_mpcs_to_neighbours_utility.cpp` in the appropriate order to assign MPCs to neighboring nodes.
- `assign_mpcs_to_neighbours_process.mdpa`: This is an MDPA file used for testing the MPC algorithm.

**Modified files:**

- `add_geometrical_utilities_to_python.cpp`: This file was modified to export the functions in `assign_mpcs_to_neighbours_utility.cpp` to Python, so that they can be called from `assign_mpcs_to_neighbours_process.py`.
- `test_processes.py`: This file was modified to include a small test to validate the functionality of `assign_mpcs_to_neighbours_process.py`.


loumalouomega:
@KratosMultiphysics/altair and specially @ddiezrod

kratos/python_scripts/assign_mpcs_to_neighbours_process.py
```diff
@@ -0,0 +1,179 @@
...
+        self -- It signifies an instance of a class.
+        """
+        # If mpcs are updated every time step, these are to me removed before being re-assigned.
+        if self.assign_mpcs_every_time_step:
+            self.RemoveConstraints()
+        
+    def RemoveConstraints(self):
+        #Remove master-slave constraints
+        for constraint in self.computing_model_part.MasterSlaveConstraints:
+            constraint.Set(KM.TO_ERASE)
```
loumalouomega: Use VariableUtils

kratos/tests/auxiliar_files_for_python_unittest/mdpa_files/assign_mpcs_to_neighbours_process.mdpa
```diff
@@ -0,0 +1,3308 @@
+Begin Properties 0
```
loumalouomega: Please, can you consider a smaller mesh?, you can put this one in the Example repository

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+    SearchNodesInRadiusForNodes(pNodes, Radius, results, MinNumOfNeighNodes);
+
+    int prev_num_mpcs = rComputingModelPart.NumberOfMasterSlaveConstraints();
+
+    NodesContainerType::ContainerType& nodes_array     = const_cast<NodesContainerType::ContainerType&>(pNodes.GetContainer());
+
+    ModelPart::MasterSlaveConstraintType const& r_clone_constraint = KratosComponents<MasterSlaveConstraint>::Get("LinearMasterSlaveConstraint");
+    
+    #pragma omp parallel
```
loumalouomega: Please ParallelUtilities

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,211 @@
...
+    KRATOS_CATCH("");
+}
+
+void AssignMPCsToNeighboursUtility::GetDofsAndCoordinatesForNode(
+NodeType::Pointer pNode,
+const Variable<double>& rVariable,
+DofPointerVectorType& rCloud_of_dofs,
+array_1d<double,3>& rSlave_coordinates
+){
+KRATOS_TRY;
```
loumalouomega: Tabs missing (spaces)

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+// Get Dofs and Coordinates arrays for a given variable double. (For nodes)
+void AssignMPCsToNeighboursUtility::GetDofsAndCoordinatesForNodes(
+ResultNodesContainerType nodes_array,
+const Variable<double>& rVariable,
+DofPointerVectorType& rCloud_of_dofs,
+Matrix& rCloud_of_nodes_coordinates
+)
+{
+KRATOS_TRY;
```
loumalouomega: Tabs missing (spaces)

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,211 @@
...
+KRATOS_CATCH("");
+}
+
+void AssignMPCsToNeighboursUtility::AssignRotationToNodes(
+NodesContainerType pNodes,
+Matrix RotationMatrix
+)
+{
+//TODO: Do it in parallel
+KRATOS_TRY;
```
loumalouomega: Tabs missing (spaces)

loumalouomega (changes requested):
Some comments

SADPR:
I will use a smaller mesh for the test and update to Parallel Utilities.
I will upload this case to the Kratos Examples.
Assign_MPCs_to_neighbours

loumalouomega:
I cannot unsee:
image

image

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+        // Increment the counter variable
+        i++;
+
+        // Get Dofs and Coordinates
+        DofPointerVectorType r_cloud_of_dofs, r_slave_dof;
+        Matrix r_cloud_of_nodes_coordinates;
+        array_1d<double,3> r_slave_coordinates;
+        GetDofsAndCoordinatesForNode(pNode, rVariable, r_slave_dof, r_slave_coordinates);
+        KRATOS_WATCH(i)
```
rubenzorrilla:
```suggestion
```
Debugging leftover.

SADPR:
I've been trying to switch from OMP to `block_for_each` the implementation, and I did the following tests to have a counter "i" that could be thread-safe. The reason for using the "i" counter is to avoid repeating an MPC's Id.

- Using `std::atomic<int>`: This test showed repeating values in some threads, which means it's not thread-safe in this case.
- Using `std::atomic_flag`: This test showed repeating values in some threads as well, indicating it's not thread-safe either.
- Using a `std::mutex with std::lock_guard`: This test worked correctly but was around 50 times slower than using pragma omp.
- Using a `std::mutex with std::unique_lock`: This test did not work, as it resulted in repeating values in some threads.
- Using `#pragma omp parallel for`: This test worked correctly and was faster than using std::mutex with std::lock_guard.

Overall, it seems like using `std::mutex` with `std::lock_guard` is the only method that worked correctly for this particular use case, but it comes with a significant performance penalty. The pragma omp approach was faster, but you mentioned you wanted to avoid using OpenMP.

For the current implementation, I kept the original code using `#pragma omp` in the function AssignMPCsToNodes, and I added the implementation with `block_for_each` (std::mutex and std::lock_guard) in the function AssignMPCsToNodesParallelUtilities. Additionally, I have tried to use the `KRATOS_CRITICAL_SECTION` as a replacement for the `#pragma omp critical` section in order to add the MPCs thread-safely, but I haven't been successful yet. I am still looking for a way to substitute the OMP critical section for adding the MPCs thread-safely. If you can point me in the right direction, it would be greatly appreciated.

SADPR:
This PR is ready for a review. All requests were done. @rubenzorrilla @loumalouomega

kratos/tests/auxiliar_files_for_python_unittest/mdpa_files/test_assign_mpcs_to_neighbours_process.mdpa
```diff
@@ -0,0 +1,220 @@
...
+       42          0    10     8    13 
+       43          0    13     8    11 
+       44          0    16    14    10 
+       45          0    16    10    13 
+       46          0    16    13    20 
+       47          0     2     4     3 
+       48          0    17    16    20 
+End Elements
+
+Begin SubModelPart GENERIC_Domain // Group Domain // Subtree GENERIC
```
loumalouomega, line 103: If this submodelpart is not used you can simply removed to make the mdpa even smaller. I understand you only need master/slave ones

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+    // Add the constraints to the rComputingModelPart in a single call
+    rComputingModelPart.AddMasterSlaveConstraints(temp_constraints.begin(), temp_constraints.end());
+
+    KRATOS_INFO("AssignMPCsToNeighboursUtility") << "Build and Assign MPCs Time: " << build_and_assign_mpcs.ElapsedSeconds() << std::endl;
+    KRATOS_CATCH("");
+}
+
+
+
```
loumalouomega:
```suggestion

```

loumalouomega (commented):
Minor comments

kratos/tests/auxiliar_files_for_python_unittest/mdpa_files/test_assign_mpcs_to_neighbours_process.mdpa
SADPR: Since some applications require the MPCs to be applied to an specific model part (e.g. FluidDynamics -> fluid_computing_model_part). The process needs the name of this model part. In this case, GENERIC_Domain plays this role.

loumalouomega (approved):

rubenzorrilla (commented):
I've some comments about this. Will post them soon.

kratos/python_scripts/assign_mpcs_to_neighbours_process.py
```diff
@@ -0,0 +1,178 @@
+import KratosMultiphysics as KM
+import numpy as np
+
+def Factory(settings, model):
+    if not isinstance(settings, KM.Parameters):
+        raise Exception("expected input shall be a Parameters object, encapsulating a json string")
+    return AssignMPCsToNeighboursProcess(model, settings["Parameters"])
+
+## All the processes python should be derived from "Process"
+class AssignMPCsToNeighboursProcess(KM.Process):
```
rubenzorrilla, line 10: I'd call this `AssignMasterSlaveConstraintsToNeighboursProcess` to follow the naming convention of the rest of the Kratos (see the `ModelPart` for instance).

kratos/python_scripts/assign_mpcs_to_neighbours_process.py
```diff
@@ -0,0 +1,178 @@
...
+
+        Keyword arguments:
+        self -- It signifies an instance of a class.
+        Model -- the container of the different model parts.
+        settings -- Kratos parameters containing process settings.
+        """
+        KM.Process.__init__(self) # calling the baseclass constructor
+
+        default_settings = KM.Parameters("""{
+            "computing_model_part_name": "",
```
rubenzorrilla, line 32: `model_part_name` to be consistent with the rest of processes. Also note that not all the solvers in Kratos require a computing model part.

kratos/python_scripts/assign_mpcs_to_neighbours_process.py
```diff
@@ -0,0 +1,178 @@
...
+            "slave_model_part_name": "",
+            "master_model_part_name": "",
+            "variable_names": [],
+            "search_radius": 1.0,
+            "minimum_number_of_neighbouring_nodes": 3,
+            "assign_mpcs_every_time_step": false
```
rubenzorrilla, line 38: I think that `reform_constraints at_each_step` is more Kratos-like (see the `reform_dofs_at_each_step` in the solvers).

kratos/python_scripts/assign_mpcs_to_neighbours_process.py
```diff
@@ -0,0 +1,178 @@
...
+        }""")
+
+        # Add missing settings that the user did not provide but that
+        # are necessary for this process
+        settings.ValidateAndAssignDefaults(default_settings)
+        # Alternative:
+        # settings.RecursivelyValidateAndAssignDefaults(default_settings)
```
rubenzorrilla, line 45: Remove commented code.

kratos/python_scripts/assign_mpcs_to_neighbours_process.py
```diff
@@ -0,0 +1,178 @@
...
+        variable_names.sort()
+
+        self.variables_list = []
+        for var_name in variable_names:
+            if not KM.KratosGlobals.HasVariable(var_name):
+                err_msg = "\'{}\' variable in \'variable_names\' is not in KratosGlobals. Please check provided value.".format(var_name)
+            if not KM.KratosGlobals.GetVariableType(var_name):
+                err_msg = "\'{}\' variable in \'variable_names\' is not double type. Please check provide double type variables (e.g. [\"DISPLACEMENT_X\",\"DISPLACEMENT_Y\"]).".format(var_name)
+            self.variables_list.append(KM.KratosGlobals.GetVariable(var_name))
+        debug = True
```
rubenzorrilla, line 91: What is this?

kratos/python_scripts/assign_mpcs_to_neighbours_process.py
```diff
@@ -0,0 +1,178 @@
...
+    def ExecuteInitialize(self):
+        """ This method is executed at the begining to initialize the process
+
+        Keyword arguments:
+        self -- It signifies an instance of a class.
+        """
+        ## Assign master and slave model parts
+        self.computing_model_part = self.model.GetModelPart(self.computing_model_part_name)
+        self.slave_model_part = self.model.GetModelPart(self.slave_model_part_name)
+        self.master_model_part = self.model.GetModelPart(self.master_model_part_name)
```
rubenzorrilla, line 102: You have member variables for both model and model parts. Either one or the other. In this case it might be easier just to move these three lines to the constructor and avoid saving the model.

If you're doing this because you cannot assume that at construction time of this process the model will be filled, please put a comment explaining so, as this is not common for processes.

kratos/python_scripts/assign_mpcs_to_neighbours_process.py
```diff
@@ -0,0 +1,178 @@
...
+        """
+        pass
+
+    def ExecuteBeforeSolutionLoop(self):
+        """ This method is executed just before the solution-loop
+
+        Keyword arguments:
+        self -- It signifies an instance of a class.
+        """
+        pass
```
rubenzorrilla, line 126: Remove if not implemented.

kratos/python_scripts/assign_mpcs_to_neighbours_process.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def ExecuteAfterOutputStep(self):
+        """ This method is executed after writing the output (if output
+        is being written in this step)
+
+        Keyword arguments:
+        self -- It signifies an instance of a class.
+        """
+        pass
```
rubenzorrilla, line 155: Remove if not implemented.

kratos/python_scripts/assign_mpcs_to_neighbours_process.py
```diff
@@ -0,0 +1,178 @@
...
+        KM.VariableUtils().SetFlag(KM.TO_ERASE, True, self.computing_model_part.MasterSlaveConstraints)
+        self.computing_model_part.RemoveMasterSlaveConstraintsFromAllLevels(KM.TO_ERASE)
+
+    def ExecuteFinalize(self):
+        """ This method is executed after the computations, at the end of the solution-loop
+
+        Keyword arguments:
+        self -- It signifies an instance of a class.
+        """
+        pass
```
rubenzorrilla, line 178: Remove if not implemented.

kratos/python_scripts/assign_mpcs_to_neighbours_process.py
```diff
@@ -0,0 +1,178 @@
...
+        """ This method is executed in order to finalize the current step
+
+        Keyword arguments:
+        self -- It signifies an instance of a class.
+        """
+        # If mpcs are updated every time step, these are to me removed before being re-assigned.
+        if self.assign_mpcs_every_time_step:
+            self.RemoveConstraints()
+        
+    def RemoveConstraints(self):
```
rubenzorrilla, line 167: Make this "private" by mangling it. Otherwise you'd be extending the process API.

kratos/python/add_geometrical_utilities_to_python.cpp
```diff
@@ -432,7 +432,18 @@ void AddGeometricalUtilitiesToPython(pybind11::module &m)
         .def_static("CalculateShapeFunctions", [](const Matrix& rPoints, const array_1d<double,3>& rX, const double h, Vector& rN, DenseQRPointerType pDenseQR){
             return RBFShapeFunctionsUtility::CalculateShapeFunctions(rPoints, rX, h, rN, pDenseQR);})
         ;
+
```
rubenzorrilla, line 435: See my comment about the Kratos naming (change `MPC` to `MasterSlaveConstraint`).

kratos/utilities/assign_mpcs_to_neighbours_utility.h
```diff
@@ -0,0 +1,252 @@
...
+      using NodeType = ModelPart::NodeType;
+
+      //Dof types
+      using DofType = NodeType::DofType;
+
+      //Configure Types
+      using NodesContainerType = ModelPart::NodesContainerType;
+
+      //Bin
+      using NodeBinsType = BinsDynamic<3, NodeType, NodesContainerType::ContainerType>;
```
rubenzorrilla, line 65: Shouldn't the first template argument be 2 for the 2D case?

kratos/utilities/assign_mpcs_to_neighbours_utility.h
```diff
@@ -0,0 +1,252 @@
...
+      using DofPointerVectorType = std::vector< Dof<double>::Pointer >;
+
+      /// Pointer definition of AssignMPCsToNeighboursUtility
+      KRATOS_CLASS_POINTER_DEFINITION(AssignMPCsToNeighboursUtility);
+
+      ///@}
+      ///@name Life Cycle
+      ///@{
+
+      /// Default constructor.
```
rubenzorrilla, line 82: This is not the default constructor. Indeed, I'd delete the default constructor (it's being automatically added by the compiler).

kratos/utilities/assign_mpcs_to_neighbours_utility.h
```diff
@@ -0,0 +1,252 @@
...
+      ///@name Member Variables
+      ///@{
+      Kratos::unique_ptr<NodeBinsType> mpBins;
+      int mMaxNumberOfNodes;
+
+      /// Assignment operator.
+      AssignMPCsToNeighboursUtility& operator=(AssignMPCsToNeighboursUtility const& rOther)
+      {
+          return *this;
+      }
```
rubenzorrilla, line 249: I'd remove the assignment operator.

kratos/utilities/assign_mpcs_to_neighbours_utility.h
```diff
@@ -0,0 +1,252 @@
...
+
+      /// Print object's data.
+      virtual void PrintData(std::ostream& rOStream) const  {}
+
+      ///@}
+
+    private:
+      ///@name Member Variables
+      ///@{
+      Kratos::unique_ptr<NodeBinsType> mpBins;
```
rubenzorrilla, line 242: NodeBinsType::UniquePointer

kratos/utilities/assign_mpcs_to_neighbours_utility.h
```diff
@@ -0,0 +1,252 @@
...
+        DofPointerVectorType& rCloud_of_dofs,
+        array_1d<double,3>& rSlave_coordinates
+        );
+
+      void GetDofsAndCoordinatesForNode(
+        NodeType::Pointer pNode,
+        const Variable<array_1d<double, 3>>& rVariable,
+        DofPointerVectorType& rCloud_of_dofs_x,
+        DofPointerVectorType& rCloud_of_dofs_y,
+        DofPointerVectorType& rCloud_of_dofs_z,
```
rubenzorrilla, line 158: Please follow the style guide. CamelCase for method arguments.

kratos/utilities/assign_mpcs_to_neighbours_utility.h
```diff
@@ -0,0 +1,252 @@
...
+        );
+
+      /**
+       * @brief Assign rotation to nodes
+       * @details Assigns the given rotation matrix to the nodes' positions and updates their coordinates.
+       * @param pNodes Nodes container with the nodes to apply the rotation.
+       * @param RotationMatrix 3x3 rotation matrix to be applied to the nodes' positions.
+       **/
+      void AssignRotationToNodes(
+          NodesContainerType pNodes,
```
rubenzorrilla, line 186: This should be a reference. Indeed it is strange to me that this works if you pass it by copy...

kratos/utilities/assign_mpcs_to_neighbours_utility.h
```diff
@@ -0,0 +1,252 @@
...
+       * @param rStructureNodes Nodes Container.
+       * @param rNodes Nodes to obtain their respective neighbouring nodes.
+       * @param Radius Search radius.
+       * @param rResults Results container.
+       * @param MinNumOfNeighNodes Minimum Number of Neighbouring nodes (minimum cloud of nodes)
+       **/
+
+      // Search for the neighbouring nodes (in rStructureNodes) of each rNode on a given array of rNodes
+      void SearchNodesInRadiusForNodes(
+          NodesContainerType const& rNodes,
```
rubenzorrilla, line 105: Though it is valid, in Kratos we normally put the `const` in front.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+    KRATOS_TRY;
+    NodesContainerType::ContainerType& nodes_model_part = rStructureNodes.GetContainer();
+    mpBins = Kratos::make_unique<NodeBinsType>(nodes_model_part.begin(), nodes_model_part.end());
+    mMaxNumberOfNodes = rStructureNodes.size();
+    KRATOS_CATCH(""); 
+} 
+
+// Search for the neighbouring nodes (in rStructureNodes) of each rNode on a given array of rNodes
+void AssignMPCsToNeighboursUtility::SearchNodesInRadiusForNodes(
+    NodesContainerType const& rNodes,
```
rubenzorrilla, line 51: Normally in Kratos we put the `const` in front.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+    const double Radius,
+    VectorResultNodesContainerType& rResults,
+    const double MinNumOfNeighNodes)
```
rubenzorrilla, line 54: It is a good programming practice to put the results arguments after the input (const ones).

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+{
+    KRATOS_TRY;
+    NodesContainerType::ContainerType& nodes_array     = const_cast<NodesContainerType::ContainerType&>(rNodes.GetContainer());
```
rubenzorrilla, line 57: `auto& r_nodes_array`

rubenzorrilla: And why you need the `const_cast`?

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+    rResults.resize(nodes_array.size());
+
+    // Declare a counter variable outside the loop as std::atomic<int>
+    std::atomic<int> i(0);
+
+    block_for_each(nodes_array, [&](Node<3>::Pointer& pNode)
+    {
+        double localRadius = Radius;
```
rubenzorrilla, line 66: Please follow the style guide (`local_radius`).

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+void AssignMPCsToNeighboursUtility::SearchNodesInRadiusForNodes(
+    NodesContainerType const& rNodes,
+    const double Radius,
+    VectorResultNodesContainerType& rResults,
+    const double MinNumOfNeighNodes)
+{
+    KRATOS_TRY;
+    NodesContainerType::ContainerType& nodes_array     = const_cast<NodesContainerType::ContainerType&>(rNodes.GetContainer());
+
+    rResults.resize(nodes_array.size());
```
rubenzorrilla, line 59: This means you're allocating much memory than needed. I'd do a `shrink_to_fit` at the end of this function to avoid so.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+            localRadius += Radius;
+        }
+    });
+    KRATOS_CATCH("");
+}
+
+
+//Search for the neighbouring nodes (in rStructureNodes) of the given rNode
+void AssignMPCsToNeighboursUtility::SearchNodesInRadiusForNode(
+    NodeType::Pointer pNode,
```
rubenzorrilla, line 82: I'd pass a constant reference instead. Note that the `SearchInRadius` supports it.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+const Variable<double>& AssignMPCsToNeighboursUtility::GetComponentVariable(
+    const Variable<array_1d<double, 3>>& rVectorVariable, 
+    const std::size_t ComponentIndex)
+{
+    static const std::array<std::string, 3> component_suffixes = {"_X", "_Y", "_Z"};
+
+    if (ComponentIndex < 3) {
+        std::string component_variable_name = rVectorVariable.Name() + component_suffixes[ComponentIndex];
+        return KratosComponents<Variable<double>>::Get(component_variable_name);
```
rubenzorrilla, line 106: This is extremely expensive and being done, if I'm not wrong, inside a loop. An alternative approach must be used.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+void AssignMPCsToNeighboursUtility::GetDofsAndCoordinatesForNode(
+    NodeType::Pointer pNode,
+    const Variable<double>& rVariable,
+    DofPointerVectorType& rCloud_of_dofs,
+    array_1d<double,3>& rSlave_coordinates
+    )
+{
+    KRATOS_TRY;
+    rCloud_of_dofs.push_back(pNode->pGetDof(rVariable));
+    rSlave_coordinates = pNode->Coordinates();
```
rubenzorrilla, line 122: I don't see the point on doing a function of two lines. IMO it only makes more complex browsing the code with no advantage at all.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+    DofPointerVectorType& rCloud_of_dofs_z,
+    array_1d<double,3>& rSlave_coordinates
+    )
+{
+    KRATOS_TRY;
+
+    // Check if the node has the required DOFs
+    if (pNode->HasDofFor(GetComponentVariable(rVariable, 0)) &&
+        pNode->HasDofFor(GetComponentVariable(rVariable, 1)) &&
+        pNode->HasDofFor(GetComponentVariable(rVariable, 2))) {
```
rubenzorrilla, line 140: What would happen in the 2D case in which the third component DOF might not be present?

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+    KRATOS_CATCH("");
+}
+
+void AssignMPCsToNeighboursUtility::GetDofsAndCoordinatesForNode(
+    NodeType::Pointer pNode,
+    const Variable<array_1d<double, 3>>& rVariable,
+    DofPointerVectorType& rCloud_of_dofs_x,
+    DofPointerVectorType& rCloud_of_dofs_y,
+    DofPointerVectorType& rCloud_of_dofs_z,
+    array_1d<double,3>& rSlave_coordinates
```
rubenzorrilla, line 132: Please follow the style guide for function arguments.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+    for(unsigned int i = 0; i < nodes_array.size(); i++){
+        rCloud_of_dofs[i] = nodes_array[i]->pGetDof(rVariable);
+        noalias(row(rCloud_of_nodes_coordinates,i)) = nodes_array[i]->Coordinates();
+    }
+    KRATOS_CATCH("");
+}
+
+
+void AssignMPCsToNeighboursUtility::GetDofsAndCoordinatesForNodes(
+    ResultNodesContainerType nodes_array,
```
rubenzorrilla, line 173: Why passing by copy?

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+    DofPointerVectorType& rCloud_of_dofs_y,
+    DofPointerVectorType& rCloud_of_dofs_z,
+    Matrix& rCloud_of_nodes_coordinates
+)
+{
+    KRATOS_TRY;
+    rCloud_of_dofs_x.resize(nodes_array.size());
+    rCloud_of_dofs_y.resize(nodes_array.size());
+    rCloud_of_dofs_z.resize(nodes_array.size());
+    rCloud_of_nodes_coordinates.resize(nodes_array.size(), 3);
```
rubenzorrilla, line 185: I'd try to shrink these containers afterwards.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+    for(int i = 0; i < static_cast<int>(nodes_array.size()); i++)
+    {
+        NodeType::Pointer pNode = nodes_array[i];
+
+        // Check if the node has the required DOFs
+        if (pNode->HasDofFor(GetComponentVariable(rVariable, 0)) &&
+            pNode->HasDofFor(GetComponentVariable(rVariable, 1)) &&
+            pNode->HasDofFor(GetComponentVariable(rVariable, 2))) {
```
rubenzorrilla, line 194: Same comment about the 2D third component.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+}
+
+void AssignMPCsToNeighboursUtility::AssignRotationToNodes(
+    NodesContainerType pNodes,
+    Matrix RotationMatrix
+    )
+{
+    //TODO: Do it in parallel
+    KRATOS_TRY;
+    NodesContainerType::ContainerType& nodes_array     = const_cast<NodesContainerType::ContainerType&>(pNodes.GetContainer());
```
rubenzorrilla, line 215: `auto& r_nodes_array`

rubenzorrilla: Why the `const_cast`?

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+    // Assign rotation to given nodes
+    for(unsigned int i = 0; i < nodes_array.size(); i++){
+        auto coordinate_vector = nodes_array[i]->GetInitialPosition().Coordinates();
```
rubenzorrilla, line 219: You are allocating/deallocating an unnecessary copy for each node.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+        Vector rotated_position(coordinate_vector.size());
```
rubenzorrilla, line 220: Can't this be an `array_1d<double,3>`?

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+        noalias(rotated_position) = prod(RotationMatrix,coordinate_vector);
+        nodes_array[i]->X() = rotated_position[0];
+        nodes_array[i]->Y() = rotated_position[1];
+        nodes_array[i]->Z() = rotated_position[2];
```
rubenzorrilla, line 224: `noalias(nodes_array[i]->Coordinates()) = rotated_position;` I think this should work.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+    KRATOS_CATCH("");
+}
+
+void AssignMPCsToNeighboursUtility::AssignMPCsToNodes(
+    NodesContainerType pNodes,
+    double const Radius,
+    ModelPart& rComputingModelPart,
+    const Variable<array_1d<double, 3>>& rVariable,
+    double const MinNumOfNeighNodes
+    )
```
rubenzorrilla, line 308: This is basically copy & paste of the above function but modifying the section in which the Master-Slave Constraints are created. I'd do the creation within a private auxiliary function and have a unique implementation of `AssignMPCsToNodes`.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+    // Dequeue the constraints from the concurrent queue and add them to the temporary container
+    while (concurrent_constraints.try_dequeue(p_constraint)) {
+        temp_constraints.push_back(p_constraint);
+    }
+
+    // Add the constraints to the rComputingModelPart in a single call
+    rComputingModelPart.AddMasterSlaveConstraints(temp_constraints.begin(), temp_constraints.end());
+
+    KRATOS_INFO("AssignMPCsToNeighboursUtility") << "Build and Assign MPCs Time: " << build_and_assign_mpcs.ElapsedSeconds() << std::endl;
```
rubenzorrilla, line 298: This may result in a too verbose output. If you want to output this, I'd introduce an echo level.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+    ModelPart& rComputingModelPart,
+    const Variable<double>& rVariable,
+    double const MinNumOfNeighNodes
+    )
+{
+    KRATOS_TRY;
+
+    BuiltinTimer build_and_assign_mpcs;
+
+    int prev_num_mpcs = rComputingModelPart.NumberOfMasterSlaveConstraints();
```
rubenzorrilla, line 241: In here you are assuming that your new MPCs are not interacting with previous ones. Though this might be a valid assumption, I'd throw a warning message if the previous number of MPCs is non-zero to inform about this.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+    NodesContainerType::ContainerType& nodes_array = const_cast<NodesContainerType::ContainerType&>(pNodes.GetContainer());
```
rubenzorrilla, line 243: `auto& r_nodes_array`.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+    ModelPart::MasterSlaveConstraintType const& r_clone_constraint = KratosComponents<MasterSlaveConstraint>::Get("LinearMasterSlaveConstraint");
```
rubenzorrilla, line 245: const auto& r_clone_constraint`.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+        while (r_result.size() < MinNumOfNeighNodes) {
+            r_result.clear();
+            SearchNodesInRadiusForNode(pNode, local_radius, r_result);
+            local_radius += Radius;
+        }
+
+        // Get Dofs and Coordinates
+        DofPointerVectorType r_cloud_of_dofs, r_slave_dof;
+        Matrix r_cloud_of_nodes_coordinates;
+        array_1d<double, 3> r_slave_coordinates;
```
rubenzorrilla, line 270: You're allocating/deallocating these for each node. Declare them as a TLS outside the loop.

kratos/tests/test_processes.py
```diff
@@ -2521,6 +2521,84 @@ def test_split_internal_interfaces_process_3D(self):
...
+                    ]
+                }
+            """
+        )
+
+        # Create and execute processes
+        list_of_processes = process_factory.KratosProcessFactory(current_model).ConstructListOfProcesses(settings["process_list"])
+        for process in list_of_processes:
+            process.ExecuteInitialize()
+            process.ExecuteInitializeSolutionStep()
```
rubenzorrilla, line 2562: You have a unique process. I don't see the point on doing it with a list. I'd directly call the constructor.

kratos/tests/test_processes.py
```diff
@@ -2521,6 +2521,84 @@ def test_split_internal_interfaces_process_3D(self):
...
+            return all(abs(ew - ow) <= delta for ew, ow in zip(expected_weights, obtained_weights))
+
+        # Compare each expected_weights to obtained_weights, ensuring that there is a match
+        for expected_weights in expected_constraints_weights:
+            matching_found = False
+
+            for obtained_weights in obtained_constraints_weights:
+                if check_matching_weights(expected_weights, obtained_weights):
+                    for ew, ow in zip(expected_weights, obtained_weights):
+                        self.assertAlmostEqual(ew, ow, delta=1e-6)  # Assert the weights are equal within the specified tolerance
```
rubenzorrilla, line 2594: If I'm not wrong, there is an `assertAlmostEqualVector` or something of this sort.

rubenzorrilla (commented):
Though this is a working implementation, I think it doesn't suffice the quality requirements for being in the `KratosCore`. I hope you take my comments into consideration in a new PR.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+        } else {
+            KRATOS_ERROR << "The node with ID " << pNode->Id() << " does not have the required DOFs for the variable " << rVariable << std::endl;
+        }
+        noalias(row(rCloud_of_nodes_coordinates, i)) = pNode->Coordinates();
+    }
+    KRATOS_CATCH("");
+}
+
+void AssignMPCsToNeighboursUtility::AssignRotationToNodes(
```
rubenzorrilla, line 208: This is not used anywhere in this process. Besides, it is a custom function for your target application. Long story short, it must be removed from here to be placed somewhere else.

kratos/python_scripts/assign_mpcs_to_neighbours_process.py
SADPR: Debugging leftover.

kratos/utilities/assign_mpcs_to_neighbours_utility.h
SADPR: Not used anymore.

kratos/tests/test_processes.py
SADPR: assertVectorAlmostEqual

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
SADPR: Resolved, changed to:
```python
    KRATOS_TRY;
    auto& r_nodes_array = rNodes.GetContainer();

    rResults.resize(r_nodes_array.size());

    // Declare a counter variable outside the loop as std::atomic<int>
    std::atomic<int> i(0);

    block_for_each(r_nodes_array, [&](const Node<3>::Pointer& pNode)
    {
```

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
SADPR: I am allocating what I need. Nonetheless, I could updated it to this:
```cpp
    // Resize rResults vector
    rResults.resize(r_nodes_array.size());

    // Declare a counter variable outside the loop as std::atomic<int>
    std::atomic<int> i(0);

    block_for_each(r_nodes_array, [&](const Node<3>::Pointer& pNode)
    {
        double local_radius = Radius;

        IndexType it = i.fetch_add(1); // Atomically increment the counter and get the previous value

        while (rResults[it].size()<MinNumOfNeighNodes){
            rResults[it].clear();
            SearchNodesInRadiusForNode(pNode, local_radius, rResults[it]);
            local_radius += Radius;
        }
    });

    // Optimize memory usage
    rResults.shrink_to_fit();

File: kratos/utilities/assign_mpcs_to_neighbours_utility.cpp (hunk 33 above)
Comment (SADPR) [reply kratos/utilities/assign_mpcs_to_neighbours_utility.cpp:82]: Do you find it necesarry, since this is a constant pointer. Isn't it okay like this?

File: kratos/utilities/assign_mpcs_to_neighbours_utility.cpp (hunk 34 above)
Comment (SADPR) [reply kratos/utilities/assign_mpcs_to_neighbours_utility.cpp:106]: After a long afternoon, I managed to do it. Now it only acces the Kratos components once and has a pointer in the loop.

File: kratos/utilities/assign_mpcs_to_neighbours_utility.cpp (hunk 35 above)
Comment (SADPR) [reply kratos/utilities/assign_mpcs_to_neighbours_utility.cpp:122]: It is to preserve similarity. Is it really necessary to undo this function? I know it is not typical to have a two line function, but for me, it preserves similarity.

File: kratos/utilities/assign_mpcs_to_neighbours_utility.cpp (hunk 36 above)
Comment (SADPR) [reply kratos/utilities/assign_mpcs_to_neighbours_utility.cpp:140]: Is this really something that happens? 
The user can also specify ["DISPLACEMENT_X","DISPLACEMENT_Y"].

Comment (SADPR) [reply kratos/utilities/assign_mpcs_to_neighbours_utility.cpp:140]: I thought the 2D also 3 components. But solves only for X and Y components.

File: kratos/utilities/assign_mpcs_to_neighbours_utility.cpp (hunk 50 above)
Comment (SADPR) [reply kratos/utilities/assign_mpcs_to_neighbours_utility.cpp:270]: Master-Slave constraints require rSlaveDofs to be a DofPointerVectorType. To do r_slave_dofs thread_local I need to resize and assign to Dof<double>::Pointer. Or:
rCloudOfDofs.resize(1); 
rCloudOfDofs[0] = pNode->pGetDof(rVariable);
Which I dont like. I will fix it.

File: kratos/utilities/assign_mpcs_to_neighbours_utility.cpp (hunk 46 above)
Comment (SADPR) [reply kratos/utilities/assign_mpcs_to_neighbours_utility.cpp:298]: TODO.

File: kratos/utilities/assign_mpcs_to_neighbours_utility.cpp (hunk 45 above)
Comment (SADPR) [reply kratos/utilities/assign_mpcs_to_neighbours_utility.cpp:308]: I have a concern here, at first I though of the user passing a variable which could be either array_1d or double.
If array_1d, then you have to create the variable components lists, etc. If double, you can directly retrieve the varible with GetDof. Now I am thinking to always create a list of variables, e.g, if ["VELOCITY","PRESSURE], then create a list:
["DISPLACEMENT_X","DISPLACEMENT_Y","PRESSURE"] for 2D.
Or:
["DISPLACEMENT_X","DISPLACEMENT_Y","DISPLACEMENT_Z","PRESSURE"] for 3D.
We can discuss this in person, since it is taking a lot of time.

File: kratos/utilities/assign_mpcs_to_neighbours_utility.cpp (hunk 33 above)
Comment (rubenzorrilla) [reply kratos/utilities/assign_mpcs_to_neighbours_utility.cpp:82]: I don't think it is a constant pointer but nevermind, leave it like this.

File: kratos/utilities/assign_mpcs_to_neighbours_utility.cpp (hunk 32 above)
Comment (rubenzorrilla) [reply kratos/utilities/assign_mpcs_to_neighbours_utility.cpp:59]: No. You are allocating much more than what you need since you're resizing the vector to the number of nodes, which would only be the case if all the nodes are found. Indeed, the proper usage would be
```c++
rResults.reserve(r_nodes_array.size());

// ... your stuff in here ...

rResults.shrink_to_fit();
```

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
rubenzorrilla: Depends on the implementation so you cannot take it as granted.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
rubenzorrilla: I don't get your point on "similarity"... I'd remove it but if you really require it leave it. Up to you.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
rubenzorrilla: (same `reseve` and `shrink_to_fit` comment).

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
rubenzorrilla: OK.

kratos/utilities/assign_mpcs_to_neighbours_utility.cpp
SADPR: I am changing the names of the functions to make them more self-explanatory. In this function, we retrieve the degrees of freedom (DOFs) and coordinates of the cloud of nodes. Therefore, we already know the size.

PR #11085, opened 2023-05-05
[GeoMechanicsApplcation] Add toggle in to select experimental mode

### LINKED ISSUE #11084: [GeoMechanicsApplication] Toggle to select experimental behavior
We would like to introduce a toggle to distinguish between verified and research(/experimental) behavior. Default behavior must be safe to accommodate less experienced users. More experienced users can flip the toggle when they want to use not yet fully verified elements.

The suggested change will only impact the GeoApplication.

A selection mechanism could be added to the Kratos Core if other applications want a similar mechanism. Having it in the core would enable better integration. Like let the core reply/return if research elements, conditions, etc are used in the current model. When using elements from different applications this can be even more helpful. For example, I want to research or improve application X while only using verified behavior from applications Y and Z.

huhaas:
**📝 Description**
Add a toggle to select experimental mode from the Python interface.
The registration of the application happens in the python `__init__.py` file. Which limits control over the code flow.
Selection based on an environmental variable is not ideal but the least ugly variation I could think of. Python does not like global variable sharing over module boundaries. This creation keeps it local to the GeoMechanicsApplication. 
Any suggestions on how to improve are welcome.

Note: Which elements, conditions, or laws are experimental is not decided. This PR is just the mechanism.

### DIFF
--- a/applications/GeoMechanicsApplication/GeoMechanicsApplication.py
+++ b/applications/GeoMechanicsApplication/GeoMechanicsApplication.py
@@ -1,8 +1,14 @@
+import os
+
 # Application dependent names and paths
 from KratosMultiphysics import _ImportApplication
 import KratosMultiphysics.StructuralMechanicsApplication
 from KratosGeoMechanicsApplication import *
 application = KratosGeoMechanicsApplication()
 application_name = "KratosGeoMechanicsApplication"
+
+geo_experimental = os.getenv('GEO_EXPERIMENTAL')
+if geo_experimental:
+    application.SetExperimental()
 
 _ImportApplication(application, application_name)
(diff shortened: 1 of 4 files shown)


mcgicjn2 (changes requested):
Although I like the concept and idea, I think  we need to design it from the opposite view point, i.e. we activate 'GEO_SUPPORTED', functionality. 

My reasoning is generally 2 fold: 

- The master of Kratos by its nature is research\experimentally based, and as such it would appear odd to have to switch on such functionality to a typical researcher.
- By inverting the flag we put the onus on our team to add the flag around supported functionality, and equally to deal with the flag in compilation

Furthermore we must discuss this further now in the context of CIMNE's proposed potential changes, in terms of 2 tiers of application; in essence this functionality would create 4 tiers

PR #11056, opened 2023-04-26
[Core] Remove template argument from Node

loumalouomega:
**📝 Description**

This PR removes template argument (both dimension and DoF). @pooyan-dadvand ask me to put as reviewers @rubenzorrilla and @roigcarlo 

Many files involved:

1. Remove template from `node.h` (only file I suggest to look in depth, changes reduced to minimum)
2. Redefine declarations

**TODO**: 
- [ ] Part of the code should be moved to source file to reduce the compilation times. (future PR after speaking with @pooyan-dadvand)
- [ ] We should also at least explicitly instantiate `Dof<double>` in order to reduce compilation time

**🆕 Changelog**

- Remove dimension template argument
- Remove DoF template argument
- Replace definitions

PR #11086, opened 2023-05-05
[Core] Separate header source `point.h`

loumalouomega:
**📝 Description**

This PR separates `Point` into header and source, in order to reduce compilation time. We are already doing the changes to separate `Node` into header and source (see https://github.com/KratosMultiphysics/Kratos/pull/11056) to reduce compilation time, and I just realized that `Point`, despite being templateless was still header only. In addition, I added some missing documentation.

**🆕 Changelog**

- [Separate header source `point.h`. Main changes in this commit:](https://github.com/KratosMultiphysics/Kratos/commit/154f25f9b0f8ea7399246e27c0603afaeb87bf3e)
	- Separation of function implementations from the header file to the source file, `point.cpp`	
	- Introduction of comments and documentation for the class and its member functions.
	- Improved code readability by adding whitespace and newlines.
	- Modifications in the way constructors and some other functions are written for better clarity and consistency.
- [Missing KRATOS_API(KRATOS_CORE)](https://github.com/KratosMultiphysics/Kratos/pull/11086/commits/22df46732be9e4710d0478ee16b2d3efb1b3d9a2)
- [Update guard](https://github.com/KratosMultiphysics/Kratos/pull/11086/commits/a885b223457e9236941270de7d262502e0cadef6) : Replace include guard with `#pragma once` in `array_1d.h`


loumalouomega:
@roigcarlo any idea why the windows is failing (only the one without unity)?

roigcarlo:
Honestly, I see nothing wrong. I will try to compile tomorrow in win to see what is happening...

loumalouomega:
@roigcarlo in my Windows machine it works, I don't know how to reproduce this

huhaas:
@loumalouomega which configuration are you using on your local Windows machine? Custom, FullDebug, or regular Debug/Release?

loumalouomega:
> @loumalouomega which configuration are you using on your local Windows machine? Custom, FullDebug, or regular Debug/Release?

Regular release, I am trying to replicate Custom, but I cannot find the definition

roigcarlo:
I still have no idea about the error. On the other hand... 

We have been reviewing this with @RiccardoRossi abd we are not sure it is a good idea to split this particular class.

Our concern is that in this particular case, all the functions are simple enough so the compiler is able to inline (we are making sure it is the case) and by moving the implementation we are hiding the possibility of inline from any transactional unit different from the one in the core.

loumalouomega:
Gcc usually just inlines very small functions. The class is included in
many places and therefore is compiled again and again, separate it will
help in reducing compilation times.

philbucher:
I guess some measurements of compilation times and runtimes could help to make a decision here ;)

roigcarlo:
Yep it will help compilation times, the problem is that from source you can only inline into other header/sources of the same transactional unit. So it will be impossible for the compiler to inline outside the core.

roigcarlo:
Yep it will help compilation times, the problem is that from source you can only inline into other header/sources of the same transactional unit. So it will be impossible for the compiler to inline outside the core.

> it would be interesting to add a macro to force inlining

mmm we can take a loot at it sure :+1:

pooyan-dadvand:
@KratosMultiphysics/technical-committee applies the same comment as in the #11119 which for data classes that are performance critical, the priority is inlining.

loumalouomega:
> > Yep it will help compilation times, the problem is that from source you can only inline into other header/sources of the same transactional unit. So it will be impossible for the compiler to inline outside the core.
> 
> Okay, I will accept it, but it would be interesting to add a macro to force inlining (source [en.wikipedia.org/wiki/Inline_function](https://en.wikipedia.org/wiki/Inline_function)):
> 
> ```c++
> #ifdef _MSC_VER
>     #define forceinline __forceinline
> #elif defined(__GNUC__)
>     #define forceinline inline __attribute__((__always_inline__))
> #elif defined(__CLANG__)
>     #if __has_attribute(__always_inline__)
>         #define forceinline inline __attribute__((__always_inline__))
>     #else
>         #define forceinline inline
>     #endif
> #else
>     #define forceinline inline
> #endif
> ```
> 
> Otherwise, inline is not guaranteed as GCC is quite conservative with inlining.

This should be considered, otherwise the compiler will not inline in most cases

PR #11094, opened 2023-05-09
[Poro] Fixed rotation matrix sign in 2D interface elements

ipouplana:
**📝 Description**
Some months ago, we had to change the sign of the 2D interface element rotation matrix because GiD generated the 2D interface elements mesh with a clockwise numbering. Now the numbering in GiD is counter-clockwise. In this PR we reverse the previous change of sign to be consistent with respect to the GiD numbering.

**🆕 Changelog**
- The sign of the second row of the 2D interface element rotation matrix has been changed. This change has been applied in two methods in the Poromechanics Application and in another method in the Dam Application.


joaquinirazabal (approved):
Thank you, @ipouplana. I hope it is the last change in the node numbering...

PR #11095, opened 2023-05-09
[OptApp] Restructuring 21 - Implement implicit/Helmholtz filtering using AnalysisStage 

RezaNajian:
**📝 Description**
This PR restructures implicit Helmholtz filters to be used for general scalar and vector fields. 
The goal is to make them standalone and inherit from AnalysisStage to be compatible with Kratos.

**🆕 Changelog**
It would not break the current workflow however later it will be integrated in shape, thickness and topology controls. when the controls are ready.
E.g.
- Adds HelmholtzAnalysis(ImplicitFiltering), inherited from AnalysisStage
- Adds tests for "OpenMP" runs, "MPI" will be covered in a separate PR


applications/OptimizationApplication/python_scripts/filtering/implicit_filtering_analysis.py
```diff
@@ -0,0 +1,62 @@
...
+class ImplicitFiltering(AnalysisStage):
+    """
+    This class is the main-script of the implicit filtering analysis
+    It can be imported and used as "black-box"
+    """
+    def __init__(self, model, project_parameters):
+        # Making sure that older cases fail with a meaningful error message
+        solver_settings = project_parameters["solver_settings"]
+        if not solver_settings.Has("time_stepping"):
+            raise Exception("Using the old way to pass the time_step, this was removed!")
```
Igarizza: Use the Kratos::Deprecation mechanisms.

applications/OptimizationApplication/python_scripts/filtering/implicit_filtering_analysis.py
```diff
@@ -0,0 +1,62 @@
...
+            raise Exception('"solver_settings" does not have "echo_level", please add it!')
+
+        super().__init__(model, project_parameters)
+
+    #### Internal functions ####
+    def _CreateSolver(self):
+        """ Create the Solver (and create and import the ModelPart if it is not alread in the model) """
+        return implicit_filter_solvers.CreateSolver(self.model, self.project_parameters)
+
+    def _CreateProcesses(self, parameter_name, initialization_order):
```
Igarizza: I would use the base class `_CreateProcesses`

applications/OptimizationApplication/python_scripts/filtering/implicit_filtering_analysis.py
```diff
@@ -0,0 +1,62 @@
...
+                raise Exception("Using the old way to create the gid-output, this was removed!")
+        else:
+            raise NameError("wrong parameter name")
+
+        return list_of_processes
+
+    def _GetSimulationName(self):
+        return "::[Implicit Helmholtz/Sobolev Filtering]:: "
+    
+    def FilterField(self, unfiltered_field: KM.ContainerExpression.NodalNonHistoricalExpression) -> KM.ContainerExpression.NodalNonHistoricalExpression:
```
Igarizza: I would pass generic field, and do the conversion to nodal values within this function such that we can filter any field.

applications/OptimizationApplication/custom_elements/helmholtz_shell_element.h
```diff
@@ -0,0 +1,332 @@
...
+#include "utilities/geometry_utilities.h"
+#include "includes/ublas_interface.h"
+#include "includes/variables.h"
+#include "optimization_application_variables.h"
+
+
+namespace Kratos
+{
+
+///@name Kratos Globals
```
Igarizza: remove empty groups

applications/OptimizationApplication/custom_elements/helmholtz_shell_element.h
```diff
@@ -0,0 +1,332 @@
...
+
+/// Short class definition.
+/**
+ * @class HelmholtzShellElement
+ * @ingroup OptimizationApplication
+ * @brief Helmholtz filtering element for 3D geometries.
+ * @details Implements Sobolev/Helmholtz filter for filtering/regularization. This works for arbitrary geometries in 3D
+ * @author Reza Najian Asl
+ */
+class KRATOS_API(OPTIMIZATION_APPLICATION) HelmholtzShellElement
```
Igarizza:
```suggestion
class KRATOS_API(OPTIMIZATION_APPLICATION) HelmholtzSurfaceElement
```

Igarizza:
```suggestion
class KRATOS_API(OPTIMIZATION_APPLICATION) HelmholtzSurfaceElement
```

RezaNajian:
@sunethwarna @Igarizza  It is ready for being reviewed.

applications/OptimizationApplication/optimization_application_variables.h
```diff
@@ -165,6 +165,15 @@ namespace Kratos
     KRATOS_DEFINE_APPLICATION_VARIABLE(OPTIMIZATION_APPLICATION, double, YOUNG_MODULUS_SENSITIVITY);
     KRATOS_DEFINE_APPLICATION_VARIABLE(OPTIMIZATION_APPLICATION, double, POISSON_RATIO_SENSITIVITY);
 
+    // For helholtz solvers
+    KRATOS_DEFINE_VARIABLE( bool, COMPUTE_HELMHOLTZ_INVERSE);
+    KRATOS_DEFINE_VARIABLE( bool, HELMHOLTZ_INTEGRATED_FIELD);
+    KRATOS_DEFINE_VARIABLE( double, HELMHOLTZ_RADIUS);
+    KRATOS_DEFINE_VARIABLE( double, HELMHOLTZ_SCALAR);
+    KRATOS_DEFINE_VARIABLE( double, HELMHOLTZ_SCALAR_SOURCE);
```
sunethwarna:
```suggestion
    KRATOS_DEFINE_APPLICATION_VARIABLE(OPTIMIZATION_APPLICATION, bool, COMPUTE_HELMHOLTZ_INVERSE);
    KRATOS_DEFINE_APPLICATION_VARIABLE(OPTIMIZATION_APPLICATION, bool, HELMHOLTZ_INTEGRATED_FIELD);
    KRATOS_DEFINE_APPLICATION_VARIABLE(OPTIMIZATION_APPLICATION, double, HELMHOLTZ_RADIUS);
    KRATOS_DEFINE_APPLICATION_VARIABLE(OPTIMIZATION_APPLICATION, double, HELMHOLTZ_SCALAR);
    KRATOS_DEFINE_APPLICATION_VARIABLE(OPTIMIZATION_APPLICATION, double, HELMHOLTZ_SCALAR_SOURCE);
```

applications/OptimizationApplication/tests/filtering/implicit_filters_tests.py
```diff
@@ -0,0 +1,400 @@
...
+                                    "echo_level"    : 0,
+                                    "start_time"    : 0.0,
+                                    "end_time"      : 1.0,
+                                    "parallel_type" : "OpenMP"
+                                }
+                            }""")
+
+        my_filter = HelmholtzAnalysis(model, general_scalar_parameters)
+
+        model_part_io.ReadModelPart(model_part)
```
sunethwarna: I would use `Kratos.testing.utilities.ReadModelPart` instead of `ModelPartIO::ReadModelPart`. Then the same test can be used OpenMp and MPI without any change.

applications/OptimizationApplication/tests/filtering/implicit_filters_tests.py
```diff
@@ -0,0 +1,400 @@
...
+
+        unfiltered_uniform_field_nodal = KM.ContainerExpression.NodalNonHistoricalExpression(model_part)
+        unfiltered_uniform_field_nodal.SetData(1.0)
+
+        filtered_field = my_filter.FilterField(unfiltered_uniform_field_nodal)
+        self.assertAlmostEqual(LA.norm(filtered_field.Evaluate()), 3.741657, 4)
```
sunethwarna: I would use l2 norm from `ContainerExpressionUtilities` to minimize the use of numpy. This will be relevant if we are to use the same test for MPI in future because numpy if not mpi compatible, but `ContainerExpressionUtilities` is..

applications/OptimizationApplication/tests/filtering/implicit_filters_tests.py
```diff
@@ -0,0 +1,400 @@
...
+
+        for element in model_part.Elements:
+            for node in element.GetNodes():
+                node.SetValue(KM.NODAL_VOLUME, node.GetValue(KM.NODAL_VOLUME)+element.GetGeometry().Volume()/4.0)
+
+        nodal_volume = KM.ContainerExpression.NodalNonHistoricalExpression(model_part)
+        nodal_volume.Read(KM.NODAL_VOLUME)
+
+        filtered_field = my_filter.FilterIntegratedField(nodal_volume)
+        self.assertAlmostEqual(LA.norm(filtered_field.Evaluate()), 3.741657, 4)
```
sunethwarna: As in the above comment

applications/OptimizationApplication/tests/filtering/implicit_filters_tests.py
```diff
@@ -0,0 +1,400 @@
...
+                                    "echo_level"    : 0,
+                                    "start_time"    : 0.0,
+                                    "end_time"      : 1.0,
+                                    "parallel_type" : "OpenMP"
+                                }
+                            }""")
+
+        my_filter = HelmholtzAnalysis(model, filter_parameters)
+
+        model_part_io.ReadModelPart(model_part)
```
sunethwarna: Please change to `ReadModelPart` from testing.utilities

applications/OptimizationApplication/tests/elements/bulk_surface_implicit_filters_tests.py
```diff
@@ -0,0 +1,95 @@
+import KratosMultiphysics as KM
```
sunethwarna: Please reformulate this test to use Kratos testing framework.

applications/OptimizationApplication/python_scripts/filtering/python_solvers_wrapper_implicit_filters.py
```diff
@@ -0,0 +1,56 @@
...
+        raise Exception(err_msg)
+
+    module_full = 'KratosMultiphysics.OptimizationApplication.filtering.' + solver_module_name
+    solver = import_module(module_full).CreateSolver(model, solver_settings)
+
+    return solver
+
+def CreateSolver(model, custom_settings):
+
+    if (type(model) != KratosMultiphysics.Model):
```
sunethwarna:
```suggestion
    if isinstance(model, KratosMultiphysics.Model):
```

applications/OptimizationApplication/python_scripts/filtering/python_solvers_wrapper_implicit_filters.py
```diff
@@ -0,0 +1,56 @@
...
+        raise Exception("input is expected to be provided as a Kratos Model object")#
+
+    if (type(custom_settings) != KratosMultiphysics.Parameters):
```
sunethwarna:
```suggestion
    if isinstance(custom_settings, KratosMultiphysics.Parameters):
```

applications/OptimizationApplication/python_scripts/filtering/helmholtz_vector_solver.py
```diff
@@ -0,0 +1,86 @@
...
+        num_conds_nodes = None
+        for cond in self.original_model_part.Conditions:
+            geom = cond.GetGeometry()
+            num_conds_nodes = len(cond.GetNodes())
+            break
+
+        if self.bulk_surface_shape_filter:
+
+            if num_elems_nodes != 4:
+                raise Exception('::[HelmholtzVectorSolver]:: given model part must have only tetrahedral elemenst')
```
sunethwarna, line 54:
```suggestion
                raise Exception('::[HelmholtzVectorSolver]:: Given model part must have only tetrahedral elements')
```

applications/OptimizationApplication/python_scripts/filtering/helmholtz_vector_solver.py
```diff
@@ -0,0 +1,86 @@
...
+            if num_conds_nodes != 3:
+                raise Exception('::[HelmholtzVectorSolver]:: given model part must have only triangular conditions')
```
sunethwarna, line 56:
```suggestion
                raise Exception('::[HelmholtzVectorSolver]:: Given model part must have only triangular conditions')
```

applications/OptimizationApplication/python_scripts/filtering/helmholtz_vector_solver.py
```diff
@@ -0,0 +1,86 @@
...
+            if is_surface:
+                raise Exception('::[HelmholtzVectorSolver]:: given model part must have only solid elemenst')
```
sunethwarna, line 58:
```suggestion
                raise Exception('::[HelmholtzVectorSolver]:: Given model part must have only solid elements')
```

applications/OptimizationApplication/python_scripts/filtering/helmholtz_vector_solver.py
```diff
@@ -0,0 +1,86 @@
...
+    def PrepareModelPart(self):
+
+        num_elems_nodes = None
+        is_surface = False
+        for elem in self.original_model_part.Elements:
+            geom = elem.GetGeometry()
+            if geom.WorkingSpaceDimension() != geom.LocalSpaceDimension():
+                is_surface = True
+            num_elems_nodes = len(elem.GetNodes())
+            break
```
sunethwarna, line 43: Bit curious, what is the purpose of this check? Usually, elements are not present in surfaces.

1. In the case of shell elements, they represent not a surface, but still a domain with thickness. If a surface representation is required for this, it is usually created sepeatly with condtions.
2. Then we have beam elements where the working space dimension is 3 and local space dimension is 1, and it will identify this as a surface eventhough it represent a domain.

I am not sure the proper way of doing this. But I would consider a model part a domain if it has elements, and a surface if it only has conditions. May be I misunderstood the purpose of this check.

applications/OptimizationApplication/python_scripts/filtering/helmholtz_vector_solver.py
```diff
@@ -0,0 +1,86 @@
...
+                elif num_elems_nodes == 4:
+                    KM.ConnectivityPreserveModeler().GenerateModelPart(
+                            self.original_model_part, self.helmholtz_model_part, "HelmholtzVectorSurfaceElement3D4N")
+            else:
+                if num_elems_nodes == 4:
+                    KM.ConnectivityPreserveModeler().GenerateModelPart(
+                            self.original_model_part, self.helmholtz_model_part, "HelmholtzVectorSolidElement3D4N")
+                elif num_elems_nodes == 8:
+                    KM.ConnectivityPreserveModeler().GenerateModelPart(
+                            self.original_model_part, self.helmholtz_model_part, "HelmholtzVectorSolidElement3D8N")
```
sunethwarna: This will fail in the case of beams.

applications/OptimizationApplication/python_scripts/filtering/helmholtz_vector_solver.py
sunethwarna, line 54: I am little bit confused, how can a tetrahedral element represent a surface (you are using `self.bulk_surface_shape_filter`)?

applications/OptimizationApplication/python_scripts/filtering/helmholtz_scalar_solver.py
```diff
@@ -0,0 +1,54 @@
...
+            elif num_nodes == 4:
+                KM.ConnectivityPreserveModeler().GenerateModelPart(
+                        self.original_model_part, self.helmholtz_model_part, "HelmholtzSurfaceElement3D4N")
+        else:
+            if num_nodes == 4:
+                KM.ConnectivityPreserveModeler().GenerateModelPart(
+                        self.original_model_part, self.helmholtz_model_part, "HelmholtzSolidElement3D4N")
+            elif num_nodes == 8:
+                KM.ConnectivityPreserveModeler().GenerateModelPart(
+                        self.original_model_part, self.helmholtz_model_part, "HelmholtzSolidElement3D8N")
```
sunethwarna: Again the problem of surface detection as commented in the vector_solver.

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+
+    KRATOS_CATCH("");
+}
+
+//************************************************************************************
+//************************************************************************************
+void HelmholtzSurfaceShapeCondition::CalculateLocalSystem(MatrixType& rLeftHandSideMatrix,
+                                            VectorType& rRightHandSideVector,
+                                            const ProcessInfo& rCurrentProcessInfo)
+{
```
sunethwarna, line 90:
```suggestion
void HelmholtzSurfaceShapeCondition::CalculateLocalSystem(
    MatrixType& rLeftHandSideMatrix,
    VectorType& rRightHandSideVector,
    const ProcessInfo& rCurrentProcessInfo)
{
```

Kratos styleguide.

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+
+    KRATOS_ERROR_IF_NOT(rCurrentProcessInfo.Has(HELMHOLTZ_INTEGRATED_FIELD))
+    << "HELMHOLTZ_INTEGRATED_FIELD not defined in the ProcessInfo!" << std::endl;
+
+    KRATOS_ERROR_IF_NOT(rCurrentProcessInfo.Has(HELMHOLTZ_RADIUS))
+    << "HELMHOLTZ_RADIUS not defined in the ProcessInfo!" << std::endl;
+
+    // Check that parents have been computed
+    // These are required to retrieve the material properties and the viscous stress
+    auto& parentElement = this->GetValue(NEIGHBOUR_ELEMENTS);
```
sunethwarna, line 104: There should be a Has check for `NEIGHBOUR_ELEMENTS`

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+
+    auto& r_geometry = this->GetGeometry();
+    const SizeType number_of_nodes = r_geometry.size();
+    const SizeType dimension = r_geometry.WorkingSpaceDimension();
+
+    // Resizing as needed the LHS
+    const SizeType mat_size = number_of_nodes * dimension;
+
+    if ( rLeftHandSideMatrix.size1() != mat_size )
+        rLeftHandSideMatrix.resize( mat_size, mat_size, false );
```
sunethwarna, line 116:
```suggestion
    if ( rLeftHandSideMatrix.size1() != mat_size ) {
    rLeftHandSideMatrix.resize( mat_size, mat_size, false );
    }
```

Kratos Syleguide, Please don't use the suggestion, it kills the tabbing.

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+
+    noalias( rLeftHandSideMatrix ) = ZeroMatrix( mat_size, mat_size ); //resetting LHS
+
+    // Resizing as needed the RHS
+    if ( rRightHandSideVector.size() != mat_size )
```
sunethwarna, line 121: Kratos styleguide

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+        rRightHandSideVector.resize( mat_size, false );
+
+    rRightHandSideVector = ZeroVector( mat_size ); //resetting RHS
+
+    MatrixType K;
+    CalculateStiffnessMatrix(K,rCurrentProcessInfo);
+
+    if(!rCurrentProcessInfo[COMPUTE_HELMHOLTZ_INVERSE])
```
sunethwarna: Kratos style guide

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+        noalias(rLeftHandSideMatrix) += K;
+
+    const unsigned int number_of_points = r_geometry.size();
+    Vector nodal_vals(number_of_points*3);
```
sunethwarna, line 135: Would it make sense to template this class with `TDim` and `TNumNodes` so you dont have to heap allocate the vectors and use `BoundedVectors`?. This heap allocation will be an expensive operation.

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+    for(unsigned int node_element = 0; node_element<number_of_points; node_element++)
+    {
```
sunethwarna, line 137:
```suggestion
    for(unsigned int node_element = 0; node_element<number_of_points; node_element++)  {
```

Kratos style guide

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+        const auto &source = r_geometry[node_element].GetValue(HELMHOLTZ_VECTOR_SOURCE);
+        nodal_vals[3 * node_element + 0] = source[0];
+        nodal_vals[3 * node_element + 1] = source[1];
+        nodal_vals[3 * node_element + 2] = source[2];
```
sunethwarna, line 141: I would do it with a `local_index` incrementor to avoid multiplication

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+    }
+
+    if (rCurrentProcessInfo[COMPUTE_HELMHOLTZ_INVERSE])
+        noalias(rRightHandSideVector) += prod(K,nodal_vals);
```
sunethwarna: Kratos style guide

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+
+    //apply drichlet BC
+    Vector temp;
+    GetValuesVector(temp,0);
```
sunethwarna, line 149: This vector heap allocation also can be avoided if templated `TDim` and `TNumNodes` are used.

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+    noalias(rRightHandSideVector) -= prod(rLeftHandSideMatrix,temp);
+
+    KRATOS_CATCH("")
+}
+
+//******************************************************************************
+//******************************************************************************
+void HelmholtzSurfaceShapeCondition::GetValuesVector(VectorType &rValues,
+                                            int Step) const {
```
sunethwarna, line 158: Kratos style guide

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+  const GeometryType &rgeom = this->GetGeometry();
+  const SizeType num_nodes = rgeom.PointsNumber();
+  const unsigned int dimension = GetGeometry().WorkingSpaceDimension();
+  const unsigned int local_size = num_nodes * dimension;
+
+  if (rValues.size() != local_size)
+    rValues.resize(local_size, false);
```
sunethwarna, line 165: Kratos style guide

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+    rValues[index++] =
+        rgeom[i_node].FastGetSolutionStepValue(HELMHOLTZ_VECTOR_Y, Step);
+    rValues[index++] =
+        rgeom[i_node].FastGetSolutionStepValue(HELMHOLTZ_VECTOR_Z, Step);
+  }
+}
+//******************************************************************************
+//******************************************************************************
+
+void HelmholtzSurfaceShapeCondition::Calculate(const Variable<double>& rVariable, double& rOutput, const ProcessInfo& rCurrentProcessInfo)
```
sunethwarna, line 180:
```suggestion
void HelmholtzSurfaceShapeCondition::Calculate(
const Variable<double>& rVariable, 
double& rOutput, 
const ProcessInfo& rCurrentProcessInfo)
```

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+    if (rResult.size() != dimension * number_of_nodes)
+        rResult.resize(dimension * number_of_nodes,false);
+
+    const SizeType pos = this->GetGeometry()[0].GetDofPosition(HELMHOLTZ_VECTOR_X);
+
+    for (IndexType i = 0; i < number_of_nodes; ++i) {
+        const SizeType index = i * 3;
+        rResult[index] = GetGeometry()[i].GetDof(HELMHOLTZ_VECTOR_X,pos).EquationId();
+        rResult[index + 1] = GetGeometry()[i].GetDof(HELMHOLTZ_VECTOR_Y,pos+1).EquationId();
+        rResult[index + 2] = GetGeometry()[i].GetDof(HELMHOLTZ_VECTOR_Z,pos+2).EquationId();
```
sunethwarna, line 240: or you can use a `local_index` which is incremented.

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+    }
+
+    KRATOS_CATCH("")
+
+}
+
+//************************************************************************************
+//************************************************************************************
+void HelmholtzSurfaceShapeCondition::GetDofList(DofsVectorType& rElementalDofList,const ProcessInfo& rCurrentProcessInfo) const
```
sunethwarna, line 249: style guide

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+    if ( rLeftHandSideMatrix.size1() != mat_size )
+        rLeftHandSideMatrix.resize( mat_size, mat_size, false );
+
+    noalias( rLeftHandSideMatrix ) = ZeroMatrix( mat_size, mat_size ); //resetting LHS
+
+    // Resizing as needed the RHS
+    if ( rRightHandSideVector.size() != mat_size )
+        rRightHandSideVector.resize( mat_size, false );
+
+    rRightHandSideVector = ZeroVector( mat_size ); //resetting RHS
```
sunethwarna, line 124:
```suggestion
    noalias(rRightHandSideVector) = ZeroVector( mat_size ); //resetting RHS
```

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+    KRATOS_CATCH("");
+
+}
+/***********************************************************************************/
+/***********************************************************************************/
+void HelmholtzSurfaceShapeCondition::GetParentElementShapeFunctionsValues(
+    MatrixType& rNMatrix,
+    const IntegrationMethod& rIntegrationMethod,
+    const ProcessInfo& rCurrentProcessInfo
+    ) const
```
sunethwarna, line 349: Nice with style :) :+1:

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+        r_elem_geom.PointLocalCoordinates(elem_cond__gp_local_pt,cond_gp_global_pt);
+
+        for(IndexType cond_point_number = 0; cond_point_number < r_cond_geom.size(); ++cond_point_number ){
+            for ( IndexType elem_point_number = 0; elem_point_number < r_elem_geom.size(); ++elem_point_number ){
+                if(r_cond_geom[cond_point_number].Id()==r_elem_geom[elem_point_number].Id()){
+                    double elem_shape_func_value = r_elem_geom.ShapeFunctionValue(elem_point_number,elem_cond__gp_local_pt);
+                    rNMatrix(point_number,cond_point_number) = elem_shape_func_value;
+                }
+            }
+        }
```
sunethwarna, line 379: Curious, this is calculating shape function values for gauss points on conditions using elements. Why is it not possible to calculate shape function gauss point values on conditions it self. You can avoid this extra for loop and it will be way faster. If the integration method is the same, both should give the same value.

https://github.com/KratosMultiphysics/Kratos/blob/1f89ad0f5ab9b554f80013bd0bd39077cd083a6e/applications/RANSApplication/custom_utilities/rans_calculation_utilities.cpp#L62-L89

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+        rStiffnessMatrix.resize( mat_size, mat_size, false );
+    rStiffnessMatrix = ZeroMatrix( mat_size, mat_size );
+
+    const IntegrationMethod& integration_method = r_cond_geom.GetDefaultIntegrationMethod();
+    const GeometryType::IntegrationPointsArrayType& integration_points = r_cond_geom.IntegrationPoints(integration_method);
+
+
+    // get the normal
+    VectorType n_surf;
+    CalculateNormal(n_surf);
```
sunethwarna, line 410: normal calculation can be done outside the element using normal calculation utils, so you can read it here.

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+    MatrixType A_dirc = ZeroMatrix(number_of_nodes,number_of_nodes);
+    for ( IndexType point_number = 0; point_number < integration_points.size(); ++point_number ) {
+
+        Matrix J0;
+        GeometryUtils::JacobianOnInitialConfiguration(r_cond_geom, integration_points[point_number], J0);
+        double detJ0 = MathUtils<double>::GeneralizedDet(J0);
+
+        const double integration_weight = integration_points[point_number].Weight() * detJ0;
+        MatrixType DN_DX;
+        GetParentElementShapeFunctionsGlobalGradients(DN_DX,point_number,integration_method,rCurrentProcessInfo);
```
sunethwarna, line 422: I guess you need `DN_DZ` as well here ryt?

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+        MatrixType DN_DX_t = prod(DN_DX,tangent_projection_matrix);
+        const double r_helmholtz = rCurrentProcessInfo[HELMHOLTZ_RADIUS];
+        noalias(A_dirc) += integration_weight * r_helmholtz * r_helmholtz * prod(DN_DX_t, trans(DN_DX_t));
+    }
+
+    //contruct the stifness matrix in all dims
+    for(IndexType i=0;i<number_of_nodes;i++)
+        for(IndexType j=0;j<dimension;j++)
+            for(IndexType k=0;k<number_of_nodes;k++)
+                rStiffnessMatrix(dimension*i+j,dimension*k+j) = A_dirc(i,k);
```
sunethwarna, line 432: Kratos style guide

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+    MatrixType DN_De;
+    r_elem_geom.ShapeFunctionsLocalGradients(DN_De,elem_cond__gp_local_pt);
+
+    Matrix J0,InvJ0;
+    GeometryUtils::JacobianOnInitialConfiguration(r_elem_geom, elem_cond__gp_local_pt, J0);
+    double detJ0;
+    MathUtils<double>::InvertMatrix(J0, InvJ0, detJ0);
+
+
+    MatrixType elem_DN_DX = prod(DN_De,InvJ0);
```
sunethwarna, line 477:
```suggestion
    const MatrixType elem_DN_DX = prod(DN_De,InvJ0);
```

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+}
+
+/***********************************************************************************/
+/***********************************************************************************/
+void HelmholtzSurfaceShapeCondition::GetParentElementShapeFunctionsGlobalGradients(
+    MatrixType& rDN_DX,
+    const IndexType PointNumber,
+    const IntegrationMethod& rIntegrationMethod,
+    const ProcessInfo& rCurrentProcessInfo
+    ) const
```
sunethwarna, line 445: This is an expensive method, you have cond_local->cond_global->parent_element_local conversion, then cond_node == parent_element_node matching as well per gauss point. I don't have a solution now, but it would be better if we can come up with a better solution.

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.cpp
```diff
@@ -0,0 +1,511 @@
...
+
+    v2[0] = r_cond_geom[2].X0() - r_cond_geom[0].X0();
+    v2[1] = r_cond_geom[2].Y0() - r_cond_geom[0].Y0();
+    v2[2] = r_cond_geom[2].Z0() - r_cond_geom[0].Z0();
+
+    r_n.resize(3);
+    MathUtils<double>::CrossProduct(r_n,v1,v2);
+    double norm = MathUtils<double>::Norm3(r_n);
+    r_n /= norm;
+}
```
sunethwarna, line 511: This depends on whther you need outward pointing consistent normals. I would suggest to use the `NormalCalculationUtilities` and use the variable `NORMAL` to read in the normals.

Note: If the outward pointing consistency is required, then tetrahedral mesh orientation check should be done.

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.h
```diff
@@ -0,0 +1,238 @@
...
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:		 BSD License
+//					 license: OptimizationApplication/license.txt
+//
+//  Main authors:    Reza Najian Asl
+//
```
sunethwarna:
```suggestion
//    |  /           |
//    ' /   __| _` | __|  _ \   __|
//    . \  |   (   | |   (   |\__ `
//   _|\_\_|  \__,_|\__|\___/ ____/
//                   Multi-Physics
//
//  License:		 BSD License
//                              license: OptimizationApplication/license.txt
//
//  Main authors:    Reza Najian Asl
//
```

sunethwarna: Please remove the use of tabbing from (It is present in the old headers, but remove them and replace them with spaces when copying)

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.h
```diff
@@ -0,0 +1,238 @@
...
+
+/// Short class definition.
+/**
+ * @class HelmholtzSurfaceShapeCondition
+ * @ingroup OptimizationApplication
+ * @brief Helmholtz filtering condition for 3D geometries.
+ * @details Implements Sobolev/Helmholtz filter for filtering/regularization. This works for arbitrary geometries in 3D
+ * @author Reza Najian Asl
+ */
+class KRATOS_API(OPTIMIZATION_APPLICATION) HelmholtzSurfaceShapeCondition
```
sunethwarna, line 42: I would suggest to template this with `TDim` and `TNumNodes` and leave only the template specialization which are valid.

Could you please add some more documentation if possible like the equation being used in here?

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.h
```diff
@@ -0,0 +1,238 @@
...
+    /// Counted pointer of HelmholtzSurfaceShapeCondition
+    KRATOS_CLASS_INTRUSIVE_POINTER_DEFINITION(HelmholtzSurfaceShapeCondition);
+
+    ///@}
+    ///@name Life Cycle
+    ///@{
+
+    /// Default constructor.
+    HelmholtzSurfaceShapeCondition(IndexType NewId, GeometryType::Pointer pGeometry);
+    HelmholtzSurfaceShapeCondition(IndexType NewId, GeometryType::Pointer pGeometry,  PropertiesType::Pointer pProperties);
```
sunethwarna:
```suggestion
    HelmholtzSurfaceShapeCondition(
    IndexType NewId, 
    GeometryType::Pointer pGeometry);
    
    HelmholtzSurfaceShapeCondition(
    IndexType NewId, 
    GeometryType::Pointer pGeometry,  
    PropertiesType::Pointer pProperties);
```

Kratos style, don't commit the suggestion, it kills the tabbing

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.h
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    ///@}
+
+    ///@name Private Operations
+    ///@{
+    void CalculateStiffnessMatrix(MatrixType& rStiffnessMatrix,const ProcessInfo& rCurrentProcessInfo) const;
+    void CalculateNormal(VectorType & r_n) const;
+    void GetParentElementShapeFunctionsValues(MatrixType& rNMatrix,const IntegrationMethod& rIntegrationMethod, const ProcessInfo& rCurrentProcessInfo) const;
+    void GetParentElementShapeFunctionsGlobalGradients(MatrixType& rDN_DX,const IndexType PointNumber,const IntegrationMethod& rIntegrationMethod, const ProcessInfo& rCurrentProcessInfo) const;
```
sunethwarna: Kratos style guide

applications/OptimizationApplication/custom_elements/helmholtz_solid_element.cpp
```diff
@@ -0,0 +1,327 @@
...
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:		 BSD License
+//					 license: OptimizationApplication/license.txt
+//
+//  Main authors:    Reza Najian Asl
+//
+
+// System includes
```
sunethwarna: remove tabs

applications/OptimizationApplication/custom_elements/helmholtz_solid_element.h
```diff
@@ -0,0 +1,227 @@
...
+
+/// Short class definition.
+/**
+ * @class HelmholtzSolidElement
+ * @ingroup OptimizationApplication
+ * @brief Helmholtz filtering element for 3D geometries.
+ * @details Implements Sobolev/Helmholtz filter for filtering/regularization. This works for arbitrary geometries in 3D
+ * @author Reza Najian Asl
+ */
+class KRATOS_API(OPTIMIZATION_APPLICATION) HelmholtzSolidElement
```
sunethwarna: templating and adding latex stlye equation would be nice for the documentation

applications/OptimizationApplication/custom_elements/helmholtz_solid_shape_element.cpp
```diff
@@ -0,0 +1,494 @@
...
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:		 BSD License
+//					 license: OptimizationApplication/license.txt
+//
+//  Main authors:    Reza Najian Asl
+//
+
```
sunethwarna: replace tabbing

applications/OptimizationApplication/custom_elements/helmholtz_solid_shape_element.cpp
```diff
@@ -0,0 +1,494 @@
...
+    const unsigned int number_of_points = r_geometry.size();
+    Vector nodal_vals(number_of_points*3);
+    for(unsigned int node_element = 0; node_element<number_of_points; node_element++)
+    {
+        const auto &source = r_geometry[node_element].GetValue(HELMHOLTZ_VECTOR_SOURCE);
+        auto node_weight = r_geometry[node_element].GetValue(NUMBER_OF_NEIGHBOUR_ELEMENTS);
+        nodal_vals[3 * node_element + 0] = source[0];
+        nodal_vals[3 * node_element + 1] = source[1];
+        nodal_vals[3 * node_element + 2] = source[2];
+        if(rCurrentProcessInfo[HELMHOLTZ_INTEGRATED_FIELD]){
```
sunethwarna: this lookup for `HELMHOLTZ_INTEGRATED_FIELD` should not be in the for loop.

applications/OptimizationApplication/custom_elements/helmholtz_solid_shape_element.cpp
```diff
@@ -0,0 +1,494 @@
...
+    noalias(rLeftHandSideMatrix) += M;
+    if(!rCurrentProcessInfo[COMPUTE_HELMHOLTZ_INVERSE])
+        noalias(rLeftHandSideMatrix) += K;
+
+    const unsigned int number_of_points = r_geometry.size();
+    Vector nodal_vals(number_of_points*3);
+    for(unsigned int node_element = 0; node_element<number_of_points; node_element++)
+    {
+        const auto &source = r_geometry[node_element].GetValue(HELMHOLTZ_VECTOR_SOURCE);
+        auto node_weight = r_geometry[node_element].GetValue(NUMBER_OF_NEIGHBOUR_ELEMENTS);
```
sunethwarna: this should be only read if the below if block is true.

sunethwarna:
```suggestion
        const auto node_weight = r_geometry[node_element].GetValue(NUMBER_OF_NEIGHBOUR_ELEMENTS);
```

applications/OptimizationApplication/custom_elements/helmholtz_solid_shape_element.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+    for(std::size_t i_point = 0; i_point<integration_points.size(); ++i_point)
+    {
+
+        Matrix J0,InvJ0;
+        GeometryUtils::JacobianOnInitialConfiguration(r_geom, integration_points[i_point], J0);
+        double detJ0;
+        MathUtils<double>::InvertMatrix(J0, InvJ0, detJ0);
+
+        MatrixType B = CalculateBMatrix(i_point,rCurrentProcessInfo);
```
sunethwarna:
```suggestion
        const MatrixType& B = CalculateBMatrix(i_point,rCurrentProcessInfo);
```

applications/OptimizationApplication/custom_elements/helmholtz_solid_shape_element.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+        MatrixType constitutive_matrix = SetAndModifyConstitutiveLaw(i_point,rCurrentProcessInfo);
```
sunethwarna:
```suggestion
        const MatrixType& constitutive_matrix = SetAndModifyConstitutiveLaw(i_point,rCurrentProcessInfo);
```

applications/OptimizationApplication/custom_elements/helmholtz_solid_shape_element.cpp
```diff
@@ -0,0 +1,494 @@
...
+  const auto& r_integration_points = rgeom.IntegrationPoints(this_integration_method);
+  GeometryType::ShapeFunctionsGradientsType DN_De = rgeom.ShapeFunctionsLocalGradients(this_integration_method);
+
+  Matrix J0,InvJ0;
+  GeometryUtils::JacobianOnInitialConfiguration(rgeom, r_integration_points[PointNumber], J0);
+  double detJ0;
+  MathUtils<double>::InvertMatrix(J0, InvJ0, detJ0);
+
+
+  Matrix DN_DX = prod(DN_De[PointNumber], InvJ0);
```
sunethwarna:
```suggestion
  const Matrix& DN_DX = prod(DN_De[PointNumber], InvJ0);
```

applications/OptimizationApplication/custom_elements/helmholtz_solid_shape_element.cpp
```diff
@@ -0,0 +1,494 @@
...
+  return B;
+
+  KRATOS_CATCH("");
+}
+//******************************************************************************
+//******************************************************************************
+HelmholtzSolidShapeElement::MatrixType
+HelmholtzSolidShapeElement::SetAndModifyConstitutiveLaw(
+    const int PointNumber, const ProcessInfo& rCurrentProcessInfo) const {
+  KRATOS_TRY;
```
sunethwarna: Shoudn't this be done in the constitutive law?

applications/OptimizationApplication/custom_elements/helmholtz_solid_shape_element.h
```diff
@@ -0,0 +1,239 @@
...
+
+/// Short class definition.
+/**
+ * @class HelmholtzSolidShapeElement
+ * @ingroup OptimizationApplication
+ * @brief Helmholtz filtering element for 3D geometries.
+ * @details Implements Sobolev/Helmholtz filter for filtering/regularization. This works for arbitrary geometries in 3D
+ * @author Reza Najian Asl
+ */
+class KRATOS_API(OPTIMIZATION_APPLICATION) HelmholtzSolidShapeElement
```
sunethwarna: templating and latex equation

applications/OptimizationApplication/custom_elements/helmholtz_surface_element.cpp
```diff
@@ -0,0 +1,484 @@
...
+
+    const unsigned int number_of_points = r_geometry.size();
+    Vector nodal_vals(number_of_points);
+    for(unsigned int node_element = 0; node_element<number_of_points; node_element++)
+    {
+        const auto &source = r_geometry[node_element].GetValue(HELMHOLTZ_SCALAR_SOURCE);
+        auto node_weight = r_geometry[node_element].GetValue(NUMBER_OF_NEIGHBOUR_ELEMENTS);
+        nodal_vals[node_element] = source;
+        if(rCurrentProcessInfo[HELMHOLTZ_INTEGRATED_FIELD])
+            nodal_vals[node_element] /= node_weight;
```
sunethwarna: should not lookup `HELMHOLTZ_INTEGRATED_FIELD` within the for loop, and nodal weigh should only be read when required.

applications/OptimizationApplication/custom_elements/helmholtz_surface_element.cpp
```diff
@@ -0,0 +1,484 @@
...
+
+    // create a pseudo bulk element here
+    double height = r_geom.Length();
+    Point geom_center = r_geom.Center();
+    const auto surf_points = r_geom.Points();
+    VectorType n_surf;
+    CalculateAvgSurfUnitNormal(n_surf);
+    PointPtrType p0 = new PointType(surf_points[0].Id(),surf_points[0]);
+    PointPtrType p1 = new PointType(surf_points[1].Id(),surf_points[1]);
+    PointPtrType p2 = new PointType(surf_points[2].Id(),surf_points[2]);
```
sunethwarna: Wow. this is a memory leak... at least use `uniqe_ptr` which is really expensive in the element level.

applications/OptimizationApplication/custom_elements/helmholtz_surface_element.cpp
```diff
@@ -0,0 +1,484 @@
...
+            r_geom.GlobalCoordinates(gp_global_pt,gp_local_pt);
+            Point elem_gp_local_pt;
+            pseudo_tetrahedra.PointLocalCoordinates(elem_gp_local_pt,gp_global_pt);
+
+            for(IndexType i = 0; i < number_of_nodes; ++i )
+                rNMatrix(point_number,i) = pseudo_tetrahedra.ShapeFunctionValue(i,elem_gp_local_pt);
+
+        }
+
+    }else if(number_of_nodes==4) {
```
sunethwarna: I don't think this is required. What is done in here is, cond_local_point->cond_global_point->element_local_point->element_gp_on_local_point. This should be basically the same as doing shape function values on the condition geometry. Then you won't require any of these, and will be much cheaper operation.

applications/OptimizationApplication/custom_elements/helmholtz_surface_element.cpp
```diff
@@ -0,0 +1,484 @@
...
+        PointPtrType p3 = new PointType(surf_points[3].Id(),surf_points[3]);
+        PointPtrType p4 = new PointType(surf_points[0].Id()+4, geom_center.Coordinates() + height * n_surf);
+        PyramidGeometryType pseudo_pyramid(p0,p1,p2,p3,p4);
```
sunethwarna: This is not allowed, and leaks lots of memory.

applications/OptimizationApplication/custom_elements/helmholtz_surface_element.cpp
```diff
@@ -0,0 +1,484 @@
...
+        pseudo_pyramid.PointLocalCoordinates(elem_surf_gp_local_pt,surf_gp_global_pt);
+        MatrixType DN_De;
+        pseudo_pyramid.ShapeFunctionsLocalGradients(DN_De,elem_surf_gp_local_pt);
+        MatrixType InvJ0;
+        pseudo_pyramid.InverseOfJacobian(InvJ0,elem_surf_gp_local_pt);
+        pseudo_elem_DN_DX = prod(DN_De,InvJ0);
+    }
```
sunethwarna: This leaks memory. I would suggest to calculate `DN_DX` in the element by passing the cond_local_potin->condition_global_point->element_local_point and then take that matrix accordingly.

Then you don't have to create new points. Creation of points per gauss loop is very expensive.

applications/OptimizationApplication/custom_elements/helmholtz_surface_element.cpp
```diff
@@ -0,0 +1,484 @@
...
+    const GeometryType::IntegrationPointsArrayType& integration_points = r_geom.IntegrationPoints(integration_method);
+
+    rNormal.resize(3);
+    rNormal = ZeroVector(3);
+    for ( IndexType point_number = 0; point_number < integration_points.size(); ++point_number )
+        rNormal += r_geom.UnitNormal(point_number,integration_method);
+
+    rNormal /= integration_points.size();
+    rNormal /= MathUtils<double>::Norm3(rNormal);
+}
```
sunethwarna: Use of `NormalCalculationUtilities`

applications/OptimizationApplication/custom_elements/helmholtz_surface_element.h
```diff
@@ -0,0 +1,258 @@
...
+
+/// Short class definition.
+/**
+ * @class HelmholtzSurfaceElement
+ * @ingroup OptimizationApplication
+ * @brief Helmholtz filtering element for 3D geometries.
+ * @details Implements Sobolev/Helmholtz filter for filtering/regularization. This works for arbitrary geometries in 3D
+ * @author Reza Najian Asl
+ */
+class KRATOS_API(OPTIMIZATION_APPLICATION) HelmholtzSurfaceElement
```
sunethwarna: templating and latex equation please :D

applications/OptimizationApplication/custom_elements/helmholtz_vector_solid_element.cpp
```diff
@@ -0,0 +1,372 @@
...
+        const auto &source = r_geometry[node_element].GetValue(HELMHOLTZ_VECTOR_SOURCE);
+        auto node_weight = r_geometry[node_element].GetValue(NUMBER_OF_NEIGHBOUR_ELEMENTS);
+        nodal_vals[3 * node_element + 0] = source[0];
+        nodal_vals[3 * node_element + 1] = source[1];
+        nodal_vals[3 * node_element + 2] = source[2];
+        if(rCurrentProcessInfo[HELMHOLTZ_INTEGRATED_FIELD]){
+            nodal_vals[3 * node_element + 0] /= node_weight;
+            nodal_vals[3 * node_element + 1] /= node_weight;
+            nodal_vals[3 * node_element + 2] /= node_weight;
+        }
```
sunethwarna: as before, constant value lookups should not be done within the for loop

applications/OptimizationApplication/custom_elements/helmholtz_vector_surface_element.cpp
```diff
@@ -0,0 +1,555 @@
...
+        MatrixType InvJ0;
+        pseudo_pyramid.InverseOfJacobian(InvJ0,elem_surf_gp_local_pt);
+        pseudo_elem_DN_DX = prod(DN_De,InvJ0);
+    }
+    else
+        KRATOS_ERROR<<"HelmholtzVectorSurfaceElement: this element only supports 3 and 4 noded surface elements"<<std::endl;
+
+    for(IndexType i = 0; i<mat_size1; i++)
+        for(IndexType j = 0; j<mat_size2; j++)
+            rDN_DX(i,j) = pseudo_elem_DN_DX(i,j);
```
sunethwarna: no use of `new` please as mentioned in above comments.

applications/OptimizationApplication/custom_elements/helmholtz_vector_surface_element.cpp
```diff
@@ -0,0 +1,555 @@
...
+    const GeometryType::IntegrationPointsArrayType& integration_points = r_geom.IntegrationPoints(integration_method);
+
+    rNormal.resize(3);
+    rNormal = ZeroVector(3);
+    for ( IndexType point_number = 0; point_number < integration_points.size(); ++point_number )
+        rNormal += r_geom.UnitNormal(point_number,integration_method);
+
+    rNormal /= integration_points.size();
+    rNormal /= MathUtils<double>::Norm3(rNormal);
+}
```
sunethwarna: as previously mentioned

applications/OptimizationApplication/custom_elements/helmholtz_vector_surface_element.cpp
```diff
@@ -0,0 +1,555 @@
...
+            r_geom.GlobalCoordinates(gp_global_pt,gp_local_pt);
+            Point elem_gp_local_pt;
+            pseudo_pyramid.PointLocalCoordinates(elem_gp_local_pt,gp_global_pt);
+
+            for(IndexType i = 0; i < number_of_nodes; ++i )
+                rNMatrix(point_number,i) = pseudo_pyramid.ShapeFunctionValue(i,elem_gp_local_pt);
+        }
+    }
+    else
+        KRATOS_ERROR<<"HelmholtzVectorSurfaceElement: this element only supports 3 and 4 noded surface elements"<<std::endl;
```
sunethwarna: This is a memory leak

applications/OptimizationApplication/custom_python/add_custom_utilities_to_python.cpp
```diff
@@ -248,6 +249,12 @@ void  AddCustomUtilitiesToPython(pybind11::module& m)
         .def("ComputeNodalVariableProductWithEntityMatrix", &ContainerExpressionUtils::ComputeNodalVariableProductWithEntityMatrix<ModelPart::ConditionsContainerType>, py::arg("output_nodal_container_expression"), py::arg("input_nodal_values_container_expression"), py::arg("matrix_variable"), py::arg("entities"))
         .def("ComputeNodalVariableProductWithEntityMatrix", &ContainerExpressionUtils::ComputeNodalVariableProductWithEntityMatrix<ModelPart::ElementsContainerType>, py::arg("output_nodal_container_expression"), py::arg("input_nodal_values_container_expression"), py::arg("matrix_variable"), py::arg("entities"))
         ;
+
+    m.def_submodule("ImplicitFilterUtils")
+        .def("CalculateNodeNeighbourCount", &ImplicitFilterUtils::CalculateNodeNeighbourCount)
+        .def("SetBulkRadiusForShapeFiltering", &ImplicitFilterUtils::SetBulkRadiusForShapeFiltering)
```
sunethwarna: I would add `py::args` in here

applications/OptimizationApplication/python_scripts/filtering/helmholtz_analysis.py
```diff
@@ -0,0 +1,129 @@
...
+    def _SetHelmHoltzSourceMode(self, integrated_field=False):
+        self._GetComputingModelPart().ProcessInfo.SetValue(KOA.HELMHOLTZ_INTEGRATED_FIELD, integrated_field)
+
+    def _AssignDataExpressionToNodalSource(self, data_exp: ContainerExpressionTypes):
+        mapped_values = KM.ContainerExpression.NodalNonHistoricalExpression(data_exp.GetModelPart())
+        if type(data_exp) == KM.ContainerExpression.NodalNonHistoricalExpression:
+            mapped_values = data_exp
+        elif type(data_exp) == KM.ContainerExpression.ElementNonHistoricalExpression:
+            neighbour_elems = KM.ContainerExpression.NodalNonHistoricalExpression(data_exp.GetModelPart())
+            KOA.ContainerExpressionUtils.ComputeNumberOfNeighbourElements(neighbour_elems)
```
sunethwarna, line 43: If you do this already, you don't need `ImplicitFilterUtils::CalculateNodeNeighbourCount`, you can just say
neighbour_elems.Evaluate(Kratos.NUMBER_OF_NEIGHBOUR_ELEMENTS). So you only calculate neighbous once for all. You may have to move this neighbour calculation outside the if block.

sunethwarna (commented):
Hi @RezaNajian , Thanks for all the work. Following summarize my overall review:

I have following major concerns regarding this PR:
1. Creation of on the fly `Point` with `new` keyword, and not managing the memory is a memory leak. This is really dangerous. For the shape function value computation, as I mentioned in the comments, I would use the condition geometry only. And for `DN_DX` computation, I would cond_local_pt->cond_global_pt->element_local_pt then calculate the `DN_DX` in the element and find the corresponding condition DN_DX from that.
2. All the test should use the kratos testing framework
3. There is a lot of code duplication in elements/conditions, I would suggest to refactor the code.

Followings are the minor comments:
1. No need to have another element neighbour count mechanism, the method in `ContainerExpressionUtils` can be used there.
2. Use of kratos style guide.
3. Replacing tabs with spaces in the headers
4.  I would templatize the elements and conditions with `TDim` and `TNumNodes`.

Question:
Would it be possible to templatize based on the `scalar` or `vector` also in the elements? As far as I have seen, it may be possible. So we can remove lots of code. :D

May be we can have a meeting and you could explain the code a little bit as well to clarify my misunderstandings :D

applications/OptimizationApplication/custom_conditions/helmholtz_surface_shape_condition.h
RezaNajian: I added. Please check and let me know.

applications/OptimizationApplication/custom_elements/helmholtz_solid_element.cpp
RezaNajian: done

applications/OptimizationApplication/custom_elements/helmholtz_solid_element.h
RezaNajian: can you give me an example of how to put ?

applications/OptimizationApplication/python_scripts/filtering/helmholtz_analysis.py
RezaNajian: The point is that in general, I need to know the NUMBER_OF_NEIGHBOUR_ELEMENTS for PDEs. What you say is valid for filtering the element container field. However, the solver can be called standalone.

sunethwarna:
Hi Reza,

Could you make the element/condition test something similar to the followng given example? We usually create element/condition tests in cpp because it is way faster,  there is no need of testing the python exposure (python exposure is already tested in core) and there is little benefit on using the same model because a new model part is needed to be created for each new element/condition type.

Once you make the test, I can start porting to templates... I think I found a way, but cannot be sure until I try.

Be sure to add the test folder with cpps to the CMakeLists.txt for the compilation.

You can run the given example (assuming it is in OptimizationApplication) cpp tests manually by the following python code block:
```python
import KratosMultiphysics.OptimizationApplication as KratosOA

Kratos.Tester.SetVerbosity(Kratos.Tester.Verbosity.TESTS_OUTPUTS)
Kratos.Tester.RunTestCases("*ElementStokesTwoFluidPositiveSide3D4N*")
```

Following is an example of a CPP test:
```c++
KRATOS_TEST_CASE_IN_SUITE(ElementStokesTwoFluidPositiveSide3D4N, FluidDynamicsApplicationFastSuite)
{
    Model current_model;
    ModelPart& modelPart = current_model.CreateModelPart("Main");
    
    PrepareModelPart(modelPart);
    
    auto pElement = modelPart.pGetElement(1);

    pElement->GetGeometry()[0].FastGetSolutionStepValue(DISTANCE) = 1.0;
    pElement->GetGeometry()[1].FastGetSolutionStepValue(DISTANCE) = 1.0;
    pElement->GetGeometry()[2].FastGetSolutionStepValue(DISTANCE) = 1.0;
    pElement->GetGeometry()[3].FastGetSolutionStepValue(DISTANCE) = 1.0;

    // Compute RHS and LHS
    Vector RHS = ZeroVector(16);
    Matrix LHS = ZeroMatrix(16, 16);

    const auto& r_process_info = modelPart.GetProcessInfo();
    pElement->Initialize(r_process_info); // Initialize the element to initialize the constitutive law
    pElement->CalculateLocalSystem(LHS, RHS, r_process_info);

    // Check the RHS values (the RHS is computed as the LHS x previous_solution,
    // hence, it is assumed that if the RHS is correct, the LHS is correct as well)
    std::vector<double> theoretical_rhs = {0.039501,0.0600744,-0.328102,7026.02,0.0117953,0.0213953,-0.377754,411.433,0.0115203,0.0211214,-0.378029,685.739,0.0112459,0.0208464,-0.378303,-8123.29}; // namespace Testing
    KRATOS_CHECK_VECTOR_NEAR(theoretical_rhs, RHS, 1e-2);
}
```

You can find examples here:
https://github.com/KratosMultiphysics/Kratos/blob/master/applications/FluidDynamicsApplication/tests/cpp_tests/test_two_fluid_navier_stokes_element.cpp

Or any other application element tests for that matter.

RezaNajian:
@sunethwarna I cleaned the tests for testing Helmholtz solver. I can make the cpp tests as well.

RezaNajian:
@sunethwarna cpp tests added. You may proceed with it. Tnx.

RezaNajian:
@sunethwarna it is ready for you to continue. Tnx.

sunethwarna (dismissed):
@RezaNajian Thanks :)

sunethwarna (approved):

RezaNajian:
@sunethwarna Thanks for the collab. Highly appreciated. :-)

PR #11096, opened 2023-05-09
[Core] Bugfix - leave one empty mesh when calling clear Model Part

### LINKED ISSUE #10880: [Core] Calling Clear model part "breaks" model part in FullDebug in Windows

ddiezrod:
**📝 Description**
As described in https://github.com/KratosMultiphysics/Kratos/issues/10880, calling model part Clear leads to error as we should always have one default mesh. This PR add the fix proposed by @matekelemen there.

### DIFF
--- a/kratos/sources/model_part.cpp
+++ b/kratos/sources/model_part.cpp
@@ -82,6 +82,7 @@ void ModelPart::Clear()
 
     // Clear meshes list
     mMeshes.clear();
+    mMeshes.emplace_back(Kratos::make_shared<MeshType>());
 
     // Clear geometries
     mGeometries.Clear();


loumalouomega (approved):
Mesh

philbucher:
> Mesh

image

PR #11100, opened 2023-05-09
[Core] Adding `AllGatherv` methods to `DataCommunicator`/`MPIDataCommunicator`

loumalouomega:
**📝 Description**

This PR adds the `AllGatherv` method to the `DataCommunicator`/`MPIDataCommunicator`class, which allows efficient communication of data with varying sizes among multiple processes. It also adds tests for the method to ensure its correct implementation.

**🆕 Changelog**

- [Adding `AllGatherv`methods to `DataCommunicator`](https://github.com/KratosMultiphysics/Kratos/commit/1c51f28829a009facbf76969137708124fa7158b)
- [Expose `AllGatherv`to python](https://github.com/KratosMultiphysics/Kratos/commit/81879a4b6ffc08ad9e9dbb6f7524c1cffc7691c8)
- [Adding serial C++ test](https://github.com/KratosMultiphysics/Kratos/commit/e68e499d133e3aaac206c6d331603e86bcd9ebb5)
- [Adding MPI implementation in `MPIDataCommunicator`](https://github.com/KratosMultiphysics/Kratos/commit/f0aa31dd90ab5a841ad85aeb6a181157d2292a9d)
- [Adding MPI test](https://github.com/KratosMultiphysics/Kratos/commit/377309e5d070f3327be1b0e819265a1b59450d6c)
- [Add python test](https://github.com/KratosMultiphysics/Kratos/pull/11100/commits/4bc1cf824da9ab0e06968321cffbf84fb0d03391)


philbucher:
LGTM, but I leave approval to @jcotela 

Also please add tests for the python interface

kratos/mpi/tests/cpp_tests/sources/test_mpi_data_communicator.cpp
```diff
@@ -12,9 +12,10 @@
 //
 
 // System includes
-#include "mpi.h"
+#include <numeric>
```
jcotela, line 15: Why is this needed?

jcotela:
Looks good overall but it does not compile in debug mode. It seems that there is some problem in the debug-only part of the tests in test_mpi_data_communicator.cpp.

loumalouomega:
> Looks good overall but it does not compile in debug mode. It seems that there is some problem in the debug-only part of the tests in test_mpi_data_communicator.cpp.
Copy paste from Gatherv methods, I will clean up

kratos/mpi/tests/cpp_tests/sources/test_mpi_data_communicator.cpp
loumalouomega: To do std::accumulate

loumalouomega:
Okay, only python tests are missing. Tomorrow

kratos/mpi/tests/cpp_tests/sources/test_mpi_data_communicator.cpp
jcotela: but I can't find that in this PR : /

jcotela:
I would prefer to keep the tests for the debug functionality, the existing methods in the mpi data communicator have some basic sanity checks to help developers when building in debug mode and I think it would be good to provide similar functionality (appropriately tested) for new methods, where reasonable. We can discuss that on the phone later.

sunethwarna:
This is really cool :), may be we can extend the interfaces for `std::string` as well?

loumalouomega:
> This is really cool :), may be we can extend the interfaces for `std::string` as well?

I can.. but in a future PR, because I would like to extend it consistently for other methods. BTW, do you expect to use a std::string directly of a vector of strings?

RiccardoRossi:
btw, a bit off scope, but we really should add some tutorial to the wiki about how to use it in c++ and in python.

it is not obvious at all to use all of this without a guidance or an example

sunethwarna:
I am talking about  `std::vector<std::string>`.. :)

loumalouomega:
> I am talking about `std::vector<std::string>`.. :)

Future PR better i think

jcotela:
I have updated one of the debug validation checks you are doing, I think that the local send and receive sizes do not need to match (since the size of the piece of the full message sent from i to j does not need to be the same as the size of the piece sent from j to i) so I would communicate the expected sizes to be sure.

I have no issues with the code otherwise, so if you agree with the changes I will approve.

loumalouomega:
> I have updated one of the debug validation checks you are doing, I think that the local send and receive sizes do not need to match (since the size of the piece of the full message sent from i to j does not need to be the same as the size of the piece sent from j to i) so I would communicate the expected sizes to be sure.
> 
> I have no issues with the code otherwise, so if you agree with the changes I will approve.

I agree

jcotela:
I just updated two more tests that were failing since the check being tested is not done in the serial implementation. I think it is ok as is (it is also consistent with the AllGather interface) so I just removed the failing verification.

jcotela (approved):
No more issues from my side, approving.

PR #11103, opened 2023-05-10
ASCII logo for Optimization Application

Igarizza:
**📝 Description**
I have tried the full-name Optimization Application, but the name is too big. I have used the abbreviation we have used a lot. 

Here, is how it looks:

Screenshot from 2023-05-10 10-29-54

- Api Breaker, Mpi, etc...

**🆕 Changelog**
E.g.
- Added feature X to Y
- Added Foo Application
- Fixed X (#XXXX Reference to issue if apply) 
- etc...


RezaNajian:
@Igarizza @sunethwarna we have App/Application three times repeated. I would suggest to drop App/Application and only keep Optimization. What do you think ?!

RezaNajian (commented):
@Igarizza   we have App/Application three times repeated. I would suggest to drop App/Application and only keep Optimization. What do you think ?!

Igarizza:
> @Igarizza we have App/Application three times repeated. I would suggest to drop App/Application and only keep Optimization. What do you think ?!

It doesn't bother me. In my opinion "Optimization" is way too long for logo. Having just "Opt" is too short. "Optimiz", etc. doesn't sound like a correct shortening. Also, two other prints are just some standart Kratos logs and they may change in the future; hence, I think Logo should be in some sence independent from them.

RezaNajian:
@Igarizza ok fine then please merge it. Tnx.

RezaNajian (approved):

sunethwarna (approved):

PR #11106, opened 2023-05-10
Adding Run() methods to RomManager

Rbravo555:
**📝 Description**
So far, the RomManager allowed to **train** and **test** a reduced order model. However, we were missing a method in the RomManager for cheaply launching the generated models.
 
This PR adds the methods in the RomManager for running a set of simulations using either of the models. Clearly, if any of the required files for launching them is missing, the simulations will fail, as no training or snapshots matrices storage is considered. 

In order not to generate any conflict with currently open PRs, this PR is not tackling known shortcomings of the current approach, for example if a user wants to launch a ROM using a different amount of modes each time, she would have to modify the RomParameters.json directly. That is, not all of the ROM simulation parameters are controlled from the RomManager and a few are still relying on the RomParameters.json. A possible solution would necessarily involve modifying the RomAnalysis class to pass the RomParameters object instead of reading it from disk. This or other solutions will be deffered to future PRs.

**🆕 Changelog**
Added six methods to the RomManager.

Methods to be called from the instantiated RomManager 
- RunFOM(self, mu_run=None)
- RunROM(self, mu_run=None)
- RunHROM(self, mu_run=None)

Auxiliary methods inside the RomManager class:
- LaunchRunFOM(self, mu_run)
- LaunchRunROM(self, mu_run)
- LaunchRunHROM(self, mu_run)

The branch is compatible with the RomManager example in the Examples repo [here](https://github.com/KratosMultiphysics/Examples/tree/master/rom_application/RomManager)

### DIFF
--- a/applications/RomApplication/python_scripts/rom_manager.py
+++ b/applications/RomApplication/python_scripts/rom_manager.py
@@ -33,11 +33,11 @@ class RomManager(object):
         ######  Galerkin ######
         if chosen_projection_strategy == "galerkin":
             if any(item == "ROM" for item in training_stages):
-                fom_snapshots = self.LaunchTrainROM(mu_train)
+                fom_snapshots = self.__LaunchTrainROM(mu_train)
                 if store_all_snapshots or store_fom_snapshots:
                     self._StoreSnapshotsMatrix('fom_snapshots', fom_snapshots)
                 self._ChangeRomFlags(simulation_to_run = "GalerkinROM")
-                rom_snapshots = self.LaunchROM(mu_train)
+                rom_snapshots = self.__LaunchROM(mu_train)
                 if store_all_snapshots or store_rom_snapshots:
                     self._StoreSnapshotsMatrix('rom_snapshots', rom_snapshots)
                 self.ROMvsFOM_train = np.linalg.norm(fom_snapshots - rom_snapshots)/ np.linalg.norm(fom_snapshots)
@@ -45,24 +45,23 @@ class RomManager(object):
             if any(item == "HROM" for item in training_stages):
                 #FIXME there will be an error if we only train HROM, but not ROM
                 self._ChangeRomFlags(simulation_to_run = "trainHROMGalerkin")
-                self.LaunchTrainHROM(mu_train, store_residuals_projected)
+                self.__LaunchTrainHROM(mu_train, store_residuals_projected)
                 self._ChangeRomFlags(simulation_to_run = "runHROMGalerkin")
-                hrom_snapshots = self.LaunchHROM(mu_train)
+                hrom_snapshots = self.__LaunchHROM(mu_train)
                 if store_all_snapshots or store_hrom_snapshots:
                     self._StoreSnapshotsMatrix('hrom_snapshots', hrom_snapshots)
                 self.ROMvsHROM_train = np.linalg.norm(rom_snapshots - hrom_snapshots) / np.linalg.norm(rom_snapshots)
         #######################
 
-
         #######################################
         ##  Least-Squares Petrov Galerkin   ###
         elif chosen_projection_strategy == "lspg":
             if any(item == "ROM" for item in training_stages):
-                fom_snapshots = self.LaunchTrainROM(mu_train)
+                fom_snapshots = self.__LaunchTrainROM(mu_train)
                 if store_all_snapshots or store_fom_snapshots:
                     self._StoreSnapshotsMatrix('fom_snapshots', fom_snapshots)
                 self._ChangeRomFlags(simulation_to_run = "lspg")
-                rom_snapshots = self.LaunchROM(mu_train)
+                rom_snapshots = self.__LaunchROM(mu_train)
                 if store_all_snapshots or store_rom_snapshots:
                     self._StoreSnapshotsMatrix('rom_snapshots', rom_snapshots)
                 self.ROMvsFOM_train = np.linalg.norm(fom_snapshots - rom_snapshots)/ np.linalg.norm(fom_snapshots)
@@ -74,22 +73,22 @@ class RomManager(object):
         ###  Petrov Galerkin   ###
         elif chosen_projection_strategy == "petrov_galerkin":
             if any(item == "ROM" for item in training_stages):
-                fom_snapshots = self.LaunchTrainROM(mu_train)
+                fom_snapshots = self.__LaunchTrainROM(mu_train)
                 if store_all_snapshots or store_fom_snapshots:
                     self._StoreSnapshotsMatrix('fom_snapshots', fom_snapshots)
                 self._ChangeRomFlags(simulation_to_run = "TrainPG")
-                self.TrainPG(mu_train)
+                self.__LaunchTrainPG(mu_train)
                 self._ChangeRomFlags(simulation_to_run = "PG")
-                rom_snapshots = self.LaunchROM(mu_train)
+                rom_snapshots = self.__LaunchROM(mu_train)
                 if store_all_snapshots or store_rom_snapshots:
                     self._StoreSnapshotsMatrix('rom_snapshots', rom_snapshots)
                 self.ROMvsFOM_train = np.linalg.norm(fom_snapshots - rom_snapshots)/ np.linalg.norm(fom_snapshots)
             if any(item == "HROM" for item in training_stages):
                 #FIXME there will be an error if we only train HROM, but not ROM
                 self._ChangeRomFlags(simulation_to_run = "trainHROMPetrovGalerkin")
-                self.LaunchTrainHROM(mu_train, store_residuals_projected)
+                self.__LaunchTrainHROM(mu_train, store_residuals_projected)
                 self._ChangeRomFlags(simulation_to_run = "runHROMPetrovGalerkin")
-                hrom_snapshots = self.LaunchHROM(mu_train)
+                hrom_snapshots = self.__LaunchHROM(mu_train)
                 if store_all_snapshots or store_hrom_snapshots:
                     self._StoreSnapshotsMatrix('hrom_snapshots', hrom_snapshots)
                 self.ROMvsHROM_train = np.linalg.norm(rom_snapshots - hrom_snapshots) / np.linalg.norm(rom_snapshots)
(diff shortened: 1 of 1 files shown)


SADPR:
It looks good to me. I would wait for @rubenzorrilla 's opinion to merge it. But it is something very useful.

NicolasSR (commented):
Very nice addition. Seems fine to me

applications/RomApplication/python_scripts/rom_manager.py
```diff
@@ -154,6 +154,52 @@ def Test(self, mu_test=None):
 
 
 
+    def RunFOM(self, mu_run=None):
+        if mu_run is None:
```
rubenzorrilla:
```suggestion
        if not mu_run:
```
I think it's more pythonic this way.

applications/RomApplication/python_scripts/rom_manager.py
```diff
@@ -154,6 +154,52 @@ def Test(self, mu_test=None):
...
+            mu_run = ['single case with parameters already contained in the ProjectParameters.json and CustomSimulation']
+        self.LaunchRunFOM(mu_run)
+
+
+
+
```
rubenzorrilla:
```suggestion
```
Avoid extra whitespaces.

applications/RomApplication/python_scripts/rom_manager.py
```diff
@@ -154,6 +154,52 @@ def Test(self, mu_test=None):
 
 
 
+    def RunFOM(self, mu_run=None):
+        if mu_run is None:
+            mu_run = ['single case with parameters already contained in the ProjectParameters.json and CustomSimulation']
```
rubenzorrilla: Why assigning a value to it? Wouldn't it be better to keep the `None` default value?

applications/RomApplication/python_scripts/rom_manager.py
```diff
@@ -154,6 +154,52 @@ def Test(self, mu_test=None):
...
+        self.LaunchRunFOM(mu_run)
+
+
+
+
+    def RunROM(self, mu_run=None):
+        if mu_run is None:
+            mu_run = ['single case with parameters already contained in the ProjectParameters.json and CustomSimulation']
```
rubenzorrilla: (same comments)

applications/RomApplication/python_scripts/rom_manager.py
```diff
@@ -154,6 +154,52 @@ def Test(self, mu_test=None):
...
+        if self.general_rom_manager_parameters["projection_strategy"].GetString() == "galerkin":
+            self._ChangeRomFlags(simulation_to_run = "GalerkinROM")
+        #######################################
+        ##  Least-Squares Petrov Galerkin   ###
+        elif self.general_rom_manager_parameters["projection_strategy"].GetString() == "lspg":
+            self._ChangeRomFlags(simulation_to_run = "lspg")
+        ##########################
+        ###  Petrov Galerkin   ###
+        elif self.general_rom_manager_parameters["projection_strategy"].GetString() == "petrov_galerkin":
+            self._ChangeRomFlags(simulation_to_run = "PG")
```
rubenzorrilla, line 171:
```suggestion
            self._ChangeRomFlags(simulation_to_run = "PG")
        else:
            err_msg = f"Provided projection strategy '{self.general_rom_manager_parameters["projection_strategy"].GetString()}' is not supported. Available options are 'galerkin', 'lspg' and 'petrov_galerkin'."
            raise Exception(err_msg)
```

applications/RomApplication/python_scripts/rom_manager.py
```diff
@@ -154,6 +154,52 @@ def Test(self, mu_test=None):
...
+        self.LaunchRunROM(mu_run)
+
+
+
```
rubenzorrilla:
```suggestion
```
Avoid extra whitespaces.

applications/RomApplication/python_scripts/rom_manager.py
```diff
@@ -154,6 +154,52 @@ def Test(self, mu_test=None):
...
+            self._ChangeRomFlags(simulation_to_run = "runHROMGalerkin")
+        #######################################
+        ##  Least-Squares Petrov Galerkin   ###
+        elif self.general_rom_manager_parameters["projection_strategy"].GetString() == "lspg":
+            raise Exception('Sorry, Hyper Reduction not yet implemented for lspg')
+        ##########################
+        ###  Petrov Galerkin   ###
+        elif self.general_rom_manager_parameters["projection_strategy"].GetString() == "petrov_galerkin":
+            self._ChangeRomFlags(simulation_to_run = "runHROMPetrovGalerkin")
+        self.LaunchRunHROM(mu_run)
```
rubenzorrilla: I'd throw the same error message I suggest for `RunROM`.

applications/RomApplication/python_scripts/rom_manager.py
```diff
@@ -154,6 +154,52 @@ def Test(self, mu_test=None):
...
+        ###  Petrov Galerkin   ###
+        elif self.general_rom_manager_parameters["projection_strategy"].GetString() == "petrov_galerkin":
+            self._ChangeRomFlags(simulation_to_run = "PG")
+        self.LaunchRunROM(mu_run)
+
+
+
+    def RunHROM(self, mu_run=None):
+        if mu_run is None:
+            mu_run = ['single case with parameters already contained in the ProjectParameters.json and CustomSimulation']
```
rubenzorrilla: (same comment)

applications/RomApplication/python_scripts/rom_manager.py
```diff
@@ -167,6 +213,7 @@ def PrintErrors(self):
             print("approximation error in test set ROM vs HROM: ",  self.ROMvsHROM_test)
 
 
+
```
rubenzorrilla:
```suggestion
```

applications/RomApplication/python_scripts/rom_manager.py
```diff
@@ -192,6 +239,7 @@ def LaunchTrainROM(self, mu_train):
         return SnapshotsMatrix
 
 
+
```
rubenzorrilla:
```suggestion
```

applications/RomApplication/python_scripts/rom_manager.py
```diff
@@ -365,6 +413,54 @@ def LaunchTestHROM(self, mu_test):
         return SnapshotsMatrix
 
 
+    def LaunchRunFOM(self, mu_run):
```
rubenzorrilla:
```suggestion
    def __LaunchRunFOM(self, mu_run):
```
As far as I see these can be mangled.

applications/RomApplication/python_scripts/rom_manager.py
```diff
@@ -365,6 +413,54 @@ def LaunchTestHROM(self, mu_test):
...
+        for Id, mu in enumerate(mu_run):
+            parameters = self.UpdateProjectParameters(parameters, mu)
+            parameters = self._StoreResultsByName(parameters,'FOM_Run',mu,Id)
+            model = KratosMultiphysics.Model()
+            analysis_stage_class = self._GetAnalysisStageClass(parameters)
+            simulation = self.CustomizeSimulation(analysis_stage_class,model,parameters)
+            simulation.Run()
+
+
+    def LaunchRunROM(self, mu_run):
```
rubenzorrilla:
```suggestion
    def __LaunchRunROM(self, mu_run):
```

applications/RomApplication/python_scripts/rom_manager.py
```diff
@@ -365,6 +413,54 @@ def LaunchTestHROM(self, mu_test):
...
+        for Id, mu in enumerate(mu_run):
+            parameters = self.UpdateProjectParameters(parameters, mu)
+            parameters = self._StoreResultsByName(parameters,'ROM_Run',mu,Id)
+            model = KratosMultiphysics.Model()
+            analysis_stage_class = type(SetUpSimulationInstance(model, parameters))
+            simulation = self.CustomizeSimulation(analysis_stage_class,model,parameters)
+            simulation.Run()
+
+
+    def LaunchRunHROM(self, mu_run):
```
rubenzorrilla:
```suggestion
    def __LaunchRunHROM(self, mu_run):
```

applications/RomApplication/python_scripts/rom_manager.py
Rbravo555: In the past, when using `not mu_run`, sometime I came across the error 
    ValueError: The truth value of an array with more than one element is ambiguous.
On the other hand, `mu_run is None` always works independently of the type of mu_run

I must recognize that given the type of `mu` (either a Python list or None), the suggested code will always work. 

I would stick to the less Pythonic, yet more English and robust expression. But if you insist I will change it across the script

applications/RomApplication/python_scripts/rom_manager.py
Rbravo555: If the user doesn't  pass a list of parameters in `mu_run`, that means that a single case will be run with the parameters already contained in the Project Parameters.json. That is exactly the string we add to `mu_run`. Why we add this string? because later on (in `LaunchRunFOM`), we loop over the elements of `mu_run`, and if `mu_run` is None, we would have an error at

        for Id, mu in enumerate(mu_run):
            parameters = self.UpdateProjectParameters(parameters, mu)
            parameters = self._StoreResultsByName(parameters,'FOM_Run',mu,Id)
            model = KratosMultiphysics.Model()
            analysis_stage_class = self._GetAnalysisStageClass(parameters)
            simulation = self.CustomizeSimulation(analysis_stage_class,model,parameters)
            simulation.Run()

It's not too bad a trick, I believe.

rubenzorrilla: I think it is a bit hacky. What if we default as `mu_run = [None]`?

Rbravo555: This one I like more. That would also solve the comment above

Rbravo555:
I have considered all suggestions, thank you all. Now I only wait for green light to merge @rubenzorrilla

rubenzorrilla (approved):

PR #11110, opened 2023-05-11
[PFEMFluid] Minor API update in PFEM elements

rubenzorrilla:
Update in the `CalcElasticPlasticCauchySplitted` method API. I removed the time step argument as it is passed within the `ProcessInfo` and added the shape functions vector. On top of being required for the subintegration, this avoids calculating twice the shape functions.

I also took the chance to remove some compilation warnings.


AlejandroCornejo:
@AFranci

AFranci (approved):

PR #11111, opened 2023-05-11
[StructuralMechanicsApplication] Fix distributed load and added parallel utilities in the `distribute_load_on_surface_process`

### LINKED ISSUE #11109: Mesh size changes the value of distributed load.
**Description**

When distributed loading is applied to a surface using "distribute_load_on_surface_process" as shown below, the analysis results will vary depending on the mesh size.

```json
"loads_process_list": [
      {
        "python_module": "distribute_load_on_surface_process",
        "kratos_module": "KratosMultiphysics.StructuralMechanicsApplication",
        "process_name": "DistributeLoadOnSurfaceProcess",
        "help": "This process distributes a load on surface load conditions belonging to a modelpart. The load is distributed according to the surface area.",
        "Parameters": {
          "load": [
            0.0,
            0.0,
            100000.0
          ],
          "model_part_name": "plate.distribute_load_nodes_conditions"
        }
```

The deformation contours are visualized in Paraview. Both analyses are performed with the above settings, but the finer mesh results in smaller displacements.

image

Checking the reaction force, the one on the left is 634 and the one on the right is 107, and the reaction force values are different.
Since the reaction forces are different, it is not likely that the mesh size makes a difference in the accuracy of the analysis.

**Scope**
Where in Kratos is this bug happening. 

- StructuralMechanicsApplication

**To Reproduce**

Run "run_analysis.py" in the attached file
[model.zip](https://github.com/KratosMultiphysics/Kratos/files/11450402/model.zip)

**Expected behavior**

Both will have the same displacement and the same reaction force.

**Environment**

- OS: windows11
- Branch: use Release 9.3 via PYPI
- Python 3.9

loumalouomega:
**📝 Description**

Fixes https://github.com/KratosMultiphysics/Kratos/issues/11109

In this PR, the `distribute_load_on_surface_process` in the StructuralMechanicsApplication bugfrom https://github.com/KratosMultiphysics/Kratos/issues/11109 has been fixed, and part of the code now uses the parallel utilities.

**🆕 Changelog**

- [Fix distributed load process](https://github.com/KratosMultiphysics/Kratos/commit/9cf5b9d3cd8ee11658f74611a2c1233c100aa848)
	- `SURFACE_LOAD` now is a distributed load, instead of a total load, fixing the bug from https://github.com/KratosMultiphysics/Kratos/issues/11109
	- `parallel_utilities` and `reduction_utilities`, are included in the `distribute_load_on_surface_process.cpp`. These utilities allow for parallel computation. The calculation of total surface area has been improved, likely for better performance and parallel computation. Instead of a simple loop, `block_for_each<SumReduction<double>>` function is used, which is likely a parallel operation for summing all elements of a container.
	- The documentation and class definition in the `distribute_load_on_surface_process.h` file have been improved for better readability and understanding.

- [Update test](https://github.com/KratosMultiphysics/Kratos/commit/4623f8b4cb96ca71a97d46acab0cb1cbad730067)
	-  The test file, `test_distribute_load_on_surface_process.py`, is also updated to reflect these changes.

### DIFF
--- a/applications/StructuralMechanicsApplication/custom_processes/distribute_load_on_surface_process.cpp
+++ b/applications/StructuralMechanicsApplication/custom_processes/distribute_load_on_surface_process.cpp
@@ -16,30 +16,32 @@
 // Project includes
 #include "distribute_load_on_surface_process.h"
 #include "utilities/interval_utility.h"
+#include "utilities/parallel_utilities.h"
+#include "utilities/reduction_utilities.h"
 #include "structural_mechanics_application_variables.h"
 
 namespace Kratos
 {
 
-DistributeLoadOnSurfaceProcess::DistributeLoadOnSurfaceProcess(ModelPart& rModelPart,
-                                                                Parameters Settings)
-                                                                : mrModelPart(rModelPart),
-                                                                mParameters(Settings)
+DistributeLoadOnSurfaceProcess::DistributeLoadOnSurfaceProcess(
+    ModelPart& rModelPart,
+    Parameters Settings)
+    : mrModelPart(rModelPart),
+      mParameters(Settings)
 {
     Parameters default_parameters(R"(
         {
             "help"            : "This process distributes a load on surface load conditions belonging to a modelpart. The load is distributed according to the surface area.",
             "model_part_name" : "please_specify_model_part_name",
             "interval"        : [0.0, 1e30],
-            "load"           : [1.0, 0.0, 0.0]
+            "load"            : [1.0, 0.0, 0.0]
         }  )"
     );
 
     IntervalUtility interval_utility(mParameters);
 
     mParameters.RecursivelyValidateAndAssignDefaults(default_parameters);
-    KRATOS_ERROR_IF(mParameters["load"].GetVector().size() != 3) <<
-        "'load' has to be a vector of doubles with size 3!" << std::endl;
+    KRATOS_ERROR_IF(mParameters["load"].GetVector().size() != 3) << "'load' has to be a vector of doubles with size 3!" << std::endl;
 }
 
 void DistributeLoadOnSurfaceProcess::ExecuteInitializeSolutionStep()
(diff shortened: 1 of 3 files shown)


applications/StructuralMechanicsApplication/custom_processes/distribute_load_on_surface_process.cpp
```diff
@@ -48,18 +50,23 @@ void DistributeLoadOnSurfaceProcess::ExecuteInitializeSolutionStep()
...
     IntervalUtility interval_utility(mParameters);
     if (interval_utility.IsInInterval(current_time)) {
-        double total_area = 0.0;
-        for (auto& r_cond : mrModelPart.Conditions()) {
-            total_area += r_cond.GetGeometry().Area();
-        }
+        // Calculate the total area
+        const auto& r_communicator = mrModelPart.GetCommunicator();
+        const auto& r_local_mesh = r_communicator.LocalMesh();
+        const auto& r_conditions_array = r_local_mesh.Conditions();
```
rubenzorrilla: I think there is no need to take the local mesh as conditions will always be local.

applications/StructuralMechanicsApplication/custom_processes/distribute_load_on_surface_process.cpp
```diff
@@ -48,18 +50,23 @@ void DistributeLoadOnSurfaceProcess::ExecuteInitializeSolutionStep()
...
 
-        Vector force_by_area = mParameters["load"].GetVector() / global_total_area;
+        // Getting force by area
+        const Vector force_by_area = mParameters["load"].GetVector() / global_total_area;
 
+        // Assign on conditions
         for (auto& r_cond : mrModelPart.Conditions()) {
-            const double area = r_cond.GetGeometry().Area();
-            r_cond.SetValue(SURFACE_LOAD, force_by_area * area);
+            r_cond.SetValue(SURFACE_LOAD, force_by_area);
```
rubenzorrilla: I think this can be parallel.

loumalouomega: 100% true

rubenzorrilla (approved):
Fixes current issue plus some performance additions.

loumalouomega:
@roigcarlo should we do a minor release?

PR #11112, opened 2023-05-11
[GeoMechanicsApplication] Added a helper function that converts an angle in degrees to one in radians

### LINKED ISSUE #10946: [GeoMechanicsApplication] Add utility function for Radians → Degrees and Degrees → Radians conversions.
As described.

To remove inline conversions in calculations.

eg. In the piping calculations

avdg81:
**📝 Description**
Added a helper function that converts an angle in degrees to one in radians. With this function in place, the hard-coded conversions have been replaced by calls to this function.

**🆕 Changelog**
- Introduced nonmember function `DegreesToRadians`.
- Replaced the hard-coded angle conversions in the GeoMechanicsApplication by calling this new function.

### DIFF
--- a/applications/GeoMechanicsApplication/custom_elements/steady_state_Pw_piping_element.cpp
+++ b/applications/GeoMechanicsApplication/custom_elements/steady_state_Pw_piping_element.cpp
@@ -13,7 +13,10 @@
 // Application includes
 #include "custom_elements/steady_state_Pw_piping_element.hpp"
 #include "custom_utilities/element_utilities.hpp"
+#include "utilities/math_utils.h"
 #include <cmath>
+
+
 namespace Kratos
 {
 
@@ -328,7 +331,7 @@ double SteadyStatePwPipingElement<TDim,TNumNodes>:: CalculateEquilibriumPipeHeig
         return 1e10;
     }
 	
-    return modelFactor * M_PI / 3.0 * particle_d * (SolidDensity / FluidDensity - 1) * eta  * sin((theta  + pipeSlope) * M_PI / 180.0) / cos(theta * M_PI / 180.0) / dhdx;
+    return modelFactor * Globals::Pi / 3.0 * particle_d * (SolidDensity / FluidDensity - 1) * eta  * std::sin(MathUtils<>::DegreesToRadians(theta  + pipeSlope)) / std::cos(MathUtils<>::DegreesToRadians(theta)) / dhdx;
 
 }
 
--- a/kratos/tests/cpp_tests/utilities/test_math_utils.cpp
+++ b/kratos/tests/cpp_tests/utilities/test_math_utils.cpp
@@ -892,5 +892,16 @@ namespace Kratos
             KRATOS_CHECK_MATRIX_NEAR(exp_A, A, 1.0e-8);
         }
 
+        /** Checks whether the angle conversion from degrees to radians is performed correctly
+         */
+        KRATOS_TEST_CASE_IN_SUITE(MathUtilsDegreesToRadians, KratosCoreFastSuite)
+        {
+            const double abs_tolerance = 1e-9;
+            KRATOS_CHECK_NEAR(          0.0, MathUtils<>::DegreesToRadians(   0.0), abs_tolerance);
+            KRATOS_CHECK_NEAR(Globals::Pi/2, MathUtils<>::DegreesToRadians(  90.0), abs_tolerance);
+            KRATOS_CHECK_NEAR( -Globals::Pi, MathUtils<>::DegreesToRadians(-180.0), abs_tolerance);
+            KRATOS_CHECK_NEAR(4*Globals::Pi, MathUtils<>::DegreesToRadians( 720.0), abs_tolerance);
+        }
+
     } // namespace Testing
 }  // namespace Kratos.
--- a/kratos/utilities/math_utils.h
+++ b/kratos/utilities/math_utils.h
@@ -25,6 +25,7 @@
 // Project includes
 #include "input_output/logger.h"
 #include "includes/ublas_interface.h"
+#include "includes/global_variables.h"
 
 namespace Kratos
 {
@@ -1817,6 +1818,12 @@ public:
                 break;
             series_term++;
         }
+    }
+
+
+    static double DegreesToRadians(double AngleInDegrees)
+    {
+        return (AngleInDegrees * Globals::Pi) / 180.0;
     }
 
     ///@}


applications/GeoMechanicsApplication/custom_utilities/math_utilities.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include "math_utilities.hpp"
+
+
+namespace Kratos
+{
+
+double DegreesToRadians(double AngleInDegrees)
+{
+    return (AngleInDegrees * M_PI) / 180.0;
```
philbucher:
```suggestion
    return (AngleInDegrees * Globals::Pi) / 180.0;
```

philbucher (commented):
would be cool to add this to the math_utils in the Core

applications/GeoMechanicsApplication/custom_utilities/math_utilities.cpp
```diff
@@ -0,0 +1,24 @@
...
+//  Main authors:    Anne van de Graaf
+//
+
+#include "math_utilities.hpp"
+
+
+namespace Kratos
+{
+
+double DegreesToRadians(double AngleInDegrees)
```
AlejandroCornejo:
```suggestion
double DegreesToRadians(const double AngleInDegrees)
```

applications/GeoMechanicsApplication/custom_utilities/math_utilities.hpp
```diff
@@ -1236,5 +1236,8 @@ class StructuralMechanicsMathUtilities
 
 private:
 };// class StructuralMechanicsMathUtilities
+
+KRATOS_API(GEO_MECHANICS_APPLICATION) double DegreesToRadians(double AngleInDegrees);
```
rubenzorrilla: You're declaring the function at the global Kratos namespace. Though this is technically possible, in Kratos we always prefer the `static` but inside a class approach. As suggested by @philbucher, I'd place this within the Kratos core math utils.

avdg81: Thanks, I will move the function to the Kratos core math utils as suggested by @philbucher and you. However, what is the rationale behind using a static member function rather than a nonmember function? If you're trying to limit its scope (which is a fair point), I feel that using a nested namespace is a simpler and cleaner way to achieve just that.
I will follow your suggestion anyway, but I'm just asking you to think about this one and to consider whether we may want to change this policy. Thanks.

applications/GeoMechanicsApplication/custom_utilities/math_utilities.cpp
avdg81: I'm not sure whether this actually improves the code. I have checked the C++ Core Guidelines and there they explicitly mention this situation as an exception to the rule that objects should be immutable whenever possible. From [section Con 1: By default, make objects immutable](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#con1-by-default-make-objects-immutable) it says (under the heading Exception):

> Function parameters passed by value are rarely mutated, but also rarely declared `const`. To avoid confusion and lots of false positives, don’t enforce this rule for function parameters.

Also the accompanying example illustrates this:
```
void g(const int i) { ... }  // pedantic
```

Since I don't see a good reason to deviate from that recommendation here, I'd like to keep it the way it is. (Unless you have a convincing reason why we should make it immutable here.) Thanks.

applications/GeoMechanicsApplication/custom_utilities/math_utilities.hpp
rubenzorrilla: Yes, my point is to basically limit its scope. About how to achieve it, i.e. the class vs namespace, this is a long term discussion in Kratos. At some point we tried to provide some guidelines & conclusions in #7434.

avdg81:
> would be cool to add this to the math_utils in the Core

I have looked into that, but it's not as straightforward as it may seem at first sight. Let me explain. `MathUtils` is actually a _template_ class, which means it needs a template argument (`TDataType`) in order to be instantiated. Surely we wouldn't want the return type to be of this type, since unexpected things may happen. For instance, consider `TDataType` to be set to `int`, it would simply truncate the returned value (which almost always will be a floating point number). That's probably not what you want. So a fixed return type of `double` seems more logical. You might argue that the input parameter could be of type `TDataType`, but in the end it needs to be converted to a `double` anyway. So, if we would express that explicitly (which is how my current function signature looks like), then the template parameter is not used in any way. But if you want to call the function, you'd still have to supply a template argument, which opens the door to silly code constructs like `MathUtils<std::string>::DegreesToRadians(90);`.

The whole problem here stems from the fact that the template parameter is actually redundant for this particular case. To me, whichever way we go here, it will never be really elegant in my opinion (please correct me if I have overlooked a good option). Therefore, I'm inclined to keep it the way it is now and leave it up to you to move it to a more general location (say somewhere in Kratos core) in a way that you feel best fits the way Kratos does things. Any feedback is greatly appreciated. Thanks.

philbucher:
sounds reasonable

@rubenzorrilla @loumalouomega maybe it is time to drop the template arg for the math utils? I don't think any type other than `double` is used

loumalouomega:
> sounds reasonable
> 
> @rubenzorrilla @loumalouomega maybe it is time to drop the template arg for the math utils? I don't think any type other than `double` is used

I started to remove it long time ago in a branch, I can retake it

rubenzorrilla:
> > sounds reasonable
> > @rubenzorrilla @loumalouomega maybe it is time to drop the template arg for the math utils? I don't think any type other than `double` is used
> 
> I started to remove it long time ago in a branch, I can retake it


> sounds reasonable
> 
> @rubenzorrilla @loumalouomega maybe it is time to drop the template arg for the math utils? I don't think any type other than `double` is used

I was about to suggest this.

applications/GeoMechanicsApplication/custom_elements/steady_state_Pw_piping_element.cpp
```diff
@@ -328,7 +331,7 @@ double SteadyStatePwPipingElement<TDim,TNumNodes>:: CalculateEquilibriumPipeHeig
         return 1e10;
     }
 	
-    return modelFactor * M_PI / 3.0 * particle_d * (SolidDensity / FluidDensity - 1) * eta  * sin((theta  + pipeSlope) * M_PI / 180.0) / cos(theta * M_PI / 180.0) / dhdx;
+    return modelFactor * M_PI / 3.0 * particle_d * (SolidDensity / FluidDensity - 1) * eta  * sin(MathUtils<>::DegreesToRadians(theta  + pipeSlope)) / cos(MathUtils<>::DegreesToRadians(theta)) / dhdx;
```
philbucher:
```suggestion
    return modelFactor * Globals::Pi / 3.0 * particle_d * (SolidDensity / FluidDensity - 1) * eta  * std::sin(MathUtils<>::DegreesToRadians(theta  + pipeSlope)) / std::cos(MathUtils<>::DegreesToRadians(theta)) / dhdx;
```

philbucher (commented):
thanks!

philbucher (dismissed):

applications/GeoMechanicsApplication/custom_elements/steady_state_Pw_piping_element.cpp
avdg81: Done. Thanks for the suggestion!

philbucher (approved):
Nice!

huhaas (approved):

roigcarlo (approved):

PR #11114, opened 2023-05-11
[FreeSurface] Remove useless files

rubenzorrilla:
**📝 Description**
In this PR I'm removing a few do-nothing files. Note that the processes are the ones automatically generated as a sample by the application generator. The Python export ones are basically doing nothing as there is no custom strategy to be exported.


joaquinirazabal (approved):
Thanks for the clean up, @rubenzorrilla

PR #11116, opened 2023-05-12
[Core] Added GetThreadId function to ParallelUtilities for thread ID retrieval

loumalouomega:
**📝 Description**

This commit mainly focuses on the addition of the `GetThreadId` function in the `ParallelUtilities` class. This function is implemented to provide the thread number for the current thread, and returns 0 if it is a scalar run. The implementation is done considering different scenarios such as `KRATOS_SMP_OPENMP` and `KRATOS_SMP_CXX11`.

For `KRATOS_SMP_OPENMP`, it retrieves the thread number using `omp_get_thread_num()`. For `KRATOS_SMP_CXX11`, an error message is returned, indicating the lack of an equivalent definition for `GetThreadId` in C++11. If neither of these macros are defined, the function defaults to return 0.

In addition to this main change, the indentation of the license comments in parallel_utilities.cpp file has been corrected for better readability.

https://github.com/KratosMultiphysics/Kratos/pull/9985 will be the proper solution for this, not requiring the `GetThreadId` at all, but it will require to refactor the classes currently using `GetThreadId` and `ThisThread` (`OMPUtils`)

**🆕 Changelog**

- [Adding method to reduce dependency in Legacy OMPUtils](https://github.com/KratosMultiphysics/Kratos/commit/9438f45d02766ea5ec1089dc3712a17d6e7ff624)

### DIFF
--- a/kratos/utilities/parallel_utilities.cpp
+++ b/kratos/utilities/parallel_utilities.cpp
@@ -4,8 +4,8 @@
 //   _|\_\_|  \__,_|\__|\___/ ____/
 //                   Multi-Physics
 //
-//  License:		 BSD License
-//					 Kratos default license: kratos/license.txt
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
 //
 //  Main authors:    Riccardo Rossi
 //                   Denis Demidov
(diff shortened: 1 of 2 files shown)


philbucher (changes requested):
as said before, this should not be added, rather the corresponding fct in the OMPutils should be deprecated

the affected code should be replaced with #9985

The ParallelUtils only contain code that we know has solutions for both versions of parallelization

loumalouomega:
> as said before, this should not be added, rather the corresponding fct in the OMPutils should be deprecated
> 
> the affected code should be replaced with #9985
> 
> The ParallelUtils only contain code that we know has solutions for both versions of parallelization

Okay, if I add a deprecation to the current OMPUtils?

loumalouomega:
> then there will be lots of warnings, without a proper way for solving them
> 
> => I would do it after #9985 is merged

Okay, but in the other PR, I will simply revert the include

PR #11119, opened 2023-05-12
[Core] Separate  `Node` into source and header file

loumalouomega:
**📝 Description**

This PR involves a significant refactoring of the Node class located in `kratos/includes/node.h` and creates a new file `kratos/sources/node.cpp`. The `Node` class has been thoroughly documented with detailed comments explaining the purpose and functionality of methods and member variables.

Overall, this PR brings significant improvements to code readability and maintainability due to the added comments and the clear structuring of the class, and also reduce compilation times.

**🆕 Changelog**

- [Separate Node into source and header file](https://github.com/KratosMultiphysics/Kratos/commit/7b96e08b88869ae97713bc235abc9a699f2eb6df)
- [Missing includes](https://github.com/KratosMultiphysics/Kratos/commit/00120b56c5d9619d83f26a7fbce08da40520e4cb)


kratos/includes/node.h
```diff
@@ -60,10 +49,15 @@ class Element;
...
-*/
-class Node : public Point, public Flags
+/**
+ * @class Node
+ * @brief This class defines the node
+ * @details The node class from Kratos is defined in this class
+ * @ingroup KratosCore
+ * @author Pooyan Dadvand
+ */
+class Node
```
philbucher:
```suggestion
class KRATOS_API(KRATOS_CORE) Node

```

kratos/includes/node.h
```diff
@@ -200,92 +162,76 @@ class Node : public Point, public Flags
...
-    }
+    ~Node() override;
 
     //*********************************************
     //public API of intrusive_ptr
-    unsigned int use_count() const noexcept
-    {
-        return mReferenceCounter;
-    }
+    unsigned int use_count() const noexcept;
```
roigcarlo: Much as I like, all these need to remain in the header. Its the same problem as with the point split. Everything that we are sure that can be inlined needs to remain in the header (Getter, Setters and non complex functions mainly).

loumalouomega: okay

loumalouomega: Done

roigcarlo: @loumalouomega **Everything** means ALL getters and setters (and all the functions that do simple access with no modifications)

loumalouomega: Okay, sorry for the noise. Anyway if we don't put inline the compiler may ignore it, specially gcc which is not very friend of inline

roigcarlo: Yep, we were reading it with @RiccardoRossi last week and seems that `inline`is just a hint, it may ignore it if you put it, or it may add it if you don't. Seems to be under no control. 

What is true is that if we divide it then it only can perform the inlining in the same compilation unit, so it could do it in the core (regardless of the function attributes) but won't do it under any circumstance for example in the FluidDynamics. The later is the situation we want to avoid.

loumalouomega: I improved the inlining

kratos/includes/node.h
```diff
@@ -14,33 +14,22 @@
 #pragma once
 
 // System includes
-#include <string>
-#include <iostream>
-#include <sstream>
-#include <cstddef>
-#include <atomic>
```
philbucher: this is used, please dont remove it

loumalouomega: But it is included in other dependencies..., I will revert it

loumalouomega:
This is ready @roigcarlo

kratos/sources/node.cpp
```diff
@@ -0,0 +1,354 @@
...
+    mData = rOther.mData;
+    mInitialPosition = rOther.mInitialPosition;
+
+    return *this;
+}
+
+/***********************************************************************************/
+/***********************************************************************************/
+
+bool Node::operator==(const Node& rOther)
```
RiccardoRossi: this should be in header

kratos/sources/node.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+double& Node::operator[](const IndexType ThisIndex)
+{
+    return BaseType::operator[](ThisIndex);
+}
+
+/***********************************************************************************/
+/***********************************************************************************/
+
+double Node::operator[](const IndexType ThisIndex) const
```
RiccardoRossi: same

kratos/sources/node.cpp
```diff
@@ -0,0 +1,354 @@
...
+/***********************************************************************************/
+/***********************************************************************************/
+
+void Node::SetInitialPosition(
+    const double X,
+    const double Y,
+    const double Z
+    )
+{
+    mInitialPosition.X() = X;
```
RiccardoRossi: same

kratos/sources/node.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+void Node::CloneSolutionStepData()
+{
+    SolutionStepData().CloneFront();
+}
+
+/***********************************************************************************/
+/***********************************************************************************/
+
+Node::IndexType Node::GetBufferSize() const
```
RiccardoRossi: header

kratos/sources/node.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+VariablesList::Pointer Node::pGetVariablesList()
+{
+    return SolutionStepData().pGetVariablesList();
+}
+
+/***********************************************************************************/
+/***********************************************************************************/
+
+const VariablesList::Pointer Node::pGetVariablesList() const
```
RiccardoRossi: header

kratos/sources/node.cpp
```diff
@@ -0,0 +1,354 @@
...
+{
+    mInitialPosition.X() = X;
+    mInitialPosition.Y() = Y;
+    mInitialPosition.Z() = Z;
+}
+
+/***********************************************************************************/
+/***********************************************************************************/
+
+VariablesList::Pointer Node::pGetVariablesList()
```
RiccardoRossi: header

kratos/sources/node.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+bool Node::operator==(const Node& rOther)
+{
+    return PointType::operator ==(rOther);
+}
+
+/***********************************************************************************/
+/***********************************************************************************/
+
+double& Node::operator[](const IndexType ThisIndex)
```
RiccardoRossi: header

RiccardoRossi (changes requested):
in essence all small functions should remain in header to ensure that they are inlined not just in the core but also in derived applications. 

at this point i think it does not really make sense to split the class in header and cpp

As i recall @roigcarlo was mentioning this...

loumalouomega:
> in essence all small functions should remain in header to ensure that they are inlined not just in the core but also in derived applications.
> 
> at this point i think it does not really make sense to split the class in header and cpp
> 
> As i recall @roigcarlo was mentioning this...

Comments adressed

loumalouomega:
@RiccardoRossi comments addressed

loumalouomega:
> @RiccardoRossi comments addressed

Hello

pooyan-dadvand:
@KratosMultiphysics/technical-committee would **not approve** this one as our objective is to let compiler make inline all the methods it considers that can be inlined. 

In general, we would not suggest dividing into header and cpp the classes that are mostly data and with methods that can be inlined. This includes many classes (mostly the ones in the modelpart) like PointerVectors, Point, Variable, IndexedObject, etc.

loumalouomega:
See my comments in https://github.com/KratosMultiphysics/Kratos/pull/11086

PR #11121, opened 2023-05-12
[Mapping] optionally allow to not compute approximation

philbucher:
This allows to disable the computation of an approximation with the nearest element mapper
This is actually how I thought it would work in the first place, but I implemented it wrong

@akhilkmat can you please test if this now works for you? (still with setting the local coord tolerance to zero)

### DIFF
--- a/applications/MappingApplication/custom_mappers/nearest_element_mapper.cpp
+++ b/applications/MappingApplication/custom_mappers/nearest_element_mapper.cpp
@@ -14,6 +14,7 @@
 //  Framework for Non-Matching Grid Mapping"
 
 // System includes
+#include <limits>
 
 // External includes
 
@@ -56,7 +57,7 @@ void NearestElementInterfaceInfo::SaveSearchResult(const InterfaceObject& rInter
     if (is_full_projection) {
         SetLocalSearchWasSuccessful();
     } else {
-        if (!ComputeApproximation || mLocalCoordTol < std::numeric_limits<double>::epsilon) {
+        if (!ComputeApproximation || mLocalCoordTol < std::numeric_limits<double>::epsilon()) {
             return;
         } else {
             SetIsApproximation();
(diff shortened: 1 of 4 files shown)


applications/MappingApplication/custom_mappers/nearest_element_mapper.cpp
akhilkmat, line 1: Hi Philipp,

Thanks for the changes. It solved the main issue of 'particles outside the element (while doing dem->fem) getting mapped' with "local_coord_tolerance"  : 0, however the 'conservative fem->dem mapping (with use transpose)' no longer works ( non-conservative fem->dem mapping works). I think this is probably so because it used the approximation part of the code for the conservative fem->dem mapping. 

If someone wants to do the conservative way, probably creating a second mapper with non zero, positive "local_coord_tolerance" could be a solution.

akhilkmat (dismissed):

applications/MappingApplication/custom_mappers/nearest_element_mapper.cpp
philbucher: > I think this is probably so because it used the approximation part of the code for the conservative fem->dem mapping.

Yes now that you say it, this is expected, makes sense

> probably creating a second mapper

this anyway happens internally

Have you considered using the barycentric mapper for mapping dem->fem? This way you could still get the interpolative properties of the mapping

philbucher:
glad that it works. I will do some more refactorings, then it will be ready for merging

philbucher:
@akhilkmat I changed how to disable the usage of the approximation: set `use_approximation` to `false`, pretty straight forward.
Can you please check it again?

akhilkmat:
> @akhilkmat I changed how to disable the usage of the approximation: set `use_approximation` to `false`, pretty straight forward. Can you please check it again?

Yes, it is working fine. thanks.

philbucher:
can you please approve?

akhilkmat (approved):

PR #11128, opened 2023-05-15
[StructuralMechanicsApplication] Damping parameters fix for explicit simulations

### LINKED ISSUE #11107: FEM explicit dynamic solver - No difference - changing Damping parameters.
**Description**

Hi all,

As part of initial problem setup for a coupled DEM-FEM problem, I am trying to simulate a simple FEM structural analysis problem using Dynamic solver.

The problem consists of a small column resting on 4 fixed nodes in gravity.
Domain 0.02 x 0.02 x 0.16 (height) - 8 cube elements of size 0.02 x 0.02 x 0.02. 
The bottom most 4 nodes are fixed.
The simulations lasts 0.5 sec real time.

Problem:
1. When i analyze the reaction from the bottom fixed nodes (images attached), it keeps on oscillating about a mean value, even after changing the Rayleigh_alpha and Rayleigh_beta coefficients ( coefficients are changed over a wide range), there is no change in the graph at all for varying coefficient values; i wonder whether the solver actually reads the input values? or am I missing something? (Ps: i tried both linear and non linear analysis as well)

2. I ran the same problem in static solver and it gives me expected results ie.. weight = reaction. Although the reaction in the static solver is positive and in the dynamic solver it is negative but the magnitude of the mean value of reaction in the dynamic solver is almost same as the magnitude of static solver. The reaction is expected to be positive since gravity acts in the -ive direction. I wonder whether the dynamic solver has a different sign convention compared to the static solver?

I am attaching the projectparameter.json file (in .txt extention) and some images for your reference, the displacement graph shows the displacement of one of the nodes just above the fixed node.
[ProjectParameters.txt](https://github.com/KratosMultiphysics/Kratos/files/11444162/ProjectParameters.txt)
reaction
displacement_lowest element-top nod
[ProjectParameters.txt
e 32](https://github.com/KratosMultiphysics/Kratos/assets/69186170/d84c6b9a-0611-43e4-bf81-ea079dce6b16)

Please help,
Best,
Akhil Mathews

**Scope**
FEM
structure

loumalouomega:
**📝 Description**

Fixes https://github.com/KratosMultiphysics/Kratos/issues/11107. 

The Rayleigh damping parameters, `RAYLEIGH_ALPHA` and `RAYLEIGH_BETA`, are set for each property in the main_model_part. A warning message is also added, mentioning that this support is legacy and will be removed in the future. The changes are done in the `_CreateScheme(self)` method of the `structural_mechanics_explicit_dynamic_solver.py` script.

This patch addresses a bug fix in the `base_solid_element.cpp` file. A missing condition is added in the `AddExplicitContribution` method, which checks if `RAYLEIGH_ALPHA` or `RAYLEIGH_BETA` exists in `rCurrentProcessInfo`, in addition to checking in `r_prop`.

**🆕 Changelog**

- [Clean up](https://github.com/KratosMultiphysics/Kratos/commit/8f679242cf49c1a6721efc65e1e2abbe13e51e6a) and [more cleaning](https://github.com/KratosMultiphysics/Kratos/commit/91fba4c444cbe2692056580e7639196029760cec): Unnecessary parentheses in if conditions are removed, improving the readability of the code.
- [Setting variable](https://github.com/KratosMultiphysics/Kratos/commit/e9b8f74fd4a6ff5f3336147f1b67e82600847386)
- [Missing check in base solid element](https://github.com/KratosMultiphysics/Kratos/commit/a961352dd973c7197c71a65ccf861440b8e66f23)

### DIFF
--- a/applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
+++ b/applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
@@ -501,7 +501,7 @@ void BaseSolidElement::AddExplicitContribution(
     Vector damping_residual_contribution = ZeroVector(element_size);
 
     // Calculate damping contribution to residual -->
-    if (r_prop.Has(RAYLEIGH_ALPHA) || r_prop.Has(RAYLEIGH_BETA)) {
+    if (r_prop.Has(RAYLEIGH_ALPHA) || r_prop.Has(RAYLEIGH_BETA) || rCurrentProcessInfo.Has(RAYLEIGH_ALPHA) || rCurrentProcessInfo.Has(RAYLEIGH_BETA)) {
         Vector current_nodal_velocities = ZeroVector(element_size);
         this->GetFirstDerivativesVector(current_nodal_velocities);
 
--- a/applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
+++ b/applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
@@ -501,7 +501,7 @@ void BaseSolidElement::AddExplicitContribution(
     Vector damping_residual_contribution = ZeroVector(element_size);
 
     // Calculate damping contribution to residual -->
-    if (r_prop.Has(RAYLEIGH_ALPHA) || r_prop.Has(RAYLEIGH_BETA) || rCurrentProcessInfo.Has(RAYLEIGH_ALPHA) || rCurrentProcessInfo.Has(RAYLEIGH_BETA)) {
+    if (StructuralMechanicsElementUtilities::HasRayleighDamping(r_prop, rCurrentProcessInfo)) {
         Vector current_nodal_velocities = ZeroVector(element_size);
         this->GetFirstDerivativesVector(current_nodal_velocities);
 
--- a/applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_explicit_dynamic_solver.py
+++ b/applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_explicit_dynamic_solver.py
@@ -1,6 +1,9 @@
 # Importing the Kratos Library
 import KratosMultiphysics
 
+# Import sys
+import sys
+
 # Import applications
 import KratosMultiphysics.StructuralMechanicsApplication as StructuralMechanicsApplication
 
@@ -97,16 +100,23 @@ class ExplicitMechanicalSolver(MechanicalSolver):
 
         # Setting the Rayleigh damping parameters
         rayleigh_alpha = self.settings["rayleigh_alpha"].GetDouble()
+        if rayleigh_alpha > sys.float_info.epsilon:
+            for prop in self.main_model_part.Properties:
+                if prop.Has(StructuralMechanicsApplication.RAYLEIGH_ALPHA):
+                    prop.SetValue(StructuralMechanicsApplication.RAYLEIGH_ALPHA, rayleigh_alpha)
         rayleigh_beta = self.settings["rayleigh_beta"].GetDouble()
-        for prop in self.main_model_part.Properties:
-            prop.SetValue(StructuralMechanicsApplication.RAYLEIGH_ALPHA], rayleigh_alpha)
-            prop.SetValue(StructuralMechanicsApplication.RAYLEIGH_BETA], rayleigh_beta)
+        if rayleigh_beta > sys.float_info.epsilon:
+            for prop in self.main_model_part.Properties:
+                if prop.Has(StructuralMechanicsApplication.RAYLEIGH_BETA):
+                    prop.SetValue(StructuralMechanicsApplication.RAYLEIGH_BETA, rayleigh_beta)
 
         # Setting the Rayleigh damping parameters (legacy on ProcessInfo)
         KratosMultiphysics.Logger.PrintWarning("ExplicitMechanicalSolver", "RAYLEIGH_ALPHA and RAYLEIGH_BETA are set in explicit simulation, this support is legacy and will be removed at some point")
         process_info = self.main_model_part.ProcessInfo
-        process_info[StructuralMechanicsApplication.RAYLEIGH_ALPHA] = rayleigh_alpha
-        process_info[StructuralMechanicsApplication.RAYLEIGH_BETA] = rayleigh_beta
+        if rayleigh_alpha > sys.float_info.epsilon:
+            process_info[StructuralMechanicsApplication.RAYLEIGH_ALPHA] = rayleigh_alpha
+        if rayleigh_beta > sys.float_info.epsilon:
+            process_info[StructuralMechanicsApplication.RAYLEIGH_BETA] = rayleigh_beta
 
         # Setting the time integration schemes
         if scheme_type == "central_differences":
--- a/applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_explicit_dynamic_solver.py
+++ b/applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_explicit_dynamic_solver.py
@@ -1,9 +1,6 @@
 # Importing the Kratos Library
 import KratosMultiphysics
 
-# Import sys
-import sys
-
 # Import applications
 import KratosMultiphysics.StructuralMechanicsApplication as StructuralMechanicsApplication
 
@@ -99,14 +96,9 @@ class ExplicitMechanicalSolver(MechanicalSolver):
         scheme_type = self.settings["scheme_type"].GetString()
 
         # Setting the Rayleigh damping parameters
-        KratosMultiphysics.Logger.PrintWarning("ExplicitMechanicalSolver", "RAYLEIGH_ALPHA and RAYLEIGH_BETA are set in explicit simulation, this support is legacy and will be removed at some point")
         process_info = self.main_model_part.ProcessInfo
-        rayleigh_alpha = self.settings["rayleigh_alpha"].GetDouble()
-        if rayleigh_alpha > sys.float_info.epsilon:
-            process_info[StructuralMechanicsApplication.RAYLEIGH_ALPHA] = rayleigh_alpha
-        rayleigh_beta = self.settings["rayleigh_beta"].GetDouble()
-        if rayleigh_beta > sys.float_info.epsilon:
-            process_info[StructuralMechanicsApplication.RAYLEIGH_BETA] = rayleigh_beta
+        process_info[StructuralMechanicsApplication.RAYLEIGH_ALPHA] = self.settings["rayleigh_alpha"].GetDouble()
+        process_info[StructuralMechanicsApplication.RAYLEIGH_BETA] = self.settings["rayleigh_beta"].GetDouble()
 
         # Setting the time integration schemes
         if scheme_type == "central_differences":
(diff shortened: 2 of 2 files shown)


applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
```diff
@@ -501,7 +501,7 @@ void BaseSolidElement::AddExplicitContribution(
     Vector damping_residual_contribution = ZeroVector(element_size);
 
     // Calculate damping contribution to residual -->
-    if (r_prop.Has(RAYLEIGH_ALPHA) || r_prop.Has(RAYLEIGH_BETA)) {
+    if (r_prop.Has(RAYLEIGH_ALPHA) || r_prop.Has(RAYLEIGH_BETA) || rCurrentProcessInfo.Has(RAYLEIGH_ALPHA) || rCurrentProcessInfo.Has(RAYLEIGH_BETA)) {
```
loumalouomega: Note, the check for the ProcessInfo is done later, the problem was the initial check

applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_explicit_dynamic_solver.py
```diff
@@ -96,18 +96,25 @@ def _CreateScheme(self):
...
 
         # Setting the Rayleigh damping parameters
+        rayleigh_alpha = self.settings["rayleigh_alpha"].GetDouble()
+        rayleigh_beta = self.settings["rayleigh_beta"].GetDouble()
+        for prop in self.main_model_part.Properties:
+            prop.SetValue(StructuralMechanicsApplication.RAYLEIGH_ALPHA], rayleigh_alpha)
+            prop.SetValue(StructuralMechanicsApplication.RAYLEIGH_BETA], rayleigh_beta)
+
+        # Setting the Rayleigh damping parameters (legacy on ProcessInfo)
+        KratosMultiphysics.Logger.PrintWarning("ExplicitMechanicalSolver", "RAYLEIGH_ALPHA and RAYLEIGH_BETA are set in explicit simulation, this support is legacy and will be removed at some point")
```
loumalouomega: Do you agree adding this warning or better just remove it?

ipouplana (commented):
It seems right to me. Let’s see what @KratosMultiphysics/structural-mechanics have to say.

loumalouomega:
> It seems right to me. Let’s see what @KratosMultiphysics/structural-mechanics have to say.

And @rubenzorrilla as he participated in #11107

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
rubenzorrilla: You mean that we're already checking for both properties and process info in `CalculateDampingMatrixWithLumpedMass` right?

loumalouomega: Yep

applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_explicit_dynamic_solver.py
rubenzorrilla: I'm not an expert on the topic but if I'm not wrong you may compute the damping coefficients based on a previous modal analysis. Then these would be global, that is, the same for all the properties so in my opinion it would make sense to set them through the project parameters.

@RiccardoRossi I'd like to hear your opinion about this.

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
philbucher: Please use the fct from the element utils, otherwise it is inconsistent with the implicit dynamics

applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_explicit_dynamic_solver.py
```diff
@@ -96,18 +96,25 @@ def _CreateScheme(self):
         scheme_type = self.settings["scheme_type"].GetString()
 
         # Setting the Rayleigh damping parameters
+        rayleigh_alpha = self.settings["rayleigh_alpha"].GetDouble()
+        rayleigh_beta = self.settings["rayleigh_beta"].GetDouble()
+        for prop in self.main_model_part.Properties:
+            prop.SetValue(StructuralMechanicsApplication.RAYLEIGH_ALPHA], rayleigh_alpha)
+            prop.SetValue(StructuralMechanicsApplication.RAYLEIGH_BETA], rayleigh_beta)
+
+        # Setting the Rayleigh damping parameters (legacy on ProcessInfo)
```
philbucher: This is not legacy!

Please check the implicit solver, there is an established workflow

philbucher (changes requested):
Please check the implicit workflow, and make it consistent

applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_explicit_dynamic_solver.py
philbucher: This workflow is already established, see the implicit solver

loumalouomega: Yes, Rayleigh parameters are estimated from modal analysis (there is a process I implemented myself), see https://github.com/KratosMultiphysics/Kratos/blob/master/applications/StructuralMechanicsApplication/custom_utilities/rayleigh_damping_coefficients_utilities.cpp

But you may calculate in advance and set it manually to avoid the initial cost (is quite costly the modal analysis). And is not global, you should compute it for every structural "object", if all the structure is bound is global, but for a contact problem the two contacting domains may have independent damping coefficients.

The point is, how do we do in that case, because in that case setting it on the solver may overwrite damping coefficients...

Okay I will patch this

applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_explicit_dynamic_solver.py
loumalouomega: We should make it legacy for inconsistent

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
loumalouomega: Nope, in explicit we can not set the same strategies. The computational cost is increased by a lot and makes the simulation very expensive

applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_explicit_dynamic_solver.py
loumalouomega: As I said, we should reconsider to consider a property, not ProcessInfo. Damping is a property of a material/structural set, not something global

applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_explicit_dynamic_solver.py
loumalouomega: I did changes accordingly to my reasoning

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
loumalouomega: In fact, I would not consider at all RAYLEIGH_BETA (stiffness damping) as in explicit is totally killer. In fact, the coherent thing would be tom directly provide the damping coefficient in explicit simulations. BTW, someone mentioned that we confuse damping ratio with damping coefficient, we should check that

philbucher:
Sorry I think I confused sth
Need some time to look at this in detail

applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_explicit_dynamic_solver.py
loumalouomega: And as I said, explicit cannot be totally consistent with implicit because otherwise computational cost will be huge

applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
philbucher:
```suggestion
    if (HasRayleighDamping(r_prop, rCurrentProcessInfo)) {
```

applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_explicit_dynamic_solver.py
philbucher: I took a look now. The idea with the damping is that one can specify it globally (via `solver_settings` aka `ProcessInfo`), or locally (via the `Properties`). The locally defined ones take precedence

What one does with that is up to the user. I dont like the idea of excluding some stuff upfront, just because it might be expensive.

Keep in mind that Rayleigh damping is after all an estimation of the real structural damping.

AlejandroCornejo:
@atfard FYI

applications/StructuralMechanicsApplication/python_scripts/structural_mechanics_explicit_dynamic_solver.py
philbucher, line 1: the changes in this file should be reverted, it is correct as is.

With the fix in the element it should work as expected

loumalouomega: Huum, I do not agree. I would at least for the check that the values are different of zero, so I can set it on properties manually or in case I don't want to set it to skip it. I will do a minor refactor

loumalouomega: Should be Ok now

philbucher: this check is handled inside `HasRayleighDamping`, thats why the changes in the solver are not required
https://github.com/KratosMultiphysics/Kratos/blob/a52dd47025384ff0d6e21fa101ea18a69a0d708f/applications/StructuralMechanicsApplication/custom_utilities/structural_mechanics_element_utilities.cpp#L125-L130

loumalouomega: Okay

loumalouomega: I already checked:

~~~cpp
double GetRayleighAlpha(
    const Properties& rProperties,
    const ProcessInfo& rCurrentProcessInfo)
{
    // giving the locally defined setting (through Properties) priority
    // over the globally defined one (through ProcessInfo)
    if (rProperties.Has(RAYLEIGH_ALPHA)) {
        return rProperties[RAYLEIGH_ALPHA];
    } else if (rCurrentProcessInfo.Has(RAYLEIGH_ALPHA)) {
        return rCurrentProcessInfo[RAYLEIGH_ALPHA];
    }

    return 0.0;
}
~~~

Okay

loumalouomega: Done

philbucher (approved):
Perfect 👍🏻

PR #11130, opened 2023-05-15
[Core] Transition to remove template argument `MathUtils`

loumalouomega:
**📝 Description**

Related https://github.com/KratosMultiphysics/Kratos/pull/11112.

Transition to replace template types with `double` and improve type definitions. Removes three functions (`Abs`, `Min`, and `Max`) from the `MathUtils` class. Also moving `Solve` method to cpp.

**🆕 Changelog**

- [Remove template argument `MathUtils`](https://github.com/KratosMultiphysics/Kratos/commit/cf25333b9e8e1bf3aa34ee7bb805754452c7ac2c):
  - **Refactoring the `MathUtils` class**: The templated data type `TDataType` has been replaced with double in various functions. This suggests a design change to standardize the numerical type in these functions to `double` precision.
  - **Changing typedefs to using**: The old-style `typedef` statements have been replaced with the newer using syntax in defining types such as `MatrixType`, `VectorType`, `SizeType`, and `IndexType`.
- [Remove methods duplicate STL methods](https://github.com/KratosMultiphysics/Kratos/commit/2f63c5d19c82e9bb0ea430b0ce469bc60997c5a9): This commit removes three functions (`Abs`, `Min`, and `Max`) from the `MathUtils` class. The removal of these functions may indicate that they were either unnecessary or their functionalities are being replaced by standard library functions or methods.
- [Move Solve method to cpp](https://github.com/KratosMultiphysics/Kratos/commit/315230476f413f2c881c609f8b2b2b29671e75ab): The implementation of the `Solve` function, which was previously in the math_utils.h header file, has been moved to a new source file `math_utils.cpp`. This is generally considered good practice as it separates the interface (in the header file) from the implementation (in the source file), which can make the code easier to understand and maintain.
- [Revert template](https://github.com/KratosMultiphysics/Kratos/pull/11130/commits/06f812df316f2e63bdab94cee19144f6db46c70d): Default to `double`.


kratos/utilities/math_utils.h
```diff
@@ -56,7 +56,6 @@ namespace Kratos
  * @author Riccardo Rossi
  * @author Pooyan Dadvand
  */
-template<class TDataType>
 class MathUtils
```
philbucher:
```suggestion
template<class TDataType=double>
class MathUtils
```

this way it is backward compatible, and you can replace `MathUtils<double>` => `MathUtils` one-by-one

see https://github.com/KratosMultiphysics/Kratos/blob/4920be1d8b4d3d97f4a992ef57673d96eb2bf844/kratos/utilities/parallel_utilities.h#L406

avdg81: I agree. If we can split this up in smaller (more manageable) pieces, that would be very helpful for the review process. Also, I prefer to have logically distinct changes (e.g. replacing `MathUtils<>::Min` by `std::min`) in separate PRs. Having said that, those are my preferences, but perhaps it's not practical here since the work has already been done...?

loumalouomega: I will separate the other methods

loumalouomega: I think that if you put a default template argument you need to change for MathUtils<> anyway, so it will be the same effect

philbucher: > you need to change for MathUtils<> anyway

since C++17 you don't ;)

loumalouomega:
Lets see if it works

kratos/utilities/math_utils.h
loumalouomega: Done

huhaas:
I suggest squashing the commits in this PR. Otherwise, the massive change (of 450 files) is still in history including the full revert commit.

kratos/utilities/math_utils.h
```diff
@@ -211,7 +173,7 @@ class MathUtils
     template<class TMatrixType>
     static MatrixType CofactorMatrix(const TMatrixType& rMat)
     {
-        static_assert(std::is_same<TDataType, double>::value, "Bad value type.");
+        static_assert(std::is_same<double, double>::value, "Bad value type.");
```
avdg81: This assert no longer makes sense, since it will always evaluate to `true`.

kratos/utilities/math_utils.h
```diff
@@ -713,19 +666,19 @@ class MathUtils
...
-    static inline TDataType Dot(
+    static inline double Dot(
         const Vector& rFirstVector,
         const Vector& rSecondVector
         )
     {
         Vector::const_iterator i = rFirstVector.begin();
         Vector::const_iterator j = rSecondVector.begin();
-        TDataType temp = TDataType();
+        double temp = double();
```
avdg81: Since this is a built-in type, I'd rather be explicit about the desired initial value:
```suggestion
        double temp = 0.0;
```

kratos/utilities/math_utils.h
```diff
@@ -713,19 +666,19 @@ class MathUtils
...
         while(i != rFirstVector.end()) {
             temp += *i++ * *j++;
         }
         return temp;
-        //return std::inner_product(rFirstVector.begin(), rFirstVector.end(), rSecondVector.begin(), TDataType());
+        //return std::inner_product(rFirstVector.begin(), rFirstVector.end(), rSecondVector.begin(), double());
```
avdg81: Same here:
```suggestion
        //return std::inner_product(rFirstVector.begin(), rFirstVector.end(), rSecondVector.begin(), 0.0);
```

kratos/utilities/math_utils.h
```diff
@@ -747,10 +700,10 @@ class MathUtils
      * @param a Input vector
      * @return The resulting norm
      */
-    static inline TDataType Norm(const Vector& a)
+    static inline double Norm(const Vector& a)
     {
         Vector::const_iterator i = a.begin();
-        TDataType temp = TDataType();
+        double temp = double();
```
avdg81:
```suggestion
        double temp = 0.0;
```

avdg81 (commented):
Thanks for narrowing down the scope of this PR. It really helped me during the review process.

kratos/utilities/math_utils.h
loumalouomega:
```suggestion
```

avdg81 (commented):
The only suggestion I have left is to squash the commits in order to keep the history clean. Thanks.

avdg81:
@rubenzorrilla could you please have a look at this PR and approve it when you agree on the changes made? Thanks. Once these changes are available on `master` I will update the related PR #11112.

rubenzorrilla (approved):
Changes look good to me.

kratos/utilities/math_utils.h
rubenzorrilla: This is a game changer for this sort of things. Thanks for the hint @philbucher.

PR #11137, opened 2023-05-16
[Core] Expression Iterators

sunethwarna:
**📝 Description**
This PR implements the `ExpressionIterators` which are iterators for `Expressions`.  This allows iterating through lazy expressions as well as concrete flat data expressions.

**🆕 Changelog**
- Added `ExpressionIterators`


kratos/containers/container_expression/expressions/expression_iterator.h
```diff
@@ -0,0 +1,283 @@
...
+    ///@}
+    ///@name Life cycle
+    ///@{
+
+    ExpressionIterator(Expression::Pointer pExpression)
+        : mpExpression(pExpression),
+          mEntityIndex(0),
+          mEntityDataBeginIndex(0),
+          mComponentIndex(0),
+          mFlattenedShapeSize(pExpression->GetFlattenedShapeSize())
```
matekelemen: flattened shape size? Isn't that **always** 1 by definition?

I'd stick to something like `mFlatSize` or `mFlatEntitySize`.

kratos/containers/container_expression/expressions/expression_iterator.h
```diff
@@ -0,0 +1,283 @@
...
+
+    Expression::Pointer mpExpression;
+
+    IndexType mEntityIndex;
+
+    IndexType mEntityDataBeginIndex;
+
+    IndexType mComponentIndex;
+
+    IndexType mFlattenedShapeSize;
```
matekelemen: I can't say I'm comfortable with this representation. As far as I understand, this iterator represents a scalar component within a 1D array holding flattened entities (entities can be scalars, vectors or matrices). I assume the expression can be queried for the flattened size of a single entity. If so, `mEntityIndex`, `mEntityDataBeginIndex`, and `mComponentIndex` can all be deduced from an absolute index within the 1D array, so they shouldn't be separately stored and managed by the iterator.

The scary downside of keeping track of all these separate indices is that implementing `operator-`, `operator<` and the like, which are required for random access iterators, becomes incredibly tedious and prone to bugs.

I recommend storing only the pointer to the expression and the absolute index of the component within the 1D array. Comparisons and increments/decrements/steps could be performed directly on the absolute index. Other indices could then be computed only when necessary (eg.: ˙operator*`).

kratos/containers/container_expression/expressions/expression_iterator.h
```diff
@@ -0,0 +1,283 @@
...
+    ///@name Private member variables
+    ///@{
+
+    typename ExpressionType::Pointer mpExpression;
+
+    ///@}
+};
+
+template<>
+class ExpressionIterator<LiteralFlatExpression<int>> {
```
matekelemen: I think macros are OK to be used here to eliminate code duplication.

kratos/containers/container_expression/expressions/expression.h
```diff
@@ -37,14 +37,87 @@ class KRATOS_API(KRATOS_CORE) Expression {
...
+    ///@}
+    ///@name Public classes
+    ///@{
+
+    class ExpressionIterator {
+    public:
+        ///@name Type definitions
+        ///@{
+
+        KRATOS_CLASS_POINTER_DEFINITION(ExpressionIterator);
```
matekelemen, line 54: You're missing member aliases (such as `value_type`) that'd make this a valid iterator. Check out [`std::iterator_traits`](https://en.cppreference.com/w/cpp/iterator/iterator_traits) to see what the requirements of an iterator are.

kratos/containers/container_expression/expressions/expression.cpp
```diff
@@ -20,6 +20,72 @@
...
+}
+
+double Expression::ExpressionIterator::operator*() const
+{
+    return mpExpression->Evaluate(mEntityIndex, mEntityDataBeginIndex, mComponentIndex);
+}
+
+bool Expression::ExpressionIterator::operator==(const ExpressionIterator& rOther) const
+{
+    return (&*mpExpression == &*rOther.mpExpression) && (mEntityIndex == rOther.mEntityIndex) && (mComponentIndex == rOther.mComponentIndex);
```
matekelemen: Assuming `mpExpression` is an `std::shared_ptr`:


```diff
-   &*mpExpression == &*rOther.mpExpression
+   mpExpression.get() == rOther.mpExpression.get() && mpExpression.get() != nullptr
```

kratos/containers/container_expression/expressions/expression.cpp
```diff
@@ -20,6 +20,72 @@
...
+
+Expression::ExpressionIterator& Expression::ExpressionIterator::operator++()
+{
+    ++mComponentIndex;
+    if (mComponentIndex == mFlattenedShapeSize) {
+        mComponentIndex = 0;
+        ++mEntityIndex;
+        mEntityDataBeginIndex = mEntityIndex * mFlattenedShapeSize;
+    }
+    return *this;
```
matekelemen: I'm not blocking but I just have to note here for the record that I still believe an implementation in terms of an absolute index would be cleaner.

kratos/containers/container_expression/expressions/literal/literal_flat_expression.h
```diff
@@ -89,13 +87,15 @@ class KRATOS_API(KRATOS_CORE) LiteralFlatExpression : public Expression {
...
+    TRawDataType* begin() noexcept { return mData.begin(); }
 
-    inline TRawDataType* DataBegin() noexcept { return mData.DataBegin(); }
+    TRawDataType* end() noexcept { return mData.end(); }
 
-    inline TRawDataType const* DataBegin() const noexcept { return mData.DataBegin(); }
+    TRawDataType const* cbegin() const noexcept { return mData.cbegin(); }
 
-    inline TRawDataType const* DataEnd() const noexcept { return mData.DataEnd(); }
+    TRawDataType const* cend() const noexcept { return mData.cend(); }
```
matekelemen:
```suggestion
    TRawDataType* begin() noexcept { return mData.begin(); }

    TRawDataType* end() noexcept { return mData.end(); }

    const TRawDataType* begin() const noexcept { return mData.begin(); }

    const TRawDataType* end() const noexcept { return mData.end(); }

    TRawDataType const* cbegin() const noexcept { return mData.cbegin(); }

    TRawDataType const* cend() const noexcept { return mData.cend(); }
```

kratos/containers/container_expression/expressions/expression.h
```diff
@@ -82,11 +162,21 @@ class KRATOS_API(KRATOS_CORE) Expression {
     ///@name Input and output
     ///@{
 
+    ExpressionIterator cbegin() const;
+
+    ExpressionIterator cend() const;
```
matekelemen:
```suggestion
    ExpressionIterator begin() const;

    ExpressionIterator end() const;

    ExpressionIterator cbegin() const;

    ExpressionIterator cend() const;
```

Does the same thing with standard `begin` and `end`.

kratos/containers/container_expression/expressions/expression.h
```diff
@@ -37,14 +37,87 @@ class KRATOS_API(KRATOS_CORE) Expression {
     ///@name Type definitions
     ///@{
 
-    using Pointer = Kratos::intrusive_ptr<Expression>;
+    using Pointer = Kratos::intrusive_ptr<const Expression>;
 
     using IndexType = std::size_t;
```
matekelemen, line 43: If you want `Expression` to act as a proper container, I suggest adding
- member alias `value_type` (the "stored" type - same as `value_type` of `iterator`)
- member alias `size_type` (return type of `Expression::size()`)
- member alias `const_iterator` (and `iterator` if you allow mutable access)
- member function `size` (number of components in the container)

kratos/containers/container_expression/expressions/literal/literal_flat_expression.h
matekelemen: (I just added `const` versions of `begin` and `end`)

kratos/containers/container_expression/expressions/expression.h
sunethwarna: I was too lazy to do all those.... :/ :D... Now I have to fix them :P 

Btw, all requirements cannot be satisfied, because, `std::forward_iterator` requires a default constructor as well which does not make sense in this case... :/

kratos/containers/container_expression/expressions/expression.h
sunethwarna: My laziness strikes again....

It is a pain to have iterators :/

kratos/containers/container_expression/expressions/expression.h
matekelemen: You can add a default constructor that throws.

kratos/containers/container_expression/expressions/expression_iterator.h
sunethwarna: changed to `mItemComponentCount`

kratos/containers/container_expression/expressions/expression.cpp
sunethwarna: done :)

kratos/containers/container_expression/expressions/expression.h
sunethwarna: done :)

kratos/containers/container_expression/expressions/expression.h
sunethwarna: done :)

kratos/containers/container_expression/expressions/literal/literal_flat_expression.h
sunethwarna: done

kratos/containers/container_expression/expressions/expression.h
sunethwarna: done :)

kratos/containers/container_expression/expressions/expression_iterator.h
sunethwarna: removed since I moved the iterators to the respective classes.

matekelemen (approved):

PR #11138, opened 2023-05-16
[Core] Change default diagonal value in block B&S, unify with Parameters constructor

loumalouomega:
**📝 Description**

This PR updates the default value of the `mScalingDiagonal` member in the `ResidualBasedBlockBuilderAndSolver` class. The default setting is changed from `NO_SCALING` to `CONSIDER_MAX_DIAGONAL`. This change could impact the scaling considered for the Dirichlet degrees of freedom in the solving strategy. 

Related https://github.com/KratosMultiphysics/Kratos/pull/10706

**🆕 Changelog**

- [This commit updates the default value of the `mScalingDiagonal` member in the `ResidualBasedBlockBuilderAndSolver` class. The default setting is changed from `NO_SCALING` to `CONSIDER_MAX_DIAGONAL` ](https://github.com/KratosMultiphysics/Kratos/commit/cf32f3a73c979f03941541b7a5a7d4e75a1ec5ab)


rubenzorrilla (commented):
This unifies the scaling defaults in the block B&S. I also checked the tests reference values update and it makes sense to me.

However, I'd modify current test to use the parameters-based B&S so we can specify the no scaling option. If we do so, we circumvent the need of updating the test results.

loumalouomega:
> This unifies the scaling defaults in the block B&S. I also checked the tests reference values update and it makes sense to me.
> 
> However, I'd modify current test to use the parameters-based B&S so we can specify the no scaling option. If we do so, we circumvent the need of updating the test results.

Done

rubenzorrilla (commented):
Very good. Now I think we can safely merge this.

rubenzorrilla (approved):

Issue #10907, opened 2023-03-22
[MPM] Missing dependency pyevtk


jginternational:
**Description**
When running an MPM example, with vtk output, this error shows.
![image](https://user-images.githubusercontent.com/5918085/226880506-d0c19665-2d97-46b4-b096-b5ad696f77a9.png)

Seems to be missing this package -> pyevtk

**Scope**
I only found it in the MPM app, when VTK output is enabled

**To Reproduce**
Just execute any example for MPM

**Expected behavior**
MPM should have it as registered dependency

**Environment**
- Windows
- GID 16.1.4d > internet retrieve Kratos 9.3.2

@KratosMultiphysics/mpm

jginternational:
Probably should be added here
https://github.com/KratosMultiphysics/Kratos/blob/master/applications/ParticleMechanicsApplication/ParticleMechanicsApplication.json

mpentek:
Recently come across this myself.

@antonialarese @VeronikaSinger

philbucher:
afaik some special treatment is required to make MPM work with the core vtk output
I.e. some derived class (`MPMCtkOutput`) would need to be implemented with the required changes. I designed the `VtkOutput` with this in mind, similar to how the `EigenVtkOutput` in StructuralMechanics works

The current solution based on pyevtk was a rapid prototype iirc

ncrescenzio:
I am writing a `ParticleVtkOutput` class for the `ParticleMechanicsApplication` derived from the `VtkOutput` class in the core. The dependency on `pyevtk` will be removed soon.

ncrescenzio:
Issue fixed by #11140

PR #11140, opened 2023-05-16
[MPM] New `ParticleVtkOutput` process

ncrescenzio:
**📝 Description**
- Implementation of a `ParticleVtkOutput` class that can be found inside the new [`custom_io`](https://github.com/KratosMultiphysics/Kratos/tree/mpm/refactor-particle-vtk-output/applications/ParticleMechanicsApplication/custom_io) directory
- The class `ParticleVtkOutput` is derived from the `VtkOutput` class in the core.
- Dependency on the Python package `pyevtk` is removed (see issue #10907).
- [Added python tests](https://github.com/KratosMultiphysics/Kratos/blob/mpm/refactor-particle-vtk-output/applications/ParticleMechanicsApplication/tests/test_particle_vtk_output_process.py)
- Both the `ascii` and `binary` formats can now be used

---

These are the default parameters for the output process:
```bash
   "Parameters"   :  {
         "model_part_name"                             : "PLEASE_SPECIFY_MODEL_PART_NAME",
         "file_format"                                 : "binary",
         "output_precision"                            : 7,
         "output_control_type"                         : "step",
         "output_interval"                             : 1.0,
         "output_sub_model_parts"                      : false,
         "output_path"                                 : "MPM_VTK_Output",
         "custom_name_prefix"                          : "",
         "custom_name_postfix"                         : "",
         "save_output_files_in_folder"                 : true,
         "entity_type"                                 : "automatic",
         "write_ids"                                   : false,
         "element_flags"                               : [],
         "condition_flags"                             : [],
         "gauss_point_variables_in_elements"           : []
   }
```
`"entity_type"` can be set to `"elements"`, `"conditions"` or `"automatic"` and is used to decide whether to write the elements or conditions of the specified model part (see also the base class [`VtkOutput`](https://github.com/KratosMultiphysics/Kratos/blob/master/kratos/input_output/vtk_output.h)). 

The new [`particle_vtk_output_process.py`](https://github.com/KratosMultiphysics/Kratos/blob/mpm/refactor-particle-vtk-output/applications/ParticleMechanicsApplication/python_scripts/particle_vtk_output_process.py) replaces both the previous [`particle_vtk_output_process.py`](https://github.com/KratosMultiphysics/Kratos/blob/master/applications/ParticleMechanicsApplication/python_scripts/particle_vtk_output_process.py) and the [`particle_vtk_condition_output_process.py`](https://github.com/KratosMultiphysics/Kratos/blob/master/applications/ParticleMechanicsApplication/python_scripts/particle_vtk_condition_output_process.py).


applications/ParticleMechanicsApplication/python_scripts/particle_vtk_output_process.py
```diff
@@ -1,249 +1,104 @@
...
+            grid_settings = KratosMultiphysics.Parameters()
+            grid_settings.AddString("model_part_name","Background_Grid")
+            grid_settings.AddString("file_format",settings["file_format"].GetString())
+            grid_settings.AddDouble("output_precision",settings["output_precision"].GetDouble())
+            grid_settings.AddBool("output_sub_model_parts",settings["output_sub_model_parts"].GetBool())
+            grid_settings.AddString("output_path",settings["output_path"].GetString())
+            grid_settings.AddBool("save_output_files_in_folder",settings["save_output_files_in_folder"].GetBool())
+            background_grid = model["Background_Grid"]
+            output_filename = f"Background_Grid_{background_grid.GetCommunicator().MyPID()}"
+            KratosMultiphysics.VtkOutput(background_grid, grid_settings).PrintOutput(output_filename)
```
ncrescenzio: Notice that now, when the process is defined, the background grid is also written using the `VtkOutput` class in the core. I think it is useful since in this way we write the vtk data of the domain where the particles are moving without adding the `vtk_output_process` in the `ProjectParameters.json`

applications/ParticleMechanicsApplication/custom_io/particle_vtk_output.cpp
```diff
@@ -0,0 +1,430 @@
...
+{
+    const auto& r_local_mesh = rModelPart.GetCommunicator().LocalMesh();
+
+    if (mOutputEntities == ParticleVtkOutput::OutputEntities::ELEMENTS) {
+        const SizeType num_elements = rModelPart.GetCommunicator().GetDataCommunicator().SumAll(static_cast<int>(r_local_mesh.NumberOfElements()));
+        // Write points data
+        rFileStream << "POINTS " << num_elements << " float\n";
+        for (auto itr_element = r_local_mesh.ElementsBegin(); itr_element != r_local_mesh.ElementsEnd(); itr_element++) {
+            std::vector<array_1d<double, 3>> mp_coord = { ZeroVector(3) };
+            itr_element->CalculateOnIntegrationPoints(MP_COORD, mp_coord, rModelPart.GetProcessInfo());
```
philbucher, line 74: at first glance this seems to be the only change, the rest looks very similar to the Core Output

If there was a function `WritePoints`, could the rest of the code then be removed?

I would advise to avoid the duplication of code, as some of the writing things are quite delicate

ncrescenzio: I removed part of the duplicated code. I cannot remove methods `WritePropertiesIdsToFile` and `WriteIdsToFile` because they are defined as private in the `VtkOutput` class.

What do you mean when you say that a function `WritePoints` could be sufficient? Should I write everything (`POINTS`, `CELLS`, `CELL_TYPES` and `POINT_DATA`) in this function?

philbucher: Let me reformulate my question, what is different for the mpm vtk output?

I.e. what is the minimal changes necessary to make it work?

ncrescenzio: The function `WriteNodesToFile` must be changed because the nodes now are the coordinates of the material points (this basically is the line of code you highlighted above).

Moreover, I have changed the function `WriteConditionsAndElementsToFile` because in the `VtkOutput` implemented in the core if a model part has both elements and conditions then only the elements are written, while in the implementation I proposed here the entity that is written (elements or conditions) is the one specified in the input parameters. As far as I have understood, it is not possible to write conditions when both elements and conditions are defined in a model part and this should be why `particle_vtk_condition_output_process.py` was implemented before. For the same reason it is also required to change the function `WriteConditionResultsToFile`.

Concerning the functions `WritePropertiesIdsToFile` and `WriteIdsToFile`, as I said in the previous comment, they are copied only because in the base class they are defined as private.

philbucher: Ok understood

Have a look at #11156, I think with this all you need to do is to override the `WriteNodesToFile` method right?

ncrescenzio: I think so, but now there is a new problem.

I have tried to merge your branch `core/vtk-enhancements` to this one locally and to override only the method `WriteNodesToFile`. Kratos is compiling and the process is working without returning errors but data is not written in the VTK files.

I guess this is because the mpm model part that is used to initialize the `ParticleVtkOutput` has no elements and conditions in that moment (they are created in the `Initialize` stage of the `PythonSolver`, so after all the processes have already been defined). Then, even if in the parameters we specify `"entity_type" : "element"`, the value is of `mEntityType` is set to `EntityType::NONE` because the number of elements is zero:

https://github.com/KratosMultiphysics/Kratos/blob/63814884db4b81537ea46c814cec736c2a0a1d4d/kratos/input_output/vtk_output.cpp#L142-L144

I do not know if it is better to adjust your implementation in `core/vtk-enhancements` or if it is better to generate material point elements and conditions in an earlier stage of the simulation.

philbucher: Ok I see

I did not consider this case, I will update the implementation accordingly

sunethwarna: This would be a nice addition for me too... :D... I was also thinking to do the same... :D

philbucher: now it should work :)

ncrescenzio: Ok, the class is writing data on the VTK files but there are other problems 😅

I think they are both related to the fact that the geometry type of mpm elements and conditions is `Kratos_Quadrature_Point_Geometry`.

When the method `WriteConnectivity` is called, the nodes of the background geometry are written instead of just a number referring to the points in the `POINTS` section

Moreover, when the method `WriteCellType` is called, the following error is shown

https://github.com/KratosMultiphysics/Kratos/blob/0affec97cbeabff2f75f07fddda0b66ff1767ed8/kratos/input_output/vtk_output.cpp#L425-L427

because `Kratos_Quadrature_Point_Geometry` is not included here

https://github.com/KratosMultiphysics/Kratos/blob/0affec97cbeabff2f75f07fddda0b66ff1767ed8/kratos/input_output/vtk_output.cpp#L64-L66

Can I also override these two methods?

philbucher: Do you really need to override? Would adding this type to the map be enough?

ncrescenzio: For `WriteCellType` yes, it should be enough to add `Kratos_Quadrature_Point_Geometry` to `geo_type_vtk_cell_type_map` (didn't know if for you it's ok to add this in the core).

But this does not solve the problem related to `WriteConnectivity`.

philbucher: looking at the code I think you need to implement your own `WriteConnectivity` (the one in the baseclass is templated and can thus not be overridden)

I guess you can do this in this function (`WriteConditionsAndElementsToFile`), or have a `WriteParticleConnectivity`, which is called from here

ncrescenzio: Summarizing, now I'm using the following methods:
* `WriteNodesToFile`: it writes the coordinates of the material points elements/conditions (virtual method in the core since #11156)
* `WriteCellType`: it is necessary since `Kratos_Quadrature_Point_Geometry` is not contained in `vtk_definitions.cpp`. I can add it to the map `VtkDefinitions::KratosVtkGeometryTypes` and remove this method but in my opinion in the map there should be only geometries that are compatible with the vtk class in the core, while `Kratos_Quadrature_Point_Geometry` requires other methods for writing the coordinates and the connectivity
* `WriteConnectivity`: it is necessary since the base method writes the IDs of the nodes of the backgroud mesh instead of the ones of the elements
* `WriteConditionsAndElementsToFile`: used to call the new methods `WriteConnectivity` and `WriteCellType` (requires to be defined as `virtual` in the core with a separate PR)
* `WriteNodalResultsToFile`: empty method that overrides the one in the core (requires to be defined as `virtual` in the core with a separate PR). This is necessary because, when the method in the base class is called, even if we don't want to write nodal results (like in mpm) it still writes `POINT_DATA number_of_nodes` and an error occur when trying to open the file with paraview

applications/ParticleMechanicsApplication/python_scripts/particle_vtk_condition_output_process.py
philbucher, line 1: just a thought, maybe you can add a deprecation warning first? (and internally construct the new process?)

applications/ParticleMechanicsApplication/tests/test_particle_vtk_output_process.py
```diff
@@ -0,0 +1,211 @@
+import KratosMultiphysics
+import KratosMultiphysics.ParticleMechanicsApplication as KratosParticle
+import KratosMultiphysics.KratosUnittest as KratosUnittest
+
+import KratosMultiphysics.kratos_utilities as kratos_utils
+import KratosMultiphysics.ParticleMechanicsApplication.particle_vtk_output_process as particle_vtk_output_process
+from KratosMultiphysics.compare_two_files_check_process import CompareTwoFilesCheckProcess
+
+import os
```
philbucher, line 9:
```suggestion
```

unused

applications/ParticleMechanicsApplication/custom_io/particle_vtk_output.cpp
philbucher: sounds good to me

Please feel free to make the methods `virtual` if you need them to be, and tag me in the PR

applications/ParticleMechanicsApplication/custom_io/particle_vtk_output.h
```diff
@@ -0,0 +1,146 @@
...
+     * @param Parameters Parameters including settings for the output
+     */
+    explicit ParticleVtkOutput(
+        ModelPart& rModelPart,
+        Parameters ThisParameters = Parameters(R"({})" )
+        );
+
+    /// Destructor.
+    virtual ~ParticleVtkOutput() = default;
+
```
philbucher:
```suggestion
```

you can let the compiler do its job here :)

applications/ParticleMechanicsApplication/custom_io/particle_vtk_output.cpp
```diff
@@ -0,0 +1,159 @@
...
+        "save_output_files_in_folder"                 : true,
+        "entity_type"                                 : "automatic",
+        "write_ids"                                   : false,
+        "element_flags"                               : [],
+        "condition_flags"                             : [],
+        "gauss_point_variables_in_elements"           : []
+    })" );
+
+    return default_parameters;
+}
```
philbucher: do you need this method? Seems unused

philbucher (dismissed):
LGTM

applications/ParticleMechanicsApplication/custom_io/particle_vtk_output.cpp
ncrescenzio: I'm using these parameters for the validation of the input settings in the constructor of the Python process. 
See the code here:

https://github.com/KratosMultiphysics/Kratos/blob/5cb788e0361ea7125df5c4a2a653dfce5afc679b/applications/ParticleMechanicsApplication/python_scripts/particle_vtk_output_process.py#L22-L23

I know that the validation of the input settings is already performed in the constructor of the C++ process, but if I am not wrong it always uses the default parameters of the base class `VtKOutput`. Now the problem is that the parameters of the base class `VtkOutput` contains some fields, such as `element_data_value_variables`, `condition_data_value_variables`, `nodal_data_value_variables`, `nodal_flags` and `nodal_solution_step_data_variables`, that must not be used when using the class `ParticleVtkOutput` because some errors arise (they basically have no meaning in this context). I just want to avoid an external user to copy the parameters from the base process and add some variables where they shouldn't be added.

I don't like too much the solution that I am proposing, but I failed to come up with a better idea. 
Do you have any suggestion?

applications/ParticleMechanicsApplication/custom_io/particle_vtk_output.cpp
ncrescenzio: Done in #11482

matekelemen: I won't suggest rewriting the process, but I think the logic of this output deviates from `VtkOutput` so much that it doesn't make sense to derive from it. Default params are restricted instead of expanded, some functions (such as `WriteNodalResultsToFile`) are overridden to do nothing, etc. Furthermore, there isn't anything to be gained from dynamic polymorphism here.

My 2 cents: it would've probably been better to compose `VtkOutput` instead of inheriting from it.

ncrescenzio: Sorry, but I am not sure to have understood how you wold like the process to be implemented. 
Could you please provide me an example of something similar implemented in kratos?

matekelemen: It's just a suggestion for the future, you don't need to change this now.

```diff
- class ParticleVtkOutput : public VtkOutput
+ class ParticleVtkOutput
{
    /*...*/
+ private: 
+     VtkOutput mVtkOutput;
};
```

applications/ParticleMechanicsApplication/python_scripts/particle_vtk_condition_output_process.py
ncrescenzio: Done!

applications/ParticleMechanicsApplication/custom_io/particle_vtk_output.cpp
ncrescenzio: Ok, thanks for your comment and suggestion, I will consider it in the future!

philbucher:
hm the setup seems correct, I cannot tell why it fails on windows :/
I know that we have had some issues in the past with vtk in binary in windows.

If you dont need it, then I suggest to skip the binary tests in win (`if os.name == "nt"`) for now
If someone needs it and it is not working, then lets take that opportunity to debug

applications/ParticleMechanicsApplication/custom_io/particle_vtk_output.cpp
philbucher: @matekelemen makes full sense to me, back then the design with derivation seemed the more useful, but perhaps it is not the case here

Regardless, I would not focus too much on vtk, but rather on vtu in the future

applications/ParticleMechanicsApplication/tests/test_particle_vtk_output_process.py
```diff
@@ -0,0 +1,215 @@
...
+
+class TestParticleVtkOutputProcess(KratosUnittest.TestCase):
+    def test_ascii_particle_conditions_vtk_output_2D(self):
+        ExecuteBasicParticleVTKOutputProcessCheck("ascii", "condition")
+
+    def test_ascii_particle_elements_vtk_output_2D(self):
+        ExecuteBasicParticleVTKOutputProcessCheck("ascii", "element")
+
+    def test_binary_particle_conditions_vtk_output_2D(self):
+        ExecuteBasicParticleVTKOutputProcessCheck("binary", "condition")
```
philbucher:
```suggestion
    def test_binary_particle_conditions_vtk_output_2D(self):
        if os.system == "nt:
            self.skipTest("Binary output currently not working in windows.")
        ExecuteBasicParticleVTKOutputProcessCheck("binary", "condition")
```

this is the proper way of skipping tests, this way it gets reported correctly by the test runner

ncrescenzio: Ok, thanks for the suggestion

philbucher (dismissed):

lauramoremar (dismissed):

VeronikaSinger (approved):
Thanks @ncrescenzio

PR #11141, opened 2023-05-16
[HDF5] Replace Dimension in XDMF Generation

matekelemen:
## Description
`Dimension` got replaced with `LocalSpaceDimension` and `WorkingSpaceDimension` and as a result, XDMF generation has been crashing because it didn't get updated (there aren't any tests for it in the CI). I've replaced `Dimension` with `WorkingSpaceDimension` in `XdmfConnectivitiesWriterProcess` though I'm not a 100% sure it's the right way to go.

## ToDo
Write XDMF generation tests and include them in the CI (requires `h5py`).


sunethwarna:
@matekelemen Are you sure you pushed all the commits? I see there are several palces needs replacing....

sunethwarna (approved):
I trust you that you have created an xdmf file with this update without any troubles :).

PR #11147, opened 2023-05-17
[GeoMechanicsApplication] Removed apply_double_table_process, because it only contained a copy …

WPK4FEM:
Removed apply_double_table_process, because it only contained a copy of apply_component_table_process

**📝 Description**
Removed code duplication and extra class.

**🆕 Changelog**
Removed apply_double_table_process.hpp and all the references to it. Replaced its use in python_scripts with the equivalent in apply_component_table_process.hpp


avdg81 (dismissed):
Nice cleanup. Thanks!

avdg81 (approved):

PR #11151, opened 2023-05-18
[Structural][Hotfix] Integration order base solid element correction

rubenzorrilla:
**📝 Description**
This fixes the usage of non-default integration orders in the base solid element. As it is right now, the constitutive law vector has a wrong size, leading to a segmentation fault when evaluating the material response.


rubenzorrilla:
BTW, a few lines of code above the fix in here, there is the check to use the integration rules from the geometry (i.e. `UseGeometryIntegrationMethod`), which is always `true`. I might be wrong, but as far as I see, we can safely remove it (I'd do so in a different PR).

AlejandroCornejo (commented):
Makes sense

AlejandroCornejo:
@loumalouomega

loumalouomega (approved):
Ups

PR #11152, opened 2023-05-18
Exporting accessor to python extension

roigcarlo:
**📝 Description**
Extension for #11069.

Since I touched more things than I intended to, I prefer to review it as a separate PR.

**🆕 Changelog**
- Accessor python interface through opaque std::unique_ptr
- Simplified wrapper code
- Added exceptions
- Added tests for corner cases and example of use.

### DIFF
--- a/kratos/python/add_accessors_to_python.cpp
+++ b/kratos/python/add_accessors_to_python.cpp
@@ -22,28 +22,21 @@
 #include "includes/accessor.h"
 #include "includes/properties.h"
 
+using AcccessorBindType = std::unique_ptr<Kratos::Accessor>;
+
+PYBIND11_MAKE_OPAQUE(AcccessorBindType);
+
 namespace Kratos::Python
 {
 
-template<class TDataType, class TAccessorType, class... TAccessorArgs>
-auto& InternalCreateAndSetAccessorToProperties(
-    Properties& rProperties,
-    const Variable<TDataType>& rVariable,
-    TAccessorArgs... rAccessorArgs)
-{
-    auto p_new = Kratos::make_unique<TAccessorType>(std::forward<TAccessorArgs>(rAccessorArgs)...);
-    rProperties.SetAccessor(rVariable, std::move(p_new));
-    return rProperties.GetAccessor(rVariable);
-}
+namespace py = pybind11;
 
-void AddAccessorsToPython(pybind11::module& m)
+void AddAccessorsToPython(py::module& m)
 {
-    namespace py = pybind11;
-
-    py::class_<Accessor, Accessor::UniquePointer>(m, "Accessor")
-        .def("Create", [](Properties& rProp, Variable<double>& rVariable) -> auto& 
-        { return InternalCreateAndSetAccessorToProperties<double, Accessor>(rProp,rVariable); },
-        py::return_value_policy::reference_internal)
+    py::class_<AcccessorBindType>(m, "Accessor")
+        .def_static("Create", []() { 
+            return std::unique_ptr<Accessor>(new Accessor());
+        })
         ;
 }
 
--- a/kratos/python/add_accessors_to_python.cpp
+++ b/kratos/python/add_accessors_to_python.cpp
@@ -22,9 +22,9 @@
 #include "includes/accessor.h"
 #include "includes/properties.h"
 
-using AcccessorBindType = std::unique_ptr<Kratos::Accessor>;
+using AccessorBindType = std::unique_ptr<Kratos::Accessor>;
 
-PYBIND11_MAKE_OPAQUE(AcccessorBindType);
+PYBIND11_MAKE_OPAQUE(AccessorBindType);
 
 namespace Kratos::Python
 {
@@ -33,7 +33,7 @@ namespace py = pybind11;
 
 void AddAccessorsToPython(py::module& m)
 {
-    py::class_<AcccessorBindType>(m, "Accessor")
+    py::class_<AccessorBindType>(m, "Accessor")
         .def_static("Create", []() { 
             return std::make_unique<Accessor>();
         })
--- a/kratos/python/add_accessors_to_python.cpp
+++ b/kratos/python/add_accessors_to_python.cpp
@@ -35,7 +35,7 @@ void AddAccessorsToPython(py::module& m)
 {
     py::class_<AcccessorBindType>(m, "Accessor")
         .def_static("Create", []() { 
-            return std::unique_ptr<Accessor>(new Accessor());
+            return std::make_unique<Accessor>();
         })
         ;
 }
--- a/kratos/python/add_properties_to_python.cpp
+++ b/kratos/python/add_properties_to_python.cpp
@@ -161,18 +161,17 @@ void AddInterfaceToAccessorFold(pybind11::class_<Properties, Properties::Pointer
                 
             return accessor; 
         }, py::return_value_policy::reference_internal)
-    .def("SetAccessor", [](Properties &rProperties, Variable<TVariableType> &rVariable, std::unique_ptr<Accessor> & rpAccessor) {
-            if (rpAccessor == nullptr)
+    .def("SetAccessor", [](Properties &rProperties, Variable<TVariableType> &rVariable, std::unique_ptr<Accessor>& rAccessor) {
+            if (rAccessor == nullptr)
                 KRATOS_ERROR << "Trying to set a consumed or invalid Accessor. Accessors are unique. Please create a different one." << std::endl;
 
-            rProperties.SetAccessor(rVariable, std::move(rpAccessor));
+            rProperties.SetAccessor(rVariable, std::move(rAccessor));
             return &rProperties.pGetAccessor(rVariable);
         }, py::return_value_policy::reference_internal)
     .def("HasAccessor", [](Properties &rProperties, Variable<TVariableType> &rVariable) { 
             return rProperties.HasAccessor(rVariable);
         })
-    ;
-        
+    ; 
 }
 
 template<typename... TArgs>
--- a/kratos/python/add_properties_to_python.cpp
+++ b/kratos/python/add_properties_to_python.cpp
@@ -14,7 +14,6 @@
 // System includes
 
 // External includes
-#include <pybind11/pybind11.h>
 
 // Project includes
 #include "includes/define_python.h"
@@ -23,13 +22,16 @@
 #include "includes/condition.h"
 #include "includes/properties.h"
 #include "includes/constitutive_law.h"
-#include "python/add_mesh_to_python.h"
+#include "python/add_properties_to_python.h"
 #include "python/containers_interface.h"
 
-namespace Kratos
+using AcccessorBindType = std::unique_ptr<Kratos::Accessor>;
+
+PYBIND11_MAKE_OPAQUE(AcccessorBindType);
+
+namespace Kratos::Python
 {
-namespace Python
-{
+
 namespace py = pybind11;
 
 using NodeType = Node<3>;
@@ -147,10 +149,40 @@ bool HasTableHelperFunction1( TContainerType& rContainer,
     return rContainer.HasTable(XVar, YVar);
 }
 
+template<typename TVariableType>
+void AddInterfaceToAccessorFold(pybind11::class_<Properties, Properties::Pointer, Properties::BaseType>& pyProperties) {
+
+    pyProperties
+    .def("GetAccessor", [](Properties &rProperties, Variable<TVariableType> &rVariable) { 
+            auto accessor = &rProperties.pGetAccessor(rVariable);
+            
+            if (*accessor == nullptr)
+                KRATOS_ERROR << "Trying to get a consumed or invalid Accessor." << std::endl;
+                
+            return accessor; 
+        }, py::return_value_policy::reference_internal)
+    .def("SetAccessor", [](Properties &rProperties, Variable<TVariableType> &rVariable, std::unique_ptr<Accessor> & rpAccessor) {
+            if (rpAccessor == nullptr)
+                KRATOS_ERROR << "Trying to set a consumed or invalid Accessor. Accessors are unique. Please create a different one." << std::endl;
+
+            rProperties.SetAccessor(rVariable, std::move(rpAccessor));
+            return &rProperties.pGetAccessor(rVariable);
+        }, py::return_value_policy::reference_internal)
+    .def("HasAccessor", [](Properties &rProperties, Variable<TVariableType> &rVariable) { 
+            return rProperties.HasAccessor(rVariable);
+        })
+    ;
+        
+}
+
+template<typename... TArgs>
+void AddInterfaceToAccessor(pybind11::class_<Properties, Properties::Pointer, Properties::BaseType>& pyProperties) {
+    (AddInterfaceToAccessorFold<TArgs>(pyProperties), ...);
+}
 
 void  AddPropertiesToPython(pybind11::module& m)
 {
-    py::class_<Properties, Properties::Pointer, Properties::BaseType>(m, "Properties")
+    auto properties_module = py::class_<Properties, Properties::Pointer, Properties::BaseType>(m, "Properties")
         .def(py::init<Kratos::Properties::IndexType>())
         .def(py::init<const Properties &>())
         .def("__setitem__", SetValueHelperFunction1<Properties, Variable<array_1d<double, 6>>>)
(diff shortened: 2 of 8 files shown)


loumalouomega (approved):
Okidoki

AlejandroCornejo (approved):
Looks perfect!!

kratos/python/add_accessors_to_python.cpp
```diff
@@ -22,28 +22,21 @@
 #include "includes/accessor.h"
 #include "includes/properties.h"
 
+using AcccessorBindType = std::unique_ptr<Kratos::Accessor>;
+
+PYBIND11_MAKE_OPAQUE(AcccessorBindType);
+
```
rubenzorrilla: Shouldn't we define this within the `Kratos::Python` namespace for the case in which we export a custom accessor from a property?

kratos/python/add_properties_to_python.h
```diff
@@ -0,0 +1,29 @@
...
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
+//
+//  Main authors:    Pooyan Dadvand
+//                   Riccardo Rossi
```
rubenzorrilla:
```suggestion
//  Main authors:    Carlos Roig
```

kratos/python/add_properties_to_python.cpp
```diff
@@ -23,13 +22,16 @@
...
-#include "python/add_mesh_to_python.h"
+#include "python/add_properties_to_python.h"
 #include "python/containers_interface.h"
 
-namespace Kratos
-{
-namespace Python
+using AcccessorBindType = std::unique_ptr<Kratos::Accessor>;
+
+PYBIND11_MAKE_OPAQUE(AcccessorBindType);
```
rubenzorrilla: Shouldn't we define this once within the `Kratos::Python` namespace? As it is right now it is both here and in the `add_accessors_to_python.cpp`.

kratos/python/add_properties_to_python.cpp
```diff
@@ -147,10 +149,39 @@ bool HasTableHelperFunction1( TContainerType& rContainer,
...
+    pyProperties
+    .def("GetAccessor", [](Properties &rProperties, Variable<TVariableType> &rVariable) { 
+            auto accessor = &rProperties.pGetAccessor(rVariable);
+            
+            if (*accessor == nullptr)
+                KRATOS_ERROR << "Trying to get a consumed or invalid Accessor." << std::endl;
+                
+            return accessor; 
+        }, py::return_value_policy::reference_internal)
+    .def("SetAccessor", [](Properties &rProperties, Variable<TVariableType> &rVariable, std::unique_ptr<Accessor>& rAccessor) {
```
rubenzorrilla: Long story short and for the record, we can pass the `unique_ptr` thanks to the `PYBIND11_MAKE_OPAQUE`  that circumvents the type conversion.

kratos/python/add_properties_to_python.cpp
```diff
@@ -147,10 +149,39 @@ bool HasTableHelperFunction1( TContainerType& rContainer,
...
 
+template<typename TVariableType>
+void AddInterfaceToAccessorFold(pybind11::class_<Properties, Properties::Pointer, Properties::BaseType>& pyProperties) {
+
+    pyProperties
+    .def("GetAccessor", [](Properties &rProperties, Variable<TVariableType> &rVariable) { 
+            auto accessor = &rProperties.pGetAccessor(rVariable);
+            
+            if (*accessor == nullptr)
+                KRATOS_ERROR << "Trying to get a consumed or invalid Accessor." << std::endl;
```
rubenzorrilla:
```suggestion
            KRATOS_ERROR_IF(*accessor == nullptr) << "Trying to get a consumed or invalid Accessor." << std::endl;
```

kratos/python/add_properties_to_python.cpp
rubenzorrilla:
```suggestion
    .def("SetAccessor", [](Properties &rProperties, Variable<TVariableType> &rVariable, std::unique_ptr<Accessor>& rpAccessor) {
```

kratos/python/add_properties_to_python.cpp
```diff
@@ -147,10 +149,39 @@ bool HasTableHelperFunction1( TContainerType& rContainer,
...
+                
+            return accessor; 
+        }, py::return_value_policy::reference_internal)
+    .def("SetAccessor", [](Properties &rProperties, Variable<TVariableType> &rVariable, std::unique_ptr<Accessor>& rAccessor) {
+            if (rAccessor == nullptr)
+                KRATOS_ERROR << "Trying to set a consumed or invalid Accessor. Accessors are unique. Please create a different one." << std::endl;
+
+            rProperties.SetAccessor(rVariable, std::move(rAccessor));
```
rubenzorrilla:
```suggestion
            rProperties.SetAccessor(rVariable, std::move(rpAccessor));
```

rubenzorrilla: As we're moving in here, wouldn't this end with invalid variables in the Python side?

kratos/python/add_accessors_to_python.cpp
```diff
@@ -22,28 +22,21 @@
...
 {
-    namespace py = pybind11;
-
-    py::class_<Accessor, Accessor::UniquePointer>(m, "Accessor")
-        .def("Create", [](Properties& rProp, Variable<double>& rVariable) -> auto& 
-        { return InternalCreateAndSetAccessorToProperties<double, Accessor>(rProp,rVariable); },
-        py::return_value_policy::reference_internal)
+    py::class_<AcccessorBindType>(m, "Accessor")
+        .def_static("Create", []() { 
+            return std::unique_ptr<Accessor>(new Accessor());
```
rubenzorrilla:
```suggestion
            return Kratos::make_unique<Accessor>();
```

kratos/tests/test_properties.py
```diff
@@ -71,10 +71,49 @@ def test_accessor(self):
...
+
+        properties_1 = model_part.CreateNewProperties(1)
+        properties_2 = model_part.CreateNewProperties(2)
+
+        accessor_binding = KM.Accessor.Create()
+        
+        accessor_from_set = properties_1.SetAccessor(KM.TEMPERATURE, accessor_binding)
+
+        with self.assertRaisesRegex(Exception, "Trying to set a consumed or invalid Accessor. Accessors are unique. Please create a different one."):
+            accessor_from_set = properties_2.SetAccessor(KM.TEMPERATURE, accessor_binding)
```
rubenzorrilla: This boils down to the fact that the same accessor instance can't be used in more than one properties at the same time. I think this is OK, however we should confirm this together with the @KratosMultiphysics/technical-committee.

kratos/tests/test_properties.py
```diff
@@ -71,10 +71,49 @@ def test_accessor(self):
...
+        properties_1 = model_part.CreateNewProperties(1)
+        properties_2 = model_part.CreateNewProperties(2)
+
+        accessor_binding = KM.Accessor.Create()
+        
+        accessor_from_set = properties_1.SetAccessor(KM.TEMPERATURE, accessor_binding)
+        accessor_from_set = properties_2.SetAccessor(KM.TEMPERATURE, accessor_from_set)
+
+        with self.assertRaisesRegex(Exception, "Trying to get a consumed or invalid Accessor."):
+            accessor_from_get = properties_1.GetAccessor(KM.TEMPERATURE)
```
rubenzorrilla: Taking this into account (in this particular case the error would be thrown at the time the `GetAccessor` is called), I'd remove the return from the `SetAccessor` to always enforce the retrieval through the get.

rubenzorrilla (changes requested):
Blocking to prevent an accidental merge in the prior branch that would complicate the @KratosMultiphysics/technical-committee discussion.

kratos/python/add_accessors_to_python.cpp
roigcarlo: No, it needs no to be defined outisde the Kratos namespace.

kratos/python/add_properties_to_python.cpp
roigcarlo: Yep but I don't know a clean way to do it. It needs to be defined in the python layer. If we put in either add_properties.h or add_accessor.h then we have to import the other in the file does not belong to. If we really want to have it defined one I would put it in the properties.

In any case as before, needs to be outside Kratos namespace

kratos/python/add_properties_to_python.cpp
roigcarlo: Yep. This is why I put all the new checks for nullptr.

Ideally I would create an interface in python that makes the movement transparent by replacing the binding with the one that SetAccessor returned, but its not possible because the binding is immutable. 

I can't manipulate the reference argument in C++ either.

If someone knows how to do it I will happily code it.

kratos/python/add_accessors_to_python.cpp
roigcarlo: I've changed to `std::make_unique`. we should get rid of `Kratos::make_unique`(and half of the `smart_pointers.h` )

AlejandroCornejo:
what is the state of this?

kratos/python/add_properties_to_python.cpp
```diff
@@ -147,10 +149,37 @@ bool HasTableHelperFunction1( TContainerType& rContainer,
...
+            
+            KRATOS_ERROR_IF(*accessor == nullptr) << "Trying to get a consumed or invalid Accessor." << std::endl;
+                
+            return accessor; 
+        }, py::return_value_policy::reference_internal)
+    .def("SetAccessor", [](Properties &rProperties, Variable<TVariableType> &rVariable, std::unique_ptr<Accessor>& rpAccessor) {
+            if (rpAccessor == nullptr)
+                KRATOS_ERROR << "Trying to set a consumed or invalid Accessor. Accessors are unique. Please create a different one." << std::endl;
+
+            rProperties.SetAccessor(rVariable, std::move(rpAccessor));
```
RiccardoRossi, line 167: i think you should set rpAccessor to a nullptr explicilty, otherwise what is left on rpAccessor is undefined

kratos/python/add_properties_to_python.cpp
```diff
@@ -254,70 +283,22 @@ void  AddPropertiesToPython(pybind11::module& m)
...
-             { return rProperties.HasAccessor(rVariable); })
-        .def("HasAccessor", [](Properties &rProperties, const Variable<array_1d<double, 9>> &rVariable)
-             { return rProperties.HasAccessor(rVariable); })
-        .def("HasAccessor", [](Properties &rProperties, const Variable<array_1d<double, 4>> &rVariable)
-             { return rProperties.HasAccessor(rVariable); })
-        .def("HasAccessor", [](Properties &rProperties, const Variable<std::string> &rVariable)
-             { return rProperties.HasAccessor(rVariable); });
+        ;
+
+        AddInterfaceToAccessor<
```
RiccardoRossi, line 288: won't it take forever to compile a function with so many template arguments? maybe alternatively we could use a few macros here.

kratos/python/add_properties_to_python.cpp
```diff
@@ -147,10 +149,37 @@ bool HasTableHelperFunction1( TContainerType& rContainer,
...
+    pyProperties
+    .def("GetAccessor", [](Properties &rProperties, Variable<TVariableType> &rVariable) { 
+            auto accessor = &rProperties.pGetAccessor(rVariable);
+            
+            KRATOS_ERROR_IF(*accessor == nullptr) << "Trying to get a consumed or invalid Accessor." << std::endl;
+                
+            return accessor; 
+        }, py::return_value_policy::reference_internal)
+    .def("SetAccessor", [](Properties &rProperties, Variable<TVariableType> &rVariable, std::unique_ptr<Accessor>& rpAccessor) {
+            if (rpAccessor == nullptr)
```
RiccardoRossi: cosmetic comment, but there is an officil way of doing this:

https://en.cppreference.com/w/cpp/memory/unique_ptr/operator_bool

kratos/python/add_properties_to_python.cpp
```diff
@@ -147,10 +149,37 @@ bool HasTableHelperFunction1( TContainerType& rContainer,
...
+            rProperties.SetAccessor(rVariable, std::move(rpAccessor));
+        }, py::return_value_policy::reference_internal)
+    .def("HasAccessor", [](Properties &rProperties, Variable<TVariableType> &rVariable) { 
+            return rProperties.HasAccessor(rVariable);
+        })
+    ; 
+}
+
+template<typename... TArgs>
+void AddInterfaceToAccessor(pybind11::class_<Properties, Properties::Pointer, Properties::BaseType>& pyProperties) {
```
RiccardoRossi, line 176: why do we need variadic arguments here? 

cannot we simply call this function as many times as types we have, but each time just with one template parameter?

kratos/python/add_properties_to_python.cpp
RiccardoRossi: btw, you have too many cs in the name

kratos/python/add_properties_to_python.cpp
roigcarlo: I cannot change rpAccessor because its a refference pointer to the python wrapper. I can only checks that points to nullptr because of either the move ir or in another place.

kratos/python/add_properties_to_python.cpp
roigcarlo: No, this is not a normal template. This is a folding expression. In general compilation performance of folding expression is waaaay better than recursive templates (which would be another way to do this, and performance the reason we werent using it)

kratos/python/add_properties_to_python.cpp
roigcarlo: Because of the folding, it forwards the types you are using to instantiate to the rest of the function.

kratos/python/add_properties_to_python.cpp
RiccardoRossi: well, then you have no guarantees about the status in which it will be left after moving it (as i recall it is undefined behaviour) ... this may be a problem since you are checking for multiple usage

kratos/python/add_properties_to_python.cpp
RiccardoRossi: but why cannot you write

 ~~~cpp
AddInterfaceToAccessorFold<double>(...)
AddInterfaceToAccessorFold<bool>(...)
AddInterfaceToAccessorFold<int>(...)
AddInterfaceToAccessorFold<unsifned int>(...)
...
 ~~~

which would be a much simpler code?

roigcarlo: Well, yes I could, but then I have to write one line with all the args per type I want to expose. Imo the other way is better (note that we could have a macro with a type bundle (ie KRATOS_PYTHON_SCALAR_VARS, KRATOS_PYTHON_VECTOR_VARS) and  then it will look as:

```c++
#define KRATOS_PYTHON_SCALAR_VARS bool, int, double, string
#define KRATOS_PYTHON_VECTOR_VARS matrix, vector, array_1d<double, 3>, array_1d<double, 4>, array_1d<double, 6>, .... 
```

```c++
AddInterfaceToAccessor<
    KRATOS_PYTHON_SCALAR_VARS,
    KRATOS_PYTHON_VECTOR_VARS 
>(properties_module);
```
 
Also if we leave it as I did, then when templated lambdas become usable we will be able to do something like:

```c++
AddInterfaceToPyhonGeneric<T1,T2...TN>(args, []<template TX>(arg) {
     return myclass<TX>(args)
})
```

So we don't need the auxiliar function, the folder is hidden and all the interface is written once and looks like any other lambda export from pybind.

kratos/python/add_properties_to_python.cpp
roigcarlo: The I don't know how yo check the correctness of this :S

kratos/tests/test_properties.py
```diff
@@ -68,13 +68,51 @@ def test_has_table(self):
...
+            accessor_from_get = properties.GetAccessor(KM.TEMPERATURE)
+
+    def test_accessor_get_consumed(self):
+        current_model = KM.Model()
+        model_part= current_model.CreateModelPart("Main")
+
+        properties_1 = model_part.CreateNewProperties(1)
+        properties_2 = model_part.CreateNewProperties(2)
+
+        accessor_binding = KM.Accessor.Create()
```
pooyan-dadvand: I would not create the variable because after calling the `SetAccessor `will be invalidated. I would put the `KM.Accessor.Create()` directly in the `SetAccessor `argument list

pooyan-dadvand (commented):
@KratosMultiphysics/technical-committee approves this mechanism and waits for the reviewers to approve it.

rubenzorrilla (approved):
Ready to go

AlejandroCornejo:
FINALLY! :)

PR #11154, opened 2023-05-18
RomApp Bugfix HROM Modelpart Part Creation

Rbravo555:
**📝 Description**
This PR adresses a bug in the current implementation of the workflow when using either the RomManager or when runing the workflow by hand inputting the correct flags to the RomParameters.json

The bug was not detected before because it is only present if the basis format selected is Numpy and if the simulation contains CONDITIONS which are selected by the HROM selection algorithm.

In future PRs, we will add more comprehensive tests including tests for the RomManager. And we shall also update the tests of the HROM model part creation with a simulation that does select ELEMENTS + CONDITIONS. I will update the TODOs in the RomApp project accodingly.

**🆕 Changelog**
Changed the placement of 3 lines in a single file: hrom_training_utility.py


SADPR (approved):
Nice catch.

PR #11155, opened 2023-05-18
[Structural] Remove UseGeometryIntegrationMethod from base solid

rubenzorrilla:
**📝 Description**
Remove the always  true `UseGeometryIntegrationMethod` from the `BaseSolidElement`. On top of the fact that it's not being used, resulting in a more cumbersome implementation, from current code I think that the case of switching it to false would have led to some missbehaviours.


loumalouomega (approved):

loumalouomega:
@rubenzorrilla it is actually used in @KratosMultiphysics/altair , we are going to revert, and then we can adjust to something simpler

PR #11156, opened 2023-05-18
[Core][VTKOutput] make type of entity selectable

philbucher:
This PR makes the selection of entities written by the VtkOutput selectable, which was previously hardcoded
The behavior has not changed, still by default elements have a higher prio than conditions
Also the MPI-performance is slightly increased by removing some collective communication operations

This change is necessary for
- Preparation for #11140
- Preparation for Output of geometries

FYI @ncrescenzio

### DIFF
--- a/kratos/input_output/vtk_output.cpp
+++ b/kratos/input_output/vtk_output.cpp
@@ -7,7 +7,8 @@
 //  License:		 BSD License
 //					 Kratos default license: kratos/license.txt
 //
-//  Main authors:    Aditya Ghantasala, Philipp Bucher
+//  Main authors:    Aditya Ghantasala
+//                   Philipp Bucher (https://github.com/philbucher)
 //  Collaborator:    Vicente Mataix Ferrandiz
 //
 //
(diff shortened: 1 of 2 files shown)


sunethwarna:
@philbucher : I have a small question regarding this. Assume you want to write `Conditions` output and points output of a model part which has elements and conditions. In this case, we have to find out the nodes from `ModelPart::Nodes` which are relevant for conditions (and not consider the element nodes) ryt? Writing all the nodes and their data from `ModelPart::Nodes` will be wrong as far as I see. Am I missing something?

philbucher:
This is true, all the nodes would be written

This would require extra "filtering".

Note though that this is also the existing behavior, not sure if we need the filtering. Could be added I guess if it becomes relevant

philbucher:
Maybe we should also distinguish for nodes
all or from_entities

All is what we do now, but probably from_entities would be a better option for most cases

philbucher:
I was waiting for your response/approval.

I believe there are still some discussions open in your PR, but we can merge this one already

ncrescenzio:
Yes, but I think that the discussions in #11140 are not related to the topic of this PR. The selection of the entity type seems to work also there. So if other different changes are needed we can open a new PR, no?

philbucher:
yes that makes sense

can you then please approve this PR?

ncrescenzio (approved):

PR #11148, opened 2023-05-17
[Core][MPI] Enhanced testing framework for distributed test cases

loumalouomega:
**📝 Description**

This PR introduces several enhancements to the Kratos testing framework, specifically focusing on handling distributed test cases. 

Summary of changes:

1. In `distributed_test_case.cpp` and `distributed_test_case.h`, new methods `IsDistributedTest()` have been added, which return true for distributed test cases. 
2. In `test_case.cpp` and `test_case.h`, modifications have been made to accommodate the changes related to the distribution of test cases.
3. The `Tester` class has been updated to include a new method `SelectOnlyDistributedTestCases()`, which selects only the distributed test cases to run.
4. In `add_testing_to_python.cpp`, new static methods `RunAllDistributedTestCases()` and `ProfileAllDistributedTestCases()` have been added to run and profile all distributed test cases, respectively.
5. The `KratosUnittest.py` script has been modified to handle `mpi` levels properly in the `runTests()` function.
6. The `test_StructuralMechanicsApplication_mpi.py` file has been modified, but the exact changes are not clear from the provided patch.
7. The `run_cpp_mpi_tests.py` script now runs all distributed test cases by default, and the verbosity is set based on whether a test suite is provided as an argument.
8. Finally, changes have been made to the CI/CD scripts to run the MPI C++ tests and Python MPI tests using the new testing functionality introduced in this commit. The naming in the CI/CD scripts has also been updated to reflect the changes.
9. Changes from https://github.com/KratosMultiphysics/Kratos/pull/11161,  the changes specifically include:
	- Replacing the `sys` module with `argparse`, which is a more robust and user-friendly way of managing command-line arguments.
	- Instead of directly checking command-line arguments, the code now defines a `main` function that sets up a parser for command-line arguments. This parser is designed to handle an argument `--using-mpi` to force MPI mode, and another argument to specify a match string to run specific test cases.
	- The execution flow based on the command-line arguments has also been updated. Now, if a match string is provided, the verbosity is set to `TESTS_OUTPUTS` and only the matching test cases are run. Otherwise, all distributed test cases are run with the verbosity set to `TESTS_LIST`.

Related https://github.com/KratosMultiphysics/Kratos/pull/10706

**NOTE:** In my machine some C++ MPI tests are failing, we didn't notice because until now these tests were not run in the CI/CD. Let's see if the CI passes.

**🆕 Changelog**

[...]

PR #11157, opened 2023-05-19
[Core][MPI] Remove unnecessary argument so MPI C++ tests are run

loumalouomega:
**📝 Description**

Part of https://github.com/KratosMultiphysics/Kratos/pull/11148

The PR consists of changes in the continuous integration (CI) workflow file `ci.yml`. This file is typically responsible for setting up the environment and running tests automatically whenever changes are pushed to the repository. 

Until now `--using-mpi` was provoking that any test was run at all, due to the implementation of https://github.com/KratosMultiphysics/Kratos/blob/1f82bd8d96560ed962935f629fa61ed5e69bc126/kratos/python_scripts/testing/run_cpp_mpi_tests.py#L12

~~~sh
 |  /           |                  
 ' /   __| _` | __|  _ \   __|    
 . \  |   (   | |   (   |\__ \  
           Multi-Physics 9.3."1"--02ef2d1-Custom-x86_64
           Compiled for GNU/Linux and Python3.10 with GCC-12.1
Maximum number of threads: 1.
MPI world size:         2.
Process Id: 4523 
Ran 0 of 1544 test cases in 0.000481801s. OK
~~~

**🆕 Changelog**

- [Remove unnecessary argument](https://github.com/KratosMultiphysics/Kratos/commit/fe4298008639c83259071eceadb50999500e6e8a)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,7 +91,7 @@ jobs:
         source /opt/intel/oneapi/setvars.sh
         export PYTHONPATH=${PYTHONPATH}:${GITHUB_WORKSPACE}/bin/${{ matrix.build-type }}
         export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:${GITHUB_WORKSPACE}/bin/${{ matrix.build-type }}/libs
-        mpiexec -np 2 --hostfile ${GITHUB_WORKSPACE}/ci_hostfile python3 kratos/python_scripts/testing/run_cpp_mpi_tests.py --using-mpi
+        mpiexec -np 2 --hostfile ${GITHUB_WORKSPACE}/ci_hostfile python3 kratos/python_scripts/testing/run_cpp_mpi_tests.py
 
     - name: Running MPICore C++ tests (3 Cores)
       shell: bash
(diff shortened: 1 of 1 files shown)


philbucher:
I would rather change the runner file than the CI file

The argument is used to endure that MPI is initialized correctly

loumalouomega:
> I would rather change the runner file than the CI file
> 
> The argument is used to endure that MPI is initialized correctly

loumalouomega:
> I would rather change the runner file than the CI file
> 
> The argument is used to endure that MPI is initialized correctly

@jcotela agrees with you

jcotela:
@philbucher yes, this is actually a bug that I introduced by modifying the runner script in #11055 and that part needs a separate fix in any case. I will do a PR for that.

Having said that, I am not sure that the --using-mpi flag makes a difference here, the mpi module is automatically initialized if certain mpi-relevant environment variables are set, the command line argument only forces the initialization if the auto-detection fails.

philbucher:
I am aware of the auto initialization

For the CI I feel more comfortable explicitly doing the initialization, in case we update the underlying libs and the auto-init no longer works (e.g. newer OS)

Having said this I think the risk is low for this to happen, up to you guys

PR #11161, opened 2023-05-19
Correctly parse --using-mpi if given

jcotela:
**📝 Description**
Add proper command-line argument parsing to run_cpp_mpi_tests.py to prevent the `--using-mpi` command-line option from being incorrectly interpreted as a test name.

This fixes a bug introduced in #11055 which was causing the CI to not run the MPI tests. #11157 addresses the same issue.

**🆕 Changelog**
- Add command-line argument parsing to run_cpp_mpi_tests.py

### DIFF
--- a/kratos/python_scripts/testing/run_cpp_mpi_tests.py
+++ b/kratos/python_scripts/testing/run_cpp_mpi_tests.py
@@ -1,12 +1,29 @@
-import sys
+import argparse
 
 from KratosMultiphysics import mpi, Tester
 
-if __name__ == '__main__':
-    #Tester.SetVerbosity(Tester.Verbosity.TESTS_LIST)
-    Tester.SetVerbosity(Tester.Verbosity.TESTS_OUTPUTS)
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--using-mpi',
+        action='store_true',
+        help="Force MPI mode if auto-detection fails")
+    parser.add_argument(
+        'match_string',
+        metavar='MATCH-STRING',
+        nargs='?',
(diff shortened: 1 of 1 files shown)


loumalouomega (approved):
I close mine and patch the other one

loumalouomega (dismissed):

loumalouomega (approved):

kratos/python_scripts/testing/run_cpp_mpi_tests.py
```diff
@@ -1,12 +1,27 @@
-import sys
+import argparse
 
 from KratosMultiphysics import mpi, Tester
 
-if __name__ == '__main__':
-    #Tester.SetVerbosity(Tester.Verbosity.TESTS_LIST)
-    Tester.SetVerbosity(Tester.Verbosity.TESTS_OUTPUTS)
+def main():
+    parser = argparse.ArgumentParser()
```
philbucher, line 6: how about adding a check for running distributed?

Sth like `assert Kratos.IsDistibutedRun()`

but with proper error message

jcotela: Depends on what we want to accomplish. `IsDistributed` will always return true if `--using-mpi` is passed, even for something like `python run_cpp_mpi_tests.py  --using-mpi` which I guess is what we would be trying to prevent here...

If we want to have a better control of the number of threads used and the like, I would go for something like what is done for the python mpi tests, where there is a serial launcher and the mpi environment is created from a subprocess as needed.

loumalouomega (approved):

philbucher (approved):

PR #11162, opened 2023-05-19
RomApp Launch HROM simulations using HROM mdpa

Rbravo555:
**📝 Description**
This PR allows to seamlessly launch a simulation using only a subset of Elements and Conditions, by passing a flag to the RunHROM method of the RomManager.

PhilipaHROM

**🆕 Changelog**
- Added flag to the method RunHROM(mu, use_full_model part) to the RomManager
   - RunHROM method sets the name of the model part as "{original_name}HROM.mdpa" if the flag "use_full_model part" is left as False. Such a file should have been created during the training stage


Rbravo555:
After a breve internal discussion, the decision was taken to simplify the PR. Now the user only passes a flag if she wants to use the full_model_part. Otherwise, the HROM simulations will be run using the HROM mdpa by default

roigcarlo (approved):

PR #11164, opened 2023-05-21
[OptApp] Adding VtuOutput for OptimizationProblem data container

sunethwarna:
**📝 Description**
This PR adds `OptimizationProblemVtuOutputProcess` which is able to output selected (or all) components `ContanerExpression` or `CollectiveExpression` with the provided names and output them in vtu (ascii or binary) format. This process looks through all the expressions and finds the model parts in each `ContainerExpression` or `CollectiveExpression` and writes all the data relevant to one model part in one vtu file. 

**🆕 Changelog**
- Added `OptimizationProblemVtuOutputProcess`
- Added tests


RezaNajian (dismissed):
Another good job :-)

RezaNajian (dismissed):
@sunethwarna Thanks for the work. However, I tried to open the test vtu files with paraview and it gives error. could you please have a look. Also the names of the output files make me concerned, could you please comment on these points ?!

sunethwarna:
@RezaNajian Nice catch. I was never testing vtu for reall small data arrays such as data array of size 1 byte. Thats the reason why this was not caught. Now I changed the core base64 encoder as in #11203. I have changed the reference files also accordingly. Could you also check whether the vtu files are readable?

About the file names: `OptimizationProblem` has data for different model parts, and their model part names are little bit not so readable because we are creating unique names when using `Union` operators. And multiple components can use the same model part if possible. So I need one name for the model parts. The current approach is to find out all the components which share the same model part, and put all those component names with a prefix given by the user. I am open for suggestions on how to name these.

applications/OptimizationApplication/tests/process_tests/test_optimization_problem_vtu_output_process.py
```diff
@@ -0,0 +1,187 @@
...
+        cls.components_list.append(TestOptimizationProblemVtuOutputProcess.DummyControl("control_3", cls.model_part1))
+        cls.components_list.append(TestOptimizationProblemVtuOutputProcess.DummyExecutionPolicy("policy_1", cls.model_part1))
+        cls.components_list.append(TestOptimizationProblemVtuOutputProcess.DummyExecutionPolicy("policy_2", cls.model_part2))
+        cls.components_list.append(TestOptimizationProblemVtuOutputProcess.DummyExecutionPolicy("policy_3", cls.model_part1))
+
+        for component in cls.components_list:
+            cls.optimization_problem.AddComponent(component)
+            ComponentDataView(component, cls.optimization_problem).SetDataBuffer(2)
+
+        # cls.components_list.append("algorithm")
```
RezaNajian: if the commented not needed please remove

RezaNajian (dismissed):
There was a minor point you may address or not. Tnx for the work, we need to discuss name handling/style because it is bothering me :-)

applications/OptimizationApplication/tests/process_tests/test_optimization_problem_vtu_output_process.py
sunethwarna: This will be removed :)

sunethwarna: done :)

RezaNajian (approved):

PR #11166, opened 2023-05-22
[GeoMechanicsApplication] Geo/k0procedure

### LINKED ISSUE #10334: [GeoMechanicsApplication] Investigate the validity of the k0 procedure
- [ ] Test cases: 

  * [ ] Column of one soil material
  * [ ] Columns of horizontal layers
  * [ ] Slope with a uniform material
  * [ ] Columns of skewed layers
  * [ ] Slope with skewed layers

WPK4FEM:
**📝 Description**
A process to apply a k0 procedure to a stress field

**🆕 Changelog**
- Added k0 procedure process based on:
  * k0 for normal consolidation, sigma_horizontal = k0 * sigma_vertical, sigma_shear = 0.
  * taking into account a provided OCR parameter field
  * able to use friction angle retrieved from user supplied materials ( UMAT, UDSM )
- Tests for the 3 above variants.


applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
```diff
@@ -1042,9 +1042,14 @@ void SmallStrainUPwDiffOrderElement::
...
-        mConstitutiveLawVector[GPoint]->SetValue( rVariable, rValues[GPoint], rCurrentProcessInfo );
-
+        for (unsigned int GPoint = 0; GPoint < mConstitutiveLawVector.size(); ++GPoint) {
+            if (rVariable == CAUCHY_STRESS_VECTOR) {
+               mStressVector[GPoint] = rValues[GPoint];
+            }
+            else {
+                mConstitutiveLawVector[GPoint]->SetValue(rVariable, rValues[GPoint], rCurrentProcessInfo);
+            }
+        }
```
avdg81: If I'm not mistaken, the point here was that we needed different behavior for the Cauchy stress vector. The current implementation tests this for every iteration in the `for` loop, whereas `rVariable` won't change inside of the loop. Perhaps we can express this more directly as follows:
```suggestion
        if (rVariable == CAUCHY_STRESS_VECTOR) {
            std::copy(rValues.begin(), rValues.end(), mStressVector.begin());
        }
        else {
            for (unsigned int GPoint = 0; GPoint < mConstitutiveLawVector.size(); ++GPoint) {
                mConstitutiveLawVector[GPoint]->SetValue(rVariable, rValues[GPoint], rCurrentProcessInfo);
            }
        }
```
What do you think?

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.hpp
```diff
@@ -0,0 +1,161 @@
...
+
+    //----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+
+    /// Constructor
+    ApplyK0ProcedureProcess(ModelPart&  model_part,
+                           Parameters& rParameters) : Process(Flags()), mrModelPart(model_part)
+    {
+        KRATOS_TRY
+
+        mModelPartName      =  rParameters["model_part_name"].GetString();
```
avdg81: As far as I could see, this data member (`mModelPartName`) isn't used anywhere. If that is correct, I would suggest to delete it.

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.hpp
```diff
@@ -0,0 +1,161 @@
...
+    /// Destructor
+    ~ApplyK0ProcedureProcess() override{}
+
+    //----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+
+    void ExecuteFinalizeSolutionStep() override
+    {
+        KRATOS_TRY
+
+        if (mrModelPart.NumberOfElements() > 0) {
```
avdg81: To me, this condition seems redundant. The only thing that is done when the elements container is not empty, is to calculate the K0 stress for each element in it. But if the container is empty, the `block_for_each` effectively does nothing, right?

mcgicjn2:
This branch appears not to have been built. Can you look at this @WPK4FEM @avdg81

mcgicjn2:
@WPK4FEM Can you look into the test cases defined into #10334 I guess these might form our definition of done / acceptance criteria. @EleniSmyrniou Could you perhaps help Wijtze Pieter look into the tests you have defined.

mcgicjn2 (changes requested):
Requires further tests according to previous issue (or reasoning why should be avoid) Also currently not building.

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.hpp
```diff
@@ -0,0 +1,161 @@
...
+ 
+        KRATOS_CATCH("")
+    }
+
+    //----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+   
+  protected:
+    /// Member Variables
+
+    ModelPart& mrModelPart;
```
avdg81: In line with the recommendations from the C++ Core Guidelines, I would make this data member `private` instead of `protected`. See [C.133 Avoid `protected` data](https://sonarcloud.io/summary/new_code?id=KratosGeomechanics&branch=geo%2FK0procedure).

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.hpp
```diff
@@ -0,0 +1,161 @@
...
+    std::string mModelPartName;
+
+  private:
+
+      void CalculateK0Stresses(Element& rElement)
+      {
+          // Get K0 material parameters of this element ( probably there is something more efficient )
+          const Element::PropertiesType& rProp = rElement.GetProperties();
+          ConstitutiveLaw::Pointer pConstitutiveLaw = rProp.GetValue(CONSTITUTIVE_LAW);
+          const int& K0MainDirection = rProp[K0_MAIN_DIRECTION];
```
avdg81: I would change this reference-to-const to a const-value, since this is about a built-in type (`int`).
```suggestion
          const int K0MainDirection = rProp[K0_MAIN_DIRECTION];
```

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.hpp
```diff
@@ -0,0 +1,161 @@
...
+          if (K0MainDirection < 0 || K0MainDirection > 1) {
+              KRATOS_ERROR << "undefined K0_MAIN_DIRECTION in ApplyK0ProcedureProcess: " << K0MainDirection << std::endl;
+          }
+          KRATOS_INFO("apply_k0_procedure_process") << " K0_NC " << rProp.Has(K0_NC) << " index voor Phi  " << rProp.Has(NUMBER_OF_UMAT_PHI_PARAMETER) << std::endl;
+
+          //Check for alternative K0 specifications
+          double PoissonUR = 0.;
+          if (rProp.Has(POISSON_UNLOADING_RELOADING)) PoissonUR = rProp[POISSON_UNLOADING_RELOADING];
```
avdg81: `PoissonUR` is declared and initialized here, but only used much later (at line 141). I'd rather have these two closer to each other. For instance, you could move it to just before the `for` loop at line 136.

Furthermore, you could combine the two statements using the ternary operator. The advantage in that case is that you can make `PoissonUR` immutable:
```suggestion
          const auto PoissonUR = rProp.Has(POISSON_UNLOADING_RELOADING) ? rProp[POISSON_UNLOADING_RELOADING] : 0.;
```

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.hpp
```diff
@@ -0,0 +1,161 @@
...
+              K0Vector[1] = rProp[K0_VALUE_YY];
+              K0Vector[2] = rProp[K0_VALUE_ZZ];
+          }
+          else {
+              KRATOS_ERROR << "Insufficient material dat for K0 procedure process: " << std::endl;
+          }
+
+          //Loop over integration points
+          const Element::GeometryType& rGeom = rElement.GetGeometry();
+          const Element::GeometryType::IntegrationPointsArrayType& IntegrationPoints = rGeom.IntegrationPoints(rElement.GetIntegrationMethod());
```
avdg81: In general, for local variables in C++ code we are supposed to use `snake_case`. See the [Style Guide](https://github.com/KratosMultiphysics/Kratos/wiki/Style-Guide#local-variables).

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.hpp
```diff
@@ -0,0 +1,157 @@
...
+          rElement.CalculateOnIntegrationPoints(CAUCHY_STRESS_VECTOR, rStressVectors, rCurrentProcessInfo);
+
+          for (unsigned int GPoint = 0; GPoint < IntegrationPoints.size(); ++GPoint) {
+
+              // Determine OCR dependent K0 values
+              if ((rProp.Has(K0_NC) || rProp.Has(NUMBER_OF_UMAT_PHI_PARAMETER)) && rProp.Has(OCR)) {
+                  //Modify for presence of OCR (or POP?) field values
+                  double K0Value = rProp[K0_NC] * rProp[OCR] - (PoissonUR / (1.0 - PoissonUR)) * (rProp[OCR] - 1.0);
+                  std::fill(K0Vector.begin(), K0Vector.end(), K0Value);
+              }
```
avdg81: I'm not sure, but I don't think the OCR dependent K0 value is integration point dependent. (I do believe it is element dependent.) If that is correct, then we can move this piece of code out of the `for` loop.

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -134,10 +134,10 @@ void SetParameterFieldProcess::ExecuteInitialize()
 
     KRATOS_TRY
 
-
     const auto& r_var = KratosComponents< Variable<double> >::Get(mParameters["variable_name"].GetString());
 
     // set parameter field from input function
+    KRATOS_INFO("SetParameterFieldProcess ") << mParameters["func_type"].GetString() << r_var << std::endl;
```
avdg81: Was this statement intended for debugging? Or is it intended to be added permanently?

applications/GeoMechanicsApplication/python_scripts/geomechanics_U_Pw_solver.py
```diff
@@ -71,6 +71,7 @@ def GetDefaultParameters(cls):
             "number_cycles"              : 5,
             "increase_factor"            : 2.0,
             "reduction_factor"           : 0.5,
+            "realised_factor"            : 1.0,
```
avdg81: Perhaps this is a mistake?

applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
WPK4FEM: Incorporated, with length checks for vector sizes.

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.hpp
WPK4FEM: Done

applications/GeoMechanicsApplication/python_scripts/geomechanics_U_Pw_solver.py
WPK4FEM: realised_factor is removed. Now corrected.

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
WPK4FEM: Debugging, so removed.

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.hpp
WPK4FEM: Correct, done

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.hpp
WPK4FEM: I agree. Done.

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.hpp
WPK4FEM: Done.

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.hpp
WPK4FEM: Moved it closer, but still with the other material properties ( at 111 ).
Used the ternary operator.

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.hpp
WPK4FEM: Yes indeed, so changed. I consider this a flaw in the possibilities and would have liked a field that had values per integrationpoint.

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.hpp
WPK4FEM: Did this one ( and some others )

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.hpp
```diff
@@ -0,0 +1,139 @@
...
+          const Element::GeometryType& rGeom = rElement.GetGeometry();
+          const Element::GeometryType::IntegrationPointsArrayType& integration_points = rGeom.IntegrationPoints(rElement.GetIntegrationMethod());
+          for (unsigned int g_point = 0; g_point < integration_points.size(); ++g_point) {
+
+             // Apply K0 procedure
+             for (int i_dir = 0; i_dir <= 2; ++i_dir) {
+                  if (i_dir != k0_main_direction) {
+                      rStressVectors[g_point][i_dir] = k0_vector[i_dir] * rStressVectors[g_point][k0_main_direction];
+                  }
+              }
```
mcgicjn2, line 128:
```suggestion
             for (int dim_ind = 0; dim_ind <= 2; ++dim_ind) {
                  if (dim_ind != k0_main_direction) {
                      rStressVectors[g_point][dim_ind] = k0_vector[dim_ind] * rStressVectors[g_point][k0_main_direction];
                  }
              }
```

mcgicjn2 (dismissed):
If this has been code reviewed by @avdg81 then it can be merged

mcgicjn2 (commented):
k0 procdure simple dike tests - Is testing more than the k0 procedure on a dike ??? Is this wanted ?

Yes - See that it is a K0 is in stage 2 here - Fine

mcgicjn2 (changes requested):
Need better descriptions of tests \ files to reproduce

Maybe a readme in each test folder, or in documentation. 

and/or a plaxis comparison

mcgicjn2 (approved):

PR #11167, opened 2023-05-22
[Core] Check Container Template Arguments

matekelemen:
## Description

Check whether `block_for_each` overloads expecting containers really do get a valid container type as template argument.

### In a nutshell

If someone tries using `block_for_each` with iterators other than random access iterators, failed overload resolutions result in a cryptic error message. This PR adds compile-time checks to previously unchecked overloads, effectively disabling `block_for_each` if invalid template arguments are provided.

### Details

The current call hierarchy looks like this:
1) `block_for_each` on containers
2) `block_for_each` on iterators
3) `BlockPartition`

Each of the above steps can be invoked individually as well. In step 2, the iterator type is checked and if it's not a random access iterator, the overload is disabled. If the user wants to call this overload but provides an iterator of some other category (e.g.: a bidirectional iterator), the overload is still dropped and overload resolution will try invoking functions in step 1. Since there aren't any checks in step 1, the function isn't discarded but template instantiation fails because the function expects a container-like object but gets an iterator. The resulting error message misleads the user about where the problem truly lies.

This PR adds compile-time checks in step 1 that check whether the expected container-like object satisfied the requirements. Now if someone passes the wrong iterator type, these functions also get discarded and the compiler won't find a suitable overload. The error message correctly informs the user that no suitable overload was found for the provided arguments.

## Implementation

The check itself just tests whether `TContainerType` has a member function `begin` and whether it returns a valid iterator. This doesn't properly check whether `TContainerType` is a standard-conforming container, but it's better than nothing and doesn't pollute the template argument list that much.

Another reason why C++20 concepts would be nice to have.

### DIFF
--- a/kratos/utilities/parallel_utilities.h
+++ b/kratos/utilities/parallel_utilities.h
@@ -354,42 +354,82 @@ template <class TReduction,
 }
 
 /** @brief simplified version of the basic loop (without reduction) to enable template type deduction
- * @param v - containers to be looped upon
- * @param func - must be a unary function accepting as input TContainerType::value_type&
+ *  @tparam TContainerType A standard-conforming container type.
+ *  @tparam TFunctionType Functor operating on @a TContainerType::value_type.
+ *  @param v - containers to be looped upon
+ *  @param func - must be a unary function accepting as input TContainerType::value_type&
  */
-template <class TContainerType, class TFunctionType>
+template <class TContainerType,
+          class TFunctionType,
+          std::enable_if_t<!std::is_same_v<
+            std::iterator_traits<typename decltype(std::declval<std::remove_cv_t<TContainerType>>().begin())::value_type>,
+            void
+          >, bool> = true
+         >
 void block_for_each(TContainerType &&v, TFunctionType &&func)
 {
     block_for_each(v.begin(), v.end(), std::forward<TFunctionType>(func));
 }
 
 /** @brief simplified version of the basic loop with reduction to enable template type deduction
- * @param v - containers to be looped upon
- * @param func - must be a unary function accepting as input TContainerType::value_type&
+ *  @tparam TReducer Reduction type to apply. See @ref SumReduction as an example.
+ *  @tparam TContainerType A standard-conforming container type.
+ *  @tparam TFunctionType Functor operating on @a TContainerType::value_type.
+ *  @param v - containers to be looped upon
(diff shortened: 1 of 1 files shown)


sunethwarna:
Nice addition... Ok from my side :)

roigcarlo (dismissed):
:+1: Ty!!

sunethwarna:
I have a small suggestion, Instead of having the `std::enable_if` as in the following for the iterator begin and end overload,
https://github.com/KratosMultiphysics/Kratos/blob/12894f20cc485c4ed978d9b8139ea782bdabb03d/kratos/utilities/parallel_utilities.h#L293-L295

Why don't we allow all the `iterator_category` types for this, and have the following static assert to throw a meaningful error?
https://github.com/KratosMultiphysics/Kratos/blob/12894f20cc485c4ed978d9b8139ea782bdabb03d/kratos/utilities/parallel_utilities.h#L152

On top of the addition proposed in this PR.

matekelemen:
> I have a small suggestion, Instead of having the `std::enable_if` as in the following for the iterator begin and end overload,
> 
> https://github.com/KratosMultiphysics/Kratos/blob/12894f20cc485c4ed978d9b8139ea782bdabb03d/kratos/utilities/parallel_utilities.h#L293-L295
> 
> Why don't we allow all the `iterator_category` types for this, and have the following static assert to throw a meaningful error?
> 
> https://github.com/KratosMultiphysics/Kratos/blob/12894f20cc485c4ed978d9b8139ea782bdabb03d/kratos/utilities/parallel_utilities.h#L152
> 
> On top of the addition proposed in this PR.

I understand where you're coming from and think that it might be useful for less experienced developers, but at the same time it would be confusing for others that are more familiar with how C++ works (why do we check and accept a template parameter just to reject it in a static assert). 

I'll leave it up to you and @roigcarlo because you know our users better than I do.

roigcarlo:
I can see the benefits of both. @sunethwarna does your proposal prevent the code from compiling?. I would prioritize not letting the users run the code even if the compilation error es cryptic before having a more readable one during runtime ( mostly because they will be mad if it crashes after X hours of runtime )

sunethwarna:
@roigcarlo Yes, it will prevent it from compiling since we are using a `static_assert` with a meaningfull error message.

roigcarlo:
Go ahead then :+1:

sunethwarna (approved):
Nice... :)

PR #11169, opened 2023-05-22
[Core] Adding Hasher and Comparor for `array_1d<doube, 3>`

loumalouomega:
**📝 Description**

Adding `std::hash` and `operator==` for `array_1d<doube, 3>`. More changes detailed in changelog.

**🆕 Changelog**

- Changes from [[Core] Missing include in `GeometricalObject`](https://github.com/KratosMultiphysics/Kratos/pull/11179)
- [Simplify some hashes](https://github.com/KratosMultiphysics/Kratos/pull/11169/commits/a68bf30cc806ac26cf56b1a8595042bf3ff76fdd)
	1. Removal of the templating for the VariableHasher, VariableComparator, pVariableHasher, and pVariableComparator structures. This simplifies the code and makes it less error-prone.
	2. Removal of unnecessary `#include` statements, making the header lighter and reducing dependencies.
	3. The contents related to the hash computation (such as operator overloading) have been moved from the header file `key_hash.h` to a new source file `key_hash.cpp`, improving the organization of the code and reducing the compilation time.
	4. Simplified the usage of `std::unordered_map` in `integration_values_extrapolation_to_nodes_process.h` by removing the previously required specific hasher and comparator. This is possible due to the above-mentioned refactor.
	5. Removed some specific hashing and comparing code (like Array1D3DoubleHasher and Array1D3DoubleComparor), indicating a possible shift towards more generalized solutions.
- [Adding test for HashCombine and HashRange](https://github.com/KratosMultiphysics/Kratos/pull/11169/commits/494db4e876a9a9aaf52f392d11ca9699d19b5cb2)
	1. Improved formatting and organization: The changes include a number of formatting and spacing fixes. These adjustments aim to make the test file easier to read and maintain. 
	2. Addition of new tests: The commit introduces new tests such as `HashCombine`, `VariableComparator`, `SharedPointerComparator`, and others. These tests likely represent additional functionality or behavior that is now being checked.
In general, these changes should make the test file more robust and comprehensive, and may reflect changes in the functionality of the codebase that necessitate new or revised tests. This commit seems to be part of a larger effort to improve the testing infrastructure of the Kratos Multi-Physics project.
- [Adding `== operator` and hash method to `array_1d`](https://github.com/KratosMultiphysics/Kratos/pull/11169/commits/7e66df91113ee0e67708a762d491297f4eeab5cd)
	1. Formatting and Code Clean-up: Minor adjustments are made to improve the code's readability and maintainability. These include alignment of comments, and replacing the traditional include guard mechanism with `#pragma once`.
	2. Array Comparison: Added a comparison operator (`operator ==`) to the `array_1d` class. This operator allows for direct comparison of `array_1d` instances. It checks if the arrays are equal by comparing each of their elements.
	3. Hash Functionality: Introduced hashing functionality for `array_1d` instances. This change includes a definition of the `HashCombine` function and a specialization of `std::hash` for `Kratos::array_1d`. This allows `array_1d` to be used in STL containers like `unordered_map` or `unordered_set` that require hashing.
- [Adding new test for `array_1d`](https://github.com/KratosMultiphysics/Kratos/pull/11169/commits/e7554993d9b64064828e14709fd1ad8948699231)
	1. A new `#include <functional>` directive is added to include functional library.
	2. A new test `Array1DTest` is added, which checks if two 1D arrays are equal or not.
	3. Another new test, `SizeFunction`, is added which tests if the size function returns the expected value for a given array.
	4. The `IndexOperator` test is introduced to check the index operator.
	5. Two new test cases for addition (`AdditionOperator`) and subtraction (`SubtractionOperator`) of two arrays of 3 doubles are introduced.
	6. Finally, a test case (`HashesCorrectly`) is added to check if a Kratos array of doubles is hashed correctly.
- [Document hashercombine origin](https://github.com/KratosMultiphysics/Kratos/pull/11169/commits/1be6b2ee5bbf541ef0cf91065cab55bdae3363a2)
    - [Origin](https://www.boost.org/doc/libs/1_55_0/doc/html/hash/reference.html#boost.hash_combine)


kratos/includes/key_hash.h
```diff
@@ -456,9 +457,52 @@ namespace Kratos
...
+         * @brief Overloads the function call operator to hash a 1D array of three doubles.
+         * @param rArray The 1D array of doubles to hash.
+         * @return The hash value of the input array.
+         */
+        HashType operator()(const array_1d<double, 3>& rArray) const
+        {
+            HashType seed = 0;
+            HashCombine(seed, std::hash<double>{}(rArray[0]));
+            HashCombine(seed, std::hash<double>{}(rArray[1]));
+            HashCombine(seed, std::hash<double>{}(rArray[2]));
```
matekelemen:
```suggestion
            for (auto component : rArray) {HashCombine(seed, component);}
```

sunethwarna:
I would do it for `array_1d<double, T>` using fold expressions. Woudn't it be possible?

loumalouomega:
> I would do it for `array_1d<double, T>` using fold expressions. Woudn't it be possible?

Possible?, yes. But we mostly use array_1d<double,3>. BTW, first I was implementing for the Point class, but then I decided to use the array to make a bit more generic.

matekelemen (commented):
If you want to use `array_1d` in a hash map, I suggest specializing `std::hash` for `array_1d` instead of creating a separate class that does the same thing. It's the standard way to do it. As for `Array1D3DoubleComparor`, I'd prefer to add an overload for `operator==` or a specialization for `std::equal_to`.

loumalouomega:
> If you want to use `array_1d` in a hash map, I suggest specializing `std::hash` for `array_1d` instead of creating a separate class that does the same thing. It's the standard way to do it. 

Well, I tried std::hash in `array_1d` and it doesn't work, that's why the manual implementation.

> As for `Array1D3DoubleComparor`, I'd prefer to add an overload for `operator==` or a specialization for `std::equal_to`.

The comparers are implemented that way, Is what the map expects

matekelemen:
> > If you want to use `array_1d` in a hash map, I suggest specializing `std::hash` for `array_1d` instead of creating a separate class that does the same thing. It's the standard way to do it.
> 
> Well, I tried std::hash in `array_1d` and it doesn't work, that's why the manual implementation.
> 
> > As for `Array1D3DoubleComparor`, I'd prefer to add an overload for `operator==` or a specialization for `std::equal_to`.
> 
> The comparers are implemented that way, Is what the map expects

Ok, I'll be blunt then. This is what I want to see:
```cpp
namespace std {
template <>
struct hash<Kratos::array_1d<double,3>>
{
    std::size_t operator()(const Kratos::array_1d<double,3>& rArray) {...}
};
}
```
```cpp
namespace Kratos {
bool operator==(const array_1d<double,3>& rLeft, const array_1d<double,3>& rRight) {...}
}
```

There's no such thing as a "comparor". There's "comparison", "comparison operator", etc.

loumalouomega:
> > > If you want to use `array_1d` in a hash map, I suggest specializing `std::hash` for `array_1d` instead of creating a separate class that does the same thing. It's the standard way to do it.
> > 
> > 
> > Well, I tried std::hash in `array_1d` and it doesn't work, that's why the manual implementation.
> > > As for `Array1D3DoubleComparor`, I'd prefer to add an overload for `operator==` or a specialization for `std::equal_to`.
> > 
> > 
> > The comparers are implemented that way, Is what the map expects
> 
> Ok, I'll be blunt then. This is what I want to see:
> 
> ```c++
> namespace std {
> template <>
> struct hash<Kratos::array_1d<double,3>>
> {
>     std::size_t operator()(const Kratos::array_1d<double,3>& rArray) {...}
> };
> }
> ```
> 
> ```c++
> namespace Kratos {
> bool operator==(const array_1d<double,3>& rLeft, const array_1d<double,3>& rRight) {...}
> }
> ```
> 
> There's no such thing as a "comparor". There's "comparison", "comparison operator", etc.

Okay, according to this, if i override the std namespace and add the hash method then the maps could be automatically constructed?. If yes I will change it accordingly, but then we will need to replace the others ones

matekelemen:
> Okay, according to this, if i override the std namespace and add the hash method then the maps could be automatically constructed?. If yes I will change it accordingly, but then we will need to replace the others ones

You don't have to override anything. There's no runtime polymorphism involved. [`std::unordered_map`](https://en.cppreference.com/w/cpp/container/unordered_map) is a template with default parameters. You're providing definitions for those parameters. You'll be able to instantiate `std::unordered_map`. I don't know what other maps you have in mind so you'll have to check them yourself.

loumalouomega:
I added your suggestions @matekelemen (and something more, like tests)

kratos/containers/array_1d.h
```diff
@@ -761,4 +768,34 @@ class	array_1d	: public boost::numeric::ublas::vector_expression< array_1d<T, N>
...
+     * @todo Once the hashers and comparors are moved from key_hash.h, include key_hash and remove this. Right now there is a cross inclussion
+     */
+    template <class TClassType>
+    inline void HashCombine(
+        std::size_t& rSeed,
+        const TClassType& rValue
+        )
+    {
+        std::hash<TClassType> hasher;
+        rSeed ^= hasher(rValue) + 0x9e3779b9 + (rSeed<<6) + (rSeed>>2);
```
matekelemen, line 788: this smells like a copy-pasta from stackoverflow. Can you please elaborate on what this does and why these magic numbers are chosen?

loumalouomega: It is a copy paste from my own code, which probably I took from stackoverflow

loumalouomega: https://github.com/KratosMultiphysics/Kratos/blob/1f89ad0f5ab9b554f80013bd0bd39077cd083a6e/kratos/includes/key_hash.h#L57

loumalouomega: It is boost has_combine: https://stackoverflow.com/questions/35985960/c-why-is-boosthash-combine-the-best-way-to-combine-hash-values

loumalouomega: (used in Kratos since several years ago)

loumalouomega:
Interesting, hash returns different values in Windows and Linux..., how do I check this?...

matekelemen:
> Interesting, hash returns different values in Windows and Linux..., how do I check this?...

I'm guessing the standard doesn't force an implementation, so you shouldn't check its value directly.

loumalouomega:
> > Interesting, hash returns different values in Windows and Linux..., how do I check this?...
> 
> I'm guessing the standard doesn't force an implementation, so you shouldn't check its value directly.

Yes, std::hash is implemented differently, so I will check with a value of same value

kratos/containers/array_1d.h
matekelemen: aight, but can you plz move this outside the `std` namespace? Only the `hash` specialization should be there (we wouldn't want unintentional name collisions).

matekelemen: Oh and please leave a comment that this impl comes from boost.

loumalouomega: Okay (this means more compilation time...)

loumalouomega: > Oh and please leave a comment that this impl comes from boost.

Yes, as I said I just now rediscover, at that time I just copied from stackoverflow

image

loumalouomega: I copied from here (Firefox still remembers): https://stackoverflow.com/questions/7222143/unordered-map-hash-function-c

matekelemen: I don't have a problem with copying a snippet, but those magic numbers are a bit unnerving without proper explanation/sources.

matekelemen: > I copied from here (Firefox still remembers): https://stackoverflow.com/questions/7222143/unordered-map-hash-function-c

The [primary source](https://www.boost.org/doc/libs/1_55_0/doc/html/hash/reference.html#boost.hash_combine) would be better.

loumalouomega: If you take into account that I introduced in Kratos 6 years ago when Kratos was still moving from subversion and everything was super chaotic and documentation was close to 0 (closer than now), we are lucky that I can trackback my own web history and see what I did instead of being unknown legacy code

loumalouomega: Done @matekelemen

matekelemen (approved):

PR #11170, opened 2023-05-22
[GeoMechanicsApplication] Refactored Python class ApplyVectorConstraintTableProcess

avdg81:
**📝 Description**
Refactored Python class `ApplyVectorConstraintTableProcess`

**🆕 Changelog**
- Separated the construction of `Parameters` objects and processes.
- When a component is inactive, don't create a `Parameters` object nor add a process.
- Extracted the conversion from component to index to make the code easier to understand.
- Rearranged the `import` statements.
- Several stylistic changes.

### DIFF
--- a/applications/GeoMechanicsApplication/python_scripts/apply_vector_constraint_table_process.py
+++ b/applications/GeoMechanicsApplication/python_scripts/apply_vector_constraint_table_process.py
@@ -1,67 +1,83 @@
-import KratosMultiphysics
-import KratosMultiphysics.GeoMechanicsApplication as KratosGeo
+import KratosMultiphysics as Core
+import KratosMultiphysics.GeoMechanicsApplication as Geo
+
 
 def Factory(settings, Model):
-    if(type(settings) != KratosMultiphysics.Parameters):
+    if(type(settings) != Core.Parameters):
         raise Exception("expected input shall be a Parameters object, encapsulating a json string")
     return ApplyVectorConstraintTableProcess(Model, settings["Parameters"])
 
+
 ## All the python processes should be derived from "python_process"
 
-class ApplyVectorConstraintTableProcess(KratosMultiphysics.Process):
-    def __init__(self, Model, settings ):
-        KratosMultiphysics.Process.__init__(self)
-
+class ApplyVectorConstraintTableProcess(Core.Process):
+    def __init__(self, Model, settings):
+        super().__init__()
         self.model_part = Model[settings["model_part_name"].GetString()]
-        variable_name = settings["variable_name"].GetString()
+        self._CreateParametersForComponents(settings)
+        self._AddProcesses()
 
+
+    def _CreateParametersForComponents(self, settings):
+        self.x_params = self._CreateParametersForComponent(settings, "X")
+        self.y_params = self._CreateParametersForComponent(settings, "Y")
+        self.z_params = self._CreateParametersForComponent(settings, "Z")
+
+
+    def _AddProcesses(self):
         self.components_process_list = []
+        self._AddProcessUsing(self.x_params)
+        self._AddProcessUsing(self.y_params)
+        self._AddProcessUsing(self.z_params)
 
-        if settings["active"][0].GetBool() == True:
-            self.x_params = KratosMultiphysics.Parameters("{}")
-            self.x_params.AddValue("model_part_name",settings["model_part_name"])
-            if settings.Has("is_fixed"):
-                self.x_params.AddValue("is_fixed",settings["is_fixed"][0])
-            self.x_params.AddValue("value",settings["value"][0])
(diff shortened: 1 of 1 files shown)


WPK4FEM (dismissed):
Functionality has now a clearer split. Less code repetition.

applications/GeoMechanicsApplication/python_scripts/apply_vector_constraint_table_process.py
```diff
@@ -1,67 +1,82 @@
...
+
+    raise LookupError(f"Unknown component '{component}'")
+
+
 ## All the python processes should be derived from "python_process"
 
 class ApplyVectorConstraintTableProcess(KratosMultiphysics.Process):
-    def __init__(self, Model, settings ):
+    def __init__(self, Model, settings):
         KratosMultiphysics.Process.__init__(self)
```
huhaas: Can you please change this line to: `super().__init__(this)` ?

applications/GeoMechanicsApplication/python_scripts/apply_vector_constraint_table_process.py
```diff
@@ -1,67 +1,82 @@
...
+        index = ComponentToIndex(component)
+        if not settings["active"][index].GetBool():
+            return None
+
+        parameters = KratosMultiphysics.Parameters("{}")
+        parameters.AddValue("model_part_name", settings["model_part_name"])
+        if settings.Has("is_fixed"):
+            parameters.AddValue("is_fixed", settings["is_fixed"][index])
+        parameters.AddValue("value", settings["value"][index])
+        variable_name = settings["variable_name"].GetString() + "_" + component
```
huhaas: This:
`settings["variable_name"].GetString() + "_" + component`
Can be improved to this:
`f"{settings["variable_name"].GetString()}_{component}"`

applications/GeoMechanicsApplication/python_scripts/apply_vector_constraint_table_process.py
```diff
@@ -1,67 +1,82 @@
...
+from KratosMultiphysics.GeoMechanicsApplication import ApplyComponentTableProcess
+from KratosMultiphysics import ApplyConstantScalarValueProcess
 
 def Factory(settings, Model):
     if(type(settings) != KratosMultiphysics.Parameters):
         raise Exception("expected input shall be a Parameters object, encapsulating a json string")
     return ApplyVectorConstraintTableProcess(Model, settings["Parameters"])
 
+
+def ComponentToIndex(component):
```
huhaas: Why is this a separate function? Is it only relevant inside the class or also outside? @classmethod or private member function seems more fitting.

applications/GeoMechanicsApplication/python_scripts/apply_vector_constraint_table_process.py
```diff
@@ -1,67 +1,82 @@
 import KratosMultiphysics
-import KratosMultiphysics.GeoMechanicsApplication as KratosGeo
+from KratosMultiphysics.GeoMechanicsApplication import ApplyComponentTableProcess
+from KratosMultiphysics import ApplyConstantScalarValueProcess
```
huhaas: If Kratos is already imported in line 1, why have this on line 3?

huhaas (changes requested):
Can you fix these 4 minor things? :)

applications/GeoMechanicsApplication/python_scripts/apply_vector_constraint_table_process.py
avdg81: Agreed. We shouldn't repeat the base class name. The `self` argument (rather than `this`) is not needed here.

applications/GeoMechanicsApplication/python_scripts/apply_vector_constraint_table_process.py
avdg81: I agree that f-strings should be preferred over traditional string concatenations. I have adopted your suggestion in a slightly different form.

applications/GeoMechanicsApplication/python_scripts/apply_vector_constraint_table_process.py
avdg81: A class method probably makes more sense here. Changed according to your suggestion.

applications/GeoMechanicsApplication/python_scripts/apply_vector_constraint_table_process.py
avdg81: Reduced the three `import` statements to just two. We can now use `Core` as a qualifier when we mean `KratosMultiphysics` and `Geo` when we mean `KratosMultiphysics.GeoMechanicsApplication`. This makes the distinction between the two libraries clearer.

huhaas (approved):

PR #11175, opened 2023-05-23
[StructuralMechanicsApplication] Remove `MathUtils`template, replace `MathUtils<double>` with `MathUtils`

loumalouomega:
**📝 Description**

This PR refactors and cleans up code in the `StructuralMechanicsApplication`. The key change across all affected files is the replacement of `MathUtils<double>::` with `MathUtils::`, which streamlines the syntax and avoids unnecessary template specification.

These changes are primarily refactoring and should not alter the functionality of the code, but rather improve its readability and maintainability. 

**🆕 Changelog**

- [Remove MathUtils template](https://github.com/KratosMultiphysics/Kratos/commit/42dbb80e35762194751861a6ab2a1f42bb43b2a7)

### DIFF
--- a/applications/StructuralMechanicsApplication/custom_conditions/line_load_condition.cpp
+++ b/applications/StructuralMechanicsApplication/custom_conditions/line_load_condition.cpp
@@ -132,7 +132,7 @@ void LineLoadCondition<TDim>::CalculateOnIntegrationPoints(
             GetLocalAxis1(tangent_xi, J);
 
             // Computing normal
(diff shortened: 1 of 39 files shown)


AlejandroCornejo (approved):

loumalouomega:
@philbucher "'template<class TDataType> class Kratos::MathUtils' used without template arguments" I think it is not as you said

philbucher:
Hm maybe it only works within the same TU (aka Core) 

Too bad :/

loumalouomega:
> Hm maybe it only works within the same TU (aka Core)
> 
> Too bad :/

In that case the whole replacement is needed, and we can just do as intermediate step the Core

PR #11177, opened 2023-05-23
[Core][MPI] Add exit code to run_cpp_mpi_tests for use in CI

jcotela:
**📝 Description**
As @loumalouomega noticed in #11148, the MPI cpp tests are always reported as a success in CI, even if some of them fail. It seems to me that this is due to the fact that `run_cpp_mpi_tests.py` does not set the exit code (unlike other test scripts used in CI), so I'm creating a draft PR to test it.

**🆕 Changelog**
- add exit code to run_cpp_mpi_tests for use in CI


loumalouomega (dismissed):
If it passes the CI okay

loumalouomega (dismissed):

jcotela:
image
Looks like it is working now, so I'm reverting the test failure and moving from draft to full PR

loumalouomega (approved):

loumalouomega:
> image Looks like it is working now, so I'm reverting the test failure and moving from draft to full PR

image

PR #11180, opened 2023-05-23
[Core] Remove unused includes in DoF

loumalouomega:
**📝 Description**

1. It replaces the traditional include guard `#if !defined(KRATOS_DOF_H_INCLUDED)` with a `#pragma once` directive, which is a cleaner and more modern way of preventing multiple inclusions of the same header file. 
2. Some unnecessary includes are removed ("containers/array_1d.h"`), potentially reducing dependencies and improving compilation time. 

**🆕 Changelog**

- [Remove unused includes in DoF](https://github.com/KratosMultiphysics/Kratos/commit/e5e556f689677abd8988ee4bf14f3145b68e7600)

### DIFF
--- a/kratos/includes/dof.h
+++ b/kratos/includes/dof.h
@@ -4,15 +4,14 @@
 //   _|\_\_|  \__,_|\__|\___/ ____/
 //                   Multi-Physics
 //
-//  License:		 BSD License
-//					 Kratos default license: kratos/license.txt
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
 //
 //  Main authors:    Pooyan Dadvand
 //
 //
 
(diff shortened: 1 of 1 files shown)


RiccardoRossi:
i would say that the "define.h" is needed because of the definition of the pointersso i would expect that header to be needed everywhere, no?

loumalouomega:
> i would say that the "define.h" is needed because of the definition of the pointersso i would expect that header to be needed everywhere, no?

Technically is already included in other dependencies, but I can readded again

loumalouomega:
> i would say that the "define.h" is needed because of the definition of the pointersso i would expect that header to be needed everywhere, no?

Done

loumalouomega:
Anyone can approve this?

philbucher (approved):
LGTM

PR #11184, opened 2023-05-25
[GeoMechanicsApplication] Phreatic multi line

### LINKED ISSUE #10290: [GeoMechanicsApplication] Phreatic error with discontinuous geometry
As a user, I would like to be able to apply a multiple-point phreatic line, such that I can model water pressure fields.

### Description 

When multiple phreatic lines are assigned to a surface it seems that the first point in the left is always extrapolated to the right. As it can be seen in the figures bellow.

image

image

mcgicjn2:
**📝 Description**
Added phreatic multi line both statically (constant) and transiently (table) processes

**🆕 Changelog**
- Added apply constant phreatic multiline process 
- Added apply phreatic multiline table process
- Some changes to test helper

Closes #10290


mcgicjn2:
This is now being reopened as is required by several projects.

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
...
+{
+
+class ApplyConstantPhreaticMultiLinePressureProcess : public Process
+{
+
+public:
+
+    KRATOS_CLASS_POINTER_DEFINITION(ApplyConstantPhreaticMultiLinePressureProcess);
+
+///----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
rfaasse: In my opinion, these separator lines should not be necessary, could you remove them (and others in this PR?)

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
...
+
+    /// Constructor
```
rfaasse: Similar to the separator lines, the comments stating 'Constructor, Destructor, Assignment operator' etc are trivial so only result in extra clutter of the file. Could you remove them? (I know they are everywhere in our codebase, but still we can make newly added code a bit more clean)

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
...
+    ApplyConstantPhreaticMultiLinePressureProcess(ModelPart& model_part,
+                                                Parameters rParameters
+                                                ) : Process(Flags()) , mrModelPart(model_part)
+    {
+        KRATOS_TRY
+
+        //only include validation with c++11 since raw_literals do not exist in c++03
```
rfaasse: I don't think this comment is still relevant (I don't see any pre-compiler flags checking for C++ version)

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
...
+
+        mHorizontalDirection = 0;
+        for (unsigned int i=0; i<N_DIM_3D; ++i)
+           if (i!=GravityDirection() && i != OutOfPlaneDirection()) mHorizontalDirection = i;
+
+        mXCoordinates = rParameters["x_coordinates"].GetVector();
+        mYCoordinates = rParameters["y_coordinates"].GetVector();
+        mZCoordinates = rParameters["z_coordinates"].GetVector();
+
+        if (HorizontalDirection() == 0) {
```
rfaasse: This could be converted to a switch statement, which I think is slightly more readable than if, else if, else if.

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
...
+        else if (HorizontalDirection() == 1)
+        {
+            mHorizontalDirectionCoordinates = YCoordinates();
+        }
+        else if (HorizontalDirection() == 2)
+        {
+            mHorizontalDirectionCoordinates = ZCoordinates();
+        }
+
+        if (GravityDirection() == 0) {
```
rfaasse: Same comment here about the switch statement

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
...
+///----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+
+    /// this function is designed for being called at the beginning of the computations
+    /// right after reading the model and the groups
+    void ExecuteInitialize() override
+    {
+        KRATOS_TRY
+
+        if (mrModelPart.NumberOfNodes() <= 0) {
+            return;
```
rfaasse: Do you think it might be better to throw an error here instead of just returning, or does it make sense to continue when the number of nodes is <= 0?

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
...
+        const Variable<double> &var = KratosComponents< Variable<double> >::Get(VariableName());
+
+        if (IsSeepage()) {
+            block_for_each(mrModelPart.Nodes(), [&var, this](Node& rNode) {
+                const double pressure = CalculatePressure(rNode);
+
+                if ((PORE_PRESSURE_SIGN_FACTOR * pressure) < 0) {
+                    rNode.FastGetSolutionStepValue(var) = pressure;
+                    if (IsFixed()) rNode.Fix(var);
+                } else {
```
rfaasse: This alignment is a bit confusing, you have to look twice to see what the else is related to

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
...
+        rParameters["variable_name"];
+        rParameters["model_part_name"];
+
+        // Now validate agains defaults -- this also ensures no type mismatch
+        rParameters.ValidateAndAssignDefaults(default_parameters);
+
+        mVariableName = rParameters["variable_name"].GetString();
+        mIsFixed = rParameters["is_fixed"].GetBool();
+        mIsSeepage = rParameters["is_seepage"].GetBool();
+        mGravityDirection = rParameters["gravity_direction"].GetInt();
```
rfaasse: In general this constructor has quite a bit of lines. I would suggest to extract some functions and call them in the constructor, such that you don't need to see all details when looking at the constructor itself. Example:
```
Constructor
...
        mIsSeepage = rParameters["is_seepage"].GetBool();
        SetGravityDirection(rParameters);
...


    void SetGravityDirection(const Parameters &rParameters)
    {
        mGravityDirection = rParameters["gravity_direction"].GetInt();
        if (GravityDirection() == 0) {
            mGravityDirectionCoordinates = XCoordinates();
        }
        else if (GravityDirection() == 1)
        {
            mGravityDirectionCoordinates = YCoordinates();
        }
        else if (GravityDirection() == 2)
        {
            mGravityDirectionCoordinates = ZCoordinates();
        }
    }
```

And this could be similar for setting some of the other members (coordinates, horizontal direction etc)

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
...
+//   //  ____  //___) ) //   ) )
+//  //    / / //       //   / /
+// ((____/ / ((____   ((___/ /  MECHANICS
+//
+//  License:         geo_mechanics_application/license.txt
+//
+//  Main authors:    Vahid Galavi
+//                   Jonathan Nuttall
+
+#if !defined(KRATOS_GEO_APPLY_CONSTANT_PHREATIC_MULTI_LINE_PRESSURE_PROCESS )
```
rfaasse: Let's use #pragma once here for consistency

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
+// KRATOS___
```
rfaasse: Since this is not a templated class, there is no reason to keep all logic in the .hpp. In my opinion we should split it out into a .h and .cpp file.

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
...
+            KRATOS_ERROR << "The Horizontal Elements Coordinates are not ordered."
+                         << rParameters
+                         << std::endl;
+        }
+
+        mSpecificWeight = rParameters["specific_weight"].GetDouble();
+        if (rParameters.Has("pressure_tension_cut_off"))
+          mPressureTensionCutOff = rParameters["pressure_tension_cut_off"].GetDouble();
+        else
+          mPressureTensionCutOff = 0.0;
```
rfaasse: The alignment of this if/else is slightly off.

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
...
+
+    /// Member Variables
+
+    ModelPart& mrModelPart;
+
+    int findIndex(const Node& rNode) const
+    {
+        int index = 0;
+        auto coords = rNode.Coordinates();
+        auto noCoordinates = static_cast<int>(HorizontalDirectionCoordinates().size());
```
rfaasse: Maybe it's better to rename this to numberOfCoordinates

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
...
+    double PressureTensionCutOff() const
+    {
+        return mPressureTensionCutOff;
+    }
+
+///----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+
+protected:
+
+    /// Member Variables
```
rfaasse: Could you remove this comment?

applications/GeoMechanicsApplication/python_scripts/apply_scalar_constraint_table_process.py
```diff
@@ -58,11 +58,30 @@ def __init__(self, Model, settings ):
                 if settings.Has("is_seepage"):
                     self.params.AddValue("is_seepage",settings["is_seepage"])
 
-                if settings["table"][0].GetInt() == 0 and settings["table"][1].GetInt() == 0:
+                if all(i.GetInt() == 0 for i in settings["table"]):
                     self.process = KratosGeo.ApplyConstantPhreaticLinePressureProcess(self.model_part, self.params)
                 else:
                     self.params.AddValue("table",settings["table"])
                     self.process = KratosGeo.ApplyPhreaticLinePressureTableProcess(self.model_part, self.params)
+            elif settings["fluid_pressure_type"].GetString() == "Phreatic_Multi_Line":
```
rfaasse: This is probably going to result into a merge conflict with the work of @indigocoral, where the logic for this process is taken out of the python script and put into the C++ equivalent, so let's coordinate that accordingly.

applications/GeoMechanicsApplication/custom_processes/apply_phreatic_multi_line_pressure_table_process.hpp
```diff
@@ -0,0 +1,180 @@
+// KRATOS___
```
rfaasse: Since this is also no templated class, we should split it into .h and .cpp

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
...
+
+    ModelPart& mrModelPart;
+
+    int findIndex(const Node& rNode) const
+    {
+        int index = 0;
+        auto coords = rNode.Coordinates();
+        auto noCoordinates = static_cast<int>(HorizontalDirectionCoordinates().size());
+    	for ( ; index < noCoordinates; ++index)
```
rfaasse: The int index = 0 can be moved into the first part of this for-loop

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
...
+protected:
+
+    /// Member Variables
+
+    ModelPart& mrModelPart;
+
+    int findIndex(const Node& rNode) const
+    {
+        int index = 0;
+        auto coords = rNode.Coordinates();
```
rfaasse: both the coords and noCoordinated can be const.

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
...
+    double CalculatePressure(const Node &rNode) const
+    {
+        double height = 0.0;
+        int firstPointIndex;
+        int secondPointIndex;
+        double slope;
+
+        // find nodes in horizontalDirectionCoordinates
+
+        firstPointIndex = findIndex(rNode);
```
rfaasse: This could be just `const int firstPointIndex`, then we don't need `int firstPointIndex` in line 292. The same holds for secondPointIndex. Also for the naming to follow the convention it should be `first_point_index`.

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
...
+        secondPointIndex = firstPointIndex + 1;
+
+        slope = (GravityDirectionCoordinates()[secondPointIndex] - GravityDirectionCoordinates()[firstPointIndex]) /
+            (HorizontalDirectionCoordinates()[secondPointIndex] - HorizontalDirectionCoordinates()[firstPointIndex]);
+
+        height = slope * (rNode.Coordinates()[HorizontalDirection()] - HorizontalDirectionCoordinates()[firstPointIndex]) + GravityDirectionCoordinates()[firstPointIndex];
+
+        const double distance = height - rNode.Coordinates()[GravityDirection()];
+        const double pressure = - PORE_PRESSURE_SIGN_FACTOR * SpecificWeight() * distance;
```
rfaasse: Just for my understanding: why do we need the extra - sign here? Maybe it makes it more clear if we just change the order in calculating the distance so it gets the opposite sign

applications/GeoMechanicsApplication/custom_processes/apply_phreatic_multi_line_pressure_table_process.hpp
```diff
@@ -0,0 +1,180 @@
...
+//     //   ) )
+//    //         ___      ___
+//   //  ____  //___) ) //   ) )
+//  //    / / //       //   / /
+// ((____/ / ((____   ((___/ /  MECHANICS
+//
+//  License:         geo_mechanics_application/license.txt
+//
+//  Main authors:    Vahid Galavi
```
rfaasse: Could you add your name to the main authors?

applications/GeoMechanicsApplication/custom_processes/apply_phreatic_multi_line_pressure_table_process.hpp
```diff
@@ -0,0 +1,180 @@
...
+//
+
+#if !defined(KRATOS_GEO_APPLY_PHREATIC_MULTI_LINE_PRESSURE_TABLE_PROCESS )
```
rfaasse: Also here let's use #pragma once

applications/GeoMechanicsApplication/custom_processes/apply_phreatic_multi_line_pressure_table_process.hpp
```diff
@@ -0,0 +1,180 @@
...
+{
+
+public:
+
+    KRATOS_CLASS_POINTER_DEFINITION(ApplyPhreaticMultiLinePressureTableProcess);
+
+    /// Defining a table with double argument and result type as table type.
+    using TableType = Table<double, double>;
+
+///----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
rfaasse: Also here, we should get rid of the redundant comments (separator, constructor etc)

applications/GeoMechanicsApplication/custom_processes/apply_phreatic_multi_line_pressure_table_process.hpp
```diff
@@ -0,0 +1,180 @@
...
+
+    /// Copy constructor.
+    ApplyPhreaticMultiLinePressureTableProcess(ApplyPhreaticMultiLinePressureTableProcess const&) = delete;
+
+///----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+
+    /// this function will be executed at every time step BEFORE performing the solve phase
+    void ExecuteInitializeSolutionStep() override
+    {
+        KRATOS_TRY
```
rfaasse: There is duplicate code here which is there in ExecuteInitialize(). If I didn't miss anything, we could change this function to 
```
        ApplyConstantPhreaticMultiLinePressureProcess::ExecuteInitialize()
        const double Time = mrModelPart.GetProcessInfo()[TIME]/mTimeUnitConverter;
        std::vector<double> deltaH;
        std::transform(mpTable.begin(), mpTable.end(), std::back_inserter(deltaH),
                       [Time](auto element){return element ? element->GetValue(Time) : 0.0;});
```
However, that is still a bit strange, since Initialize should only be called once before the calculations according to the comment at that function. Let's discuss this

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
```diff
@@ -0,0 +1,346 @@
...
+    int findIndex(const Node& rNode) const
+    {
+        int index = 0;
+        auto coords = rNode.Coordinates();
+        auto noCoordinates = static_cast<int>(HorizontalDirectionCoordinates().size());
+    	for ( ; index < noCoordinates; ++index)
+        {
+	        if (HorizontalDirectionCoordinates()[index] >= coords[HorizontalDirection()])
+	        {
+                if (index == 0)
```
rfaasse: This if/else can be a one-liner: 
`return index == 0 ? index : index - 1`

applications/GeoMechanicsApplication/custom_processes/apply_phreatic_multi_line_pressure_table_process.hpp
```diff
@@ -0,0 +1,180 @@
...
+    {
+        rOStream << "ApplyPhreaticMultiLinePressureTableProcess";
+    }
+
+///----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+
+protected:
+    double CalculateTimeDependentPressure(const Node &rNode, std::vector<double> &deltaH) const
+    {
+
```
rfaasse: There is quite some overlap here with CalculatePressure in the base-class. We should be able to extract something general that we can use in both places

applications/GeoMechanicsApplication/custom_processes/apply_phreatic_multi_line_pressure_table_process.hpp
```diff
@@ -0,0 +1,180 @@
...
+#include "includes/table.h"
+
+#include "custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp"
+#include "geo_mechanics_application_variables.h"
+
+namespace Kratos
+{
+
+class ApplyPhreaticMultiLinePressureTableProcess : public ApplyConstantPhreaticMultiLinePressureProcess
```
rfaasse: The inheritance seems to be the wrong way around. The Constant... should inherit from the more generic one instead of this way around.

applications/GeoMechanicsApplication/tests/test_helper.py
```diff
@@ -172,51 +172,33 @@ def get_nodal_variable_from_ascii(filename: str, variable: str):
 
     # read data
     with open(filename, "r") as f:
-        all_data = f.readlines()
+        all_lines = f.readlines()
```
rfaasse, line 174: @avdg81 mentioned he and @indigocoral worked on a more general way of doing this. Maybe we can discuss if that can be applied here?

rfaasse (changes requested):
The code itself is quite clear, so most comments I have are minor changes to the structure and naming that can be done very quickly (~minutes).

However, there are three topics that will require more work:
1. The inheritance direction of these two classes is the wrong way around. The 'Constant' version is a more specific version of the 'normal' one, so 'Constant' should inherit from 'normal', not the other way around as it is now. This is the biggest item of this list.
2. The changes made to the apply_scalar_constraint_table_process should be moved to the C++ equivalent after these changes are merged to master.
3. The .hpp files for the processes should be split into .h and .cpp. This is not a lot of work, but slightly more than trivial.

I know this code was for a large part copied from earlier development, but since it's new code, adding it to master in the current state would add to our technical debt. Therefore, I'd propose that we first spend some time to improve it.

applications/GeoMechanicsApplication/custom_processes/apply_constant_phreatic_multi_line_pressure_process.hpp
rfaasse: I added a throw here, let me know if you agree

applications/GeoMechanicsApplication/custom_processes/apply_phreatic_multi_line_pressure_table_process.hpp
rfaasse: As discussed, we'll pick this up when looking at refactoring the phreatic lines anyway

applications/GeoMechanicsApplication/python_scripts/apply_scalar_constraint_table_process.py
rfaasse: Moved to the C++ file

applications/GeoMechanicsApplication/tests/test_absorbing_boundary.py
```diff
@@ -106,11 +106,14 @@ def run_and_assert_1d_column(self, file_path, node_nbrs, direction):
             expected_ini_time = dist/self.vp
 
             # find index of expected time of wave arrival in time list
-            ini_time_idx = test_helper.find_closest_index_greater_than_value(res["time"], expected_ini_time)
+            res_keys = list(res.keys())
+            ini_time_idx = test_helper.find_closest_index_greater_than_value(res_keys, expected_ini_time)
```
rfaasse, line 110: @mcgicjn2 These changes seem to give trouble in the pipeline for the slow tests (see [run logs](https://dpcbuild.deltares.nl/buildConfiguration/GEOFEA_KratosGeo_RunTests/2827080?expandBuildProblemsSection=true&hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&showLog=2827080_769_665&logView=flowAware))
image

I'm not too sure what the intentions were of making this change, maybe you can have a look at it?

rfaasse (dismissed):

applications/GeoMechanicsApplication/tests/test_absorbing_boundary.py
mcgicjn2: I had made changes which appear to have been reverted which meant that the dictionary output was more structured, i.e. first index is time_step, 2nd is node and then 3rd is the value of the variable, this was needed in the transient time stepping. Here I was just updating all tests which used this facility to operate correctly. It was failing due to a revert that must have occured during the review process. I have now re-added the functionality.

mnabideltares (dismissed):

applications/GeoMechanicsApplication/tests/test_absorbing_boundary.py
rfaasse: Okay, could be it happened when bringing the branch up to date with master, or I changed something I shouldn't have during the refactoring. In any case, thanks for double checking the PR and fixing this issue!

rfaasse (approved):

PR #11187, opened 2023-05-26
[Core] Container expressions slice, comb, reshape capabilities

sunethwarna:
**📝 Description**
This PR adds the slice, comb and reshape capabilities to container expressions. Few tests were added to explain the uses of these.

All of these slice, combine and reshape operations are creating new lazy expressions combining existing ones, hence has constant cost complexity.

@RezaNajian : Now you can slice a `array3` shaped container expression to 3 `scalar` expressions, and do computations seperately, and combine 3 `scalar` expressions to one `array3` container expression.

**🆕 Changelog**
- Add `UnarySliceExpression`, `UnaryReshapeExpression` and `UnaryCombineExpression`
- Added `SpecializedContainerExpression::Slice`, `SpecializedContainerExpression::Reshape` and `SpecializedContainerExpression::Comb` methods
- Expose them to python
- Add tests


kratos/containers/container_expression/specialized_container_expression.h
```diff
@@ -168,6 +168,16 @@ class SpecializedContainerExpression : public ContainerExpression<TContainerType
...
 
+    SpecializedContainerExpression Slice(
+        const IndexType Offset,
+        const IndexType Stride) const;
+
+    SpecializedContainerExpression Reshape(const std::vector<IndexType>& rShape) const;
+
+    SpecializedContainerExpression Comb(const BaseType& rOther) const;
+
+    SpecializedContainerExpression Comb(const std::vector<typename BaseType::Pointer>& rListOfOthers) const;
```
matekelemen, line 179: I'd prefer an implementation with iterators. The constructors with `std::vector` could then just call the ones with the iterators.

matekelemen: I'd prefer an implementation with iterators. The constructors with `std::vector` could then just call the ones with the iterators.

+referring to the docs of the operations would be nice. You can use `@ref` `@see`, `@copydoc`, etc.

kratos/containers/container_expression/expressions/unary/unary_combine_expression.h
```diff
@@ -0,0 +1,67 @@
...
+
+// Project includes
+#include "containers/container_expression/expressions/expression.h"
+
+namespace Kratos {
+
+///@name Kratos Classes
+///@{
+
+class KRATOS_API(KRATOS_CORE) UnaryCombineExpression : public Expression {
```
matekelemen: docs

kratos/containers/container_expression/expressions/unary/unary_combine_expression.cpp
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    KRATOS_ERROR << "Component index is greater than the current expressions stride.\n";
+
+    return 0.0;
+}
+
+const std::vector<std::size_t> UnaryCombineExpression::GetItemShape() const
+{
+    return {*(mStrides.begin() + mStrides.size() - 1)};
```
matekelemen:
```suggestion
    return {mStrides.back()};
```

matekelemen (commented):
I have 2 comments, but otherwise nice and clean 👍:
1) missing documentation
2) I'd create an interface with iterators for each function that takes `std::vector` arguments. For example, separate the following function
```cpp
void f(const std::vector<Stuff>& rArrayOfStuff)
{
    // do stuff
}
```
into two functions:
```cpp
void f(const Stuff* pBegin, const Stuff* pEnd)
{
    // do stuff
}

void f(const std::vector<Stuff>& rStuff)
{
    return f(rStuff.begin(), rStuff.end());
}
```

kratos/containers/container_expression/expressions/unary/unary_slice_expression.cpp
```diff
@@ -0,0 +1,76 @@
...
+    // Following will always convert shape = [1] to shape = [].
+    // Hence, if slicing is used, the output for scalars always will be
+    // shape = [] irrespective of the input scalar shape. If they still want to
+    // have the shape = [1], then UnaryReshapeExpression may be used.
+    if (mStride == 1) {
+        return {};
+    } else {
+        return {mStride};
+    }
+}
```
matekelemen, line 67: on second thought, why is the shape empty if the stride is 1? An empty shape would mean an empty expression.

kratos/containers/container_expression/specialized_container_expression.h
sunethwarna: done :)

kratos/containers/container_expression/expressions/unary/unary_slice_expression.cpp
sunethwarna: it is following the same as in numpy. numpy also return an empty shape if it is a scalar.

kratos/containers/container_expression/expressions/unary/unary_combine_expression.h
sunethwarna: done :)

matekelemen (dismissed):

sunethwarna:
@matekelemen I had to make `UnaryReshapeExpression` and `UnaryCombineExpression` headers only due to the fact that, having templated methods and part in cpp (even if the templated methods are in the .h) will fail in msvc compilation. I tried putting it in `impl` header, but then msvc can compile, but gcc and clang complains that multiple definitions are found for non-templated member functions (such as `Evaluate`). I don't know any other solution than putting everything in the same header.

matekelemen (approved):

PR #11188, opened 2023-05-26
[Core] Container Expression - Skip Expression::Evaluate vtable if possible

sunethwarna:
**📝 Description**
This PR checks whether the underlying expressions is of type `LiteralFlatExpression<T>` which has an underlying T contiguous array, if so, skips the `Expression::Evaluate` method  call and directly iterate through the contiguous array when `Evaluate` is called on the Container.

**🆕 Changelog**
- Skips `Expression::Evaluate` call which goes through the vtable if possible.


huhaas:
I'm curious, did you perform any measurements showing this improvement? or that this is a bottleneck? What is the typical (or average) size of the `r_expression` container? Looks like a classic example of readability vs performance.

Please add some comments above the first `if` to explain this addition but also the assumptions on which the change is based upon.

PR #11189, opened 2023-05-26
[Core] Refactor `GlobalPointerUtilities`: Add required includes to `global_pointer_utilities.h`

loumalouomega:
**📝 Description**

The PR adds four new header file includes to the `global_pointer_utilities.h`. These include the 'node', and 'variables' header files. These inclusions are needed to ensure that the 'global_pointer_utilities.h' file has access to all the necessary data types and functions that are defined in these other files. 

In addition, it changes the way the system identifies local objects by replacing the use of `Element` and `Condition` classes with the `GeometricalObject` class. The inclusion of the header files for the `Element` and `Condition` classes has been removed. The `GeometricalObject` header file is now included instead. The function `ObjectIsLocal` which previously checked if an `Element` or a `Condition` is local, has been refactored to now check if a `GeometricalObject` is local. This simplifies the code by reducing the number of separate checks and improves maintainability by reducing dependencies on the `Element` and `Condition` classes.

**🆕 Changelog**

- [Missing includes](https://github.com/KratosMultiphysics/Kratos/pull/11189/commits/3254eca02720a625001f2be8b100d79742d6ded0)
- [Using geometrical object instead of element/condition](https://github.com/KratosMultiphysics/Kratos/pull/11189/commits/1ced41a14d6e7f45210846732f6f1a4d42d78911)


kratos/utilities/global_pointer_utilities.cpp
```diff
@@ -0,0 +1,73 @@
...
+bool GlobalPointerUtilities::ObjectIsLocal(const Element& rElem, const int CurrentRank)
+{
+    return true; //if the iterator was found, then it is local!
+}
+
+/***********************************************************************************/
+/***********************************************************************************/
+
+bool GlobalPointerUtilities::ObjectIsLocal(const Condition& rCond, const int CurrentRank)
+{
```
RiccardoRossi: same as below

kratos/utilities/global_pointer_utilities.h
```diff
@@ -49,6 +49,10 @@ namespace Kratos
 ///@name Kratos Classes
 ///@{
 
+class Element;   /// Forward declaration
```
RiccardoRossi: we decided long ago that we should NOT use forward declarations with the goal of avoiding includes

loumalouomega: In that case I need to do a lot of  heavy includes

loumalouomega: Solved

loumalouomega:
Done @RiccardoRossi now just I added missing includes that in my machine provoke compilation to fail. I used the forward includes in order to reduce includes and increase even more compilations times.

philbucher:
This seems like code smell to me, hiding the actual problem

can you post the error?

loumalouomega:
> This seems like code smell to me, hiding the actual problem
> 
> can you post the error?

Condition/Element not defined. PARTITION_INDEX not defined. Basically missing includes

loumalouomega:
I will simplify with one include

philbucher:
But where is the error coming from?

Where is it used?

loumalouomega:
> But where is the error coming from?
> 
> Where is it used?

In a class I am developing

philbucher (approved):
Those includes were indeed missing, the Objects are used inside this class

roigcarlo:
Yep its fine, probably hidden because of the unitary builds :+1:

PR #11193, opened 2023-05-27
[Core] Fixed interval recurring event utility

sunethwarna:
**📝 Description**
This PR introduces the `FixedIntervalRecurringEventUtility` which has the same mechanism which was copy pasted in some of the output processes such as `unv_output_process.py`, `vtk_output_process.py` and `vtu_output_process.py`.

One behaviour change:
The existing implementations assumes output control type as "step" if the `output_control_type` is not `time`. This is changed, and now only supports `time` and `step` as the output control type, if any other is given an error is thrown.

**🆕 Changelog**
- Added `FixedIntervalRecurringEventUtility`
- Updated `unv_output_process.py`, `vtk_output_process.py`, and `vtu_output_process.py`.

### DIFF
--- a/kratos/python/add_other_utilities_to_python.cpp
+++ b/kratos/python/add_other_utilities_to_python.cpp
@@ -69,6 +69,7 @@
 #include "utilities/particles_utilities.h"
 #include "utilities/string_utilities.h"
 #include "utilities/model_part_operation_utilities.h"
+#include "utilities/fixed_interval_recurring_event_utility.h"
 
 namespace Kratos {
 namespace Python {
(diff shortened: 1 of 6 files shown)


loumalouomega (approved):

rubenzorrilla (changes requested):
Shouldn't we promote the use of controllers instead?

philbucher:
Also fyi with the current version of the vtk output Paraview will not order the timesteps correctly 

(We need to implement a small fix to write the file labels lexicographically)

sunethwarna:
@rubenzorrilla I totally forgot about the controller. I will change the implementation based on that and will create backward compatible Json settings...

@philbucher if everyone agrees, then i can change it too.

rubenzorrilla:
> @rubenzorrilla I totally forgot about the controller. I will change the implementation based on that and will create backward compatible Json settings...
> 
> @philbucher if everyone agrees, then i can change it too.

Indeed, if I'm not  wrong there's already an output with controller implemented by @ddiezrod. I'd follow that approach.

matekelemen:
`IntervalUtility` can already handle this functionality if you offset its input value and apply a modulo.

philbucher:
> @philbucher if everyone agrees, then i can change it too.

it is unfortunately not straight forward, it has been on my todo list since forever
Basically sth similar to what I did in the eigenvalue-post-process is required:
https://github.com/KratosMultiphysics/Kratos/blob/70f363bbcf6927fdb2a0b2fbae62ccb95b8eb053/applications/StructuralMechanicsApplication/custom_processes/postprocess_eigenvalues_process.cpp#L264

However unlike there, you dont know the max-time and thus not how many zeros to put at the front
In any case, I think it would be better do discuss this separately

matekelemen:
> > @philbucher if everyone agrees, then i can change it too.
> 
> it is unfortunately not straight forward, it has been on my todo list since forever Basically sth similar to what I did in the eigenvalue-post-process is required:
> 
> https://github.com/KratosMultiphysics/Kratos/blob/70f363bbcf6927fdb2a0b2fbae62ccb95b8eb053/applications/StructuralMechanicsApplication/custom_processes/postprocess_eigenvalues_process.cpp#L264
> 
> However unlike there, you dont know the max-time and thus not how many zeros to put at the front In any case, I think it would be better do discuss this separately

I'm not quite familiar with how VTK output sorts its files, but encoding data in a file's name has never been a good idea so I'd suggest a different solution to that problem (probably [something that keeps track of output files](https://github.com/KratosMultiphysics/Kratos/pull/10380)).

philbucher:
> > > @philbucher if everyone agrees, then i can change it too.
> > 
> > 
> > it is unfortunately not straight forward, it has been on my todo list since forever Basically sth similar to what I did in the eigenvalue-post-process is required:
> > https://github.com/KratosMultiphysics/Kratos/blob/70f363bbcf6927fdb2a0b2fbae62ccb95b8eb053/applications/StructuralMechanicsApplication/custom_processes/postprocess_eigenvalues_process.cpp#L264
> > 
> > However unlike there, you dont know the max-time and thus not how many zeros to put at the front In any case, I think it would be better do discuss this separately
> 
> I'm not quite familiar with how VTK output sorts its files, but encoding data in a file's name has never been a good idea so I'd suggest a different solution to that problem (probably [something that keeps track of output files](https://github.com/KratosMultiphysics/Kratos/pull/10380)).

Paraview sorts the files in lexicographical order, there is not much we can do abt this :man_shrugging:

sunethwarna:
Implemented using `Controller` in #11509 .

loumalouomega:
@sunethwarna close the branch if not needed

PR #11197, opened 2023-05-30
[CI] Update mpi test level name.

jcotela:
**📝 Description**
Updating test level name for mpi nightly runs after #11148 

**🆕 Changelog**

- Update nightly test suite name to mpi_nightly in CI.


loumalouomega (approved):
I though it was already changed

PR #11201, opened 2023-05-30
[GeoMechanicsApplication] Removed an unused data member

avdg81:
**📝 Description**
Removed unused data member `mModelPartName` from class `ApplyExcavationProcess`.


WPK4FEM (approved):
Clearly unused mModelPartName, so removal does decreases the maintenance burden.

Code in-between the changes seems to use a lot of doubling statements.

Consider to:

1. remove the if ( xxx > 0 ) conditions, as the block_for_each implies doing nothing if not present.
2. rElement.Set(ACTIVE, mDeactivateSoilPart) i.s.o. setting false or true in an if or else branch.
3. similar to 2 for rCondition.Set
4. Do the rElement.ResetConsitutiveLaw(); and the rNode.Set(ACTIVE, true) on a separate condition if 1 and 2 have been done

PR #11202, opened 2023-05-30
[Core] Added `KRATOS_FORCE_INLINE` macro for function inlining across different compilers

loumalouomega:
**📝 Description**

Mentioned in https://github.com/KratosMultiphysics/Kratos/pull/11086

This commit introduces the `KRATOS_FORCE_INLINE` macro to the `kratos/includes/define.h` file in order to force inline function behavior across different compilers. The macro's definition changes based on the detected compiler:

1. If the compiler is Microsoft's MSVC, it uses `__forceinline`.
2. If the compiler is GNU GCC, it uses `inline` with the `__always_inline__` attribute.
3. If the compiler is Clang and supports the `__always_inline__` attribute, it uses `inline` with the `__always_inline__` attribute. If the attribute is not supported, it defaults to `inline`.
4. For any other compilers, it defaults to `inline`. 

The introduction of this macro allows the Kratos codebase to optimize function inlining and enhance performance across different compilers.

**🆕 Changelog**

- [Adding macros to force inlining](https://github.com/KratosMultiphysics/Kratos/commit/148ae879b73840c5260c95a80009e66484ea582d)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,6 +57,12 @@ endif(NOT DEFINED ENV{KRATOS_INSTALL_PYTHON_USING_LINKS})
 # Setting the libs folder for the shared objects built in kratos
 set(CMAKE_INSTALL_RPATH "$ORIGIN/../libs")
 
+# If no inlining policy disable by default
+option(KRATOS_FORCE_INLINING "KRATOS_FORCE_INLINING defines if the compiler enforces the inlining. Default setting is OFF" OFF)
+if(${KRATOS_FORCE_INLINING})
+   add_definitions( -DKRATOS_FORCE_INLINING )
+endif(${KRATOS_FORCE_INLINING})
+
 # If no test policy enable by default
 option(KRATOS_BUILD_TESTING "KRATOS_BUILD_TESTING defines if the C++ tests are compiled. These increase compilation time, but if not set only python tests can be executed. Default setting is ON" ON)
 
--- a/kratos/includes/define.h
+++ b/kratos/includes/define.h
@@ -774,6 +774,20 @@ __pragma(warning(pop))
 #define KRATOS_STOP_IGNORING_DEPRECATED_FUNCTION_WARNING // not implemented for other compilers, hence does nothing
 #endif
 
+// Defining macros to force inlining
+#ifdef _MSC_VER
+    #define KRATOS_FORCE_INLINE __forceinline
+#elif defined(__GNUC__)
+    #define KRATOS_FORCE_INLINE inline __attribute__((__always_inline__))
+#elif defined(__CLANG__)
+    #if __has_attribute(__always_inline__)
+        #define KRATOS_FORCE_INLINE inline __attribute__((__always_inline__))
+    #else
+        #define KRATOS_FORCE_INLINE inline
+    #endif
+#else
+    #define KRATOS_FORCE_INLINE inline
+#endif
 
 namespace Kratos
 {
--- a/kratos/includes/define.h
+++ b/kratos/includes/define.h
@@ -785,6 +785,12 @@ __pragma(warning(pop))
     #else
         #define KRATOS_FORCE_INLINE inline
     #endif
+#elif defined(__INTEL_COMPILER)
+    #if __has_attribute(__always_inline__)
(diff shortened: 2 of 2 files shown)


loumalouomega:
It would depend of the compiler, but a priori it enforces the inline.

philbucher:
what about the intel compiler (ICPX)?

loumalouomega:
> what about the intel compiler (ICPX)?

The default is just inline, if you find the macro to be used we can add it

philbucher:
no clue, was just curious if you checked it out

loumalouomega:
> no clue, was just curious if you checked it out

I found that is the same as gcc and clang

loumalouomega:
**NOTE**: For greater control over inlining, we should consider using the GCC optimization options `-finline-functions`, `-finline-small-functions`, `-findirect-inlining`, `-finline-functions-called-once`, and others as needed.

matekelemen:
Are we sure we want to be smarter than the compiler?

loumalouomega:
> Are we sure we want to be smarter than the compiler?

I am not afraid, I know what I am doing

philbucher:
How about making this optional with a cmake flag? KRATOS_FORCE_INLINE
Then you could also add the compiler flags

If it proves to work well then it can be made default

philbucher:
as CMake `option`, see e.g. `KRATOS_BUILD_TESTING`

loumalouomega:
> as CMake `option`, see e.g. `KRATOS_BUILD_TESTING`

Okay, but otherwise to be a simple inline?

philbucher:
> > as CMake `option`, see e.g. `KRATOS_BUILD_TESTING`
> 
> Okay, but otherwise to be a simple inline?

I guess so, but thats up to you

loumalouomega:
> > > as CMake `option`, see e.g. `KRATOS_BUILD_TESTING`
> > 
> > 
> > Okay, but otherwise to be a simple inline?
> 
> I guess so, but thats up to you

Done

philbucher:
LGTM

this way can be tested without affecting others

@matekelemen wdyt?

matekelemen:
if it's optional, fine by me

RiccardoRossi:
just to tell that i would be in favor of this (as a very minimum i do not see this harmful)

loumalouomega:
> Let us discuss first in the @KratosMultiphysics/technical-committee

Is this already discussed, or the commentary of @RiccardoRossi is in his own?

roigcarlo:
We did not, but if @RiccardoRossi agrees, ok.

loumalouomega:
> We did not, but if @RiccardoRossi agrees, ok.

Then Okay, approved?

loumalouomega:
> > We did not, but if @RiccardoRossi agrees, ok.
> 
> Then Okay, approved?

Hello?

pooyan-dadvand:
@KratosMultiphysics/technical-committee is not sure what is the use case for having such a macro. Usually, compilers and optimizers are tend to decide better if it should be inlined or not. And how do we avoid miss use such a macro?

loumalouomega:
> @KratosMultiphysics/technical-committee is not sure what is the use case for having such a macro. Usually, compilers and optimizers are tend to decide better if it should be inlined or not. And how do we avoid miss use such a macro?

The idea of this PR is because compilers do not always take the best decision, there are many examples, particularly in MSVC. At the end by using these techniques, you provide stronger suggestions to the compiler to inline certain functions, but the compiler may still decide not to inline functions if it deems it inappropriate (e.g., the function is too large, recursive, etc.).

loumalouomega:
> > @KratosMultiphysics/technical-committee is not sure what is the use case for having such a macro. Usually, compilers and optimizers are tend to decide better if it should be inlined or not. And how do we avoid miss use such a macro?
> 
> The idea of this PR is because compilers do not always take the best decision, there are many examples, particularly in MSVC. At the end by using these techniques, you provide stronger suggestions to the compiler to inline certain functions, but the compiler may still decide not to inline functions if it deems it inappropriate (e.g., the function is too large, recursive, etc.).

Do i close this @KratosMultiphysics/technical-committee ?

pooyan-dadvand:
Yes, would you please close it?

loumalouomega:
> Yes, would you please close it?

I was asking for

PR #11203, opened 2023-05-30
[Core] Base64 encoder bugfix for data arrays of size 1 byte.

sunethwarna:
**📝 Description**
The `Base4EncodedOutput` may omit writing data array of `char` with 1 element if the previous data set had 1 byte leftover writing the encoded byte triplet. So this data in the array is never written. Suggested fix always forces to write all the data in the stream. This does not affect existing tests since such small data sets were not tested (Such small data sets usually appear only in unit testing) . Hence a proper small data set unit test is also added.

**🆕 Changelog**
- Fixed base64 encoding for small data sets.
- Added a test.

### DIFF
--- a/kratos/input_output/base_64_encoded_output.h
+++ b/kratos/input_output/base_64_encoded_output.h
@@ -24,7 +24,7 @@
 // Project includes
 #include "includes/define.h"
 
-namespace Kratos 
+namespace Kratos
 {
 ///@name Kratos Classes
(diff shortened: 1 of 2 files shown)


roigcarlo (approved):
I presume this has been very fun to debug :+1:

sunethwarna:
@roigcarlo Yeah, had a lot of fun debugging ;)

PR #11204, opened 2023-05-31
[Core] Move Expression Operations To Utilities

matekelemen:
## Description

Move operations defined on expressions (`Clone`, `Slice`, `Reshape`, `Comb`, `Pow`) from member functions to a utility. The new functions operate on their arguments in-place as opposed to the old behaviour, which was to create a new expression and return it.

## Reason

Expressions are implemented with runtime polymorphism, so users intending to interface with them have to do so through a base class (`ContainerExpression<TContainer>`). As a result, users cannot clone provided expressions since information on derived classes' types is lost, so operations that return a new object cannot be carried out. This PR exposes operations to the base class which were only defined on derived classes, with the caveat that each operation mutates its arguments instead of returning a newly constructed expression.


matekelemen:
@sunethwarna and I agreed to take this a step further and move `Read` and `Evaluate` functionalities to their own polymorphic class structure, which means that we'll be able to get rid of `SpecializedContainerExpression` alltogether. As a result, this PR is now obsolete.

PR #11207, opened 2023-05-31
[Core] Adding test for spatial search result

loumalouomega:
**📝 Description**

This PR introduces a new file `test_spatial_search_result.cpp` within the `kratos/tests/cpp_tests/spatial_containers` directory. It includes unit tests for a class named `SpatialSearchResult`. 

The added tests are as follows:
1. `SpatialSearchResultDefaultConstruction`: This test checks the default constructor of the `SpatialSearchResult` class. It asserts that initially, no object is found, no distance is calculated, and the object pointer is null.
2. `SpatialSearchResultObjectFound`: This test verifies the functionality of the `SetIsObjectFound` method by checking if the `IsObjectFound` method correctly reports an object as found after it has been set so.
3. `SpatialSearchResultDistanceCalculated`: Similar to `SpatialSearchResultObjectFound`, this test checks the functionality of `SetIsDistanceCalculated` method.
4. `SpatialSearchResultPointer`: This test constructs a `SpatialSearchResult` object with a `GeometricalObject` element and checks if the object returned by the `Get` method is the same as the one initialized.
5. `SpatialSearchResultDistance`: This test checks if the `SetDistance` method correctly sets the distance that can be retrieved by the `GetDistance` method.

This suite of tests are designed to ensure that the `SpatialSearchResult` class behaves as expected and that all of its methods function properly.

**🆕 Changelog**

- [Adding test for spatial search result](https://github.com/KratosMultiphysics/Kratos/commit/e1e42beaa38d560ffdf7a8781ab1148ca0318411)

### DIFF
--- a/kratos/spatial_containers/spatial_search_result.h
+++ b/kratos/spatial_containers/spatial_search_result.h
@@ -136,18 +136,38 @@ public:
         mIsDistanceCalculated = true;
     }
 
+    /// Getting if the object is found
+    bool GetIsObjectFound() const {
+        return mIsObjectFound;
+    }
+
+    /// Getting if the ibject is found
+    void SetIsObjectFound(const bool IsObjectFound) {
+        mIsObjectFound = IsObjectFound;
+    }
+
+    /// Getting if the distance is calculated
+    bool GetIsDistanceCalculated() const {
+        return mIsDistanceCalculated;
+    }
+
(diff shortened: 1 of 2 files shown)


loumalouomega:
@roigcarlo

roigcarlo (approved):

PR #11209, opened 2023-05-31
[TEST: DO. NOT. MERGE.] Testing pch performance in ci

roigcarlo:
**📝 Description**
This is a test. Ignore it.


loumalouomega:
But model part has a source file...

roigcarlo:
I know. I am just interested in checking something about the current unitary interactions and the impact on the compile time of the non unitary core build.

I chose that header because is a good trade-of between amount of code that will be pre-compiled and number of files in wich the pch is redundant or useless.

roigcarlo:
Btw @loumalouomega did you have a branch with an implementation of doctests in Kratos?

loumalouomega:
> Btw @loumalouomega did you have a branch with an implementation of doctests in Kratos?

I had, current one is uploaded is not working, and one I make it work I lost as local changes in the computer of my previous work. It is not very difficult to reimplement.

matekelemen:
you'll probably need the `REUSE_FROM` syntax in all applications to see a difference in compile times. In my experience, Core isn't the bottleneck.

roigcarlo:
> > Btw @loumalouomega did you have a branch with an implementation of doctests in Kratos?
> 
> I had, current one is uploaded is not working, and one I make it work I lost as local changes in the computer of my previous work. It is not very difficult to reimplement.

Na don't worry, I just wanted to make sure before redoing work

roigcarlo:
> you'll probably need the `REUSE_FROM` syntax in all applications to see a difference in compile times. In my experience, Core isn't the bottleneck.

I am getting about 50% performance increase in the core in my computer with that, but I suspect most of it comes from the combination of that with parallel compilation. I just want to know how impactful can be.

I cannot use `REUSE_FROM` directly in other applications because of the visibility attributes, but if the improvement is huge I will create pch to use for the core and the apps.

roigcarlo:
@matekelemen You see, its x3 faster in the CI for linux and x4.46 in windows (16 min vs 46 min, looking at ubuntu core without unity, 1h 7 min to 15 min for windows). This is very very interesting.

Jumbo builds do not change time, which is what I would expect, as pch is basically a subset of the unitary optimization.

Seems that intel es the only one that performs worse than the non pch version.

If I would have to bet, I would bet they are using very slow HDD, and the pch prevents the most of the reading (not parsing) time and dramatically reduces the number of disk lectures.

matekelemen:
> You see, its x3 faster in the CI for linux and x4.46 in windows (16 min vs 46 min, looking at ubuntu core without unity, 1h 7 min to 15 min for windows). This is very very interesting.

That's surprising. I would've guessed that windows would be the least affected since they have control over both their compiler **and** the linker, but I guess they aren't doing optimizations for compile times.

roigcarlo:
I wanted to do more tests but I don't have really the time for it. Since this is a straight forward increase on performance I would just merge and continue with this topic later. @KratosMultiphysics/technical-committee

philbucher (approved):

kratos/CMakeLists.txt
```diff
@@ -129,6 +129,11 @@ endif()
 pybind11_add_module(Kratos MODULE THIN_LTO ${KRATOS_PYTHON_SOURCES})
 target_link_libraries(Kratos PRIVATE ${KRATOS_TEST_LIBRARIES} KratosCore)
 
+# Precompile most demanding / widely included headers
+if(KRATOS_USE_PCH MATCHES ON)
+    target_precompile_headers(KratosCore PRIVATE "$<$<COMPILE_LANGUAGE:CXX>:${CMAKE_CURRENT_SOURCE_DIR}/includes/model_part.h>")
```
philbucher, line 134: can we easily extend this or proceed with caution?

Could we add all core-headers here?

roigcarlo: Caution better, the problem is that this gets included everywhere, so all the sources that do not make use of it will pay an overhead.

The trick is to only pch something that is shared by virtually everyone

philbucher: so what about e.g. `define.h` or `array_1d.h`? Those are already included in `model_part.h`, do they need to be added separately?

I guess `kratos_parameters.h` would be another good addition :thinking:

roigcarlo: It includes everything that hangs from that header, so array_1d and define are included, along with others

Issue #10696, opened 2023-01-30
Proposal to add a function "GetStageFinalData" to the analysis stage


RiccardoRossi:
**Description**
As of now there is no (non hacky) way to return data which are eventually computer within a stage. This makes cumbersome to have a stage to compute something (not to be written at disk!).

Two recent examples in which this capability would be useful are the following:
1 - computation of training data for ROM simulation (ping @Rbravo555 )
2 - i would like to change some constitutive parameters and compute the distribution of stresses that corresponds to this (i need it as a numpy array to use it later in an optimization loop 

in essence i would like to be able to do something of the type

~~~py
     ...within stage...
     def GetFinalData(self):
          results =  { "STRESSES" , self.stresses } by default results is an empty dictionary, but i could use it to return what i need
          return results
~~~

usage would be on the line of 

~~~py
   opt_params = {...}
   stage = MyCustomStage(Model, parameters, opt_params) # in here for example i apply opt_params and compute the stresses as a numpy array
   stage.Run()
   data = stage.GetFinalData() # here we return a python dictionary which contains the data that we will use for our optimization loop
~~~

roigcarlo:
For me it makes sense :+1:

rubenzorrilla:
Just saying that in the base class I'd return `None`.

roigcarlo:
I am closing this as it is already Implemented in some branches and currently discussed in the context of the multistage.

PR #11210, opened 2023-05-31
RomApp Get simulation residuals and store them as numpy files

NicolasSR:
**📝 Description**

This PR allows the user to obtain the residuals of the simulations at the FinalizeSolutionStep() of the analysis stage and to save them as .npy files.
It gets and stores the residuals from all simulations performed within the __LaunchTrainROM, __LaunchROM and __LaunchHROM methods of the RomManager To activate/deactivate this functionality, the user just needs to add the parameter "store_residuals=False/True" when calling the Fit() method of the RomManager.
The residuals are obtained by adding an intermediate analysis stage before the one defined in the user's own CustomizeSimulation method.
The residuals are stored in a folder named "./ResidualsMatrices/" in the current directory

**🆕 Changelog**

- Added file residual_output_analysis.py, which contains the new method GetResidualOutputAnalysisClass.
- Added parameter store_residuals with default value to the RomManager.Fit() method.
- Added method _StoreResidualsMatrix to the RomManager to store the residuals in .npy files.
- Modified RomManager.__LaunchTrainROM, RomManager.__LaunchROM and RomManager.__LaunchHROM to take the parameter store_residuals with default value and to apply the GetResidualOutputAnalysisClass and the RomManager._StoreResidualsMatrix methods if indicated.

### DIFF
--- a/applications/RomApplication/python_scripts/rom_manager.py
+++ b/applications/RomApplication/python_scripts/rom_manager.py
@@ -1,5 +1,6 @@
 import KratosMultiphysics
 from KratosMultiphysics.RomApplication.rom_testing_utilities import SetUpSimulationInstance
+from KratosMultiphysics.RomApplication.residual_output_analysis import GetResidualOutputAnalysisClass
 from KratosMultiphysics.RomApplication.calculate_rom_basis_output_process import CalculateRomBasisOutputProcess
 from KratosMultiphysics.RomApplication.randomized_singular_value_decomposition import RandomizedSingularValueDecomposition
 import numpy as np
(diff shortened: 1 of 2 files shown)


applications/RomApplication/python_scripts/rom_manager.py
```diff
@@ -660,3 +684,16 @@ def _StoreSnapshotsMatrix(self, string_numpy_array_name, numpy_array):
...
+
+    def _StoreResidualsMatrix(self, string_numpy_array_name, numpy_array):
+
+        # Define the directory and file path
+        directory = './ResidualsMatrices'
+        file_path = os.path.join(directory, f'{string_numpy_array_name}.npy')
+
+        # Create the directory if it doesn't exist
+        if not os.path.exists(directory):
+            os.makedirs(directory)
```
philbucher, line 696: I recommend to use `pathlib.Path` instead of `os.path`

NicolasSR: Thanks for the suggestion! For now I will keep as it is to be consistent with the rest of the RomManager, which I didn't implement. But maybe we can change it for the whole class

rubenzorrilla (changes requested):
I'm very hesitant to add a new analysis stage class to the ROM application (we did a huge effort in order to unify things). Instead, I'd follow the approach discussed in #10696 (of course we'll need to implement it first but it should be a pretty straightforward task).

Besides, I'd follow @philbucher's suggestion in here, even though it is not consistent with what has being done before in the app.

PR #11211, opened 2023-05-31
[GeoMechanicsApplication] Cleaned up class `ApplyExcavationProcess`

avdg81:
**📝 Description**
Cleaned up and simplified class `ApplyExcavationProcess`.

**🆕 Changelog**
Most changes are due to review comments from PR #11201:
- The checks whether some sequence is non-empty were redundant. They have been removed.
- When activating or deactivating parts of the model, directly pass the value of flag `mDeactivateSoilPart` rather than splitting it out explicitly.
- Avoid `protected` data members. Rather use `private` ones.
- Removed comments that had no additional value.
- Replaced (incorrectly positioned) inclusion guard by `#pragma once`.
- Removed some redundant blank lines and whitespace.
- Removed some unused type aliases.
- Don't provide an empty function body for the destructor, but use `= default` instead.
- Removed several redundant `#include`s.
- Replaced some occurrences of `block_for_each` by `VariableUtils::SetFlag`.

### DIFF
--- a/applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.hpp
+++ b/applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.hpp
@@ -11,30 +11,18 @@
 //                   Aron Noordam,
 //                   Vahid Galavi
 //
-//
-
-
-// System includes
-#include <cmath>
-#include <iostream>
-#include<string>
+#pragma once
 
 // Project includes
 #include "includes/model_part.h"
-#include "utilities/openmp_utils.h"
-#include "utilities/math_utils.h"
 #include "includes/element.h"
+#include "utilities/variable_utils.h"
 
 // Application includes
-
-#if !defined(KRATOS_GEO_APPLY_EXCAVATION_PROCESS )
-#define  KRATOS_GEO_APPLY_EXCAVATION_PROCESS
-
 #include "includes/kratos_flags.h"
 #include "includes/kratos_parameters.h"
 #include "processes/process.h"
 
-#include "geo_mechanics_application_variables.h"
 
 namespace Kratos
 {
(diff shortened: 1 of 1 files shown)


WPK4FEM (changes requested):
Please have a look at the necessity of the rather long list of includes.
E.g. are geo_mechanics_application_variables.h, cmath, and iostream necessary here

avdg81:
> Please have a look at the necessity of the rather long list of includes. E.g. are geo_mechanics_application_variables.h, cmath, and iostream necessary here

Done. Thanks for your feedback.

WPK4FEM (dismissed):
Now every remaining include is explainable.
Thank you.

applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.hpp
```diff
@@ -11,118 +11,68 @@
...
-
+        mDeactivateSoilPart = rParameters["deactivate_soil_part"].GetBool();
         KRATOS_CATCH("")
     }
 
-    ///------------------------------------------------------------------------------------
 
-    /// Destructor
-    ~ApplyExcavationProcess() override{}
+    ~ApplyExcavationProcess() override = default;
```
philbucher, line 45:
```suggestion
```

better let the compiler do its job ;)

applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.hpp
```diff
@@ -11,118 +11,68 @@
...
-                // Same nodes for both computing model part
-                if (mrModelPart.NumberOfNodes() > 0) {
-                    block_for_each(mrModelPart.Nodes(), [&](Node& rNode) {
-                        rNode.Set(ACTIVE, true);
-                    });
-                }
-            }
+        block_for_each(mrModelPart.Elements(), [this](Element& rElement) {
+            rElement.Set(ACTIVE, !mDeactivateSoilPart);
+        });
```
philbucher: better use the VariableUtils for this

applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.hpp
```diff
@@ -11,118 +11,68 @@
...
+
+        if (mDeactivateSoilPart) {
+            block_for_each(mrModelPart.Elements(), [](Element& rElement) {
+                rElement.ResetConstitutiveLaw();
+            });
+        } else {
+            block_for_each(mrModelPart.Nodes(), [](Node& rNode) {
+                rNode.Set(ACTIVE, true);
+            });
```
philbucher: variableUtils

applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.hpp
```diff
@@ -11,118 +11,68 @@
...
-    typedef std::size_t IndexType;
-    typedef Table<double, double> TableType;
-
     KRATOS_CLASS_POINTER_DEFINITION(ApplyExcavationProcess);
 
-    //----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 
-    /// Constructor
     ApplyExcavationProcess(ModelPart&  model_part,
                            Parameters& rParameters) : Process(Flags()), mrModelPart(model_part)
```
philbucher:
```suggestion
                           Parameters rParameters) : Process(Flags()), mrModelPart(model_part)
```

due to internal management, `Parameters` should be passed by value

applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.hpp
philbucher, line 1: I recommend to move the implementation to a cpp file

philbucher (commented):
some suggestions for improvements

applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.hpp
avdg81: My understanding is that when using the `= default` construct we actually get a compiler-generated one. See the explanation of variant number 3 on [this page](https://en.cppreference.com/w/cpp/language/destructor).

applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.hpp
avdg81: I agree that the code becomes more compact and perhaps more readable when we follow your suggestion. However, what I don't like here is that the `SetFlag` member function is _not_ `static`. That is, I need to create a _temporary_ `VariableUtils` object in order to call it. I would suggest to change `VariableUtils::SetFlag` into a `static` member function to avoid that (but in a separate commit, of course). Do you agree?

applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.hpp
avdg81: Thanks for pointing that out. I didn't know that.

applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.hpp
avdg81: I agree, but I had planned to do that in a separate commit. I prefer to take small steps, to keep PRs small and to make the review easier to digest ;-)

applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.hpp
avdg81: Done.

WPK4FEM (approved):
Further improvement indeed.

applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.hpp
philbucher: Yes, but that is always generated by the compiler, i.e. can be omitted

applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.hpp
philbucher: Makes full sense to have it static

Leftover from dark times :)

PR #11215, opened 2023-06-01
[Core] Refactor Expressions

matekelemen:
This PR is a work in progress.

## Description

A followup on #11204. Move reading/writing functionality from `ContainerExpression` to a new interface class `ExpressionIO` operating on `Expression`s directly. This effectively separates responsibilities and detaches `Expression`s from `Kratos` types (especially `ModelPart`), and allows more freedom when manipulating / writing them.

As a result, `Expression` will take the role of the primary interface for operations as well, with `ContainerExpression`s referring to the corresponding implementation.


sunethwarna (approved):
Thanks @matekelemen  for all the work :)

PR #11216, opened 2023-06-01
[GeoMechanicsApplication] Minor enhancements for Python function `Factory`

avdg81:
**📝 Description**
Addressed two kinds of code smells that were reported by SonarCloud.

**🆕 Changelog**
- Use builtin function `isinstance` rather than `type`.
- Raise a `TypeError` rather than the more generic `Exception` when the given settings object is of an unexpected type.


WPK4FEM (approved):
Consistent 2 thing fix.

PR #11227, opened 2023-06-02
Rom-based optimization

RezaNajian:
**📝 Description**
This PR implements ROM-based optimization using RomApplication and OptimizationApplication of Kratos. The idea is to use ROMs to accelerate optimization. This PR is the first proposal that should be reviewed and improved by the ROM and Optimization teams.

There are two major changes in RomApplication. First, in rom_builder_and_solver, the number of ROM modes is now updated on the fly. Second, I have introduced fom-rom-analysis.py to be used in optimization loop of OptApp. 

To provide an overview, I have conducted a comparison of thickness optimization results for a shell bridge using a basic strategy that involves a fixed number of iterations to update the ROM bases. The primary focus of the comparison lies in analyzing the distribution of thickness:

g18431

The plot illustrates the thickness distribution along the longitudinal section, with the full-fom represented in red and the fom-rom shown in green.

Overall, the results are consistent, and we achieved a 30% acceleration with the FOM-ROM analysis.

A test case has also been added to OptimizationApplication under fom-rom-opt-test.

I look forward to receiving comments and reviews.
Reza


rubenzorrilla (changes requested):
Very delicate. Blocking to prevent accidental merge.

Besides I see that you're using a custom analysis stage. Can't the standard `rom_analysis` be used?

RezaNajian:
@rubenzorrilla, thank you for your feedback. I couldn't find a clean way to use or inherit from rom_analysis. As a result, I took inspiration from it and introduced fom_rom_analysis instead. This new addition aims to encompass all "fom" analyses and maintain consistency with the other scripts.

Could you take a look at it and share your opinion?

applications/RomApplication/python_scripts/fom_rom_analysis.py
```diff
@@ -0,0 +1,114 @@
...
+                aux = KratosMultiphysics.Matrix(self.n_nodal_unknowns, num_rom_dofs)
+                for j in range(self.n_nodal_unknowns):
+                    for i in range(num_rom_dofs):
+                        aux[j,i] = self._current_basis.item((node_counter * self.n_nodal_unknowns +j,i))
+                node.SetValue(KratosROM.ROM_BASIS, aux ) # ROM basis
+                node_counter += 1
+            self.rom_solver._GetBuilderAndSolver().SetNumberOfROMModes(num_rom_dofs)
+            #self._GetComputingModelPart().ProcessInfo.SetValue(KratosROM.NUMBER_OF_ROM_MODES,num_rom_dofs)
+
+        def UpdateRomBases(self,snapshots_matrix,svd_truncation_tolerance=1e-6):
```
Rbravo555, line 90: This would be an "update" of the current basis. 

In this suggested implementation, the basis is updated only if the new snapshots are not captured by the old basis. 

Before, a new basis was being computed and imposed, independently of whether it was required.

Also, I would not use the randomized SVD, but the truncated SVD for this. The randomization makes sense when the number of columns in the matrix is much larger than its rank.

applications/RomApplication/custom_strategies/rom_builder_and_solver.h
```diff
@@ -229,7 +229,12 @@ class ROMBuilderAndSolver : public BuilderAndSolver<TSparseSpace, TDenseSpace, T
     SizeType GetNumberOfROMModes() const noexcept
     {
         return mNumberOfRomModes;
-    } 
+    }
```
Rbravo555, line 232: By only exporting this SetNumberOfROMModes method to python, we are less intrusive in the RomB&S and no new variable should be created

Rbravo555 (commented):
We in the ROM team were reluctant to adding a new analysis stage to the RomApplication, but although in the future this could be tackled using the multistage; as of now we do not see an alternative approach to your specific case than having this new fom_rom_analysis with both, a FOM and a ROM solver. 

It is worth noting that this PR cannot take full advantage of the RomApp features, for example the Hyper-Reduction was not initialized in this new fom_rom_analysis. For your expected usage it will not be necessary?

applications/RomApplication/python_scripts/fom_rom_analysis.py
Rbravo555: Admittedly, now we keep the basis stored in self._current_basis

rubenzorrilla (changes requested):
552 files changed. This definitively must be split.

applications/ConstitutiveLawsApplication/constitutive_laws_application_variables.cpp
```diff
@@ -107,6 +107,12 @@ KRATOS_CREATE_VARIABLE(Vector, MULTI_LINEAR_ELASTICITY_STRAINS)
 KRATOS_CREATE_VARIABLE(Vector, STRAIN_DAMAGE_CURVE)
 KRATOS_CREATE_VARIABLE(Vector, DELAMINATION_DAMAGE_VECTOR_MODE_ONE)
 KRATOS_CREATE_VARIABLE(Vector, DELAMINATION_DAMAGE_VECTOR_MODE_TWO)
+KRATOS_CREATE_VARIABLE(Vector, INTERFACIAL_NORMAL_STRENGTH_VECTOR)
+KRATOS_CREATE_VARIABLE(Vector, INTERFACIAL_SHEAR_STRENGTH_VECTOR)
+KRATOS_CREATE_VARIABLE(Vector, MODE_ONE_FRACTURE_ENERGY_VECTOR)
```
AlejandroCornejo, line 112: are you sure this change is yours?

applications/ConstitutiveLawsApplication/custom_constitutive/composites/rule_of_mixtures_law.cpp
```diff
@@ -680,6 +680,7 @@ Vector& ParallelRuleOfMixturesLaw<TDim>::CalculateValue(
         r_flags.Set( ConstitutiveLaw::COMPUTE_CONSTITUTIVE_TENSOR, flag_const_tensor );
         r_flags.Set( ConstitutiveLaw::COMPUTE_STRESS, flag_stress );
     } else {
+        if (rThisVariable != DELAMINATION_DAMAGE_VECTOR_MODE_ONE && rThisVariable != DELAMINATION_DAMAGE_VECTOR_MODE_TWO) {
```
AlejandroCornejo, line 683: idem @atfard

applications/ConstitutiveLawsApplication/custom_processes/element_deactivation_process.cpp
```diff
@@ -0,0 +1,127 @@
...
+//  Main authors:    Alejandro Cornejo
+//
+//
+
+// System includes
+
+// External includes
+
+// Project includes
+#include "includes/model_part.h"
```
AlejandroCornejo, line 20: my process? something is messed up IMO

AlejandroCornejo (changes requested):
There are many unrelated changes I think...

applications/DEMApplication/tests/test_history_dependent_CLs.py
```diff
@@ -46,20 +46,20 @@ def FinalizeSolutionStep(self):
 
             node = self.spheres_model_part.GetNode(5)
             force = node.GetSolutionStepValue(DEM.CONTACT_FORCES)
-            expected_value = -1219.77524
+            expected_value = -1152.7767
             self.CheckValueOfForce(force, 0, expected_value, tolerance)
             expected_value = 0.0
             self.CheckValueOfForce(force, 1, expected_value, tolerance)
-            expected_value = 1854.85877
+            expected_value = 1752.9904
```
maceligueta, line 53: How come this test had to be re-written? Have the results changed?

maceligueta: @joaquinirazabal this might affect your code

Issue #11208, opened 2023-05-31
Point output process - error trying to write after ExecuteFinalize 


ddiezrod:
We are not able to use the point output process after https://github.com/KratosMultiphysics/Kratos/pull/9908 in our altair stage. This stage changes the Finalize function from analysis stage to add a final print, so the function looks like this (pseudocode)

AltairAnalysis
```
def Finalize(self):

   processes.ExecuteFinalize() #Here we compute some final post-processes results

   solver.Finalize()

   OutputLastStep()


def OutputLastStep(self):
    processes.ExecuteBeforeOutputStep()
    output_processes.PrintOutput() #here we print the results computed in the ExecuteFinalize()
    processes.ExecuteAfterOutputStep()

```

The reason for this structure is that there are some post-process results that can be only computed at the very end of the stage, so that is why we added the OuputLastStep function.

The problem with the point output process is that this process closes its output file in ExecuteFinalize() so when we call PrintOutput the process throws an error becuse the file is close.
https://github.com/KratosMultiphysics/Kratos/blob/e36816eb641329d2da9462ae8b829a9f8c72b70a/kratos/python_scripts/point_output_process.py#L141

The first solution I thought was to move the last print inside the ExecuteFinalize of the processes that need it, the problem with this is that then we are not calling ExecuteBefore/AfterOutputStep for the rest of the processes in the list, so I do not think this is valid solution.

Other possible solutions are:
- Add a Free method to the processes that frees the resources at the very end of the stage. (This could also maybe be done calling the Clear method which is not being called now in the analysis stage)
- Close the file in the process destructor (we are not completely sure this works well in python)
- Somehow modify the process so after the file is closed , calling PrintOutput just passes (or throws a warning)

Any other ideas are very welcome!!
Pinging to hear your thoughts: @KratosMultiphysics/technical-committee @KratosMultiphysics/implementation-committee @philbucher @jcotela @KratosMultiphysics/altair

matekelemen:
I've heard other complaints about the point output process keeping files open (problems hitting the max number of open files supported by the OS) before. Seems to me that if you change the process to open file/write to file/close file, your (and others') problems will be solved.

ddiezrod:
> I've heard other complaints about the point output process keeping files open (problems hitting the max number of open files supported by the OS) before. Seems to me that if you change the process to open file/write to file/close file, your (and others') problems will be solved.

that'd be fine with me 👍

philbucher:
[...]

PR #11229, opened 2023-06-05
[Core] Point output process open/close file every time we write

ddiezrod:
This fixes the problems described in https://github.com/KratosMultiphysics/Kratos/issues/11208 , opening and closing the file after every print.

### DIFF
--- a/kratos/python_scripts/point_output_process.py
+++ b/kratos/python_scripts/point_output_process.py
@@ -134,11 +134,10 @@ class PointOutputProcess(KratosMultiphysics.OutputProcess):
                     out += " " + format(value,self.format)
 
             out += "\n"
-            f.write(out)
+            with open(f.name,'a') as file:
+                file.write(out)
 
-    def ExecuteFinalize(self):
-        for f in self.output_file:
-            f.close()
+
 
     def __SearchPoint(self):
         # retrieving the entity type
--- a/kratos/python_scripts/point_output_process.py
+++ b/kratos/python_scripts/point_output_process.py
@@ -187,8 +187,10 @@ class PointOutputProcess(KratosMultiphysics.OutputProcess):
             file_header = GetFileHeader(self.params["search_configuration"].GetString(),
                 entity_type, found_id, self.point, self.output_variables[0])
 
-            self.output_file.append(TimeBasedAsciiFileWriterUtility(
-                self.model_part, file_handler_params, file_header).file)
+            new_file = TimeBasedAsciiFileWriterUtility(
+                self.model_part, file_handler_params, file_header).file
+            new_file.close()
+            self.output_file.append(new_file)
 
 
     def __CheckVariableIsSolutionStepVariable(self, var):


matekelemen:
This results in a behaviour change at process construction time if a file already exists. I'd add an extra flag to the input parameters that defines what to do in such cases, and have a default value that ensures the old behaviour. For example:
```json
{
     ...
    "overwrite" : true // overwrite existing files at process construction if true, throw an exception otherwise
}
```

philbucher:
I think it is ok as is if the tests pass

Existing files is ok, used for restarts

mpentek:
What is the performance change, when you have 1000 files or similar order? I know it is not efficient to have some many ASCII files open, but is opening and closing at each time step better?

ddiezrod:
> What is the performance change, when you have 1000 files or similar order? I know it is not efficient to have some many ASCII files open, but is opening and closing at each time step better?

I'll run a test tomorrow and give you some numbers

ddiezrod:
@matekelemen That is true, although the entity search is done only once at the very beginning, so I assume that the difference wont be too big (I might be wrong)

mpentek:
> Hm that is more than I thought I don't oppose the changes, but perhaps someone could tests this also on the cluster?
> 
> @mpentek could you do this?
> 
> @ddiezrod can you please post your script?

I can first test on a cluster in 1-2 weeks. We have had cases with 2000-3000 files, although rarely. Also, this there is the option of piping point output to H5.

@ddiezrod is that an option?

philbucher (approved):
I think this is ok as is, since it solves relevant issues.

IMO if this becomes a bottleneck, then whatever it is used for should probably be replaced by sth more appropriate

@mpentek please block if you dont agree

@ddiezrod please wait until Monday with the merge to see if others have concerns too

mpentek:
I do agree with @philbucher, that such an output is anyway a bottleneck. So one thing is to have the correct syntax for the ascii-based point out, other is to see if an h5-based one meanwhile works and is the proper way for multiple points.

@ddiezrod is this PR urgent?
@matekelemen we had the discussion on piping to h5, is that in a state that we (@ddiezrod, @mtnzguillermo) can test?

As some of our campaigns use the point output a lot, on clusters, I will block this change before we have a working and tested solution.

mpentek (changes requested):
@ddiezrod we need a test on a cluster, with a real model part and many output points, including restart.

ddiezrod:
fine to me @mpentek , we need to fix this but it is not urgent

ddiezrod:
@mpentek Were you able to test this? its becoming more urgent in our side...

mpentek:
> @mpentek Were you able to test this? its becoming more urgent in our side...

Sorry, did not happen and most probably I will cause delay, if you depend on my input. I strongly urge a change like this to be tested on a cluster, with MPI, a large model, many output files and restart.

If you consider to proceed without doing so, I will not block, but I am not in favor of doing so.

ddiezrod:
Ok, I'll wait and add some fix in our private repo.

PR #11230, opened 2023-06-05
[CLApp] Adding Deactivate Element Process

AlejandroCornejo:
**📝 Description**
In this PR I am adding a process to deactivate elements according to a certain `Variable `and `Threshold`

~~~json
{
            "python_module" : "element_deactivation_process",
            "kratos_module" : "KratosMultiphysics.ConstitutiveLawsApplication",
            "process_name"  : "ElementDeactivationProcess",
            "Parameters"    : {
				"model_part_name"             : "Structure.Parts_Solid_Solid_Auto1",
				"variable_name"               : "DAMAGE",
				"variable_maximum_threshold"  : 0.99,
				"average_calculation_over_ip" : true
            }
        }
~~~

I though to implement this in the master based on the work of @atfard in his branch


AlejandroCornejo:
@atfard you can try to use this and let me know if it works properly.

applications/ConstitutiveLawsApplication/python_scripts/element_deactivation_process
```diff
@@ -0,0 +1,67 @@
+
+import KratosMultiphysics as KM
+
+def Factory(settings, Model):
```
loumalouomega: This could be a C++ process, it would be faster

loumalouomega (changes requested):
Add test + move to C++

AlejandroCornejo:
> Add test + move to C++

Yes indeed, after @atfard tries it, I will move it to parallel c++

atfard:
I checked it in a DCB test with isotropic damage, and is working properly. The point is that in Traction Separation CL, damage variable is vector so we should think about generalizing it.
image

loumalouomega (dismissed):
I applied the suggestions myself to avoid disturb you

atfard (dismissed):

atfard (approved):

PR #11236, opened 2023-06-07
Extend porous navier stokes

jgonzalezusua:
**📝 Description**
This PR is aimed to include the porous version of Navier-Stokes via VMS for both projections, ASGS and OSS.

- Applications
- Feature

**🆕 Changelog**
- Renovated `alternative_qs_vms_dem_coupled.cpp`, `alternative_d_vms_dem_coupled.cpp`, `qs_vms_dem_coupled.cpp` and `d_vms_dem_coupled.cpp`
- Added support for quadratic elements
- Included tests for quadratic elements to test second derivatives of the shape functions
- Added porosity term in cpp tests of the `alternative_qs_vms_dem_coupled.cpp`, `alternative_d_vms_dem_coupled.cpp`, `qs_vms_dem_coupled.cpp` and `d_vms_dem_coupled.cpp` elements.


applications/FluidDynamicsApplication/custom_elements/fluid_element.h
```diff
@@ -229,7 +229,7 @@ class FluidElement : public Element
      * @param rRightHandSideVector Local finite element residual vector (output)
      * @param rCurrentProcessInfo Current ProcessInfo values (input)
      */
-    void CalculateLocalVelocityContribution(
+    virtual void CalculateLocalVelocityContribution(
```
rubenzorrilla: Why is this required to be `virtual`?

applications/FluidDynamicsApplication/custom_strategies/schemes/bdf2_turbulent_scheme.h
```diff
@@ -270,7 +270,7 @@ class BDF2TurbulentScheme : public Scheme<TSparseSpace, TDenseSpace>
      * @param Dx Newton-Raphson iteration solution
      * @param b Newton-Raphson right hand side (unused)
      */
-    void Update(
+    virtual void Update(
```
rubenzorrilla: Why is the `virtual` required?

applications/FluidDynamicsApplication/custom_strategies/schemes/residualbased_simple_steady_scheme.h
```diff
@@ -213,6 +213,17 @@ class ResidualBasedSimpleSteadyScheme : public Scheme<TSparseSpace, TDenseSpace>
...
+        ModelPart &rModelPart,
+        TSystemMatrixType &A,
+        TSystemVectorType &Dx,
+        TSystemVectorType &b) override
+    {
+
+        BaseType::InitializeNonLinIteration(rModelPart, A, Dx, b);
+
+    }
+
```
rubenzorrilla: This is doing nothing. Please remove it and leave current base class implementation.

applications/FluidDynamicsApplication/tests/cpp_tests/test_alternative_qs_vms_dem_coupled_quadratic_element.cpp
```diff
@@ -0,0 +1,113 @@
+//    |  /           |
```
rubenzorrilla: I'd add this test to current `test_alternative_qs_vms_dem_coupled_element.cpp`. Note that you're testing the very same implementation but with different template instantiation.

applications/FluidDynamicsApplication/tests/cpp_tests/test_alternative_qs_vms_dem_coupled_quadratic_element_3D.cpp
```diff
@@ -0,0 +1,138 @@
+//    |  /           |
```
rubenzorrilla: I'd add this test to current `test_alternative_qs_vms_dem_coupled_element.cpp` under the very same reasoning I posted above.

applications/FluidDynamicsApplication/custom_strategies/schemes/bdf2_turbulent_scheme.h
jgonzalezusua: Because this function is derived in `bdf2_turbulent_schemeDEMCoupled.h`, which will be merge to master after this PR.

applications/FluidDynamicsApplication/tests/cpp_tests/test_alternative_qs_vms_dem_coupled_quadratic_element_3D.cpp
jgonzalezusua: I thought that it would be better to split these tests in three in order to ease the detection of errors. However, I can add them in only one.

applications/FluidDynamicsApplication/custom_strategies/schemes/residualbased_simple_steady_scheme.h
jgonzalezusua: Yes, it was for a testing implementation, I forgot to remove it.

applications/FluidDynamicsApplication/custom_elements/fluid_element.h
jgonzalezusua: This function is derived in the elements I have renovated. I need to initialize the second derivatives of the shape functions.

applications/FluidDynamicsApplication/custom_strategies/schemes/bdf2_turbulent_scheme.h
rubenzorrilla: But this is already `virtual` in the base class... I mean, it makes no sense to add it also here.

applications/FluidDynamicsApplication/tests/cpp_tests/test_alternative_qs_vms_dem_coupled_quadratic_element_3D.cpp
rubenzorrilla: The tests are already split by the macro.

applications/FluidDynamicsApplication/custom_elements/fluid_element.h
rubenzorrilla: The point is that by doing so you're introducing the vtable overhead. Considering that you know at compile time which elements are high order as we've a dimension and number of nodes template argument, I think we can achieve this in a more generic and efficient way by using an `if constexpr` together with an overload of the method that calculates the kinematics. Note that this approach would make possible to keep a unique implementation for all the elements deriving from the `FluidElement`.

applications/FluidDynamicsApplication/custom_strategies/schemes/bdf2_turbulent_scheme.h
jgonzalezusua: It is true

applications/FluidDynamicsApplication/tests/cpp_tests/test_alternative_qs_vms_dem_coupled_quadratic_element_3D.cpp
jgonzalezusua: Ok, so you mean to write them in the same .cpp file. I thought you meant to join them in only one test.  To me it is ok

rubenzorrilla: Exactly this.

applications/FluidDynamicsApplication/custom_elements/fluid_element.h
jgonzalezusua: Ok, but we shall consider to implement `GetShapeSecondDerivatives` in `FluidElement` and modify 'CalculateLocalVelocityContribution' using an `if constexpr`.

jgonzalezusua: Do you agree with that, @rubenzorrilla?

rubenzorrilla: No. `GetShapeSecondDerivatives` must be removed. Note that all the kinematics are computed at the `CalculateGeometryData`. Hence, my suggestion is to do overload of the `CalculateGeometryData` so we have two signatures, current one, without second order derivatives and a new one, with first and second order ones for high order geometries. Then you can filter which one is used at compile time by using and `if constexpr` with the dimension and number of nodes template argument (indeed, we can make a static member var to filter this).

Pointing that these changes must be definitively done in a separate PR.

jgonzalezusua: @rubenzorrilla, I am still confused. If I remove `GetShapeSecondDerivatives`, where do I implement the transform of the second shape derivatives?

rubenzorrilla: In a new overload of the `CalculateGeometryData`.

jgonzalezusua: @rubenzorrilla, this means that we must implement `GetShapeSecondDerivatives` in the `fluid_element` and overload the `UpdateIntegrationPointData` to call the second derivatives from data.

rubenzorrilla: No, you do another version of the `CalculateGeometryData` (method overload) that does what we're currently doing plus the second derivatives.

RiccardoRossi:
is this done?

jgonzalezusua:
> is this done?

@RiccardoRossi, in theory it is done, but we have decided to move the second derivatives transform to KratosCore. We are waiting for the approval of #11253.

jgonzalezusua:
It is very strange since in my local machine all tests run OK. But in github, the results in fluid cpp tests are different... How is it possible?

rubenzorrilla:
> It is very strange since in my local machine all tests run OK. But in github, the results in fluid cpp tests are different... How is it possible?

It could be a memory error. Are you running the tests in FullDebug?

jgonzalezusua:
> > It is very strange since in my local machine all tests run OK. But in github, the results in fluid cpp tests are different... How is it possible?
> 
> It could be a memory error. Are you running the tests in FullDebug?

Yes, I am running in FullDebug. That's why is so strange.

jgonzalezusua:
@rubenzorrilla tests are runnning. We are ready to merge.

applications/FluidDynamicsApplication/custom_elements/fluid_element.h
rubenzorrilla: It seems that this is not solved yet.

rubenzorrilla (commented):
The issue with the `virtual` seems to be still pending. Aside of this everything looks good to me.

jgonzalezusua:
@rubenzorrilla, I have removed the `virtual` from `fluid_element.h`

applications/FluidDynamicsApplication/custom_elements/alternative_d_vms_dem_coupled.cpp
```diff
@@ -170,11 +181,20 @@ void AlternativeDVMSDEMCoupled<TElementData>::Initialize(const ProcessInfo& rCur
...
     DVMS<TElementData>::Initialize(rCurrentProcessInfo);
 
+    if(Dim == 2){
+        if (NumNodes == 9 || NumNodes == 6 || NumNodes == 4)
+            mInterpolationOrder = 2;
+    }
+    else if(Dim == 3){
+        if (NumNodes == 10 || NumNodes == 27)
+            mInterpolationOrder = 2;
+    }
```
rubenzorrilla, line 191: As these are template parameters, in this case you can use `if constexpr` so you evaluate the if at compile time.

(I'd do this in a future PR considering that this is already taking a while...)

rubenzorrilla (approved):
No more changes in base elements. Thanks.

applications/FluidDynamicsApplication/custom_elements/alternative_d_vms_dem_coupled.cpp
jgonzalezusua: Yeah, you are right

PR #11239, opened 2023-06-07
[Core] Fix issues in stl reader with line ends and trailing spaces

jrubiogonzalez:
**📝 Description**
This PR fixes an issue when reading STL solids with trailing whitespaces. It also adds encloses between quotation marks the output modelpart names when a submodelpart is not found.

### DIFF
--- a/kratos/input_output/stl_io.cpp
+++ b/kratos/input_output/stl_io.cpp
@@ -11,7 +11,7 @@
 //
 
 // System includes
-
+#include <regex>
 // External includes
 
 // Project includes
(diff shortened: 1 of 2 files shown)


loumalouomega (approved):

loumalouomega:
I don't know why my changes doesn't appear

kratos/input_output/stl_io.cpp
```diff
@@ -217,7 +217,17 @@ void StlIO::ReadSolid(
...
+    std::size_t pos = word.find("COMMENT:");
+    if(pos!=std::string::npos){
+        word = word.substr(0,pos);
+    }
+    std::size_t pos = word.find(";");
+    if(pos!=std::string::npos){
+        word = word.substr(0,pos);
+    }
+    // Remove newline and spaces characters from string in C++
+    word = std::regex_replace( word, std::regex("\\r\\n|\\r|\\n| "), "");
```
matekelemen, line 230: I think this would be a bit cleaner, and you wouldn't have to `#include <regex>`.
```cpp
for (const auto& symbol : {"COMMENT:", ";"}) {
        const auto position = word.find(symbol);
        if (position != word.npos) {
            word.erase(position);
        }
    }
    word.erase(std::remove_if(
        word.begin(),
        word.end(),
        [](auto character) -> bool {
            return character == '\r' || character == '\n';
        }
    ), word.end());
```

PR #11241, opened 2023-06-07
[DO NOT MERGE] [Core] CMake Updates for Trilinos 14

matekelemen:
## Description

At long last Trilinos was updated in the AUR, bringing along a lot of changes. One of them is a delightful break with backward compatibility on their CMake interface ([they adopted CMake's `IMPORTED` syntax](https://github.com/trilinos/Trilinos/blob/2f8614ce01b61271bad2ccbc96e8d226124ebae6/RELEASE_NOTES#L40-L65)). This PR gets Kratos to compile with `Trilinos 14.0.0-1` (on my system) but I can't promise anything more.

Don't merge this PR; it's meant to be a guideline for people who want to update their Trilinos and get it working with Kratos.

### DIFF
--- a/applications/TrilinosApplication/CMakeLists.txt
+++ b/applications/TrilinosApplication/CMakeLists.txt
@@ -67,7 +67,7 @@ endif(${KRATOS_BUILD_TESTING} MATCHES ON)
 # ###############################################################
 ## TrilinosApplication core library (C++ parts)
 add_library( KratosTrilinosCore SHARED ${KRATOS_TRILINOS_APPLICATION_CORE_SOURCES} ${KRATOS_TRILINOS_TEST_SOURCES} )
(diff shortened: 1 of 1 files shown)


loumalouomega:
Just to say that in Trilinos 15 Epetra support will be removed (maybe 16), and we will need to migrate to TPetra

philbucher:
hopefully we will have dropped trilinos by the time they remove epetra, and replaced it by our own distributed linear algebra

FYI @RiccardoRossi ;)

philbucher:
> We'll have to pick a version and stick with it for Kratos because it seems to me that each Trilinos version has its own set of breaking changes, and it would be a nightmare to support multiple versions.

As good as this sounds, the reality on clusters can look quite different :/
Sometimes only one specific version workd

PR #11242, opened 2023-06-08
[Core] Extend definitions from `define_python`, adding `VectorToPyList` and `MatrixToPyList`

loumalouomega:
**📝 Description**

This PR primarily introduces two new functions `VectorToPyList` and `MatrixToPyList` that provide conversion from C++ vectors and matrices to Python lists using pybind11. These functions aim to enhance the interoperability between Python and the Kratos codebase.

The changes made include:
1. Two additional includes for `pybind11/stl.h` and `pybind11/stl/filesystem.h` have been moved to the beginning of the file to adhere to the proper structure of including external libraries.
2. A detailed comment block has been added to describe the `PrintObject` function, improving readability and understanding of its purpose and usage. 
3. The creation of `VectorToPyList`, a template function that converts C++ vectors into Python lists. It iterates over the input vector, appending each element to a pybind11 list which is then returned.
4. The creation of `MatrixToPyList`, another template function that converts C++ matrices into nested Python lists (list of lists). It iterates over the input matrix and for each row, it creates a new pybind11 list, appends the elements of the row to this list and then appends the row list to the final result list.

These new conversion functions are significant as they facilitate data transfer between the C++ and Python parts of the codebase.

**🆕 Changelog**

- [Extend definitions from define_python, adding `VectorToPyList` and `MatrixToPyList`](https://github.com/KratosMultiphysics/Kratos/commit/13dee6782a6a4fce6db4f27eb9ba04436145b5fa)

### DIFF
--- a/kratos/includes/define_python.h
+++ b/kratos/includes/define_python.h
@@ -16,15 +16,14 @@
 
 // External includes
 #include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+#include <pybind11/stl/filesystem.h>
 
 // Project includes
 #include "intrusive_ptr/intrusive_ptr.hpp"
 
 // Always needed for custom holder types
 PYBIND11_DECLARE_HOLDER_TYPE(T, Kratos::intrusive_ptr<T>);
-
-#include <pybind11/stl.h>
-#include <pybind11/stl/filesystem.h>
 
 #include "includes/define.h"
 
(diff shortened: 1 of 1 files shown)


kratos/includes/define_python.h
```diff
@@ -103,15 +102,53 @@ PYBIND11_DECLARE_HOLDER_TYPE(T, Kratos::intrusive_ptr<T>);
...
+ * @return The converted Python list.
+ */
+template<typename T>
+pybind11::list VectorToPyList(const T& results) {
+    pybind11::list list_results;
+    for (auto& r_result : results) {
+        list_results.append(r_result);
+    }
+    return list_results;
+}
```
matekelemen, line 134: [`std::vector` and other standard containers are implicitly converted to python lists/dicts if you `#include "pybind11/stl.h"`](https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html)

By the way, `results` is passed as a **const** reference `const T& results` but is iterated on by mutable reference `for (auto& r_results : results)`.

RiccardoRossi: also our vectors and matrices follow the buffer protocol, so it should be possible to convert them to list by simply doing 

~~~py
list(kratos_vector)
~~~

if you want that conversion to be automatic, than the way of doing it should be this one

https://github.com/KratosMultiphysics/Kratos/blob/5f204140be9345b9cb1c5420539415ca6f1556c7/kratos/python/add_vector_to_python.cpp#L248

loumalouomega: I will give a try

loumalouomega:
You are right, including STL the conversion is automatic, sorry for the noise

PR #11247, opened 2023-06-08
[Core][MPI] Add of `MPISearchUtilities`

loumalouomega:
**📝 Description**

The PR involves adding a new file to the project, specifically `mpi_search_utilities.h` under `kratos/mpi/utilities`. This new header file introduces new functionalities related to MPI-based (Message Passing Interface) point search operations.

In this PR, two main functions were introduced:
1. `MPISynchronousPointSynchronization` - This function prepares the coordinates of the points for an MPI search in a synchronized manner across all processes. It checks whether all points are identical across all processes. If they are, it collects their coordinates; if they are not, it gathers the points per partition and coordinates across all processes using the `AllGather` and `AllGatherv` MPI routines.
2. `CheckAllPointsAreTheSame` - This function verifies if all partitions have the same number of points and whether the points themselves are identical across all partitions. The verification is performed by summing coordinates across all partitions and checking if the average coordinate value (sum divided by the number of partitions) matches each coordinate value within a zero tolerance.

The commit provides new utilities for MPI searches, helping to synchronize and verify the consistency of point data across multiple processes.

**🆕 Changelog**

- [Initial add of `MPISearchUtilities`](https://github.com/KratosMultiphysics/Kratos/commit/fd0aeeb0664b54a41a9e80f0e6908770d40e6f75)
- [Adding tests for MPISearchUtilities](https://github.com/KratosMultiphysics/Kratos/pull/11247/commits/dc1be97aaec88b92a5760816975ed8668d01d14b)
	1. `MPISynchronousPointSynchronizationAllPointsAreTheSame`: This test case is designed to test the `MPISearchUtilities::MPISynchronousPointSynchronization` method with a scenario where all points are the same. The test checks that the number of points and their coordinates match the expected values after synchronization.
	2. `MPISynchronousPointSynchronizationAllPointsAreDifferent`: This test case is designed to validate the same `MPISynchronousPointSynchronization` method but with a scenario where all points are different. It checks that the size of all_points_coordinates matches the expected size and also verifies the correctness of the coordinates after synchronization.


loumalouomega:
Ready @pooyan-dadvand

loumalouomega:
@roigcarlo from all MPI search related PR this is the simplest one to review

sunethwarna:
I am curious. Is this gonna be first step towards `MPI` enabled `Tree` searches? :D  In any case, is there a specific reason why the `GlobalPointers` are not used in this and returning list of coordinates?

loumalouomega:
> I am curious. Is this gonna be first step towards `MPI` enabled `Tree` searches? :D In any case, is there a specific reason why the `GlobalPointers` are not used in this and returning list of coordinates?

Because this is not used as output in the solution, but as an input (must be copied in every partition). This is the synchronous version, I will add the asynchronous one in a future PR.

kratos/mpi/utilities/mpi_search_utilities.h
```diff
@@ -0,0 +1,284 @@
...
+        std::vector<double>& rAllPointsCoordinates,
+        const DataCommunicator& rDataCommunicator,
+        int AllPointsAreTheSame = -1,
+        int NumberOfPoints = -1,
+        int TotalNumberOfPoints = -1
+        )
+    {
+        // First check that the points are the same in all processes
+        if (AllPointsAreTheSame < 0) {
+            AllPointsAreTheSame = static_cast<int>(CheckAllPointsAreTheSame(itPointBegin, itPointEnd, NumberOfPoints, TotalNumberOfPoints, rDataCommunicator));
```
roigcarlo: There are several things here that I don't fully understand. I put is as a comment of this function because they are mainly related with this.

1 - If function does anything else than checking if all the points are the same, for example calculating `TotalNumberOfElements`, it should not be called check. Otherwise is very misleading.

2 - `TotalNumberOfPoints` is the actual total or the different ones? If is the total, it would be more performant to just calculate that number outside the function in order to prevent the the collective in line 97 (the `AllGather`) which essentially you are using for the same (`SumAll` = `AllGather` + `Reduce`, but you are repeating the `AllGather` below). If the total number is the different ones, then either `Check` does something more behind the scenes, or the line 97 collective is calculating something incorrect.

kratos/mpi/utilities/mpi_search_utilities.h
```diff
@@ -0,0 +1,284 @@
...
+        KRATOS_DEBUG_ERROR_IF(NumberOfPoints < 0) << "The number of points is negative" << std::endl;
+        KRATOS_DEBUG_ERROR_IF(TotalNumberOfPoints < 0) << "The total number of points is negative" << std::endl;
+
+        // If all points are the same
+        if (AllPointsAreTheSame) {
+            rAllPointsCoordinates.resize(NumberOfPoints * 3);
+            std::size_t counter = 0;
+            array_1d<double, 3> coordinates;
+            unsigned int i_coord;
+            for (auto it_point = itPointBegin ; it_point != itPointEnd ; it_point++) {
```
roigcarlo: `it_point++` + `counter++` or `++it_point` + `++counter` but don't mix.

kratos/mpi/utilities/mpi_search_utilities.h
```diff
@@ -0,0 +1,284 @@
...
+                noalias(coordinates) = it_point->Coordinates();
+                for (i_coord = 0; i_coord < 3; ++i_coord) {
+                    send_points_coordinates[3 * counter + i_coord] = coordinates[i_coord];
+                }
+                ++counter;
+            }
```
roigcarlo: Note that, if you move the calculation of the number of points outside the check function all this could be outside the if, and the only branching code will be the creation of the buffers and the `sendrecv`:

```c++

...
        send_points_coordinates[3 * counter + i_coord] = coordinates[i_coord];
....
if (AllPointsAreTheSame) {
    rAllPointsCoordinates = send_points_coordinates
} else {
   // What you have below...
}
```

kratos/mpi/utilities/mpi_search_utilities.h
```diff
@@ -0,0 +1,284 @@
...
+            std::vector<int> recv_offsets(world_size, 0);
+            for (int i_rank = 1; i_rank < world_size; ++i_rank) {
+                recv_offsets[i_rank] = recv_offsets[i_rank - 1] + recv_sizes[i_rank - 1];
+            }
+
+            // Invoque AllGatherv
+            rDataCommunicator.AllGatherv(send_points_coordinates, rAllPointsCoordinates, recv_sizes, recv_offsets);
+        }
+
+        return NumberOfPoints;
```
roigcarlo: This is calculated inside the `Check` and its misleading. Does include the repeated or not? Wouldn't be more natural to return `rAllPointsCoordinates/3` or something?

kratos/mpi/utilities/mpi_search_utilities.h
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    /**
+     * @brief MPISynchronousPointSynchronization prepares synchronously the coordinates of the points for MPI search.
```
roigcarlo: This does nothing with the coordinates, only with the sizes. Note that, even if `rAllPointsCoordinates` is propagated during all the stack call, you only use `all_points_are_the_same`, `number_of_points` and `total_number_of_points`, which do not need from the coordinate transfer to be calculated.

Seems odd that the coordinate stuff is inside this function. 

Also it repeats a lot of code. This function is calling `CheckAllPointsAreTheSame` in line 150, but then in line 153 you are calling `MPISynchronousPointSynchronization` which again has the same call to `CheckAllPointsAreTheSame` inside.

kratos/mpi/utilities/mpi_search_utilities.h
```diff
@@ -0,0 +1,284 @@
...
+            }
+        } else if (world_size == 1) { // In case only one process is used
+            recv_sizes[0] = number_of_points;
+        }
+
+        return recv_sizes;
+    }
+
+    /**
+     * @brief MPISynchronousPointSynchronization prepares synchronously the coordinates of the points for MPI search.
```
roigcarlo: Same comments as above

roigcarlo (changes requested):
Overall, after looking at my comments globally, I think that some changes are needed in the structure of the code.

My suggestion would be to make the code inside the current `MPISynchronousPointSynchronization*` functions more restricted and private (only the loops). For example:
- Check
- SynchronizePoints
- SynchronizeSize
- SynchronizeSizeWithRadius

Make the check function so its input parameters are the vector containing the partition numbers / partition coordinates in order to avoid calculating things twice or thrice.

Then create the three operations that you want in terms of the private functions you previously defined:

Ie:
```c++
MPISynchronousPointSynchronizationWithRadius() {
    std::vector<int> points_per_partition(world_size);
    std::vector<int> send_points_per_partition(1, number_of_points);
    rDataCommunicator.AllGather(send_points_per_partition, points_per_partition);

    CheckAndCalc(itBegin, itEnd, rNumber, rTotal, send_points_per_partition, rDataComm)

    SynchronizePoints(itBegin, itEnd, rNumber, rTotal, send_points_per_partition, rDataComm)
    SynchronizeSizeWithRadius(itBegin, itEnd, rNumber, rTotal, send_points_per_partition, rDataComm)
}
```

If `SynchronizeSize` and `SynchronizeSizeWithRadius` are to be called from other contexts, then those should be the ones responsible of maintaining the global data you need (or make a specific function for that), but the way is it right now too many collectives are repeated.

roigcarlo:
Also, could you provide an example in which you need to programmatically detect that all points are the same? I've been thinking for a while and unless is fabricated, I cannot come up with any scenario that satisfies that condition.

loumalouomega:
> Also, could you provide an example in which you need to programmatically detect that all points are the same? I've been thinking for a while and unless is fabricated, I cannot come up with any scenario that satisfies that condition

If you call the method passing `model_part.Nodes()` then the nodes will be different in each partition, but imagine you call the method in MPI from a list of points you have created in every partition, in that case you will get the same result N times.

kratos/mpi/utilities/mpi_search_utilities.h
loumalouomega: Okidoki

kratos/mpi/utilities/mpi_search_utilities.h
loumalouomega: Okay, true

kratos/mpi/utilities/mpi_search_utilities.h
loumalouomega: 1- Okay, improving name.
2- TotalNumberOfPoints is the sum of the nodes in all partitions, I will improve name.

kratos/mpi/utilities/mpi_search_utilities.h
loumalouomega: Once the check is done there are not repeated points

kratos/mpi/utilities/mpi_search_utilities.h
loumalouomega: Okay

kratos/mpi/utilities/mpi_search_utilities.h
loumalouomega: I think I will need to call you later, because I have questions

roigcarlo:
@loumalouomega Its ok, I had not considered that you could be creating the same list everywhere. If this is a use case for you no problem.

loumalouomega:
> @loumalouomega Its ok, I had not considered that you could be creating the same list everywhere. If this is a use case for you no problem.

More than case use for me, it is more like something that potentially someone (developer/advanced user) may do and then will be confused with the output.

loumalouomega:
Suggestions applied @roigcarlo (I hope)

roigcarlo (approved):
:heart:

PR #11249, opened 2023-06-09
Revert "[Structural] Remove UseGeometryIntegrationMethod from base solid"

loumalouomega:
Reverts KratosMultiphysics/Kratos#11155

### DIFF
--- a/applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
+++ b/applications/StructuralMechanicsApplication/custom_elements/base_solid_element.cpp
@@ -626,7 +626,7 @@ void BaseSolidElement::CalculateMassMatrix(
 
         Matrix J0(dimension, dimension);
 
-        IntegrationMethod integration_method = UseGeometryIntegrationMethod() ? IntegrationUtilities::GetIntegrationMethodForExactMassMatrixEvaluation(r_geom) : mThisIntegrationMethod ;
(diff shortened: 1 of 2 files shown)


rubenzorrilla (changes requested):
Instead of reverting and doing the more simpler later on, I'd directly do the more simpler.

loumalouomega:
> Instead of reverting and doing the more simpler later on, I'd directly do the more simpler.

After reading the code, I think it is no possible to do it simpler, and this completely breaks our code and it is essential

loumalouomega:
The only thing I can think as simpler is to define it as a property...

ddiezrod:
Just for you to have the context @rubenzorrilla , we have a couple of elements in altair where we move the integration nodes at each time step to correctly integrate the region of interest, thats why we override this method with UseGeometryIntegrationMethod = false

pooyan-dadvand:
> Instead of reverting and doing the more simpler later on, I'd directly do the more simpler.

What is your suggestion for simpler?

rubenzorrilla:
> > Instead of reverting and doing the more simpler later on, I'd directly do the more simpler.
> 
> What is your suggestion for simpler?

Not mine, @loumalouomega's one (he said so in the PR related to this one).

rubenzorrilla (approved):

PR #11254, opened 2023-06-11
[Core] move DataLocation to Globals

philbucher:
This has been on my todo list for a while, but never got to it
Now motivated by #11217 
FYI @AlejandroCornejo


loumalouomega (dismissed):
It's okay IMHO

matekelemen:
while we're here, can you expose it to python plz?

Mesh types could also have their own ```enum``` as well (local, ghost, interface - whatever that is).

philbucher:
> while we're here, can you expose it to python plz?

Done @matekelemen

matekelemen (approved):
thanks!

PR #11256, opened 2023-06-11
[Core] include ext libraries as SYSTEM

philbucher:
Same as we do for MPI, Trilinos, eigen etc ...
Not sure why we forgot about this one

Basically allows to ignore compiler warnings from the ext libs that we use, which would conflict with our rather strict CI


sunethwarna (approved):
Nice 👍

PR #11261, opened 2023-06-12
[Core[ Add maximimum and minimum edge calculation to hex3d8n geometry

jrubiogonzalez:
**📝 Description**
This PR contains adds the MaxedgeLength and MinEdgeLenght methods to the hex3d8n geometry

**🆕 Changelog**
In `hexahedra_3d_8n.h` both methods are added looping through all the edges. Two new tests have been added to `test_hexahedra_3d_8.cpp`

### DIFF
--- a/kratos/geometries/hexahedra_3d_8.h
+++ b/kratos/geometries/hexahedra_3d_8.h
@@ -613,7 +613,7 @@ public:
      */
     double MaxEdgeLength() const override {
         const auto edges = GenerateEdges();
-        double max_edge_length = -std::numeric_limits<double>::max();
+        double max_edge_length = 0.0;
         for (const auto& r_edge: edges) {
             max_edge_length = std::max(max_edge_length, r_edge.Length());
         }
--- a/kratos/geometries/hexahedra_3d_8.h
+++ b/kratos/geometries/hexahedra_3d_8.h
@@ -603,6 +603,40 @@ public:
             MathUtils<double>::Norm3(p3-p7)) /12.0;
     }
 
+    /** This method calculates and returns the maximum edge
+     * length of the geometry
+     *
+     * @return double value with the maximum edge length
+     *
+     * @see MinEdgeLength()
+     * @see AverageEdgeLength()
+     */
+    double MaxEdgeLength() const override {
+        const auto edges = GenerateEdges();
+        double max_edge_length = -std::numeric_limits<double>::max();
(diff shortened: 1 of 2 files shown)


kratos/geometries/hexahedra_3d_8.h
```diff
@@ -603,6 +603,40 @@ template<class TPointType> class Hexahedra3D8 : public Geometry<TPointType>
...
+     * length of the geometry
+     *
+     * @return double value with the maximum edge length
+     *
+     * @see MinEdgeLength()
+     * @see AverageEdgeLength()
+     */
+    double MaxEdgeLength() const override {
+        const auto edges = GenerateEdges();
+        double max_edge_length = -std::numeric_limits<double>::max();
```
philbucher:
```suggestion
        double max_edge_length = 0;
```

loumalouomega: Huuuum, Now I am confused

loumalouomega: in fact not, because length is not negative, is not easier to set to 0?

loumalouomega (commented):
Same comment as @philbucher

kratos/geometries/hexahedra_3d_8.h
philbucher: good point :D

jrubiogonzalez: lowest might not work with an all 0.00 distance, the idea is to use something negative, and as per code coherency I used -`-std::numeric_limits<double>::max()` the biggest negative number you can get. In fact both max and minimum were already computed in the `ShortestToLongestEdgeQuality`

loumalouomega (approved):

kratos/geometries/hexahedra_3d_8.h
philbucher: if you really want the largest negative number then `std::numeric_limits<double>::lowest()` is what you need: https://en.cppreference.com/w/cpp/types/numeric_limits/lowest

`
Returns the lowest finite value representable by the numeric type T, that is, a finite value x such that there is no other finite value y where y < x. This is different from [std::numeric_limits](https://en.cppreference.com/w/cpp/types/numeric_limits)<T>::min() for floating-point types. Only meaningful for bounded types.
`

still, I believe 0.0 would be the appropriate choice

philbucher (approved):

PR #11263, opened 2023-06-12
[CI] debug ICX failures

philbucher:
see title


kratos/tests/cpp_tests/strategies/schemes/test_residual_based_adjoint_bossak_scheme.cpp
```diff
@@ -1069,6 +1069,7 @@ double RunAdjointSensitivityTest(
 KRATOS_TEST_CASE_IN_SUITE(ResidualBasedAdjointBossak_TwoMassSpringDamperSystem_Elements, KratosCoreFastSuite)
 {
     Model current_model;
+    KRATOS_SKIP_TEST << "this test segfaults" << std::endl;
```
roigcarlo, line 1072: This will disable the test in all CI right? I would prefer not running them on ICX than disabling it for everyone :S

philbucher: dont worry, this is just for debugging

ofc this will be solved properly ;)

PR #11265, opened 2023-06-13
[MPM] Kinematic update of particles

VeronikaSinger:
Hey all,

the Newmark integration is used to update the velocity of the particles. The previous approach which updates the velocity using the accelerations is wrong for the first time step unless the particle acceleration is not set explicitly before the solution loop (which was done for gravity)

### DIFF
--- a/applications/ParticleMechanicsApplication/custom_elements/mpm_updated_lagrangian.cpp
+++ b/applications/ParticleMechanicsApplication/custom_elements/mpm_updated_lagrangian.cpp
@@ -963,6 +963,7 @@ void MPMUpdatedLagrangian::UpdateGaussPoint( GeneralVariables & rVariables, cons
     const unsigned int dimension = GetGeometry().WorkingSpaceDimension();
 
     const array_1d<double,3> & MP_PreviousVelocity = mMP.velocity;
+    const array_1d<double,3> & MP_PreviousAcceleration = mMP.acceleration;
 
     array_1d<double,3> delta_xg = ZeroVector(3);
     array_1d<double,3> MP_acceleration = ZeroVector(3);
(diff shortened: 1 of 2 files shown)


VeronikaSinger:
I did the changes as discussed yesterday for the velocity update as otherwise the calculation in the very first time step is not correct

applications/ParticleMechanicsApplication/custom_elements/mpm_updated_lagrangian.cpp
```diff
@@ -998,9 +998,12 @@ void MPMUpdatedLagrangian::UpdateGaussPoint( GeneralVariables & rVariables, cons
     }
 
     /* NOTE:
-    Another way to update the MP velocity (see paper Guilkey and Weiss, 2003).
-    This assume newmark (or trapezoidal, since n.gamma=0.5) rule of integration*/
-    mMP.velocity = MP_PreviousVelocity + 0.5 * delta_time * (MP_acceleration + MP_PreviousAcceleration);
+    Update velocity with Newmark integration rule 
+    This assumes newmark (since n.gamma=0.5 and beta=0.25) rule of integration*/
+    if (std::sqrt(MP_PreviousAcceleration[0]*MP_PreviousAcceleration[0]+MP_PreviousAcceleration[1]*MP_PreviousAcceleration[1]+MP_PreviousAcceleration[2]*MP_PreviousAcceleration[2])<std::numeric_limits<double>::epsilon())
```
ncrescenzio, line 1003:
```suggestion
    const double norm_MP_PreviousAcceleration = std::sqrt(MP_PreviousAcceleration[0]*MP_PreviousAcceleration[0]+MP_PreviousAcceleration[1]*MP_PreviousAcceleration[1]+MP_PreviousAcceleration[2]*MP_PreviousAcceleration[2]);
    
    if (norm_MP_PreviousAcceleration < std::numeric_limits<double>::epsilon())
```

Maybe defining an additional variable the code is a bit easier to read

PR #11266, opened 2023-06-13
[MPM] Refactor mpm_boundary_rotation_utility.h to reduce code duplication

gzjing:
**📝 Description**
Modified ApplySlipCondition and ConditionApplySlipCondition to allow RHS-only versions to reuse code from the LHS+RHS version. 

Main idea: the RHS-only version passes in a dummy LocalMatrix to the LHS+RHS version, which is modified to have no effect if the rLocalMatrix.size1() == 0.

### DIFF
--- a/applications/ParticleMechanicsApplication/custom_utilities/mpm_boundary_rotation_utility.h
+++ b/applications/ParticleMechanicsApplication/custom_utilities/mpm_boundary_rotation_utility.h
@@ -157,19 +157,19 @@ public:
 					array_1d<double,3> rN = rGeometry[itNode].FastGetSolutionStepValue(NORMAL);
 					this->Normalize(rN);
 
-					for( unsigned int i = 0; i < j; ++i)// Skip term (i,i)
-					{
-						rLocalMatrix(i,j) = 0.0;
-						rLocalMatrix(j,i) = 0.0;
-					}
-					for( unsigned int i = j+1; i < LocalSize; ++i)
-					{
-						rLocalMatrix(i,j) = 0.0;
-						rLocalMatrix(j,i) = 0.0;
-					}
+                    // Zero out row/column corresponding to normal displacement DoF except diagonal term (set to 1)
+                    // Applied IFF the local matrix passed is not empty [otherwise does nothing -- RHS only case]
+                    if (rLocalMatrix.size1() != 0) {
+                        for( unsigned int i = 0; i < LocalSize; ++i)
+                        {
+                            rLocalMatrix(i,j) = 0.0;
+                            rLocalMatrix(j,i) = 0.0;
+                        }
+                        rLocalMatrix(j, j) = 1.0; // set diagonal term to 1.0
+                    }
 
+                    // Set value of normal displacement at node directly to the normal displacement of the boundary mesh
 					rLocalVector[j] = inner_prod(rN,displacement);
-					rLocalMatrix(j,j) = 1.0;
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


applications/ParticleMechanicsApplication/custom_utilities/mpm_boundary_rotation_utility.h
```diff
@@ -179,26 +179,10 @@ class MPMBoundaryRotationUtility: public CoordinateTransformationUtils<TLocalMat
...
-					this->Normalize(rN);
-
-					rLocalVector[j] = inner_prod(rN,displacement);
-
-				}
-			}
-		}
+        // creates an empty dummy matrix to pass into the 'full' ApplySlipCondition -- this dummy matrix is
+        // ignored, effectively only updating the RHS
+        TLocalMatrixType dummyMatrix = *(new TLocalMatrixType());
```
philbucher: This is a memory leak, please do not allocate with new

philbucher (changes requested):

applications/ParticleMechanicsApplication/custom_utilities/mpm_boundary_rotation_utility.h
gzjing: Understood - would using ZeroMatrix() instead solve the problem?

gzjing: Oh wait sorry I see what you mean and where the problem lies now: I should just change it to
`TLocalMatrixType dummyMatrix = TLocalMatrixType();`

Thanks for the feedback!

gzjing: I've made the changes to both occurrences in the latest commit d7ad7d57dc4cb8a83ecf73410c8dea4a9c4be5f3

applications/ParticleMechanicsApplication/custom_utilities/mpm_boundary_rotation_utility.h
```diff
@@ -179,26 +179,10 @@ class MPMBoundaryRotationUtility: public CoordinateTransformationUtils<TLocalMat
...
-					this->Normalize(rN);
-
-					rLocalVector[j] = inner_prod(rN,displacement);
-
-				}
-			}
-		}
+        // creates an empty dummy matrix to pass into the 'full' ApplySlipCondition -- this dummy matrix is
+        // ignored, effectively only updating the RHS
+        TLocalMatrixType dummyMatrix = TLocalMatrixType();
```
philbucher:
```suggestion
        TLocalMatrixType dummyMatrix;
```

this should be enough

applications/ParticleMechanicsApplication/custom_utilities/mpm_boundary_rotation_utility.h
```diff
@@ -157,19 +157,19 @@ class MPMBoundaryRotationUtility: public CoordinateTransformationUtils<TLocalMat
...
-
+                    // Zero out row/column corresponding to normal displacement DoF except diagonal term (set to 1)
+                    // Applied IFF the local matrix passed is not empty [otherwise does nothing -- RHS only case]
+                    if (rLocalMatrix.size1() != 0) {
+                        for( unsigned int i = 0; i < LocalSize; ++i)
+                        {
+                            rLocalMatrix(i,j) = 0.0;
+                            rLocalMatrix(j,i) = 0.0;
+                        }
+                        rLocalMatrix(j, j) = 1.0; // set diagonal term to 1.0
```
philbucher, line 168: I think there is an identity matrix for this in boost

philbucher (commented):
looks good from my side
once @KratosMultiphysics/mpm approves I will too

applications/ParticleMechanicsApplication/custom_utilities/mpm_boundary_rotation_utility.h
gzjing: I'm not too sure if the identity matrix will help in this case, since this operation is only for certain row/columns - the `j` index does not go through all DoFs, but only the first DoF for nodes flagged as `SLIP`:

```
for(unsigned int itNode = 0; itNode < rGeometry.PointsNumber(); ++itNode)
			{
				if(this->IsSlip(rGeometry[itNode]) )
				{
					// We fix the first displacement dof (normal component) for each rotated block
					unsigned int j = itNode * this->GetBlockSize();
```

VeronikaSinger (approved):
Thanks!

philbucher (approved):

PR #11268, opened 2023-06-13
[Core] Test XML Utilities - fix test in MVSC 2019

ddiezrod:
For some reason. std::setprecision(0) is not working with MSVC 2019 (at least it did not work properly in the two machines I tried with this compiler). Instead of setting number of printing the number without decimal values, it uses the default settings (5 decimals if I remember correctly).
If I use any other value it works correctly. As we do not have license for 2022 version in @KratosMultiphysics/altair I changed the test to use setprecision(1). I hope this is ok with you @sunethwarna


roigcarlo:
Maybe unrelated, but I think in order to use `std::setprecision(0)` and prevent decimals from being printed, you should also use  `**** << std::fixed`

sunethwarna (approved):
@roigcarlo AFAIK you are correct, but in this case I would like to keep the scientific notation for `double` values, because fixing may lose unintentionally the precision.

@ddiezrod The changes are fully ok with me. Sorry for the mess.

ddiezrod:
Thanks @sunethwarna !

Issue #10983, opened 2023-04-10
Container Expression test fails in MPI with 8 cores


ddiezrod:
TestConditionContainerExpression.test_WeightedProduct in test_container_expression.py is failing in MPI with 8 cores. I think the problem rises because there is an empty partition (I found also that you get a segmentation error in openMP if you try to add empty containers). We can skip this test if the number of cores is greater than 7, but I think this should be properly fixed.

Traceback:
```
test_WeightedProduct (test_container_expression.TestConditionContainerExpression) ... ==== backtrace (tid:    179) ====
 0 0x0000000000053573 ucs_debug_print_backtrace()  /var/tmp/ucx-1.10.0/src/ucs/debug/debug.c:656
 1 0x0000000001187aa7 Kratos::BinaryExpression<Kratos::BinaryOperations::Addition>::BinaryExpression()  ???:0
 2 0x0000000001188203 Kratos::BinaryExpression<Kratos::BinaryOperations::Addition>::Create()  ???:0
 3 0x00000000004b7cb2 Kratos::Exception::operator<< <Kratos::SpecializedContainerExpression<Kratos::PointerVectorSet<Kratos::Condition, Kratos::IndexedObject, std::less<unsigned long>, std::equal_to<unsigned long>, Kratos::intrusive_ptr<Kratos::Condition>, std::vector<Kratos::intrusive_ptr<Kratos::Condition>, std::allocator<Kratos::intrusive_ptr<Kratos::Condition> > > >, Kratos::ContainerDataIO<Kratos::ContainerDataIOTags::NonHistorical> > >()  ???:0
 4 0x000000000052ee43 Kratos::VariableUtils::SetNonHistoricalVariableToZero<double, Kratos::PointerVectorSet<Kratos::Node<3ul, Kratos::Dof<double> >, Kratos::IndexedObject, std::less<unsigned long>, std::equal_to<unsigned long>, Kratos::intrusive_ptr<Kratos::Node<3ul, Kratos::Dof<double> > >, std::vector<Kratos::intrusive_ptr<Kratos::Node<3ul, Kratos::Dof<double> > >, std::allocator<Kratos::intrusive_ptr<Kratos::Node<3ul, Kratos::Dof<double> > > > > > >()  ???:0
 5 0x0000000000259703 pybind11::error_already_set::what()  ???:0
 6 0x0000000000093d41 cfunction_call_varargs()  /usr/src/Python-3.8.10/Objects/call.c:743
[...]

PR #11270, opened 2023-06-14
[Core] TestConditionContainerExpression - skip test with more than 7 cores

ddiezrod:
This test fails with more than 7 cores as mentioned here https://github.com/KratosMultiphysics/Kratos/issues/10983 because the condition container is empty, I'm skipping the test with more than 7 cores until a proper fix is done.


loumalouomega (approved):

loumalouomega:
Problems of the first world, my test fail because my computer is too powerful

sunethwarna (approved):

sunethwarna:
I will work on the Fix in the mean time :)

ddiezrod:
that's be great, thanks @sunethwarna !

PR #11271, opened 2023-06-14
[Core] Bugfix for ContainerExpression segfaulting in MPI when some ranks are empty.

### LINKED ISSUE #10983: Container Expression test fails in MPI with 8 cores

sunethwarna:
**📝 Description**
This fixes #10983. 

The failure was happening due to some containers in some ranks not having any entities, hence could not deduce the shape information of the variable types of `Vector` and `Matrix` types. The fix properly communicates the sizes now.

@ddiezrod Could you also please check this?

**🆕 Changelog**
- Fixes #10983 .

### DIFF
--- a/kratos/expression/container_expression.cpp
+++ b/kratos/expression/container_expression.cpp
@@ -255,7 +255,7 @@ void ContainerExpression<TContainerType, TMeshType>::Evaluate(
         << NumberOfEntities
         << ", local container size = " << this->GetContainer().size() << " ].\n";
 
-    CArrayExpressionIO::Output(pBegin, NumberOfEntities * std::accumulate(pShapeBegin, pShapeBegin+ShapeSize, 1, [](const int V1, const int V2) { return V1 * V2; })).Execute(**this->mpExpression);
+    CArrayExpressionIO::Output(pBegin, NumberOfEntities * std::accumulate(pShapeBegin, pShapeBegin+ShapeSize, 1, [](const int V1, const int V2) { return V1 * V2; })).Execute(*this->mpExpression.value());
 
     KRATOS_CATCH("");
 }
@@ -289,14 +289,14 @@ bool ContainerExpression<TContainerType, TMeshType>::HasExpression() const
 template <class TContainerType, MeshType TMeshType>
 const Expression& ContainerExpression<TContainerType, TMeshType>::GetExpression() const
 {
-    return *(*mpExpression);
+    return *mpExpression.value();
 }
 
 
 template <class TContainerType, MeshType TMeshType>
 Expression::ConstPointer ContainerExpression<TContainerType, TMeshType>::pGetExpression() const
 {
-    return *mpExpression;
+    return mpExpression.value();
 }
 
 template <class TContainerType, MeshType TMeshType>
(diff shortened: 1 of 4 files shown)


ddiezrod:
It works with 8 cores now, but I was testing a random number of cores and it seems test gets stuck when running with more than 20 cores.

loumalouomega:
> It works with 8 cores now, but I was testing a random number of cores and it seems test gets stuck when running with more than 20 cores.

image

sunethwarna:
@ddiezrod I think getting slower with large number of cores is always expected in MPI when the problem is a small one. I also tried with 16 cores, it takes lots of time because, most of the ranks has empty containers. :/

matekelemen (approved):

ddiezrod:
Hmm I am not sure. The thing is that with 20 cores it runs more or less fast (less than 1 s), but with more than that it seems it gets stuck (I have to kill it manually after several minutes) so it seems there might be which is not completely ok... It should not happen in real cases though

sunethwarna:
@ddiezrod Just the timing. I merged it the moment you commented :/ . I have checked it with couple of cases, but not with a really big case...  Have you noticed the same reduction in speed in othe mpi test cases where there is communication (MPI Allgatherv is used)?

PR #11274, opened 2023-06-14
[Core] Fix unity build conflict in test

loumalouomega:
**📝 Description**

Overall, the changes in the PR include modifying the function name `CheckExactlyOneIntervalContains` to `CountIntervalsContaining` and updating its return type, as well as replacing the function calls in the test case with the modified function name. Also add a namespace to avoid conflicts with the definition `TOLERANCE`.

**🆕 Changelog**

- [Fix unity build conflict in test](https://github.com/KratosMultiphysics/Kratos/commit/816108204bfb3b9b9d281ee0b0184e328135fe6b)

### DIFF
--- a/kratos/tests/cpp_tests/utilities/test_polynomial_utilities.cpp
+++ b/kratos/tests/cpp_tests/utilities/test_polynomial_utilities.cpp
@@ -19,13 +19,13 @@
 #include "testing/testing.h"
 #include "utilities/polynomial_utilities.h"
 
-namespace Kratos::Testing {
+namespace Kratos::Testing::PolynomialUtilitiesTests {
 
 namespace {
     using Polynomial = PolynomialUtilities::PolynomialType;
     using Interval = PolynomialUtilities::IntervalType;
     using RootIntervals = std::vector<PolynomialUtilities::IntervalType>;
-    constexpr double POLYNOMIAL_TOLERANCE = 1e-9;
+    constexpr double TOLERANCE = 1e-9;
 
     void CheckExactlyOneIntervalContains(const RootIntervals& rIntervals, double Coordinate) {
         std::size_t containing = 0;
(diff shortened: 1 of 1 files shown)


jcotela:
Just to understand the PR: why is this needed? Are we not using unity builds on any of the PR pipelines?

loumalouomega:
It is failing some CI, I don't know why not everywhere, I pulled that
branches and those fail. Luckily it is easily solved with this fix.

philbucher:
I didn't find a failing build, can you link it please?

If this was the issue it should fail consistently 

I am worried sth else is wrong if it fails randomly

loumalouomega:
> I didn't find a failing build, can you link it please?
> 
> If this was the issue it should fail consistently
> 
> I am worried sth else is wrong if it fails randomly

https://github.com/KratosMultiphysics/Kratos/actions/runs/5267610454/jobs/9523087010

jcotela:
The issue in this case comes from a conflict with a `constexpr double TOLERANCE` defined in `test_geometry.h`. If the unity build decides to compile this test together with one of the geometry ones a name clash happens.

To avoid similar issues, @pooyan-dadvand suggests to move the definitions in test_geometry.h to their own namespace, so that they don't pollute `Kratos::Testing` (and I agree with the suggestion) but to fix the issue without further disruption I am proposing to move the tests in test_polynomial_utilities.cpp to their own namespace instead.

If we decide to go ahead and use a dedicated namespace in test_geometry.h, we can just move the tests here back to Kratos::Testing without further disruption.

loumalouomega:
> The issue in this case comes from a conflict with a `constexpr double TOLERANCE` defined in `test_geometry.h`. If the unity build decides to compile this test together with one of the geometry ones a name clash happens.
> 
> To avoid similar issues, @pooyan-dadvand suggests to move the definitions in test_geometry.h to their own namespace, so that they don't pollute `Kratos::Testing` (and I agree with the suggestion) but to fix the issue without further disruption I am proposing to move the tests in test_polynomial_utilities.cpp to their own namespace instead.
> 
> If we decide to go ahead and use a dedicated namespace in test_geometry.h, we can just move the tests here back to Kratos::Testing with a straightforward change of namespace.

Okay, remember that I created this PR, I cannot approve myself

jcotela (approved):
Approving to solve CI issues that are blocking other PR, let's discuss about introducing a namespace for test_geometry.h on a separate PR if we think it is necessary.

PR #11276, opened 2023-06-15
[Core] Bug fix for ContainerExpression hagging in large number of mpi ranks with lots of empty model parts

sunethwarna:
**📝 Description**
@ddiezrod Thanks a lot for spotting this bug in `ContainerExpressions`. #11271 Only fixed the problem in the `Read` part of the `ContainerExpression`. The write part also had a `Synchronize` call wrapped around an if condition which caused the hanging since some communication from some ranks were missing. This introduces proper communication.

We were really lucky that we got the tests passing with upto even 20 ranks. Now I have tested this with **120** ranks (super slow around 6 mins, but not hanging).

**🆕 Changelog**
- Fix hanging issue in `ContainerExpression` for large number of ranks with lots of empty containers.

### DIFF
--- a/kratos/expression/expression_io_utils.h
+++ b/kratos/expression/expression_io_utils.h
@@ -114,9 +114,11 @@ public:
 
         const IndexType number_of_entities = rContainer.size();
 
-        if (number_of_entities > 0) {
+        std::visit([&, number_of_entities](auto pVariable) {
             KRATOS_TRY
-            std::visit([&, number_of_entities](auto pVariable) {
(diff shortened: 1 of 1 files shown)


ddiezrod:
Thanks a lot for the fix @sunethwarna I suspected that something else was going on but I was not familiar with the code, approving!

ddiezrod (approved):

ddiezrod:
You could also remove the skip from the test ;)

PR #11277, opened 2023-06-16
[Core][Mapping] Mapping from Expressions.

matekelemen:
## Description

Add an interface for mapping from `Expression`s. The mapper is still constructed between two `ModelPart`s, but the values to be mapped can now be mutated via `Expression`s before the mapping takes place. Reverse mapping to `Expression`s is not supported (yet). 

An implementation is provided for mappers deriving from `InterpolativeMapper`.


philbucher (changes requested):
No objections, but I need some time to review 

FYI our initial interface was unmodified since 2016, guess we did a good job designing it 😃

matekelemen:
> FYI our initial interface was unmodified since 2016, guess we did a good job designing it smiley

No complaints here; this PR just adds stuff and doesn't modify the existing interface.

(I'm not a fan of using the same object for mapping in both directions, but that's just my preference.)

matekelemen:
I changed the interface a bit so it's a bit more straightforward now.

Mapping now happens completely between `Expression`s and neither the origin `ModelPart`, nor the destination `ModelPart` is mutated. Fetching/assigning data between `ModelPart`s and `Expression`s via [`VariableExpressionIO`](https://github.com/KratosMultiphysics/Kratos/blob/bc3459b242f6b04bcf209f54ba7aeea0c0ed3168/kratos/expression/variable_expression_io.h#L30) but it's up to the user when that happens.

kratos/python/add_mapper_to_python.h
```diff
@@ -145,7 +145,10 @@ void ExposeMapperToPython(pybind11::module& m)
             .def("Map",                 MapWithoutOptionsVector<TSparseSpace, TDenseSpace>)
             .def("Map",                 MapWithOptionsScalar<TSparseSpace, TDenseSpace>)
             .def("Map",                 MapWithOptionsVector<TSparseSpace, TDenseSpace>)
-
+            .def("Map",
+                 py::overload_cast<Expression::Pointer,Flags>(&MapperType::Map),
+                 py::arg("source_expression"),
```
philbucher: Mapper terminology is origin and destination

philbucher (commented):
Sorry I forgot abt this, will do my best to do it this week

kratos/python/add_mapper_to_python.h
```diff
@@ -145,7 +145,10 @@ void ExposeMapperToPython(pybind11::module& m)
             .def("Map",                 MapWithoutOptionsVector<TSparseSpace, TDenseSpace>)
             .def("Map",                 MapWithOptionsScalar<TSparseSpace, TDenseSpace>)
             .def("Map",                 MapWithOptionsVector<TSparseSpace, TDenseSpace>)
-
+            .def("Map",
+                 py::overload_cast<Expression::Pointer,Flags>(&MapperType::Map),
```
philbucher, line 149: could `overload_cast` also be applied to the other functions in this way?

kratos/mappers/mapper.h
```diff
@@ -110,6 +111,10 @@ class Mapper
         const Variable< array_1d<double, 3> >& rDestinationVariable,
         Kratos::Flags MappingOptions) = 0;
 
+    /// @brief Mapping interface for @ref Expression s.
+    virtual Expression::Pointer Map(Expression::Pointer pOriginExpression,
+                                    Kratos::Flags MappingOptions) = 0;
```
philbucher, line 116: this differs from the other interfaces, which take both input and output as argument, and return nothing
Any objections to making it consistent?

applications/MappingApplication/tests/basic_mapper_tests.py
```diff
@@ -75,6 +95,15 @@ def test_Map_non_constant_scalar(self):
         self.mapper.Map(KM.PRESSURE, KM.TEMPERATURE)
         mapper_test_case.CheckHistoricalNonUniformValues(self.interface_model_part_destination, KM.TEMPERATURE, GetFilePath(self._GetFileName("map_scalar")))
 
+    def test_Map_non_constant_scalar_expression(self) -> None:
+        SetHistoricalNonUniformSolutionScalar(self.interface_model_part_origin.Nodes, KM.PRESSURE)
+        source_expression = KM.Expression.NodalExpression(self.interface_model_part_origin)
+        KM.Expression.VariableExpressionIO.Read(source_expression, KM.PRESSURE, True)
```
philbucher, line 101:
```suggestion
        KM.Expression.VariableExpressionIO.Read(origin_expression, KM.PRESSURE, True)
```

please use `origin` and `destination`

applications/MappingApplication/custom_mappers/interpolative_mapper_base.h
```diff
@@ -470,6 +486,80 @@ class KRATOS_API(MAPPING_APPLICATION) InterpolativeMapperBase : public Mapper<TS
...
+        using VectorSubscriptType = decltype(std::declval<typename TSparseSpace::VectorType>()[0]);
+        const auto get_vector_begin = [](InterfaceVectorContainerType& rVector) -> std::optional<ComponentType*> {
+            auto p_begin = std::optional<ComponentType*>();
+            if constexpr (std::is_same_v<VectorSubscriptType,ComponentType*&>) {
+                p_begin = rVector.GetVector()[0];
+            } else if constexpr (std::is_same_v<VectorSubscriptType,ComponentType&>) {
+                p_begin = &rVector.GetVector()[0];
+            }
+            return p_begin;
+        };
```
philbucher, line 524: the unified treatment of ublas and trilinos vectors was the origin of the `InterfaceVectorContainer`. Could this be solved better by moving this part there? Maybe then the nasty tricks would not be required?

applications/MappingApplication/custom_mappers/interpolative_mapper_base.h
```diff
@@ -470,6 +486,80 @@ class KRATOS_API(MAPPING_APPLICATION) InterpolativeMapperBase : public Mapper<TS
...
+
+        ComponentType* p_origin_begin = get_vector_begin(*mpInterfaceVectorContainerOrigin).value();
+
+        // Create an expression for each component of the destination expression
+        // These will be filled one by one and combed at the end.
+        const unsigned stride = pOriginExpression->GetItemComponentCount();
+        std::vector<Expression::Pointer> destination_components;
+        destination_components.reserve(stride);
+
+        KRATOS_ERROR_IF_NOT(stride) << "Cannot map an empty expression";
```
philbucher, line 534: this could be checked at the very beginning of this function no?
Seems clearer to me

philbucher (commented):
Few more questions
- does this also work in MPI? It seems so right?
- please add a brief explanation for this in the readme
- I feel like a better place for integrating the expressions would be the `InterfaceVectorContainer`. I might be wrong, did you consider this?

philbucher:
> (I'm not a fan of using the same object for mapping in both directions though, but that's just my preference.)

this is required as for mapping with the transpose one can reuse the mapping matrix. I.e. in standard FSI one only needs one mapping matrix when using regular mapping for displacements and transpose (aka conservative) for the forces

matekelemen:
> this is required as for mapping with the transpose one can reuse the mapping matrix. I.e. in standard FSI one only needs one mapping matrix when using regular mapping for displacements and transpose (aka conservative) for the forces

Makes sense, but I'd keep the mapping interface one-way and make the matrix a shared object between related mappers. Anyway, as you said the current system has worked for a long time so I won't suggest changing it.

kratos/mappers/mapper.h
matekelemen: I would, but unfortunately [`Expression`](https://github.com/KratosMultiphysics/Kratos/blob/5d3b297575ec0accb88648c14d718a9844c50a32/kratos/expression/expression.h#L38)s don't allow that. @sunethwarna decided to implement them in a way that they can't be mutated after construction.

sunethwarna: :O In this case, if we are to be consistent with the other interface methods, we could have the following (not a big fan, but it is a possibility):
```c++
virtual void Map(LiteralFlatExpression<double>& rOutputExpression,  Expression::Pointer pOriginExpression, Kratos::Flags MappingOptions) = 0;
```

matekelemen: I don't like that either. It will prevent reverse mapping.

matekelemen: One possibility is to pass a mutable reference to a pointer to the destination expression, and reset it to point to a different expression. It's not a great solution either because other pointers pointing to the original expression won't get updated.

In any case, the destination expression would have to be empty, and the current system isn't meant to represent empty expressions. I think the current signature reflects this well.

Another argument for the current signature is that the function differs from the others because it doesn't mutate any of the model parts.

philbucher:
ping @matekelemen any more input needed from my side?
Dont think so, but dont want to hold it up bcs I forgot sth

matekelemen:
> ping @matekelemen any more input needed from my side? Dont think so, but dont want to hold it up bcs I forgot sth

There's just one issue that needs discussing: **the signature**. To recap, the existing interface looks like something like this:
```cpp
void Map(const Variable& origin, const Variable& destination, ... settings);
```

Depending on the settings, either the destination **or** the origin `ModelPart` is mutated.

The problem is that `Expression`s cannot be mutated, so there are 2 options:
1) break the signature style and return the mapped `Expression` (works only for one-way mapping).
```cpp
Expression::Pointer Map(Expression::Pointer origin, ... settings) const;
```
2) keep the signature style and pass references to `Expression` pointers instead (doesn't work in python). For example:
```cpp
void Map(Expression::Pointer& origin, Expression::Pointer& destination, ... settings);
```

I'd go with *1)* because it would be consistent between C++ and python, and because mapping `Expression`s doesn't mutate the origin or destination `ModelPart`s, which is reflected in that signature.

##

@philbucher  let me know what you think.

philbucher:
Seems like I was a blocker here

Do you still need this?

matekelemen:
> Seems like I was a blocker here
> 
> Do you still need this?

Not really. The project I needed this for was iced a while ago, and `Expression`s are about to be changed drastically, so I wouldn't want to maintain this over that ordeal.

I still think that mapping to or from an array would be very useful.

PR #11278, opened 2023-06-16
[Core] Nearest point utilities

pooyan-dadvand:
**📝 Description**
Adding a new NearestPointUtilities class with methods for finding nearest point on line segment and triangle.

These methods should substitute the corresponding methods in geometry which are wrong now. But I would put that in a separate PR as it changes the current (wrong) behavior.

**🆕 Changelog**
A new NearestPointUtilities class
Tests


loumalouomega (dismissed):

loumalouomega:
After corrections, Okay from my side

loumalouomega (dismissed):

loumalouomega (approved):

PR #11281, opened 2023-06-16
[Core] Remove SpecializedContainers

sunethwarna:
**📝 Description**
This PR removes `SpecializedContainers` which was old way of handling `Read` and `Write` functionalities of `ContainerExpressions`. No tests were changed, only file and method renames are present in this PR.

**🆕 Changelog**
- Removes `SpecializedContainers` and its python counter parts.


matekelemen (approved):
👍

PR #11282, opened 2023-06-16
[Core] update test scripts

philbucher:
Some modernizations

Including better handling of the C++-tests verbosity

The `print_xxx` functions from `testing.utilities` should also properly flush the output, i.e. now no more scrambled CI testing outputs ;)

### DIFF
--- a/kratos/python_scripts/testing/run_tests.py
+++ b/kratos/python_scripts/testing/run_tests.py
@@ -125,7 +125,7 @@ class Commander(object):
                         file=sys.stderr)
                     sys.stderr.flush()
 
-    def RunCppTests(self, applications):
+    def RunCppTests(self, applications, verbosity):
         ''' Calls the cpp tests directly
         '''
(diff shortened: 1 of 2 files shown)


kratos/python_scripts/testing/run_tests.py
```diff
@@ -125,7 +125,7 @@ def RunTestSuit(self, application, applicationPath, path, level, verbose, comman
                         file=sys.stderr)
                     sys.stderr.flush()
 
-    def RunCppTests(self, applications):
+    def RunCppTests(self, applications, verbosity):
```
sunethwarna:
```suggestion
    def RunCppTests(self, applications, verbosity = 1):
```
I would make this optional

sunethwarna (dismissed):
Minor comment only.

sunethwarna (approved):
Thanks @philbucher 👍

PR #11283, opened 2023-06-16
[Core] use proper check macros in ParUtils test

philbucher:
floating point values should be compared with the appropriate check macros

### DIFF
--- a/kratos/tests/cpp_tests/utilities/test_parallel_utilities.cpp
+++ b/kratos/tests/cpp_tests/utilities/test_parallel_utilities.cpp
@@ -66,7 +66,7 @@ KRATOS_TEST_CASE_IN_SUITE(BlockPartitioner, KratosCoreFastSuite)
     //error check
     for(auto& item : data_vector)
     {
-        KRATOS_CHECK_EQUAL(item, std::pow(5.0, 0.1));
+        KRATOS_CHECK_DOUBLE_EQUAL(item, std::pow(5.0, 0.1));
     }
 
     //shorter form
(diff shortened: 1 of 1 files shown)


matekelemen:
`KRATOS_CHECK_DOUBLE_EQUAL` will have to be updated then because its current definition makes no sense:
```cpp
#define KRATOS_CHECK_DOUBLE_EQUAL(a,b) KRATOS_CHECK_NEAR(a,b,std::numeric_limits<double>::epsilon())
```
```cpp
#define KRATOS_CHECK_NEAR(a,b, tolerance) if(!(std::abs(a - b) <= tolerance)) KRATOS_ERROR << "Check failed because " << #a << " = " << a << \
" is not near to " << #b << " = " << b << " within the tolerance " << tolerance
```

Comparing against `std::numeric_limits<double>::epsilon()` completely disregards the scale of the inputs as well as any rounding errors in the test value.

philbucher:
> `KRATOS_CHECK_DOUBLE_EQUAL` will have to be updated then because its current definition makes no sense:
> 
> ```c++
> #define KRATOS_CHECK_DOUBLE_EQUAL(a,b) KRATOS_CHECK_NEAR(a,b,std::numeric_limits<double>::epsilon())
> ```
> 
> ```c++
> #define KRATOS_CHECK_NEAR(a,b, tolerance) if(!(std::abs(a - b) <= tolerance)) KRATOS_ERROR << "Check failed because " << #a << " = " << a << \
> " is not near to " << #b << " = " << b << " within the tolerance " << tolerance
> ```
> 
> Comparing against `std::numeric_limits<double>::epsilon()` completely disregards the scale of the inputs as well as any rounding errors in the test value.

could not agree more!

I might push sth smart for this at some point ...

sunethwarna (approved):
> I might push sth smart for this at some point ...

May be we can use the same mechanism as python abs with rel and abs tol ?

PR #11284, opened 2023-06-18
[OptApp] Modify model part utilities

sunethwarna:
**📝 Description**
This PR simplifies the `ModelPartUtiltiies::Union`, `ModelPartUtiltiies::Intersect` methods and removes `ModelPartUtiltiies::Substract`. This allows first to create an empty model part using the `UNION` and `INTERSECT` operations without filling entities so these model parts can be used within solvers to add `SolutionStepVariables` and `Dofs`. Thereafter, `ModelPartUtilities::ExecuteOperationOnModelPart` is called to actually perform the aforementioned operation.

The naming of the model parts are also not cryptic, hence now on model part names can be used to output.

**🆕 Changelog**
- Changes `ModelPartUtiltiies::Union`, `ModelPartUtiltiies::Intersect` methods.
- Removes `ModelPartUtiltiies::Substract` method
- Add tests


applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
```diff
@@ -65,7 +63,7 @@ def CalculateValue(self) -> float:
 
     def CalculateGradient(self, physical_variable_collective_expressions: dict[SupportedSensitivityFieldVariableTypes, KratosOA.CollectiveExpression]) -> None:
         # first merge all the model parts
-        merged_model_part_map = ModelPartUtilities.GetMergedMap(self.model_part, physical_variable_collective_expressions, False)
+        merged_model_part_map = ModelPartUtilities.GetMergedMap(physical_variable_collective_expressions, False)
 
         # now get the intersected model parts
         intersected_model_part_map = ModelPartUtilities.GetIntersectedMap(self.model_part, merged_model_part_map, True)
```
RezaNajian, line 69: From my perspective, these operations are costly and can significantly slow down the optimization iteration, depending on the size and setup. I would suggest performing these operations only once, specifically during the first call of CalculateGradient, and then storing the intersected_model_part_map. This is because the intersected_model_part_map does not change throughout the optimization process. What are your thoughts on this?

RezaNajian (dismissed):
@sunethwarna Thanks for it. There is only one concern please have a look.

applications/OptimizationApplication/python_scripts/responses/linear_strain_energy_response_function.py
sunethwarna: It is already like that. This get intersected map within itself does the checking, if it is not there, then only the intersected model part is created and filled. So from the outside, you can call this method any number of times. As long as the input args are the same, it will only create and fill the model part once.

sunethwarna: This is like that with all the model part operations.

RezaNajian: Great, got it, Thanks.

RezaNajian (approved):

Issue #4795, opened 2019-05-06
[Core] Consistency between Element/Condition and Node clone


loumalouomega:
The clone of the node does not consider new Id:

https://github.com/KratosMultiphysics/Kratos/blob/fbb7be7bcf6c1b6c5c5cd2233488cf382816adcd/kratos/includes/node.h#L328

But the condition and element does:

https://github.com/KratosMultiphysics/Kratos/blob/231a4d1451377b593791cc330ec1a052a1594ea0/kratos/includes/element.h#L233

Maybe we should discuss this inconsistency

loumalouomega:
I think this discussion is still valid

loumalouomega:
> I think this discussion is still valid

@KratosMultiphysics/technical-committee

roigcarlo:
@loumalouomega We are currently considering options to "merge" elements and conditions. As long as the PR is ready we will ping everyone.

In the meantime, I agree with you, its odd that they have different interfaces :S

loumalouomega:
> @loumalouomega We are currently considering options to "merge" elements and conditions. As long as the PR is ready we will ping everyone.
> 
> In the meantime, I agree with you, its odd that they have different interfaces :S

It is nice to unify elements and conditions, as implementation is 99.9% identical. In any case, this discussion is still valid.

pooyan-dadvand:
@KratosMultiphysics/technical-committee don't see any reason for having the Clone method for the node as we don't derive from it. So we propose to remove the Clone of the node (and implicitly we don't have inconsistency anymore)

loumalouomega:
There is a PR, where this is discussed as well

loumalouomega:
> There is a PR, where this is discussed as well

https://github.com/KratosMultiphysics/Kratos/pull/11286

(@RiccardoRossi you contributed recently there)

PR #11286, opened 2023-06-19
[Core] Remove `Clone` from `Node`

loumalouomega:
**📝 Description**

Following the conclusion of https://github.com/KratosMultiphysics/Kratos/issues/4795#issuecomment-1594326929 this PR removes the Clone() function from the Node class in the file node.h. The function was responsible for creating a new instance of the Node class with copied data and properties from the current instance. 

**🆕 Changelog**

- [Remove `Clone` from `Node`](https://github.com/KratosMultiphysics/Kratos/commit/505c739be8b9868b35479076f77c056b4fbfb56d)


sunethwarna:
I think this will not compile. @RezaNajian We have to remove `Node::Clone` usages from places such as `LinearStrainEnergyOptResponse` in OptApp.

loumalouomega:
> I think this will fail in CI. `Node::Clone` usages from places such as `LinearStrainEnergyOptResponse` in OptApp needs to be removed.
> 
> FYI @RezaNajian

I did a fast search and I didn't found any usage, but yes, probably will fail if my fast search is not correct

loumalouomega:
> > I think this will fail in CI. `Node::Clone` usages from places such as `LinearStrainEnergyOptResponse` in OptApp needs to be removed.
> > FYI @RezaNajian
> 
> I did a fast search and I didn't found any usage, but yes, probably will fail if my fast search is not correct

Doing a second search, yes definitively will fail

sunethwarna:
Hmm.... apparently, even in the core, we still use `Node::Clone`... :|

loumalouomega:
> Hmm.... apparently, even in the core, we still use `Node::Clone`... :|

And the copy constructor has been removed... I am not sure we can remove the Clone method

sunethwarna:
The use cases in `OptApp` will be removed in the near future :D .... But, the tests in core need to re-worked with the new Ids and same nodal positions i guess :/

roigcarlo:
I took a look at this and it does not seems we can remove the method easily. I've pushed the changes to the clone is consistent with the ones in element and condition, but I have a few questions regarding this (to be commented here or with @KratosMultiphysics/technical-committee ):

1 - `fast_transfer_between_model_parts_process.cpp`

What is the point of this process? As I understand, copies of nodes, elements, etc... are being made, new Id's are assigned for such entities, and those copies are assigned to the `mrDestinationModelPart`, but then elements and conditions in `mrDestinationModelPart` keep pointing to the old nodes in `mrOriginModelPart` (If I am not mistaken) Shouldn't the elements and conditions on `mDestinationModelPart` point to the newly created nodes? what is then point on cloning the elements if not?

2 - Copy constructors in the hierarchy / function visibility:

I firstly tried to avoid reintroducing the clone by making a explicit constructor where one could pass the `NodalData`. This has several caveats: 
- `Node` copy constructor is deleted.
- `mNodalData` cannot be retrieved from the node
- `NodalData` copy constructor is private, but not deleted and not implemented
- `SolutionStepsNodalData` inside mNodalData has a copy constructor implemented and public

So the questions are: 

- is `Node` copy constructor prohibited why? If we do have cases where we need to use clone that would be the preferred way to go (IMHO)
- is `mNodalData` not exposed for some reason? Do we want to explicitly avoid the user to get to the data? It does not makes much sense to me because we do allow to access the `mSolutionStepData`
- `NodalData` copy constructor is alive for some reason? I would make it private to keep consistence with other implementations or allow `Node` copy, but one private and other public is odd.
- Same with `SolutionStepsNodalData` copy constructor, either we leave them all public or all private.

TL;DR I would remove `Clone` and leave the Copy constructor of `Node`.

loumalouomega:
> I took a look at this and it does not seems we can remove the method easily. I've pushed the changes so the clone is consistent with the ones in element and condition, but I have a few questions regarding this (to be commented here or with @KratosMultiphysics/technical-committee ):
> 
> 1 - `fast_transfer_between_model_parts_process.cpp`
> What is the point of this process? As I understand, copies of nodes, elements, etc... are being made, new Id's are assigned for such entities, and those copies are assigned to the `mrDestinationModelPart`, but then elements and conditions in `mrDestinationModelPart` keep pointing to the old nodes in `mrOriginModelPart` (If I am not mistaken) Shouldn't the elements and conditions on `mDestinationModelPart` point to the newly created nodes? what is the point on cloning the nodes if not?

I should remove/refactor that process, I hink it can be done ina  more efficient manner, it could be a good moment to get rip off of the Clone.

> 2 - Copy constructors in the hierarchy / function visibility:
> 
> I firstly tried to avoid reintroducing the clone by making a explicit constructor where one could pass the `NodalData`. This has several caveats:
> 
>     * `Node` copy constructor is deleted.
> 
>     * `mNodalData` cannot be retrieved from the node
> 
>     * `NodalData` copy constructor is private, but not deleted and not implemented
> 
>     * `SolutionStepsNodalData` inside `mNodalData` has a copy constructor implemented and public
> 
> 
> So the questions are:
> 
>     * is `Node` copy constructor prohibited why? If we do have cases where we need to use clone that would be the preferred way to go (IMHO)
> 
>     * is `mNodalData` not exposed for some reason? Do we want to explicitly avoid the user to get to the data? It does not makes much sense to me because we do allow to access the `mSolutionStepData`
> 
>     * `NodalData` copy constructor is alive for some reason? I would make it private to keep consistence with other implementations or allow `Node` copy, but one private and other public is odd.
> 
>     * Same with `SolutionStepsNodalData` copy constructor, either we leave them all public or all private.

Basically everthing you say, we need the Clone because we can not copy the class.

> TL;DR I would remove `Clone` and leave the Copy constructor of `Node`.

I agree with this.

RiccardoRossi:
guys, if you remove the clone in favour of the copy constructor, please ensure to make it explicit.

loumalouomega:
I am confuse now, we remove the Clone or we replace with the Clone with id as done by @roigcarlo ?

roigcarlo:
The problem is that in a first stage we cannot remove the clone because we don't have control on where it is used ( the examples above are not exhaustive ). We should add the explicit copy constructor, base the current con on that operator, deprecate clone, an eventually remove it.

loumalouomega:
> The problem is that in a first stage we cannot remove the clone because we don't have control on where it is used ( the examples above are not exhaustive ). We should add the explicit copy constructor, base the current con on that operator, deprecate clone, an eventually remove it.

In that case should be Okay?

RiccardoRossi:
Well, the point of Clone was to make explicit the copyright constructor. The explicit keyword was not there pre c++11.

Having said this 
, and in light if the discussion of the last day about creatina nkdes only through the modelpart, i wonder if we should not prohibit it overall...

loumalouomega:
> Well, the point of Clone was to make explicit the copyright constructor. The explicit keyword was not there pre c++11.
> 
> Having said this , and in light if the discussion of the last day about creatina nkdes only through the modelpart, i wonder if we should not prohibit it overall...

Huuuum, nodes are created outside model part everywhere, that would be aclompete crash of the current codebase.

RiccardoRossi:
@KratosMultiphysics/technical-committee understands that "Clone" is a copy of the node. Our impression is that using "Clone" instead of the copy constructor is clearer (for example you can easily search for it in the code).

@pooyan-dadvand takes this over and will speak with @loumalouomega to understand what are the driving need for this proposal.

loumalouomega:
@roigcarlo any idea why Intel is failing?

roigcarlo:
The `FindIntersectedConditionsProcessIntersectionExtendedOBB3D` Segfaulted because an illegal memory access. That's all I can tell :/

loumalouomega:
> The `FindIntersectedConditionsProcessIntersectionExtendedOBB3D` Segfaulted because an illegal memory access. That's all I can tell :/

But but...

loumalouomega:
> The `FindIntersectedConditionsProcessIntersectionExtendedOBB3D` Segfaulted because an illegal memory access. That's all I can tell :/

Now that tehre is not Intel COmpiler should be fine

loumalouomega:
> > The `FindIntersectedConditionsProcessIntersectionExtendedOBB3D` Segfaulted because an illegal memory access. That's all I can tell :/
> 
> Now that tehre is not Intel COmpiler should be fine

Indeed, compiled

roigcarlo (commented):
@KratosMultiphysics/technical-committee After reviewing and checking that the CI works, for me this can go in. 

I leave tomorrow so I would feel better if someone else approve, as I will not be here to take care in case something breaks.

loumalouomega:
> @KratosMultiphysics/technical-committee After reviewing and checking that the CI works, for me this can go in.
> 
> I leave tomorrow so I would feel better if someone else approve, as I will not be here to take care in case something breaks.

Now then?

RiccardoRossi:
@roigcarlo will take final decision about this one.

roigcarlo (approved):
Good to go for me. From here we can proceed to delete the clone now that we have a replacement. ty!

PR #11288, opened 2023-06-19
Let PrintInfo use Info() in process base class.

WPK4FEM:
Info() and PrintInfo() contain repeated text strings that made override in derived classes necessary.

This commit changes the PrintInfo() implementation such that it uses Info(). This avoids overrides in derived classes if only its default behavior is wanted.

### DIFF
--- a/kratos/processes/process.h
+++ b/kratos/processes/process.h
@@ -203,7 +203,7 @@ public:
     /// Print information about this object.
     void PrintInfo(std::ostream& rOStream) const override
     {
-        rOStream << "Process";
+        rOStream << Info();
     }
 
     /// Print object's data.


philbucher (changes requested):
I am pretty sure this does not work reliably across compilers. To my knowledge there is no portable way to get the name of the class
Waiting for comments from @KratosMultiphysics/technical-committee

avdg81:
> I am pretty sure this does not work reliably across compilers. To my knowledge there is no portable way to get the name of the class Waiting for comments from @KratosMultiphysics/technical-committee

I agree that different compilers may yield different results. According to [cppreference.com](https://en.cppreference.com/w/cpp/types/type_info/name):
> Returns an implementation defined null-terminated character string containing the name of the type. No guarantees are given; in particular, the returned string can be identical for several types and change between invocations of the same program.

> Some implementations (such as MSVC, IBM, Oracle) produce a human-readable type name. Others, most notably gcc and clang, return the mangled name, which is specified by the Itanium C++ ABI. [...]

The intent here was to avoid repeating the class name (in `Info` and `PrintInfo`). Depending on how these functions are being used (we assumed debugging is their primary purpose), these implementation differences may be acceptable. We would like to add that the current approach where the name is hard-coded is also not ideal. At present, each and every class that derives from `Process` must override the `Info` and `PrintInfo` member functions, or else they will yield wrong results. Also, when a developer decides to rename a class, he/she must not forget to also update these member functions. That is automatically taken care of when the proposed changes are accepted, which helps to keep derived classes clean, since the base class already provides an adequate implementation (in fact, cleaning up our code is what triggered this change in the first place).

In essence, the question is whether we would like to continue repeating class names (at the cost of more code and the possibility that the information becomes inaccurate at some point in time) or that we would like to have an adequate (but not perfect) solution that works for most practical cases. If there are any usages of `Info` or `PrintInfo` that require certain guarantees that we are unaware of, we would like to learn about them. Thanks.

matekelemen:
compiler-dependent behaviour is bad already, but there's just no point in sending mangled names to the user.

It's not a great solution, but you could just write a test in python that scans a module for classes, and catches those whose type name doesn't match up with what `PrintInfo` returns. The downside is obviously that this only works for classes exposed to python.

Another solution would be to make `Info` a pure virtual function, but that won't help us with tall class hierarchies.

avdg81 (commented):
I feel this is a nice improvement. Could someone from the @KratosMultiphysics/technical-committee please have a look? Thanks.

huhaas (approved):

roigcarlo:
I agree that would be usefull but @philbucher and @matekelemen. Imo (personal opinion, not as tc) if we want to go ahead with this we should an utility of a macro that demangles it for the main compilers (gcc, clang, icc) and extract the name using a trick like using `__PRETTY_FUNCTION__` over a function with a template arguments of the class we want to quiery:

```C++
template<typename TType>
std::string type_name() { return __PRETTY_FUNCTION__; }

std::string demangle(std::string mangled) {
# compilers ifdefs....
}

std::cout << demangle(type_name<Class>) << std::endl;
```

avdg81:
> I agree that would be usefull but @philbucher and @matekelemen are right. Imo (personal opinion, not as tc) if we want to go ahead with this we should use an utility or a macro that demangles it for the main compilers (gcc, clang, icc) and extract the name using a trick like using `__PRETTY_FUNCTION__` over a function with a template arguments of the class we want to query:
> 
> ```c++
> template<typename TType>
> std::string type_name() { return __PRETTY_FUNCTION__; }
> 
> std::string demangle(std::string mangled) {
> # compilers ifdefs....
> }
> 
> std::cout << demangle(type_name<Class>) << std::endl;
> ```

Please note that we have taken out the change to `Info()` since that sparked a lot of debate (which is good by the way). Now we are only proposing to change `PrintInfo()`. The reason is that with that change in place we can remove a considerable amount of code duplication on our end (which we will do in a separate commit).

philbucher (approved):
Seems ok to me now :)

PR #11224, opened 2023-06-02
[Core][MPI] Adding `GeometricalObjectsBinsMPI` class, which provides MPI interface to `GeometricalObjectsBins`

loumalouomega:
**📝 Description**

Following the implementation detailed in https://github.com/KratosMultiphysics/Kratos/pull/11205, this Pull Request introduces `GeometricalObjectsBinsMPI`. This new feature operates similarly to `GeometricalObjectsBins`, but with added compatibility for MPI. Please note that this feature can only be used in conjunction with the container from PR #11205.

`GeometricalObjectsBinsMPI` functions by establishing local searches using a `GeometricalObjectsBins` within each partition. Leveraging the containers defined in PR #11205, the solution is then synchronized across all partitions. This approach ensures efficient and consistent data sharing in a distributed computing environment.

**TODO**: Implement asynchronous version (I will add it to this PR, but I open the PR now, so I can easily track myself)

**🆕 Changelog**

- Adding `GeometricalObjectsBinsMPI` via `kratos/mpi/spatial_containers/geometrical_objects_bins_mpi.cpp`
- Expose to python in `kratos/mpi/python/add_mpi_search_strategies_to_python.cpp`
- Adding C++ test via `kratos/mpi/spatial_containers/geometrical_objects_bins_mpi.cpp`
- Update python serial test to support MPI test
- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11205
- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11046
- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11223
- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11247
- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11250
- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11290

PR #11290, opened 2023-06-19
[Core] Adding more BB methods to `SearchUtilities`

loumalouomega:
**📝 Description**

Part of https://github.com/KratosMultiphysics/Kratos/pull/11224

This pull request introduces changes to the `SearchUtilities` class, specifically to the `PointIsInsideBoundingBox` method. The method is responsible for determining whether a given point is inside a bounding box.

The changes include three different versions of the method, each with its own purpose:

1. **PointIsInsideBoundingBox**:
   - Parameters:
     - `const BoundingBox<Point>& rBoundingBox`: The bounding box object containing the maximum and minimum points.
     - `const array_1d<double, 3>& rCoords`: The coordinates of the point to check.
   - Description:
     - The method checks if the `rCoords` point is inside the specified `rBoundingBox` by comparing the coordinates in each direction (x, y, and z) to the maximum and minimum points of the bounding box.
     - If the point is inside the bounding box, `true` is returned; otherwise, `false` is returned.

2. **PointIsInsideBoundingBoxWithTolerance**:
   - Parameters:
     - `const BoundingBox<Point>& rBoundingBox`: The bounding box object containing the maximum and minimum points.
     - `const array_1d<double, 3>& rCoords`: The coordinates of the point to check.
     - `const double Tolerance`: The tolerance value to be applied to the bounding box.
   - Description:
     - This version of the method includes an additional tolerance parameter.
     - The method first applies the tolerance to the maximum and minimum points of the bounding box.
     - After applying the tolerance, the method checks if the `rCoords` point is inside the adjusted bounding box, similar to the previous versions.
     - If the point is inside the bounding box, `true` is returned; otherwise, `false` is returned.

**🆕 Changelog**

- [Adding more BB methods to SearchUtilities](https://github.com/KratosMultiphysics/Kratos/commit/2c124c6142385a47069556d0b71de27fe5c3cc52)

### DIFF
--- a/kratos/utilities/search_utilities.cpp
+++ b/kratos/utilities/search_utilities.cpp
@@ -41,7 +41,7 @@ bool SearchUtilities::PointIsInsideBoundingBox(
 /***********************************************************************************/
 /***********************************************************************************/
 
-bool SearchUtilities::PointIsInsideBoundingBoxWithTolerance(
+bool SearchUtilities::PointIsInsideBoundingBox(
     const BoundingBox<Point>& rBoundingBox,
     const array_1d<double, 3>& rCoords,
     const double Tolerance
--- a/kratos/utilities/search_utilities.cpp
+++ b/kratos/utilities/search_utilities.cpp
@@ -41,22 +41,6 @@ bool SearchUtilities::PointIsInsideBoundingBox(
 /***********************************************************************************/
 /***********************************************************************************/
 
-bool SearchUtilities::PointIsInsideBoundingBox(
-    const BoundingBoxType& rBoundingBox,
-    const array_1d<double, 3>& rCoords
-    )
-{
-    // The Bounding Box should have some tolerance already!
-    if (rCoords[0] < rBoundingBox[0] && rCoords[0] > rBoundingBox[1])           // check x-direction
-        if (rCoords[1] < rBoundingBox[2] && rCoords[1] > rBoundingBox[3])       // check y-direction
-            if (rCoords[2] < rBoundingBox[4] && rCoords[2] > rBoundingBox[5])   // check z-direction
-                return true;
-    return false;
-}
-
-/***********************************************************************************/
-/***********************************************************************************/
-
 bool SearchUtilities::PointIsInsideBoundingBoxWithTolerance(
     const BoundingBox<Point>& rBoundingBox,
     const array_1d<double, 3>& rCoords,
--- a/kratos/utilities/search_utilities.cpp
+++ b/kratos/utilities/search_utilities.cpp
@@ -21,6 +21,72 @@
 namespace Kratos
 {
 
+bool SearchUtilities::PointIsInsideBoundingBox(
+    const BoundingBox<Point>& rBoundingBox,
+    const array_1d<double, 3>& rCoords
+    )
+{
+    // Get the bounding box points
+    const auto& r_max_point = rBoundingBox.GetMaxPoint();
+    const auto& r_min_point = rBoundingBox.GetMinPoint();
+
+    // The Bounding Box check
+    if (rCoords[0] < r_max_point[0] && rCoords[0] > r_min_point[0])           // check x-direction
+        if (rCoords[1] < r_max_point[1] && rCoords[1] > r_min_point[1])       // check y-direction
+            if (rCoords[2] < r_max_point[2] && rCoords[2] > r_min_point[2])   // check z-direction
+                return true;
+    return false;
+}
+
+/***********************************************************************************/
+/***********************************************************************************/
+
+bool SearchUtilities::PointIsInsideBoundingBox(
+    const BoundingBoxType& rBoundingBox,
+    const array_1d<double, 3>& rCoords
+    )
+{
+    // The Bounding Box should have some tolerance already!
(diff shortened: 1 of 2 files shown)


kratos/utilities/search_utilities.h
```diff
@@ -73,22 +74,40 @@ class SearchUtilities
...
+        );
+
+    /**
+     * @brief This method checks if a point is inside any bounding box of the global bounding boxes considering a certain tolerance
+     * @param rBoundingBox The bounding box
+     * @param rCoords The coordinates of the point
+     * @param Tolerance The tolerance
+     * @return True if the point is inside the bounding box
+     */
+    static bool PointIsInsideBoundingBoxWithTolerance(
```
matekelemen: the function's signature is different and it's pretty straightforward that the extra argument is a tolerance, so this function could just be an overload of `PointIsInsideBoundingBox`. No need to explain the argument in the function's name too.

kratos/utilities/search_utilities.cpp
```diff
@@ -21,6 +21,72 @@
...
+    auto min_point = rBoundingBox.GetMinPoint();
+    
+    // Apply Tolerances (only in non zero BB cases)
+    const double epsilon = std::numeric_limits<double>::epsilon();
+    if (norm_2(max_point) > epsilon && norm_2(min_point) > epsilon) {
+        for (unsigned int i=0; i<3; ++i) {
+            max_point[i] += Tolerance;
+            min_point[i] -= Tolerance;
+        }
+    }
```
matekelemen, line 61: This makes no sense to me. Why is this check any better than just checking the norm of the input point?

Furthermore, the docs read `This method checks if a point is inside any bounding box of the global bounding boxes considering a certain tolerance`, but the tolerance is applied **only if** the bounding box is triple-degenerate and exactly on the origin.

kratos/utilities/search_utilities.h
```diff
@@ -73,22 +74,40 @@ class SearchUtilities
...
-        if (rCoords[0] < rBoundingBox[0] && rCoords[0] > rBoundingBox[1])           // check x-direction
-            if (rCoords[1] < rBoundingBox[2] && rCoords[1] > rBoundingBox[3])       // check y-direction
-                if (rCoords[2] < rBoundingBox[4] && rCoords[2] > rBoundingBox[5])   // check z-direction
-                    return true;
-        return false;
-    }
+        );
+
+    /**
+     * @brief This method checks if a point is inside any bounding box of the global bounding boxes considering a certain tolerance
```
matekelemen: what's a _global bounding box_?

matekelemen (commented):
this check is almost identical to what `IntervalUtility` does, except it checks 3 axes instead of 1. I suggest creating an implementation that both objects can use to reduce code duplication.

kratos/utilities/search_utilities.h
loumalouomega: huuuuum, true. I will think about changing it

kratos/utilities/search_utilities.h
loumalouomega: My bad, corrected

kratos/utilities/search_utilities.cpp
loumalouomega: Corrected the doc

kratos/utilities/search_utilities.h
loumalouomega: Done

kratos/utilities/search_utilities.cpp
```diff
@@ -21,6 +21,56 @@
...
+        if (rCoords[1] < max_point[1] && rCoords[1] > min_point[1])       // check y-direction
+            if (rCoords[2] < max_point[2] && rCoords[2] > min_point[2])   // check z-direction
+                return true;
+    return false;
+}
+
+/***********************************************************************************/
+/***********************************************************************************/
+
 void SearchUtilities::ComputeBoundingBoxesWithTolerance(
```
philbucher, line 74: changing this fct name (and the usages in the mappingapp) would be awesome, but I am also fine if you do it in a separate PR
:pray:

kratos/utilities/search_utilities.cpp
```diff
@@ -21,6 +21,56 @@
...
+{
+    // Get the bounding box points
+    const auto& r_max_point = rBoundingBox.GetMaxPoint();
+    const auto& r_min_point = rBoundingBox.GetMinPoint();
+
+    // The Bounding Box check
+    if (rCoords[0] < r_max_point[0] && rCoords[0] > r_min_point[0])           // check x-direction
+        if (rCoords[1] < r_max_point[1] && rCoords[1] > r_min_point[1])       // check y-direction
+            if (rCoords[2] < r_max_point[2] && rCoords[2] > r_min_point[2])   // check z-direction
+                return true;
```
philbucher, line 37: maybe there would be a way to unify this logic, since it seems prone to copy-paste errors :thinking:

philbucher (approved):

kratos/utilities/search_utilities.cpp
loumalouomega: Not done in this PR, I changed the name of a new function

kratos/utilities/search_utilities.cpp
loumalouomega: Yep, we could

PR #11297, opened 2023-06-21
[OptApp] Adding explicit vertex morphing 

sunethwarna:
**📝 Description**
This PR adds explicit vertex morphing (matrix free) features to `OptimizationApplication` for `ContainerExpression`s. Hence this supports `double`, `array_1d<double, 3>`, `array_1d<double, 4>`, `array_1d<double, 6>`, `array_1d<double, 9>`, `Vector` and `Matrix` variables types. This is not **MPI** compatible. Following filters are added.
1. `NodalExplicitVertexMorphingFilter` which is based on the nodal coordinates.
2. `ConditionExplicitVertexMorphingFilter` which is based on the center point coordinates of the conditions.
3. `ElementExplicitVertexMorphingFilter` which is based on the center point coordinates of the elements.

The reference values for tests are generated from the matrix free vertex morphing mechanism in `ShapeOptimizationApplication` for all three types of new filters.

These newly introduced filters allows having `ContainerExpression` for filter radius, hence one can have different radius for different entities (adaptive radius). The current tests uses constant radius.

**🆕 Changelog**
- Added `NodalExplicitVertexMorphingFilter`
- Added `ConditionExplicitVertexMorphingFilter`
- Added `ElementExplicitVertexMorphingFilter`
- Added tests.


applications/OptimizationApplication/custom_utilities/filtering/explicit_vertex_morphing_filter.h
```diff
@@ -0,0 +1,114 @@
...
+#include "entity_point.h"
+#include "filter_function.h"
+
+namespace Kratos {
+
+///@name Kratos Classes
+///@{
+
+template<class TContainerType>
+class KRATOS_API(OPTIMIZATION_APPLICATION) ExplicitVertexMorphingFilter
```
RezaNajian, line 37: Since this is going to be used in filtering any field it is not ExplicitVertexMorphingFilter, it should be named as simply ExplicitFiltering.

applications/OptimizationApplication/custom_utilities/filtering/explicit_vertex_morphing_filter.cpp
```diff
@@ -0,0 +1,311 @@
...
+}
+
+template<class TContainerType>
+void ExplicitVertexMorphingFilter<TContainerType>::ExplicitVertexMorphingFilter::Update()
+{
+    KRATOS_TRY
+
+    BuiltinTimer timer;
+
+    KRATOS_INFO("OptimizationApplication") << "Creating search tree to perform mapping..." << std::endl;
```
RezaNajian: I would add ExplicitFilter::Update to KRATOS_INFO and ExplicitFilter to all Infos in this class.

applications/OptimizationApplication/custom_utilities/filtering/explicit_vertex_morphing_filter.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+        const IndexType current_data_begin = Index * stride;
+
+        for (IndexType j = 0; j < stride; ++j) {
+            double& current_index_value = *(p_expression->begin() + current_data_begin + j);
+            current_index_value = 0.0;
+
+            for(IndexType neighbour_index = 0 ; neighbour_index < number_of_neighbors; ++neighbour_index) {
+                const IndexType neighbour_id = rTLS.mNeighbourEntityPoints[neighbour_index]->Id();
+                const double weight = list_of_weights[neighbour_index] / sum_of_weights;
```
RezaNajian, line 287: You need to include the area for surface field filtering and the volume for the volume field filtering. So the weight of each node is multiplied by area/volume and also considered in the sum of weights. In this way, we can make sure that the filtering of a discrete field is mesh independent.

RezaNajian: Please also implement FilterIntegratedField which only considers integration in the sum of the weights because the given field is already integrated, like the discrete sensitivity of structures and fluids.

applications/OptimizationApplication/custom_utilities/filtering/explicit_vertex_morphing_filter.cpp
```diff
@@ -0,0 +1,311 @@
...
+        }
+    });
+
+    return result;
+
+    KRATOS_CATCH("");
+}
+
+template<class TContainerType>
+ContainerExpression<TContainerType> ExplicitVertexMorphingFilter<TContainerType>::UnFilterField(const ContainerExpression<TContainerType>& rContainerExpression) const
```
RezaNajian: I think here MapBackward is misunderstood as unfiltering a field. Please note that Unfilter field for explicit filtering requires solution of a linear system of equations.

applications/OptimizationApplication/custom_utilities/filtering/explicit_vertex_morphing_filter.cpp
RezaNajian: I would suggest to only have FilterField and FilterIntegratedField, which do the job for forward and backward mapping.

RezaNajian (commented):
@sunethwarna thanks a lot, there are some points from my side.

applications/OptimizationApplication/custom_utilities/filtering/explicit_vertex_morphing_filter.cpp
```diff
@@ -0,0 +1,311 @@
...
+    const IndexType MaxNumberOfNeighbours)
+    : mrModelPart(rModelPart),
+      mMaxNumberOfNeighbors(MaxNumberOfNeighbours)
+{
+    mpKernelFunction = Kratos::make_unique<FilterFunction>(rKernelFunctionType);
+    Update();
+}
+
+template<class TContainerType>
+void ExplicitVertexMorphingFilter<TContainerType>::SetFilterRadius(const ContainerExpression<TContainerType>& rContainerExpression)
```
Igarizza, line 159: @sunethwarna very nice function! Thnx!

applications/OptimizationApplication/custom_utilities/filtering/explicit_vertex_morphing_filter.h
Igarizza: I think ExplicitFiltering is too general as there are a lot of different methods. I will keep VM to explicetly define that VM methodology is used to compute A.

applications/OptimizationApplication/custom_utilities/filtering/explicit_vertex_morphing_filter.cpp
Igarizza: What does UnFilterField mean? A^-1 ? In VM we don't unfilter anything as far as I know ...

Igarizza (commented):
@sunethwarna very nice work. specially thnx for keeping radius as a field. I would use it to implement AVM.

Before merging, let's prepare and run some some actual example to check how does it work. It is hard to judge based on the unittest.

applications/OptimizationApplication/custom_utilities/filtering/explicit_vertex_morphing_filter.h
RezaNajian: @Igarizza As far as I know, based on an extensive literature study I did recently, convolution-based filtering is coined as explicit filtering and it has been used in node-based shape optimization (VM) as well as topology and other field optimizations. As described in the description, this kind of filtering is container agnostic, so can be used for any nodal, elemental, and conditional field.

applications/OptimizationApplication/custom_utilities/filtering/explicit_vertex_morphing_filter.cpp
sunethwarna: Added area contribution and `FilterIntegratedField`

applications/OptimizationApplication/custom_utilities/filtering/explicit_vertex_morphing_filter.cpp
sunethwarna: Removed.

sunethwarna:
This is ready for your review again @RezaNajian @Igarizza.

applications/OptimizationApplication/custom_utilities/filtering/explicit_vertex_morphing_filter.cpp
sunethwarna: done :)

applications/OptimizationApplication/custom_utilities/filtering/explicit_vertex_morphing_filter.h
RezaNajian: @sunethwarna my main concerns are remaining. As the function names and operations are suggesting, it is cleaner, correct and more consistent to name the folders, class simply explicit filtering without VertexMorphing. Its counterpart, implicit filtering is also implemented in a general manner.

applications/OptimizationApplication/custom_utilities/filtering/explicit_vertex_morphing_filter.cpp
```diff
@@ -0,0 +1,342 @@
...
+
+    return p_expression;
+}
+
+template<class TEntityType>
+double GetDomainSize(
+    const EntityPoint<TEntityType>& rPoint,
+    Expression const * const pExpression)
+{
+    return rPoint.GetEntity().GetGeometry().DomainSize();
```
RezaNajian, line 93: I am concerned here. If we have a nodalexpression to filter, then we do node search on the nodes to find the neighbors and weights. Then we need to apply integration, depending the field is integrated or not,  we need to get the NODAL area or volume as the integration weight. However, DomainSize does the job in case the expression is an element entity but what about elemental/conditional expressions?

applications/OptimizationApplication/custom_utilities/filtering/explicit_vertex_morphing_filter.cpp
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    KRATOS_INFO("ExplicitVertexMorphingFilter") << "Search tree created in: " << timer.ElapsedSeconds() << " s" << std::endl;
+
+    KRATOS_CATCH("");
+}
+
+template<class TContainerType>
+template<class TWeightIntegrationType>
+ContainerExpression<TContainerType> ExplicitVertexMorphingFilter<TContainerType>::GenericFilterField(const ContainerExpression<TContainerType>& rContainerExpression) const
```
RezaNajian, line 223: If we have elementalexpression do you do kdtree search on nodes or element centers ?

applications/OptimizationApplication/tests/filtering/explicit_filters_tests.py
```diff
@@ -0,0 +1,215 @@
...
+        cls.model = Kratos.Model()
+        cls.model_part = cls.model.CreateModelPart("test")
+        cls.model_part.AddNodalSolutionStepVariable(Kratos.VELOCITY)
+        cls.model_part.AddNodalSolutionStepVariable(Kratos.ACCELERATION)
+        with kratos_unittest.WorkFolderScope(".", __file__):
+            ReadModelPart("solid", cls.model_part)
+
+        for node in cls.model_part.Nodes:
+            node.SetSolutionStepValue(Kratos.VELOCITY, Kratos.Array3([node.Id, node.Id + 1, node.Id + 2]))
+
```
RezaNajian: could you please add consistency checks, I mean make sure that for a uniform field and an integrated uniform field, we get uniform values. because these are critical checks for arbitrary non-unifom meshes. 
Say we a nodal/elemental uniform 1 field and then we filter it with arbitrary filter sizes, we need to get exactly 1 at each node/element. Also the same for an integrated uniform nodal/elemental field.

RezaNajian (dismissed):

applications/OptimizationApplication/custom_utilities/filtering/explicit_vertex_morphing_filter.cpp
sunethwarna: In this case, Kd tree search is done based on the element centeres (ConitionsExpressions -> Condition centeres)

applications/OptimizationApplication/custom_utilities/filtering/explicit_vertex_morphing_filter.cpp
sunethwarna: In the condition and element case, the domain size is used. In the specialized node case, the representative nodal area/volumes are used.

applications/OptimizationApplication/tests/filtering/explicit_filters_tests.py
sunethwarna: Done :)

RezaNajian:
@sunethwarna shall we merge this ?

RezaNajian (approved):

PR #11298, opened 2023-06-21
[GeomechanicsApplication] added validation for geomechanics solver settings

aronnoordam:
added a check if the scaling solver setting is set on false if rebuild level < 2  or prebuild_dynamics is true

### DIFF
--- a/applications/GeoMechanicsApplication/python_scripts/geomechanics_solver.py
+++ b/applications/GeoMechanicsApplication/python_scripts/geomechanics_solver.py
@@ -145,15 +145,15 @@ class GeoMechanicalSolver(PythonSolver):
 
         # Checks if scaling is used in combination with rebuild level lower than 2 and prebuild dynamics, if so it
         # throws an error
-        if self.settings.Has("linear_solver_settings"):
-            if self.settings["linear_solver_settings"].Has("scaling"):
-                if self.settings["linear_solver_settings"]["scaling"].GetBool():
-                    if self.settings.Has("rebuild_level"):
-                        if self.settings["rebuild_level"].GetInt() < 2:
-                            raise ValueError("Scaling can only be used if rebuild level is 2")
-                    if self.settings.Has("prebuild_dynamics"):
-                        if self.settings["prebuild_dynamics"].GetBool():
-                            raise ValueError("Scaling can not be used if prebuild dynamics is true")
+        if (self.settings.Has("linear_solver_settings") and
+            self.settings["linear_solver_settings"].Has("scaling") and
+            self.settings["linear_solver_settings"]["scaling"].GetBool()):
+            if (self.settings.Has("rebuild_level") and
+                self.settings["rebuild_level"].GetInt() < 2):
+                raise ValueError("Scaling can only be used if rebuild level is 2")
+            if (self.settings.Has("prebuild_dynamics" and
+                self.settings["prebuild_dynamics"].GetBool()):
+                raise ValueError("Scaling can not be used if prebuild dynamics is true")
 
 
 
--- a/applications/GeoMechanicsApplication/python_scripts/geomechanics_solver.py
+++ b/applications/GeoMechanicsApplication/python_scripts/geomechanics_solver.py
@@ -141,6 +141,8 @@ class GeoMechanicalSolver(PythonSolver):
         """This function validates the settings of the solver
         """
 
+        super().ValidateSettings()
+
         # Checks if scaling is used in combination with rebuild level lower than 2 and prebuild dynamics, if so it
         # throws an error
         if self.settings.Has("linear_solver_settings"):
--- a/applications/GeoMechanicsApplication/python_scripts/geomechanics_solver.py
+++ b/applications/GeoMechanicsApplication/python_scripts/geomechanics_solver.py
@@ -144,14 +144,14 @@ class GeoMechanicalSolver(PythonSolver):
         # Checks if scaling is used in combination with rebuild level lower than 2 and prebuild dynamics, if so it
         # throws an error
         if self.settings.Has("linear_solver_settings"):
-            if self.settings["linear_solver_settings"].Has("scaling") and self.settings.Has("rebuild_level"):
-                if self.settings["linear_solver_settings"]["scaling"].GetBool() and \
-                        self.settings["rebuild_level"].GetInt() < 2:
-                    raise ValueError("Scaling can only be used if rebuild level is 2")
-            if self.settings["linear_solver_settings"].Has("scaling") and self.settings.Has("prebuild_dynamics"):
-                if self.settings["linear_solver_settings"]["scaling"].GetBool() and \
-                            self.settings["prebuild_dynamics"].GetBool():
-                    raise ValueError("Scaling can not be used if prebuild dynamics is true")
+            if self.settings["linear_solver_settings"].Has("scaling"):
+                if self.settings["linear_solver_settings"]["scaling"].GetBool():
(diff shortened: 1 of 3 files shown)


applications/GeoMechanicsApplication/python_scripts/geomechanics_solver.py
```diff
@@ -137,6 +137,24 @@ def GetDefaultParameters(cls):
...
+        # throws an error
+        if self.settings.Has("linear_solver_settings"):
+            if self.settings["linear_solver_settings"].Has("scaling") and self.settings.Has("rebuild_level"):
+                if self.settings["linear_solver_settings"]["scaling"].GetBool() and \
+                        self.settings["rebuild_level"].GetInt() < 2:
+                    raise ValueError("Scaling can only be used if rebuild level is 2")
+            if self.settings["linear_solver_settings"].Has("scaling") and self.settings.Has("prebuild_dynamics"):
+                if self.settings["linear_solver_settings"]["scaling"].GetBool() and \
+                            self.settings["prebuild_dynamics"].GetBool():
+                    raise ValueError("Scaling can not be used if prebuild dynamics is true")
```
WPK4FEM: The check on scaling is repeated here, which is avoidable. Something like:
```
if self.settings.Has("linear_solver_settings"):
    if self.settings["linear_solver_settings"].Has("scaling"):
        if self.settings["linear_solver_settings"]["scaling"].GetBool():
            if self.settings.Has("rebuild_level"):
                if self.settings["rebuild_level"].GetInt() < 2:
                    raise ValueError("Scaling can only be used if rebuild level is 2")
            if self.settings.Has("prebuild_dynamics"):
                if self.settings["prebuild_dynamics"].GetBool():
                    raise ValueError("Scaling can not be used if prebuild dynamics is true")
```

mnabideltares: Preferable to have maximum 3 nested if. Can you somehow reduce it?

aronnoordam: thanks for the review, @WPK4FEM and @mnabideltares , these review comments are conflicting. Can you discuss with eachother which is prefered, I will then implement that.

applications/GeoMechanicsApplication/python_scripts/geomechanics_solver.py
```diff
@@ -137,6 +137,26 @@ def GetDefaultParameters(cls):
...
+        # throws an error
+        if self.settings.Has("linear_solver_settings"):
+            if self.settings["linear_solver_settings"].Has("scaling"):
+                if self.settings["linear_solver_settings"]["scaling"].GetBool():
+                    if self.settings.Has("rebuild_level"):
+                        if self.settings["rebuild_level"].GetInt() < 2:
+                            raise ValueError("Scaling can only be used if rebuild level is 2")
+                    if self.settings.Has("prebuild_dynamics"):
+                        if self.settings["prebuild_dynamics"].GetBool():
+                            raise ValueError("Scaling can not be used if prebuild dynamics is true")
```
avdg81: To avoid deep nesting I would suggest to string conditions together by using `and`:
```suggestion
        if (self.settings.Has("linear_solver_settings") and
            self.settings["linear_solver_settings"].Has("scaling") and
            self.settings["linear_solver_settings"]["scaling"].GetBool()):
            if (self.settings.Has("rebuild_level") and
                self.settings["rebuild_level"].GetInt() < 2):
                raise ValueError("Scaling can only be used if rebuild level is 2")
            if (self.settings.Has("prebuild_dynamics" and
                self.settings["prebuild_dynamics"].GetBool()):
                raise ValueError("Scaling can not be used if prebuild dynamics is true")
```

aronnoordam: this wont work, since it can only check self.settings["linear_solver_settings"]["scaling"].GetBool(), if it actually has "linear_solver_settings",

same for the other checks

avdg81: I think it does work. The conditions have been ordered such that they first check that an item exists (using `Has`) before attempting to access the value. If some item does not exist (i.e. `Has` returns `False`) then the other conditions are not checked (and hence that code is not executed). See the [Python reference documentation about `and`](https://docs.python.org/3/reference/expressions.html#booleans):

> The expression `x and y` first evaluates x; if x is false, its value is returned; otherwise, y is evaluated and the resulting value is returned.

aronnoordam: aha, thanks

applications/GeoMechanicsApplication/python_scripts/geomechanics_solver.py
```diff
@@ -21,8 +21,8 @@ def __init__(self, model, custom_settings):
 
         super().__init__(model, custom_settings)
 
-        # # Overwrite the default settings with user-provided parameters.
-        # self.settings.ValidateAndAssignDefaults(default_settings)
+        # Validate settings.
```
avdg81: I think this comment is redundant. The function name (`ValidateSettings`) says it all 😃

applications/GeoMechanicsApplication/python_scripts/geomechanics_solver.py
```diff
@@ -137,6 +137,26 @@ def GetDefaultParameters(cls):
...
+        super().ValidateSettings()
+
+        # Checks if scaling is used in combination with rebuild level lower than 2 and prebuild dynamics, if so it
+        # throws an error
+        if (self.settings.Has("linear_solver_settings") and
+            self.settings["linear_solver_settings"].Has("scaling") and
+            self.settings["linear_solver_settings"]["scaling"].GetBool()):
+            if (self.settings.Has("rebuild_level") and
+                self.settings["rebuild_level"].GetInt() < 2):
+                raise ValueError("Scaling can only be used if rebuild level is 2")
```
avdg81: The error message says that the rebuild level must be set to 2 if scaling is used. But the check suggests that any value greater than 1 will work. So shouldn't the error message say that the rebuild level must be _at least_ equal to 2?

applications/GeoMechanicsApplication/python_scripts/geomechanics_solver.py
```diff
@@ -137,6 +137,26 @@ def GetDefaultParameters(cls):
...
+            if (self.settings.Has("prebuild_dynamics" and
```
avdg81: I think a parenthesis went missing:
```suggestion
            if (self.settings.Has("prebuild_dynamics") and
```

applications/GeoMechanicsApplication/tests/test_geomechanics_solver.py
```diff
@@ -0,0 +1,45 @@
...
+        parameters["linear_solver_settings"]["scaling"].SetBool(True)
+        parameters["rebuild_level"].SetInt(1)
+
+        with self.assertRaises(ValueError) as context:
+            self.create_solver(Kratos.Model(), parameters)
+        self.assertTrue("Scaling can only be used if rebuild level is 2" in str(context.exception))
+
+        # reinitialize parameters
+        parameters = geomechanics_solver.GeoMechanicalSolver.GetDefaultParameters()
+        # set scaling on true and rebuild level on 1, such that an exception is thrown
```
avdg81: This comment appears incorrect to me.

avdg81 (approved):
Thanks for your contribution, Aron.

PR #11299, opened 2023-06-21
[Core] Fix triangle division when one of the nodes has exactly zero distance

jcotela:
**📝 Description**
In the current master DivideTriangle3D3 throws an error if the nodal distance is exactly zero, even though it is actually doing the right thing. This is due to the part where it counts positive and negative nodes, which always counts 0 as positive and reports a change of sign for the negative side triangle.

I assume the same is true for the 2D version, since the proposed change actually involves triangle_2d_3.cpp, but I have not used it.

This PR addresses the issue and introduces a test illustrating the expected behavior.

**🆕 Changelog**
- Fixed issue in DivideTriangle3D3 when one of the nodes has exactly zero distance.

### DIFF
--- a/kratos/tests/cpp_tests/utilities/test_divide_triangle_2d_3.cpp
+++ b/kratos/tests/cpp_tests/utilities/test_divide_triangle_2d_3.cpp
@@ -393,5 +393,42 @@ namespace Kratos
 			KRATOS_CHECK_EQUAL(divider.GetPositiveSubdivisions().size(), 1);
 			KRATOS_CHECK_EQUAL(divider.GetNegativeSubdivisions().size(), 1);
 		}
+
+
+		KRATOS_TEST_CASE_IN_SUITE(DivideGeometryTriangle2D3TwoZeroNodes, KratosCoreFastSuite)
+		{
+			Model current_model;
+
+			// Generate a model part with the previous
+			ModelPart& base_model_part = current_model.CreateModelPart("Triangle");
+
+			// Fill the model part geometry data
+			base_model_part.CreateNewNode(1, 0.0, 0.6, 0.0);
+			base_model_part.CreateNewNode(2, 0.0, 0.9, 0.0);
+			base_model_part.CreateNewNode(3,-0.3, 0.3, 0.0);
+			Properties::Pointer p_properties(new Properties(0));
+			auto p_elem = base_model_part.CreateNewElement("Element2D3N", 1, {1, 2, 3}, p_properties);
+
+			Vector nodal_distances = ZeroVector(3);
+			// cut at y = 0.6, across node 0
+			nodal_distances[0] = 0.0;
+			nodal_distances[1] = 0.0;
+			nodal_distances[2] = -0.3;
+
+			auto divider_one_negative = DivideTriangle2D3<Node>(p_elem->GetGeometry(), nodal_distances);
+			divider_one_negative.GenerateDivision();
+
+			KRATOS_CHECK_IS_FALSE(divider_one_negative.mIsSplit);
+			KRATOS_CHECK_EQUAL(divider_one_negative.mDivisionsNumber, 1);
+			KRATOS_CHECK_EQUAL(divider_one_negative.mSplitEdgesNumber, 0);
+
+			nodal_distances[2] = 0.3;
+			auto divider_one_positive = DivideTriangle2D3<Node>(p_elem->GetGeometry(), nodal_distances);
+			divider_one_positive.GenerateDivision();
+
+			KRATOS_CHECK_IS_FALSE(divider_one_positive.mIsSplit);
+			KRATOS_CHECK_EQUAL(divider_one_positive.mDivisionsNumber, 1);
+			KRATOS_CHECK_EQUAL(divider_one_positive.mSplitEdgesNumber, 0);
+		}
 	}
 }  // namespace Kratos.
--- a/kratos/tests/cpp_tests/utilities/test_divide_triangle_2d_3.cpp
+++ b/kratos/tests/cpp_tests/utilities/test_divide_triangle_2d_3.cpp
@@ -5,7 +5,7 @@
 //                   Multi-Physics
 //
 //  License:     BSD License
-//  			 Kratos default license: kratos/license.txt
+//               Kratos default license: kratos/license.txt
 //
 //  Main authors:    Ruben Zorrilla
 //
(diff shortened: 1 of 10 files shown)


kratos/utilities/divide_triangle_2d_3.cpp
```diff
@@ -160,12 +160,12 @@ namespace Kratos
...
-                if(i0 <= 2) {nodal_distances(i0) < 0.0 ? neg++ : pos++;}
-                if(i1 <= 2) {nodal_distances(i1) < 0.0 ? neg++ : pos++;}
-                if(i2 <= 2) {nodal_distances(i2) < 0.0 ? neg++ : pos++;}
+                if(i0 <= 2) {if(nodal_distances(i0) < 0.0) neg++; else if(nodal_distances(i0) > 0.0) pos++;};
+                if(i1 <= 2) {if(nodal_distances(i1) < 0.0) neg++; else if(nodal_distances(i1) > 0.0) pos++;};
+                if(i2 <= 2) {if(nodal_distances(i2) < 0.0) neg++; else if(nodal_distances(i2) > 0.0) pos++;};
 
                 if(neg > 0 && pos > 0)
-                    KRATOS_ERROR << "The subgeometry " << i0 << " " << i1 << " " << i2 << " in triange has nodes in both positive and negative sides." << std::endl;
+                    KRATOS_ERROR << "The subgeometry " << i0 << " " << i1 << " " << i2 << " in triangle has nodes in both positive and negative sides." << std::endl;
```
loumalouomega: KRATOS_ERROR_IF

rubenzorrilla (commented):
This is a very delicate corner case, which I admittedly haven't had the time to carefully look into. Though the splitting algorithm may work, I think we should test how this splitting pattern behaves when used in combination with the modified shape functions calculation.
Hence, I'd suggest to add the very same test set in there as well. Besides, I think this is also happening in the tetrahedron.

kratos/tests/cpp_tests/utilities/test_divide_triangle_3d_3.cpp
```diff
@@ -269,5 +266,33 @@ KRATOS_TEST_CASE_IN_SUITE(DivideGeometryTriangle3D3, KratosCoreFastSuite)
...
+    nodal_distances[0] = 0.0;
+    nodal_distances[1] = 0.3;
+    nodal_distances[2] = -0.3;
+
+    auto divider = DivideTriangle3D3<Node>(p_elem->GetGeometry(), nodal_distances);
+    divider.GenerateDivision();
+
+    // Should split a single edge and return two triangles
+    KRATOS_CHECK_EQUAL(divider.GetPositiveSubdivisions().size(), 1);
+    KRATOS_CHECK_EQUAL(divider.GetNegativeSubdivisions().size(), 1);
```
rubenzorrilla, line 295: I think this test must be added in the `test_divide_triangle_2d_3.cpp` instead of this one.

jcotela:
I am not using this in the context of enriched shape functions, but only as part of a tool to subdivide triangles in a poor quality mesh that comes from an STL, so what you see here is just a reformulation of a weird case I found as a test and its proposed fix... I guess that in the context of enrichment this may be a non-issue, since otherwise you would have noticed by now (I would think the nudging to prevent bad cuts prevents it?).

All this is to say that I will need some guidance if extended testing is needed, since I am not that sure of what to look for in the context of enrichment.

kratos/tests/cpp_tests/utilities/test_divide_triangle_3d_3.cpp
jcotela: I would add both (it would not be the first time in Kratos that something that works in XY fails in YZ)

rubenzorrilla:
> I am not using this in the context of enriched shape functions, but only as part of a tool to subdivide triangles in a poor quality mesh that comes from an STL, so what you see here is just a reformulation of a weird case I found as a test and its proposed fix... I guess that in the context of enrichment this may be a non-issue, since otherwise you would have noticed by now (I would think the nudging to prevent bad cuts prevents it?).
> 
> All this is to say that I will need some guidance if extended testing is needed, since I am not that sure of what to look for in the context of enrichment.

My point is that even though you aim at using the splitting utilities out of the context of the enrichment, the behavior of the former is crucial for the latter, so I want to be sure that changing what is considered OK to be split doesn't affect the enrichment utils. I don't think we can rely on the ill-conditioned cuts prevention as this is one of the possible alternatives in literature to make the formulation stable.  In this regard, I note that a zero distance value might not imply a "zero"-cut.

Having said this, my suggestion would be to mimic one of the tests in [here](https://github.com/KratosMultiphysics/Kratos/blob/8067b99315faa2a828f277b7febdcbcc246e1729/kratos/tests/cpp_tests/modified_shape_functions/test_triangle_2d_3_modified_shape_functions.cpp) but with the triangle and distance pattern you're using in the test in this PR.

jcotela:
I agree that it is important to keep the enriched element working, but the situation that this PR attempts to fix results in a thrown exception, so it is quite evident when it happens in practice. If you are not seeing simulations fail at random as soon as an element is cut in this way, it seems to me that it is not a common situation.

For the 2D triangle test, I reproduced the 3D configuration for the XY plane in [98985b3](https://github.com/KratosMultiphysics/Kratos/pull/11299/commits/98985b347d2fc3a0553d4025ebfbd5624ec84fee). I think that the only thing missing would be a test using the relevant enriched element for a minimal simulation (or at least to get its linear system contribution), but I am not sure how to do it.

RiccardoRossi:
The enrichment will definitely fail in this context, but as @jcotela suggests it should not be happening as of now since if it was an error would be thrown.

the right fix would be to have the error thrown in the calculation of the enrichment funciton rather than in the splitting utility, so i would agree with   @jcotela about merging this.

RiccardoRossi (dismissed):
to me the argument of @jcotela is valid so i am approving, however in case it is not urgent and u can wait for further feedback from  @rubenzorrilla please do

kratos/tests/cpp_tests/utilities/test_divide_triangle_2d_3.cpp
```diff
@@ -365,5 +365,33 @@ namespace Kratos
 			KRATOS_CHECK_EQUAL(triangle_splitter.mSplitEdgesNumber, 0);
 
 		}
+
+		KRATOS_TEST_CASE_IN_SUITE(DivideGeometryTriangle2D3ZeroNode, KratosCoreFastSuite)
```
loumalouomega: This tests have tabs!

RiccardoRossi:
@jcotela why not merging?

rubenzorrilla:
> @jcotela why not merging?

Well, I'd like to have the modified shape functions test I suggested above before merging this.

jcotela:
@rubenzorrilla which element should the tests use? Simplest thing would be to point me to a test using an element that requires this utility and I can replicate it with the configuration that triggers this error.

rubenzorrilla:
> > I am not using this in the context of enriched shape functions, but only as part of a tool to subdivide triangles in a poor quality mesh that comes from an STL, so what you see here is just a reformulation of a weird case I found as a test and its proposed fix... I guess that in the context of enrichment this may be a non-issue, since otherwise you would have noticed by now (I would think the nudging to prevent bad cuts prevents it?).
> > All this is to say that I will need some guidance if extended testing is needed, since I am not that sure of what to look for in the context of enrichment.
> 
> My point is that even though you aim at using the splitting utilities out of the context of the enrichment, the behavior of the former is crucial for the latter, so I want to be sure that changing what is considered OK to be split doesn't affect the enrichment utils. I don't think we can rely on the ill-conditioned cuts prevention as this is one of the possible alternatives in literature to make the formulation stable. In this regard, I note that a zero distance value might not imply a "zero"-cut.
> 
> Having said this, my suggestion would be to mimic one of the tests in [here](https://github.com/KratosMultiphysics/Kratos/blob/8067b99315faa2a828f277b7febdcbcc246e1729/kratos/tests/cpp_tests/modified_shape_functions/test_triangle_2d_3_modified_shape_functions.cpp) but with the triangle and distance pattern you're using in the test in this PR.

@jcotela I attached a link in my previous post (see the one in "here"). My idea is to test the modified shape functions but with the "new" splitting pattern (i.e. the zero distance).

jcotela:
@rubenzorrilla I have added a test based on the one you provided. To make it pass I had to fix the calculation of the interface too, but I think it is working now.

I also did a second small modification to make sure that triangles with two zero nodes are not incorrectly cut (and added a test). I can do an extra commit with whitespace since changes as @loumalouomega noticed that the test files use tabs, but otherwise I think this is good to go.

kratos/utilities/divide_triangle_2d_3.cpp
```diff
@@ -160,12 +160,11 @@ namespace Kratos
 
                 // Determine if the subdivision is wether in the negative or the positive side
```
rubenzorrilla:
```suggestion
                // Determine if the subdivision is whether in the negative or the positive side
                // Note that zero distance nodes are also identified and stored in here 
```

kratos/utilities/divide_triangle_2d_3.cpp
```diff
@@ -219,7 +218,7 @@ namespace Kratos
 
                     // Check the nodal keys to state which nodes belong to the interface
                     // If the indexed keys is larger or equal to the number of nodes means that they are the auxiliar interface points
```
rubenzorrilla:
```suggestion
                    // If the indexed keys is larger or equal to the number of nodes means that they are the auxiliary interface points
                    // For the zero distance case, the corresponding node is considered as part of the interface
```

kratos/utilities/divide_triangle_2d_3.cpp
```diff
@@ -244,7 +243,7 @@ namespace Kratos
...
```
rubenzorrilla:
```suggestion
                    // If the indexed keys is larger or equal to the number of nodes means that they are the auxiliary interface points
                    // For the zero distance case, the corresponding node is considered as part of the interface
```

rubenzorrilla (commented):
Now it looks good to me. I'll approve if you agree with my documentation suggestions.

However, right now the behavior of the triangle and tetrahedron splitting is inconsistent. I think we should update it accordingly.

jcotela:
No problem regarding the comments (I don't think I touched those, though?)

Regarding the tets, I fully agree. I believe that the same approach used here should be applicable, but it is something that should be done and tested carefully

rubenzorrilla:
> No problem regarding the comments (I don't think I touched those, though?)
> 
> Regarding the tets, I fully agree. I believe that the same approach used here should be applicable, but it is something that should be done and tested carefully

I think that for a zero node distance we'll get 2 lets say positive subtets and a negative one. For two zero node distance case I think we'll get one positive subtet and a negative one. For three or more zero distance nodes there won't be splitting.

Having said this, shall we do it in this same PR or in a future one?

jcotela:
@rubenzorrilla I've gone ahead and applied the same changes to the tetrahedra divider. I think it works (I added a test) but I won't be using it, so I'd encourage you to make sure it is ok.

kratos/utilities/divide_tetrahedra_3d_4.cpp
```diff
@@ -215,12 +214,12 @@ namespace Kratos
                     int node_i_key = r_face[0].Id();
                     int node_j_key = r_face[1].Id();
                     int node_k_key = r_face[2].Id();
-                    
+
                     // Check the nodal keys to state which nodes belong to the interface
                     // If the indexed keys is larger or equal to the number of nodes means that they are the auxiliar interface points
```
rubenzorrilla:
```suggestion
                    // If the indexed keys is larger or equal to the number of nodes means that they are the auxiliary interface points
                    // For the zero distance case, the corresponding node is considered as part of the interface
```

kratos/utilities/divide_tetrahedra_3d_4.cpp
```diff
@@ -243,12 +242,12 @@ namespace Kratos
...
```
rubenzorrilla:
```suggestion
                    // If the indexed keys is larger or equal to the number of nodes means that they are the auxiliary interface points
                    // For the zero distance case, the corresponding node is considered as part of the interface
```

kratos/utilities/divide_tetrahedra_3d_4.h
```diff
@@ -182,6 +185,8 @@ class KRATOS_API(KRATOS_CORE) DivideTetrahedra3D4 : public DivideGeometry<TPoint
     ///@name Private Inquiry
     ///@{
 
+    bool NodeIsInterface(int NodeKey) const;
+
```
rubenzorrilla, line 189: This is admittedly a bit picky, but I'd add this method to the triangle divider as well (it makes the code more self-comprehensive).

kratos/utilities/divide_tetrahedra_3d_4.cpp
```diff
@@ -152,17 +152,17 @@ namespace Kratos
...
-                    this->mAuxPointsContainer(i0), 
-                    this->mAuxPointsContainer(i1), 
+                    this->mAuxPointsContainer(i0),
+                    this->mAuxPointsContainer(i1),
                     this->mAuxPointsContainer(i2),
                     this->mAuxPointsContainer(i3));
-                
-                // Determine if the subdivision is wether in the negative or the positive side                                                                                                 
+
+                // Determine if the subdivision is wether in the negative or the positive side
```
rubenzorrilla:
```suggestion
                // Determine if the subdivision is whether in the negative or the positive side
                // Note that zero distance nodes are also identified and stored in here 
```

rubenzorrilla (commented):
I think that what you did is correct. My unique suggestion would be to add a test for the zero node tetrahedra modified shape functions calculator as you did for the triangles one.

jcotela:
I think this is done now

rubenzorrilla (approved):
Great. Thanks for the patience.

PR #11310, opened 2023-06-22
[OptApp] Improving vtu output process

RezaNajian:
This PR improves output handling, especially for outputting analysis data fields. Furthermore, more options were added to vtu_output_process. Another point is that steps and times of Kratos analyses will no longer be controlled by optimization, treating them as black box analyses.

### DIFF
--- a/applications/OptimizationApplication/python_scripts/execution_policies/kratos_analysis_execution_policy.py
+++ b/applications/OptimizationApplication/python_scripts/execution_policies/kratos_analysis_execution_policy.py
@@ -35,6 +35,7 @@ class KratosAnalysisExecutionPolicy(ExecutionPolicy):
         self.parameters = parameters
         self.optimization_problem = optimization_problem
         self.parameters.ValidateAndAssignDefaults(default_settings)
+        self.parameters["analysis_output_settings"].ValidateAndAssignDefaults(default_settings["analysis_output_settings"])
 
         analysis_module = parameters["analysis_module"].GetString()
         analysis_type = parameters["analysis_type"].GetString()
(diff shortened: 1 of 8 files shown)


applications/OptimizationApplication/python_scripts/processes/optimization_problem_vtu_output_process.py
```diff
@@ -151,16 +167,27 @@ def __init__(self, parameters: Kratos.Parameters, optimization_problem: Optimiza
...
+            self.list_of_components.append(GetComponentHavingDataByFullName(component_name, self.optimization_problem))
+
+        optimization_problem_data = self.optimization_problem.GetProblemDataContainer()
+
+        # check if there are vtu output requests from some other processes
+        if not optimization_problem_data.HasValue("requested_container_expression_outputs"):
+            optimization_problem_data["requested_container_expression_outputs"] = set()
+
+        for component in self.list_of_components:
+            optimization_problem_data["requested_container_expression_outputs"].add(component)
```
sunethwarna: This mechanism can be fully removed once `kratos_analysis_execution_policy` does not need it anymore.

applications/OptimizationApplication/python_scripts/execution_policies/kratos_analysis_execution_policy.py
```diff
@@ -0,0 +1,105 @@
...
+        if analysis_settings["output_processes"].Has("vtk_output"):
+           self._GetOutputData(analysis_settings["output_processes"]["vtk_output"][0]["Parameters"])
+        elif analysis_settings["output_processes"].Has("gid_output"):
+           self._GetOutputData(analysis_settings["output_processes"]["gid_output"][0]["Parameters"])
+
+    def _GetOutputData(self,output_process_params):
+        self.nodal_solution_step_data_variables = output_process_params["nodal_solution_step_data_variables"]
+        self.nodal_data_value_variables = output_process_params["nodal_data_value_variables"]
+        self.element_data_value_variables = output_process_params["element_data_value_variables"]
+        self.condition_data_value_variables = output_process_params["condition_data_value_variables"]
```
sunethwarna: I am sorry, I forgot, This is gonna break in following three cases:
1. The parameters group `vtk_output` and `gid_output` are merely arbitrary names, user can have any name in there, hence this will break if a user specifies a different name.
2. This does not work with the `HDF5`.
3. This does not work with `vtu` output for primal analysis.

Hence, I would remove this mechanism fully, and have something like the following:
```json
{
            "model_part_names"        : [],
            "analysis_module"         : "KratosMultiphysics",
            "analysis_type"           : "",
            "analysis_settings"       : {},
            "analysis_output_settings": {
                     "nodal_solution_step_data_variables": [],
                     "nodal_data_value_variables"        : [],
                     "element_data_value_variables"      : [],
                     "condition_data_value_variables"    : []
            }
}
```

Then we can remove the check also which this process checks whether a vtu output requests this. This is way cleaner.

sunethwarna (commented):
Basically, I am ok, but except for the mechanism of getting required output variables from analysis parameters which can break at any point. I would remove it.

RezaNajian:
@sunethwarna i have updated the analysis_output_settings as you wished, however the changes from your side do not write out the data since the following if is always false
        optimization_problem_data = self.optimization_problem.GetProblemDataContainer()
        if optimization_problem_data.HasValue("requested_container_expression_outputs") and self in optimization_problem_data["requested_container_expression_outputs"]:
            self._OutputAnalysisData()

applications/OptimizationApplication/python_scripts/execution_policies/kratos_analysis_execution_policy.py
```diff
@@ -0,0 +1,114 @@
...
+        self.analysis.time = self.analysis.project_parameters["problem_data"]["start_time"].GetDouble()
+        for model_part in self.model_parts:
+            model_part.ProcessInfo.SetValue(Kratos.STEP, 0)
+            model_part.ProcessInfo.SetValue(Kratos.TIME, 0)
+            model_part.ProcessInfo.SetValue(Kratos.DELTA_TIME, 0)
+        self.analysis.RunSolutionLoop()
+
+        optimization_problem_data = self.optimization_problem.GetProblemDataContainer()
+        if optimization_problem_data.HasValue("requested_container_expression_outputs") and self in optimization_problem_data["requested_container_expression_outputs"]:
+            self._OutputAnalysisData()
```
sunethwarna: We don't need this check anymore ryt? we can check `self.nodal_solution_step_data_variables`, etc variables, and write accordingly which is done in `_OutputAnalysisData`. So If the user gives an empty list, this does not do anything.

applications/OptimizationApplication/python_scripts/processes/optimization_problem_vtu_output_process.py
sunethwarna: This thing is not required anymore, and to be honest, this creates a circular dependency which is not so good.

RezaNajian:
@sunethwarna could you please review this ?!

sunethwarna (approved):
Thanks @RezaNajian :)

PR #11315, opened 2023-06-24
[Core] KratosMultiphysics on MacOS

matekelemen:
## Description

`KratosMultiphysics` on an M1.

The two main headaches of compiling on MacOS are
1) the default `clang` provided by the system lacks `OpenMP`
2) `CMake`, `pybind11`, or our importing scheme isn't consistent with how `@rpath` works on MacOS

1 is solved by installing another `llvm` from `Homebrew`, which seems to be the standard solution among Mac users. Unfortunately, this means that the config script must explicitly set this specific toolchain.

I couldn't solve 2 nicely because I can't find who's messing up the `@rpath`, though our import gymnastics in the root `__init__.py` sure look suspicious. Anyway, I baked the absolute install path into binaries instead of relying on `@rpath`, so `${CMAKE_INSTALL_PREFIX}` must be correctly set at configure time. This shouldn't be a problem in most cases tho, unless someone does `cmake -DCMAKE_INSTALL_PREFIX=some_install_path ... && cmake --build ... --prefix some_completely_different_install_path`.

## Changelog

- embed the full install path in built binaries' `RPath` on MacOS (I just couldn't figure out what the problem with `@rpath` was, so I baked the full path of `${CMAKE_INSTALL_PREFIX}/libs` into every target's `RPath`.)
- update stub generation to consider libraries with `Application` in their names, not just `.cpython` (Another weird thing: only `KratosCore` and the `LinearSolversApplication` have `.cpython` in their python bindings' names. All others just end with `...Application.so`)
- add a build script that helps setting up Kratos from scratch

##

I'm open to guidance and suggestions if there's a MacOS veteran aroun.

##

Edit: a companion PR for `CoSimIO` KratosMultiphysics/CoSimIO#348 is required to get the `CoSimulationApplication` working.

### DIFF
--- a/scripts/mac_build
+++ b/scripts/mac_build
@@ -1,0 +1,246 @@
+#!/bin/bash
+# authors: Máté Kelemen
+# Run this script with the -h flag for info.
+
+# Name of this script
+script_name="$(basename ${BASH_SOURCE[0]})"
+
+# Function for printing usage info
+print_help() {
+    echo "$script_name - Configure, build, and install KratosMultiphysics."
+    echo "-h                    : print this help and exit"
+    echo "-C                    : clean build and install directories, then exit"
+    echo "-b build_path         : path to the build directory (created if it does not exist yet)"
+    echo "-i install_path       : path to the install directory (created if it does not exist yet)"
+    echo "-t build_type         : build type [FullDebug, Debug, Release, RelWithDebInfo] (Default: Release)"
+    echo "-a application_name   : name of the application to build (can be passed repeatedly to add more applications)"
+    echo
+    echo "This script provides a build environment for KratosMultiphysics targeting systems running on Apple Silicon."
+    echo "The interface is minimal, so users seeking more control over the build process are invited to tweak this"
+    echo "script to suit their requirements better."
+    echo
+    echo "By default, Kratos is installed to the site-packages directory of the available python"
+    echo "interpreter. This makes KratosMultiphysics and its applications immediately available from"
+    echo "anywhere on the system without having to append PYTHONPATH, provided that the same interpreter"
+    echo "is used. Note however, that it is recommend to use a virtual python environment to avoid tainting"
+    echo "the system python."
+    echo
+    echo "Build requirements:"
+    echo " - Homebrew"
+    echo " - CMake (can be installed from homebrew via 'brew install cmake')"
+    echo " - LLVM (can be installed from homebrew via 'brew install llvm')"
+    echo " - Boost (can be installed from homebrew via 'brew install boost')"
+    echo
+    echo "Recommended build tools:"
+    echo " - ccache ('brew install ccache')"
+    echo " - ninja ('brew install ccache')"
+    echo 
+    echo "Caveats:"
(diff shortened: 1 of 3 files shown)


scripts/mac_build
```diff
@@ -0,0 +1,246 @@
+#!/bin/bash
```
loumalouomega, line 1: Can you update https://github.com/KratosMultiphysics/Kratos/blob/master/scripts/standard_configure_mac.sh?

matekelemen: Not really, it has a fundamentally different assumption about the host system than `mac_build` :/

`standard_configure_mac.sh` is hard-coded to use the system `clang` and `python`, and requires the user to provide `boost`. I don't know whether these were deliberate decisions, so I don't want to change them.

`mac_build` uses `clang` provided by `llvm` separately installed from `Homebrew`. `boost` is also pulled in from `Homebrew`, and the environment `python` is used.

Essentially, the main difference is that `standard_configure_mac.sh` requires the user to modify their system packages and know what they're doing, while `mac_build` provides a specific way of installing dependencies without touching system packages.

roigcarlo: You are free to change anything related to mac, as I have no clue on the environment of iOs or how packages are managed. If your solution is more convenient for the final user go ahead

roigcarlo (approved):

scripts/mac_build
matekelemen: > You are free to change anything related to mac, as I have no clue on the environment of iOs or how packages are managed. If your solution is more convenient for the final user go ahead

I'm new to macos, so I've no idea how conventient this is for the average/veteran user, but I managed to install Kratos on 2 different kinds of hardware so far, so I'm quite happy with it.

I'll update our `standard_configure_mac.sh` once I'm more confident about what I'm doing.

pasenau:
Hallo @matekelemen ,
i did just compiled kratos on a mac mini (intel) and found a problem with the script `scripts/mac_build `:
When i launched `./mac_build -a FluidDynamicsApplication` it freezes the mac to the point that the SO's watchdog restarts the computer.
After looking into it i found that the script launches as many jobs as it has to compile kratos, more tha 600 jobs, thus collapsing the machine.
At the end of the script there is following code:
```bash
if ! cmake --build "$build_dir" --target install -j; then
    exit 1 
fi  
    
exit 0
```
with the flag '-j' and not specifying the number of jobs, it launches these may compile tasks.

By adding the number of tasks to generate, the problem is solved:
```bash
NUM_PROC=$(sysctl -n machdep.cpu.thread_count)
echo "Using $NUM_PROC jobs to build."
if ! cmake --build "$build_dir" --target install -j $NUM_PROC; then
    exit 1 
fi  
    
exit 0
```

`sysctl -n machdep.cpu.thread_count`  -->  returns the number of threads, 4 in my case, and
`sysctl -n machdep.cpu.core_count`  --> returns the number of cores, 2 in my case.

With this modification i was able to build kratos on macOS Monterey 12.7.1.

Thanks and best

Miguel Pasenau

PR #11320, opened 2023-06-27
[GeoMechanics] Adding dummy class "GeomechanicsAnalysis"

Rbravo555:
This is the **1st of 3** PRs for using the RomApp with the GeoMechanicsApp 

**📝 Description**
In the RomApplication, the analysis stage class is obtained by transforming the name of the file containing it from "snake_case" to "PascalCase". The class GeoMechanicsAnalysis is contained in the file "geomechanics_analysis.py" which get transformed into GeomechanicsAnalysis, without the capital M. 

This PR therefore defines a dummy class with the expected name allowing to use the ROM and the GeoMechanics appps as they are

**🆕 Changelog**
- Added dummy class GeomechanicsAnalysis to geomechanics_analysis.py
- Added GeoMechanicsApp to the Rom's solvers wrapper

### DIFF
--- a/applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
+++ b/applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
@@ -199,12 +199,6 @@ class GeoMechanicsAnalysis(GeoMechanicsAnalysisBase):
             self.OutputSolutionStep()
 
 
-class GeomechanicsAnalysis(GeoMechanicsAnalysis):
-
-    def __init__(self, model, project_parameters):
-        super().__init__(model, project_parameters)
-
-
 if __name__ == '__main__':
     from sys import argv
 
--- a/applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
+++ b/applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
@@ -199,6 +199,11 @@ class GeoMechanicsAnalysis(GeoMechanicsAnalysisBase):
             self.OutputSolutionStep()
 
 
+class GeomechanicsAnalysis(GeoMechanicsAnalysis):
+
+    def __init__(self, model, project_parameters):
+        super().__init__(model, project_parameters)
+
 
 if __name__ == '__main__':
     from sys import argv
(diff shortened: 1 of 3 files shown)


applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
```diff
@@ -199,6 +199,11 @@ def RunSolutionLoop(self):
             self.OutputSolutionStep()
 
 
+class GeomechanicsAnalysis(GeoMechanicsAnalysis):
```
huhaas: We like the suggestion. Our application needs to be updated to the Kratos formatting.

Could you update this PR by creating a new python file with the correct naming `geo_mechanics_analysis.py` and move this wrapper class to it. This should satisfy the ROM application. While providing us with an easier transition to the new expected naming scheme. Do you agree?

huhaas (changes requested):

applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
Rbravo555: After having a short internal discussion, we think the creation of a new script `geo_mechanics_analysis.py ` will not be required, because from our side we can come up with a more robust solution for loading the analysis stage class of the physics. 

It seems to us that the naming of the file as `geomechanics_analysis.py` cannot be considered an innapropriate name. For now, the addition of this dummy class will allow us to effortlessly move forward, and eventually it can be destroyed equally easily, as no other scripts will depend on it.

Regarding PR #11322 in which one function is splitted into two:

> CreateSolver(model, custom_settings) ==> CreateSolver(model, custom_settings) + CreateSolverByParameters(model, custom_settings, parallelism) 

this was done to keep the backward compatibility, and is present on all apps. 

We could arrange to have a discussion about the extend to which the format of the files is fixed, and how we are envisioning to evolve it in the future, which should help you in getting/staying close to the Kratos formatting. @roigcarlo @RiccardoRossi

huhaas:
I have added a commit on this branch. This should work for your use case. Can you please confirm? If yes, we can add this to the master.

Rbravo555:
> I have added a commit on this branch. This should work for your use case. Can you please confirm? If yes, we can add this to the master and rebase the other 2 PR's.

The ROM-GeoMechanics example works well with this change. I have rebased both #11321 and #11322

huhaas (approved):

PR #11321, opened 2023-06-27
[GeoMechanics] Matching Function Names in GeoMechanicsAnalysis

Rbravo555:
depends on #11320

This is the **2nd of 3** PRs for using the RomApp with the GeoMechanicsApp 

**📝 Description**
The RomApp will use all of the features of the GeoMechanicsAnalysis, except for the BuilderAndSolver, which will be set as the RomB&S. For this, the required functions should be named _CreateBuilderAndSolver rather than _ConstructBuilderAndSolver, and there should be a function returning the linear solver.

**🆕 Changelog**
- changed _ConstructBuilderAndSolver(self, param1) ==> _CreateBuilderAndSolver(self)
- added method _GetLinearSolver(self)

PR #11322, opened 2023-06-27
[GeoMechanics] Modifications on geomechanics_solvers_wrapper.py

Rbravo555:
depends on #11320 and #11321

This is the **3rd of 3** PRs for using the RomApp with the GeoMechanicsApp 

**📝 Description**
The mechanism inside the RomApp for setting the RomBuilderAndSolver requires the existance of the CreateSolver and CreateSolverByParameters functions.

Notice that this PR still requires some work because the parameters being passed to the CreateSolverByParameters function are in general not the complete parameters contained in the ProjectParameters.json, but only the "solver_settings". Therefore,  **end_time** is a problematic term because it is not contained in the "solver_settings" but in the "problem_data". What I have done here was to set it to 1.0, but we need to find a more robust alternative. 

**🆕 Changelog**
- splitted operations  CreateSolver(model, custom_settings) ==> CreateSolver(model, custom_settings) + CreateSolverByParameters(model, custom_settings, parallelism)

### DIFF
--- a/applications/GeoMechanicsApplication/python_scripts/geo_mechanics_analysis.py
+++ b/applications/GeoMechanicsApplication/python_scripts/geo_mechanics_analysis.py
@@ -1,0 +1,6 @@
+from KratosMultiphysics.GeoMechanicsApplication import geomechanics_analysis
+
+class GeoMechanicsAnalysis(geomechanics_analysis.GeoMechanicsAnalysis):
+
+    def __init__(self, model, project_parameters):
+        super().__init__(model, project_parameters)
--- a/applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
+++ b/applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
@@ -199,12 +199,6 @@ class GeoMechanicsAnalysis(GeoMechanicsAnalysisBase):
             self.OutputSolutionStep()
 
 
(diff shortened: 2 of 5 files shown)


avdg81:
> depends on #11320 and #11321
> 
> This is the **3rd of 3** PRs for using the RomApp with the GeoMechanicsApp
> 
> **📝 Description** The mechanism inside the RomApp for setting the RomBuilderAndSolver requires the existance of the CreateSolver and CreateSolverByParameters functions.
> 
> Notice that this PR still requires some work because the parameters being passed to the CreateSolver function are in general not the complete parameters contained in the ProjectParameters.json, but only the "solver_settings". Therefore, **end_time** is a problematic term because it is not contained in the "solver_settings" but in the "problem_data". What I have done here was to set it to 1.0, but we need to find a more robust alternative.
> 
> **🆕 Changelog**
> 
>     * splitted operations  CreateSolver(model, custom_settings) ==> CreateSolver(model, custom_settings) + CreateSolverByParameters(model, custom_settings, parallelism)

We will investigate at our end how to accommodate your needs (e.g. to rename files to comply with expected naming conventions). We also need to investigate what our options are with respect to the `end_time` parameter which is at present defined under `"problem_data"`. Therefore, we will make a proposal once we understand these things better. Thanks.

Rbravo555:
Moved PR to #11429

PR #11331, opened 2023-06-28
[Core] Adding `GatherEntitiesFromOtherPartitions` to `GatherModelPartUtility`

loumalouomega:
**📝 Description**

This PR adds `GatherEntitiesFromOtherPartitions`, which allows to easily bring entities (nodes, elements, conditions) from other partitions to the current one. The idea is to be used to bring nodes on the fly in cases where MPC are linking DoFs of nodes distributed in different partitions. But I also extended to other entities like elements and conditions for the sake of generality.

@AlejandroCornejo this will allow to eventually make contact MPI compatible

**🆕 Changelog**

- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11338
- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11339
- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11340
- Add `GatherEntitiesFromOtherPartitions` to  `ParallelFillCommunicator` and `FillCommunicator`
- Expose `GatherEntitiesFromOtherPartitions` to python
- Add C++ test for `GatherEntitiesFromOtherPartitions` for `ParallelFillCommunicator`


kratos/utilities/fill_communicator.cpp
```diff
@@ -39,6 +39,17 @@ void FillCommunicator::Execute()
...
 
+void FillCommunicator::BringEntitiesFromOtherPartitions(
+    const std::map<int, std::vector<std::size_t>>& rNodesToBring,
+    const std::map<int, std::vector<std::size_t>>& rElementsToBring,
+    const std::map<int, std::vector<std::size_t>>& rConditionsToBring,
+    const bool CallExecuteAfterBringingEntities
+    )
+{
+    KRATOS_TRY
+    KRATOS_CATCH("");
```
matekelemen:
```suggestion
```

kratos/includes/fill_communicator.h
```diff
@@ -102,6 +114,22 @@ class KRATOS_API(KRATOS_CORE) FillCommunicator
...
+    /**
+     * @brief Function to bring entities from other partitions
+     * @details This function is intended to bring entities from other partitions. The map indicates the partitions to bring and the vector the entities to bring from each partition. In the current serial case it does nothing.
+     * @note For the parallel implementation see ParallelFillCommunicator.
+     * @param rNodesToBring Nodes to bring from other partitions
+     * @param rElementsToBring Elements to bring from other partitions
+     * @param rConditionsToBring Conditions to bring from other partitions
+     * @param CallExecuteAfterBringingEntities Call Execute after bringing entities
+     */
+    virtual void BringEntitiesFromOtherPartitions(
```
matekelemen: > `BringEntitiesFromOtherPartitions`

sooo, `Gather`? The rest of the signature should be self-explanatory.

loumalouomega: Okay

philbucher:
this is more or less the same as the `GatherModelPartUtility` no (which I think does not work properly)?

Why is it integrated with the FillComm? Why not a separate utility?

loumalouomega:
> this is more or less the same as the `GatherModelPartUtility` no (which I think does not work properly)?

More specific to entities and not whole model parts

> Why is it integrated with the FillComm? Why not a separate utility?

I can move it, but I thought it was a proper place to add it. Sometimes adding more and more utilities makes the codebase very confuse to use.

kratos/includes/fill_communicator.h
matekelemen: I meant replacing the entire name `BringEntitiesFromOtherPartitions` with `Gather`, or is that too short?

loumalouomega: GatherFromOtherPartitions? (to me just Gather is too short)

jcotela:
I am not up to date on this, but last time I touched it (which admittedly was 3-4 years ago), the job of the ParallelFillCommunicator was to compute the communication plan (coloring + interface meshes) for a given modelpart. I would prefer to keep it focused on that job, so if this new capability does not go in that direction I would put it in a dedicated utility.

loumalouomega:
> I am not up to date on this, but last time I touched it (which admittedly was 3-4 years ago), the job of the ParallelFillCommunicator was to compute the communication plan (coloring + interface meshes) for a given modelpart. I would prefer to keep it focused on that job, so if this new capability does not go in that direction I would put it in a dedicated utility.

Okay, moving now

kratos/mpi/utilities/gather_modelpart_utility.h
```diff
@@ -176,6 +241,20 @@ class KRATOS_API(KRATOS_MPI_CORE) GatherModelPartUtility
...
+     * @brief Function to gather entities from other partitions
+     * @details This function is intended to gather entities from other partitions. The map indicates the partitions to gather and the vector the entities to gather from each partition. In the current serial case it does nothing.
+     * @param rModelPart Model part to gather entities from other partitions
+     * @param rEntitiesToBring Entities to gather from other partitions
+     * @param EchoLevel The verbosity level
+     */
+    template <class TObjectType>
+    static void GatherEntityFromOtherPartitions(
+        ModelPart& rModelPart,
+        const std::map<int, std::vector<std::size_t>>& rEntitiesToBring,
```
philbucher, line 254: why map and not unordered map?

Also is it the ids if the entities to bring? The template seems unused, maybe it should be `std::vector<TObjectType>` insetad?

philbucher (commented):
Basic comment regarding a confusing interface, rest looks ok to me

kratos/mpi/utilities/gather_modelpart_utility.h
loumalouomega: Huuum, I need to be ordered to avoid conflicts in the communications

kratos/mpi/utilities/gather_modelpart_utility.cpp
```diff
@@ -182,6 +183,256 @@ template void GatherModelPartUtility::GatherOnMaster(const Variable<array_1d<dou
...
+        buffer << "\nRank " << rank << " has to bring " << entity_name << "s from other partitions:" << std::endl;
+        for (auto bring : rEntitiesToBring) {
+            buffer << "\tFrom rank " << bring.first << ": " << bring.second.size() << " " << entity_name << "s\n\t" << bring.second << std::endl;
+        }
+        buffer << "\nRank " << rank << " has to send " << entity_name << "s from other partitions:" << std::endl;
+        for (auto send : send_entities) {
+            buffer << "\tTo rank " << send.first << ": " << send.second.size() << " " << entity_name << "s\n\t" << send.second << std::endl;
+        }
+        buffer << std::endl;
+        std::cout << buffer.str();
```
sunethwarna: Use of KRATOS_INFO? Would be nice to have the ability to turn them off with logger status as well.

kratos/mpi/utilities/gather_modelpart_utility.h
philbucher: what exactly does need to be ordered?
I mean one could still pass the entities in a different order on different ranks no?

> Also is it the ids if the entities to bring? The template seems unused, maybe it should be std::vector<TObjectType> insetad?

and what about this?

loumalouomega: I will try, I think I tried and it was required

kratos/mpi/utilities/gather_modelpart_utility.cpp
loumalouomega: Done

kratos/mpi/utilities/gather_modelpart_utility.h
loumalouomega: It requires to be ordered, otherwise doesn't work. I need to implement the asynchronous. I will try bring to the data communicator your implementation in the MappingApplication.

loumalouomega: BTW, the template is used, look at thde code. There are constexpr

philbucher: ok

can you then please update the function descriptions to include this info?
I.e. what those maps represent and what restrictions there are

loumalouomega: Done

loumalouomega:
A priori all is solved can can be merged

kratos/mpi/utilities/gather_modelpart_utility.h
```diff
@@ -131,6 +132,70 @@ class KRATOS_API(KRATOS_MPI_CORE) GatherModelPartUtility
...
+     * @param rModelPart Model part to gather entities from other partitions
+     * @param rNodesToBring Nodes to gather from other partitions. It must be an ordered map to ensure the order of the ranks. This is a synchronous MPI implementation, and order matters until asynchronous is implemented.
+     * @param rElementsToBring Elements to gather from other partitions. It must be an ordered map to ensure the order of the ranks. This is a synchronous MPI implementation, and order matters until asynchronous is implemented.
+     * @param rConditionsToBring Conditions to gather from other partitions. It must be an ordered map to ensure the order of the ranks. This is a synchronous MPI implementation, and order matters until asynchronous is implemented.
+     * @param CallExecuteAfterBringingEntities Call Execute after gathering entities
+     * @param EchoLevel The verbosity level
+     */
+    static void GatherEntitiesFromOtherPartitions(
+        ModelPart& rModelPart,
+        const std::map<int, std::vector<std::size_t>>& rNodesToBring,
```
philbucher, line 147: can you please indicate what these maps contain? The Ids or the entities? Or their position in the vector of entities?

philbucher (approved):
thanks!

PR #11334, opened 2023-06-29
[FastPR][Core] Butcher tableau usability

rubenzorrilla:
**📝 Description**
In this PR I'm slightly modifying the Butcher tableau class in order to use `std::vector`'s as return types (instead of stack arrays based on the number of Runge-Kutta steps). This makes the return types unique regardless the version of the table we are using (FE, CD, RK3 or RK4). Note that the performance impact of this is negligible as these arrays are allocated when the table is instantiated, something that occurs once per time step or simulation.

Besides, I also changed a bit the way we return the tableau rows to exploit structured bindings, available since we moved to c++17. This allows to remove the auxiliary `ArraySlice` return type (again this also makes possible to have unique return types among tables). I also added a `GetMatrixRowBegin` and `GetMatrixRowEnd` to help using these in the context of parallel utils as c++17 doesn't support lambda capture of structured bindings (introduced in c++20, see [here](https://burnicki.pl/en/2021/04/19/capture-structured-bindings.html)).

Finally, the reasoning to have unique return types among tables is to be able to use these in combination with the `std::variant` and `std::visit`, which requires unique types by construction.

@EduardGomezEscandell most probably you forget about the technicalities of these, but I'm pinging you in any case :sweat_smile:.


kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -120,12 +122,19 @@ class ButcherTableau
...
      */
 
-    ArraySlice GetMatrixRow(const unsigned int SubStepIndex) const
+    std::tuple<RowType::const_iterator, RowType::const_iterator> GetMatrixRow(const unsigned int SubStepIndex) const
     {
-        return ArraySlice{
-            mA[SubStepIndex - 1].begin(),
-            mA[SubStepIndex - 1].begin() + SubStepIndex
-        };
+        return {mA[SubStepIndex - 1].begin(), mA[SubStepIndex - 1].begin() + SubStepIndex};
```
EduardGomezEscandell:
```suggestion
        return {GetMatrixRowBegin(SubStepIndex), GetMatrixRowEnd(GetMatrixRowEnd)};
```

This avoids duplicate code, should compile to the same code.

kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -120,12 +122,19 @@ class ButcherTableau
...
+    }
+
+    RowType::const_iterator GetMatrixRowBegin(const unsigned int SubStepIndex) const
```
EduardGomezEscandell, line 122:
```suggestion
    RowType::const_iterator GetMatrixRowBegin(const unsigned int SubStepIndex) const noexcept
```

This is not really done in Kratos but I guess we have to start somewhere.
It's going to have negligible performance impact here, but it's more a good practices kinda thing.

kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -120,12 +122,19 @@ class ButcherTableau
...
+    {
+        return mA[SubStepIndex - 1].begin();
+    }
+
+    RowType::const_iterator GetMatrixRowEnd(const unsigned int SubStepIndex) const
```
EduardGomezEscandell, line 128: Same thing about noexcept

kratos/solving_strategies/strategies/explicit_solving_strategy_runge_kutta.h
```diff
@@ -388,7 +389,7 @@ class ExplicitSolvingStrategyRungeKutta : public ExplicitSolvingStrategy<TSparse
                 if (!it_dof->IsFixed()) {
                     const double mass = r_lumped_mass_vector(i_dof);
                     const auto k = row(rIntermediateStepResidualVectors, i_dof);
-                    r_u = r_u_old + (dt / mass) * std::inner_product(alphas.begin, alphas.end, k.begin(), 0.0);
+                    r_u = r_u_old + (dt / mass) * std::inner_product(alphas_begin, alphas_end, k.begin(), 0.0);
```
EduardGomezEscandell, line 392: The comment right below this (Github won't let me select it) is still true but no longer relevant, as we are now forced to use std::inner_product due to the types involved.

EduardGomezEscandell (changes requested):
Minor changes, looks good overall.

kratos/solving_strategies/strategies/butcher_tableau.h
rubenzorrilla: Makes sense :+1: .

kratos/solving_strategies/strategies/butcher_tableau.h
rubenzorrilla: Done.

kratos/solving_strategies/strategies/explicit_solving_strategy_runge_kutta.h
rubenzorrilla: Agree. However, as we're quite used to the boost one, I left it as a reasoning for the usage of the `std` one. Nonetheless, removing it would be fine for me if you consider it misleading.

EduardGomezEscandell: It's fine either way, let's leave it there.

EduardGomezEscandell (dismissed):
Nice!

kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -94,6 +87,8 @@ class ButcherTableau
     ///@name Life Cycle
     ///@{
 
+    /// Destructor.
+    virtual ~ButcherTableau() = default;
```
philbucher, line 90:
```suggestion
```

let the compiler do its job ;)

No need for this to be virtual unless you derive from it

kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -241,15 +243,15 @@ class ButcherTableauForwardEuler : public ButcherTableau<ButcherTableauForwardEu
...
+        B[0] = 1.0;
         return B;
     }
 
     static const BaseType::VectorType GenerateThetasVector()
     {
-        VectorType C;
-        C(0) = 0.0;
+        VectorType C(1);
+        C[0] = 0.0;
```
philbucher:
```suggestion
        VectorType C(1);
```

this gets default constructed!

kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -241,15 +243,15 @@ class ButcherTableauForwardEuler : public ButcherTableau<ButcherTableauForwardEu
 
     static const BaseType::VectorType GenerateWeights()
     {
-        VectorType B;
-        B(0) = 1.0;
+        VectorType B(1);
+        B[0] = 1.0;
```
philbucher:
```suggestion
        VectorType B(1, 1.0);
```

kratos/solving_strategies/strategies/butcher_tableau.h
rubenzorrilla: Agree. However a got a warning in a branch I was working on because of this :confused:

kratos/solving_strategies/strategies/butcher_tableau.h
rubenzorrilla: I know, but I want to make clear which is the coefficient being used.

kratos/solving_strategies/strategies/butcher_tableau.h
rubenzorrilla: Same comment, I found it clearer in the other way (even though it is more verbose). Note that Butcher tableau might be a bit confusing for newbies.

kratos/solving_strategies/strategies/butcher_tableau.h
philbucher: really?
what is the warning?

rubenzorrilla: If I don't remember wrong `class has virtual functions but non-virtual destructor`.

kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -120,12 +115,19 @@ class ButcherTableau
      *          intended to be used with std::inner_product.
      */
 
-    ArraySlice GetMatrixRow(const unsigned int SubStepIndex) const
+    std::tuple<RowType::const_iterator, RowType::const_iterator> GetMatrixRow(const unsigned int SubStepIndex) const
+    {
+        return {GetMatrixRowBegin(SubStepIndex), GetMatrixRowEnd(GetMatrixRowEnd)};
```
roigcarlo:
```suggestion
        return {GetMatrixRowBegin(SubStepIndex), GetMatrixRowEnd(SubStepIndex)};
```

kratos/solving_strategies/strategies/butcher_tableau.h
philbucher: Ah yes that makes sense

Then I would remove the virtual on the functions

kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -120,12 +115,19 @@ class ButcherTableau
...
+        return {GetMatrixRowBegin(SubStepIndex), GetMatrixRowEnd(SubStepIndex)};
+    }
+
+    RowType::const_iterator GetMatrixRowBegin(const unsigned int SubStepIndex) const noexcept
     {
-        return ArraySlice{
-            mA[SubStepIndex - 1].begin(),
-            mA[SubStepIndex - 1].begin() + SubStepIndex
-        };
+        return mA[SubStepIndex - 1].begin();
```
matekelemen, line 125: can `SubStepIndex` be 0? I´d feel a lot safer if there was a (debug) check for that.

kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -120,12 +115,19 @@ class ButcherTableau
...
+    }
+
+    RowType::const_iterator GetMatrixRowEnd(const unsigned int SubStepIndex) const noexcept
+    {
+        return mA[SubStepIndex - 1].begin() + SubStepIndex;
```
matekelemen, line 131: same

kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -135,7 +137,7 @@ class ButcherTableau
 
     constexpr double GetIntegrationTheta(const unsigned int SubStepIndex) const
     {
-        return mC(SubStepIndex - 1);
+        return mC[SubStepIndex - 1];
```
matekelemen, line 142: same

kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -241,15 +243,15 @@ class ButcherTableauForwardEuler : public ButcherTableau<ButcherTableauForwardEu
 
     static const BaseType::VectorType GenerateWeights()
     {
-        VectorType B;
-        B(0) = 1.0;
+        VectorType B(1);
+        B[0] = 1.0;
         return B;
```
matekelemen:
```suggestion
        return VectorType {1.0};
```

kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -241,15 +243,15 @@ class ButcherTableauForwardEuler : public ButcherTableau<ButcherTableauForwardEu
...
     }
 
     static const BaseType::VectorType GenerateThetasVector()
     {
-        VectorType C;
-        C(0) = 0.0;
+        VectorType C(1);
+        C[0] = 0.0;
         return C;
```
matekelemen:
```suggestion
        return VectorType {0.0};
```

kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -272,24 +274,24 @@ class ButcherTableauMidPointMethod : public ButcherTableau<ButcherTableauMidPoin
...
 
     static const BaseType::VectorType GenerateWeights()
     {
-        VectorType B;
-        B(0) = 0.0;
-        B(1) = 1.0;
+        VectorType B(2);
+        B[0] = 0.0;
+        B[1] = 1.0;
         return B;
```
matekelemen:
```suggestion
        return VectorType {0.0, 1.0};
```

kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -272,24 +274,24 @@ class ButcherTableauMidPointMethod : public ButcherTableau<ButcherTableauMidPoin
...
 
     static const BaseType::VectorType GenerateThetasVector()
     {
-        VectorType C;
-        C(0) = 0.0;
-        C(1) = 0.5;
+        VectorType C(2);
+        C[0] = 0.0;
+        C[1] = 0.5;
         return C;
```
matekelemen:
```suggestion
        return VectorType {0.0, 0.5};
```

kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -327,19 +329,19 @@ class ButcherTableauRK3TVD : public ButcherTableau<ButcherTableauRK3TVD, 3, 3>
...
     {
-        VectorType B;
-        B(0) = 1.0 / 6.0;
-        B(1) = 1.0 / 6.0;
-        B(2) = 2.0 / 3.0;
+        VectorType B(3);
+        B[0] = 1.0 / 6.0;
+        B[1] = 1.0 / 6.0;
+        B[2] = 2.0 / 3.0;
         return B;
```
matekelemen:
```suggestion
        return VectorType {1.0 / 6.0,
                           1.0 / 6.0,
                           2.0 / 3.0};
```

kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -327,19 +329,19 @@ class ButcherTableauRK3TVD : public ButcherTableau<ButcherTableauRK3TVD, 3, 3>
...
     {
-        VectorType C;
-        C(0) = 0.0;
-        C(1) = 1.0;
-        C(2) = 0.5;
+        VectorType C(3);
+        C[0] = 0.0;
+        C[1] = 1.0;
+        C[2] = 0.5;
         return C;
```
matekelemen:
```suggestion
        return VectorType {0.0,
                           1.0,
                           0.5};
```

kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -371,21 +372,21 @@ class ButcherTableauRK4 : public ButcherTableau<ButcherTableauRK4, 4, 4>
...
-        B(0) = 1.0 / 6.0;
-        B(1) = 1.0 / 3.0;
-        B(2) = 1.0 / 3.0;
-        B(3) = 1.0 / 6.0;
+        VectorType B(4);
+        B[0] = 1.0 / 6.0;
+        B[1] = 1.0 / 3.0;
+        B[2] = 1.0 / 3.0;
+        B[3] = 1.0 / 6.0;
         return B;
```
matekelemen:
```suggestion
        return VectorType {1.0 / 6.0,
                           1.0 / 3.0,
                           1.0 / 3.0,
                           1.0 / 6.0};
```

kratos/solving_strategies/strategies/butcher_tableau.h
```diff
@@ -371,21 +372,21 @@ class ButcherTableauRK4 : public ButcherTableau<ButcherTableauRK4, 4, 4>
...
-        C(0) = 0.0;
-        C(1) = 0.5;
-        C(2) = 0.5;
-        C(3) = 1.0;
+        VectorType C(4);
+        C[0] = 0.0;
+        C[1] = 0.5;
+        C[2] = 0.5;
+        C[3] = 1.0;
         return C;
```
matekelemen:
```suggestion
        return VectorType {0.0,
                           0.5,
                           0.5,
                           1.0};
```

matekelemen (commented):
My only real concern is a potential integer underflow and resulting segfault, the rest is just style.

kratos/solving_strategies/strategies/butcher_tableau.h
rubenzorrilla: Fair enough. `SubStepIndex` is always larger or equal to 1 as it represents the RK steps. Nonetheless, bad usages of this class can lead to the situation you're pointing to. I'll add it.

kratos/solving_strategies/strategies/butcher_tableau.h
rubenzorrilla: We're deriving so we need to leave it as it is.

kratos/solving_strategies/strategies/butcher_tableau.h
rubenzorrilla: Done.

kratos/solving_strategies/strategies/butcher_tableau.h
rubenzorrilla: Done.

kratos/solving_strategies/strategies/butcher_tableau.h
rubenzorrilla: Done.

philbucher (approved):

PR #11335, opened 2023-06-29
[FastPR][Core] Add GetAnalysisStageFinalData to AnalysisStage

rubenzorrilla:
**📝 Description**
As agreed in #10696, this adds the `GetAnalysisStageFinalData` to the `AnalysisStage`.

### DIFF
--- a/kratos/python_scripts/analysis_stage.py
+++ b/kratos/python_scripts/analysis_stage.py
@@ -126,6 +126,12 @@ class AnalysisStage(object):
         KratosMultiphysics.Logger.PrintInfo(self._GetSimulationName(), "Analysis -END- ")
 
     def GetFinalData(self):
+        """Returns the final data dictionary.
+        
+        The main purpose of this function is to retrieve any data (in a key-value format) from outside the stage.
+        Note that even though it can be called at any point, it is intended to be called at the end of the stage run. 
+        """
+        
         return {}
 
     def InitializeSolutionStep(self):
--- a/kratos/python_scripts/analysis_stage.py
+++ b/kratos/python_scripts/analysis_stage.py
@@ -125,6 +125,9 @@ class AnalysisStage(object):
 
         KratosMultiphysics.Logger.PrintInfo(self._GetSimulationName(), "Analysis -END- ")
 
+    def GetAnalysisStageFinalData(self):
+        return {}
+
     def InitializeSolutionStep(self):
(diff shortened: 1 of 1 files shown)


loumalouomega (dismissed):

kratos/python_scripts/analysis_stage.py
```diff
@@ -125,6 +125,9 @@ def Finalize(self):
 
         KratosMultiphysics.Logger.PrintInfo(self._GetSimulationName(), "Analysis -END- ")
 
+    def GetAnalysisStageFinalData(self):
```
philbucher:
```suggestion
    def GetFinalData(self):
```

analysis_stage.GetAnalyisStageFinalData

Reminds me of condition.Condition_CalculateLocalSystem :sweat_smile:

rubenzorrilla: I was thinking the same :smile:. However this is the name we agreed in the @KratosMultiphysics/technical-committee to make it clear from where the data is retrieved when using the method. We considered that in the multistage context this might not be that obvious. Nonetheless, I'll be fine with changing it.

philbucher: ok, no strong opinion :)

matekelemen: > this is the name we agreed in the https://github.com/orgs/KratosMultiphysics/teams/technical-committee to make it clear from where the data is retrieved when using the method. We considered that in the multistage context this might not be that obvious.

(maybe we should require using type hints in python if that really is a problem)

rubenzorrilla: (wait for the upcoming multistage PR :wink:)

roigcarlo (dismissed):

kratos/python_scripts/analysis_stage.py
```diff
@@ -125,6 +125,9 @@ def Finalize(self):
 
         KratosMultiphysics.Logger.PrintInfo(self._GetSimulationName(), "Analysis -END- ")
 
+    def GetFinalData(self):
+        return {}
```
philbucher: Sorry for the late request, but can you please add some documentation?

For this function the intended purpose/usage would be very helpful!

rubenzorrilla: Done.

philbucher (approved):

Issue #11062, opened 2023-04-27
Compilation Failing


rlrangel:
I have just updated Master branch and I am now getting the following error when compiling it (also shown in image below):

_C:\Users\rrangel\Desktop\Repos\Kratos\kratos\processes\apply_ray_casting_interface_recognition_process.cpp(74,73): error C2946: explicit instantiation; 'Kratos::ApplyRayCastingInterfaceRecognitionProcess<3>::<lambda_ce7114c54fe906ddd2f12c6de41d62e9>' is not a template-class specialization [C:\Users\rrangel\Desktop\Repos\Kratos\build\Debug\kratos\KratosCore.vcxproj]_

_C:\Users\rrangel\Desktop\Repos\Kratos\kratos\processes\apply_ray_casting_process.cpp(565,53): error C2946: explicit instantiation; 'Kratos::ApplyRayCastingProcess<2>::<lambda_5d12caca6125160d56f841004477be35>' is not a template-class specialization [C:\Users\rrangel\Desktop\Repos\Kratos\build\Debug\kratos\KratosCore.vcxproj]_

It happens both in Debug and Release modes with any application that I try to compile.

If I go back to the previous state of two weeks ago, it compiles normally.

I tried to delete Kratos and clone it again, but the error persists.

I'm using Windows 10, Python 3.9, Cmake 3.18, Boost 1.74.

![image](https://user-images.githubusercontent.com/69159206/234843151-c44aa773-bfbe-4f1c-bfcf-1f0c8103322a.png)

ddiezrod:
Hmm its weird that it compiles fine in the rest of the machines , which visual studio version are you using?

ddiezrod:
Even weirder as it the same compiler we use in our CI... I honestly don't know how to debug this without being able to reproduce it. Do you have any clue? @matekelemen @roigcarlo

roigcarlo:
I will take a look

roigcarlo:
Mm I am afraid I can't also reproduce it, but I might have a slight idea about what is happening.

The error seems to be caused by a bug in the compiler. For more info read:
https://developercommunity.visualstudio.com/t/msvc-instantation-issue/1059795
https://jeffpar.github.io/kbarchive/kb/263/Q263633/

Could you try a couple of changes:

At the bottom of the file try to change 

```C++

    template class Kratos::ApplyRayCastingProcess<2>;
    template class Kratos::ApplyRayCastingProcess<3>;

[...]

PR #11345, opened 2023-07-03
Template autoregister alternative

roigcarlo:
**📝 Description**
This should fix new problems in #11062  

After considering suggestions from @matekelemen, this implementation should avoid the usage of template alias, template recursion, and explicitly using instantiations inside the class definition.

The drawback is that registry key is more complex to generate.
Current solution will fail if the type of the template argument is not supported by `std::to_string`. For complex cases it can be implemented, but its a known limitation.

@rlrangel please, try this branch to see if the error is solved.

**🆕 Changelog**
- Re-implemented auto registry of template classes.

### DIFF
--- a/kratos/includes/registry.h
+++ b/kratos/includes/registry.h
@@ -104,6 +104,13 @@ public:
         KRATOS_CATCH("")
     }
 
+    template<typename... Types>
+    static std::string RegistryTemplateToString(Types&&... args) {
+        std::string f_name = (... += ("," + std::to_string(args)));
+        f_name.erase(0,1);
+        return f_name;
+    }
+
     ///@}
     ///@name Access
     ///@{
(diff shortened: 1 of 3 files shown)


kratos/includes/registry.h
```diff
@@ -104,6 +104,13 @@ class KRATOS_API(KRATOS_CORE) Registry final
         KRATOS_CATCH("")
     }
 
+    template<typename... Types>
+    static std::string RegistryTemplateToString(Types&&... args) {
+        std::string f_name = (... += ("," + std::to_string(args)));
+        f_name.erase(0,1);
+        return f_name;
```
matekelemen, line 111: How about this:

```suggestion
        return (std::to_string(args) + ...);
```

https://godbolt.org/z/8bs1n5Gex

matekelemen: ah you need the comma as delimiter, nevermind then 🤦

matekelemen (approved):

kratos/includes/registry.h
roigcarlo: yep sadly... I am not the best out there dealing with strings in C++ :P. It also hurts my eyes, I will change it as soon as I find something better

rlrangel:
> @rlrangel please, try this branch to see if the error is solved.

@roigcarlo Apparently it has solved the problem in my machine.
Thank you very much for that.

PR #11350, opened 2023-07-04
[GeoMechanicsApplication] Fixed several problems found by clang-tidy in DGeoFlow

avdg81:
**🆕 Changelog**
- Removed two unused member functions.
- Improved `const` correctness.
- Removed an intermediate (redundant) local variable.
- Removed some commented-out code.
- Declared several member functions `static`.
- Pass by reference-to-const rather than pass by value.
- Renamed several function parameters to be in line with Kratos' Style Guide.
- Use raw string literals to avoid having to escape certain characters.
- Used the more effective overload to find a single character.
- Avoid copying loop variables when a `const` reference will do.
- Reformatted some function signatures such that each parameter is on its own line. Also lined up the parameter names.
- Removed some narrowing conversions.
- Removed a local variable that shadowed another one.
- Removed an empty statement.
- Replaced `NULL` by `nullptr`.
- Removed a destructor that can be generated by the compiler.
- Removed some unused aliases.
- Marked a function `[[nodiscard]]`.

### DIFF
--- a/applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
+++ b/applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
@@ -38,16 +38,6 @@ public:
     {
         return mr_model_part;
     }
-
-    double GetProcessDoubleValue()
-    {
-        return mdouble_value;
-    }
-
-    void SetProcessDoubleValue(double value)
-    {
-        mdouble_value = value;
-    }
 };
 
 class GeoFlowApplyConstantHydrostaticPressureProcess : public Kratos::ApplyConstantHydrostaticPressureProcess
@@ -99,7 +89,7 @@ namespace Kratos
         this->SetEchoLevel(0);
     }
 
-    int KratosExecute::GetEchoLevel()
+    int KratosExecute::GetEchoLevel() const
     {
         return echoLevel;
     }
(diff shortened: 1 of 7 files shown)


applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -253,61 +237,66 @@ namespace Kratos
...
     }
 
-    int KratosExecute::execute_flow_analysis(std::string workingDirectory, std::string projectName,
-                                             double minCriticalHead, double maxCriticalHead, double stepCriticalHead,
-                                             std::string criticalHeadBoundaryModelPartName,
-                                             std::function<void(char *)> logCallback,
-                                             std::function<void(double)> reportProgress,
-                                             std::function<void(char *)> reportTextualProgress,
-                                             std::function<bool()> shouldCancel)
+    int KratosExecute::ExecuteFlowAnalysis(const std::string&                       rWorkingDirectory,
```
philbucher, line 290: I recommend std::filesystem::path for this purpose

avdg81: Yes, I totally agree. But let's do that in a separate PR. There are many more things that can be improved here, but I prefer to take small steps. Anyway, your suggestion is greatly appreciated. Thanks.

philbucher: Sure, just wanted to point it out in general

WPK4FEM (approved):
Understandable changes towards the style guide and removal of code smells.

PR #11355, opened 2023-07-05
[Core] Project-based multistage

rubenzorrilla:
**📝 Description**
This PR adds the upgraded version of our current multistage capabilities. The enhancements can be summarized into refactoring the code to be much more flexible/extensible and adding checkpointing (i.e. restarting) between stages.

The main changes are the following
- Creation of the (deliberately) pure python `Project` container. This new object is conceived to hold all the simulation data, that is to say the `Model`, the settings, the active stages dictionary and the stages output data dictionary (see #11335). Note that it is also in charge of doing the `Save` and `Load` of the multistage.
- The `multistage_analysis.py` is now the `multistage_orchestrator.py`. We conceived the orchestrators as the class that knows the multistage simulation workflow. Note that this implies knowing at which points we are calling the Project save and load in order to do so.
- Creation of a derived `sequential_multistage_orchestrator.py` that does the standard one-after-each-other (note that this emulates our current `multistage_analysis.py`. On top of that, we added some features to it that we believe could help usability such as the output of the project parameters to be used together with each restarting file.

There are also some minor changes that we require in order to make this work. There are
- Registering the orchestrators as well as the analysis stage. I note that we still support the creation of unregistered stages but throwing a warning.
- Adding a `Save` and `Load` method to current `analysis_stage.py`. This will remain empty by default as the current `SequentialMultistageOrchestrator` does the checkpointing between stages or, in other words, no stage instances are pickled. Nonetheless, this might not be the case in the future or in some forks, so we are providing the interface to achieving so.

As a final note I'd like to add that these changes require some minor modifications in our current multistage `MainKratos.py` and I/O. Having said this, I'll update current examples accordingly once this PR is merged.


philbucher (changes requested):
I have a couple of comments, but only minor, will address them in the next days

Looking good overall, happy to see that this makes progress!

loumalouomega (commented):
Okay from my side. It would be nice to have a test or at least an example in the Example repo

applications/FluidDynamicsApplication/python_registry_lists.py
```diff
@@ -5,4 +5,10 @@
...
     "apply_inlet_process.ApplyInletProcess"
-]
\ No newline at end of file
+]
+
+python_stages_to_be_registered = [
+    "fluid_dynamics_analysis.FluidDynamicsAnalysis"
+]
+
+python_multistage_orchestrators_to_be_registered = []
```
pooyan-dadvand: Maybe not so related to this PR but shall we add the empty lists?

kratos/python_interface/python_registry_utilities.py
```diff
@@ -17,6 +19,14 @@ def RegisterProcessesList(PythonModuleName, PythonRegistryListModule):
     __CheckRegistryListIsInModule(PythonRegistryListModule, "processes")
     __RegisterItemList(PythonModuleName, PythonRegistryListModule.python_processes_to_be_registered, "Processes")
 
+def RegisterStagesList(PythonModuleName, PythonRegistryListModule):
+    __CheckRegistryListIsInModule(PythonRegistryListModule, "stages")
+    __RegisterItemList(PythonModuleName, PythonRegistryListModule.python_stages_to_be_registered, "Stages")
+
+def RegisterMultistageOrchestratorsList(PythonModuleName, PythonRegistryListModule):
```
pooyan-dadvand: Wouldn't be easier to call it Orchestrator? The Multistage (or MultiStage) seems to be redundant here

Apart from that, the SingleStageOrchestrator would be a weird derive class from Multistage.

kratos/python_scripts/analysis_stage.py
```diff
@@ -211,6 +211,20 @@ def ChangeMaterialProperties(self):
         """this function is where the user could change material parameters as a part of the solution step """
         pass
 
+    def Save(self, serializer : KratosMultiphysics.StreamSerializer):
+        """Serializes current analysis stage instance
+
+        This method is intended to make the class pure Python (pickable). This means serialize all the Kratos objects with the given serializer to then destroy them.
```
pooyan-dadvand: Why destroy them?

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
```diff
@@ -0,0 +1,180 @@
+import sys
+import importlib
+
+import KratosMultiphysics
+from KratosMultiphysics.project import Project
+from KratosMultiphysics.model_parameters_factory import KratosModelParametersFactory
+
+class MultistageOrchestrator():
```
pooyan-dadvand: Documentation is missing

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
```diff
@@ -0,0 +1,180 @@
...
+        project -- Kratos project instance
+        """
+
+        # Store pointer to current project
+        # Note that the project already contains the multistage simulation settings
+        self.__project = project
+
+        # Add Kratos version and compilation to settings
+        kratos_version = f"{KratosMultiphysics.KratosGlobals.Kernel.Version()}-{KratosMultiphysics.KratosGlobals.Kernel.BuildType()}"
+        self.__project.GetSettings().AddString("kratos_version", kratos_version)
```
pooyan-dadvand: This should be done in project. It is a project data and shall be initialized there. Orchestrator has no additional info

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
```diff
@@ -0,0 +1,180 @@
...
+        This methods performs the check of the given stage name settings.
+        Note that the Check() method of the stage instance is intentionally not called in here.
+        This is due to the fact that the Check() of the stage instance is to be called within its Initialize() method.
+        """
+
+        if self.__project.GetSettings()["stages"][stage_name].Has("stage_postprocess"):
+            if self.__project.GetSettings()["stages"][stage_name]["stage_postprocess"].Has("modelers"):
+                err_msg = f"Found 'modelers' field in 'stage_postprocess' of stage {stage_name}."
+                err_msg += " Place the 'modelers' section in the next stage 'stage_preprocess'."
+                raise Exception(err_msg)
```
pooyan-dadvand: This shouldn't be in the stage itself?

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
```diff
@@ -0,0 +1,180 @@
...
+        kratos_version = f"{KratosMultiphysics.KratosGlobals.Kernel.Version()}-{KratosMultiphysics.KratosGlobals.Kernel.BuildType()}"
+        self.__project.GetSettings().AddString("kratos_version", kratos_version)
+
+    def Run(self):
+        """Main function that runs the complete multistage simulation."""
+
+        err_msg = "Calling base MultistageOrchestrator Run() method. This must be implemented in derived orchestrators."
+        raise NotImplementedError (err_msg)
+
+    def CheckStageSettings(self, stage_name : str):
```
pooyan-dadvand: I don't get the idea. Why this is not done by the stage itself? 
Instead I would add a method CheckProject() which also has the settings and we may do some sanity checks there.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
```diff
@@ -0,0 +1,180 @@
...
+                err_msg += " Place the 'modelers' section in the next stage 'stage_preprocess'."
+                raise Exception(err_msg)
+
+    def CreateStage(self, stage_name : str):
+        """This method creates a stage instance
+
+        Given a stage name, this method creates the corresponding stage instance from the "analysis_stage" registry entry in the settings.
+        The stage instance is to be created from the registry. If not registered, it tries to use "analysis_stage" as Python module.
+        """
+
```
pooyan-dadvand: I think this should call the factory for the creation and the below code should go to the stage factory.

Nevertheless, the method is useful if we want to do some house keeping before and while creating the stage.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
```diff
@@ -0,0 +1,180 @@
...
+        elif hasattr(analysis_stage_module, "Create"):
+            # If Kratos convention is not fulfilled we search for a Create method
+            stage_instance = analysis_stage_module.Create(self.__project.GetModel(),  KratosMultiphysics.Parameters(self.__project.GetSettings()["stages"][stage_name]))
+        else:
+            err_msg = f"Analysis stage in '{analysis_stage_module_name}' Python module cannot be created. Please check class name or provide a 'Create' method."
+            raise Exception(err_msg)
+        
+        return stage_instance
+
+    def RunCurrentStagePreprocess(self, stage_name : str, data=None):
```
pooyan-dadvand: I like the idea of calling the preprocess from orchestrator. But then we should remove them from the Run of the stages, no?

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
```diff
@@ -0,0 +1,180 @@
...
+                    modeler.PrepareGeometryModel()
+                for modeler in modelers_list:
+                    modeler.SetupModelPart()
+                del modelers_list
+
+            if self.__project.GetSettings()["stages"][stage_name]["stage_preprocess"].Has("operations"):
+                operations_list = self.__CreateListOfOperations("stage_preprocess", stage_name)
+                for operation in operations_list:
+                    operation.Execute()
+                del operations_list
```
pooyan-dadvand: IMO here we should call the stage preprocess and this code should be in the stage itself.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
```diff
@@ -0,0 +1,180 @@
...
+        Keyword arguments:
+        data -- Custom data argument to be used in derived classes
+        """
+
+        if self.__project.GetSettings()["stages"][stage_name].Has("stage_postprocess"):
+            if self.__project.GetSettings()["stages"][stage_name]["stage_postprocess"].Has("operations"):
+                operations_list = self.__CreateListOfOperations("stage_postprocess", stage_name)
+                for operation in operations_list:
+                    operation.Execute()
+                del operations_list
```
pooyan-dadvand: Here again, we should call the method of the stage

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def GetProject(self):
+        """Returns the project."""
+
+        return self.__project
+   
+    def __CreateListOfModelers(self, stage_name : str):
+        """This method creates the modelers at the preprocess execution point."""
```
pooyan-dadvand: Why we are creating the modelers for the stage? Stage should be responsible of its modeler

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
```diff
@@ -0,0 +1,180 @@
...
+
+        execution_point_settings = self.__project.GetSettings()["stages"][stage_name]["stage_preprocess"]
+
+        factory = KratosModelParametersFactory(self.__project.GetModel())
+        return factory.ConstructListOfItems(execution_point_settings["modelers"])
+
+    def __CreateListOfOperations(self, execution_point : str, stage_name : str):
+        """This method creates the operations at any execution point."""
```
pooyan-dadvand: Like the modelers, this should be done in the stage

rubenzorrilla:
@sunethwarna @matekelemen there's a test from the `OptimizationApplication` that crashes. I've investigated a bit and I think that the problem is that you are not considering the possibility of having submodules in `GetClassModuleFromKratos` from `helper_utilities.py`. Can you please do the hotfix for it in a different branch so we can merge it in here?

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
```diff
@@ -0,0 +1,175 @@
...
+
+        # Check if loading from checkpoint is required and load
+        if self.GetProject().GetSettings()["orchestrator"]["settings"].Has("load_from_checkpoint"):
+            if not self.GetProject().GetSettings()["orchestrator"]["settings"]["load_from_checkpoint"].IsNull():
+                if not self.GetProject().GetSettings()["orchestrator"]["settings"]["load_from_checkpoint"].IsString():
+                    err_msg = "'load_from_checkpoint' is expected to be null or a string with the loading point."
+                    raise TypeError(err_msg)
+                else:
+                    loading_point = self.GetProject().GetSettings()["orchestrator"]["settings"]["load_from_checkpoint"].GetString()
+                    self.GetProject().Load(loading_point)
```
pooyan-dadvand: A protected method in the base class would be useful to avoid code duplication in other derived classes

loumalouomega:
> Okay from my side. It would be nice to have a test or at least an example in the Example repo

Speaking with @pooyan-dadvand we think it is important to add a test (with dummy solvers/analysis)

rubenzorrilla:
> > Okay from my side. It would be nice to have a test or at least an example in the Example repo
> 
> Speaking with @pooyan-dadvand we think it is important to add a test (with dummy solvers/analysis)

Sure. I'll add it once we finish the review, just in case we slightly modify the API.

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
```diff
@@ -0,0 +1,175 @@
...
+                self._execution_list = self.GetProject().GetSettings()["orchestrator"]["settings"]["execution_list"].GetStringArray()
+            # If not provided, create an auxiliary execution list from the stages declaration order
+            else:
+                KratosMultiphysics.Logger.PrintInfo("'execution_list' is not provided. Stages will be executed according to their declaration order.")
+                self._execution_list = list(self.GetProject().GetSettings()["stages"].keys())
+
+        return self._execution_list
+
+    def __GetStagesToCheckpointList(self):
+        '''Creates and returns the stages checkpoint list.
```
pooyan-dadvand: If it creates then I would not call it Get...

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
```diff
@@ -0,0 +1,175 @@
...
+    def GetNumberOfStages(self):
+        '''Returns the number of stages.'''
+
+        return len(self.__GetExecutionList())
+
+    def __GetExecutionList(self):
+        '''Creates and returns the execution list.
+        This method creates the execution list, either from a user-defined execution list or,
+        if this is not provided, from the stages declaration order.
+        '''
```
pooyan-dadvand: If it creates I would not call it Get....

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
```diff
@@ -0,0 +1,175 @@
...
+        return self._stages_to_checkpoint_list
+    
+    def __GetStagesCheckpointFolder(self):
+        '''Gets the folder to store the checkpoint files from the settings.'''
+
+        if self.GetProject().GetSettings()["orchestrator"]["settings"].Has("stage_checkpoints_folder"):
+            return self.GetProject().GetSettings()["orchestrator"]["settings"]["stage_checkpoints_folder"].GetString()
+        else:
+            warn_msg = "'stage_checkpoints_folder' is not provided. Creating a default 'checkpoints' one."
+            KratosMultiphysics.Logger.PrintWarning(warn_msg)
```
pooyan-dadvand: I would say this is an info and not a warning. It would be normal to use the default one

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
```diff
@@ -0,0 +1,175 @@
...
+            return 'checkpoints'
+        
+    def __PrepareOutputSettings(self, stage_name : str):
```
pooyan-dadvand: Shouldn't be PrepareCheckPointSettings?

pooyan-dadvand (changes requested):
This is an important feature and I think a very good addition to Kratos!

Apart from comments in the code there are some additional observations:

- Tests are missing
- I would use Control class for the check point

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
rubenzorrilla: mmm... I don't think so. The checkpoint settings are a tiny portion of all the settings we are outputting.

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
rubenzorrilla: It creates the first time is called only (see the ' hasattr` check below). Then I think it is a get.

kratos/python_scripts/analysis_stage.py
rubenzorrilla: If you don't destroy (i.e. assigning them to `None`) the non pure Python object the class wouldn't be pickable.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: Agree. This is a leftover of the version without storing the settings in the project.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: This is checking the multistage part of current stage, that is pre and postprocessing. This cannot be done at the stage level.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: No. It cannot be there as we are not passing the  `stage_postprocess` and `stage_preprocess` settings to the `AnalysisStage`.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: Agree. In the short term I think we should move this piece of code to a stage factory module. I note that this can be used in both the single stage and multistage contexts.

Nonetheless, I'd leave it for a future PR.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: The `Run` of the stages includes neither the preprocess nor the postprocess. In other words, the pre and postprocess only occur at the orchestrator level (i.e. the run of an `AnalysisStage` instance is the very same we used to have).

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: I don't think so. Note that preprocess modelers/operations might depend on previous stages. If introducing them into there you might end up manipulating or removing data from a previous stage from within the next one, something that is in my opinion very counterintuitive.

As it is right now, preprocess and postprocess occur between stages. Note that this doesn't preclude using `Modelers` within a stage as we agreed on #9258 (already implemented).

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: The same observations I did for the preprocess apply also here.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: Note that we give the possibility of having modelers at both the ochestrator and the stage level.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: We designed operations to be executable between stages. If bringing them down to the `AnalysisStage` we'll be killing the possibility of doing stuff between the stages occur (for instance an SVD of previous stage data which might be required for the next stage).

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
rubenzorrilla: I think I'm not getting your point at all. You mean creating a `__GetProject` method instead?

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
rubenzorrilla: (same comment)

applications/FluidDynamicsApplication/python_registry_lists.py
rubenzorrilla: We might check if the attribute is in the module but I have the feeling that having them empty will ease developers in populating the registry. If we put it empty people only needs to fill it with their stuff. Removing it would require them to browse through the code to find which is the list name the registry is expecting.

kratos/python_interface/python_registry_utilities.py
rubenzorrilla: We may rename it but eventually you might have orchestrators for different stuff (e.g. control problems or this sort of things). This is why I put the `Multistage` keyword in front.

About the single stage orchestrator, this is nothing but an instance of the `SequentialMultistageOrchestrator` with a single stage so I don't see the need of implementing it.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: Together with @pooyan-dadvand we've been discussing about this. The misunderstanding comes from the fact that now we have the capability of calling the modellers at the preprocess of the stage and within the stage. We should discuss about this point in the future.

Thanks to this, we realized that current input settings might be a bit misleading. Hence we propose to encapsulate the stage settings as
```json
"fluid_stage" : {
    "name" : "registry_point",
    "stage_preprocess" : {},
    "stage_settings" : {
        "problem_data" : {},
        "solver_setttings" : {},
        ...
    },
    "stage_postprocess" : {}
}
```

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
rubenzorrilla: Together with @pooyan-dadvand we've been discussing about this and we think that it's better to leave it like this as the base class doesn't have `Run` so it doesn't really know when it should restart.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: Done. I did so for the `SequentialMultistageOrchestrator` and the `Project` classes as well.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: Done.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: @pooyan-dadvand I'm resolving this after our discussion. Feel free to unresolve it if required.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: @pooyan-dadvand I'm resolving this after our discussion. Feel free to unresolve it if required.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: @pooyan-dadvand I'm resolving this after our discussion. Feel free to unresolve it if required.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: @pooyan-dadvand I'm resolving this after our discussion. Feel free to unresolve it if required.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: @pooyan-dadvand I'm resolving this after our discussion. Feel free to unresolve it if required.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: @pooyan-dadvand I'm resolving this after our discussion. Feel free to unresolve it if required.

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
rubenzorrilla: Done.

sunethwarna:
OptApp is extended to support `MultistageOrchestrator` in #11371.

kratos/python_scripts/project.py
```diff
@@ -0,0 +1,127 @@
...
+
+        # Set the list of modules (Kratos and non-Kratos) that have been added up to current save
+        required_modules = list(sys.modules.keys())
+
+        # Create save folder
+        checkpoint_path = pathlib.Path(save_folder_name)
+        KratosMultiphysics.FilesystemExtensions.MPISafeCreateDirectories(checkpoint_path)
+
+        # Save current status
+        with open(os.path.join(checkpoint_path, checkpoint_file_name), 'wb+') as checkpoint_file:
```
philbucher:
```suggestion
        with open(checkpoint_path / checkpoint_file_name, 'wb+') as checkpoint_file:
```

There is NO reason to use `os.path` in modern python ;)

kratos/python_scripts/analysis_stage.py
```diff
@@ -211,6 +211,20 @@ def ChangeMaterialProperties(self):
         """this function is where the user could change material parameters as a part of the solution step """
         pass
 
+    def Save(self, serializer : KratosMultiphysics.StreamSerializer):
```
philbucher: 2 things:
1. why restrict to StreamSerializer? I would put the baseclass here
2. I STRONGLY recommend to either remove the type annotations entirely, or do it properly (i.e. add the return type). This will otherwise be a huge PITA when using e.g. mypy for checking the types (source: my own struggles with these problems :see_no_evil: )

kratos/python_scripts/analysis_stage.py
```diff
@@ -211,6 +211,20 @@ def ChangeMaterialProperties(self):
...
         pass
 
+    def Save(self, serializer : KratosMultiphysics.StreamSerializer):
+        """Serializes current analysis stage instance
+
+        This method is intended to make the class pure Python (pickable). This means serialize all the Kratos objects with the given serializer to then destroy them.
+        """
+        pass
+
+    def Load(self, serializer : KratosMultiphysics.StreamSerializer):
```
philbucher: same
```suggestion
    def Load(self, serializer: KratosMultiphysics.Serializer) -> None:

```

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
```diff
@@ -0,0 +1,182 @@
+import sys
+import importlib
+
+import KratosMultiphysics
+from KratosMultiphysics.project import Project
+from KratosMultiphysics.model_parameters_factory import KratosModelParametersFactory
+
+class MultistageOrchestrator():
```
philbucher:
```suggestion
class MultistageOrchestrator:
```

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
```diff
@@ -0,0 +1,187 @@
...
+                    self._stages_to_checkpoint_list = self.__GetExecutionList()
+                else:
+                    # Create an empty list to avoid errors
+                    self._stages_to_checkpoint_list = []
+
+            elif self.GetProject().GetSettings()["orchestrator"]["settings"]["stage_checkpoints"].IsStringArray():
+                # Check that the stages asked to be checkpointed are in the execution list
+                stages_asked_to_checkpoint = self.GetProject().GetSettings()["orchestrator"]["settings"]["stage_checkpoints"].GetStringArray()
+                aux_diff = set(stages_asked_to_checkpoint) - set(self.__GetExecutionList())
+                if len(aux_diff) == 0:
```
philbucher:
```suggestion
                if aux_diff:
```

kratos/python_scripts/project.py
```diff
@@ -0,0 +1,127 @@
...
+                "output_data" : self.__output_data,
+                "stage_names" : stage_names_list,
+                "stage_instances" : stage_instances_list,
+                "required_modules" : required_modules
+            }, checkpoint_file, protocol=2)
+
+        # Output current settings
+        if KratosMultiphysics.ParallelEnvironment.GetDefaultDataCommunicator().Rank() == 0:
+            if output_settings_file_name:
+                with open(os.path.join(checkpoint_path, output_settings_file_name), 'w') as parameter_output_file:
```
philbucher:
```suggestion
                with open(checkpoint_path / output_settings_file_name, 'w') as parameter_output_file:
```

kratos/python_scripts/project.py
```diff
@@ -0,0 +1,127 @@
...
+                    parameter_output_file.write(self.__settings.PrettyPrintJsonString())
+   
+    def Load(self, loading_point : str):
```
philbucher:
```suggestion
    def Load(self, loading_point: Path) -> None:
```

kratos/python_scripts/project.py
```diff
@@ -0,0 +1,127 @@
...
+        '''Loads a saved Project status into current one.'''
+
+        # Load save path file
+        loading_path = pathlib.Path(loading_point)
```
philbucher:
```suggestion
```

kratos/python_scripts/project.py
```diff
@@ -0,0 +1,127 @@
...
+            err_msg = f"Stage '{stage_name}' is already active and cannot be added to active stages."
+            raise Exception(err_msg)
+        self.__active_stages[stage_name] = stage_instance
+
+    def RemoveActiveStage(self, stage_name : str):
+        '''Removes an active stage instance from the current stages dictionary.'''
+
+        del self.__active_stages[stage_name]
+
+    def Save(self, save_folder_name : str, checkpoint_file_name : str, output_settings_file_name : str = None):
```
philbucher:
```suggestion
    def Save(self, save_folder_name: Path, checkpoint_file_name: Path output_settings_file_name: Optional[Path] = None) -> None:
```

=> requires `from typing import Optional`

kratos/python_scripts/project.py
```diff
@@ -0,0 +1,127 @@
...
+        '''Returns the current multistage simulation output data container.'''
+
+        return self.__output_data
+    
+    def GetActiveStages(self):
+        '''Returns the current multistage simulation active stages dictionary.'''
+
+        return self.__active_stages
+    
+    def AddActiveStage(self, stage_name : str, stage_instance):
```
philbucher: either add all or no type annotations

kratos/python_scripts/project.py
```diff
@@ -0,0 +1,127 @@
+import os
```
philbucher:
```suggestion
```

=> use `pathlib`

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
```diff
@@ -0,0 +1,182 @@
...
+    Member variables:
+    __project -- Current multistage simulation project container
+    """
+
+    def __init__(self, project : Project) -> None:
+        """Constructs a multistage orchestrator instance"""
+
+        # Store pointer to current project
+        # Note that the project already contains the multistage simulation settings
+        self.__project = project
```
philbucher:
```suggestion
        self.__project: Project = project
```

if you want to do it properly

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def Run(self):
+        """Main function that runs the complete multistage simulation."""
+
+        err_msg = "Calling base MultistageOrchestrator Run() method. This must be implemented in derived orchestrators."
+        raise NotImplementedError (err_msg)
```
philbucher:
```suggestion
    @abstractmethod
    def Run(self):
        """Main function that runs the complete multistage simulation."""
```

this is a more modern and safer solution (requires to derive the class from `ABC`)

philbucher (commented):
bunch of comments

kratos/python_scripts/analysis_stage.py
RiccardoRossi: about the first point i tend to disagree. We need to serialize this to a list of characters to later pickle it. for this the StreamSerializer is the proper object...

RiccardoRossi: about the first point i tend to disagree. We need to serialize this to a list of characters to later pickle it. for this the StreamSerializer is the proper object...

kratos/python_scripts/analysis_stage.py
RiccardoRossi: same comment

kratos/python_scripts/analysis_stage.py
rubenzorrilla: - About 1: see @RiccardoRossi 's comment.
- About 2: Basically I forgot to add the return types. I'll do it right away.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: In my first implementation I did it like this. However, it gave me problems. If I'm nor remembering wrong, it required the entire class to be abstract, and we're implementing a bunch of methods in here.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: I wasn't aware of this. Thanks for the hint.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
sunethwarna: I agree with @philbucher , you can have bunch of method implemented, and bunch to be abstract. This is safer, so at interpretation time you get the error message rather than the run time.

This is not possible when you have a base class like `Kratos.Process` since the pybind abstract meta class and python abstract meta class is in conflict.

kratos/python_scripts/analysis_stage.py
rubenzorrilla: Return type annotation added.

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
rubenzorrilla: Fair enough (I did it like this to make it more readable). In any case I think that it should be `if not aux_diff`.

rubenzorrilla: Done.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: OK. If so most probably I did something wrong. Let me give it another try.

matekelemen: > This is not possible when you have a base class like Kratos.Process since the pybind abstract meta class and python abstract meta class is in conflict.

That's a very unfortunate limitation of the `abc` implementation in python. I opened an [issue](https://github.com/python/cpython/issues/102077) for it a while back, but it didn't get any attention :'(. I'd very much like to use `abc` with `pybind` classes, so I'm open to giving it another go or roll our own implementation if someone else is also interested.

kratos/python_scripts/project.py
rubenzorrilla: Done.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: @philbucher @sunethwarna @matekelemen I did it again and it worked. I've discussed briefly with @roigcarlo to figure out what could have happened the last time we tried, we believe that is something related to the version before the refactor to orchestrators, long story short, nothing to be worried about.

kratos/python_interface/python_registry_utilities.py
pooyan-dadvand: This is what I want to avoid. Having the MultiStage in the front gives to the confusion that it is a type of stage and doesn't add any readability to it.

kratos/python_scripts/analysis_stage.py
pooyan-dadvand: I don't get the point. A typical use case would be save and continue. No?

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
pooyan-dadvand: So would you please ad a TODO comment to avoid more copy paste of it?

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
pooyan-dadvand: So I would correct the documentation

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
pooyan-dadvand: So I would state it in the documentation

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
pooyan-dadvand: I don't get the design then. What are the outputs?

kratos/python_scripts/analysis_stage.py
pooyan-dadvand: @RiccardoRossi I don't get your point. Are you using some non-base class interface?

kratos/python_interface/python_registry_utilities.py
rubenzorrilla: OK. Fair enough. Removing it.

kratos/python_scripts/analysis_stage.py
rubenzorrilla: Yes. But we are using Python `pickle` for doing so. Pickle is only capable to save / load pure Python types. Hence, we need to convert all Kratos objects to a pure Python type (this is what we do with our stream serializer). Then you "destroy" these objects in order to make the class Pickable. Let me try to write an example.

We have a `MyStage` analysis stage class with whatever Kratos object e.g. a Newton-Raphson instance. At this point, the class is not pickable as `self.strategy` cannot be dump. In order to do so, we call the stream serializer within the `Save` method of `MyClass` to create a string "representing" it. At this point, the `MyStage` is still not pickable as `self.strategy` is not a Python type but a Kratos one. Then, we need to do `self.strategy = None` after the stream serialization. At this point the clas IS pickable. This is what I meant by destroy. In pseudocode will be something like
```py
def Save(self, serializer: KratosMultiphysics.StreamSerializer):
    serializer.Save(self.strategy) # class is not pickable yet
    self.strategy = None # class is now pickable
```

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: Sure.

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
rubenzorrilla: The name of the output settings file (e.g. `ProjectParametersOutput.json`) . Also note that the function edits current settings accordingly.

kratos/python_scripts/analysis_stage.py
matekelemen: > Pickle is only capable to save / load pure Python types. 

https://pybind11.readthedocs.io/en/latest/advanced/classes.html#pickling-support

kratos/python_interface/python_registry_utilities.py
rubenzorrilla: Done.

kratos/python_scripts/analysis_stage.py
rubenzorrilla: > > Pickle is only capable to save / load pure Python types.
> 
> https://pybind11.readthedocs.io/en/latest/advanced/classes.html#pickling-support

Well, we need a poor soul to be sacrificed if we are to do this.

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
rubenzorrilla: Done.

kratos/python_scripts/multistage_orchestrators/sequential_multistage_orchestrator.py
rubenzorrilla: Done.

kratos/python_scripts/multistage_orchestrators/multistage_orchestrator.py
rubenzorrilla: Done.

rubenzorrilla:
@philbucher @pooyan-dadvand I went through all your requirements / suggestions. Besides I added a test running a "fake" multistage workflow.

Last but not least I'm sharing a working "real" example in here so you can see how the `MainKratos.py` will look like. You can execute it with `python3 MainKratos.py ProjectParametersCustom.json`. I'll update the examples repository and documentation accordingly once this is merged.

[1x1_square_multistage_project.zip](https://github.com/KratosMultiphysics/Kratos/files/12004797/1x1_square_multistage_project.zip)

kratos/python_scripts/analysis_stage.py
pooyan-dadvand: But the documentation you are referring don't says anything about making it null. It only says that `__setstate__` and `__getstate__` methods to be defined. Would you please give me the reference about making it null?

pooyan-dadvand:
Just to add that, going through your example, adding the stage setting make it much more readable than without!

rubenzorrilla:
> Just to add that, going through your example, adding the stage setting make it much more readable than without!

Agree. Besides, it makes easier the inclusion of already existing json files.

kratos/python_scripts/analysis_stage.py
matekelemen: Here's my take on it:

An object is `pickle`able if it defines `__setstate__` and `__getstate__`, **or** all its member variables are `pickle`able. Note the recursion here.

## Before `AnalysisStage::Save`

The analysis stage may have objects coming from pybind, such as solvers, processes, etc. By default, `pickle` cannot handle these because classes bound from C++ are only picklable if someone explicitly added support for it (i.e.: defined `__setstate__` and `__getstate__` in the bindings).

## In `AnalysisStage::Save`, before `pickle`

Before passing the analysis stage on to `pickle`, these members coming from C++ need to be dealt with. I'm guessing @rubenzorrilla proposes to use the kratos serializers to take over `pickle`'s job for these classes. After the kratos serializers did their job, the C++ classes are captured by the serializer so they can be removed from the analysis stage to make it `pickle`able (`self.cpp_class_instance = None`).

## `pickle`

The analysis stage is now free of un`pickle`able members, so it can be safely passed on to `pickle`.

##

My problem with this approach is that the analysis stage is left in an invalid state after `AnalysisStage::Save`. This means that at every checkpoint, we must not only `Save` the analysis stage but also `Load` it back in => i.e.: perform a restart every time.

kratos/python_scripts/orchestrators/orchestrator.py
```diff
@@ -0,0 +1,186 @@
...
+        elif hasattr(analysis_stage_module, "Create"):
+            # If Kratos convention is not fulfilled we search for a Create method
+            stage_instance = analysis_stage_module.Create(self.__project.GetModel(), input_stage_settings)
+        else:
+            err_msg = f"Analysis stage in '{analysis_stage_module_name}' Python module cannot be created. Please check class name or provide a 'Create' method."
+            raise Exception(err_msg)
+
+        return stage_instance
+
+    def RunCurrentStagePreprocess(self, stage_name: str, data: typing.Optional[dict] = None):
```
philbucher:
```suggestion
    def RunCurrentStagePreprocess(self, stage_name: str, data: typing.Optional[typing.Dict] = None):
```

pretty internal types can only be used from 3.9
then you could use it even better: `data: dict | None = None`

kratos/python_scripts/orchestrators/orchestrator.py
```diff
@@ -0,0 +1,186 @@
...
+                    modeler.SetupModelPart()
+                del modelers_list
+
+            if self.__project.GetSettings()["stages"][stage_name]["stage_preprocess"].Has("operations"):
+                operations_list = self.__CreateListOfOperations("stage_preprocess", stage_name)
+                for operation in operations_list:
+                    operation.Execute()
+                del operations_list
+
+    def RunCurrentStagePostprocess(self, stage_name: str, data: typing.Optional[dict] = None):
```
philbucher: I would do `from typing import Dict, Optional` at the top

kratos/python_scripts/analysis_stage.py
matekelemen: We could skip `Load`ing everything back in after `Save` if all members of the analysis stage were `pickle`able to begin with (i.e. the bindings define `__setstate__` and `__getstate__`). That means adding these extra bindings for **every** C++ type that may end up as a member of an `AnalysisStage`.

Yes that's a lot of work. However, it's just mundane copy-pasta work. We already have this functionality implemented with serializers, so we could just reuse them for the bindings:
```cpp
pybind11::class_<KratosClass, ...>(rModule, "KratosClass")
    ...
    .def(pybind11::pickle(
        [](const KratosClass& rInstance) {
                // use the serializer to convert rInstance into a string
                return pybind11::make_tuple(/*get string from the serializer*/);
        },
        [](pybind11::tuple Arguments) {
                // Arguments contains exactly one string
                // use the serializer to construct an instance from a string
         }
    ))
    ;
```

rubenzorrilla: Just to say that I added more details to the docstring.

kratos/python_scripts/orchestrators/orchestrator.py
rubenzorrilla: Done. Both in this file and in `project.py`.

kratos/python_scripts/orchestrators/orchestrator.py
```diff
@@ -0,0 +1,186 @@
...
+import sys
+import importlib
+from typing import Dict, Optional
+
+import KratosMultiphysics
+from KratosMultiphysics.project import Project
+from KratosMultiphysics.analysis_stage import AnalysisStage
+from KratosMultiphysics.model_parameters_factory import KratosModelParametersFactory
+
+class Orchestrator(abc.ABC):
```
loumalouomega, line 11: abc?

loumalouomega: I see https://docs.python.org/3/library/abc.html

loumalouomega (approved):
Removing my block

philbucher (approved):

kratos/python_scripts/analysis_stage.py
RiccardoRossi: the point is not that we cannot use the base class serializer. It would work.

The point is that we specifically want to pass a StreamSerializer to later pickle it, so the current code is doing exactly what we expect.

kratos/python_scripts/analysis_stage.py
pooyan-dadvand: > Here's my take on it:
> 
> An object is `pickle`able if it defines `__setstate__` and `__getstate__`, **or** all its member variables are `pickle`able. Note the recursion here.
> 

This is what I understood

> ## Before `AnalysisStage::Save`
> The analysis stage may have objects coming from pybind, such as solvers, processes, etc. By default, `pickle` cannot handle these because classes bound from C++ are only picklable if someone explicitly added support for it (i.e.: defined `__setstate__` and `__getstate__` in the bindings).
> 
> ## In `AnalysisStage::Save`, before `pickle`
> Before passing the analysis stage on to `pickle`, these members coming from C++ need to be dealt with. I'm guessing @rubenzorrilla proposes to use the kratos serializers to take over `pickle`'s job for these classes. After the kratos serializers did their job, the C++ classes are captured by the serializer so they can be removed from the analysis stage to make it `pickle`able (`self.cpp_class_instance = None`).

This is exactly my question: Why we should do self.cpp_class_instance=None? Would you please point me to the documentation in this respect?

> 
> ## `pickle`
> The analysis stage is now free of un`pickle`able members, so it can be safely passed on to `pickle`.
> 
> My problem with this approach is that the analysis stage is left in an invalid state after `AnalysisStage::Save`. This means that at every checkpoint, we must not only `Save` the analysis stage but also `Load` it back in => i.e.: perform a restart every time.

And this is my direct concern. In the case of this PR the stage is already deleted. But this limits the usage very much.

pooyan-dadvand:
Thanks @rubenzorrilla for considering the suggestions.

I am unblocking as the pickleable problem would not affect the current implementation.

pooyan-dadvand (approved):

kratos/python_scripts/analysis_stage.py
matekelemen: > This is exactly my question: Why we should do `self.cpp_class_instance=None`? Would you please point me to the documentation in this respect?

I can't point you to the docs, because this is just one possible roundabout solution to a problem we created for ourselves, so I don't think the docs would mention anything like this. The followings are just as valid solutions:
`self.cpp_class_instance = 0`
`self.cpp_class_instance = "whatever"`
`self.cpp_class_instance = 6.0 * 1e23`
`self.cpp_class_instance = anything_that_is_pickleable`
`del self.cpp_class_instance`

The point is to remove any un`pickle`able member.

The proper (albeit more laborious) solution is to make our C++ types `pickle`able.

RiccardoRossi: the thing is more delicate than simple copy-pasting.

Imagine you want to store a Model and then some other python object.

one way of doing this could be to create a serializer object in the pickle pybind call, write it to a string and then write the resulting char array to pickle.

the problem of this approach is that if we now want to have a different save for the second object, we don't have access to the serializer used to save the first object. We could generate a second StreamSerializer insance, however this implies that common things (like the kernel) would be serialized twice (and differently) in the first object and in the second.

What i am trying to say is that we need to first serialize all of the cpp objects USING THE FIRST SERIALIZER, and then write the serialized object using pickle. That is exactly what we are doing.
no doubts we could have a global serializer used in python ... however this complicates things quite a lot.

having said this, this is an implementation detail. If you like, let's open a different thread to discuss about that

PR #11356, opened 2023-07-05
[FastPR][Core] ModelPartIO geometries submodelparts fix

rubenzorrilla:
**📝 Description**
Hotfix to also do the parsing of the submodelpart geometries.

@jginternational this should fix the issue you reported to us yesterday.

### DIFF
--- a/kratos/sources/model_part_io.cpp
+++ b/kratos/sources/model_part_io.cpp
@@ -3639,7 +3639,9 @@ void  ModelPartIO::ReadSubModelPartConditionsBlock(ModelPart& rMainModelPart, Mo
     KRATOS_CATCH("")
 }
 
-void  ModelPartIO::ReadSubModelPartGeometriesBlock(ModelPart& rMainModelPart, ModelPart& rSubModelPart)
+void  ModelPartIO::ReadSubModelPartGeometriesBlock(
+    ModelPart& rMainModelPart,
+    ModelPart& rSubModelPart)
 {
     KRATOS_TRY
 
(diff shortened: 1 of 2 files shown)


kratos/sources/model_part_io.cpp
```diff
@@ -3637,6 +3639,29 @@ void  ModelPartIO::ReadSubModelPartConditionsBlock(ModelPart& rMainModelPart, Mo
...
+    while (!mpStream->eof())
+    {
+        ReadWord(word); // Reading the node id or End
+        if (CheckEndBlock("SubModelPartGeometries", word))
+            break;
+
+        ExtractValue(word, geometry_id);
+                    ordered_ids.push_back(geometry_id);
+    }
+    std::sort(ordered_ids.begin(), ordered_ids.end());
```
philbucher, line 3661: Why the sort?

This makes it inconsistent with the mdpa file

Or is this also done for elems and conds?

kratos/sources/model_part_io.cpp
```diff
@@ -3637,6 +3639,29 @@ void  ModelPartIO::ReadSubModelPartConditionsBlock(ModelPart& rMainModelPart, Mo
...
+{
+    KRATOS_TRY
+
+    SizeType geometry_id;
+    std::string word;
+    std::vector<SizeType> ordered_ids;
+
+    while (!mpStream->eof())
+    {
+        ReadWord(word); // Reading the node id or End
```
philbucher: Wrong comment

philbucher (commented):
Minor otherwise looks good

kratos/sources/model_part_io.cpp
rubenzorrilla: Basically I took what we do for elements and conditions and I applied here as well.

kratos/sources/model_part_io.cpp
rubenzorrilla: Copy & paste error. Thanks for spotting.

kratos/sources/model_part_io.cpp
rubenzorrilla: I've been checking the `AddGeometries` of the `ModelPart` and I don't see why the `std::sort` is needed. As far as I understand it shouldn't affect the performance of the key (id) search that we do in there as the underlying geometry container is an `std::unordered_map`.

philbucher (approved):

kratos/sources/model_part_io.cpp
philbucher: Seems that you didn't remove it but up to you

rubenzorrilla: Yes, I want to confirm the removal with the @KratosMultiphysics/technical-committee guys.

rubenzorrilla:
For the record. We merge the hotfix as it is and we'll discuss later on about the sorting.

PR #11351, opened 2023-07-04
[OptApp] Vertex-Morphing (Node-based) shape optimization

RezaNajian:
This PR implements Vertex-Morphing shape control using implicit and explicit filtering techniques. The goal is to enable node-based shape optimization for solid and shell geometries of arbitrary shapes. Tests and examples will be provided.

PR #11359, opened 2023-07-06
[OptApp] Adding explicit filtering including fixing/damping

RezaNajian:
**📝 Description**
This PR adds explicit (convolution-based) filtering (matrix free) features to `OptimizationApplication` for `ContainerExpression`s. Hence this supports `double`, `array_1d<double, 3>`, `array_1d<double, 4>`, `array_1d<double, 6>`, `array_1d<double, 9>`, `Vector` and `Matrix` variables types. This is not **MPI** compatible. It also handles fixed/not filtered areas, if the fixed domain is provided as an extra model part. Damping is implemented in a consistent way. We also note that the new explicit filter provides two main filtering functions, FilterField for filtering a NOT integrated field like pressure or control field and FilterIntegratedField for filtering already integrated fields like sensitivities of integrated functionals. Consistency tests are made and provided, to make sure that a constant field over non-uniform mesh is filtered to a uniform field.

Following filters are added.
1. `NodalExplicitFilter` which is based on the nodal coordinates.
2. `ConditionExplicitFilter` which is based on the center point coordinates of the conditions.
3. `ElementExplicitFilter` which is based on the center point coordinates of the elements.

The reference values for tests are generated from the matrix free vertex morphing mechanism in `ShapeOptimizationApplication` for all three types of new filters.

These newly introduced filters allows having `ContainerExpression` for filter radius, hence one can have different radius for different entities (adaptive radius). The current tests uses constant radius.

**🆕 Changelog**
- Added `NodalExplicitFilter`
- Added `ConditionExplicitFilter`
- Added `ElementExplicitFilter`
- Added tests.


applications/OptimizationApplication/custom_python/add_custom_filters_to_python.cpp
```diff
@@ -17,7 +17,7 @@
 // Project includes
 
 // Application includes
-#include "custom_utilities/filtering/explicit_vertex_morphing_filter.h"
```
Igarizza, line 20: I don't understand why do you suggest removing vertex morphing from the name if we implement vertex morphing filter. Can you please clarify? 

As Vertex Morphing is a discrete explicit filter based on Euclidian distances by definition, hence, in my opinion, I would keep it as a simple vertex morphing. Modified version of it can have any addition to its name.

RezaNajian: @Igarizza The reason is simple. The implemented functionality is basically convolution-based filtering used in many disciplines. The application of explicit filtering to regularize node-based shape optimization is coined as Vertex Morphing. The implemented convolution-based filtering can be used in other kinds of optimizations, like Topology optimization. Also, note that the implemented filtering functions can filter any field type, such as face or volume fields, so I highly recommend being consistent with the literature and others.

RezaNajian: Also note that Vertex Morphing is implemented as a node-based shape control in PR #11351

RezaNajian:
The damping feature will be addressed in a separate place, due to the fact that we can have different settings for different damping model parts. It would be advantageous since the ExplicitFilter will be pure filtering.

PR #11360, opened 2023-07-06
[Core] Refactor nurbs volume quad point interface.

manuelmessmer:
**📝 Description**
Update CreateQuadraturePointGeometries() of NurbsVolumeGeometry.
So far, one QuadraturePointGeometry was created for each integration point. Consequently, each point was associated with a separate element. This was very expensive for the assembly of the system matrices since the contributions of each point were assembled to the global matrices. 
Now, the quadrature point geometries are created per element and hence contain multiple integration points.

**🆕 Changelog**
- CreateQuadraturePointGeometries() of NurbsVolumeGeometry


manuelmessmer:
Will close this. Need some additional changes.

PR #11362, opened 2023-07-06
[Core] Refactor nurbs volume quadrature point interface.

manuelmessmer:
**📝 Description**
Refactor CreateQuadraturePointGeometries() of NurbsVolumeGeometry.
So far, in IGA  in Kratos one QuadraturePointGeometry is created for each integration point. Consequently, each point is associated with a separate element. This is quite expensive (at least for solids) for the assembly of the system matrices since the contributions of each point are assembled to the global matrices.
I added an additional interface that allows to create one quadrature point geometry per knot span, which contains all integration points.

**🆕 Changelog**
- CreateQuadraturePointGeometries() of NurbsVolumeGeometry

### DIFF
--- a/kratos/geometries/nurbs_volume_geometry.h
+++ b/kratos/geometries/nurbs_volume_geometry.h
@@ -690,7 +690,7 @@ public:
     ///@{
 
     /**
-     * @brief Creates a list of quadrature point geometries.
+     * @brief Creates a list of quadrature point geometries
      *        from a list of integration points.
      * @param rResultGeometries List of quadrature point geometries.
      * @param NumberOfShapeFunctionDerivatives the number of evaluated
@@ -758,7 +758,6 @@ public:
             ShapeFunctionsValuesContainerType shape_function_values;
             ShapeFunctionsLocalGradientsContainerType shape_function_gradients;
 
-            // Only default integration method Gauss_1=0 is used.
             integration_points[0] = rIntegrationPoints;
             shape_function_gradients[0].resize(rIntegrationPoints.size());
             shape_function_values[0].resize(rIntegrationPoints.size(), num_nonzero_cps);
--- a/kratos/geometries/nurbs_volume_geometry.h
+++ b/kratos/geometries/nurbs_volume_geometry.h
@@ -700,26 +700,49 @@ public:
         const IntegrationPointsArrayType& rIntegrationPoints,
         IntegrationInfo& rIntegrationInfo) override
     {
-        // shape function container.
+        // Shape function container.
         NurbsVolumeShapeFunction shape_function_container(
             mPolynomialDegreeU, mPolynomialDegreeV, mPolynomialDegreeW, NumberOfShapeFunctionDerivatives);
 
         // Resize containers.
-        if (rResultGeometries.size() != rIntegrationPoints.size())
-            rResultGeometries.resize(rIntegrationPoints.size());
+        if (rResultGeometries.size() != 1){
+            rResultGeometries.resize(1);
+        }
 
-        auto default_method = this->GetDefaultIntegrationMethod();
-        SizeType num_nonzero_cps = shape_function_container.NumberOfNonzeroControlPoints();
+        const auto default_method = this->GetDefaultIntegrationMethod();
+
+        const SizeType num_nonzero_cps = shape_function_container.NumberOfNonzeroControlPoints();
+        const SizeType num_points = rIntegrationPoints.size();
+
+        KRATOS_ERROR_IF(num_points < 1) << "List of integration points is empty.\n";
+
+        KRATOS_ERROR_IF(NumberOfShapeFunctionDerivatives != 2) << "NumberOfShapeFunctionDerivatives must be 2.\n";
 
         Matrix N(1, num_nonzero_cps);
         DenseVector<Matrix> shape_function_derivatives(NumberOfShapeFunctionDerivatives - 1);
 
-        for (IndexType i = 0; i < NumberOfShapeFunctionDerivatives - 1; ++i) {
-            const IndexType num_derivatives = (2 + i) * (3 + i) / 2;
-            shape_function_derivatives[i].resize(num_nonzero_cps, num_derivatives);
+        GeometryShapeFunctionContainer<GeometryData::IntegrationMethod>::IntegrationPointsContainerType integration_points;
+        GeometryShapeFunctionContainer<GeometryData::IntegrationMethod>::ShapeFunctionsValuesContainerType shape_function_values;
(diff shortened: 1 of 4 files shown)


kratos/geometries/nurbs_volume_geometry.h
```diff
@@ -700,69 +734,160 @@ class NurbsVolumeGeometry : public Geometry<typename TContainerPointType::value_
...
 
-        for (IndexType i = 0; i < rIntegrationPoints.size(); ++i)
-        {
-            // Attention: Weights are not yet implemented.
-            // if (IsRational()) {
-            //     shape_function_container.ComputeNurbsShapeFunctionValues(
-            //         mKnotsU, mKnotsV, mKnotsW, mWeights, rIntegrationPoints[i][0], rIntegrationPoints[i][1], rIntegrationPoints[i][2]);
-            // }
+            for( IndexType i_point = 0; i_point < num_points; ++i_point){
+                for (IndexType i = 0; i < NumberOfShapeFunctionDerivatives - 1; ++i) {
```
matekelemen: `NumberOfShapeFunctionDerivatives` is guaranteed to be 2, right? Then everything in the loop scope runs for `i=0` only and doesn´t need the loop at all.

kratos/geometries/nurbs_volume_geometry.h
```diff
@@ -700,69 +734,160 @@ class NurbsVolumeGeometry : public Geometry<typename TContainerPointType::value_
...
+
+                /// Get Shape Functions.
+                for (IndexType j = 0; j < num_nonzero_cps; ++j) {
+                    shape_function_values[0](i_point, j) = shape_function_container(j, 0);
+                }
 
+                // Get Shape Function Derivatives DN_De, ...
+                if (NumberOfShapeFunctionDerivatives > 0) {
+                    IndexType shape_derivative_index = 1;
+                    for (IndexType n = 0; n < NumberOfShapeFunctionDerivatives - 1; ++n) {
```
matekelemen: same

rickyaristio:
> **📝 Description** Refactor CreateQuadraturePointGeometries() of NurbsVolumeGeometry. So far, in IGA in Kratos one QuadraturePointGeometry is created for each integration point. Consequently, each point is associated with a separate element. This is quite expensive (at least for solids) for the assembly of the system matrices since the contributions of each point are assembled to the global matrices. I added an additional interface that allows to create one quadrature point geometry per knot span, which contains all integration points.
> 
> **🆕 Changelog**
> 
> * CreateQuadraturePointGeometries() of NurbsVolumeGeometry

Basically, I am not against the change :). How efficient is this way?

kratos/geometries/nurbs_volume_geometry.h
rickyaristio: It is not always. In the input, we assign the shape function order, which can be 1 or 2 and so on (case dependent).

matekelemen: but there's a check in L740 at the head of this branch:
```cpp
KRATOS_ERROR_IF(NumberOfShapeFunctionDerivatives != 2) << "NumberOfShapeFunctionDerivatives must be 2.\n";
```

rickyaristio: My bad. Yes, in this branch, we can omit the loop.

manuelmessmer: Yes, that is correct. Actually, the current interface does not provide the necessary containers for the higher-order derivatives. I thought I keep it in case we change it at some point. I could comment it out. In fact, for the solids we most likely never require second-order derivatives..

manuelmessmer:
> > **memo Description** Refactor CreateQuadraturePointGeometries() of NurbsVolumeGeometry. So far, in IGA in Kratos one QuadraturePointGeometry is created for each integration point. Consequently, each point is associated with a separate element. This is quite expensive (at least for solids) for the assembly of the system matrices since the contributions of each point are assembled to the global matrices. I added an additional interface that allows to create one quadrature point geometry per knot span, which contains all integration points.
> > **new Changelog**
> > 
> > * CreateQuadraturePointGeometries() of NurbsVolumeGeometry
> 
> Basically, I am not against the change :). How efficient is this way?

System matrix resize time and build is much more efficient. I am not sure how severe this problem is for the shells. But in 3D it is horrible. Here I tried to make it work with very limited changes. However, to change everything to a more 'element-oriented' approach would involve drastic changes (also in the base geometry class), e.g. the CreateIntegrationPoint function return only a list of integration points. There we would need at least a `vector<IntegrationPointsArray>` to group the integration points. However, this would probably contradict with the requirements for the MPMApplication. 

Therefore, the current changes only use this element-wise approach for my customized integration points coming from TIBRA. If you use the standard functions, it still creates one Geometry per integration points.

kratos/geometries/nurbs_volume_geometry.h
manuelmessmer: The problem is that the current interface does not support what I want to do. Therefore, this is a bit hacky. I think in the long term we should think about changing the entire interface. However, if this is not important for the shells etc. I guess it is not worth the effort.

matekelemen: It's OK to keep the loop if you're planning to add support for other orders later on; I just though there's an algorithmic restriction to `2`.

rickyaristio:
> > > **memo Description** Refactor CreateQuadraturePointGeometries() of NurbsVolumeGeometry. So far, in IGA in Kratos one QuadraturePointGeometry is created for each integration point. Consequently, each point is associated with a separate element. This is quite expensive (at least for solids) for the assembly of the system matrices since the contributions of each point are assembled to the global matrices. I added an additional interface that allows to create one quadrature point geometry per knot span, which contains all integration points.
> > > **new Changelog**
> > > 
> > > * CreateQuadraturePointGeometries() of NurbsVolumeGeometry
> > 
> > 
> > Basically, I am not against the change :). How efficient is this way?
> 
> System matrix resize time and build is much more efficient. I am not sure how severe this problem is for the shells. But in 3D it is horrible. Here I tried to make it work with very limited changes. However, to change everything to a more 'element-oriented' approach would involve drastic changes (also in the base geometry class), e.g. the CreateIntegrationPoint function return only a list of integration points. There we would need at least a `vector<IntegrationPointsArray>` to group the integration points. However, this would probably contradict with the requirements for the MPMApplication.
> 
> Therefore, the current changes only use this element-wise approach for my customized integration points coming from TIBRA. If you use the standard functions, it still creates one Geometry per integration points.

Ok, as the 'element oriented' approach is much more efficient for solids, I approve the change. 
For the shells, that might be interesting. I will try and see whether it is possible to be applied.

rickyaristio (dismissed):

manuelmessmer:
Thanks a lot guys. 
@rickyaristio we can test the shells at some point! Could you approve again. I made some minor additional changes.

rickyaristio (approved):

PR #11366, opened 2023-07-06
[MPM] Erase particle elements and conditions not found

ncrescenzio:
The process `ParticleEraseProcess` is called after the search of particle elements and conditions is performed and it removes those elements and conditions not found in the background grid.


VeronikaSinger (approved):
Thanks @ncrescenzio . I approve

PR #11376, opened 2023-07-11
[Core] Adding `EntitityIdentifier` to `EntitiesUtilities`

loumalouomega:
**📝 Description**

Part of https://github.com/KratosMultiphysics/Kratos/pull/10726

This PR adds a new Entity Identifier class and related utilities. The patch includes additions to the `entities_utilities.cpp` and `entities_utilities.h` files. 

This class has as objective to provide the corresponding entity prototype (element/condition) of a given geometry, prior name of the entity provided in construction. The name can be:

- `Single`:  Single definition, the geometry is not checked later (only in debug) as it is assumed that it is always the same geometry as the definition is unique. Example: "Element3D4N"
- `Multiple`: Multiple definitions is when a list of potential elements/conditions is considered and then assigned with the corresponding geometry. Example: "Element3D4N;Element3D8N"
- `Templated`: Templated definition, the geometry is checked, and the name is adapted to the given geometry. Example: "Element#D#N", where # will be replaced depending on the geometry provided.

The implementation tries to minimize the penalty of checking the geometry and providing the prototype for the `Multiple` and `Templated` case (`Single` is not checked), this is done defining the prototype as a pointer and reassigned in the case that the geometry type does not correspond with the prototype. In the particular case of the `Multiple` case, the different cases are saved in a map geometry type/entity name for later execution, avoiding extracting the names on execution.

FYI @rubenzorrilla 

**🆕 Changelog**

- [Adding `EntitityIdentifier` to `EntitiesUtilities`](https://github.com/KratosMultiphysics/Kratos/commit/12c02479130dbba677045b890d1a18af004f477e)


kratos/utilities/entities_utilities.cpp
```diff
@@ -17,11 +17,144 @@
...
+#include "utilities/geometry_utilities.h"
 
-namespace Kratos
+namespace Kratos::EntitiesUtilities
 {
-namespace EntitiesUtilities
+
+template<class TEntity>
+EntitityIdentifier<TEntity>::EntitityIdentifier(const std::string& rName)
+    : mName(rName)
```
RiccardoRossi: i do not fully understand this code, so i would appreciate if you explain a bit in words what you are trying to accomplish.

However i am under the impression that you are assuming that registered elements follow a strict naming convention.
what happens if you have a nonstandard name? say that i have something registered as "MyShell" what would happen in that case?

kratos/utilities/entities_utilities.h
```diff
@@ -32,6 +33,148 @@ namespace Kratos
...
+        /// Print object's data.
+        void PrintData(std::ostream& rOStream) const;
+
+        ///@}
+    private:
+        ///@name Member Variables
+        ///@{
+
+        typename TEntity::Pointer mpPrototypeEntity = nullptr;                    /// The prototype entity
+        DefinitionType mDefinitionType = DefinitionType::Single;                  /// The type of definition
```
RiccardoRossi: what is this??

loumalouomega: It is an enum, is defined some lines above

kratos/utilities/entities_utilities.cpp
loumalouomega: I will extend description, but if the element/condition is not registered it throws an error

loumalouomega: Definition extended

RiccardoRossi:
I am writing here some comments that stem out of our offline conversation about the PR.

The first comment is that i would modify the "mTypes" to contain a pointer to the reference Entity, as obtained from the registry.
That will allow using two nested maps involving a string search.
this would mean using

~~~cpp
std::unordered_map<GeometryData::KratosGeometryType, TEntityType* > mTypes;
~~~

per regards the "templated" case, i would 
1. see if there is a # in the input
2. if there is an input populate automatically the mTypes with all the available types (the registered ones)

the advantage is that now you would not have to do anything special at runtime for the "templated" case.



other than that, i would attempt simplifying the function "EntitityIdentifier" dividing it in smaller functions. Something on the line of

~~~cpp
if(#in input) 
    if ( there is a ";" ) ERROR cannot use #and ; together
    populate mTypes with all the possible registered types of the chosen element family
else if ( there is a ";" )
    pouplate the mTypes with the given list
else
   just retrieve the one given   
~~~

with the gimnastics of string manipulation hidden in helper functions


other than that ok to me

loumalouomega:
@RiccardoRossi comments applied

kratos/utilities/entities_utilities.cpp
```diff
@@ -176,24 +177,16 @@ void EntitityIdentifier<TEntity>::GenerateTemplatedTypes(const std::string& rNam
     // Check the type of entity considered
     const std::string entity_type_name = GetEntityTypeName();
 
+    // Prepare regex 
+    const std::string replace_dimension = StringUtilities::ReplaceAllSubstrings(rName, "#D", "(2D|3D)");
+    const std::string replace_number_of_nodes = StringUtilities::ReplaceAllSubstrings(replace_dimension, "#N", "[0-9]+N");
```
RiccardoRossi, line 152: I am not into regex (@roigcarlo could you take a look into this?) but as i understand you are replacing the # by a number in the range 0,9

you may have elements with 27 nodes in 3D ... but for example there are no elements with 26 ...

kratos/utilities/entities_utilities.cpp
```diff
@@ -12,16 +12,201 @@
...
+        const auto& r_reference_geometry_type = r_reference_geometry.GetGeometryType();
+        mTypes.insert({r_reference_geometry_type, &KratosComponents<TEntity>::Get(r_entity_name)});
+    }
+
+    // Set the definition type
+    mDefinitionType = EntitiesUtilities::DefinitionType::Multiple;
+
+    // Assign dumy prototype
+    typename GeometryType::Pointer p_geometry_type = Kratos::make_shared<GeometryType>();
+    mpPrototypeEntity = KratosComponents<TEntity>::Get(entity_type_name + "2D2N").Create(0, p_geometry_type, nullptr);
```
RiccardoRossi: i don't understadn what you do here... why do you generate a type 2D2N?

if it is something you cannot use "as is" i would leave it a nullptr

RiccardoRossi (commented):
will take a longer loop later on

kratos/utilities/entities_utilities.cpp
```diff
@@ -12,16 +12,201 @@
...
+            mTypes.insert({r_entity.GetGeometry().GetGeometryType(), &r_entity});
+        }
+    }
+
+    // Set the definition type
+    mDefinitionType = EntitiesUtilities::DefinitionType::Multiple;
+
+    // Assign dumy prototype
+    typename GeometryType::Pointer p_geometry_type = Kratos::make_shared<GeometryType>();
+    mpPrototypeEntity = KratosComponents<TEntity>::Get(entity_type_name + "2D2N").Create(0, p_geometry_type, nullptr);
```
RiccardoRossi: what is this dummy? i would leave it a "nullptr" if we don't use it

kratos/utilities/entities_utilities.cpp
roigcarlo: `[0-9]+N` means "at least one number followed by an N", which can be more than one. You could use `\d+N` as well.

if you want to restrict to specific numbers you could put them in a list, but that kills the purpose of the regex.

loumalouomega: I would work with any integer

RiccardoRossi: ok, after the didactic explanation of @roigcarlo in my office i understood the code.

i post here the explanation for future use:

~~~cpp
Surface#D#N //this is the input

replace_dimension = "Surface(2D|3D)#N" //after the first substitution
replace_number_of_nodes = "Surface(2D|3D)[0-9]+N" //after the second substitution

std::regex pattern(replace_number_of_nodes); //we will look for a regular expression as in "replace_number_of_nodes".
if we find one we add it to the types
~~~

loumalouomega:
~~~c++
#include <iostream>
#include <regex>

int main() {
    std::vector<std::string> strings = {
        "Element2D42N",
        "Element3D123N",
        "Element2D987N",
        "Element3D5N",
        "InvalidString"
    };

    std::regex pattern("Element(2D|3D)[0-9]+N");

    for (const auto& str : strings) {
        if (std::regex_match(str, pattern)) {
            std::cout << str << " matches the pattern." << std::endl;
        } else {
            std::cout << str << " does not match the pattern." << std::endl;
        }
    }

    return 0;
}
~~~

Will return:

~~~sh
Element2D42N matches the pattern.
Element3D123N matches the pattern.
Element2D987N matches the pattern.
Element3D5N matches the pattern.
InvalidString does not match the pattern.
~~~

kratos/utilities/entities_utilities.cpp
loumalouomega: If I leave nullptr it will fail when checking geometrytype, and if I check if nullptr I will add one more check. What I can do it is to check the first one in mTypes

kratos/utilities/entities_utilities.cpp
loumalouomega: AS I said before, will fail when checking geometrytype, and adding a extra check has an extra cost

kratos/utilities/entities_utilities.cpp
loumalouomega: I saw your self explanation later, Okay

kratos/utilities/entities_utilities.cpp
RiccardoRossi: well, to be honesty i don't think this is a good solution. 

the entity type is a nullptr up to the point at which you have the correct entity... 

both p_geometry_type and mpPrototypeEntity should be either nullptrs or have the correct type

kratos/utilities/entities_utilities.cpp
RiccardoRossi: i would never create a wrong type...

loumalouomega: Okay; I will provide an alternative solution

loumalouomega: Now

kratos/utilities/entities_utilities.cpp
loumalouomega: Fixed

loumalouomega:
Ready @RiccardoRossi

kratos/utilities/entities_utilities.cpp
```diff
@@ -12,16 +12,199 @@
...
+        const auto& r_reference_geometry = r_ref_entity.GetGeometry();
+        const auto& r_reference_geometry_type = r_reference_geometry.GetGeometryType();
+        mTypes.insert({r_reference_geometry_type, &KratosComponents<TEntity>::Get(r_entity_name)});
+    }
+
+    // Set the definition type
+    mDefinitionType = EntitiesUtilities::DefinitionType::Multiple;
+
+    // Assign default prototype
+    const auto* p_entity = mTypes.begin()->second;
```
RiccardoRossi: this is wrong again... if i have more than one type i am creating it here with the wrong type.

why do you need mpPrototypeEntity if you have multiple types?

kratos/utilities/entities_utilities.cpp
```diff
@@ -12,16 +12,199 @@
...
+            mTypes.insert({r_entity.GetGeometry().GetGeometryType(), &r_entity});
+        }
+    }
+
+    // Set the definition type
+    mDefinitionType = EntitiesUtilities::DefinitionType::Multiple;
+
+    // Assign default prototype
+    const auto* p_entity = mTypes.begin()->second;
+    mpPrototypeEntity = p_entity->Create(0, p_entity->pGetGeometry(), nullptr);
```
RiccardoRossi: here it must be a nullptr...

kratos/utilities/entities_utilities.cpp
```diff
@@ -12,16 +12,199 @@
...
+    switch (mDefinitionType) {
+        case DefinitionType::Single:
+            KRATOS_DEBUG_ERROR_IF_NOT(pGeometry->GetGeometryType() == mpPrototypeEntity->GetGeometry().GetGeometryType()) << "Trying to replace an entity with a different geometry type. Reference entity " << mpPrototypeEntity->GetGeometry().Info() << " vs  " << pGeometry->Info() << "\n Entity info: " << mpPrototypeEntity->Info() << std::endl;
+            return *mpPrototypeEntity;
+        case DefinitionType::Multiple:
+            // We check if the current type is correct or retrieving is required
+            if (pGeometry->GetGeometryType() != mpPrototypeEntity->GetGeometry().GetGeometryType()) {
+                mpPrototypeEntity = (mTypes[pGeometry->GetGeometryType()])->Create(0, pGeometry, nullptr);
+            }
+            return *mpPrototypeEntity;
```
RiccardoRossi: here you should return 

mTypes[pGeometry->GetGeometryType()])->second 

mpPrototypeEntity should never be used in DefinitionType::Multiple

kratos/utilities/entities_utilities.cpp
```diff
@@ -12,16 +12,199 @@
...
+    }
+}
+
+/***********************************************************************************/
+/***********************************************************************************/
+
+template<class TEntity>
+void EntitityIdentifier<TEntity>::PrintData(std::ostream& rOStream) const
+{
+    rOStream << "Name: " << mpPrototypeEntity->Info() << "\n";
```
RiccardoRossi: this should be only valid in the single case. In the Multiple case you have a lot of different prototypes

kratos/utilities/entities_utilities.cpp
loumalouomega: Because I want to avoid to look in a map every time (with the corresponding cost), this is a buffer to avoid search in a map

kratos/utilities/entities_utilities.cpp
loumalouomega: No, as I said if nullptr it will fail later when checking the geometry type is the same as the buffer (which is required to minimize the number of searches in the map)

kratos/utilities/entities_utilities.cpp
loumalouomega: Read my previous message, please, this is needed to minimize costs

kratos/utilities/entities_utilities.cpp
RiccardoRossi: but you are verifying every time if the type is correct...

kratos/utilities/entities_utilities.cpp
loumalouomega: True

kratos/utilities/entities_utilities.cpp
loumalouomega: Verify enum (integer) << search in a map

loumalouomega: Maybe will be some corner cases, but I am quite confident that in most cases it will be like this. @roigcarlo opinion?

RiccardoRossi: not just that, you also have 

~~~cpp
if (pGeometry->GetGeometryType() != mpPrototypeEntity->GetGeometry().GetGeometryType())
~~~

but at least now i understand the reasoning. HOWEVER the big problem with your proposed implementation is that it is inherently not threadsafe

loumalouomega: Yes it is threadsafe if you copy (private) the EntityIdentifier, see https://github.com/KratosMultiphysics/Kratos/pull/10726, where is actually used

loumalouomega:
Done @RiccardoRossi

loumalouomega:
@RiccardoRossi I applied the last suggestion you gave me (using directly an array to avoid searches)

kratos/utilities/entities_utilities.h
```diff
@@ -32,6 +33,179 @@ namespace Kratos
...
+    {
+    public:
+        ///@name Type Definitions
+        ///@{
+
+        /// Geometry type definition
+        using GeometryType = typename TEntity::GeometryType;
+
+        /// Length of array definition
+        constexpr static std::size_t LengthArray = static_cast<int>(GeometryData::KratosGeometryType::NumberOfGeometryTypes) - 2;
```
RiccardoRossi: shouldn't this be 
static_cast<std::size_t> 
?

RiccardoRossi (commented):
much better!

let me very picky with one last thing:
to determine the size of the Geometries enum, take a look at 

https://stackoverflow.com/questions/14989274/is-it-possible-to-determine-the-number-of-elements-of-a-c-enum-class

i think it is more robust than the trick you are employing

RiccardoRossi:
btw, when using the array, does it still make sense to distinguish the case of single and multiple? i am telling since the multiple case became very very efficient, and if we avoid distinguishing we could simplify even more the code and avoid an if

kratos/utilities/entities_utilities.h
loumalouomega: Yep

loumalouomega:
> much better!
> 
> let me very picky with one last thing: to determine the size of the Geometries enum, take a look at
> 
> [stackoverflow.com/questions/14989274/is-it-possible-to-determine-the-number-of-elements-of-a-c-enum-class](https://stackoverflow.com/questions/14989274/is-it-possible-to-determine-the-number-of-elements-of-a-c-enum-class)
> 
> i think it is more robust than the trick you are employing

But in that example they do what I do

kratos/utilities/entities_utilities.h
loumalouomega: Yep

loumalouomega:
> btw, when using the array, does it still make sense to distinguish the case of single and multiple? i am telling since the multiple case became very very efficient, and if we avoid distinguishing we could simplify even more the code and avoid an if

Done

kratos/utilities/entities_utilities.cpp
```diff
@@ -12,16 +12,165 @@
...
+
+    // Split the name
+    const auto splitted_names = StringUtilities::SplitStringByDelimiter(rName, ';');
+    for (auto& r_entity_name : splitted_names) {
+        // Some values need to be mandatorily prescribed since no meaningful default value exist. For this reason try accessing to them so that an error is thrown if they don't exist
+        KRATOS_ERROR_IF(r_entity_name != "" && !KratosComponents<TEntity>::Has(r_entity_name)) << entity_type_name << " name not found in KratosComponents<" << entity_type_name << "> -- name is " << r_entity_name << std::endl;
+        const auto& r_ref_entity = KratosComponents<TEntity>::Get(r_entity_name);
+        const auto& r_reference_geometry = r_ref_entity.GetGeometry();
+        const auto& r_reference_geometry_type = r_reference_geometry.GetGeometryType();
+        const std::size_t index = static_cast<std::size_t>(r_reference_geometry_type) - 1;
```
RiccardoRossi: ouch, i just checked and the numeration of enums starts with 0 ... so if you substract -1 there is a problem ...

kratos/utilities/entities_utilities.h
```diff
@@ -32,6 +33,167 @@ namespace Kratos
...
+    {
+    public:
+        ///@name Type Definitions
+        ///@{
+
+        /// Geometry type definition
+        using GeometryType = typename TEntity::GeometryType;
+
+        /// Length of array definition
+        constexpr static std::size_t LengthArray = static_cast<std::size_t>(GeometryData::KratosGeometryType::NumberOfGeometryTypes) - 2;
```
RiccardoRossi: this should be a -1, why -2?

kratos/utilities/entities_utilities.cpp
```diff
@@ -12,16 +12,165 @@
...
+    return mIsInitialized;
+}
+
+/***********************************************************************************/
+/***********************************************************************************/
+
+template<class TEntity>
+const TEntity& EntitityIdentifier<TEntity>::GetPrototypeEntity(typename GeometryType::Pointer pGeometry)
+{
+    return *mTypes[static_cast<std::size_t>(pGeometry->GetGeometryType()) - 1];
```
RiccardoRossi: i would say that the -1 is wrong here

RiccardoRossi (commented):
i think your use of the size is not correct. you have a -1 or -2 that does not belong there, since enums already start at 0

this is the code to check it

~~~cpp
#include <vector>
#include <iostream>

// Type your code here, or load an example.
int main() {

    enum my_enum{
        ZERO,
        ONE,
        TWO,
        SIZE
    };

    std::cout << "ZERO=" << static_cast<int>(my_enum::ZERO) << std::endl;
    std::cout << "ONE=" << static_cast<int>(my_enum::ONE) << std::endl;
    std::cout << "TWO=" << static_cast<int>(my_enum::TWO) << std::endl;
    std::cout << "SIZE=" << static_cast<int>(my_enum::SIZE) << std::endl;

    return 0;
}~~~

loumalouomega:
> i think your use of the size is not correct. you have a -1 or -2 that does not belong there, since enums already start at 0
> 
> this is the code to check it
> 
> ```c++
> #include <vector>
> #include <iostream>
> 
> // Type your code here, or load an example.
> int main() {
>     auto pa = new B;
> 
>     enum my_enum{
>         ZERO,
>         ONE,
>         TWO,
>         SIZE
>     };
> 
>     std::cout << "ZERO=" << static_cast<int>(my_enum::ZERO) << std::endl;
>     std::cout << "ONE=" << static_cast<int>(my_enum::ONE) << std::endl;
>     std::cout << "TWO=" << static_cast<int>(my_enum::TWO) << std::endl;
>     std::cout << "SIZE=" << static_cast<int>(my_enum::SIZE) << std::endl;
> 
>     return 0;
> }~~~
> ```

We are considering the following enum:

~~~c++
enum class KratosGeometryType
    {
        Kratos_generic_type,
        Kratos_Hexahedra3D20,
        Kratos_Hexahedra3D27,
        Kratos_Hexahedra3D8,
        Kratos_Prism3D15,
        Kratos_Prism3D6,
        Kratos_Pyramid3D13,
        Kratos_Pyramid3D5,
        Kratos_Quadrilateral2D4,
        Kratos_Quadrilateral2D8,
        Kratos_Quadrilateral2D9,
        Kratos_Quadrilateral3D4,
        Kratos_Quadrilateral3D8,
        Kratos_Quadrilateral3D9,
        Kratos_Tetrahedra3D10,
        Kratos_Tetrahedra3D4,
        Kratos_Triangle2D3,
        Kratos_Triangle2D6,
        Kratos_Triangle2D10,
        Kratos_Triangle2D15,
        Kratos_Triangle3D3,
        Kratos_Triangle3D6,
        Kratos_Line2D2,
        Kratos_Line2D3,
        Kratos_Line2D4,
        Kratos_Line2D5,
        Kratos_Line3D2,
        Kratos_Line3D3,
        Kratos_Point2D,
        Kratos_Point3D,
        Kratos_Sphere3D1,
        Kratos_Nurbs_Curve,
        Kratos_Nurbs_Surface,
        Kratos_Nurbs_Volume,
        Kratos_Nurbs_Curve_On_Surface,
        Kratos_Surface_In_Nurbs_Volume,
        Kratos_Brep_Curve,
        Kratos_Brep_Surface,
        Kratos_Brep_Curve_On_Surface,
        Kratos_Quadrature_Point_Geometry,
        Kratos_Coupling_Geometry,
        Kratos_Quadrature_Point_Curve_On_Surface_Geometry,
        Kratos_Quadrature_Point_Surface_In_Volume_Geometry,
        NumberOfGeometryTypes // Note that this entry needs to be always the last to be used as geometry types counter
    };
~~~

See https://github.com/KratosMultiphysics/Kratos/blob/6c0cdd1fbb5f1e4e5345a86691da7de9d48697fe/kratos/geometries/geometry_data.h#L109C5-L155C7

In order to save 2 pointers of memory, we don't consider `Kratos_generic_type` (the zero, that's why the -1) and `NumberOfGeometryTypes`.

kratos/utilities/entities_utilities.cpp
loumalouomega: See my response

kratos/utilities/entities_utilities.h
loumalouomega: See my response

kratos/utilities/entities_utilities.cpp
loumalouomega: See my response

RiccardoRossi:
> > i think your use of the size is not correct. you have a -1 or -2 that does not belong there, since enums already start at 0
> > this is the code to check it
> > ```c++
> > #include <vector>
> > #include <iostream>
> > 
> > // Type your code here, or load an example.
> > int main() {
> >     auto pa = new B;
> > 
> >     enum my_enum{
> >         ZERO,
> >         ONE,
> >         TWO,
> >         SIZE
> >     };
> > 
> >     std::cout << "ZERO=" << static_cast<int>(my_enum::ZERO) << std::endl;
> >     std::cout << "ONE=" << static_cast<int>(my_enum::ONE) << std::endl;
> >     std::cout << "TWO=" << static_cast<int>(my_enum::TWO) << std::endl;
> >     std::cout << "SIZE=" << static_cast<int>(my_enum::SIZE) << std::endl;
> > 
> >     return 0;
> > }~~~
> > ```
> 
> We are considering the following enum:
> 
> ```c++
> enum class KratosGeometryType
>     {
>         Kratos_generic_type,
>         Kratos_Hexahedra3D20,
>         Kratos_Hexahedra3D27,
>         Kratos_Hexahedra3D8,
>         Kratos_Prism3D15,
>         Kratos_Prism3D6,
>         Kratos_Pyramid3D13,
>         Kratos_Pyramid3D5,
>         Kratos_Quadrilateral2D4,
>         Kratos_Quadrilateral2D8,
>         Kratos_Quadrilateral2D9,
>         Kratos_Quadrilateral3D4,
>         Kratos_Quadrilateral3D8,
>         Kratos_Quadrilateral3D9,
>         Kratos_Tetrahedra3D10,
>         Kratos_Tetrahedra3D4,
>         Kratos_Triangle2D3,
>         Kratos_Triangle2D6,
>         Kratos_Triangle2D10,
>         Kratos_Triangle2D15,
>         Kratos_Triangle3D3,
>         Kratos_Triangle3D6,
>         Kratos_Line2D2,
>         Kratos_Line2D3,
>         Kratos_Line2D4,
>         Kratos_Line2D5,
>         Kratos_Line3D2,
>         Kratos_Line3D3,
>         Kratos_Point2D,
>         Kratos_Point3D,
>         Kratos_Sphere3D1,
>         Kratos_Nurbs_Curve,
>         Kratos_Nurbs_Surface,
>         Kratos_Nurbs_Volume,
>         Kratos_Nurbs_Curve_On_Surface,
>         Kratos_Surface_In_Nurbs_Volume,
>         Kratos_Brep_Curve,
>         Kratos_Brep_Surface,
>         Kratos_Brep_Curve_On_Surface,
>         Kratos_Quadrature_Point_Geometry,
>         Kratos_Coupling_Geometry,
>         Kratos_Quadrature_Point_Curve_On_Surface_Geometry,
>         Kratos_Quadrature_Point_Surface_In_Volume_Geometry,
>         NumberOfGeometryTypes // Note that this entry needs to be always the last to be used as geometry types counter
>     };
> ```
> 
> See https://github.com/KratosMultiphysics/Kratos/blob/6c0cdd1fbb5f1e4e5345a86691da7de9d48697fe/kratos/geometries/geometry_data.h#L109C5-L155C7
> 
> In order to save 2 pointers of memory, we don't consider `Kratos_generic_type` (the zero, that's why the -1) and `NumberOfGeometryTypes`.

well, you could have an element with a base geometry type (a cloud of nodes in mpm for example).

I would not do this simplification to allow for that, and also as in my opinion it obfuscates a bit the code. (for a price of 1pointer!?!)

If you are strongly against my comment, please at least document this very clearly in the code

loumalouomega:
> > > i think your use of the size is not correct. you have a -1 or -2 that does not belong there, since enums already start at 0
> > > this is the code to check it
> > > ```c++
> > > #include <vector>
> > > #include <iostream>
> > > 
> > > // Type your code here, or load an example.
> > > int main() {
> > >     auto pa = new B;
> > > 
> > >     enum my_enum{
> > >         ZERO,
> > >         ONE,
> > >         TWO,
> > >         SIZE
> > >     };
> > > 
> > >     std::cout << "ZERO=" << static_cast<int>(my_enum::ZERO) << std::endl;
> > >     std::cout << "ONE=" << static_cast<int>(my_enum::ONE) << std::endl;
> > >     std::cout << "TWO=" << static_cast<int>(my_enum::TWO) << std::endl;
> > >     std::cout << "SIZE=" << static_cast<int>(my_enum::SIZE) << std::endl;
> > > 
> > >     return 0;
> > > }~~~
> > > ```
> > 
> > 
> > We are considering the following enum:
> > ```c++
> > enum class KratosGeometryType
> >     {
> >         Kratos_generic_type,
> >         Kratos_Hexahedra3D20,
> >         Kratos_Hexahedra3D27,
> >         Kratos_Hexahedra3D8,
> >         Kratos_Prism3D15,
> >         Kratos_Prism3D6,
> >         Kratos_Pyramid3D13,
> >         Kratos_Pyramid3D5,
> >         Kratos_Quadrilateral2D4,
> >         Kratos_Quadrilateral2D8,
> >         Kratos_Quadrilateral2D9,
> >         Kratos_Quadrilateral3D4,
> >         Kratos_Quadrilateral3D8,
> >         Kratos_Quadrilateral3D9,
> >         Kratos_Tetrahedra3D10,
> >         Kratos_Tetrahedra3D4,
> >         Kratos_Triangle2D3,
> >         Kratos_Triangle2D6,
> >         Kratos_Triangle2D10,
> >         Kratos_Triangle2D15,
> >         Kratos_Triangle3D3,
> >         Kratos_Triangle3D6,
> >         Kratos_Line2D2,
> >         Kratos_Line2D3,
> >         Kratos_Line2D4,
> >         Kratos_Line2D5,
> >         Kratos_Line3D2,
> >         Kratos_Line3D3,
> >         Kratos_Point2D,
> >         Kratos_Point3D,
> >         Kratos_Sphere3D1,
> >         Kratos_Nurbs_Curve,
> >         Kratos_Nurbs_Surface,
> >         Kratos_Nurbs_Volume,
> >         Kratos_Nurbs_Curve_On_Surface,
> >         Kratos_Surface_In_Nurbs_Volume,
> >         Kratos_Brep_Curve,
> >         Kratos_Brep_Surface,
> >         Kratos_Brep_Curve_On_Surface,
> >         Kratos_Quadrature_Point_Geometry,
> >         Kratos_Coupling_Geometry,
> >         Kratos_Quadrature_Point_Curve_On_Surface_Geometry,
> >         Kratos_Quadrature_Point_Surface_In_Volume_Geometry,
> >         NumberOfGeometryTypes // Note that this entry needs to be always the last to be used as geometry types counter
> >     };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > See [`6c0cdd1`/kratos/geometries/geometry_data.h#L109C5-L155C7](https://github.com/KratosMultiphysics/Kratos/blob/6c0cdd1fbb5f1e4e5345a86691da7de9d48697fe/kratos/geometries/geometry_data.h#L109C5-L155C7)
> > In order to save 2 pointers of memory, we don't consider `Kratos_generic_type` (the zero, that's why the -1) and `NumberOfGeometryTypes`.
> 
> well, you could have an element with a base geometry type (a cloud of nodes in mpm for example).
> 
> I would not do this simplification to allow for that, and also as in my opinion it obfuscates a bit the code. (for a price of 1pointer!?!)
> 
> If you are strongly against my comment, please at least document this very clearly in the code

Okay good point, updating

loumalouomega:
> > > > i think your use of the size is not correct. you have a -1 or -2 that does not belong there, since enums already start at 0
> > > > this is the code to check it
> > > > ```c++
> > > > #include <vector>
> > > > #include <iostream>
> > > > 
> > > > // Type your code here, or load an example.
> > > > int main() {
> > > >     auto pa = new B;
> > > > 
> > > >     enum my_enum{
> > > >         ZERO,
> > > >         ONE,
> > > >         TWO,
> > > >         SIZE
> > > >     };
> > > > 
> > > >     std::cout << "ZERO=" << static_cast<int>(my_enum::ZERO) << std::endl;
> > > >     std::cout << "ONE=" << static_cast<int>(my_enum::ONE) << std::endl;
> > > >     std::cout << "TWO=" << static_cast<int>(my_enum::TWO) << std::endl;
> > > >     std::cout << "SIZE=" << static_cast<int>(my_enum::SIZE) << std::endl;
> > > > 
> > > >     return 0;
> > > > }~~~
> > > > ```
> > > 
> > > 
> > > We are considering the following enum:
> > > ```c++
> > > enum class KratosGeometryType
> > >     {
> > >         Kratos_generic_type,
> > >         Kratos_Hexahedra3D20,
> > >         Kratos_Hexahedra3D27,
> > >         Kratos_Hexahedra3D8,
> > >         Kratos_Prism3D15,
> > >         Kratos_Prism3D6,
> > >         Kratos_Pyramid3D13,
> > >         Kratos_Pyramid3D5,
> > >         Kratos_Quadrilateral2D4,
> > >         Kratos_Quadrilateral2D8,
> > >         Kratos_Quadrilateral2D9,
> > >         Kratos_Quadrilateral3D4,
> > >         Kratos_Quadrilateral3D8,
> > >         Kratos_Quadrilateral3D9,
> > >         Kratos_Tetrahedra3D10,
> > >         Kratos_Tetrahedra3D4,
> > >         Kratos_Triangle2D3,
> > >         Kratos_Triangle2D6,
> > >         Kratos_Triangle2D10,
> > >         Kratos_Triangle2D15,
> > >         Kratos_Triangle3D3,
> > >         Kratos_Triangle3D6,
> > >         Kratos_Line2D2,
> > >         Kratos_Line2D3,
> > >         Kratos_Line2D4,
> > >         Kratos_Line2D5,
> > >         Kratos_Line3D2,
> > >         Kratos_Line3D3,
> > >         Kratos_Point2D,
> > >         Kratos_Point3D,
> > >         Kratos_Sphere3D1,
> > >         Kratos_Nurbs_Curve,
> > >         Kratos_Nurbs_Surface,
> > >         Kratos_Nurbs_Volume,
> > >         Kratos_Nurbs_Curve_On_Surface,
> > >         Kratos_Surface_In_Nurbs_Volume,
> > >         Kratos_Brep_Curve,
> > >         Kratos_Brep_Surface,
> > >         Kratos_Brep_Curve_On_Surface,
> > >         Kratos_Quadrature_Point_Geometry,
> > >         Kratos_Coupling_Geometry,
> > >         Kratos_Quadrature_Point_Curve_On_Surface_Geometry,
> > >         Kratos_Quadrature_Point_Surface_In_Volume_Geometry,
> > >         NumberOfGeometryTypes // Note that this entry needs to be always the last to be used as geometry types counter
> > >     };
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > See [`6c0cdd1`/kratos/geometries/geometry_data.h#L109C5-L155C7](https://github.com/KratosMultiphysics/Kratos/blob/6c0cdd1fbb5f1e4e5345a86691da7de9d48697fe/kratos/geometries/geometry_data.h#L109C5-L155C7)
> > > In order to save 2 pointers of memory, we don't consider `Kratos_generic_type` (the zero, that's why the -1) and `NumberOfGeometryTypes`.
> > 
> > 
> > well, you could have an element with a base geometry type (a cloud of nodes in mpm for example).
> > I would not do this simplification to allow for that, and also as in my opinion it obfuscates a bit the code. (for a price of 1pointer!?!)
> > If you are strongly against my comment, please at least document this very clearly in the code
> 
> Okay good point, updating

Done @RiccardoRossi

PR #11382, opened 2023-07-12
Do not use depth serializer on gp

roigcarlo:
**📝 Description**
Prevents the new Project class from serializing `gobal_pointer` contents as it would cause deadlock (Same as restart).

**🆕 Changelog**
- Global Pointer contents are now not serialized during checkpoint.

### DIFF
--- a/kratos/python_scripts/project.py
+++ b/kratos/python_scripts/project.py
@@ -79,6 +79,8 @@ class Project:
         with open(save_folder_path / checkpoint_file_path, 'wb+') as checkpoint_file:
             # Serialize current model and stages
             serializer = KratosMultiphysics.StreamSerializer()
+            serializer.Set(KratosMultiphysics.Serializer.SHALLOW_GLOBAL_POINTERS_SERIALIZATION)
+            
             serializer.Save("Model", self.__model)
             stage_names_list = []
             stage_instances_list = []


RiccardoRossi:
why?

roigcarlo:
Because of the `NEIGHBOR_NODES`/`NEIGHBOUR_ELEMENTS` variable:

Imagine you have two nodes, all of them contain `NEIGHBOR_NODES` pointing to the other.

When we are serializing, if we follow the global_pointer serialization code:

```c++
if(rSerializer.Is(Serializer::SHALLOW_GLOBAL_POINTERS_SERIALIZATION))
{
    rSerializer.save("D", reinterpret_cast<std::size_t>(mDataPointer));
}
else
{
    rSerializer.save("D", mDataPointer);
}
```

If `SHALLOW_GLOBAL_POINTERS_SERIALIZATION` is not set the `serializer` will try to serialize `mDataPointer` (which is a `GlobalPointerVectorSet`), which will repeat the process.

Now, if we start from the begining

- Serialization of N1 begins
- Serialization of N1.NEIGHBOUR_NODES begin
- Serialization of N2 begins (as is the first node from NEIGBOUR_NODES from n1)
- Serialization of N2.NEIGHBOUR_NODES begin
Here N1 has not been fully serialized so its @ does not appear in the pointer table of the serializer yet.
- Serialization of N1 begins (as is the first node from NEIGBOUR_NODES from n1)
- Repeat infinitely.

RiccardoRossi (approved):

PR #11383, opened 2023-07-12
[CoSim] Minor Performance Improvement for MVQN Accelerator 

matekelemen:
Replace matrix assignment using python loops by a slice-based numpy assignment.

_Also, added type hints for the matrices participating in the assignment._


loumalouomega (commented):
I would approve, but code is not mine

philbucher (approved):
I trust you :)

PR #11386, opened 2023-07-13
fix computation of the derivative of the equivalent strain

dbcavalcanti:
**📝 Description**
This PR fixes a bug in the computation of the derivative of the equivalent strain in the cohesive isotropic damage 3D model in the Poromechanics application.

**🆕 Changelog**
- Fix computation of the derivative of the equivalent strain in the 3D cohesive isotropic damage model to avoid division by zero.

### DIFF
--- a/applications/PoromechanicsApplication/custom_constitutive/isotropic_damage_cohesive_3D_law.cpp
+++ b/applications/PoromechanicsApplication/custom_constitutive/isotropic_damage_cohesive_3D_law.cpp
@@ -163,10 +163,12 @@ void IsotropicDamageCohesive3DLaw::FinalizeMaterialResponseCauchy (Parameters& r
 }
 
 //----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
-
-Vector& IsotropicDamageCohesive3DLaw::GetValue( const Variable<Vector>& rThisVariable, Vector& rValue )
+double& IsotropicDamageCohesive3DLaw::GetValue( const Variable<double>& rThisVariable, double& rValue )
 {
-    rValue = mStateVariable;
+    if( rThisVariable == DAMAGE_VARIABLE )
+    {
+        rValue = mDamageVariable;
+    }
     return( rValue );
 }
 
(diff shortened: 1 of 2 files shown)


ipouplana (approved):
Nice! Thank you.

ipouplana (changes requested):
GetValue of damage variable is missing

dbcavalcanti:
The damage variable is now declared a member variable in the cohesive isotropic constitutive law. This was done to update the GetValue method to allow plotting in GiD.

ipouplana (approved):
Perfect Danilo, thanks!

applications/PoromechanicsApplication/custom_constitutive/isotropic_damage_cohesive_3D_law.cpp
```diff
@@ -276,8 +278,11 @@ void IsotropicDamageCohesive3DLaw::ComputeEquivalentStrain(ConstitutiveLawVariab
     rVariables.OldEquivalentStrain = mOldStateVariable[1] + rVariables.BetaEqStrainShearFactor * mOldStateVariable[0];
 
     // Compute the vector with the derivatives of the equivalent strain wrt to the components of the strain vector
-    rVariables.DerivativeEquivalentStrain[0] = rVariables.BetaEqStrainShearFactor * StrainVector[0] / resShear;
-    rVariables.DerivativeEquivalentStrain[1] = rVariables.BetaEqStrainShearFactor * StrainVector[1] / resShear;
+    rVariables.DerivativeEquivalentStrain = ZeroVector(3);
+    if (resShear > 0.0) {
+        rVariables.DerivativeEquivalentStrain[0] = rVariables.BetaEqStrainShearFactor * StrainVector[0] / resShear;
+        rVariables.DerivativeEquivalentStrain[1] = rVariables.BetaEqStrainShearFactor * StrainVector[1] / resShear;
+    }
```
matekelemen, line 285: I'm not familiar with the physics here, but what happens if `resShear == 0.0`? Is the model still valid? If not, this just silently swallows the error which is going to be a nightmare to debug when you only see results that don't make sense even though the program didn't crash / throw an exception.

dbcavalcanti: What this PR is doing is exactly solving the error when the resShear is equal to zero. For this case, the components of the vector DerivativeEquivalentStrain will keep with their initial value, as zero (which is not a problem).

matekelemen: > For this case, the components of the vector DerivativeEquivalentStrain will keep with their initial value, as zero (which is not a problem).

allright, sorry for the noise

PR #11390, opened 2023-07-17
[GeoMechanicsApplication] Removed default implementations of special member functions

avdg81:
**📝 Description**
Removed default implementations of special member functions. The scope of these improvements was limited to custom processes.


WPK4FEM (approved):
Consistent removal of default destructors.

PR #11395, opened 2023-07-18
[TrilinosApplication][SeveralApplications] Giving support for new TrilinosBlockBuilderAndSolvers in solvers

loumalouomega:
**📝 Description**

 Giving support for new `TrilinosBlockBuilderAndSolver` in solvers. In addition to that, support for all serial B&S has been added when missing in the other applications.

This required to add new default settings, and the corresponding validation as well.

**🆕 Changelog**

- [Update the structural solver](https://github.com/KratosMultiphysics/Kratos/commit/587f3174e1069ca751cb741960b73333df325cde)
- [Update convention diffusion solver](https://github.com/KratosMultiphysics/Kratos/commit/60e12d8d3ad395ee74af48fb394da93e9b7237d5)
- [Update fluid scripts](https://github.com/KratosMultiphysics/Kratos/commit/1f33320d6cf99c0dbe18a56844c17f7236d29c2c)


applications/FluidDynamicsApplication/python_scripts/trilinos_navier_stokes_solver_vmsmonolithic.py
```diff
@@ -131,12 +131,19 @@ def _CreateBuilderAndSolver(self):
                 epetra_communicator,
                 guess_row_size,
                 trilinos_linear_solver,
-                KratosFluid.PATCH_INDEX)
+                KratosCFD.PATCH_INDEX)
```
loumalouomega: This was wrong before, maybe I can cherry-pick the fix if needed

loumalouomega:
Maybe we can discuss defining some common utilities in order to reduce the code duplications between solvers.

applications/StructuralMechanicsApplication/python_scripts/trilinos_structural_mechanics_solver.py
```diff
@@ -80,26 +79,23 @@ def _CreateLinearSolver(self):
...
+            bs_params = self.settings["builder_and_solver_settings"]["advanced_settings"]
+            bs_params.AddEmptyValue("guess_row_size")
+            bs_params["guess_row_size"].SetInt(guess_row_size)
+            builder_and_solver = TrilinosApplication.TrilinosBlockBuilderAndSolver(epetra_communicator, linear_solver, bs_params)
         else:
-            builder_and_solver = TrilinosApplication.TrilinosEliminationBuilderAndSolver(epetra_communicator,
-                                                                                         guess_row_size,
-                                                                                         linear_solver)
+            if self.settings["multi_point_constraints_used"].GetBool():
+                raise Exception("MPCs not yet implemented in MPI")
```
philbucher, line 95: you added this in the meantime no?

applications/ConvectionDiffusionApplication/python_scripts/convection_diffusion_solver.py
```diff
@@ -128,9 +128,14 @@ def GetDefaultParameters(cls):
             "gradient_dofs" : false,
             "line_search": false,
             "compute_reactions": true,
-            "block_builder": true,
+            "builder_and_solver_settings" : {
+                "use_block_builder" : true,
+                "use_lagrange_BS"   : false,
```
rubenzorrilla, line 133:
```suggestion
                "use_lagrange_builder_and_solver"   : false,
```
to follow the convention. Indeed, we are already inside the B&S settings so I'd put
```suggestion
                "use_lagrange_multipliers"   : false,
```
to avoid being redundant.

rubenzorrilla (commented):
Just saying that thinking in the long run, introducing the B&S settings in the solver ones is not the best solution. Instad, I'd only set the B&S name (i.e. registro points) and take these from the registry. Of course this is out of this PR context.

rubenzorrilla (changes requested):
Why adding the Lagrange BS to all the CFD and Heat solvers which don't actually use it?

applications/ConvectionDiffusionApplication/python_scripts/convection_diffusion_solver.py
loumalouomega: Then we should change it in the StructuralApplication

loumalouomega:
> Why adding the Lagrange BS to all the CFD and Heat solvers which don't actually use it?

Is not used now, but can be used, it is physics agnostic. Quant més sucre més dolç :)

loumalouomega:
> > Why adding the Lagrange BS to all the CFD and Heat solvers which don't actually use it?
> 
> Is not used now, but can be used, it is physics agnostic. Quant més sucre més dolç :)

Hello?

applications/ConvectionDiffusionApplication/python_scripts/convection_diffusion_solver.py
RiccardoRossi: what does it mean "use_lagrange_bs"?

loumalouomega: use_lagrange_builder_and_solver. I t is the key currently used in the StructuralMechanicsApplication

RiccardoRossi: i was referring to "what does that BS do?"

loumalouomega: It was introduced several years ago, it solves using LM for constraints instead of the Felippa algoritihm.

loumalouomega:
@rubenzorrilla @RiccardoRossi we should retake this

rubenzorrilla:
> @rubenzorrilla @RiccardoRossi we should retake this

Maybe is worth to keep it in standby a bit longer considering that we are to change the strategies to get rid of the B&S.

loumalouomega:
> > @rubenzorrilla @RiccardoRossi we should retake this
> 
> Maybe is worth to keep it in standby a bit longer considering that we are to change the strategies to get rid of the B&S.

Well, but do we have any planned change?, because once I fix this PR the changes will be basically some python scripts and it will not affect current workflow and will add a feature that may be handy for some users/developers. It is like not adding any element because we are going to eventually replace the matrix assembly, I think we may think in consistent time intervals.

RiccardoRossi:
hi @loumalouomega Is this urgent? I am asking since i am in a peak of work until the end of november, after which i hope to be a bit more relaxed. Do you need this before then?

loumalouomega:
> hi @loumalouomega Is this urgent? I am asking since i am in a peak of work until the end of november, after which i hope to be a bit more relaxed. Do you need this before then?

It is not urgent at all, it is older than 1 year, but I would like to merge it so I can forget it about

RiccardoRossi:
then let's please postpone the discussion to the end of november ... i am seriously overwhelmed right now

RiccardoRossi:
i took a quick look into it. 

To be honest i think **you do not need my review** provided that people of the applications you modify are on board. I can take a second look if you consider it is needed, but in principle i don't think it contains nothing confilictive.

My only comment is that i do not really understand what is the use case for use_lagrange_BS in the context of CFD...but i leave this to @rubenzorrilla and or @philbucher

PR #11407, opened 2023-07-21
Addition of AssignAverageMasterSlaveConstraintsProcess and Associated Unit Test 

SADPR:
## Description:

This Pull Request introduces the `AssignAverageMasterSlaveConstraintsProcess` to KratosMultiphysics. This process is designed to assign averaged multi-point constraints (MPCs) between a slave and a master model part in a computing model part. This feature has been developed in Python, and it's included in the `kratos/python_scripts/assign_average_master_slave_constraints_process.py` file.

### Files Changed:

- `kratos/python_scripts/assign_average_master_slave_constraints_process.py`: Contains the newly introduced process.
- `kratos/tests/test_processes.py`: Updated to include the unit test for the newly introduced process.
- `kratos/tests/auxiliar_files_for_python_unittest/mdpa_files/test_assign_average_master_slave_constraints_process.mdpa`: Auxiliary file used in the unit test for the new process.

The process is implemented in Python at this moment due to immediate project deliverable requirements. It is designed for simulations where the process is only performed once, keeping the computational cost to a minimum. Nonetheless, an optimized C++ version is planned for future updates, which will leverage Kratos's parallel utilities for enhanced performance.

The new feature has been appropriately tested to ensure it works as expected under various circumstances. The provided unit test validates the correct application of the process.

This feature enriches KratosMultiphysics' capabilities for managing constraints, opening the door for new modelling opportunities.

**Note: Further testing and potential enhancements are encouraged for optimal integration with different use cases.**


loumalouomega:
Okay from my side (I reduced size of your mdpa)

kratos/python_scripts/assign_average_master_slave_constraints_process.py
```diff
@@ -0,0 +1,92 @@
...
+        # If constraints are reformed at each step, remove them at the end of each step
+        if self.reform_constraints_at_each_step:
+            self.__RemoveConstraints()
+    
+    def __AverageMultiPointConstraint(self, variable):
+        """Assigns constraints that make the value of a variable in the slave part
+        to be the average of the values in the master part."""
+        
+        # Prepare containers for the constraint data
+        number_of_master_nodes = self.slave_model_part.NumberOfNodes()
```
Rbravo555: How come `number_of_master_nodes` comes from `slave_model_part.NumberOfNodes()`?

Rbravo555 (commented):
I made a single comment in the code of the Python script.

If I understand correctly, the process will assign MPC of the type

$$u_{slave}^{j} = \sum_{i=1}^{N_{master}}  u_{master}^{i} \omega^{i} $$

Where $$\omega^{i}  = 1/N_{master}$$. Can the process compute weights with a different distribution than just the average with a minor effort? In such a case the process should be named differently.

kratos/python_scripts/assign_average_master_slave_constraints_process.py
SADPR: Good observation, the number_of_master_nodes should indeed be derived from the master model part, not the slave. This was an oversight on my part. Thanks for catching it, I've made the necessary changes.

kratos/python_scripts/assign_average_master_slave_constraints_process.py
```diff
@@ -0,0 +1,92 @@
...
+        # Overwrite the default settings with user-provided ones.
+        settings.ValidateAndAssignDefaults(default_settings)
+        
+        self.model = model
+        self.settings = settings
+
+    def ExecuteInitialize(self):
+        # Store model part names from the settings
+        self.computing_model_part_name = self.settings["computing_model_part_name"].GetString()
+        self.slave_model_part_name = self.settings["slave_model_part_name"].GetString()
```
RiccardoRossi: you actually do not need to save the name in a variable, since once you retrieve the modelpart a few lines below, you can retrieve the name with the method mp.Name()

what i am telling is to only use "slave_model_part_name" without the ".self"

kratos/python_scripts/assign_average_master_slave_constraints_process.py
```diff
@@ -0,0 +1,92 @@
...
+        weights_vector = KM.Matrix(1, number_of_master_nodes)
+        weight = 1/number_of_master_nodes
+        constant_vector = KM.Vector(number_of_master_nodes)
+        
+        counter = 1
+        master_dofs_container = []
+        i = 0
+        
+        # Loop over all master nodes to fill the containers
+        for node in self.master_model_part.Nodes:
```
RiccardoRossi, line 76: why this loop here?

btw, if you want to set a vector or a matrix to a given constant number you could do

~~~py
constant_vector.fill(0.0)
wieghts_vector.fill(weight)
~~~

RiccardoRossi (commented):
minor comments. other than that ok

kratos/python_scripts/assign_average_master_slave_constraints_process.py
SADPR: To retrieve the list of dofs for the correspoding variable.

RiccardoRossi: but what do you do with the i?

RiccardoRossi (dismissed):
approving since my comments are minimal (i think you copute and update the i for nothing in the "why that loop"

kratos/python_scripts/assign_average_master_slave_constraints_process.py
SADPR: Left-over.

SADPR:
@RiccardoRossi I erased the "i" in the loop. It was a left-over due to migrating to ".fill".

Rbravo555 (approved):
LGTM

PR #11409, opened 2023-07-21
[Core] Improve combine model part modeler

ddiezrod:
**📝 Description**
In this modeler we are readding all elements to local mesh in omp when it is not necessary, local mesh in communicator is by default the same object as the normal mesh, so Im skipping these steps in omp.
TODO: improve code in MPI to fill the local, interface meshes etc as it is really slow right now. One option could be call parallel fill communicator directly after this modeler is finished and do not touch anything related to communicators here...

**🆕 Changelog**
- Do not refill local mesh in combine model part modeler
- FIx test and add it to mpi suite

### DIFF
--- a/kratos/tests/test_combine_model_part_modeler.py
+++ b/kratos/tests/test_combine_model_part_modeler.py
@@ -8,7 +8,7 @@ if KratosMultiphysics.IsDistributedRun():
     import KratosMultiphysics.mpi as KratosMPI
     try:
         import KratosMultiphysics.MetisApplication
-    except:
+    except ImportError:
         skip_test = True
 
 def SetParameters(name, use_memory="false"):
--- a/kratos/tests/test_combine_model_part_modeler.py
+++ b/kratos/tests/test_combine_model_part_modeler.py
@@ -3,6 +3,9 @@ import KratosMultiphysics
 import KratosMultiphysics.KratosUnittest as KratosUnittest
 import KratosMultiphysics.kratos_utilities as KratosUtilities
 
+# STL imports
+import pathlib
+
 skip_test = False
 if KratosMultiphysics.IsDistributedRun():
     import KratosMultiphysics.mpi as KratosMPI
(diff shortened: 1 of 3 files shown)


loumalouomega (dismissed):
Okay from my side

kratos/tests/test_combine_model_part_modeler.py
```diff
@@ -30,6 +36,10 @@ def ImportModelPart(model_part, import_settings):
         import_flags = KratosMultiphysics.ModelPartIO.READ
         KratosMultiphysics.ModelPartIO(import_settings['model_import_settings']['input_filename'].GetString(), import_flags).ReadModelPart(model_part)
 
+def GetFilePath(fileName):
+    return os.path.join(os.path.dirname(os.path.realpath(__file__)), fileName)
+
+@KratosUnittest.skipIf(skip_test, "This test requires MetisApplication in MPI runs")
```
philbucher: Can you please use the generic importer fct from testing-utils?

Then this test file can be simplified a loz

kratos/tests/test_combine_model_part_modeler.py
```diff
@@ -30,6 +36,10 @@ def ImportModelPart(model_part, import_settings):
         import_flags = KratosMultiphysics.ModelPartIO.READ
         KratosMultiphysics.ModelPartIO(import_settings['model_import_settings']['input_filename'].GetString(), import_flags).ReadModelPart(model_part)
 
+def GetFilePath(fileName):
+    return os.path.join(os.path.dirname(os.path.realpath(__file__)), fileName)
```
philbucher: Please use pathlib.Path (I can post an example later)

philbucher (commented):
hm seems that I did not submit my review :sweat:

kratos/tests/test_combine_model_part_modeler.py
ddiezrod: Which one do you mean? The code could look better with "CheckIfApplicationsAvailable" instead of try/except.

kratos/tests/test_combine_model_part_modeler.py
ddiezrod: done!

philbucher (dismissed):
perfect!

ddiezrod:
There was an unused import, can you please approve this again? :) @philbucher

philbucher (approved):
FYI, modern python has no place for `os.path`, everything path-related should be done with `pathlib.Path`

ddiezrod:
> FYI, modern python has no place for `os.path`, everything path-related should be done with `pathlib.Path`

I had no idea, thanks for sharing 👍

PR #11412, opened 2023-07-21
[OptApp] Allowing sub-model-parts to work with ModelPartUtils

sunethwarna:
**📝 Description**
This allows sub model parts to be used with `ModelPartUtils`. This is required because, at the `__init__` of the python classes the sub model parts are not prresent hence no model part operations can be done. But we still need to add solution step variables and dofs accordingly at the init before model part is populated. This new class `ModelPartOperation` keeps track of the operation and performs the operation when `GetModelPart` method is called. It will always check whether a similar operation is already performed (hence it only requires a 'suggestive_model_part_name`), if is is already performed, then that model part is returned avoiding performing the same operation.

I updated the `HelmholtzSolver` accordingly, but unfortunately i cannot test it since I don't have proper test cases. Could you check @RezaNajian .

**🆕 Changelog**
- Adds `ModelPartOperation` class to keep track of the model part operations and not to perform the same model part operation multiple times.
- Updated responses and controls
- Added tests to new functionalities
- Updated exisiting tests.

### DIFF
--- a/applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
+++ b/applications/OptimizationApplication/python_scripts/controls/material/material_properties_control.py
@@ -1,10 +1,12 @@
+from typing import Optional
+
 import KratosMultiphysics as Kratos
 import KratosMultiphysics.OptimizationApplication as KratosOA
 from KratosMultiphysics.OptimizationApplication.controls.control import Control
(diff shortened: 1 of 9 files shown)


RezaNajian (dismissed):
@sunethwarna
Thanks a lot.

RezaNajian (approved):

PR #11413, opened 2023-07-21
[Poro] Corrected pressure drop calculation for interface elements

ipouplana:
**📝 Description**
This PR fixes an error in the calculation of the the pressure drop for the interface element.

**🆕 Changelog**
- The method that calculates the gradient of shape function matrix in U_Pw_small_strain_interface_element.cpp has been corrected to properly calculate the pressure drop at the interface.


ipouplana:
We still need to update the results of the consolidation_interface_3D…

dbcavalcanti (approved):
The consolidation_interface_3D test was temporarily removed since that we are still reviewing this element.

PR #11414, opened 2023-07-21
[Core] Making `SpatialSearchResult` derived from `IndexedObject`

loumalouomega:
**📝 Description**

This PR makes improvements to the `SpatialSearchResult` class in the `kratos/spatial_containers/spatial_search_result.h` file. The changes include adding missing include derive `SpatialSearchResult` from `IndexedObject`, updating the class documentation to provide more detailed information about the class and its functionalities, and making some adjustments to constructors and member functions.

**Details:**
- Derive `SpatialSearchResult` from `IndexedObject
- Added missing include statements for `indexed_object.h` and `logger.h`.
- Updated the class documentation to provide a detailed description of the `SpatialSearchResult` class.
- Added a new constructor with an index parameter, which sets the ID of the object if it is derived from `IndexedObject`.
- Modified the existing constructors to initialize the `BaseType` with an ID of 0 if the object is not derived from `IndexedObject`.
- Adjusted the `Set` member function to set the ID of the object if it is derived from `IndexedObject`.
- Adding override in the corresponding methods and update Serializer.

Overall, these changes will simplify using `SpatialSearchResult` in certain implementations which require to be derived from indexed object (for example, using PointerVectorSet).

**🆕 Changelog**

- [Making `SpatialSearchResult` derived from `IndexedObject`](https://github.com/KratosMultiphysics/Kratos/commit/95fb4932b32fea4a85a079ee3b03ea8b50e65886)


pooyan-dadvand:
@loumalouomega why you need to derive from indexed object?

loumalouomega:
> @loumalouomega why you need to derive from indexed object?

In order to be able to use `PointerVectorSet` in #11416, see https://github.com/KratosMultiphysics/Kratos/blob/b506cc128fa7c34c9e53fc0e7c9849877ee93602/kratos/spatial_containers/spatial_search_result_container.h#L62, which is required by the GlobalPointerUtilities in order to keep changes to the minimum. Otherwise I will need to add manually the vector of `SpatialSearchResult` the `GlobalPointerUtilities` in pourpose, with the corresponding debugging, problematics and error prone due to code duplication. This solution is therefore much cleaner.

loumalouomega:
@pooyan-dadvand can you review this, it is till a dependency of future development

loumalouomega:
A priori not needed after speaking with @pooyan-dadvand

loumalouomega:
After second check this is actullay neded :P

loumalouomega:
> After second check this is actullay neded :P

After third discussion I will keep it opened, but maybe it can be skipped with a bit of cunning

RiccardoRossi:
cannot you use the "GlobalPointerVEctor" for the results?

loumalouomega:
> cannot you use the "GlobalPointerVEctor" for the results?

Requires to have an Id the current implementation

loumalouomega:
> the GlobalPointer* continers i would say no... (it relies on a hashing function)
> 
> [`master`/kratos/containers/global_pointers_unordered_map.h](https://github.com/KratosMultiphysics/Kratos/blob/master/kratos/containers/global_pointers_unordered_map.h?rgh-link-date=2023-10-05T14%3A24%3A06Z)
> 
> [`master`/kratos/containers/global_pointers_vector.h](https://github.com/KratosMultiphysics/Kratos/blob/master/kratos/containers/global_pointers_vector.h?rgh-link-date=2023-10-05T14%3A24%3A06Z)

For the utilities it does: https://github.com/KratosMultiphysics/Kratos/blob/0fb47b9a18a0c1539c3d45500c9de2e530e36fd9/kratos/utilities/global_pointer_utilities.h#L266

RiccardoRossi:
not so ... you are only retrieving the ids from there to an std vector ...

RiccardoRossi:
what i am trying to tell is that once you have a global pointer to a given object, the PointerVectorUnorderedMap will work. The utility you pointed to is used to retrieve a global pointer to entities for which you have an Id...but if you do retrieve it on your own you do not need that

loumalouomega:
> what i am trying to tell is that once you have a global pointer to a given object, the PointerVectorUnorderedMap will work. The utility you pointed to is used to retrieve a global pointer to entities for which you have an Id...but if you do retrieve it on your own you do not need that

That's what were are going to do

PR #11417, opened 2023-07-21
[Core] Adding depth shrinking for expressions

sunethwarna:
**📝 Description**
When `ContainerExpression` or `Expression` is used with nested code such as in the following time loop block, it creates a linked lazy expression which does not allow `a` and `b` to be freed. In this case, `b` is a temporary and the time history of 100 steps of `a` is not required. So the lazy expression will keep all `a` and `b`s in memory which will at one point overflow. Hence, this PR introduces `MAX_SHRINKING_DEPTH = 20`. When a new `BinaryExpression` or an `UnaryExpression` is created, the  maxdepth of the child expressions are checked, if any exceed `MAX_SHRINKING_DEPTH`, then the new expression will be created out of shrinked (having max depth = 1) children. **It does not change the original `Expression`**. A warning message is displayed when shrinking is performed.

```python
   a = Kratos.Expression.NodalExpression(model_part)
   for t in range(100):
      a += b # here b is another dense vector different in each iteration
```

FYI: @FMeister

**🆕 Changelog**
- Added `GetMaxDepth` and `GetExpressionMaxDepth` methods to `Expression` and `ContainerExpression`.
- Added `FillUtilizedExpressionSet` to `Expression` to get the unique number of utilized `Expression`s in a given lazy expression.
- Added `GetShrinkedExpression` and `Shrink` methods to `Expression` and `ContainerExpression`.
- Added tests


matekelemen:
I'm sorry, but I don't like this solution: it extends the interface with super confusing functions and makes the behaviour less intuitive to solve a problem that stems from misusing `Expression`s.

- first of all, intensive calculations such as the one in your example should be implemented in a more performant manner, not with `Expression`s. Emitting a warning in these cases won't prevent users from further misuse and giving them additional tools to do so won't end well
- the interface is extended with functions that have nothing to do with the functionality of `Expression` and only make sense to people who are familiar with the implementation. This makes it harder for newcomers to grasp `Expression`s and encourages existing users to rely on implementation details
- avoiding intermediate arrays was one of the original design goals of `Expression`s, but this solution goes directly against that idea
- automatically collapsing the expression to create intermediate ones will make the behaviour confusing to users who aren't intricately familiar with the internals of `Expression`
- tracking the depth of `Expression`s is alright and may come in handy during testing to catch cases such as your example, but ultimately, the problem is memory, and that's exactly what we should measure. I'd make `GetDepth` private and check it internally in debug builds. Furthermore, I'd add `GetAllocatedSize` that aggregates the total amount of **allocated** memory used by the `Expression` and its subexpressions. Instead of warnings, exceptions would be a better fit to discourage users from using `Expression`s where more performant solutions are required.

In my opinion, misusing a feature **should** be difficult and hackish, and we shouldn't encourage it. I'm OK with adding tools (`GetDepth`, `GetAllocatedSize`) to detect and debug problematic cases, but I wouldn't extend `Expression`s to deal with scenarios they weren't meant for.

If you insist on having an interface for creating intermediate `Expression`s for prototyping purposes, I'd suggest to do it in an `ExpressionIO` instead of `Expression` itself.

PR #11420, opened 2023-07-24
[Meshing] do not renumber nodes when remeshing

pablobecker:
**📝 Description**
This MR avoids the renumbering of existing nodes when refining meshes. It was causing problems when used in combination with the connectivity preserve modeler, since some of the nodes of the "cousin" model parts are not seen by the model part used when refining, thus leading to repeated nodal IDs

**🆕 Changelog**
-Removes renumbering
-Adds a map and vector of Ids to move back and forth between model parts and CSR matrix, while keeping performance to reduce the coordinates in the CSR.

### DIFF
--- a/applications/MeshingApplication/custom_utilities/local_refine_triangle_mesh_conditions.hpp
+++ b/applications/MeshingApplication/custom_utilities/local_refine_triangle_mesh_conditions.hpp
@@ -200,13 +200,13 @@ public:
         rCoord.resize(pNodes.size(), pNodes.size(), false);
 
         for(auto i = it_begin; i!=it_end; i++) {
-            int index_i = i->Id() - 1; // WARNING: MESH MUST BE IN ORDER
+            int index_i = mMapNodeIdToPos[i->Id()]; // WARNING: MESH MUST BE IN ORDER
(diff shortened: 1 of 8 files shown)


pablobecker:
ping @RiccardoRossi could you look at this?

RiccardoRossi (dismissed):
well, it should now be slower...but i don't see wrong the change you did.

also if you tried it out and it works for you, fine for me too ...

applications/MeshingApplication/custom_utilities/local_refine_triangle_mesh_conditions.hpp
```diff
@@ -200,13 +200,13 @@ class LocalRefineTriangleMeshConditions
         rCoord.resize(pNodes.size(), pNodes.size(), false);
 
         for(auto i = it_begin; i!=it_end; i++) {
-            int index_i = i->Id() - 1; // WARNING: MESH MUST BE IN ORDER
+            int index_i = mMapNodeIdToPos[i->Id()]; // WARNING: MESH MUST BE IN ORDER
```
ddiezrod:
```suggestion
            int index_i = mMapNodeIdToPos[i->Id()];
```

ddiezrod (approved):

PR #11429, opened 2023-07-25
[GeoMechanics] Modifications on geomechanics_solvers_wrapper.py

Rbravo555:
This PR is the re-opened version of #11322, to which the branch name has been changed to trigger internal tests on the GeoMechanics side

Depend on #11425

This is the **3rd of 3** PRs for using the RomApp with the GeoMechanicsApp 

**📝 Description**
The mechanism inside the RomApp for setting the RomBuilderAndSolver requires the existance of the CreateSolver and CreateSolverByParameters functions.

Notice that this PR still requires some work because the parameters being passed to the CreateSolverByParameters function are in general not the complete parameters contained in the ProjectParameters.json, but only the "solver_settings". Therefore,  **end_time** is a problematic term because it is not contained in the "solver_settings" but in the "problem_data". What I have done here was to set it to 1.0, but we need to find a more robust alternative. 

**🆕 Changelog**
- splitted operations  CreateSolver(model, custom_settings) ==> CreateSolver(model, custom_settings) + CreateSolverByParameters(model, custom_settings, parallelism)


mcgicjn2 (changes requested):
We have recently updated the time stepping procedure #11471 can you merge and then I can re-review this.

Rbravo555:
The time parameter is not used anymore. 

Currently the only hacky part in this PR is the one that accounts for the fact that in the GeoMechanicsApp the complete parameters contained in the ProjectParameters.json are passed, while in other apps (in particular in the RomApp) only the "solver_settings" bit is passed. 

```python
if custom_settings.Has("solver_settings"):
    params = custom_settings["solver_settings"]
    # To conform with other apps where only the "solver_settings" are passed
else:
    params = custom_settings

PR #11421, opened 2023-07-24
[OptApp] Adding Barzilai-Borwein line search

Igarizza:
**📝 Description**
Adds Barzilai-Borwein line search that can be combined with SD and GP (after merging). 

**🆕 Changelog**
- Adding new unit test to the test_line_search
- Line_search utilities are extended

PR #11431, opened 2023-07-26
Optapp/ Adding QNBB line search technique

Igarizza:
**📝 Description**
PR implements the possibility to combine SD with QN-BB method, where the step size is computed for each design variable. New function Compute Step Size is returning collective expression with different step sizes.

**🆕 Changelog**
The PR is based on the PR #11421 
E.g.
- Added QNBB method
- New unittests and analysis based test


Igarizza:
@sunethwarna I am not sure if it is a good solution to convert all line searches to compute vector of constant step values. The reason is that typically, line search returns single value. QNBB is not a typical case and may be it is better to keep it as a unique case and do not convert other line search techniques in the same manner. What do you think? I am open for any suggestions.

applications/OptimizationApplication/python_scripts/algorithms/algorithm_steepest_descent.py
```diff
@@ -88,8 +89,12 @@ def ComputeSearchDirection(self, obj_grad) -> KratosOA.CollectiveExpression:
         return search_direction
 
     @time_decorator()
-    def ComputeControlUpdate(self, alpha) -> KratosOA.CollectiveExpression:
-        update = self.algorithm_data.GetBufferedData()["search_direction"] * alpha
+    def ComputeControlUpdate(self, alpha):
+        search_direction = self.algorithm_data.GetBufferedData()["search_direction"]
+        if type(alpha) is float:
```
matekelemen:
```suggestion
        if isinstance(alpha, float):
```

applications/OptimizationApplication/python_scripts/algorithms/algorithm_steepest_descent.py
```diff
@@ -88,8 +89,12 @@ def ComputeSearchDirection(self, obj_grad) -> KratosOA.CollectiveExpression:
...
+            update = search_direction * alpha
+        elif type(alpha) is KratosOA.CollectiveExpression:
+            update = search_direction.Scale(alpha)
```
matekelemen, line 97: it might make sense to define an overload of `Scale` that takes a scalar argument (identical to `operator*(double)`) to avoid branches like this.

What do you think @sunethwarna ?

sunethwarna: Hmm... good idea... i think it makes sense :D... Lets do it :)

sunethwarna (dismissed):
Thanks @Igarizza :)

Could you fix the codacy suggestions? They are valid for this PR.

applications/OptimizationApplication/python_scripts/utilities/opt_line_search.py
```diff
@@ -27,38 +27,100 @@ def GetDefaultParameters(cls):
...
+        self._max_step = parameters["max_step"].GetDouble()
+        self._optimization_problem = optimization_problem
+        self._gradient_scaling = parameters["gradient_scaling"].GetString()
+
+    @time_decorator()
+    def ComputeStep(self) -> KratosOA.CollectiveExpression:
+        algorithm_buffered_data = ComponentDataView("algorithm", self._optimization_problem).GetBufferedData()
+        norm = self.ComputeScaleFactor()
+        if self._optimization_problem.GetStep() == 0:
+            self.unscaled_step = self._init_step
```
sunethwarna, line 95: I don't see `self.unscaled_step` in either the base class or the `BBStep`. I would set it to some default value in the `__init__`.

applications/OptimizationApplication/python_scripts/utilities/opt_line_search.py
```diff
@@ -27,38 +27,100 @@ def GetDefaultParameters(cls):
...
+            current_search_direction = algorithm_buffered_data.GetValue("search_direction", 0)
+            previous_search_direction = algorithm_buffered_data.GetValue("search_direction", 1)
+            y = previous_search_direction - current_search_direction
+            d = algorithm_buffered_data.GetValue("control_field_update", 1)
+            dy = KratosOA.ExpressionUtils.InnerProduct(d,y)
+            dd = KratosOA.ExpressionUtils.InnerProduct(d,d)
+            if not math.isclose(dy, 0.0, abs_tol=1e-16):
+                self.unscaled_step = abs( dd / dy )
+            if math.isclose(dy, 0.0, abs_tol=1e-16):
+                self.unscaled_step = self._max_step
```
sunethwarna:
```suggestion
            if not math.isclose(dy, 0.0, abs_tol=1e-16):
                self.unscaled_step = abs( dd / dy )
            else:
                self.unscaled_step = self._max_step
```

sunethwarna (commented):
Thanks @Igarizza . Minor comments only.

sunethwarna (dismissed):

sunethwarna (approved):

PR #11433, opened 2023-07-26
[Core] Extend DataCommunicator to all the types including dynamic types

sunethwarna:
**📝 Description**
First of all **The Large number of line changes are due to addition of lots of TESTs... (around 3000 lines for testing)** :)

This PR extends all the interfaces of `DataCommunicator` to support following list of variables
1. `array_1d<double, 3>`
2. `array_1d<double, 4>`
3. `array_1d<double, 6>`
4. `array_1d<double, 9>`
5. `Vector`
6. `Matrix`

This also introduces a new method named `DataCommunicator::SynchronizeShape(TDataType& rValue)` for all the above data types as well as all the primitive types. This is used to synchronize the shape of the given `rValue` across all ranks to the maximum shape size. This is doing nothing for all the data types except for `Vector` and `Matrix` and `std::vector<TDataType>`. This is handy in the cases where we have empty ranks and to know the size of a dynamic data type. (Specially in output like in HDF5, and StatsApp, etc)

Following methods and their overloads are extended:

- [x] `Sum`, `Min`, `Max`
- [x] `SumAll`, `MinAll`, `MaxAll`
- [x] `ScanSum`
- [x] `Scatter`
- [x] `Scatterv`
- [x] `Gether`
- [x] `Gatherv`
- [x] `AllGether`
- [x] `AllGatherv`
- [x] `Broadcast`
- [x] `SendRecive`  

**🆕 Changelog**
- To be updated later.


loumalouomega (commented):
Okay from my side

roigcarlo (approved):
:+1: 

I am pretty confident that nothing breaks with the amount of tests you put in. Great job :D

sunethwarna:
Thanks @roigcarlo :D

kratos/mpi/includes/mpi_message.h
```diff
@@ -86,113 +73,409 @@ template<> struct MPIDataType<char>
...
+    }
+
+    inline int Size(const MessageDataType& rValue)
+    {
+        return rValue.size() * SubDataTypeSize(rValue);
+    }
+
+    inline int SubDataTypeSize(const MessageDataType& rValue)
+    {
+        return rValue.size() == 0 ? 0 : MPIMessage<SubDataType>().Size(rValue.front());
```
jcotela, line 432: I did not have time to review this in depth, but just reading through the code this worried me a bit. Are we sure that we cannot have a situation where the container is empty on some ranks but not on others and we get a size mismatch? If it is only used in the function above (`int Size(...`) it looks fine to me but, again, I didn't do a proper review...

sunethwarna: I thought about that, and it is checked in the message size checks. But unfortunately they are only checked in `DEBUG` mode only.

sunethwarna: @jcotela I would greately appreciate if you can review this, I can put those remaks on a seperate PR. :)

jcotela: @sunethwarna other than what I said above, I only have a couple of comments regarding MPIMessage
- there is a `template<class TDataType> class MPIMessage;` in line 88 of mpi_message.h that is not needed
- Why do you need to instance the MPIMessage class? I would have kept its methods static, since I do not see it manage any data at the moment.

I am quite happy with how this turned out overall, since you managed to fit the new features within the existing framework.

sunethwarna: @jcotela Thanks a lot for the comments. :D. Again, nice to have you back in Kratos :D

> there is a template<class TDataType> class MPIMessage; in line 88 of mpi_message.h that is not needed

Indeed, I missed it. I will fix it :)

> Why do you need to instance the MPIMessage class? I would have kept its methods static, since I do not see it manage any data at the moment.

The `MPIMessage` cannot be static in the case of non-contiguous memory arrays such as `std::vector<array_1d<double, 3>>` or `std::vector<Vector>` which is now supported. I was careful enough to reduce memeory and computational overheads while supporting these non-contiguous vectors. In there, `MPIMessage::Buffer` returns the contiguous memeory for MPI communication (if it is contiguous, then it does not do anything, returns the original data array. If it is not contiguous, then it creates a contiguous array from the original data and sends that memory location.). So because we now support non-contiguous arrys, there is this `MPIMessage::Update` method, which updates the original non-contiguous arrays given by user with the MPI communicated data.

This is how it looks like in a simple `AllReduceDetail`
https://github.com/KratosMultiphysics/Kratos/blob/972ae53b9f5984bde778088986d45f682eccbc5b/kratos/mpi/sources/mpi_data_communicator.cpp#L635-L659

Data management is done in places like following (Buffer):
https://github.com/KratosMultiphysics/Kratos/blob/972ae53b9f5984bde778088986d45f682eccbc5b/kratos/mpi/includes/mpi_message.h#L367-L394

Update method:
https://github.com/KratosMultiphysics/Kratos/blob/972ae53b9f5984bde778088986d45f682eccbc5b/kratos/mpi/includes/mpi_message.h#L452-L476

jcotela: Ok I see why it is needed. I guess my first preference in this case would have been to implement the buffering and updating explicitly on the caller function, but I can see how this helps and may reduce code duplication.

Looking into #11440  now, I think there is something going on with single-value sends of type Matrix (not std::vector<Matrix>). These are going through the serializer in the current master, but if I try to define a Send for Matrix using the mpi_data_communicator type handling the TestMPIDataCommunicatorSendAndRecvMatrix suddenly fails.

PR #11436, opened 2023-07-27
FindPython Patch for Cmake 3.27

### LINKED ISSUE #11426: Problems Compiling Kratos
Dear all,

I'm trying to compile Kratos, inside poro/test-pullout-new (PoromechanicsApplication) on my PC but I'm having problems.

The environment in which I'm trying to compile Kratos is the following:

- Windows 10, Version 1709, compilation 16299.2166
- Visual Studio 2019
- CMake 3.27.0
- Python 3.9.7
- Boost 1.78.0

I tried to compile Kratos on my laptop by using the exactly same environment expect for the operative system (my laptop uses Windows 11), and it worked well, hence, I have the latest version of the software and I can use it. So, this is why I do not understand where's the error (maybe on the operative system??).

I attache some screenshots to show you the errors I get every time I try to compile. The first two screenshots (1 and 2) are two different warnings I get at the beggining about CMake. Even though I get these messages, it seems that the compilation is working well and the procedure seems normal.

But then, when it feels like the compilation is about to finish, I get the error shown in third screenshot (3) lots of times, and the "compilation" finishes with the error seen in screenshot 4.

If anyone knows what is the problem or can help me fixing it I will really appreciate because I need it working as soon as possible.

Thank you in advance.

roigcarlo:
**📝 Description**
Should fix #11426. 

This is a ugly solution. It appears that we are **very** behind the pybind11. @KratosMultiphysics/technical-committee we should speak about how to proceed, as the build mechanism for pybind11 has changed greatly (It is instructed to add pybind11 as a sub-project rather than compiling through Pybind11Tools)

@xaviertort Could you check if you can build with cmake 3.27 with this fix?

**🆕 Changelog**
- Setting find_package policy C0148 to OLD to avoid problems with newer cmake versions

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
+cmake_minimum_required (VERSION 3.5)
 project (KratosMultiphysics)
-cmake_minimum_required (VERSION 2.8.6)
 set (CMAKE_CXX_STANDARD 17)
 set (CMAKE_CXX_STANDARD_REQUIRED ON)
 set (CMAKE_CXX_EXTENSIONS OFF)
--- a/cmake_modules/FindPythonLibsNew.cmake
+++ b/cmake_modules/FindPythonLibsNew.cmake
@@ -1,5 +1,5 @@
 # - Find python libraries
-# This module finds the libraries corresponding to the Python interpeter
+# This module finds the libraries corresponding to the Python interpreter
 # FindPythonInterp provides.
 # This code sets the following variables:
 #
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
+cmake_minimum_required (VERSION 3.5)
```
matekelemen, line 1: ❤️

matekelemen:
any chance of taking this opportunity to upgrade our other dependencies as well?
- `pybind11`
- `nlohmann_json`
- `cosimio`
- `trilinos`?
- `boost`?
- ...

philbucher:
CoSimIO should be up to date no?

philbucher:
also what about trilinos and boost? There we support also the latest versions no?

roigcarlo:
> any chance of taking this opportunity to upgrade our other dependencies as well?

Agree. We need to plan a session in the @KratosMultiphysics/technical-committee to review outdated dependencies, and set new minimal versions. Adding it to the project.

roigcarlo:
> also what about trilinos and boost? There we support also the latest versions no?

When looking for the 3.27 issue I found that basically the problem is that the old behavior of "find_package" change (now is deprecated), and now it has a series of requirements. So basically make the packages be all capital names (C0144), the python one (C0148), etc...
I have a cmake 3.27 to make tests, but in most cases the problem is more if we are discovering those libs correctly rather if they work with Kratos ( and to decide if we want to keep giving support for the old mechanism). This will insatiably cause to upgrade to versions of the libs that were upgraded to accommodate this.

PR #11439, opened 2023-07-28
[Poro] Fixing joint width for hydraulic problems

ipouplana:
**📝 Description**
We found a bug in the calculation of the joint width which affected the flow within the fractures. We corrected this bug and updated the element tests. We also cleaned up the U_Pw quasi-static and dynamic schemes.

@joaquinirazabal, this fix shouldn't affect the DamApplication behavior but now the joint width will be properly calculated for post-processing. Also bear in mind that now the MINIMUM_JOINT_WIDTH is replaced by the INITIAL_JOINT_WIDTH. In essence, we now allow the joint width to be equal or greater than zero, and the INITIAL_JOINT_WIDTH is used to define joints of width larger than zero even in cases where the geometrical joint width is zero. 

**🆕 Changelog**
- Fixed the CalculateJointWidth methods in Poromechanics and Dam applications
- Renamed MINIMUM_JOINT_WIDTH variable to INITIAL_JOINT_WIDTH
- Renamed TRANSVERSAL_PERMEABILITY variable to TRANSVERSAL_PERMEABILITY_COEFFICIENT (for consistency)
- Cleaned up the U_Pw quasi-static and dynamic schemes
- Added new element tests and cleaned up old ones.


dbcavalcanti (dismissed):
Great.

joaquinirazabal:
Great! @ipouplana, thank you for the information, I will make the proper changes in the GiD Interface.

dbcavalcanti (approved):

PR #11440, opened 2023-07-28
Adding single value send/recv implementation.

jcotela:
**📝 Description**
We are missing single-value send/recv implementations (as opposed to std::vectors of values), but those are being used in the tests, which means that MPIDataCommunicator defaults to the serialized send/recv version and generates a warning.
This PR implements single-value send/recv reusing the existing std::vector implementation.

FYI @sunethwarna, let me know if we should wait to merge this one too (or feel free to merge yourself when you think it is safe)

**🆕 Changelog**
- Add single value send-recv


sunethwarna:
Hi @jcotela, would it be ok if I integrate this change to the #11433 PR?

jcotela:
@sunethwarna as you prefer, either that or we wait until #11433 is merged, as I don't think this is particularly critical

jcotela:
Note: I expect the version after merge to fail some mpi tests (at least it is doing so on my machine), switching to draft until I can look into it some more

sunethwarna:
I am also on this :)

sunethwarna (approved):
Thanks @jcotela ...

sunethwarna:
Apparently still the CI fails with a weird error. My PC, all tests are passing without a problem... :/ I'll also keep checking...

sunethwarna:
Ok, #11458 fixed the problem. :)

jcotela:
Thanks @sunethwarna, merging.

PR #11441, opened 2023-07-28
[Meshing] Replace linear with quad elems only in flagged elems

pablobecker:
**📝 Description**
This PR modifies the behavior of the linear to quadratic utility so that only flagged elems and conditions are replaced.

**🆕 Changelog**
E.g.
- Removes flagging all conds and elems from the model part
- Thus user is in charge of setting which elems and conds are to be replaced


jcotela (dismissed):
Looks good, approving.

For users of this: the behavior change comes because it is now the caller's responsibility to mark with SPLIT_ELEMENT the elements to be made quadratic (instead of refining the entire passed model part)

RiccardoRossi:
guys, are u awqre that quad and linear meses are not conforming?

pablobecker:
> guys, are u awqre that quad and linear meses are not conforming? 

Yes, this is why I check in all non-replaced elements  (that is ,linear ones) to make sure they are not touching quadratic edges. If they do, an error is thrown

RiccardoRossi:
well, even if you replace a quafrsticnedge with two linear elements then.
mesh IS still not conforming (a parábola VS two lines).

anyhiw, as long as you have It under control do no problem on my side

jcotela (approved):
re-approving

PR #11446, opened 2023-08-01
[FastPR][Core] Geometries PrintData minor improvement

rubenzorrilla:
**📝 Description**
This adds a minor improvement to the `PrintData` function of the base geometry as well as to the "standard" derived ones. Basically, now it is checked whether the points pointers are null or not to avoid segfaulting when calling the method. If the pointers are null we will be informing the user of the situation. Note that we cannot throw an error as this might be a valid situation (e.g. the geometry prototypes).


loumalouomega (changes requested):
Can be improved by improving the base class and call the base class in the derived classes when required

rubenzorrilla:
> Can be improved by improving the base class and call the base class in the derived classes when required

Not sure at all. Note that the Jacobian might not be defined for all the geometries.

kratos/geometries/geometry.h
```diff
@@ -3840,15 +3840,22 @@ class Geometry
         rOStream << std::endl;
         rOStream << std::endl;
 
+        bool null_points = false;
         for (unsigned int i = 0; i < this->size(); ++i) {
             rOStream << "\tPoint " << i + 1 << "\t : ";
-            mPoints[i].PrintData(rOStream);
+            if (mPoints(i) != nullptr) {
```
philbucher, line 3845: this seems like a big change

are we sure we want to allow null-points?

I think better not, imagine all the checks we would need to introduce :sweat_smile:

rubenzorrilla: We are already allowing them. All the prototype geometries use `nullptr` points.

rubenzorrilla: I mean, I don't mind dropping this, the point is that I ended up spending more time debugging this than the stuff I wanted to :smile:. IMO it makes sense to do this check for the output functions as it might be quite helpful for debugging.

kratos/geometries/hexahedra_3d_20.h
```diff
@@ -878,11 +878,25 @@ template<class TPointType> class Hexahedra3D20 : public Geometry<TPointType>
...
+                break;
+            }
+        }
+
+        // If the geometry has valid points, calculate and output its data
+        if (is_valid) {
+            Matrix jacobian;
+            this->Jacobian( jacobian, PointType() );
+            rOStream << "    Jacobian in the origin\t : " << jacobian;
+        }
```
philbucher: how about adding a `protected` function `AllPointsValid` (perhaps you can come up with a better name :D), and use this here?

```suggestion

        // Check if the geometry has valid points
      if (!this->AllPointsAreValid() {
          return;
          }
        // If the geometry has valid points, calculate and output its data
        Matrix jacobian;
        this->Jacobian( jacobian, PointType() );
        rOStream << "    Jacobian in the origin\t : " << jacobian;
```

the early return makes the code also much more readable IMO

kratos/geometries/geometry.h
philbucher: I see, I made another suggestion :)

kratos/geometries/hexahedra_3d_20.h
rubenzorrilla: That would be great (I also like the naming).

EDIT: In any case, as we're not in a hurry, I'd leave this for the 1st @KratosMultiphysics/technical-committee meeting after the summer break.

kratos/geometries/geometry.h
matekelemen: I know this was implemented a long time ago, but now we have C++17 so there's no *point* in risking segfaults because of naked raw pointers => I suggest wrapping them in an `std::optional`.

If you're worried about messing up the memory alignment, I can come up with a version of `optional` for pointers that has an identical alignment to pointers, so that's not a problem.

rubenzorrilla: I think that even though @matekelemen 's comment is a good point, it will be still good to inform developers of the null points in the `PrintData`. In other words, introducing the `std::optional` when retrieving the points doesn't exclude informing about this situation to the user.

kratos/geometries/hexahedra_3d_20.h
rubenzorrilla: Done.

rubenzorrilla:
Finally I created a protected  `AllPointsAreValid` inquiry method in the base `Geometry`. This is used in the derived geometries to check if the Jacobian has to be added to the standard geometry `PrintData` output as suggested by @philbucher and @loumalouomega.

Note that the base geometry `PrintData` method still checks if the pointers are null or not. If a point is found to be null, this is reported in the output stream.

loumalouomega (commented):
Looks now Okay to me

loumalouomega (dismissed):

kratos/geometries/geometry.h
```diff
@@ -4102,6 +4083,26 @@ class Geometry
...
+    {
+        // Check if the geometry has valid (i.e. not nullptr) points
+        bool is_valid = true;
+        for (IndexType i = 0; i < this->PointsNumber(); ++i) {
+            if (this->pGetPoint(i) == nullptr) {
+                is_valid = false;
+                break;
+            }
+        }
+        return is_valid;
```
philbucher:
```suggestion
        // Check if the geometry has valid (i.e. not nullptr) points
        for (IndexType i = 0; i < this->PointsNumber(); ++i) {
            if (this->pGetPoint(i) == nullptr) {
                return false;
            }
        }
        return true;
```

kratos/geometries/hexahedra_3d_20.h
```diff
@@ -878,11 +878,16 @@ template<class TPointType> class Hexahedra3D20 : public Geometry<TPointType>
...
-        Matrix jacobian;
-        this->Jacobian( jacobian, PointType() );
-        rOStream << "    Jacobian in the origin\t : " << jacobian;
+
+        // If the geometry has valid points, calculate and output its data
+        if (this->AllPointsAreValid()) {
+            Matrix jacobian;
+            this->Jacobian( jacobian, PointType() );
+            rOStream << "    Jacobian in the origin\t : " << jacobian;
+        }
```
philbucher, line 890:
```suggestion
        // If the geometry has valid points, calculate and output its data
        if (!this->AllPointsAreValid()) {return;}
        Matrix jacobian;
        this->Jacobian( jacobian, PointType() );
        rOStream << "    Jacobian in the origin\t : " << jacobian;
```

I recommend using early exits, helps to reduce the indentation

kratos/geometries/geometry.h
philbucher: or maybe even better:

```suggestion
        return std::none_of(mPoints.begin(), mPoints.end(), [](auto& pPoint) { return pPoint == nullptr; }))
```

rubenzorrilla: OFC, copy paste leftover.

rubenzorrilla: Cool. I'll do it like this. Thanks for the hint.

kratos/geometries/hexahedra_3d_20.h
rubenzorrilla: About this, I'd leave it as it is. I think indentation in this case is not a problem as there  is only  a few levels and makes the code more readable IMO.

kratos/geometries/geometry.h
rubenzorrilla: Done.

kratos/geometries/geometry.h
```diff
@@ -4092,15 +4092,7 @@ class Geometry
...
-        // Check if the geometry has valid (i.e. not nullptr) points
-        bool is_valid = true;
-        for (IndexType i = 0; i < this->PointsNumber(); ++i) {
-            if (this->pGetPoint(i) == nullptr) {
-                is_valid = false;
-                break;
-            }
-        }
-        return is_valid;
+        return std::none_of(mPoints.ptr_begin(), mPoints.ptr_end(), [](auto& pPoint){return pPoint == nullptr;});
```
philbucher: great!

thanks to @matekelemen I learned abt this, so much nicer :tada:

philbucher (dismissed):

kratos/geometries/geometry.h
```diff
@@ -4102,6 +4083,18 @@ class Geometry
...
     ///@{
 
+    /**
+     * @brief Checks if the geometry points are valid
+     * Checks if the geometry points are valid from the pointer value
+     * Points are not valid when the pointer value is null
+     * @return true All points are valid
+     * @return false At least one point has nullptr value
+     */
+    bool AllPointsAreValid() const
```
philbucher, line 4093: BTW covering this with a test would be amazing :eyes:

kratos/geometries/geometry.h
```diff
@@ -4102,6 +4083,18 @@ class Geometry
...
+    {
+        return std::none_of(mPoints.ptr_begin(), mPoints.ptr_end(), [](auto& pPoint){return pPoint == nullptr;});
```
philbucher:
```suggestion
        return std::none_of(mPoints.ptr_begin(), mPoints.ptr_end(), [](const auto& pPoint){return pPoint == nullptr;});
```

minor nitpicking :D

kratos/geometries/geometry.h
rubenzorrilla: I tried, but I didn't come up with a simple solution since it is `protected`...

loumalouomega (approved):

PR #11449, opened 2023-08-02
[Iga] Nurbs Modeler Flexible Param Space

manuelmessmer:
**📝 Description**
Added the option to choose a flexible parameter space  (unequal to [0,0,0,][1,1,1]) in the NURBS modeler. 
The json looks the following:
```json
"lower_bound_xyz" : []
"upper_bound_xyz" : []
"lower_bound_uvw" : []
"upper_bound_uvw": []
```

**🆕 Changelog**
- Change CreateAndAddRegularGrid2D().
- Change CreateAndAddRegularGrid2D().
- Added unit test.

### DIFF
--- a/applications/IgaApplication/custom_modelers/nurbs_geometry_modeler.cpp
+++ b/applications/IgaApplication/custom_modelers/nurbs_geometry_modeler.cpp
@@ -109,7 +109,11 @@ namespace Kratos
         const Point& A_uvw, const Point& B_uvw, SizeType OrderU, SizeType OrderV,SizeType NumKnotSpansU, SizeType NumKnotSpansV)
     {
         KRATOS_ERROR_IF( B_xyz.X() <= A_xyz.X() || B_xyz.Y() <= A_xyz.Y() ) << "NurbsGeometryModeler: "
-            << "The two Points A and B must meet the following requirement: (B-A) > (0,0,0). However, (B-A)=" << B_xyz-A_xyz << std::endl;
+            << "The two Points A_xyz and B_xyz must meet the following requirement: (B_xyz-A_xyz) > (0,0,0). However, (B_xyz-A_xyz)=" << B_xyz-A_xyz << std::endl;
+
+        KRATOS_ERROR_IF( B_uvw.X() <= A_uvw.X() || B_uvw.Y() <= A_uvw.Y() ) << "NurbsGeometryModeler: "
+            << "The two Points A_uvw and B_uvw must meet the following requirement: (B_uvw-A_uvw) > (0,0,0). However, (B_uvw-A_uvw)=" << B_uvw-A_uvw << std::endl;
+
 
         PointerVector<NodeType> points;
         double delta_x = std::abs(B_xyz.X()-A_xyz.X())/OrderU;
@@ -241,10 +245,10 @@ namespace Kratos
         SizeType NumKnotSpansU, SizeType NumKnotSpansV, SizeType NumKnotSpansW )
     {
         KRATOS_ERROR_IF( B_xyz.X() <= A_xyz.X() || B_xyz.Y() <= A_xyz.Y() || B_xyz.Z() <= A_xyz.Z() ) << "NurbsGeometryModeler: "
-            << "The two Points A and B must meet the following requirement: (B-A) > (0,0,0). However, (B-A)=" << B_xyz-A_xyz << std::endl;
+            << "The two Points A_xyz and B_xyz must meet the following requirement: (B_xyz-A_xyz) > (0,0,0). However, (B_xyz-A_xyz)=" << B_xyz-A_xyz << std::endl;
 
         KRATOS_ERROR_IF( B_uvw.X() <= A_uvw.X() || B_uvw.Y() <= A_uvw.Y() || B_uvw.Z() <= A_uvw.Z() ) << "NurbsGeometryModeler: "
-            << "The two Points A and B must meet the following requirement: (B-A) > (0,0,0). However, (B-A)=" << B_uvw-A_uvw << std::endl;
+            << "The two Points A_uvw and B_uvw must meet the following requirement: (B_uvw-A_uvw) > (0,0,0). However, (B_uvw-A_uvw)=" << B_uvw-A_uvw << std::endl;
 
         // Set up control points.
         // Note: The CP's are uniformly subdivided with increasing p.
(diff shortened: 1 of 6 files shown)


applications/IgaApplication/custom_modelers/nurbs_geometry_modeler.cpp
```diff
@@ -22,22 +22,39 @@ namespace Kratos
     ///@{
 
     void NurbsGeometryModeler::SetupGeometryModel(){
-        // Get bounding box
-        KRATOS_ERROR_IF_NOT(mParameters.Has("lower_point"))
-            << "NurbsGeometryModeler: Missing \"lower_point\" section" << std::endl;
+        // Get bounding box physical space.
+        KRATOS_ERROR_IF_NOT(mParameters.Has("lower_point_xyz"))
```
tteschemacher, line 26: I like XYZ. However, I think the convention is global coordinates.

applications/IgaApplication/custom_modelers/nurbs_geometry_modeler.cpp
```diff
@@ -219,60 +240,64 @@ namespace Kratos
...
 
-    void NurbsGeometryModeler::CreateAndAddRegularGrid3D( ModelPart& r_model_part, const Point& A, const Point& B, SizeType OrderU, SizeType OrderV, SizeType OrderW,
+    void NurbsGeometryModeler::CreateAndAddRegularGrid3D( ModelPart& r_model_part, const Point& A_xyz, const Point& B_xyz,
+        const Point& A_uvw, const Point& B_uvw, SizeType OrderU, SizeType OrderV, SizeType OrderW,
         SizeType NumKnotSpansU, SizeType NumKnotSpansV, SizeType NumKnotSpansW )
     {
-        KRATOS_ERROR_IF( B.X() <= A.X() || B.Y() <= A.Y() || B.Z() <= A.Z() ) << "NurbsGeometryModeler: "
-            << "The two Points A and B must meet the following requirement: (B-A) > (0,0,0). However, (B-A)=" << B-A << std::endl;
+        KRATOS_ERROR_IF( B_xyz.X() <= A_xyz.X() || B_xyz.Y() <= A_xyz.Y() || B_xyz.Z() <= A_xyz.Z() ) << "NurbsGeometryModeler: "
+            << "The two Points A_xyz and B_xyz must meet the following requirement: (B_xyz-A_xyz) > (0,0,0). However, (B_xyz-A_xyz)=" << B_xyz-A_xyz << std::endl;
```
tteschemacher, line 248: Just out of curiosity, why does it need to be like this. You could define things in an opposite way.

tteschemacher (dismissed):
Looks ok to me. I suppose you have your reasons for that change.

With this change you should consider to complete the function of the volume geometry which returns local coordinates in return of global coordinates.

tteschemacher:
Maybe you could leave a short comment here for your reason to introduce separate spaces that we'll have it documented here in the future.

applications/IgaApplication/tests/test_modelers.py
```diff
@@ -237,15 +243,93 @@ def test_nurbs_geometry_2d_modeler_consistency(self):
             param[1] -= 0.1
             param[2] = 0.0
 
+    def test_nurbs_geometry_2d_modeler__parameter_space(self):
```
rickyaristio: typo (__) ?

applications/IgaApplication/tests/test_modelers.py
```diff
@@ -382,17 +470,108 @@ def test_nurbs_geometry_3d_modeler_consistency(self):
...
+                    param[2] += 0.1
+                    self.assertVectorAlmostEqual(coord_01, coord_02)
+                param[1] += 0.1
+                param[2] = 0.0
             param[0] += 0.1
-            param[1] -= 0.1
-            param[2] = 0.0
+            param[1] = 0.0
+
+    def test_nurbs_geometry_3d_modeler__parameter_space(self):
```
rickyaristio: idem

rickyaristio (commented):
Everything also looks fine to me.

applications/IgaApplication/custom_modelers/nurbs_geometry_modeler.cpp
manuelmessmer: It is not possible to have lower_bound larger than upper_bound.

manuelmessmer:
> Maybe you could leave a short comment here for your reason to introduce separate spaces that we'll have it documented here in the future.

Well, if you have very large patch (especially in 3D) and you map to [0,0,0][1,1,1] the knot spans become really small. Meaning the integration domains are small. In fact, if I have many elements, my integration weights are all <1e-8. So numerically, it is just much better to have a larger parametric space and hence a larger integration domain. Also, this allows to have the physical space equal to the parametric space.

applications/IgaApplication/tests/test_modelers.py
manuelmessmer: oh yes.

applications/IgaApplication/custom_modelers/nurbs_geometry_modeler.cpp
manuelmessmer: You mean: `lower_point_global_coordinates` and `lower_bound_local_coordinates`? I would be prefer xyz and uvw..

applications/IgaApplication/custom_modelers/nurbs_geometry_modeler.cpp
manuelmessmer: Well it would work for the physical space, but it still is probably not what you want. It does not work for the parametric space.

manuelmessmer:
> Looks ok to me. I suppose you have your reasons for that change.
> 
> With this change you should consider to complete the function of the volume geometry which returns local coordinates in return of global coordinates.

So far, I always assume the volume to be a regular grid. Therefore, mapping is very easy. Just a linear mapping. No Newton Raphson required. I think we will never work on boundary-fitted volumes. Anyway, I will implement this linear mapping in `ProjectionPointGlobalToLocalSpace` in the nurbs_volume_geometry. I will open a new PR.

applications/IgaApplication/custom_modelers/nurbs_geometry_modeler.cpp
tteschemacher: I think so. But I also agree that in this case, that's shorter and nicer to read.

applications/IgaApplication/custom_modelers/nurbs_geometry_modeler.cpp
tteschemacher: I don't fully understand why it wouldn't work in the parameter space. However, that's rather a detail. As the error message is sufficiently descriptive it should be fine.

tteschemacher (approved):

tteschemacher (approved):
Thanks for keeping the code up to date.

rickyaristio (approved):

manuelmessmer:
Thanks guys.

PR #11451, opened 2023-08-02
[Core] Nurbs volume mapping

manuelmessmer:
As discussed in #11449, here is the implementation of the `ProjectionPointGlobalToLocalSpace` for the `Nurbs_Volume_Geometry`.

**🆕 Changelog**
- ProjectionPointGlobalToLocalSpace()

### DIFF
--- a/applications/IgaApplication/custom_processes/map_nurbs_volume_results_to_embedded_geometry_process.cpp
+++ b/applications/IgaApplication/custom_processes/map_nurbs_volume_results_to_embedded_geometry_process.cpp
@@ -77,28 +77,13 @@ namespace Kratos
         const SizeType number_nodes_embedded = embedded_model_part.NumberOfNodes();
         IntegrationPointsArrayType integration_points(number_nodes_embedded);
 
-        const CoordinatesArrayType lower_point = p_geometry->begin()->GetInitialPosition();
-        const CoordinatesArrayType upper_point = (p_geometry->end()-1)->GetInitialPosition();
-
-        const auto p_nurbs_volume_geo = dynamic_pointer_cast<NurbsVolumeGeometry<PointerVector<Node>>>(p_geometry);
-        const auto knots_u = p_nurbs_volume_geo->KnotsU();
-        const SizeType n_knots_u = p_nurbs_volume_geo->NumberOfKnotsU();
-        const auto knots_v = p_nurbs_volume_geo->KnotsV();
-        const SizeType n_knots_v = p_nurbs_volume_geo->NumberOfKnotsV();
-        const auto knots_w = p_nurbs_volume_geo->KnotsW();
-        const SizeType n_knots_w = p_nurbs_volume_geo->NumberOfKnotsW();
-
         const auto node_itr_begin = embedded_model_part.NodesBegin();
         IndexPartition<IndexType>(embedded_model_part.NumberOfNodes()).for_each([&](IndexType i) {
             auto node_itr = node_itr_begin + i;
             // Map point into parameter space
             CoordinatesArrayType local_point;
-            local_point[0] = ((node_itr->X() - lower_point[0]) / std::abs( lower_point[0] - upper_point[0])
-                            * std::abs(knots_u[n_knots_u-1] - knots_u[0])) + knots_u[0];
-            local_point[1] = ((node_itr->Y() - lower_point[1]) / std::abs( lower_point[1] - upper_point[1])
-                            * std::abs(knots_v[n_knots_v-1] - knots_v[0])) + knots_v[0];
-            local_point[2] = ((node_itr->Z() - lower_point[2]) / std::abs( lower_point[2] - upper_point[2])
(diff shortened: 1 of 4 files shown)


applications/IgaApplication/custom_processes/map_nurbs_volume_results_to_embedded_geometry_process.cpp
```diff
@@ -77,28 +77,13 @@ namespace Kratos
...
             auto node_itr = node_itr_begin + i;
             // Map point into parameter space
             CoordinatesArrayType local_point;
-            local_point[0] = ((node_itr->X() - lower_point[0]) / std::abs( lower_point[0] - upper_point[0])
-                            * std::abs(knots_u[n_knots_u-1] - knots_u[0])) + knots_u[0];
-            local_point[1] = ((node_itr->Y() - lower_point[1]) / std::abs( lower_point[1] - upper_point[1])
-                            * std::abs(knots_v[n_knots_v-1] - knots_v[0])) + knots_v[0];
-            local_point[2] = ((node_itr->Z() - lower_point[2]) / std::abs( lower_point[2] - upper_point[2])
-                            * std::abs(knots_w[n_knots_w-1] - knots_w[0])) + knots_w[0];
+            p_geometry->ProjectionPointGlobalToLocalSpace(node_itr->Coordinates(), local_point);
```
tteschemacher, line 85: It would be actually quite interesting to know how much this slowers the procedure. Like in many calls. I wonder if it makes a difference.

kratos/geometries/nurbs_volume_geometry.h
```diff
@@ -885,6 +885,42 @@ class NurbsVolumeGeometry : public Geometry<typename TContainerPointType::value_
...
+/**
+    * @brief Returns local coordinates in return for a point in physical coordinates.
+    *        This function assumes the initial geometry to be a regular grid. Only linear mapping is applied.
+    *
+    * @param rPointGlobalCoordinates the point to which the
+    *        projection has to be found.
+    * @param rProjectedPointLocalCoordinates the location of the
+    *        projection in local coordinates.
+    * @param Tolerance not used.
+    * @return 1
```
tteschemacher: Return 1 is used to express that the protection has worked. Maybe extend the comment here that this is easier to understand.

kratos/geometries/nurbs_volume_geometry.h
```diff
@@ -885,6 +885,42 @@ class NurbsVolumeGeometry : public Geometry<typename TContainerPointType::value_
...
+        const CoordinatesArrayType& rPointGlobalCoordinates,
+        CoordinatesArrayType& rProjectedPointLocalCoordinates,
+        const double Tolerance = std::numeric_limits<double>::epsilon()
+    ) const override
+    {
+        const CoordinatesArrayType& lower_point = this->begin()->GetInitialPosition();
+        const CoordinatesArrayType& upper_point = (this->end()-1)->GetInitialPosition();
+
+        const Vector& knots_u = this->KnotsU();
+        const SizeType n_knots_u = this->NumberOfKnotsU();
```
tteschemacher: I'm not a huge fan of n_ I think it can be easily confused with m_ . Either nb_ or number_ 

Anyways, that's a matter of gusto.

tteschemacher (dismissed):
I think there isn't much to say. 

Thank you very much for adding the function to the geometry.

applications/IgaApplication/custom_processes/map_nurbs_volume_results_to_embedded_geometry_process.cpp
manuelmessmer: I guess, the only overhead is some function calls.

kratos/geometries/nurbs_volume_geometry.h
```diff
@@ -885,6 +885,42 @@ class NurbsVolumeGeometry : public Geometry<typename TContainerPointType::value_
...
+        const Vector& knots_v = this->KnotsV();
+        const SizeType n_knots_v = this->NumberOfKnotsV();
+        const Vector& knots_w = this->KnotsW();
+        const SizeType n_knots_w = this->NumberOfKnotsW();
```
manuelmessmer: The overhead should be these function calls. Since now, we do them for each projection.

kratos/geometries/nurbs_volume_geometry.h
manuelmessmer: Done to please your gusto.

applications/IgaApplication/custom_processes/map_nurbs_volume_results_to_embedded_geometry_process.cpp
tteschemacher: Most of them should compile as inline. I hope the impact is rather minor then.

kratos/geometries/nurbs_volume_geometry.h
tteschemacher: 😍

tteschemacher (approved):
Thanks for considering my suggestions.

From my side it's good to go.

PR #11460, opened 2023-08-05
[Core] Uniform Wrappers for Historical/NonHistorical Nodes, Elements, and Conditions

matekelemen:
# Ad

Are YOU tired of writing the same code for `Node`s, `Element`s and `Condition`s? You don't care whether a `Variable` of a `Node` is stored in a historical or non-historical container? Rejoice, for `EntityProxy` and `ContainerProxy` are here to release you from your suffering!

`#include` `proxies.h` in your project today and say goodbye to code duplication!

# Description

Proxy types are introduced which provide uniform access at different levels to variables stored
- historically in `Node`
- non-historically in `Node`
- in `Element`
- in `Condition`

## `EntityProxy` in C++

This is the lowest level wrapper. It provides access to variables associated with an entity without additional runtime overhead. Consequently, the implementation is heavily templated so I recommend using them through the `MakeProxy` factory instead of constructing them directly.

Example:
```cpp
#include "utilities/proxies.h"

Node node;
auto proxy = MakeProxy<Globals::DataLocation::NodeHistorical>(node);
proxy.SetValue(PRESSURE, 0.0);
```

## `ContainerProxy` in C++

This is a view class that wraps a `ModelPart::NodesContainer`, `ModelPart::ElementsContainerType`, or `ModelPart::ConditionsContainerType`.  When looping through the view, `EntityProxy` instances are created on the fly, wrapping the underlying entity. The `MakeProxy` factory function supports constructing `ContainerProxy` instances as well.

Example:
```cpp
#include "utilities/proxies.h"

ModelPart model_part;
auto proxies = MakeProxy<Globals::DataLocation::NodeHistorical>(model_part);
for (auto proxy : proxies) {
    proxy.SetValue(PRESSURE, 0.0);
}
```

## ~~`EntityProxy` in python == `DynamicEntityProxy` in C++~~

*This feature was removed. Check the discussion for details.*

`DynamicEntityProxy` is a runtime version of `EntityProxy` that can also be used in python. Obviously, there's a runtime overhead for this flexibility, but that's exactly why these are separate classes.

Examples:
```cpp
#include "utilities/dynamic_proxies.h"

Node node;
DynamicEntityProxy proxy(Globals::DataLocation::NodeHistorical, node);
node.SetValue(PRESSURE, 0.0);
```
```python
import KratosMultiphysics

node: KratosMultiphysics.Node
proxy = KratosMultiphysics.EntityProxy(KratosMultiphysics.Globals.DataLocation.NodeHistorical, node)
proxy.SetValue(KratosMultiphysics.PRESSURE, 0.0)
```

## ~~`ContainerProxy` in python~~

*This feature was removed. Check the discussion for details.*

This is a runtime equivalent of `ContainerProxy` in C++, but it's only defined in python (it wouldn't make much sense in C++). It does exactly the same thing as its C++ counterpart, except dynamically and in python. There isn't much else to be said.

Example:
```py
import KratosMultiphysics
from KratosMultiphysics.container_proxy import ContainerProxy

model_part: KratosMultiphysics.ModelPart
for proxy in ContainerProxy(KratosMultiphysics.Globals.DataLocation.NodeHistorical, model_part):
    proxy.SetValue(KratosMultiphysics.PRESSURE, 0.0)
```


loumalouomega:
FYI @KratosMultiphysics/technical-committee

loumalouomega:
I like the idea, there is any important overhead?

matekelemen:
> there is any important overhead?

Depends on which proxy you want to use.
- In C++, `EntityProxy` has no overhead in release mode because the compiler has access to everything at compile time and can fully optimize the wrapper away. [Here](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGISRqkrgAyeAyYAHI%2BAEaYxCAArGakAA6oCoRODB7evv6BaRmOAqHhUSyx8Um2mPbFDEIETMQEOT5%2BATV1WY3NBKWRMXGJyQpNLW15nWN9A%2BWVIwCUtqhexMjsHOYAzGHI3lgA1CbbbgQAnimYAPoExEyECifYhwD0r4dj6CAgeArXClYNwAbiYNABBHZ7A6YY6nJxjYiYVjPMGQiH7JgKBSHCIYWEmADsVghhzJhzCBEOLAAIkwmicSeCUl5orQ8MgQGjyRTBOYAGyHADimGEBEwKTpTQgi2OxKRBDWDGpUqYjKJNO55OBqDw6EOQlFjQlqoglMOxAQCllRIstPpTDhNItVvVhM1EI16s9GIM2MO2HobEEcqZPPN9oZ2yZLLZHK5pPJlIFwtFppt8tFSpVDrdHvBPJ1eoNaYdZpDlutocjjpOzsrebRXujTYh4pYKQM4rhbkx/oAKthBIRzqjfVicUPHBdlMRUKpzqGtWTB8OLgAqakpKcj73M1nsznLgNr86z%2BfnCCr6ejsyC4g7i6ykBbx%2BXgUP08Zqzu4/Ju%2BpgQ6ZLomPJkngVCHGgDBjJgqgpMQhwQF8Px/ACQLXKCpwoSAWCiOctwnG4167ts2CkHiWDPN%2Bx5geSCrZh2b4ALTPCKYommWix7nRZIaoctRKBSkHQbB8GIchBDfL8/yAmwmFETheFMARBBESRFzPKQgaYMGalkTRoG8WSDHEMqTGnqxZHsemPG8fxgkEsStG8TMjjINcE5xAQ1BiEo3EtkZPJekZIUFtqur6oaQFlualaGeFdEQVBAhiQhSE4WhckgopUk/MpqnqW%2BWmUZg1EgYlxkWlmZmvpZzzRcakplvFdl0Q5tBCcloniuJGV5TJ6HyVhbhKZg%2BGEacGmjmR2lBow%2BnYAlVX0TV5nbvVZHRaarWBZVwXugJnVOWGK2fE07medi3m%2BcdAWne1v6hU9kK/ntaLmgASpgVBxIwGwQKVhwMPi3Ges5Rkg1gAB0jXis10pQ5g0PsU1wGbmY93HqZypIyjRrw7Z70vR9Ib9pgYyA/iwOg02EOVW%2B54LkRpXPIcCEXhASNY0ZHMLrDpbSnz5z4zF0qyhjPOVTj7NzvzNlcXuYUcMstCcAkvB%2BBwWikKgnBuNY1ifKs6wEmY2w8KQBCaCrywANYSBogRqxwki8CwiSBFrOt6xwvAKCAgTW9rKukHAsBIGgHZ0HE5CUFHKQx/EwBcNsyRYMCHKYAAangmAAO4APJXFrls0LQ4rEAHEDRDbpDRGEzTnJwlsN8wxDnIX0TaONweW1HemFwwtDNyHpBYNEXjAG4YidS3vBYCwhjAOIY/4EiDh4MCFN13B41eOK8/kIItR1%2By0T3B3HhYHXdx4B73Ch1QBjAAoucF8XjBH/wggiGI7BSBkIIRQKh1Bj10FwfQy8UCG0sPoPA0QA6QGWKgFI9QA4cGYoXMwvBUDb2IMQPUO9kFdF7lkFwDB3CeHaHoEIYRBgVGGJAwomQBCTD8Mw9IrCGBzCGPESBdgyECF6BMaheQBG1CEQ0cY/R6HzCYbYGR7C9AzBaLwxh/DlgKBNhsCQqt1aazrr7Q4qgAAc/JmL8kkIcYAyBkCHFTtDMwSFcCEBIMcc2XBFi8GDloRYywEDIiwPEGU%2BhOBu1IB7BIXsjGcH9oHK2Nt/FhI4NsQxY9fY%2BKScsfBGRnCSCAA)'s a barebones example that shows that the generated assembly is the same as if `EntityProxy` wasn't used (the full implementation is more complex of course, but the concept is identical).
- `ContainerProxy` in C++ should also be zero runtime overhead, but I didn't dig into it.
- `DynamicEntityProxy` and the python objects do have runtime overhead, and it has 2 reasons:
  1) we need **some** sort of runtime polymorphism, and that'll always have some overhead. In this case, I used `std::variant`.
  2) Since the behaviour is decided at runtime, the compiler no longer has access to the entire call hierarchy and cannot optimize the entire wrapping away.

At the end of the day, you can choose between performance (and deal with templates) or convenience (pay the price at runtime).

matekelemen:
> there is any important overhead?

P.S.: `EntityProxy` stores an `optional` of an entity pointer to make the default constructor safe (i.e. not segfault if used in its default constructed state). This optional adds a minuscule runtime overhead, so my earlier comment is not *entirely* true.

https://github.com/KratosMultiphysics/Kratos/blob/a86652a96b0a210c248ca12b2c9a1586a31d7e32/kratos/utilities/proxies.h#L148-L150

https://github.com/KratosMultiphysics/Kratos/blob/a86652a96b0a210c248ca12b2c9a1586a31d7e32/kratos/utilities/proxies.h#L81-L84

If you think this is unnecessary, I'm open to changes here. These are our options and what I think of them:
1) *store `mpEntity` as a naked raw pointer and delete the default constructor*: this is a reasonable alternative. We gain a tiny bit of performance at the expense of some flexibility (cannot construct a null proxy to reassign later).
2) *store `mpEntity` as a naked raw pointer and define the default constructor*: I'm vetoing this one because it can lead to a segfault.
3) *store `mpEntity` as an optional raw pointer and define the default constructor*: this is the current implementation. We retain a bit more flexibility thanks to the default constructor, at the expense of a tiny runtime overhead incurred by `std::optional`.

matekelemen:
@sunethwarna I added some free-floating functions fill the same role as `EntityProxy` and `ContainerProxy` but in procedural programming style:
- `VariableUtils::HasValue`
- `VariableUtils::GetValue`
- `VariableUtils::SetValue`
- `ModelPartUtils::GetContainer`

Furthermore, I extended `EntityProxy` for the remaining major classes inheriting from `DataValueContainer` (namely `ProcessInfo` and `ModelPart`).

matekelemen:
This is ready for review.

kratos/utilities/variable_utils.h
```diff
@@ -1531,6 +1534,120 @@ class KRATOS_API(KRATOS_CORE) VariableUtils
...
+                          || std::is_same_v<TEntity,ModelPart>);
+            return rEntity.Has(rVariable);
+        }
+    }
+
+    /// @brief Fetch the value of a variable stored in an entity.
+    /// @param rEntity @ref Node, @ref Element, @ref Condition, @ref ProcessInfo, or @ref ModelPart to fetch data from.
+    /// @param rVariable @ref Variable to fetch the value of.
+    template <Globals::DataLocation TLocation, class TEntity, class TValue>
+    static std::conditional_t<std::is_integral_v<TValue> || std::is_floating_point_v<TValue>,
```
sunethwarna:
```suggestion
    static std::conditional_t<std::is_arithmetic_v<TValue>,
```
https://en.cppreference.com/w/cpp/types/is_arithmetic

kratos/utilities/variable_utils.h
```diff
@@ -1531,6 +1534,120 @@ class KRATOS_API(KRATOS_CORE) VariableUtils
...
+    }
+
+    /// @brief Overwrite the value of a variable stored in an entity.
+    /// @param rEntity @ref Node, @ref Element, @ref Condition, @ref ProcessInfo, or @ref ModelPart to set the value of.
+    /// @param rVariable @ref Variable to overwrite the value of.
+    /// @param Value new value of @a rVariable to set in @a rEntity.
+    template <Globals::DataLocation TLocation, class TEntity, class TValue>
+    static void SetValue(TEntity& rEntity,
+                         const Variable<TValue>& rVariable,
+                         std::conditional_t<std::is_integral_v<TValue> || std::is_floating_point_v<TValue>,
```
sunethwarna:
```suggestion
                         std::conditional_t<std::is_arithmetic_v<TValue>,
```

kratos/utilities/variable_utils.h
```diff
@@ -1531,6 +1534,120 @@ class KRATOS_API(KRATOS_CORE) VariableUtils
...
+                                            TValue,         /*pass scalar types by value*/
+                                            const TValue&>  /*pass non-scalar types by reference*/ Value)
+    {
+        if constexpr (TLocation == Globals::DataLocation::NodeHistorical) {
+            static_assert(std::is_same_v<TEntity,Node>);
+            rEntity.GetSolutionStepValue(rVariable) = Value;
```
sunethwarna:
```suggestion
            rEntity.FastGetSolutionStepValue(rVariable) = Value;
```

kratos/utilities/model_part_utils.h
```diff
@@ -0,0 +1,61 @@
...
+// Project includes
+#include "includes/kratos_export_api.h"
+#include "includes/model_part.h"
+#include "includes/global_variables.h"
+
+namespace Kratos {
+
+
+/// @brief Class collecting a set of free-floating utility functions for querying and mutating @ref ModelPart s.
+class KRATOS_API(KRATOS_CORE) ModelPartUtils
```
sunethwarna:
```suggestion
class ModelPartUtils
```

I don't think `KRATOS_API` is required for headers only classes.

kratos/utilities/proxies.h
```diff
@@ -0,0 +1,380 @@
...
+ *         - an @ref Element (@ref Globals::DataLocation::Element)
+ *         - a @ref Condition (@ref Globals::DataLocation::Condition)
+ *         The exposed common functionalities include checking, reading and overwriting the values
+ *         related to the provided variables associated with the entity.
+ *  @warning Default constructed @ref EntityProxy instances are in an invalid state,
+ *           and their member functions must not be called.
+ *  @throws if member functions of a default constructed instance are called.
+ */
+template <Globals::DataLocation TLocation, bool TMutable>
+class KRATOS_API(KRATOS_CORE) EntityProxy
```
sunethwarna:
```suggestion
class EntityProxy
```

kratos/utilities/proxies.h
```diff
@@ -0,0 +1,380 @@
...
+/** @brief A view with a uniform interface for @ref ModelPart::NodesContainerType, @ref ModelPart::ElementsContainerType, or @ref ModelPart::ConditionsContainerType.
+ *  @details @ref ContainerProxy provides uniform access to the @ref Variable s stored in the entities of a container.
+ *           Entities in the container are wrapped by @ref EntityProxy. In this context, an entity can refer to:
+ *         - a @ref Node with historical variables (@ref Globals::DataLocation::NodeHistorical)
+ *         - a @ref Node with non-historical variables (@ref Globals::DataLocation::NodeNonHistorical)
+ *         - an @ref Element (@ref Globals::DataLocation::Element)
+ *         - a @ref Condition (@ref Globals::DataLocation::Condition)
+ */
+template <class TEntityProxy>
+class KRATOS_API(KRATOS_CORE) ContainerProxy
```
sunethwarna:
```suggestion
class ContainerProxy
```

kratos/utilities/dynamic_proxies.h
```diff
@@ -0,0 +1,110 @@
...
+#include "includes/kratos_export_api.h" // KRATOS_API
+
+// System includes
+#include <variant> // variant
+
+
+namespace Kratos {
+
+
+class KRATOS_API(KRATOS_CORE) DynamicEntityProxy
```
sunethwarna: I would suggest to move this entire thing to pybind or delete it fully. Because, using this in cpps will have a big perfromance hit.

kratos/python_scripts/container_proxy.py
```diff
@@ -0,0 +1,65 @@
...
+
+# --- STD Imports ---
+import typing
+
+
+##! @addtogroup KratosCore
+##! @{
+
+
+class ContainerProxy:
```
sunethwarna: I would get rid of this, since this does not make much sense in python. Users should directly call the respective functions rather than relying on this. And no computationally heavy operations should be done in python.

sunethwarna (commented):
Minor comments. Basically, I am ok with this, but as you suggested, I would remove the default ctor of the proxies and get rid of the `std::optional`.

kratos/python_scripts/container_proxy.py
matekelemen: done

kratos/utilities/dynamic_proxies.h
matekelemen: deleted it

kratos/utilities/variable_utils.h
matekelemen: I hate that there are 2 versions of this

matekelemen:
- removed `DynamicEntityProxy`
- removed python proxies
- deleted the default constructor of `EntityProxy` so it now stores a naked raw pointer (instead of an optional) => now truly has zero runtime overhead in release mode.

kratos/utilities/variable_utils.h
```diff
@@ -1531,6 +1534,120 @@ class KRATOS_API(KRATOS_CORE) VariableUtils
...
+    /// @param rVariable @ref Variable to fetch the value of.
+    template <Globals::DataLocation TLocation, class TEntity, class TValue>
+    static std::conditional_t<std::is_arithmetic_v<TValue>,
+                              TValue,             // <== return by value if scalar type
+                              const TValue&>      // <== return by reference if non-scalar type
+    GetValue(const TEntity& rEntity, const Variable<TValue>& rVariable)
+    {
+        if constexpr (TLocation == Globals::DataLocation::NodeHistorical) {
+            static_assert(std::is_same_v<TEntity,Node>);
+            return rEntity.GetSolutionStepValue(rVariable);
```
sunethwarna:
```suggestion
            return rEntity.FastGetSolutionStepValue(rVariable);
```

kratos/utilities/variable_utils.h
```diff
@@ -1531,6 +1534,120 @@ class KRATOS_API(KRATOS_CORE) VariableUtils
...
+
+    /// @brief Fetch the value of a variable stored in an entity.
+    /// @param rEntity @ref Node, @ref Element, @ref Condition, @ref ProcessInfo, or @ref ModelPart to fetch data from.
+    /// @param rVariable @ref Variable to fetch the value of.
+    template <Globals::DataLocation TLocation, class TEntity, class TValue>
+    static TValue& GetValue(TEntity& rEntity, const Variable<TValue>& rVariable)
+    {
+        if constexpr (TLocation == Globals::DataLocation::NodeHistorical) {
+            static_assert(std::is_same_v<TEntity,Node>);
+            return rEntity.GetSolutionStepValue(rVariable);
```
sunethwarna:
```suggestion
            return rEntity.FastGetSolutionStepValue(rVariable);
```

sunethwarna (commented):
I think you missed two. Otherwise, I am totally fine with this :)

sunethwarna (approved):
Thanks @matekelemen :)

sunethwarna (changes requested):
On the other hand, we need to discuss somethings before merging this.

sunethwarna (approved):

PR #11461, opened 2023-08-07
[StructuralMechanicsApplication] added rot DOFS to 3D base load condition

aronnoordam:
rotation DOFS are now added to the base_load_condition in 3D


philbucher:
can you please explain why this change is needed and what it is used for?

aronnoordam:
It is needed for an upcoming update in the moving load, where I want the displacement and rotation at the location of the moving load. Where the calculation of the rotation and displacement is different in case rotation DOFS are active

philbucher:
Ok for me

@loumalouomega please approve if you agree

loumalouomega (approved):

loumalouomega:
I agree

aronnoordam:
thanks

PR #11462, opened 2023-08-08
[GeoMechanicsApplication] Reduced code smells in custom processes

avdg81:
**📝 Description**
Addressed several code smells reported by SonarCloud. Also fixed some problems reported by clang-tidy.

**🆕 Changelog**
- Replaced `typedef`s by `using` aliases.
- Made some `protected` data members `private`.
- Removed some commented out code.
- Prefer using `!empty()` over `size() > 0`.
- Use `auto` instead of stating the full iterator type.
- Made some single-argument constructors `explicit`.

Other enhancements include:
- Restored setting pressure only below phreatic for `mIsSeepage`.
- Removed class `ApplyTimeDependentInterpolateLinePressureProcess` as it had no implementation for time dependency.
- Removed `PrintInfo` overrides that were identical to the base class implementation.
- Replaced some old-style inclusion guards by `#pragma once`.
- Removed some redundant comments.
- Fixed some spelling mistakes.
- Improved the alignment of some code blocks.
- Applied the rule of zero when appropriate; otherwise applied the rule of five.


WPK4FEM (approved):
Large, but consistent increment.

Issue #11448, opened 2023-08-01
How tot Output strain to vtk


hrntsm:
**Description**

Related #11445, I created a polyline material model using the method you showed me. I tried to output the strain in vtk to confirm that the result is correct, but I could not do it. How can I do this?

I have tried the following

```JSON
  "output_processes": {
    "vtk_output": [
      {
        "Parameters": {
          "gauss_point_variables_in_elements": [
            "TOTAL_STRAIN_TENSOR",
            "GREEN_LAGRANGE_STRAIN_TENSOR",
            "CAUCHY_STRESS_TENSOR"
          ],
        },
     },
  }
```

**Scope**

- KratosCore

AlejandroCornejo:
As far as i remember, the vtk does not allow to visualize data in IP. However you can see an extrapolation to the nodes as:

~~~json
        "vtk_output" : [{
            "python_module" : "vtk_output_process",
            "kratos_module" : "KratosMultiphysics",
            "process_name"  : "VtkOutputProcess",
            "Parameters"    : {
                "model_part_name"                             : "Structure",
                "output_control_type"                         : "step",
                "output_interval"                             : 1,
                "file_format"                                 : "ascii",
                "output_precision"                            : 7,
                "output_sub_model_parts"                      : false,
                "output_path"                                 : "vtk_output",
                "save_output_files_in_folder"                 : true,
                "nodal_solution_step_data_variables"          : ["DISPLACEMENT","REACTION"],
                "nodal_data_value_variables"                  : [],
                "element_data_value_variables"                : [],
                "condition_data_value_variables"              : [],
                "gauss_point_variables_extrapolated_to_nodes" : ["GREEN_LAGRANGE_STRAIN_TENSOR"]
            }
        }]
~~~

AlejandroCornejo:
However, if what you want is to track the values of one IP of one element, I recommend to use the 

~~~json
{
            "python_module" : "print_info_in_file_process",
            "kratos_module" : "KratosMultiphysics",
            "process_name"  : "PrintInfoInFileProcess",
            "Parameters"    : {
                "model_part_name"      : "Structure.plot_element",
                "variable_name"        : "CAUCHY_STRESS_VECTOR",
                "variable_type"        : "elemental",
                "file_name"            : "CAUCHY_STRESS_VECTOR.txt",
                "output_control_type"  : "step",
                "output_interval"      : 1,
                "integration_point_number" : 0,
[...]

PR #11463, opened 2023-08-08
[HDF5] Controller and XDMF Bugfixes

matekelemen:
This PR adds some temporary fixes to controllers and XDMF generation. The entire HDF5Application will eventually need an overhaul to fix persistent problems and reliability issues, but hopefully this will be enough until that happens.

The bugs this PR fixes were detected by attempting to apply `SingleMeshTemporalOutputProcess` to the model provided in  #11448.
- XDMF generation didn't support quadratic elements. I added some XDMF topologies that we have already implemented in Kratos Core. They're untested though because one would have to manually check whether the node orderings match with what XDMF/paraview expect.
- All HDF5 output processes write the mesh in `ExecuteBeforeSolutionLoop`, which isn't great. However, they also wrote results in there which is flat out wrong. The model parts aren't necessarily fully initialized yet by that point, so non-historical values may not be present in entities yet. This results in null values for statically sized variables, but empty values for dynamic ones. => the result was empty data fields in the first output, which broke paraview.
- `TemporalController` changed its state every time `IsExecuteStep` was called because it expected that function to be called exactly once per solution step. However, this wasn't the case so output scheduling was completely broken. Furthermore, the tests relied on implementation details so they had to be heavily modified as well.

### DIFF
--- a/applications/HDF5Application/python_scripts/core/controllers.py
+++ b/applications/HDF5Application/python_scripts/core/controllers.py
@@ -26,8 +26,8 @@ class Controller(abc.ABC):
     def __init__(self,
                  model_part: KratosMultiphysics.ModelPart,
                  operation: operations.AggregateOperation):
-        self._model_part = model_part
-        self.__operation = operation
+        self._model_part: KratosMultiphysics.ModelPart = model_part
+        self.__operation: operations.AggregateOperation = operation
 
     @abc.abstractmethod
     def IsExecuteStep(self) -> bool:
--- a/applications/HDF5Application/python_scripts/core/controllers.py
+++ b/applications/HDF5Application/python_scripts/core/controllers.py
@@ -65,8 +65,8 @@ class TemporalController(Controller):
                  operation: operations.AggregateOperation,
                  settings: KratosMultiphysics.Parameters):
         super().__init__(model_part, operation)
-        time_frequency: typing.Optional[float] = None
-        step_frequency: typing.Optional[int] = None
+        time_frequency: Optional[float] = None
+        step_frequency: Optional[int] = None
         if settings.Has("time_frequency"):
             time_frequency = settings["time_frequency"].GetDouble()
         if settings.Has("step_frequency"):
@@ -82,12 +82,8 @@ class TemporalController(Controller):
         if time_frequency is None:
             time_frequency = float("inf")
 
-        # Step frequency was not defined => the output won't be triggered by changes in STEP
-        if step_frequency is None:
-            step_frequency = -1
-
-        self.__time_frequency = time_frequency
-        self.__step_frequency = step_frequency
+        self.__time_frequency: float = time_frequency
+        self.__step_frequency: Optional[int] = step_frequency
         self.__last_output_time = self._model_part.ProcessInfo[KratosMultiphysics.TIME]
         self.__last_output_step = self._model_part.ProcessInfo[KratosMultiphysics.STEP]
 
@@ -105,9 +101,10 @@ class TemporalController(Controller):
         # TODO: separately keeping track of steps and time internally is not a good
         # idea. What happens if the solution process involves jumping back and forth
         # in time (restarts, checkpointing)? @matekelemen
-        step_difference = self._model_part.ProcessInfo[KratosMultiphysics.STEP] - self.__last_output_step
-        if self.__step_frequency <= step_difference:
-            return True
+        if self.__step_frequency is not None:
+            step_difference = self._model_part.ProcessInfo[KratosMultiphysics.STEP] - self.__last_output_step
+            if self.__step_frequency <= step_difference:
+                return True
 
         time_difference = self._model_part.ProcessInfo[KratosMultiphysics.TIME] - self.__last_output_time
         if self.__time_frequency <= time_difference:
--- a/applications/HDF5Application/python_scripts/core/controllers.py
+++ b/applications/HDF5Application/python_scripts/core/controllers.py
@@ -15,7 +15,7 @@ from . import operations
 
 # STL imports
 import abc
-import typing
+from typing import Optional
 
 
 ##!@addtogroup HDF5Application
--- a/applications/HDF5Application/python_scripts/core/controllers.py
+++ b/applications/HDF5Application/python_scripts/core/controllers.py
@@ -15,17 +15,18 @@ from . import operations
 
 # STL imports
 import abc
+import typing
 
 
 ##!@addtogroup HDF5Application
 ##!@{
 ##!@name Kratos classes
 ##!@{
-class Controller(metaclass=abc.ABCMeta):
+class Controller(abc.ABC):
     def __init__(self,
                  model_part: KratosMultiphysics.ModelPart,
                  operation: operations.AggregateOperation):
-        self.model_part = model_part
+        self._model_part = model_part
         self.__operation = operation
 
     @abc.abstractmethod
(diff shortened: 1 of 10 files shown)


applications/HDF5Application/python_scripts/core/controllers.py
```diff
@@ -64,42 +65,63 @@ def __init__(self,
...
-        self.time_frequency = settings['time_frequency'].GetDouble()
-        self.step_frequency = settings['step_frequency'].GetInt()
-        self.current_time = 0.0
-        self.current_step = 0
+        time_frequency: typing.Optional[float] = None
+        step_frequency: typing.Optional[int] = None
+        if settings.Has("time_frequency"):
+            time_frequency = settings["time_frequency"].GetDouble()
+        if settings.Has("step_frequency"):
+            step_frequency = settings["step_frequency"].GetInt()
```
sunethwarna: There were long debates about the existing implementation mechanism of `time_frequency` and `step_frequency` in `HDF5`, and no one changed it because it will change the behaviour of the output. I think, it is good time we fix it. We can properly fix it in the upcoming overhauling of the `HDFApp`.

applications/HDF5Application/python_scripts/core/controllers.py
```diff
@@ -15,17 +15,18 @@
 
 # STL imports
 import abc
+import typing
```
philbucher:
```suggestion
from typing import Optional
```

I would only import what you need

applications/HDF5Application/python_scripts/core/controllers.py
```diff
@@ -15,17 +15,18 @@
...
 ##!@name Kratos classes
 ##!@{
-class Controller(metaclass=abc.ABCMeta):
+class Controller(abc.ABC):
     def __init__(self,
                  model_part: KratosMultiphysics.ModelPart,
                  operation: operations.AggregateOperation):
-        self.model_part = model_part
+        self._model_part = model_part
         self.__operation = operation
```
philbucher:
```suggestion
        self._model_part: KratosMultiphysics.ModelPart = model_part
        self.__operation: operations.AggregateOperation = operation
```

applications/HDF5Application/python_scripts/core/controllers.py
```diff
@@ -64,42 +65,63 @@ def __init__(self,
...
+            settings.AddInt("step_frequency", 1)
+            step_frequency = 1
+
+        # Time frequency was not defined => the output won't be triggered by changes in TIME
+        if time_frequency is None:
+            time_frequency = float("inf")
+
+        # Step frequency was not defined => the output won't be triggered by changes in STEP
+        if step_frequency is None:
+            step_frequency = -1
```
philbucher: that seems :fish:y to me, I would leave it as `None`, this also makes the checking more explicit

applications/HDF5Application/python_scripts/xdmf_utils.py
```diff
@@ -595,18 +595,31 @@ def CreateXdmfTemporalGridFromSinglefile(h5_file_name, h5path_pattern_to_mesh, h
         if len(output_results_dict.keys()) == 0:
             raise RuntimeError("No results data is found in the given hdf5 file matching the given pattern [ file_name = {:s}, pattern = {:s} ].".format(h5_file_name, h5path_pattern_to_results))
 
-        for k, v in output_results_dict.items():
-            if k in output_meshes_dict:
-                sgrid = CreateXdmfSpatialGrid(file_[output_meshes_dict[k]])
+        compound_dict = {}
+        for key in output_meshes_dict.keys():
```
philbucher:
```suggestion
        for key in output_meshes_dict:
```

applications/HDF5Application/python_scripts/xdmf_utils.py
```diff
@@ -595,18 +595,31 @@ def CreateXdmfTemporalGridFromSinglefile(h5_file_name, h5path_pattern_to_mesh, h
...
+            compound_dict[key] = (True, False)
+        for key in output_results_dict.keys():
```
philbucher:
```suggestion
        for key in output_results_dict:
```

applications/HDF5Application/python_scripts/xdmf_utils.py
```diff
@@ -595,18 +595,31 @@ def CreateXdmfTemporalGridFromSinglefile(h5_file_name, h5path_pattern_to_mesh, h
         if len(output_results_dict.keys()) == 0:
```
philbucher:
```suggestion
        if not output_results_dict:
```

applications/HDF5Application/python_scripts/core/controllers.py
matekelemen: `model_part` and `operation` are function arguments that already have type hints, so the member variables should already have the correct type info. Is it not working in your IDE?

applications/HDF5Application/python_scripts/core/controllers.py
matekelemen: `step_frequency` is later compared with an integer, which is why I set it to `-1`.
https://github.com/KratosMultiphysics/Kratos/blob/a994bb1c06a7e06019d588b938b21696f3c23eb4/applications/HDF5Application/python_scripts/core/controllers.py#L109
I can keep it as `None`, but then the comparison will be a bit less readable.

applications/HDF5Application/python_scripts/core/controllers.py
matekelemen: I usually `import` the whole module because you can use the module name as a "namespace" and avoid ambiguity, but I don't have a strong opinion here.

applications/HDF5Application/python_scripts/core/controllers.py
philbucher: it might be detected by the IDE, but I am pretty sure that type-checkers like `mypy` would complain

applications/HDF5Application/python_scripts/core/controllers.py
philbucher: IMO it becomes more readable with the `None` check

applications/HDF5Application/python_scripts/core/controllers.py
philbucher: are you using e.g. mypy locally? I would suggest to do that (until it is integrated in the CI).

matekelemen: > are you using e.g. mypy locally? I would suggest to do that (until it is integrated in the CI).

Yup I have `mypy` locally, so stubs are automatically generated every time I compile, and they work just fine. My IDE relies on the generated stubs, and it does pick up the type of `self._model_part` even without the explicit hint.

applications/HDF5Application/python_scripts/core/controllers.py
matekelemen: alright; changed it

applications/HDF5Application/python_scripts/core/controllers.py
philbucher: I mean do you do `mypy applications/HDF5Application/python_scripts/core/controllers.py`?

matekelemen: Nope, I imagine every dependency (including Kratos) would have to be correctly hinted for that to work. Am I wrong about that?

philbucher: yes :)
although you can selectively disable some checks, so that it can be added later also in established frameworks. 

My point is, I think randomly adding the type annotations without enforcing them will lead to a tough time when doing them properly. Aside from them being potentially wrong...

I recommend to give mypy a try, to see what errors can pop up etc. My writing of python code has changed quite a bit since I started to use it

matekelemen: Okay I'll give it a spin; thanks for the hint!

philbucher: Pun intended? 😛

philbucher (dismissed):
apologies, I forgot to approve :sweat:

matekelemen:
No worries, thanks for the review!

sunethwarna (approved):
Thanks @matekelemen

applications/HDF5Application/python_scripts/core/controllers.py
philbucher: Does the overhaul also include the split into baseApp and MPI extension? 👀

PR #11469, opened 2023-08-10
[Core] Adding DataTypeTraits

sunethwarna:
**📝 Description**
Before re-implementing the same `DataTypTraits` for the 5th time, I thought of creating generic `DataTypeTraits` covering all the possible data types in Kratos as in this PR. This provides generic interface for all the data types (even the nested data types). Following data type methods are introduced.

- `Size` -> Gives the flattened size of the value (recursive, compile-time deduced).
- `Shape` -> Gives the shape of the value (recursive, compile-time deduced).
- `Reshape` -> Reshapes given value to the given shape. This is done recursively at compile time deduced, and if actually a reshape happens in any level, this returns true. Otherwise, this returns false.
- `GetContiguousData` -> Returns contiguous data pointer (only possible with contiguous arrays)
- `CopyToContiguousData` -> Fills the given input vector with all data in the given value (recursive filling). The passed pointer must have enough allocated memory (recursive, compile-time deduced).
- `CopyFromContiguousData` -> Fills the value with a contiguous data array. The sizes must match (recursive, compile-time deduced).

**🆕 Changelog**
- Added `DataTypeTratis`
- Updated `MPIMessage`


kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+#include <type_traits>
+#include <vector>
+
+// Project includes
+#include "containers/array_1d.h"
+#include "includes/ublas_interface.h"
+
+namespace Kratos {
+
+template<class TDataType> class DataTypeTraits
```
matekelemen, line 32: missing docstring

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+{
+public:
+    ///@name Type definitions
+    ///@{
+
+    using ContainerType = TDataType;
```
matekelemen, line 38: I'm guessing this specialization is for scalars, right? Then `ContainerType` doesn't really make sense.

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+
+    static constexpr bool HasDynamicMemoryAllocation = DataTypeTraits<ValueType>::HasDynamicMemoryAllocation;
+
+    ///@}
+    ///@name Public static operations
+    ///@{
+
+    static inline unsigned int Size(const ContainerType& rContainer)
+    {
+        return Dimension * DataTypeTraits<ValueType>::Size(rContainer[0]);
```
matekelemen:
```suggestion
        if constexpr (Dimension == 0) {
            return 0;
        } else {
            return Dimension * DataTypeTraits<ValueType>::Size(rContainer[0]);
        }
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+    }
+
+    static inline std::vector<unsigned int> Shape(const ContainerType& rContainer)
+    {
+        auto shape = DataTypeTraits<ValueType>::Shape(rContainer[0]);
+        shape.insert(shape.begin(), Dimension);
+        return shape;
```
matekelemen:
```suggestion
        std::vector<unsigned> shape {Dimension};
        if constexpr (0 < Dimension) {
            const auto value_shape = DataTypeTraits<ValueType>::Shape(rContainer[0]);
            std::copy(value_shape.begin(),
                      value_shape.end(),
                      std::back_inserter(shape));
        }
        return shape;
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+            std::for_each(rContainer.begin(), rContainer.end(), [&is_reshaped, &sub_data_type_shape](auto& rValue) { is_reshaped = DataTypeTraits<ValueType>::Reshape(rValue, sub_data_type_shape) || is_reshaped; });
+        }
+
+        return is_reshaped;
+    }
+
+    inline static PrimitiveType const * GetContiguousData(const ContainerType& rValue)
+    {
+        if constexpr(HasContiguousPrimitiveData) {
+            return rValue.data().begin();
```
matekelemen:
```suggestion
            return rValue.data().data();
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+        } else {
+            static_assert(!std::is_same_v<TDataType, TDataType>, "This should be only called if the rValue is contiguous only.");
```
matekelemen, line 442: have you tried compiling with a type that doesn't satisfy this requirement? I think it will fail to compile regardless of whether you call it anywhere. You'll either have to throw a runtime error or try to disable this function if the requirements aren't met.

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+        }
+    }
+
+    inline static PrimitiveType * GetContiguousData(ContainerType& rValue)
+    {
+        if constexpr(HasContiguousPrimitiveData) {
+            return rValue.data().begin();
```
matekelemen:
```suggestion
            return rValue.data().data();
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+        } else {
+            static_assert(!std::is_same_v<TDataType, TDataType>, "This should be only called if the rValue is contiguous only.");
```
matekelemen, line 474: same

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+        }
+    }
+
+    inline static void FillToContiguousData(
```
matekelemen: going by `std` terminology, "fill" applies when you're copying one value to each component in the destination while "copy" means that the items get copied componentwise. I think "copy" is the correct term here.

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+    ///@name Type definitions
+    ///@{
+
+    using ContainerType = TDataType;
+
+    using ValueType = TDataType;
+
+    using PrimitiveType = TDataType;
+
+    static constexpr bool HasContiguousPrimitiveData = true;
```
matekelemen:
```suggestion
    static constexpr bool IsContiguous = true;
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+
+    static constexpr bool HasDynamicMemoryAllocation = false;
```
matekelemen:
```suggestion
    static constexpr bool IsDynamic = false;
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+
+    static inline unsigned int Size(const ContainerType& rValue)
+    {
+        return (rValue.size() == 0 ? 0 : rValue.size() * DataTypeTraits<ValueType>::Size(rValue[0]));
+    }
+
+    static inline std::vector<unsigned int> Shape(const ContainerType& rValue)
+    {
+        std::vector<unsigned int> shape;
+        if (rValue.size() == 0) {
```
matekelemen:
```suggestion
        if (rValue.empty()) {
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+        bool is_reshaped = false;
+
+        if (rContainer.size() != rShape[0]) {
+            rContainer.resize(rShape[0], false);
+            is_reshaped = true;
+        }
+
+        if constexpr(DataTypeTraits<ValueType>::HasDynamicMemoryAllocation) {
+            std::vector<unsigned int> sub_data_type_shape(rShape.size() - 1);
+            std::copy(rShape.begin() + 1, rShape.end(), sub_data_type_shape.begin());
```
matekelemen: this is why I suggest implementing stuff in terms of iterators. You'd save loads of unnecessary allocations and copies here.

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+            std::for_each(rContainer.begin(), rContainer.end(), [&is_reshaped, &sub_data_type_shape](auto& rValue) { is_reshaped = DataTypeTraits<ValueType>::Reshape(rValue, sub_data_type_shape) || is_reshaped; });
```
matekelemen:
```suggestion
            std::for_each(rContainer.begin(),
                          rContainer.end(),
                          [&is_reshaped, &sub_data_type_shape](auto& rValue) {
                            is_reshaped = DataTypeTraits<ValueType>::Reshape(rValue, sub_data_type_shape) || is_reshaped;
                           });
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+        }
+
+        return is_reshaped;
+    }
+
+    inline static PrimitiveType const * GetContiguousData(const ContainerType& rValue)
+    {
+        if constexpr(HasContiguousPrimitiveData) {
+            return rValue.data().begin();
```
matekelemen:
```suggestion
            return rValue.data().data();
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+        } else {
+            static_assert(!std::is_same_v<TDataType, TDataType>, "This should be only called if the rValue is contiguous only.");
```
matekelemen: same

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+        }
+    }
+
+    inline static PrimitiveType * GetContiguousData(ContainerType& rValue)
+    {
+        if constexpr(HasContiguousPrimitiveData) {
+            return rValue.data().begin();
```
matekelemen:
```suggestion
            return rValue.data().data();
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+        } else {
+            static_assert(!std::is_same_v<TDataType, TDataType>, "This should be only called if the rValue is contiguous only.");
```
matekelemen: same

kratos/utilities/data_type_traits.h
matekelemen: by the way, this approach won't work if your `ValueType` is dynamic, since each item can theoretically have a different size. This is true for all your `Size` implementations and any other function that checks the first value in the container and disregards the rest.

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+        }
+    }
+
+    inline static PrimitiveType * GetContiguousData(ContainerType& rValue)
+    {
+        if constexpr(HasContiguousPrimitiveData) {
+            return rValue.data().begin();
```
matekelemen:
```suggestion
            return rValue.data().data();
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+            for (unsigned int i = 0; i < rContainer.size1() * rContainer.size2(); ++i) {
+                DataTypeTraits<ValueType>::FillFromContiguousData(rContainer.data()[i], pContiguousDataBegin + i * stride);
+            }
+        }
+    }
+
+    ///@}
+};
+
+template<> class DataTypeTraits<std::string>
```
matekelemen, line 1166: what do you need `std::string` for?

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+        return rValue.data();
+    }
+
+    inline static void FillToContiguousData(
+        PrimitiveType* pContiguousDataBegin,
+        const ContainerType& rContainer)
+    {
+        for (unsigned int i = 0; i < rContainer.size(); ++i) {
+            *(pContiguousDataBegin++) = rContainer[i];
+        }
```
matekelemen:
```suggestion
        std::copy(rContainer.begin(),
                  rContainer.end(),
                  pContiguousDataBegin);
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+    }
+
+    inline static void FillFromContiguousData(
+        ContainerType& rContainer,
+        PrimitiveType const * pContiguousDataBegin)
+    {
+        for (unsigned int i = 0; i < rContainer.size(); ++i) {
+            rContainer[i] = *(pContiguousDataBegin++);
+        }
```
matekelemen:
```suggestion
        std::copy(pContiguousDataBegin,
                  pContiguousDataBegin + rContainer.size(),
                  rContainer.begin());
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,642 @@
...
+    }
+
+    ///@}
+
+private:
+    ///@name Private member variables
+    ///@{
+
+    std::conditional_t<HasContiguousPrimitiveData, void *, std::vector<PrimitiveType>> mData;
```
matekelemen: functions in a *traits* class are weird enough, but doing buffering in it is on a whole new level of weirdness. I'd move this into its own class together with the functions that operate on them.

kratos/utilities/data_type_traits.h
sunethwarna: It is required for the recursion which is used in the data types which are containers.

kratos/utilities/data_type_traits.h
sunethwarna: done :)

kratos/utilities/data_type_traits.h
sunethwarna: done :)

kratos/utilities/data_type_traits.h
sunethwarna: done :)

kratos/utilities/data_type_traits.h
sunethwarna: This works as it suppose to be. It is tested, and I have been using it in few other places as well. `static_assert(true)` does not work, it will always throw compiler error irrespective of the compiler branch, but this works.

kratos/utilities/data_type_traits.h
sunethwarna: done :)

kratos/utilities/data_type_traits.h
sunethwarna: same as above.

kratos/utilities/data_type_traits.h
sunethwarna: done :)

kratos/utilities/data_type_traits.h
sunethwarna: done :)

kratos/utilities/data_type_traits.h
sunethwarna: done :)

kratos/utilities/data_type_traits.h
sunethwarna: done :)

kratos/utilities/data_type_traits.h
sunethwarna: no data for this guy, the vector type has the begin which is hard coded with primitive type.

kratos/utilities/data_type_traits.h
sunethwarna: changed :)

kratos/utilities/data_type_traits.h
sunethwarna: same as above.

kratos/utilities/data_type_traits.h
sunethwarna: No begin for `Matrix` type as well. Have to use the begin which is the primitive data type pointer.

kratos/utilities/data_type_traits.h
sunethwarna: same as above.

kratos/utilities/data_type_traits.h
sunethwarna: For MPI Communication of std::strings.

kratos/utilities/data_type_traits.h
sunethwarna: done :)

kratos/utilities/data_type_traits.h
sunethwarna: done :)

kratos/utilities/data_type_traits.h
sunethwarna: done :)

kratos/utilities/data_type_traits.h
sunethwarna: Not possible with this data type.

kratos/utilities/data_type_traits.h
sunethwarna: done :)

kratos/utilities/data_type_traits.h
sunethwarna: done :)

sunethwarna:
@matekelemen I addressed all your comments, and this is ready for your review.

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,1457 @@
...
+                if (rValue.size() > 0) {
+                    return reinterpret_cast<PrimitiveType const *>(&rValue[0]);
+                } else {
+                    // not returning nullptr so, the return value can be subjected to
+                    // arithmetic operations
+                    return 0;
+                }
+            }
+        } else {
+            static_assert(!std::is_same_v<TDataType, TDataType>, "This should be only called if the rValue is contiguous only.");
```
matekelemen:
```suggestion
            static_assert(!std::is_same_v<TDataType, TDataType>, "GetContiguousData should only be called if rValue is contiguous.");
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,1457 @@
...
+                if (rValue.size() > 0) {
+                    return reinterpret_cast<PrimitiveType*>(&rValue[0]);
+                } else {
+                    // not returning nullptr so, the return value can be subjected to
+                    // arithmetic operations
+                    return 0;
+                }
+            }
+        } else {
+            static_assert(!std::is_same_v<TDataType, TDataType>, "This should be only called if the rValue is contiguous only.");
```
matekelemen:
```suggestion
            static_assert(!std::is_same_v<TDataType, TDataType>, "GetContiguousData should only be called if rValue is contiguous.");
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,1457 @@
...
+                if (rValue.size1() > 0 && rValue.size2() > 0) {
+                    return reinterpret_cast<PrimitiveType const*>(&rValue(0, 0));
+                } else {
+                    // not returning nullptr so, the return value can be subjected to
+                    // arithmetic operations
+                    return 0;
+                }
+            }
+        } else {
+            static_assert(!std::is_same_v<TDataType, TDataType>, "This should be only called if the rValue is contiguous only.");
```
matekelemen:
```suggestion
            static_assert(!std::is_same_v<TDataType, TDataType>, "GetContiguousData should only be called if rValue is contiguous.");
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,1457 @@
...
+                if (rValue.size1() > 0 && rValue.size2() > 0) {
+                    return reinterpret_cast<PrimitiveType*>(&rValue(0, 0));
+                } else {
+                    // not returning nullptr so, the return value can be subjected to
+                    // arithmetic operations
+                    return 0;
+                }
+            }
+        } else {
+            static_assert(!std::is_same_v<TDataType, TDataType>, "This should be only called if the rValue is contiguous only.");
```
matekelemen:
```suggestion
            static_assert(!std::is_same_v<TDataType, TDataType>, "GetContiguousData should only be called if rValue is contiguous.");
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,1457 @@
...
+     * @param pShapeEnd         End of the shape array.
+     */
+    template<class TIndexType = unsigned int>
+    static inline void Shape(
+        const ContainerType& rContainer,
+        TIndexType* pShapeBegin,
+        TIndexType* pShapeEnd)
+    {
+        KRATOS_ERROR_IF_NOT(std::distance(pShapeBegin, pShapeEnd) == 1)
+            << "Invalid dimensions given to fill for primitive data type [ Expected dimension == 1, provided shape = "
```
matekelemen:
```suggestion
            << "Invalid dimensions given to fill for std::string data type [ Expected dimension == 1, provided shape = "
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,1457 @@
...
+     *
+     * @param rValue                    Container to retireve the contiguous array pointer.
+     * @return PrimitiveType const*     Contiguous array pointer.
+     */
+    inline static PrimitiveType const * GetContiguousData(const ContainerType& rValue)
+    {
+        if constexpr(IsContiguous) {
+            return rValue.data();
+        } else {
+            static_assert(!std::is_same_v<TDataType, TDataType>, "This should be only called if the rValue is contiguous only.");
```
matekelemen:
```suggestion
            static_assert(!std::is_same_v<TDataType, TDataType>, "GetContiguousData should only be called if rValue is contiguous.");
```

kratos/utilities/data_type_traits.h
```diff
@@ -0,0 +1,1457 @@
...
+     *
+     * @param rValue                    Container to retireve the contiguous array pointer.
+     * @return PrimitiveType const*     Contiguous array pointer.
+     */
+    inline static PrimitiveType * GetContiguousData(ContainerType& rValue)
+    {
+        if constexpr(IsContiguous) {
+            return rValue.data();
+        } else {
+            static_assert(!std::is_same_v<TDataType, TDataType>, "This should be only called if the rValue is contiguous only.");
```
matekelemen:
```suggestion
            static_assert(!std::is_same_v<TDataType, TDataType>, "GetContiguousData should only be called if rValue is contiguous.");
```

matekelemen (dismissed):

kratos/utilities/data_type_traits.h
```diff
@@ -231,6 +252,27 @@ template<class TDataType, std::size_t TSize> class DataTypeTraits<array_1d<TData
...
+     *
+     * This method returns true if the @ref TCheckIndex of the dimension
+     * corresponds to a dynamic type. If not, this returns false.
+     *
+     * @tparam TCheckIndex          User input dimension index.
+     * @tparam TCurrentIndex        Used for the recursion only. Do not specify this.
+     * @return true                 If the input dimension index corresponds to dynamic data type.
+     * @return false                If the input dimension index corresponds to static data type.
+     */
+    template<unsigned int TCheckIndex, unsigned int TCurrentIndex = 0>
```
matekelemen: if the user isn't supposed to touch this, don't expose it. You could hide this template param in an impl function. For example:
```cpp
template <unsigned TCheckIndex>
static constexpr bool IsDimensionDynamic()
{
    return DataTypeTraits::IsDimensionDynamicImpl<TCheckIndex,0>();
}

private:
template <unsigned TCheckIndex, unsigned TCurrentIndex>
static constexpr bool IsDimensionDynamicImpl()
{
    /**/
}
```

kratos/utilities/data_type_traits.h
```diff
@@ -231,6 +252,27 @@ template<class TDataType, std::size_t TSize> class DataTypeTraits<array_1d<TData
...
+    static constexpr inline bool IsDimensionDynamic()
```
matekelemen:
```suggestion
    static constexpr bool IsDimensionDynamic()
```

kratos/utilities/data_type_traits.h
```diff
@@ -231,6 +252,27 @@ template<class TDataType, std::size_t TSize> class DataTypeTraits<array_1d<TData
...
+    {
+        if constexpr(TCheckIndex == TCurrentIndex) {
+            return false;
+        } else {
+            return ValueTrait::template IsDimensionDynamic<TCheckIndex, TCurrentIndex + 1>();
```
matekelemen:
```suggestion
            return ValueTraits::template IsDimensionDynamic<TCheckIndex, TCurrentIndex + 1>();
```

matekelemen: btw this will shoot off to an infinite loop **at compile time** if someone messes up the template params. Plz add a `constexpr` branch to prevent such cases.

matekelemen (changes requested):
Some comments that apply to all the implemented functions.

As a general question, why is `IsDimensionDynamic` necessary? Where are you planning to use it? I could understand if you implemented checks that make sure that all dynamic components in a container are of identical size, but I don't see any such checks.

kratos/utilities/data_type_traits.h
sunethwarna: This will be used in `HDFFile` where i do branching for MPI communication based on the dynamic types only since doing MPI communication for static dimensions are redundant andcostly.

sunethwarna: As far as I see, it cannot go recursively in an infinite loop since, you cannot define a data type without having a primitive type at the end.

kratos/utilities/data_type_traits.h
sunethwarna: Good point. and done :)

kratos/utilities/data_type_traits.h
sunethwarna: BTW, changed to `ValueTraits` ;)

matekelemen (approved):

PR #11471, opened 2023-08-10
[GeoMechanicsAnalysis] start_time and end_time single definition

WPK4FEM:
On ProjectParameters.json start_time was defined twice: in problem_data and in solver_settings
end_time was defined twice: in problem_data and in solver_settings - time_stepping
The definition in problem _data is kept, the others removed.

**🆕 Changelog**
The double defaults are removed from .hpp and .py files
Reading and setting in the pythonscripts now use only the
All ProjectParameters.json in applications/GeoMechanicsApplication/tests were adapted.

### DIFF
--- a/applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
+++ b/applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
@@ -117,7 +117,7 @@ class GeoMechanicsAnalysis(GeoMechanicsAnalysisBase):
             self.max_delta_time_factor = project_parameters["solver_settings"]["time_stepping"]["max_delta_time_factor"].GetDouble()
         self.max_delta_time      = self.delta_time * self.max_delta_time_factor
         self.number_cycles       = project_parameters["solver_settings"]["number_cycles"].GetInt()
-        
+
         self.max_iterations      = project_parameters["solver_settings"]["max_iterations"].GetInt()
         self.solution_type       = project_parameters["solver_settings"]["solution_type"].GetString()
         self.reset_displacements = project_parameters["solver_settings"]["reset_displacements"].GetBool()
--- a/applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
+++ b/applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
@@ -102,35 +102,39 @@ class GeoMechanicsAnalysis(GeoMechanicsAnalysisBase):
     def __init__(self, model, project_parameters):
         super().__init__(model, project_parameters)
 
+        # time step related stuff
+        self.start_time          = project_parameters["problem_data"]["start_time"].GetDouble()
+        self.end_time            = project_parameters["problem_data"]["end_time"].GetDouble()
+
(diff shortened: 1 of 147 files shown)


applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
```diff
@@ -102,50 +102,57 @@ class GeoMechanicsAnalysis(GeoMechanicsAnalysisBase):
...
                 KratosMultiphysics.Logger.PrintInfo(self._GetSimulationName(), "reducing delta_time to max_delta_time: ", self.max_delta_time)
-            t = self._GetSolver().GetComputingModelPart().ProcessInfo[KratosMultiphysics.TIME]
-            new_time = t + self.delta_time
-            if (new_time > self.end_time):
-                new_time = self.end_time
-                self.delta_time = new_time - t
+
+            # maximize delta_time to avoid exceeding the end_time
+            t               = self._GetSolver().GetComputingModelPart().ProcessInfo[KratosMultiphysics.TIME]
+            self.delta_time = min( self.delta_time, self.end_time - t )
```
mcgicjn2, line 152: The logic of this section could be compressed into a single line. i.e 
`self.delta_time = min(self.delta_time, self.end_time -t, self.max_delta_time)`
However I see the purpose for logging purposes.

applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
```diff
@@ -102,50 +102,57 @@ class GeoMechanicsAnalysis(GeoMechanicsAnalysisBase):
...
         if self._GetSolver().settings["reset_displacements"].GetBool():
             old_total_displacements = [node.GetSolutionStepValue(KratosGeo.TOTAL_DISPLACEMENT)
                                        for node in self._GetSolver().GetComputingModelPart().Nodes]
 
         self._GetSolver().solver.SetRebuildLevel(self.rebuild_level)
 
         while self.KeepAdvancingSolutionLoop():
+            # check against max_delta_time should only be necessary when trying to scale up
             if (self.delta_time > self.max_delta_time):
                 self.delta_time = self.max_delta_time
```
mcgicjn2, line 147:
```suggestion
            self.delta_time = min(self.delta_time, self.max_delta_time)
```

mcgicjn2 (approved):
The changes are minimal and just suggestions, so I think you can merge after my one suggestion. So I have this on approve.

I will add a test issue for time stepping to our backlog

applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
WPK4FEM: Indeed

applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
WPK4FEM: That would imply that we unconditionally get the logging about "reducing delta_time to max_delta_time"
even when that doesn't happen.

PR #11473, opened 2023-08-13
[HDF5] Update - 1 - Refactor HDF5File

### LINKED ISSUE #10728: [HDF5] Replace Raw MPI Calls
## Description
The `HDF5Application` has to do MPI synchronization in its internals (obviously), but the C++ side always invokes raw MPI calls on `MPI_COMM_WORLD` instead of relying on `DataCommunicator`. This will lead to hangs and other unwanted behaviour if proper MPI communication (with subcommunicators) ever gets implemeted in core.

This is not an urgent problem and I definitely don't have time to do it because it involves extensive changes to the `HDF5Application`, but it will have to be taken care of sooner or later.

## ToDo
- update the interface of `HDF5Application` C++ internals to take the `DataCommunicator` of the `ModelPart`s they are acting on
- replace raw MPI calls with the corresponding functions of `DataCommunicator`

sunethwarna:
**📝 Description**
This PR refactors `HDF5File`. Followings changes/updates were applied.
1. `HDF5File` ctor now requires `DataCommunicator`. The python ctor still has a ctor without `DataCommunicator`, in this case a warning is thrown. This is planned to be removed once the python side of the `HDF5Application` is updated.
2. All the raw `mpi` calls are now done via `DataCommunicator`.
3. `HDF5FileSerial` and `HDF5FileParallel` are removed. Everything is done now in `HDF5File`
4. Extended the attribute and dataset interfaces for data types `array_1d<double, 4>`, `array_1d<double, 6>`, `array_1d<double, 9>`
5. The `HDF5FileParallel` version throws an error if the dataset is existing when writing, but `HDF5SerialFile` retireves the existing and changes it. The behaviour is now uniform in mpi and openmp.
6. The tests and occurances of `HDFFile` updated.

Requires #11469 

This closes #10728 

**🆕 Changelog**
- As above.


applications/HDF5Application/CMakeLists.txt
```diff
@@ -96,7 +94,11 @@ include_directories(${HDF5_INCLUDE_DIRS})
 add_definitions(${HDF5_DEFINITIONS})
 
 add_library(KratosHDF5Core SHARED ${KRATOS_HDF5_APPLICATION_CORE} ${KRATOS_HDF5_APPLICATION_TESTING_SOURCES})
-target_link_libraries(KratosHDF5Core PUBLIC KratosCore PRIVATE ${HDF5_C_LIBRARIES} ${KRATOS_HDF5_APPLICATION_EXTRA_LIBS})
+if(USE_MPI)
+  target_link_libraries(KratosHDF5Core PUBLIC KratosCore KratosMPICore PRIVATE ${HDF5_C_LIBRARIES} ${KRATOS_HDF5_APPLICATION_EXTRA_LIBS})
+else(USE_MPI)
+  target_link_libraries(KratosHDF5Core PUBLIC KratosCore PRIVATE ${HDF5_C_LIBRARIES} ${KRATOS_HDF5_APPLICATION_EXTRA_LIBS})
+endif(USE_MPI)
```
philbucher, line 101: I recommend splitting this into a separate MPI-Extension, same as the other apps

applications/HDF5Application/tests/test_utils.cpp
```diff
@@ -479,17 +480,17 @@ HDF5::File GetTestFile()
...
-HDF5::FileSerial GetTestSerialFile()
+HDF5::File GetTestSerialFile()
 {
     Parameters file_params(R"(
         {
             "file_name" : "test.h5",
             "file_driver": "core"
         })");
-    return HDF5::FileSerial(file_params);
+    return HDF5::File(ParallelEnvironment::GetDefaultDataCommunicator(), file_params);
```
philbucher:
```suggestion
    return HDF5::File(Testing::GetDefaultDataCommunicator(), file_params);
```

applications/HDF5Application/tests/test_utils.cpp
```diff
@@ -7,8 +7,9 @@
 #include "includes/model_part.h"
 #include "includes/kratos_components.h"
 #include "testing/testing.h"
-#include "custom_io/hdf5_file_serial.h"
+#include "custom_io/hdf5_file.h"
 #include "custom_utilities/registered_component_lookup.h"
+#include "includes/parallel_environment.h"
```
philbucher:
```suggestion
```

this should only be used in very special circumstances

applications/HDF5Application/tests/test_hdf5_file.cpp
```diff
@@ -18,6 +18,7 @@
 // Project includes
 #include "testing/testing.h"
 #include "includes/kratos_parameters.h"
+#include "includes/parallel_environment.h"
```
philbucher, line 21:
```suggestion
```

applications/HDF5Application/custom_processes/hdf5_xdmf_connectivities_writer_process.cpp
```diff
@@ -1,8 +1,9 @@
 #include "custom_processes/hdf5_xdmf_connectivities_writer_process.h"
 
-#include "custom_io/hdf5_file_serial.h"
+#include "custom_io/hdf5_file.h"
 #include "utilities/openmp_utils.h"
 #include "includes/kratos_parameters.h"
+#include "includes/parallel_environment.h"
```
philbucher:
```suggestion
```

applications/HDF5Application/custom_processes/hdf5_xdmf_connectivities_writer_process.cpp
```diff
@@ -18,7 +19,7 @@ XdmfConnectivitiesWriterProcess::XdmfConnectivitiesWriterProcess(const std::stri
                 "file_access_mode": "read_write"
             })");
     file_params["file_name"].SetString(rFileName);
-    mpFile = FileSerial::Pointer(new FileSerial(file_params));
+    mpFile = File::Pointer(new File(ParallelEnvironment::GetDefaultDataCommunicator(), file_params));
```
philbucher: this should come from outside

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
-    ///@name Protected Operations
-    ///@{
-    hid_t GetFileId() const;
-    ///@}
-
 private:
     ///@name Member Variables
     ///@{
+
+    DataCommunicator const * mpDataCommunicator;
```
philbucher, line 417: why not `const &`?

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -22,67 +49,9 @@ bool IsPath(const std::string& rPath)
     return regex_match(rPath, std::regex("(/[\\w\\(\\)]+)+"));
```
philbucher, line 68: perhaps this can be done better with `std::filesystem`

philbucher (commented):
:+1:

applications/HDF5Application/custom_io/hdf5_file.h
matekelemen: I'm the one to blame here, but I'll stand by my opinion. Storing a `const` member always messes up move and copy semantics. In this case, copy construction/assignment and move assignment doesn't make sense, but I don't see why move construction should be forbidden.

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
     File& operator=(File&& rOther);
 
     ///@}
-    ///@name Operations
+    ///@name Public operations
     ///@{
 
-    /// Check if path exists in HDF5 file.
+    /**
+     * @brief Checks if the given @ref rPath exists.
```
matekelemen: Doxygen can't `@ref` locally defined tokens (such as function arguments or template parameters), only tokens that have their own entry in the docs (such as functions, classes, enums, etc.). Plz change these `@ref`s to something else (`@a` for example) in the rest of the docstrings too.

```suggestion
     * @brief Checks if the given @a rPath exists.
```

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+     *
+     * @throws If a string with invalid characters are given.
```
matekelemen:
```suggestion
     * @throws If a string with invalid characters is given.
```

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+     *
+     * @param rPath     Path to be checked for.
```
matekelemen:
```suggestion
     * @param rPath     Path to be checked.
```

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+     * @return true     If @ref rPath exists.
+     * @return false    If @ref rPath does not exist.
```
matekelemen: There should only be a single `return` tag. This also applies to the rest of the docstrings.

```suggestion
     * @return true     True if @a rPath exists, otherwise false.
```

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+     * @return true     If @ref rPath exists and it is a group.
+     * @return false    Either @ref rPath does not exist or it is not a group.
+     */
     bool IsGroup(const std::string& rPath) const;
 
+    /**
+     * @brief Checks if @ref rPath is a data set.
+     *
+     * @param rPath     Path to be checked for.
+     * @return true     If @ref rPath exists and it is a data set.
```
matekelemen: Minor nitpick, but you're using both `dataset` and `data set`. Official grammar seems to support `data set` but I personally prefer `dataset` in a programming context to prevent confusion with `set`ting data.

In any case, plz pick one and stick to it.

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+     * @return false            If attribute does not exist in the dataset or group.
+     */
+    bool HasAttribute(
+        const std::string& rObjectPath,
+        const std::string& rName) const;
+
+    /**
+     * @brief Checks if the attribute is of the specified type.
+     *
+     * @tparam TDataType        Data type to check against.
```
matekelemen:
```suggestion
     * @tparam TDataType        Data type to check.
```

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+     *
+     * @param rObjectPath       Path of a dataset or group.
+     * @param rName             Attribute name to be deleted.
+     */
+    void DeleteAttribute(
+        const std::string& rObjectPath,
+        const std::string& rName);
+
+    /**
+     * @brief Get the Attribute Names list.
```
matekelemen:
```suggestion
     * @brief Get the list of attributes' names.
```

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+    /**
+     * @brief Get the data set names under a group.
+     *
+     * @param rGroupPath                    Group path.
+     * @return std::vector<std::string>     List of sub data set names.
+     */
     std::vector<std::string> GetDataSetNames(const std::string& rGroupPath) const;
 
+    /**
+     * @brief Add group to the path recursively.
```
matekelemen:
```suggestion
     * @brief Add a group to the path recursively.
```

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
-    virtual void WriteDataSet(const std::string& rPath, const Vector<array_1d<double, 3>>& rData, WriteInfo& rInfo);
+    template<class TDataType>
+    void WriteAttribute(
+        const std::string& rObjectPath,
+        const std::string& rName,
+        const TDataType& rValue);
 
-    virtual void WriteDataSet(const std::string& rPath, const Matrix<int>& rData, WriteInfo& rInfo);
+    /**
+     * @brief Write attributes in Parameters object to dataset or group.
```
matekelemen: Here's your chance to use `@ref`. `Parameters` is a class that has its own entry in the docs.

```suggestion
     * @brief Write attributes provided in a @ref Parameters object to a dataset or group.
```

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+     *
+     * @param rObjectPath           Dataset or group path.
+     * @param Attributes            Attributes to be written to.
+     */
+    void WriteAttribute(
+        const std::string& rObjectPath,
+        const Parameters Attributes);
 
-    virtual void WriteDataSet(const std::string& rPath, const Matrix<double>& rData, WriteInfo& rInfo);
```
matekelemen: Awesome!

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+    /**
+     * @brief Write data set to the hDF5 file.
```
matekelemen:
```suggestion
     * @brief Write a data set to the HDF5 file.
```

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+     *
+     * Performs collective write in MPI. The data is written blockwise according to
+     * processor rank.
+     *
+     * @tparam TDataType    Data type of the Data.
```
matekelemen:
```suggestion
```

This doesn't provide any new info to what can already be divined from the template parameter's name.

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+     * @param rPath         Path to which the data is written.
+     * @param rData         Data to be writtent.
```
matekelemen:
```suggestion
     * @param rData         Data to be written.
```

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
-    virtual void WriteDataSetIndependent(const std::string& rPath, const Vector<double>& rData, WriteInfo& rInfo);
-
-    virtual void WriteDataSetIndependent(const std::string& rPath,
-                                         const Vector<array_1d<double, 3>>& rData, WriteInfo& rInfo);
-
-    virtual void WriteDataSetIndependent(const std::string& rPath, const Matrix<int>& rData, WriteInfo& rInfo);
-
-    virtual void WriteDataSetIndependent(const std::string& rPath, const Matrix<double>& rData, WriteInfo& rInfo);
+    template<class TDataType>
+    void WriteDataSetIndependent(
```
matekelemen, line 280: You have an `enum` for `Collective` and `Independent` right? If so, I wouldn't encode that in the function's name but either pass it as an argument or a template param.

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+        const std::string& rPath,
+        const TDataType& rData,
+        WriteInfo& rInfo);
 
+    /**
+     * @brief Get the Data Dimensions at the specified path.
```
matekelemen:
```suggestion
     * @brief Get the shape of the dataset stored at the specified path.
```

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+     *
+     * @param rPath                     Dataset path.
+     * @return std::vector<unsigned>    Dimensions of the data set.
```
matekelemen: I imagine you `throw` if `rPath` doesn't exist or is a group, right? If so, mention it in the docs.

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
     std::vector<unsigned> GetDataDimensions(const std::string& rPath) const;
 
+    /**
+     * @brief Checks if the dataset at path is of int type.
+     *
+     * @param rPath                     Dataset path.
+     * @return true                     If dataset of int type.
+     * @return false                    If dataset is not of int type.
+     */
     bool HasIntDataType(const std::string& rPath) const;
```
matekelemen, line 300: **argh**

Do we need to keep these? Can't we use templates instead? If backward compatibility is the issue, just plz add a template as well (same as `Parameters::Has` and friends).

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+     * @return true                     If dataset is of TDataType.
+     * @return false                    If dataset is not of TDataType.
+     */
+    template<class TDataType>
+    bool HasDataType(const std::string& rPath) const;
+
+    /**
+     * @brief Flush the content to HDF5 file.
+     *
+     */
```
matekelemen:
```suggestion
    /// @brief Flush the buffer to the HDF5 file.
```

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
 
-    /// Terminate access to the HDF5 file.
-    /*
-     *  @throws If the underlying HDF5 call fails.
+    /**
+     * @brief Terminate access to the HDF5 file.
+     *
+     * @throws If the underlying HDF5 call fails.
+     *
+     * @return * void
```
matekelemen, line 330:
```suggestion
```

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
      */
     void Close();
 
+    /**
+     * @brief Get the File Size.
+     *
+     * @return unsigned
```
matekelemen:
```suggestion
```

You don't have to explicitly state the return type, doxygen can deduce that. What you need to provide in `@return` is an explanation for what the return value actually means.

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+     * @brief Get the Echo Level.
+     *
+     * @return int
+     */
     int GetEchoLevel() const;
 
+    /**
+     * @brief Set the Echo Level.
+     *
+     * @param Level
```
matekelemen: Either provide proper documentation, or just omit it. This is empty.


```suggestion
```

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
-    // Return the total number of processes with file access.
-    virtual unsigned GetTotalProcesses() const;
-
-    template<class TScalar>
-    void ReadAttribute(const std::string& rObjectPath, const std::string& rName, TScalar& rValue);
+    /**
+     * @brief Get the Data Communicator.
+     *
+     * @return const DataCommunicator&
+     */
```
matekelemen: Same. We don't need Vincente-style documentation.


```suggestion
```

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+     * @brief Get the process rank.
+     *
+     * @return unsigned
+     */
+    unsigned GetPID() const;
 
-    template<class TScalar>
-    void ReadAttribute(const std::string& rObjectPath, const std::string& rName, Matrix<TScalar>& rValue);
+    /**
+     * @brief Get the Total number of processes with file access.
```
matekelemen:
```suggestion
     * @brief Get the total number of processes with file access.
```

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+     *
+     * @return unsigned
+     */
+    unsigned GetTotalProcesses() const;
```
matekelemen, line 348: How about `ProcessCount` or `GetNumberOfProcesses`?

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+     */
+    template<class TDataType>
+    void ReadAttribute(
+        const std::string& rObjectPath,
+        const std::string& rName,
+        TDataType& rValue);
 
-    void ReadAttribute(const std::string& rObjectPath, const std::string& rName, array_1d<double, 3>& rValue);
+    /**
+     * @brief Read attributes to Parameters object from dataset or group.
```
matekelemen:
```suggestion
     * @brief Read attributes from a dataset or group to a @ref Parameters object.
```

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
+     * @param rObjectPath           Dataset or group path.
+     * @return Parameters           Parameters object containing attribute name, value pairs.
+     */
+    Parameters ReadAttribute(const std::string& rObjectPath);
 
-    /// Read a data set from the HDF5 file.
     /**
+     * @brief Read a data set from the HDF5 file.
+     *
      * Performs collective read in MPI. Throws if out of range.
```
matekelemen, line 375: > Throws if out of range.

=> `@throws`

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
-                                       Matrix<int>& rData,
-                                       unsigned StartIndex,
-                                       unsigned BlockSize);
-
-    virtual void ReadDataSetIndependent(const std::string& rPath,
-                                       Matrix<double>& rData,
-                                       unsigned StartIndex,
-                                       unsigned BlockSize);
+    template<class TDataType>
+    void ReadDataSetIndependent(
```
matekelemen, line 404: Same suggestion to pass `Independent` or `Collective` as a function argument or template parameter.

applications/HDF5Application/custom_io/hdf5_file.h
```diff
@@ -85,216 +99,409 @@ class KRATOS_API(HDF5_APPLICATION) File
...
     ///@name Member Variables
     ///@{
+
+    DataCommunicator const * mpDataCommunicator;
+
     std::string m_file_name;
+
     hid_t m_file_id = -1; // Default invalid file id.
+
     int m_echo_level = 0;
```
matekelemen: Now's the chance to fix the naming.


```suggestion
    std::string mFileName;

    hid_t mFileId = -1; // Default invalid file id.

    int mEchoLevel = 0;
```

matekelemen: a minor nitpick about the order of member variables:

In general, there are 2 aspects to keep in mind when deciding on the order of member variables:
- memory alignment and padding (not too important in this case because the order of variables won't have much effect; just don't put the `std::string` in the middle)
- memory access patterns; i.e.: which members get accessed most often? You can improve your chances of hitting the cache if you put your most often used member vars in the front (`m_file_id`).

matekelemen (commented):
I'll review this in chunks and start with `hdf5_file.h`

applications/HDF5Application/custom_utilities/data_type_utilities.h
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Kratos
+{
+namespace HDF5
+{
+namespace Internals
+{
+
+// H5 data types
+template <class TDataType> hid_t inline GetPrimitiveH5Type()
```
matekelemen: it's a header-only template so you don't need the `inline`.
```suggestion
template <class TDataType> hid_t GetPrimitiveH5Type()
```

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -195,80 +166,151 @@ File::~File()
...
     KRATOS_TRY;
+
     // Expects a valid path.
-    KRATOS_ERROR_IF_NOT(Internals::IsPath(rPath)) << "Invalid path: \"" << rPath << "\". Path should start with \"/\" and should only have characters A-Z, a-z, 0-9, \"/\", and \"_\"." << std::endl;
+    KRATOS_ERROR_IF_NOT(Internals::IsPath(rPath))
+        << "Invalid path: \"" << rPath
+        << "\". Path should start with \"/\" and should only have characters A-Z, a-z, 0-9, \"/\", and \"_\"."
+        << std::endl;
 
     std::vector<std::string> splitted_path = StringUtilities::SplitStringByDelimiter(rPath, '/');
```
matekelemen:
```suggestion
    std::vector<std::string> split_path = StringUtilities::SplitStringByDelimiter(rPath, '/');
```

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -195,80 +166,151 @@ File::~File()
...
+
+    if (HasPath(rPath) == false) {// Expects a valid path.
         return false;
+    }
 
     H5O_info_t object_info;
     KRATOS_ERROR_IF(H5Oget_info_by_name(m_file_id, rPath.c_str(), &object_info, H5P_DEFAULT) < 0)
         << "H5Oget_info_by_name failed." << std::endl;
 
     return (object_info.type == H5O_TYPE_DATASET);
```
matekelemen:
```suggestion
    return object_info.type == H5O_TYPE_DATASET;
```

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -195,80 +166,151 @@ File::~File()
...
     KRATOS_TRY;
-    if (HasPath(rPath) == false) // Expects a valid path.
+
+    if (HasPath(rPath) == false) {// Expects a valid path.
         return false;
+    }
 
     H5O_info_t object_info;
     KRATOS_ERROR_IF(H5Oget_info_by_name(m_file_id, rPath.c_str(), &object_info, H5P_DEFAULT) < 0)
         << "H5Oget_info_by_name failed." << std::endl;
```
matekelemen:
```suggestion
    H5O_info_t object_info;
    const auto return_code = H5Oget_info_by_name(m_file_id, rPath.c_str(), &object_info, H5P_DEFAULT);
    KRATOS_ERROR_IF(return_code < 0)
        << "H5Oget_info_by_name failed with code " << return_code << std::endl;
```

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -195,80 +166,151 @@ File::~File()
...
+    KRATOS_ERROR_IF(attr_id < 0) << "H5Aopen_by_name failed." << std::endl;
+
+    auto attr_type_id = H5Aget_type(attr_id);
+    KRATOS_ERROR_IF(attr_type_id < 0) << "H5Aget_type failed." << std::endl;
+
+    auto is_valid_type = H5Tequal(mem_type_id, attr_type_id);
+
+    KRATOS_ERROR_IF(H5Tclose(attr_type_id) < 0) << "H5Tclose failed." << std::endl;
+    KRATOS_ERROR_IF(is_valid_type < 0) << "H5Tequal failed." << std::endl;
+    KRATOS_ERROR_IF(H5Aclose(attr_id) < 0) << "H5Aclose failed." << std::endl;
```
matekelemen: print error codes too. I won't mention it elsewhere but plz consider applying it at every call to the HDF5 lib.

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -336,32 +381,42 @@ std::vector<std::string> File::GetLinkNames(const std::string& rGroupPath) const
...
     KRATOS_TRY;
+
     std::vector<std::string> names;
     std::vector<std::string> link_names = GetLinkNames(rGroupPath);
     names.reserve(link_names.size());
-    for (const auto& r_name : link_names)
-        if (IsGroup(rGroupPath + '/' + r_name))
+    for (const auto& r_name : link_names) {
+        if (IsGroup(rGroupPath + '/' + r_name)) {
             names.push_back(r_name);
```
matekelemen: you won't ever reuse the items in `link_names` so you might as well `move` them and prevent a copy.

```suggestion
    for (auto& r_name : link_names) {
        if (IsGroup(rGroupPath + '/' + r_name)) {
            names.emplace_back(std::move(r_name));
```

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -195,80 +166,151 @@ File::~File()
...
+
     KRATOS_CATCH("");
 }
 
 std::vector<std::string> File::GetAttributeNames(const std::string& rObjectPath) const
 {
     KRATOS_TRY;
+
     constexpr unsigned max_ssize = 100;
     char buffer[max_ssize];
```
matekelemen:
```suggestion
    std::array<char,100> buffer;
```

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -372,253 +427,158 @@ void File::AddPath(const std::string& rPath)
...
-    KRATOS_CATCH("");
-}
-
-void File::WriteDataSet(const std::string& rPath, const Vector<array_1d<double, 3>>& rData, WriteInfo& rInfo)
-{
-    KRATOS_TRY;
-    KRATOS_ERROR << "Calling the base class method. Please override in the derived class." << std::endl;
-    KRATOS_CATCH("");
-}
+    for (auto itr = Attributes.begin(); itr != Attributes.end(); ++itr) {
```
matekelemen, line 504: y no range based `for`?

```suggestion
    for (auto attribute : Attributes) {
```

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -372,253 +427,158 @@ void File::AddPath(const std::string& rPath)
...
-    KRATOS_CATCH("");
-}
-
-void File::WriteDataSetIndependent(const std::string& rPath, const Vector<int>& rData, WriteInfo& rInfo)
-{
-    KRATOS_TRY;
-    KRATOS_ERROR << "Calling the base class method. Please override in the derived class." << std::endl;
-    KRATOS_CATCH("");
-}
+            for (const auto& v : *itr) {
```
matekelemen, line 520:
```suggestion
            for (auto v : *itr) {
```

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -372,253 +427,158 @@ void File::AddPath(const std::string& rPath)
...
-    KRATOS_ERROR_IF((dtype_id = H5Dget_type(dset_id)) < 0)
-        << "H5Dget_type failed." << std::endl;
-    H5T_class_t type = H5Tget_class(dtype_id);
-    KRATOS_ERROR_IF(type == H5T_NO_CLASS) << "Invalid data type." << std::endl;
-    KRATOS_ERROR_IF(H5Tclose(dtype_id) < 0) << "H5Tclose failed." << std::endl;
-    KRATOS_ERROR_IF(H5Dclose(dset_id) < 0) << "H5Dclose failed." << std::endl;
-
-    return (type == H5T_INTEGER);
-    KRATOS_CATCH("");
+    return HasDataType<int>(rPath);
```
matekelemen, line 584: ok so this is for backward compatibility. Nevermind my earlier comment then, sry for the noise.

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -662,272 +622,229 @@ void File::SetEchoLevel(int Level)
...
-    KRATOS_ERROR_IF(attr_id < 0) << "H5Aopen_by_name failed." << std::endl;
-
-    // Check data type.
-    attr_type_id = H5Aget_type(attr_id);
-    KRATOS_ERROR_IF(attr_type_id < 0) << "H5Aget_type failed." << std::endl;
-    htri_t is_valid_type = H5Tequal(mem_type_id, attr_type_id);
-    KRATOS_ERROR_IF(H5Tclose(attr_type_id) < 0) << "H5Tclose failed." << std::endl;
-    KRATOS_ERROR_IF(is_valid_type < 0) << "H5Tequal failed." << std::endl;
-    KRATOS_ERROR_IF(is_valid_type == 0) << "Memory and file data types are different." << std::endl;
+    using type_traits = DataTypeTraits<TDataType>;
```
matekelemen:
```suggestion
    using TypeTraits = DataTypeTraits<TDataType>;
```

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -662,272 +622,229 @@ void File::SetEchoLevel(int Level)
...
+    Parameters result;
 
-    // Check data type.
-    attr_type_id = H5Aget_type(attr_id);
-    KRATOS_ERROR_IF(attr_type_id < 0) << "H5Aget_type failed." << std::endl;
-    htri_t is_valid_type = H5Tequal(mem_type_id, attr_type_id);
-    KRATOS_ERROR_IF(H5Tclose(attr_type_id) < 0) << "H5Tclose failed." << std::endl;
-    KRATOS_ERROR_IF(is_valid_type < 0) << "H5Tequal failed." << std::endl;
-    KRATOS_ERROR_IF(is_valid_type == 0) << "Attribute \"" << rName << "\" is not a string." << std::endl;
+    const auto& attribute_names = GetAttributeNames(rObjectPath);
```
matekelemen:
```suggestion
    const auto attribute_names = GetAttributeNames(rObjectPath);
```

I know it's the same thing, but it's confusing and it rapes my brain whenever I see a temporary captured as a ref.

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -662,272 +622,229 @@ void File::SetEchoLevel(int Level)
...
-    KRATOS_ERROR_IF(H5Aread(attr_id, mem_type_id, buffer) < 0) << "H5Aread failed." << std::endl;
-    KRATOS_ERROR_IF(H5Aclose(attr_id) < 0) << "H5Aclose failed." << std::endl;
-    rValue = std::string(buffer, dims[0]);
-    KRATOS_INFO_IF("HDF5Application", GetEchoLevel() == 2)
-        << "Read time \"" << rObjectPath << '/' << rName
-        << "\": " << timer.ElapsedSeconds() << std::endl;
-    KRATOS_CATCH("Path: \"" + rObjectPath + '/' + rName + "\".");
-}
+    for (const auto& attribute_name : attribute_names) {
+        const auto& shape = GetAttributeDimensions(rObjectPath, attribute_name);
```
matekelemen:
```suggestion
        const auto shape = GetAttributeDimensions(rObjectPath, attribute_name);
```

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -662,272 +622,229 @@ void File::SetEchoLevel(int Level)
...
-    KRATOS_TRY;
-    Vector<double> vector_value;
-    ReadAttribute(rObjectPath, rName, vector_value);
-    KRATOS_ERROR_IF(vector_value.size() > 3)
-        << "Invalid size (" << vector_value.size() << ") for array_1d!" << std::endl;
-    rValue = vector_value;
-    KRATOS_CATCH("Path: \"" + rObjectPath + '/' + rName + "\".");
-}
+        if (HasAttributeType<int>(rObjectPath, attribute_name)) {
+            if (shape.size() == 0) {
```
matekelemen:
```suggestion
            if (shape.empty()) {
```

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -662,272 +622,229 @@ void File::SetEchoLevel(int Level)
...
+                int value{};
```
matekelemen:
```suggestion
                int value;
```

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -662,272 +622,229 @@ void File::SetEchoLevel(int Level)
...
+    hid_t& rDataSpaceId,
+    const std::vector<hsize_t>& rDims,
+    const std::string& rPath)
 {
-    KRATOS_TRY;
-    KRATOS_ERROR << "Calling the base class method. Please override in the derived class." << std::endl;
+    KRATOS_TRY
+
+    if (!HasPath(rPath)) {
+        CreateNewDataSet(rDataSetId, rDataSpaceId, Internals::GetPrimitiveH5Type<TDataType>(), rDims, rPath);
```
matekelemen: why? shouldn't you throw when you want to get something that doesn't exist?

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -662,272 +622,229 @@ void File::SetEchoLevel(int Level)
...
     KRATOS_CATCH("");
 }
 
-void File::ReadDataSetIndependent(const std::string& rPath, Matrix<double>& rData, unsigned StartIndex, unsigned BlockSize)
+template<class TDataType>
+void File::GetDataSet(
+    hid_t& rDataSetId,
+    hid_t& rDataSpaceId,
+    const std::vector<hsize_t>& rDims,
+    const std::string& rPath)
```
matekelemen:
```suggestion
template<class TDataType>
void File::GetDataSet(
    hid_t& rDataSetId,
    hid_t& rDataSpaceId,
    const std::vector<hsize_t>& rDims,
    const std::string& rPath) const
```

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -971,19 +888,294 @@ void File::SetFileDriver(const std::string& rDriver, hid_t FaplId) const
...
-template void File::WriteAttribute(const std::string& rObjectPath, const std::string& rName, const Matrix<double>& rValue);
+template<class TDataType, File::DataTransferMode TDataTransferMode>
+void File::WriteDataSetImpl(
+    const std::string& rPath,
+    const TDataType& rData,
+    WriteInfo& rInfo)
+{
+    KRATOS_TRY
+
+    using type_trait = DataTypeTraits<TDataType>;
```
matekelemen:
```suggestion
    using TypeTraits = DataTypeTraits<TDataType>;
```

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -971,19 +888,294 @@ void File::SetFileDriver(const std::string& rDriver, hid_t FaplId) const
...
+            KRATOS_ERROR_IF(number_of_local_primitive_data_values > 0 && max_size != local_shape[1])
+                << "Mismatching shapes found in different ranks. All ranks should have the same shapes in data sets.";
+
+            local_shape[1] = max_size;
+        }
+    }
+
+    // local_reduced_shape holds the max 2d flattened shape if the local_shape dimensions
+    // are higher than 2.
+    std::vector<hsize_t> global_shape(global_dimension, 0), local_reduced_shape(global_dimension, 0), local_shape_start(global_dimension, 0);
```
matekelemen: `global_dimension` is a compile time constant, so these could be `std::array`s.

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -971,19 +888,294 @@ void File::SetFileDriver(const std::string& rDriver, hid_t FaplId) const
...
+
+    KRATOS_CATCH("Path: \"" + rPath + "\".");
+}
+
+template<class TDataType, File::DataTransferMode TDataTransferMode>
+void File::ReadDataSetImpl(
+    const std::string& rPath,
+    TDataType& rData,
+    const unsigned StartIndex,
+    const unsigned BlockSize)
```
matekelemen:
```suggestion
template<class TDataType, File::DataTransferMode TDataTransferMode>
void File::ReadDataSetImpl(
    const std::string& rPath,
    TDataType& rData,
    const unsigned StartIndex,
    const unsigned BlockSize) const
```

matekelemen (commented):
`hdf5_file.cpp`

applications/HDF5Application/CMakeLists.txt
sunethwarna: Would it be ok if I do it in a seperate PR? I think that will be cleaner, because there are some other files which also needs to go into this mpi-extension.

applications/HDF5Application/tests/test_utils.cpp
sunethwarna: removed :)

applications/HDF5Application/tests/test_hdf5_file.cpp
sunethwarna: removed :)

applications/HDF5Application/custom_processes/hdf5_xdmf_connectivities_writer_process.cpp
sunethwarna: removed :)

applications/HDF5Application/custom_processes/hdf5_xdmf_connectivities_writer_process.cpp
sunethwarna: Since xdmf generation requires the indices which is obtained in serial only, this process is run with the serial data communicator.

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: I am not sure how to do this with `std::filesystem` since these paths are not actual file paths, but hdf5 paths within one hdf5 file.

applications/HDF5Application/tests/test_utils.cpp
sunethwarna: removed :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: changed to `dataset` :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: yes, for backward compatibility.

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: I would keep it as it is.

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: It is already like that, this is kepy for backward compatibiltiy.

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: done :)

applications/HDF5Application/custom_utilities/data_type_utilities.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: Now this method is moved where it is used, since the new implementation requires it to be called from one place only.

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: removed the method.

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: I would keep some, otherwise I have to template some callee methods as well.

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: this is not possible, since the name of the attribute is only available in the itr.

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: I would keep it const, since it gives the indication for the reader that it is not being mutated.

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: I would keep it, because then we don't have to know what kind of a return is given by the method. If it is temporary, it is the same, if it is not a temporary, then it gets the const ref from the method.

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: Same as above.

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.h
sunethwarna: done :)

sunethwarna:
@matekelemen @philbucher I applied your suggestions. Could you have a look? Thanks :D

applications/HDF5Application/custom_io/hdf5_file.cpp
matekelemen: then

```suggestion
            for (const auto v : *itr) {
```

my main beef is with the ref (`Parameters` is **not** a container)

applications/HDF5Application/custom_io/hdf5_file.cpp
matekelemen: those callees are already templated so you'd just be adding an extra template param. If you really don't want to track the array's size, then use a bounded array or a container with small size optimization, because this line adds 3 unnecessary allocations.

applications/HDF5Application/CMakeLists.txt
philbucher: ok for me

philbucher:
LGTM

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: I would like to avoid a copy operation, and I don't see a problem by using the reference in this case since, we read the underlying data and disregard the itr.  We don't use it anymore. I think, in this case this is safe and we avoid copy operations.

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: ok done... removed the callee method totally. :D.

sunethwarna:
@matekelemen  I changed to `std::array`, and replied to others. Could you take a look?

matekelemen (approved):

PR #11476, opened 2023-08-14
Geo/umat parameters python

EleniSmyrniou:
Added UMAT_PARAMETERS as a python variable. This is almost always essential when setting up random field so it is a nice minor improvement


aronnoordam (approved):

mcgicjn2:
Why is this "almost always essential" ? It's a random field anything can be random in a material property

aronnoordam (changes requested):
placeholder, because I don't remember why umat parameters were required to be exposed to python :D

mcgicjn2:
Yes the why is important for instance if a c/phi reduction is being created on the python side, we are already developing that officially

EleniSmyrniou:
This goes a bit further than the c/phi reduction and the Mohr-Coulomb model. The problem is if you want to create a random field based on the parameters of a UMAT mode you need to be able to change the UMAT_PARAMETERS.

EleniSmyrniou:
> placeholder, because I don't remember why umat parameters were required to be exposed to python :D

They need to be exposed because to vary the parameters when using a UMAT model, the parameters provided so far in KratosGeo do not suffice.

aronnoordam:
I think that the set_parameter_field needs to be updated by also accepting vector variables. Currently only double variables can be altered via set_parameter_field

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -28,132 +28,240 @@ namespace Kratos
...
+    void SetParameterFieldProcess::ExecuteInitialize()
+    {
+        if (mrModelPart.GetProcessInfo()[IS_RESTARTED]) {
+            return;
+        }
+
+        KRATOS_TRY
 
-}  // namespace Kratos.
+            const bool umat_check = mParameters["variable_name"].GetString() == "UMAT_PARAMETERS";
```
aronnoordam: as discussed, it might be better to check if the variable is a vector rather than the check if variable == "UMAT_PARAMETERS" but i dont know how, @avdg81 do you have an idea?

aronnoordam: also for now i would change the name of umat_check to: variable_is_umat

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -28,132 +28,240 @@ namespace Kratos
...
+        }
+        else {
+            const auto& r_var = KratosComponents<Variable<double>>::Get(mParameters["variable_name"].GetString());
+            SetParameterFieldForVariableType(r_var);
+        }
+
+        KRATOS_CATCH("")
+    }
+
+    void SetParameterFieldProcess::SetParameterFieldForVariableType(const Variable<Vector>& r_var)
```
aronnoordam: also as discussed. The code is duplicated, with the difference that r_var has a different type. @avdg81 , can this be shortened?

applications/GeoMechanicsApplication/python_scripts/set_parameter_field_process.py
```diff
@@ -36,15 +36,41 @@ def __init__(self, Model, settings ):
...
         self.params = KratosMultiphysics.Parameters("{}")
-        self.params.AddValue("model_part_name",settings["model_part_name"])
+        self.params.AddValue("model_part_name", settings["model_part_name"])
         self.params.AddValue("variable_name", settings["variable_name"])
         self.params.AddValue("func_type", settings["func_type"])
         self.params.AddValue("function", settings["function"])
         self.params.AddValue("dataset", settings["dataset"])
+        is_variable_vector_type = isinstance(self.GetVariableBasedOnString(), KratosMultiphysics.VectorVariable)
+        if "input" in settings["func_type"].GetString() and is_variable_vector_type:
+            self.params.AddValue("vector_index", settings["vector_index"])
```
aronnoordam: vector_variable_indices, describes the purpose more I think. Since multiple indices can be given

applications/GeoMechanicsApplication/tests/test_parameter_field.py
```diff
@@ -12,6 +12,98 @@ class KratosGeoMechanicsParameterFieldTests(KratosUnittest.TestCase):
...
+        elements = simulation._list_of_output_processes[0].model_part.Elements
+        center_coords = [element.GetGeometry().Center() for element in elements]
+
+        # Get test results
+        results = test_helper.get_on_integration_points(simulation, KratosGeo.UMAT_PARAMETERS)
+
+        # assert
+        for center_coord, res in zip(center_coords, results):
+            expected_res = 20000 * center_coord[0] + 30000 * center_coord[1]
+            self.assertAlmostEqual(expected_res, res[0][0])
```
aronnoordam, line 45: can you also check if the non changed values in the umat parameters stay non-changed

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -28,132 +28,240 @@ namespace Kratos
...
         {
             "help"              : "This process sets a parameter field on a model part, where each element can have different material properties.",
             "model_part_name"   : "please_specify_model_part_name",
             "variable_name"     : "CUSTOM",
             "func_type"         : "input",               
             "function"          : "0",
             "dataset"           : "dummy",
-            "dataset_file_name" : "dummy"
+            "dataset_file_name" : "dummy",
+            "vector_index"      : []
```
aronnoordam: i think vector_variable_indices is describing the functionality in a better way

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -28,132 +28,240 @@ namespace Kratos
...
+        for (Element& r_element : mrModelPart.Elements())
+        {
+            SetValueAtElement(r_element, rVar, r_data_vector[i]);
+            ++i;
+        }
+    }
+
+    void SetParameterFieldProcess::SetParameterFieldUsingParametersClass(const Variable<Vector>& rVar, Parameters& rParameters)
+    {
+        const Matrix& r_data_vector = rParameters["values"].GetMatrix();
```
aronnoordam: i think you should also use vector_index here, and only set the values at the variable at the index of the vector

applications/GeoMechanicsApplication/tests/test_parameter_field/parameter_field_input_umat_parameters/ProjectParameters.json
```diff
@@ -0,0 +1,147 @@
...
+                        },
+                        "file_label":          "step",
+                        "output_control_type": "step",
+                        "output_interval":     1,
+                        "body_output":         true,
+                        "node_output":         false,
+                        "skin_output":         false,
+                        "plane_output":        [],
+                        "nodal_results":       ["DISPLACEMENT","TOTAL_DISPLACEMENT","WATER_PRESSURE","VOLUME_ACCELERATION"],
+                        "gauss_point_results": ["YOUNG_MODULUS","GREEN_LAGRANGE_STRAIN_TENSOR","ENGINEERING_STRAIN_TENSOR","CAUCHY_STRESS_TENSOR","TOTAL_STRESS_TENSOR","VON_MISES_STRESS","FLUID_FLUX_VECTOR","HYDRAULIC_HEAD", "UMAT_PARAMETERS"]
```
aronnoordam: can you remove the unused nodal and gauss point results? In order to keep the test small. So i think only "UMAT_PARAMETERS" has to stay

applications/GeoMechanicsApplication/tests/test_parameter_field/parameter_field_json_umat_parameters/ProjectParameters.json
```diff
@@ -0,0 +1,147 @@
...
+                            "MultiFileFlag":         "SingleFile"
+                        },
+                        "file_label":          "step",
+                        "output_control_type": "step",
+                        "output_interval":     1,
+                        "body_output":         true,
+                        "node_output":         false,
+                        "skin_output":         false,
+                        "plane_output":        [],
+                        "nodal_results":       ["DISPLACEMENT","TOTAL_DISPLACEMENT","WATER_PRESSURE","VOLUME_ACCELERATION"],
```
aronnoordam: please remove unused variables

applications/GeoMechanicsApplication/tests/test_parameter_field/parameter_field_python_umat_parameters/ProjectParameters.json
```diff
@@ -0,0 +1,146 @@
...
+                            "MultiFileFlag":         "SingleFile"
+                        },
+                        "file_label":          "step",
+                        "output_control_type": "step",
+                        "output_interval":     1,
+                        "body_output":         true,
+                        "node_output":         false,
+                        "skin_output":         false,
+                        "plane_output":        [],
+                        "nodal_results":       ["DISPLACEMENT","TOTAL_DISPLACEMENT","WATER_PRESSURE","VOLUME_ACCELERATION"],
```
aronnoordam: please remove unused variables

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
aronnoordam: Else I think, the back and fort communication between the external function and thememory usage is unnecessarily inefficient

aronnoordam (commented):
nice, few comments and questions

mcgicjn2 (changes requested):
Can you make sure that this change doesn't limit the functionality to only UMAT parameters. This should be optional. Checking that UMAT is present means that we limit this change to only those materials, and thus we cannot for instance do this with any of the kratos coded materials parameters (e.g. linear elastic). This should be generic.

EleniSmyrniou:
> Can you make sure that this change doesn't limit the functionality to only UMAT parameters. This should be optional. Checking that UMAT is present means that we limit this change to only those materials, and thus we cannot for instance do this with any of the kratos coded materials parameters (e.g. linear elastic). This should be generic.

This is arleady the case. All parameters defined in Kratos coded materials ( for example YOUNG_MODULUS) can be used as a input to a field. This issue is merely an extension to include the UMAT_PARAMETERS as this variable is a Vector and was not covered by the current code. Functionality of how Kratos handles the UMATS or internal umats was not changed

aronnoordam:
> > Can you make sure that this change doesn't limit the functionality to only UMAT parameters. This should be optional. Checking that UMAT is present means that we limit this change to only those materials, and thus we cannot for instance do this with any of the kratos coded materials parameters (e.g. linear elastic). This should be generic.
> 
> This is arleady the case. All parameters defined in Kratos coded materials ( for example YOUNG_MODULUS) can be used as a input to a field. This issue is merely an extension to include the UMAT_PARAMETERS as this variable is a Vector and was not covered by the current code. Functionality of how Kratos handles the UMATS or internal umats was not changed

Ideally it should work for all vector variables, if there is a way to check if variable is of type vector rather than to check for UMAT, that would be ideal

EleniSmyrniou:
> > > Can you make sure that this change doesn't limit the functionality to only UMAT parameters. This should be optional. Checking that UMAT is present means that we limit this change to only those materials, and thus we cannot for instance do this with any of the kratos coded materials parameters (e.g. linear elastic). This should be generic.
> > 
> > 
> > This is arleady the case. All parameters defined in Kratos coded materials ( for example YOUNG_MODULUS) can be used as a input to a field. This issue is merely an extension to include the UMAT_PARAMETERS as this variable is a Vector and was not covered by the current code. Functionality of how Kratos handles the UMATS or internal umats was not changed
> 
> Ideally it should work for all vector variables, if there is a way to check if variable is of type vector rather than to check for UMAT, that would be ideal. My first comment of today refers to this

Indeed if anyone has an idea on how to check from a string the type of variable in Kratos, I will be happy to include it

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
EleniSmyrniou: Corrected by Anne's changes

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
```diff
@@ -28,132 +28,132 @@ namespace Kratos
...
+    void SetParameterFieldProcess::SetParameterFieldUsingParametersClass(const Variable<Vector>& rVar,
+        const Parameters& rParameters)
+    {
+        const Matrix& r_data_matrix = rParameters["values"].GetMatrix();
+
+        KRATOS_ERROR_IF_NOT(r_data_matrix.size1() == mrModelPart.Elements().size()) << "The parameter field "
+            "does not have the same size as the amount of elements within the model part!" << std::endl;
+
+        // set new data on the elements
+        const int vector_size = r_data_matrix.size2();
```
avdg81: Please note that SonarQube reports a narrowing conversion here (i.e. an implicit conversion through which we loose precision). I would suggest you to use the `size_type` of the matrix class rather than narrowing it to an `int`.

avdg81 (changes requested):

applications/GeoMechanicsApplication/custom_processes/set_parameter_field_process.cpp
EleniSmyrniou: Ok!

EleniSmyrniou: Changing the type of the vector_size to IndexType which is the size_type of the matrix class make some ubuntu builders fail. So I will revert it back to the int since then the tests pass. Any other suggestions are welcome

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_element.cpp
```diff
@@ -722,6 +722,7 @@ void UPwSmallStrainElement<TDim,TNumNodes>::
     //Defining necessary variables
     const GeometryType& rGeom = this->GetGeometry();
     const IndexType NumGPoints = rGeom.IntegrationPointsNumber( mThisIntegrationMethod );
+    const PropertiesType& rProp = this->GetProperties();
```
rfaasse, line 725: The 'this->' is not really necessary here, so unless it is important to re-emphasize it's coming from within the class, I would propose to remove it (it is also personal preference to keep the code a bit cleaner, but I see it's done a lot in this class, so it's up to you if you want to change it or not)

rfaasse (approved):
The code looks very clean and clear (as a new joiner for the Kratos GeoMechanics team, I learned quite a bit about the UMAT process during this review)! 

I only have one minor suggestion, but that isn't blocking (it's close to a general personal preference, so feel free to implement or ignore it).

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_element.cpp
EleniSmyrniou: What would the like look like then?
`const PropertiesType& rProp = GetProperties();` ?

EleniSmyrniou: Indeed maybe we leave this as is because it happens a lot in this class?

rfaasse: That's fine with me

avdg81 (approved):

aronnoordam (approved):

mcgicjn2 (approved):

PR #11478, opened 2023-08-14
[CLApp] Including int for printing process

atfard:
In this PR, the capability of printing int-type variable is added to the print_info_in_file_process

### DIFF
--- a/kratos/python_scripts/print_info_in_file_process.py
+++ b/kratos/python_scripts/print_info_in_file_process.py
@@ -94,7 +94,7 @@ class PrintInfoInFileProcess(KratosMultiphysics.OutputProcess):
         else:
             for elem in self.model_part.Elements:
                 array_values = self.GetValueToPrint(elem)[self.integration_point]
-                if not isinstance(array_values, float): # Can be an scalar entity...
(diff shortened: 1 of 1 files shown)


loumalouomega (commented):
Looks okay,@roigcarlo?

AlejandroCornejo:
@atfard discovered that I did not expect to print int values... He need it for printing number of cycles.

kratos/python_scripts/print_info_in_file_process.py
```diff
@@ -94,7 +94,7 @@ def PrintOutput(self):
         else:
             for elem in self.model_part.Elements:
                 array_values = self.GetValueToPrint(elem)[self.integration_point]
-                if not isinstance(array_values, float): # Can be an scalar entity...
+                if not isinstance(array_values, float) and not isinstance(array_values, int): # Can be an scalar entity...
```
philbucher:
```suggestion
                if not isinstance(array_values, (float, int)): # Can be an scalar entity...
```

atfard: Changes applied

philbucher (approved):

PR #11479, opened 2023-08-14
[GeoMechanicsApplication] Fixed calls to `static` member functions

avdg81:
**📝 Description**
These changes have been overlooked in a previous commit and where now reported as code smells by SonarQube.

**🆕 Changelog**
No longer instantiate the utilities class in order to call its member functions (since those had been made `static`).


WPK4FEM (approved):
Clearer to me.

PR #11481, opened 2023-08-15
[HDF5] Update - 2 - Container component IO

sunethwarna:
**📝 Description**
This PR replaces all the `#pragma omp` calls with the `ParallelUtilities`, and unifies the behaviour between all the component IOs.

Removes the dependency on `RegisteredComponentLookup` by replacing its uses with more clear fold expressions.

** TODO **

- [x] Add zero data rank tests for component IO.

**🆕 Changelog**
- Update `ComponentDataIO`.
- Updated tests.


applications/HDF5Application/custom_utilities/container_io_utils.h
```diff
@@ -0,0 +1,477 @@
...
+{
+public:
+    ///@name Type definitions
+    ///@{
+
+    template<class TDataType>
+    using TLSType = char;
+
+    template<class TDataType>
+    using ComponentDataType = char;
```
matekelemen: Too much `Data` in type names

```suggestion
    using ComponentType = char;
```

applications/HDF5Application/custom_utilities/container_io_utils.h
```diff
@@ -0,0 +1,477 @@
...
+
+    ///@}
+    ///@name Public operations
+    ///@{
+
+    template<class TEntityType>
+    inline char GetValue(
+        const TEntityType& rEntity,
+        const Flags& rFlag,
+        char& rTLS) const
```
matekelemen, line 56: I'm not sure what `TLSType` is for here, but I guess this is one of them

```suggestion
        TLSType<TEntityType>& rTLS) const
```

applications/HDF5Application/custom_utilities/container_io_utils.h
```diff
@@ -0,0 +1,477 @@
...
+    ///@}
+
+private:
+    ///@name Private member variables
+    ///@{
+
+    const double mAlphaBossak;
+
+    ///@}
+};
```
matekelemen: Why does Bossak need a whole separate infrastructure?

applications/HDF5Application/custom_utilities/container_io_utils.h
```diff
@@ -0,0 +1,477 @@
...
+        const Variable<TDataType>& rVariable,
+        const TDataType& rValue) const
+    {
+        KRATOS_ERROR << "VertexValueIO does not support setting values from vertices";
+    }
+
+    ///@}
+};
+
+class GaussPointValueIO
```
matekelemen:
```suggestion
class GaussPointIO
```

applications/HDF5Application/custom_utilities/container_io_utils.h
```diff
@@ -0,0 +1,477 @@
...
+    ///@{
+
+    template<class TEntityType, class TDataType>
+    inline const ComponentDataType<TDataType>& GetValue(
+        const TEntityType& rEntity,
+        const Variable<TDataType>& rVariable,
+        TLSType<TDataType>& rTLS) const
+    {
+        // Since the array_1d<double, 4> and array_1d<double 9> interfaces are not present in
+        // elements and conditions.
```
matekelemen: should we define them then?

applications/HDF5Application/custom_utilities/container_io_utils.h
```diff
@@ -0,0 +1,477 @@
...
+    ///@name Private member variables
+    ///@{
+
+    const ProcessInfo& mrProcessInfo;
+
+    ///@}
+};
+
+template<class TContainerIOType>
+std::string GetContainerIOType()
```
matekelemen:
```suggestion
std::string GetContainerIOName()
```

applications/HDF5Application/custom_utilities/container_io_utils.h
```diff
@@ -0,0 +1,477 @@
...
+        return "INTERPOLATED";
+    } else if constexpr(std::is_same_v<TContainerIOType, GaussPointValueIO>) {
+        return "GAUSS_POINT";
+    } else {
+        static_assert(!std::is_same_v<TContainerIOType, TContainerIOType>, "Unsupported container io type.");
+    }
+}
+
+template<class TContainerType>
+std::string GetContainerType()
```
matekelemen:
```suggestion
std::string GetContainerName()
```

applications/HDF5Application/custom_utilities/container_io_utils.h
```diff
@@ -0,0 +1,477 @@
...
+        return rModelPart.GetCommunicator().LocalMesh().Conditions();
+    } else if constexpr(std::is_same_v<TContainerType, ModelPart::ElementsContainerType>) {
+        return rModelPart.GetCommunicator().LocalMesh().Elements();
+    } else {
+        static_assert(!std::is_same_v<TContainerType, TContainerType>, "Unsupported container type.");
+    }
+}
+
+template<class TContainerType, class TComponentType, class TContainerDataIO, class TDataType>
+void CopyToContiguousDataArray(
```
matekelemen:
```suggestion
void CopyToContiguousArray(
```

applications/HDF5Application/custom_utilities/container_io_utils.h
```diff
@@ -0,0 +1,477 @@
...
+    const TContainerType& rContainer,
+    const TComponentType& rComponent,
+    const TContainerDataIO& rContainerDataIO,
+    TDataType* pBegin)
```
matekelemen: can you please also pass the array's size as well? It'll make life much easier if we could validate that the array is big enough to fit all entries, instead of segfaulting.

```suggestion
    const TContainerType& rContainer,
    const TComponentType& rComponent,
    const TContainerDataIO& rContainerDataIO,
    TDataType* pBegin,
    const std::size_t Size)
```

If this pattern occurs frequently, we could also add a `Span` (similar to [`std::span`](https://en.cppreference.com/w/cpp/container/span)) class that models a view over a contiguous array. It's a handy, safe and lightweight thing to pass around compared to just juggling with raw pointers.

applications/HDF5Application/custom_utilities/container_io_utils.h
```diff
@@ -0,0 +1,477 @@
...
+{
+    KRATOS_TRY
+
+    using component_data_type = typename Internals::template ComponentTraits<TComponentType>::ValueType;
```
matekelemen:
```suggestion
    using component_type = typename Internals::template ComponentTraits<TComponentType>::ValueType;
```

applications/HDF5Application/custom_utilities/container_io_utils.h
```diff
@@ -0,0 +1,477 @@
...
+
+    using value_type = typename TContainerDataIO::template ComponentDataType<component_data_type>;
+
+    using value_type_traits = DataTypeTraits<value_type>;
```
matekelemen, line 406: `value` or `data`? The naming should be consistent. I'm voting for `value` in this case because it distinguishes between the `value_type` stored in an array and `component_type` which might be identical to `value_type` or compose it, while `data` doesn't say much.

applications/HDF5Application/custom_utilities/container_io_utils.h
```diff
@@ -0,0 +1,477 @@
...
+    const auto& initial_value = rContainerDataIO.GetValue(rContainer.front(), rComponent, tls);
+
+    // get the stride from the first element to support dynamic types.
+    const auto stride = value_type_traits::Size(initial_value);
+
+    IndexPartition<unsigned int>(rContainer.size()).for_each(tls, [&rContainer, &rComponent, &rContainerDataIO, pBegin, stride](const auto Index, auto& rTLS) {
+        const auto& value = rContainerDataIO.GetValue(*(rContainer.begin() + Index), rComponent, rTLS);
+        TDataType const* p_value_begin = value_type_traits::GetContiguousData(value);
+        auto p_array_start = pBegin + Index * stride;
+        std::copy(p_value_begin, p_value_begin + stride, p_array_start);
```
matekelemen:
```suggestion
        auto p_subrange_begin = pBegin + Index * stride;
        std::copy(p_value_begin, p_value_begin + stride, p_subrange_begin);
```

applications/HDF5Application/custom_io/hdf5_container_component_io.h
```diff
@@ -40,13 +40,15 @@ namespace HDF5
 ///@{
 
 /// A class for IO of element data in HDF5.
-template <typename TContainerType, typename TContainerItemType, typename... TComponents>
+template <class TContainerType, class TContainerDataIO, class... TComponents>
```
matekelemen:
```suggestion
template <class TContainerType, class TContainerIO, class... TComponents>
```

applications/HDF5Application/custom_io/hdf5_container_component_io.h
```diff
@@ -55,43 +57,72 @@ class KRATOS_API(HDF5_APPLICATION) ContainerComponentIO
...
     ///@name Private Operations
     ///@{
 
-    template <typename... Targs>
-    void WriteRegisteredComponent(const std::string& rComponentName, Targs&... args);
-
-    template <typename... Targs>
-    void ReadRegisteredComponent(const std::string& rComponentName, Targs&... args);    
+    template<class TComponentType>
+    bool WriteComponentData(
```
matekelemen:
```suggestion
    bool WriteComponents(
```

applications/HDF5Application/custom_io/hdf5_container_component_io.h
```diff
@@ -55,43 +57,72 @@ class KRATOS_API(HDF5_APPLICATION) ContainerComponentIO
...
+        const std::string& rComponentName,
+        const TContainerDataIO& rContainerDataIO,
+        const TContainerType& rLocalContainer,
+        Parameters Attributes,
+        WriteInfo& rInfo);
+
+    template<class TComponentType>
+    bool ReadComponentData(
```
matekelemen:
```suggestion
    bool ReadComponents(
```

applications/HDF5Application/custom_io/hdf5_container_component_io.h
```diff
@@ -40,13 +40,15 @@ namespace HDF5
 ///@{
 
 /// A class for IO of element data in HDF5.
-template <typename TContainerType, typename TContainerItemType, typename... TComponents>
+template <class TContainerType, class TContainerDataIO, class... TComponents>
 class KRATOS_API(HDF5_APPLICATION) ContainerComponentIO
```
matekelemen, line 44: This is probably the most important class in the entire application, and yet it's not documented. We need detailed docs here about what this class does.

```suggestion
class KRATOS_API(HDF5_APPLICATION) ContainerComponentIO
```

applications/HDF5Application/custom_io/hdf5_container_component_io.h
```diff
@@ -55,43 +57,72 @@ class KRATOS_API(HDF5_APPLICATION) ContainerComponentIO
     ///@{
 
     /// Constructor.
```
matekelemen, line 59: ... and I'm not talking about these kinds of docs. Please explain each argument, especially what's expected in `Settings`.

applications/HDF5Application/custom_io/hdf5_container_component_io.h
```diff
@@ -55,43 +57,72 @@ class KRATOS_API(HDF5_APPLICATION) ContainerComponentIO
...
+        const ModelPart& rModelPart,
+        const TContainerDataIO& rContainerDataIO,
+        const Parameters Attributes);
 
-protected:
-    ///@name Protected Member Variables
-    ///@{
+    std::map<std::string, Parameters> Read(
+        ModelPart& rModelPart,
+        const TContainerDataIO& rContainerDataIO);
```
matekelemen: missing docs about what this actually returns

applications/HDF5Application/custom_io/hdf5_container_component_io.h
```diff
@@ -55,43 +57,72 @@ class KRATOS_API(HDF5_APPLICATION) ContainerComponentIO
...
 
     ///@}
     ///@name Operations
     ///@{
 
-    ///@}
+    void Write(
+        const ModelPart& rModelPart,
+        const TContainerDataIO& rContainerDataIO,
+        const Parameters Attributes);
```
matekelemen: missing docs about what's expected in `Attributes`

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
```diff
@@ -1,384 +1,91 @@
...
-template <typename TComponent>
-class ReadElementComponent
-    : public ContainerItemComponentIO<ElementType>::ReadComponentFunctor<TComponent, ComponentElementSynchronizingFunctor>
-{
-};
-
-template <typename TComponent>
-class WriteConditionComponent
-    : public ContainerItemComponentIO<ConditionType>::WriteComponentFunctor<TComponent, void>
+namespace NewContainerComponentIOUtilities
```
matekelemen, line 35:
```suggestion
// TODO: rename this namespace once HDF5 refactoring is complete
namespace NewContainerComponentIOUtilities
```

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
```diff
@@ -1,384 +1,91 @@
...
     : mpFile(pFile)
 {
     KRATOS_TRY;
 
     Parameters default_params(R"(
         {
-            "prefix": "",
+            "prefix"           : "",
             "list_of_variables": []
         })");
```
matekelemen, line 106: this should be in a `GetDefaultParameters` static member function

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
```diff
@@ -1,384 +1,91 @@
...
 
     Settings.AddMissingParameters(default_params);
 
-    mComponentPath = Settings["prefix"].GetString() + rComponentPath;
+    mComponentPrefix = Settings["prefix"].GetString() + rLegacySuffix;
     mComponentNames = Settings["list_of_variables"].GetStringArray();
 
+    KRATOS_ERROR_IF(mComponentPrefix == "" || mComponentPrefix == "/")
+        << "The prefix should not be blank or \"/\" [ prefix = " << mComponentPrefix << " ].\n";
```
matekelemen:
```suggestion
        << "The prefix must not be empty or \"/\" [ prefix = " << mComponentPrefix << " ].\n";
```

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
```diff
@@ -389,477 +96,331 @@ ContainerComponentIO<TContainerType, TContainerItemType, TComponents...>::Contai
...
+    KRATOS_TRY
+
+    if constexpr(
+        std::is_same_v<TContainerType, NodesContainerType> ||
+        std::is_same_v<TContainerType, ConditionsContainerType> ||
+        std::is_same_v<TContainerType, ElementsContainerType>) {
+        return Read(Internals::GetLocalContainer<TContainerType>(rModelPart), rContainerDataIO, rModelPart.GetCommunicator());
+    } else {
+        KRATOS_ERROR << "Unsupported container type.";
+        return std::map<std::string, Parameters>{};
```
matekelemen:
```suggestion
        return {};
```

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
```diff
@@ -389,477 +96,331 @@ ContainerComponentIO<TContainerType, TContainerItemType, TComponents...>::Contai
...
-                                                                             Targs&... args)
-{
-    RegisteredComponentLookup<Variable<array_1d<double, 3>>, Variable<double>, Variable<int>,
-                              Variable<Vector<double>>, Variable<Matrix<double>>>(rComponentName)
-        .Execute<ReadConditionComponent>(args...);
-}
+    // Write each variable.
+    for (const auto& r_component_name : mComponentNames) {
+        const bool is_component_written = (... || (WriteComponentData<TComponents>(
+                        r_component_name, rContainerDataIO, rLocalContainer, Attributes.Clone(), info)));
```
matekelemen: not sure about this one, but shouldn't it be `&&`?


```suggestion
        const bool is_component_written = (... && (WriteComponentData<TComponents>(
                        r_component_name, rContainerDataIO, rLocalContainer, Attributes.Clone(), info)));
```

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
```diff
@@ -389,477 +96,331 @@ ContainerComponentIO<TContainerType, TContainerItemType, TComponents...>::Contai
...
-template <>
-template <class ...TArguments>
-void ContainerComponentIO<VertexContainerType, Vertex, Flags>::WriteRegisteredComponent(
-    const std::string& rComponentName, TArguments&... rArguments)
-{
-    RegisteredComponentLookup<Flags>(rComponentName).Execute<WriteVertexComponent>(rArguments...);
-}
+    if (mComponentNames.size() == 0) {
+        return attributes;
+    }
```
matekelemen: my MPI spider senses are tingling. Can a situation ever occur when `mComponentNames` is not consistent across ranks?

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
```diff
@@ -389,477 +96,331 @@ ContainerComponentIO<TContainerType, TContainerItemType, TComponents...>::Contai
...
         }
-    }
 
-    std::vector<TContainerItemType*> local_items;
-    GetContainerItemReferences(local_items, rContainer);
-    std::size_t start_index, block_size;
-    std::tie(start_index, block_size) = StartIndexAndBlockSize(*mpFile, mComponentPath);
+        // add the shape to attributes.
+        KRATOS_ERROR_IF(Attributes.Has("__data_shape"))
+            << "The reserved keyword \"__data_shape\" is found. Please remove it from attributes.";
```
matekelemen:
```suggestion
            << "reserved keyword \"__data_shape\" is found. Please remove it from the attributes.";
```

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
```diff
@@ -389,477 +96,331 @@ ContainerComponentIO<TContainerType, TContainerItemType, TComponents...>::Contai
...
-                       Vector<int> const& rData,
-                       std::vector<TContainerItemType*>& rContainerItems)
-{
-    KRATOS_TRY;
+        if (attributes.Has("__data_dimension")) attributes.RemoveValue("__data_dimension");
+        if (attributes.Has("__data_shape")) attributes.RemoveValue("__data_shape");
+        if (attributes.Has("__container_type")) attributes.RemoveValue("__container_type");
+        if (attributes.Has("__data_name")) attributes.RemoveValue("__data_name");
+        if (attributes.Has("__data_location")) attributes.RemoveValue("__data_location");
+        if (attributes.Has("__mesh_location")) attributes.RemoveValue("__mesh_location");
```
matekelemen: these should be in some central location so if we ever need to update them, we can do so more easily (and they'd also be a bit easier to look through), Also, `ContainerComponentIO`'s docs should mention these reserved attributes.

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -795,7 +795,9 @@ bool File::HasDataType(const std::string& rPath) const
     if constexpr(std::is_same_v<TDataType, int>) {
         return type == H5T_INTEGER;
     } else if constexpr(std::is_same_v<TDataType, double>) {
-        return type == H5T_FLOAT;
+        return (type == H5T_FLOAT);
```
matekelemen:
```suggestion
        return type == H5T_FLOAT;
```

applications/HDF5Application/custom_io/hdf5_file.cpp
```diff
@@ -795,7 +795,9 @@ bool File::HasDataType(const std::string& rPath) const
...
+    } else if constexpr(std::is_same_v<TDataType, char>) {
+        return (type == H5T_INTEGER);
```
matekelemen:
```suggestion
        return type == H5T_INTEGER;
```

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
matekelemen: nevermind I think I understand now ...

this loops over the types and checks whether the name is registered in the corresponding instantiation of `KratosComponents`. Most return `false` and only one should return `true` (for the type of the requested variable). Is that correct?

applications/HDF5Application/custom_io/hdf5_vertex_container_io.cpp
```diff
@@ -10,189 +10,122 @@
...
-
-void VertexContainerCoordinateIO::WriteWithoutIDs(const Detail::VertexContainerType& rVertices)
+void VertexContainerCoordinateIO::WriteWithoutIDs(
+    const Detail::VertexContainerType& rVertices,
+    Parameters Attributes)
 {
     KRATOS_TRY
 
-    const std::string path = mComponentPath + "/POSITION";
+    const std::string path = mPathPrefix + "/POSITION";
```
philbucher, line 94: Perhaps filesystem::path would be better here

applications/HDF5Application/custom_io/hdf5_vertex_container_io.cpp
```diff
@@ -10,189 +10,122 @@
...
-
-    KRATOS_CATCH("");
-}
-
-
-VertexContainerCoordinateIO::VertexContainerCoordinateIO(Parameters parameters,
-                                                         File::Pointer pFile)
-    : VertexContainerIO(VertexContainerCoordinateIO::FormatParameters(parameters), pFile)
+void VertexContainerCoordinateIO::Write(
+    const Detail::VertexContainerType& rVertices,
```
philbucher, line 47: Why vertex? Usually we use node

applications/HDF5Application/custom_python/add_custom_io_to_python.cpp
```diff
@@ -29,45 +28,192 @@
...
+
+    using data_io = typename TContainerComponentIOWrapper::ContainerIOType;
+
+    using container_type = typename TContainerComponentIOWrapper::ContainerType;
 
+    using container_data_io_type = typename TContainerComponentIOWrapper::ContainerDataIOType;
+
+    std::string container_name, arg_name;
+    if constexpr(std::is_same_v<container_type, ModelPart::NodesContainerType>) {
+        container_name = "Nodal";
```
philbucher, line 94: Nodal is inconsistent with Element and Condition

applications/HDF5Application/custom_utilities/container_io_utils.h
```diff
@@ -0,0 +1,477 @@
...
+std::string GetContainerType()
+{
+    if constexpr(std::is_same_v<TContainerType, ModelPart::NodesContainerType>) {
+        return "NODES";
+    } else if constexpr(std::is_same_v<TContainerType, ModelPart::ConditionsContainerType>) {
+        return "CONDITIONS";
+    } else if constexpr(std::is_same_v<TContainerType, ModelPart::ElementsContainerType>) {
+        return "ELEMENTS";
+    } else if constexpr(std::is_same_v<TContainerType, Detail::VertexContainerType>) {
+        return "VERTICES";
```
philbucher, line 358: Nodes would be more consistent

applications/HDF5Application/custom_utilities/container_io_utils.h
```diff
@@ -0,0 +1,477 @@
...
+        if (rVariable == ACCELERATION) {
+            return (1.0 - mAlphaBossak) * rNode.FastGetSolutionStepValue(rVariable, 0) + mAlphaBossak * rNode.FastGetSolutionStepValue(rVariable, 1);
+        } else {
+            return rNode.FastGetSolutionStepValue(rVariable);
+        }
+    }
+
+    template<class TDataType>
+    inline void SetValue(
+        ModelPart::NodeType& rNode,
```
philbucher: Just Node&

applications/HDF5Application/custom_io/hdf5_vertex_container_io.cpp
matekelemen: that wouldn't work on windows because it would translate to `...\\POSITION`, but this is a path within HDF5

applications/HDF5Application/custom_io/hdf5_vertex_container_io.cpp
matekelemen: I'm to blame for this one. I came up with [`Vertex`](https://github.com/KratosMultiphysics/Kratos/blob/358979aaeb2b2d82bf7dd63992f48e984f6a04fb/applications/HDF5Application/custom_utilities/vertex.h#L31) in the context of point output to distinguish between `Node`s and general points within the mesh.

A `Vertex` is basically a point inside an element that interpolates nodal values.

applications/HDF5Application/custom_io/hdf5_vertex_container_io.cpp
philbucher: I think you can also get the path with /

matekelemen:
@sunethwarna btw how's the plan on adding MPI tests with empty model parts doing? These cases have always been the bane of the app.

applications/HDF5Application/custom_utilities/container_io_utils.h
sunethwarna: done :)

applications/HDF5Application/custom_utilities/container_io_utils.h
sunethwarna: Since it is always char, I don't see a difference in using `char` over `TLSType<TEntityType>`, apart from the fact use of templates makes my head go crazy someties ;)

applications/HDF5Application/custom_utilities/container_io_utils.h
sunethwarna: The way transient adjoints were done originally by using this modified output scheme of adjoints because, at the early days of transient adjoints, there were so little room for transient data passing. I would keep this for time being, until the old `VMSAdjointElement` is retired. [It is the only place this is required for fluid transient adjoint computations]

What this does is, it stores the adjoint specific `ACCELERATION` value as `ACCELERATION` in the HDF5 otuput, so when the old `VMSAdjointElement` reads `ACCELERATION`, it gets the adjoint sepcific `ACCELERATION` instead of the actual `ACCELERATION`. I understand this is workaround, we had in early days but for now, lets keep it.

applications/HDF5Application/custom_utilities/container_io_utils.h
sunethwarna: done :)

applications/HDF5Application/custom_utilities/container_io_utils.h
sunethwarna: I have done it in PR #11743  :).... now we don't need them :)

applications/HDF5Application/custom_utilities/container_io_utils.h
sunethwarna: done :)

applications/HDF5Application/custom_utilities/container_io_utils.h
sunethwarna: done :)

applications/HDF5Application/custom_utilities/container_io_utils.h
sunethwarna: done :)

applications/HDF5Application/custom_utilities/container_io_utils.h
sunethwarna: done :)

applications/HDF5Application/custom_utilities/container_io_utils.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_container_component_io.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_container_component_io.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
sunethwarna: You are correct 👍

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
sunethwarna: Yes and no, It is nice to have it as a method, but then if we rely on the static methods which gives the defaults, then it will be nightmare to find all the defaults. So I would rather have the defaults like this, and in the last user exposed place (which is redundant), so user can see at one glance the whole defaults.

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
sunethwarna: Hmm.... I would say, these `mComponentNames` should be consistent among ranks (I remember your use case), but woudn't it be the responsibility of the person who gives this as input to make them consistent across ranks before providing it?

Otherwise, aren't we mixing responsibilities if we integrate communication for `mComponentNames` in this?

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_file.cpp
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_vertex_container_io.cpp
sunethwarna: I am not sure, I went through the docs, but I am not sure whether they gurantee it.

applications/HDF5Application/custom_python/add_custom_io_to_python.cpp
sunethwarna: Yep, but I need it to have backward compatibility with the legacy HDF5 writing mechanism. :/

applications/HDF5Application/custom_utilities/container_io_utils.h
sunethwarna: As @matekelemen has told, these are not exactly mesh nodes, these are some random points in the mesh used for line outputs.

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
sunethwarna: Ok, it is passed via `Kratos::Parameters`, so we don't have to make the ctor of `ContainerComponentIO` private since, user is suppose to pass the same `Kratos::Parameters` from all ranks.

applications/HDF5Application/custom_io/hdf5_container_component_io.cpp
sunethwarna: done :)

applications/HDF5Application/custom_utilities/container_io_utils.h
matekelemen: fine, just please add a `@todo` so we can get rid of it when we drop support for `VMSAdjointElement`

applications/HDF5Application/custom_utilities/container_io_utils.h
matekelemen: this one slipped under your radar

matekelemen (commented):
I think this one's mostly ready, you're just missing the empty rank tests and passing the size along with array pointers.

applications/HDF5Application/custom_utilities/container_io_utils.h
sunethwarna: done :)

sunethwarna:
Finally, I have added a zero rank test as well @matekelemen  :)

matekelemen (dismissed):
Thanks!

matekelemen:
I can't reproduce the failures locally.

matekelemen (approved):
image

PR #11486, opened 2023-08-16
[HDF5] Update - 4 - ConnectivityDataIO

sunethwarna:
**📝 Description**
This PR updates to `ConnectivityDataIO` to support recent changes.

**🆕 Changelog**
- Update `ConnectivityDataIO`


applications/HDF5Application/custom_io/hdf5_connectivities_data.h
```diff
@@ -43,92 +44,63 @@ namespace Internals
 /**
  * Acts as the intermediary between the HDF5 file and the Kratos elements and conditions.
  */
+template<class TContainerType>
```
matekelemen:
```suggestion
/**
 * @tparam TContainerType A container of @ref Element "elements" or @ref Condition "conditions".
 * @details Acts as the intermediary between the HDF5 file and the Kratos elements and conditions.
 * @see ElementsContainerType
 * @see ConditionsContainerType
 */
template<class TContainerType>
```

applications/HDF5Application/custom_io/hdf5_connectivities_data.h
```diff
@@ -43,92 +44,63 @@ namespace Internals
...
     /// Pointer definition
     KRATOS_CLASS_POINTER_DEFINITION(ConnectivitiesData);
 
     ///@}
     ///@name Life Cycle
     ///@{
+
+    ConnectivitiesData(
+        const std::string& rPrefix,
+        File::Pointer pFile);
```
matekelemen, line 73:
```suggestion
    /// @brief Construct an IO reading/writing from the specified file at the given prefix.
    /// @param rPrefix Group path within the file to read from or write to.
    /// @param pFile Pointer to the HDF5 file.
    ConnectivitiesData(
        const std::string& rPrefix,
        File::Pointer pFile);
```

applications/HDF5Application/custom_io/hdf5_connectivities_data.cpp
```diff
@@ -1,257 +1,207 @@
...
-        Clear();
-        // Correctly set the name and array sizes. Otherwise, MPI collective write
-        // can fail due to inconsistent data across processes.
-        mName = rName;
-        const Element& r_elem = KratosComponents<ElementType>::Get(mName);
-        mConnectivities.resize(0, r_elem.GetGeometry().size());
-        return;
-    }
+    std::string file_entity_name;
+    mpFile->ReadAttribute(entity_group_path, "Name", file_entity_name);
```
matekelemen, line 70: ~~Seems a bit redundant that we require both the `"Name"` attribute in the file **as well as** `rEntityName` in the function's arguments, but I guess it's necessary for distinguishing between elements and conditions (and later hopefully geometries), right~~

ok I see this is handled in `HDF5ModelPartIO`

applications/HDF5Application/custom_io/hdf5_connectivities_data.cpp
```diff
@@ -1,257 +1,207 @@
...
-        // Fill ids.
-        mIds[i] = it->Id();
-        mPropertiesIds[i] = it->GetProperties().Id();
-        const ElementType::GeometryType& r_geom = it->GetGeometry();
-        // Fill connectivities.
-        for (unsigned k = 0; k < geometry_size; ++k)
-            mConnectivities(i, k) = r_geom[k].Id();
-    }
+    mpFile->ReadDataSet(entity_group_path + "/Ids", entity_ids, start_index, block_size);
+    mpFile->ReadDataSet(entity_group_path + "/PropertiesIds", entitiy_property_ids, start_index, block_size);
```
matekelemen:
```suggestion
    mpFile->ReadDataSet(entity_group_path + "/PropertyIds", entitiy_property_ids, start_index, block_size);
```

how big of a change would this entail? Duplicate plurals in kratos names are driving me mad.

applications/HDF5Application/custom_io/hdf5_connectivities_data.cpp
```diff
@@ -1,257 +1,207 @@
...
+    std::vector<char> char_entity_name(entity_name.begin(), entity_name.end());
+    const auto& global_char_entity_names = r_data_communicator.AllGatherv(char_entity_name);
+    if (entity_name.empty()) {
+        for (const auto& rank_char_entity_name : global_char_entity_names) {
+            if (!rank_char_entity_name.empty()) {
+                entity_name = std::string(rank_char_entity_name.begin(), rank_char_entity_name.end());
+                break;
+            }
+        }
     }
```
matekelemen: what if `entity_name` is still empty after this (meaning that none of the ranks have it)? Can we just skip the rest of this function?

applications/HDF5Application/custom_io/hdf5_connectivities_data.cpp
```diff
@@ -1,257 +1,207 @@
...
+    for (const auto& rank_char_entity_name : global_char_entity_names) {
+        std::string rank_name(rank_char_entity_name.begin(), rank_char_entity_name.end());
+        KRATOS_ERROR_IF(entity_name != rank_name) << "All the ranks should have the same entities.";
+    }
+
+    const auto& entity_group_path = mPrefix + "/" + entity_name;
+
+    const unsigned int num_entities = rEntities.size();
+
+    Vector<int> entity_ids, entitiy_property_ids;
```
matekelemen:
```suggestion
    Vector<int> entity_ids, entity_property_ids;
```

applications/HDF5Application/custom_io/hdf5_connectivities_data.cpp
```diff
@@ -1,257 +1,207 @@
...
+    std::string entity_name = "";
+    unsigned int geometry_size = 0;
+    if (!rEntities.empty()) {
+        CompareElementsAndConditionsUtility::GetRegisteredName(rEntities.front(), entity_name);
+        // no need of doing communication to get the correct geometry size
+        // for all ranks, because the connectivity matrix will be correctly
+        // sized in the HDF5File write call by MPI Communication.
+        // Otherwise, this will be done twice which is redundant.
+        geometry_size = rEntities.front().GetGeometry().size();
+    }
```
matekelemen, line 126: This assumes that there's only one type of entity in the container, which is quite unlikely. Are entities sorted by type into separate containers further up the callstack? If not, how are heterogeneous entity arrays handled?

applications/HDF5Application/custom_io/hdf5_model_part_io.cpp
```diff
@@ -113,15 +113,16 @@ void ModelPartIO::ReadElements(NodesContainerType& rNodes,
...
-        connectivities.CreateEntities(rNodes, rProperties, rElements);
+
+    const auto& element_path = mPrefix + "/Elements";
+
+    for (const auto& r_name : mpFile->GetGroupNames(element_path)) {
+        Internals::ConnectivitiesData<ElementsContainerType> connectivities(element_path, mpFile);
+        connectivities.Read(r_name, rNodes, rProperties, rElements);
     }
 
+    rElements.Unique();
```
matekelemen, line 124:
```suggestion
    rElements.Unique(); // @todo remove when PointerVectorSet is fixed
```

applications/HDF5Application/custom_io/hdf5_connectivities_data.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_connectivities_data.h
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_connectivities_data.cpp
sunethwarna: I would say, we would be breaking all the old HDF5 files. Let's do it in a separate PR then.

applications/HDF5Application/custom_io/hdf5_connectivities_data.cpp
sunethwarna: hmm.... good question. at the moment, we won't get any all rank empty container to this point since we have the following block in `hdf5_model_part_io.cpp`
```
    std::vector<std::string> names;
    std::vector<ElementsContainerType> factored_elements;
    FactorElements(rElements, names, factored_elements);

    const auto& element_path = mPrefix + "/Elements";

    mpFile->AddPath(element_path);
    for (unsigned int i = 0; i < names.size(); ++i) {
        Internals::ConnectivitiesData<ElementsContainerType> connectivities(element_path, mpFile);
        connectivities.Write(factored_elements[i]);
    }
```

Above make sure we don't write any empty containers, and then if we read the written stuff from the HDF5, there won't be any empty container readins as well.

But if someone uses the `ConnectivitiesData` writer stand alone, then it will be a mess.

So it make sense to ignore considering the latter. I have added that.

Did that.

applications/HDF5Application/custom_io/hdf5_connectivities_data.cpp
sunethwarna: done :)

applications/HDF5Application/custom_io/hdf5_connectivities_data.cpp
sunethwarna: this is done in the model part_io as in the following where we call the `FactorElements`. So, this is a valid check.

```c++
    std::vector<std::string> names;
    std::vector<ElementsContainerType> factored_elements;
    FactorElements(rElements, names, factored_elements);

    const auto& element_path = mPrefix + "/Elements";

    mpFile->AddPath(element_path);
    for (unsigned int i = 0; i < names.size(); ++i) {
        Internals::ConnectivitiesData<ElementsContainerType> connectivities(element_path, mpFile);
        connectivities.Write(factored_elements[i]);
    }
```

applications/HDF5Application/custom_io/hdf5_model_part_io.cpp
sunethwarna: I would keep it until I remove the PVS::push_back.

sunethwarna:
@matekelemen This is ready for re-review. I have applied your suggestions where possible, and answered your concerns.

sunethwarna:
@matekelemen The ball is in your court now ;)

applications/HDF5Application/CMakeLists.txt
matekelemen, line 1: what's up with this commit? Why does the order of translation units passed to the build system matter?

sunethwarna: It matters only to the Unity build, otherwise, it is ok.  In the unity build, if it includes the `ModelPartIO` then all the template definitions of the `ConnectuvityData` will be used, hence the compilation fails when we export attributes using the `KRATOS_API`

matekelemen (approved):

PR #11495, opened 2023-08-22
Release 9.4

roigcarlo:
**📝 Description**
Changes and fixes for Release 9.4. This is also serves for the base commit to tag the release.

**🆕 Changelog**
- Release 9.4
- Added tests to Release
- Fixed some tests in win
- Disabled old applications

### DIFF
--- a/scripts/wheels/windows/configure.bat
+++ b/scripts/wheels/windows/configure.bat
@@ -41,7 +41,6 @@ CALL :add_app %KRATOS_APP_DIR%\CompressiblePotentialFlowApplication;
 CALL :add_app %KRATOS_APP_DIR%\HDF5Application;
 CALL :add_app %KRATOS_APP_DIR%\IgaApplication;
 CALL :add_app %KRATOS_APP_DIR%\ChimeraApplication;
-CALL :add_app %KRATOS_APP_DIR%\MultilevelMonteCarloApplication;
 CALL :add_app %KRATOS_APP_DIR%\StatisticsApplication;
 CALL :add_app %KRATOS_APP_DIR%\RomApplication;
 CALL :add_app %KRATOS_APP_DIR%\ShallowWaterApplication;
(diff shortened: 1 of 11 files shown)


scripts/wheels/windows/configure.bat
```diff
@@ -61,7 +60,7 @@ cmake -G"Visual Studio 16 2019" -H"%KRATOS_SOURCE%" -B"%KRATOS_BUILD%\%KRATOS_BU
 -DCMAKE_C_FLAGS="/MP24 /Gm- /Zm10"                                                          ^
 -DCMAKE_CXX_FLAGS="/MP24 /Gm- /Zm10"                                                        ^
 -DBOOST_ROOT=%BOOST_ROOT%                                                                   ^
--DKRATOS_BUILD_TESTING=OFF                                                                  ^
+-DKRATOS_BUILD_TESTING=ON                                                                   ^
```
philbucher, line 63: you sure about this? What would be the benefit for the user? This will only increase size of the binary and compilation times

I dont think this should be part of the wheels. Maybe if at some point the tests are in a separate binary

loumalouomega: I agree with @philbucher

roigcarlo: @pooyan-dadvand Ask me for this. Its still something you need?

sunethwarna: I also think that, no one will run the tests when they install Kratos from wheels.

roigcarlo:
@KratosMultiphysics/technical-committee We mostly decided to have the tests included and work to provide mechanism so the tests are more useful to the user (Print test list, have descriptions, `KM.Tests.RunallTests()`, etc...)

@RiccardoRossi, @pooyan-dadvand Can you comment on this so we can approve it? The release is already published so I would like to merge asap to keep consistency.

PR #11496, opened 2023-08-22
[Doc] Update INSTALL.md with Msys UCRT64 (new standard)

loumalouomega:
**📝 Description**

See https://www.msys2.org/docs/environments/

UCRT (Universal C Runtime) is a newer version which is also used by Microsoft Visual Studio by default. It should work and behave as if the code was compiled with MSVC.

- Better compatibility with MSVC, both at build time and at run time.
- It only ships by default on Windows 10 and for older versions you have to provide it yourself or depend on the user having it installed.

**🆕 Changelog**

- [Update INSTALL.md with Msys UCRT64 (new standard)](https://github.com/KratosMultiphysics/Kratos/commit/813afa11b2b22fafcd436bfe360d991b910347fc)


roigcarlo (changes requested):
No, this is for normal users.

If you want this add it to the doc in a an advanced section or elsewhere. Only simple and supertested here.

roigcarlo:
I have nothing against, but i am tired to fix installations of people that copy paste everything in this file and mix compilers, libraries etc...

loumalouomega:
> I have nothing against, but i am tired to fix installations of people that copy paste everything in this file and mix compilers, libraries etc...

Okay, I will move it to advanced settings

roigcarlo (approved):

PR #11498, opened 2023-08-22
[External] Missing include in Clipper2 library

loumalouomega:
**📝 Description**

With last versions of GCC (13) and Clang (14) compilation fails without this include.

**🆕 Changelog**

- [Missing include in Clipper2 library](https://github.com/KratosMultiphysics/Kratos/commit/33860444096485cd655167032bf4a3497ef5fd3f)

### DIFF
--- a/external_libraries/clipper/include/clipper2/clipper.core.h
+++ b/external_libraries/clipper/include/clipper2/clipper.core.h
@@ -11,6 +11,7 @@
 #define CLIPPER_CORE_H
 
 #include <cstdlib>
+#include <cinttypes>
 #include <cmath>
 #include <vector>
 #include <string>


matekelemen:
Is this still an issue with the current state of this lib? I'd prefer updating the dependency rather than taking over its maintenance within Kratos.

loumalouomega:
> Is this still an issue with the current state of this lib? I'd prefer updating the dependency rather than taking over its maintenance within Kratos.

I haven't summited yet the fix in the official repo, I will do and post the PR in here

loumalouomega:
> > Is this still an issue with the current state of this lib? I'd prefer updating the dependency rather than taking over its maintenance within Kratos.
> 
> I haven't summited yet the fix in the official repo, I will do and post the PR in here

No need, they have added a similar dependency: https://github.com/AngusJohnson/Clipper2/blob/main/CPP/Clipper2Lib/include/clipper2/clipper.core.h `cstdint`, I can just simply update the library, but this may break code

matekelemen:
> I haven't summited yet the fix in the official repo, I will do and post the PR in here

just make sure it's necessary before you do; they may have already fixed it

loumalouomega:
> > I haven't summited yet the fix in the official repo, I will do and post the PR in here
> 
> just make sure it's necessary before you do; they may have already fixed it

They did, see my answer. I propose to merge this PR and update the lib in a following one, that way we can avoid breaking the code (the use of clipper2 is code from @tteschemacher and if broken can be complex to fix)

loumalouomega:
> > > I haven't summited yet the fix in the official repo, I will do and post the PR in here
> > 
> > 
> > just make sure it's necessary before you do; they may have already fixed it
> 
> They did, see my answer. I propose to merge this PR and update the lib in a following one, that way we can avoid breaking the code (the use of clipper2 is code from @tteschemacher and if broken can be complex to fix)

Can we agree on this @matekelemen ?

matekelemen:
Fine, but please open the PR with the complete update before merging this one. I don't want this to be forgotten.

loumalouomega:
> Fine, but please open the PR with the complete update before merging this one. I don't want it to be forgotten.

Done in #11504

loumalouomega:
I will close this directly

Issue #11363, opened 2023-07-06
[Core] PointerVectorSet


sunethwarna:
**Description**
Recently me and @matekelemen (and many other also have complained in past: refer #9227) that there are some inconsistencies in the `PointerVectorSet` and how it is supposed to behave. The followings are the points which we discovered:
1. Not guranteed to be ordered or unique.
2. Has a confusing interface with mixed responsibilities.

## 1: Not guranteed to be ordered or unique

The underlying storage of the `PointerVectorSet` can be mutated and can lose its `Set` properties in the following cases.
1. `PointerVectorSet::GetContainer` - The container can be freely mutated without triggering an update (sort and unique)
2. `PointerVectorSet::push_back` -  Array is appended but not sorted or made unique.
3. `PointerVectorSet::find` - Mutates the data structure if not sorted. (Problematic in OpenMP as well)

## 2: Has a confusing interface with mixed responsibilities
The `PointerVectorSet` has following methods which give the impression it does not gurantee the `Set` properties.
1. `PointerVectorSet::IsSorted`, `PointerVectorSet::Sort`, `PointerVectorSet::Unique` - Is giving wrong impressions about the assumed set properties.
2. Mixed responsibilities (separating them would allow using the features individually as well):
a. `ordered set` data structure
b. value-like access to an array of pointers

## Necessary changes
1. Remove `PointerVectorSet::GetContainer` (both mutable and immutable)
2. Replace `push_back` with `insert(begin, end)`
3. Remove sorting in the mutable `find`
4. Remove `pop_back`
5. Remove `Sort`, `Unique`, `IsSorted`
6. Remove `mSortedPartSize`, `mMaxBufferSize` and their getters and setters.
7. Change `insert(position, value)` => `insert(hint, value)`

This would require multiple PRs with incremental changes, specially when replacing `push_back`.

## Suggestion
Instead of rolling our own implementation, we suggest using a 3rd party implementation, because this is a solved problem. One possible library is [tessil `ordered_set`](https://github.com/Tessil/ordered-map). It is **ordered** and **contiguous** in memory and is **header-only** (3 files).

FYI @philbucher @loumalouomega

loumalouomega:
Everything you said: YES

This is a class is used everywhere but as you said requires a lot of refactor.

kelofm:
[...]

PR #11503, opened 2023-08-24
[Core] Add new `AddNodesFromOrderedContainer` to model part

jrubiogonzalez:
**📝 Description**

This PR Adds a new method to the model part. This method greatly improves performance when the container to be added can be ensured to have the nodes only once and are stored ordered.

**🆕 Changelog**

- Added new method to model_part.h
- Added new test that test both previous AddNodes and new AddNodesFromOrderdContainer

### DIFF
--- a/kratos/includes/model_part.h
+++ b/kratos/includes/model_part.h
@@ -354,14 +354,14 @@ public:
     /** Inserts a list of pointers to nodes
      */
     template<class TIteratorType >
-    void AddNodes(TIteratorType nodes_begin,  TIteratorType nodes_end, IndexType ThisIndex = 0)
+    void AddNodes(TIteratorType iNodesBegin,  TIteratorType iNodesEnd, IndexType ThisIndex = 0)
     {
         KRATOS_TRY
         ModelPart::NodesContainerType  aux;
         ModelPart::NodesContainerType  aux_root; //they may not exist in the root
         ModelPart* root_model_part = &this->GetRootModelPart();
 
-        for(TIteratorType it = nodes_begin; it!=nodes_end; it++)
+        for(TIteratorType it = iNodesBegin; it!=iNodesEnd; it++)
         {
             auto it_found = root_model_part->Nodes().find(it->Id());
             if(it_found == root_model_part->NodesEnd()) //node does not exist in the top model part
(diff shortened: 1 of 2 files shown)


loumalouomega (dismissed):
Okay from my side

kratos/includes/model_part.h
```diff
@@ -399,6 +399,29 @@ class KRATOS_API(KRATOS_CORE) ModelPart final
         KRATOS_CATCH("")
     }
 
+    template<class TIteratorType >
+    void AddNodesFromOrderedContainer(TIteratorType nodes_begin,  TIteratorType nodes_end, IndexType ThisIndex = 0)
```
ddiezrod: ThisIndex is not being used

loumalouomega: Indeed

ddiezrod: Also, very minor, but we use camel case for parameters, and in this case they should start with "i" to indicate its an iterator

loumalouomega: Also input should be in CammelCase format

jrubiogonzalez: I was copying the interface from AddNodes, that TBH has same issue. It is not used. We can Remove both probably.

kratos/includes/model_part.h
```diff
@@ -2023,6 +2047,50 @@ class KRATOS_API(KRATOS_CORE) ModelPart final
...
+            }
+            aux.push_back(*(it1.base()) );
+            if(it2!=elong && (it1->Id() == it2->Id())) { //If both are the same, then we need to skip
+                it2++;
+            }
+        }
+        while(it2 != elong) {
+            aux.push_back(*(it2.base()));
+            it2++;
+        }
```
matekelemen, line 2089: this will still add duplicate nodes if `it1` or `it2` iterate through duplicate nodes

matekelemen (commented):
there are no guarantees that the output is sorted, apart from the user's good conscience that they provide a sorted input. The fact that `PointerVectorSet` exposes such manipulation of its underlying data structure is the exact problem detailed in #11363.

This should be handled by a range insert into `PointerVectorSet` and definitely not implemented from the outside.

kratos/includes/model_part.h
jrubiogonzalez: This is the exact goal of this MR.The condition is being able to provide unique and sorted nodes. It is quite common to have it as the mesh generation directly/indirectly is under our control.

jrubiogonzalez: To provide some context, we are experimenting with a voxel mesher. And for a benchmark example, with these changes it goes from 70s to 20s. Quite a lot if we want to be in the order 1min.

loumalouomega (approved):

rubenzorrilla (changes requested):
Changes in the `ModelPart` API need to be discussed in the @KratosMultiphysics/technical-committee.

EDIT: Though this is a justified demand, I tend to agree @matekelemen and @sunethwarna point about tackling this from the container itself, on top of the fact that there are many routines in Kratos assuming sorted and unique containers. 

If we are to do this from the `ModelPart` as a sort of temporary patch, I think we should consider a different alternative that eases an eventual migration rather than creating a new function to an already exorbitant API.

pooyan-dadvand:
@KratosMultiphysics/technical-committee understands that the modifications proposed by #12087 can solve the problem of sort and unique and can be used as the low level implementation for this one. However, for adding the entities to the different levels of the modelpart hierarchy, we need a separate utility (not in the modelpart itself) to perform that.

RiccardoRossi:
this one shall be closed as the ongoing changes to the PointerVectorSet shall guarantee that it is always in sorted state.

PR #11506, opened 2023-08-25
GTest

roigcarlo:
**📝 Description**
Implementation of GTests.

### Depending on:

- [x] #12317
- [x] #12306

### Tasks

- [x] Fix exit code -11 appearing on some suites
- [x] Fix windows not executing any test 
- [x] Fix `ubuntu-intel` reporting exit code as valid, and tests reporting false positives.

### Related

- [x] #12326 (Currently Skipped)

### Kernel
Kernel changes are being discussed in separate PR's (#12281, #12306). This PR enabled the de-registration of components.

### Testing Mechanism
All cpp can now be run directly from `bin/Release/tests/[SuiteName]`. In order to tests and debug there are several new options like:
- `--gtest_filter="foo"`: To filter tests by name. Accepts wildcards.
- `--gtest_break_on_failure`: To tell to stop in the first failure
- `--gtest_catch_exceptions=0`: To prevent from capturing exepctions and be able to hook gdb directly on the faulty instruction

Complete list can be found in the gtest documentation.

Tests als can be run through the testing scripts as usual:
- `testing/run_python_tests.py`
- `testing/run_python_mpi_tests.py`
- `testing/run_cpp_tests.py`
- `testing/run_cpp_mpi_tests.py`

### Compilation Targets 
Tests are now compiled in the majority of the cases with the `kratos_add_gtests` macro. As example:

```CMAKE
## CableNet Testing sources
if(${KRATOS_BUILD_TESTING} MATCHES ON)
    file(GLOB_RECURSE KRATOS_CABLE_NET_APPLICATION_TESTING_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/tests/*.cpp)
    kratos_add_gtests(TARGET KratosCableNetCore SOURCES "${KRATOS_CABLE_NET_APPLICATION_TESTING_SOURCES}" USE_CUSTOM_MAIN)
endif(${KRATOS_BUILD_TESTING} MATCHES ON)
```

This macro accepts several options:
- `TARGET`: The target name of the library you will tests. In 99% this will be `Kratos[your_app_name]Core`. It will create a  `Kratos[your_app_name]CoreTest` as result.
- `SOURCES`: The list of the cpp files containing your tests and possible utilities needed by them. From now on we will not accept testing utilities as part of the core libraries in order to reduce the weight of the binaries.
- `USE_MPI`: Added this flag if you target depends on MPI (MPICore, Trilinos, MPIExtensions, etc...)
- `USE_CUSTOM_MAIN`: Add this flag if you have defined a custom main for your tests. In this PR this flag is enabled for most of the applications as a transition mechanism to keep the PR as small as possible. More in the transition section.

There are new several compilation targets as a result of the changes in the tests:

```
(lib) KratosCore            <- Main Kratos functionality
(lib) KratosApplication     <- Python Interface (Depends on KratosCore)
(lib) KratosTestingUtils    <- Auxiliar functions for testing (Depends on KratosCore)
(exe) KratosTests           <- The actual tests (Depends on  KratosTestingUtils, KratosCore)
(lib) KratosMPICore         <- Main MPI Kratos functionality
(lib) KratosMPITestingUtils <- Auxiliar functions for testing (Depends on KratosMPICore, KratosCore)
(exe) KratosMPITests        <- The actual tests (Depends on  KratosMPITestingUtils, KratosMPICore, KratosTestingUtils, KratosCore)
```

In general you should not bother about those unless you want to do special things with your applications.

### Transition Changes
Due to some changes in the structure of how the targets are compiler, tests no longer depend on the `testing` sources from Kratos. Specifically a change that will be required is to change "include/testing.h" to an include of you application testing suit.

As this would imply changing all testing sources from Kratos in a single PR, most of applications have a transition solution in which a generic suit is defined in the kernel and the includes are the same. A placeholder is provided as well which will need to be changed to the final solution.

StructuralMechanics, Geomechanics, CoSimulationApplication and MappingApplication have the new mechanism in which the suite is included directly from the application. We expect to replace the rest of applications in future PR once this is merged.

**🆕 Changelog**
- Adding GTests to Kratos


loumalouomega:
We meet again Gtest...

loumalouomega:
Many changes not related to Gtest in here?

kratos/tests/test_utilities/test_laplacian_element.cpp
```diff
@@ -20,7 +20,7 @@
 #include "includes/variables.h"
 #include "utilities/math_utils.h"
 #include "utilities/geometry_utilities.h"
-#include "tests/cpp_tests/auxiliar_files_for_cpp_unnitest/test_laplacian_element.h"
+#include "tests/test_utilities/test_laplacian_element.h"
```
loumalouomega: Can you move these files in an independent PR?

scripts/wheels/linux/configure.sh
```diff
@@ -75,5 +75,4 @@ cmake -H"${KRATOS_SOURCE}" -B"${KRATOS_BUILD}/${KRATOS_BUILD_TYPE}" \
 -DINCLUDE_MMG=ON                                                       \
 -DMMG_ROOT="/workspace/external_libraries/mmg/mmg_5_5_1"               \
 -DKRATOS_BUILD_TESTING=OFF                                             \
--DINSTALL_RUNKRATOS=OFF                                                \
```
loumalouomega: Everything related with RunKratos should go in a independent PR?

kratos/tests/cpp_tests/utilities/test_variable_utils.cpp
```diff
@@ -59,12 +59,13 @@ KRATOS_TEST_CASE_IN_SUITE(VariableUtilsSetHistoricalVariablesToZero, KratosCoreF
...
     // Values are properly allocated
     const double tolerance = 1.0e-12;
+    const auto ZeroArray = array_1d<double, 3>(3, 0.0);
     for (const auto& r_node : r_test_model_part.Nodes()) {
-        KRATOS_CHECK_NEAR(r_node.FastGetSolutionStepValue(PRESSURE), 0.0, tolerance);
-        KRATOS_CHECK_NEAR(r_node.FastGetSolutionStepValue(TEMPERATURE), 0.0, tolerance);
-        KRATOS_CHECK_VECTOR_NEAR(r_node.FastGetSolutionStepValue(VELOCITY), ZeroVector(3), tolerance);
-        KRATOS_CHECK_VECTOR_NEAR(r_node.FastGetSolutionStepValue(DISPLACEMENT), ZeroVector(3), tolerance);
-        KRATOS_CHECK_MATRIX_NEAR(r_node.FastGetSolutionStepValue(DEFORMATION_GRADIENT), ZeroMatrix(0,0), tolerance);
+        KRATOS_EXPECT_NEAR(r_node.FastGetSolutionStepValue(PRESSURE), 0.0, tolerance);
```
loumalouomega, line 64: It was really necessary to change the checks?, it would be smarter to keep them and calling the new ones, and remove bit a bit, otherwise this comprieses many changes

kratos/testing_/test_skipped_exception.h
```diff
@@ -0,0 +1,129 @@
+//    |  /           |
```
loumalouomega: `testing_` folder?

loumalouomega: I see many renamings because of this

kratos/tests/test_utilities/test_laplacian_element.cpp
roigcarlo: I would prefer not to, but if its strictly necessary I can.

kratos/sources/kernel.cpp
```diff
@@ -36,6 +36,36 @@ Kernel::Kernel(bool IsDistributedRun) : mpKratosCoreApplication(Kratos::make_sha
     Initialize();
 }
 
+Kernel::~Kernel() {
+    KratosComponents<Variable<bool>>::Clear();
```
loumalouomega: Why this?, independent PR?

kratos/python/add_testing_to_python.cpp
```diff
@@ -15,56 +15,56 @@
...
+// void ListOfAllTestCases() {
+//     std::cout << Testing::Tester::GetInstance() << std::endl;
+// }
 
 void  AddTestingToPython(pybind11::module& m) {
-	namespace py = pybind11;
+	// namespace py = pybind11;
 
-    py::class_<Testing::Tester, Kratos::shared_ptr<Testing::Tester> > TesterPyBind(m, "Tester");
+    // py::class_<Testing::Tester, Kratos::shared_ptr<Testing::Tester> > TesterPyBind(m, "Tester");
```
loumalouomega: Remove if unused

kratos/tests/cpp_tests/utilities/test_variable_utils.cpp
roigcarlo: Yep, this check specifically fails if trying to compare boost vectors (see https://github.com/KratosMultiphysics/Kratos/pull/11506/commits/8c34e8893df63d25c803ecbb542b75758b4a317b)

kratos/sources/kernel.cpp
roigcarlo: This is needed so the fixtures work correctly. I will detail it in the PR description

kratos/includes/expect.h
```diff
@@ -0,0 +1,211 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:		 BSD License
```
loumalouomega, line 7: Tabs!

kratos/tests/test_utilities/test_laplacian_element.cpp
loumalouomega: This PR is humngous, it will simplify a lot review

kratos/tests/cpp_tests/utilities/test_variable_utils.cpp
loumalouomega: Okay

kratos/sources/kernel.cpp
loumalouomega: Can be done also in a indpendent PR

loumalouomega: Please otherwise will be impossible to check properly

kratos/tests/cpp_tests/utilities/test_variable_utils.cpp
loumalouomega: Can we keep a transition interface?, think thatthis will break many branches and independent projects using Kratos

roigcarlo: I don't see how, I've tried. As I commented in the @KratosMultiphysics/technical-committee there is currently missuse of the `KRATOS_CHECK_XXX` macros, as they are used both as a pseudo-assertions in the tests, and as checks in the rest of the code. 

Since gtest uses full fledged assertions for the test checks, I cannot use the same macro for both, so either we change the KRATOS_CHECK in the code to something else (either be another macro or a proper check) or we make another interface for the assertions based on the existing one (what I did)

If you see any other solution tell me and I will implement it.

kratos/sources/kernel.cpp
roigcarlo: Agree and not. If I make a separate PR for this you will want a test, and the test can only be done if we do this. It's a snake chasing his tail :S

kratos/tests/test_utilities/test_laplacian_element.cpp
roigcarlo: Ok I'll try, but this is only like the base of the PR. I will have to change all the apps as well. I will prepare smaller PR to merge to this one but there is no way that both systems coexists :S so changes need to be in a single PR

kratos/tests/cpp_tests/utilities/test_variable_utils.cpp
loumalouomega: In that case we can change the macro from CHECK to ASSERT in a independent PR, so this PR is not 250 changes and that way other projects can be adapted

loumalouomega:
BTW, no CTest?

roigcarlo:
Its a draft, I already wrote in the desc that this is incomplete. So far the aim is to see of the lib is downloaded and both gtest and gmock are compiled correctly (which they seem to do). Patience.

kratos/testing_/test_skipped_exception.h
roigcarlo: This is leftover, I will remove it, don't worry. Its there because I still need to use things from the older testing.

kratos/tests/cpp_tests/utilities/test_variable_utils.cpp
roigcarlo: The problem is that it will still be a 250+ files changed PR, but as you like.

loumalouomega: Yes, but it will be a simple change-> Just renaming macro. Here we have renaming + gtest + runkratos + ...

roigcarlo:
@markelov208 @avdg81, I am merging the master and some of the most recent changes directly to this branch in preparation for the next week. 

If you don't mind, I will also move the changes from #11896 (specially the ones done be @markelov208  In the geo) to this branch as I have a better grip on the changes so we can have a functional basis to solve the mpi issues.

roigcarlo:
@rfaasse @avdg81 Could you check that the last commit did not break your CI? I have changed the way tests are compiled to applications by default inherit the main that initialized the kernel instead of using the default `gmok` one. I remember you having problems with this with 0 tests detected.

Should not affect but just in case.

rfaasse:
> @rfaasse @avdg81 Could you check that the last commit did not break your CI? I have changed the way tests are compiled to applications by default inherit the main that initialized the kernel instead of using the default `gmok` one. I remember you having problems with this with 0 tests detected.
> 
> Should not affect but just in case.

@roigcarlo Thanks for making these changes, on our pipeline the build and tests pass. For our local build, we need to deal with the dlls being in the bin folder instead of libs, but that shouldn't block this PR. 

Since @avdg81 and I are back again this week, we would like to have a meeting tomorrow afternoon to discuss which steps still need to be done to get this through!

roigcarlo:
Perfect!, Meeting is ok as long as its before 16:30, feel free to send me the invite :+1:

roigcarlo:
@rfaasse @avdg81 Could you check why the windows script is not running any tests :S it works on the machines I've tested (probably a problem with the path separator)

applications/CoSimulationApplication/tests/cpp_distributed_tests/co_simulation_distributed_suite.cpp
```diff
@@ -0,0 +1,27 @@
+// KRATOS  / ___|___/ ___|(_)_ __ ___  _   _| | __ _| |_(_) ___  _ ___
```
RiccardoRossi, line 1: i understand that  these modifications have been agreed with  @philbucher correct?

applications/ConstitutiveLawsApplication/tests/cpp_tests/constitutive_laws_fast_suite.h
```diff
@@ -0,0 +1,29 @@
...
+#include "testing/testing.h"
+
+namespace Kratos::Testing {
+
+class KratosConstitutiveLawsFastSuite : public KratosCoreFastSuite {
+public:
+  KratosConstitutiveLawsFastSuite();
+
+private:
+  KratosConstitutiveLawsApplication::Pointer mpConstitutiveLawsApp;
```
RiccardoRossi, line 25: why do we need this but not the StructuralMechanics?

applications/ContactStructuralMechanicsApplication/tests/cpp_tests/contact_structural_mechanics_fast_suite.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "contact_structural_mechanics_application.h"
+
+int main(int argc, char* argv[]) 
+{
+    ::testing::InitGoogleTest(&argc, argv);
+
+    Kratos::Testing::mApplicationInitializerList.push_back([](std::vector<Kratos::KratosApplication::Pointer> & rRegisteredApplications, Kratos::Kernel & rKernel) {
+      if (!rKernel.IsImported("ContactStructuralMechanicsApplication")) {
+        auto pApplication = std::make_shared<Kratos::KratosContactStructuralMechanicsApplication>();
+        rKernel.ImportApplication(pApplication);
```
RiccardoRossi, line 28: i am surprised that the StructuralMEchanics is not required by these tests.

applications/IgaApplication/tests/cpp_tests/test_iga_shell_3p.cpp
```diff
@@ -50,6 +50,8 @@ namespace Testing
     // Tests the stiffness matrix of the Shell3pElement with a polynomial degree of p=3.
     KRATOS_TEST_CASE_IN_SUITE(IgaShell3pElementP3, KratosIgaFastSuite)
     {
+        KRATOS_SKIP_TEST_IF_NOT(KratosComponents<ConstitutiveLaw>::Has("LinearElasticPlaneStress2DLaw")) << "This test needs the StructuralMechanicsApplication" << std::endl;
```
RiccardoRossi, line 53: as i understand this means that these tests are currently being skipped. @rickyaristio @NickNick9 please take into account that they need to be moved to python ASAP

kratos/tests/cpp_tests/testing/test_testing.cpp.disabled
RiccardoRossi, line 1: this is intentional or is it a leftover?

RiccardoRossi (commented):
some comments. I cannot claim this is a detailed review, nevertheless this looks good to me

kratos/tests/cpp_tests/testing/test_testing.cpp.disabled
roigcarlo: Leftover. Will fix

applications/IgaApplication/tests/cpp_tests/test_iga_shell_3p.cpp
roigcarlo: this is being tracked in https://github.com/KratosMultiphysics/Kratos/issues/12326

jcotela:
@roigcarlo I have pushed the changes for the issue we checked earlier. Since I had added the test to the wrong suite originally, I just moved the test to the standard core suite and removed the "extra" suite class.

roigcarlo:
Hi @KratosMultiphysics/technical-committee can someone make a final review to this? As far as I am concern seems to be working well for all the CI external to CIMNE as well as our in house CI.

pooyan-dadvand:
@KratosMultiphysics/technical-committee approves this PR as it passes all the tests and provides mechanisms to handle diverse usage of the cpp tests in Kratos.

rubenzorrilla (approved):

loumalouomega:
Gtest won...

PR #11507, opened 2023-08-25
[HDF5] Update - 9 - python ModelPart script

sunethwarna:
**📝 Description**
Updated the `model_part.py` script to use the new c++ side features in HDF5.

**🆕 Changelog**
- Update `model_part.py`.


RiccardoRossi:
i am completely outside of the discussion about the hdf5 application. Is anyone in @KratosMultiphysics/technical-committee (aside @sunethwarna who is authoring this) taking care? 

In any case this is nonintrusive i believe, in the sense that as long as it is portable to the windows there should not be any issue to have you guys deciding. @roigcarlo do you have a specific position about this? (i know that you have a firm opinion about HDF5)

matekelemen:
@sunethwarna is refactoring the app and I'm reviewing it. There's quite a few already merged and you'll see plenty more in the pipeline if you search for HDF5 among open PRs.

We're in the middle of a new motivational wave and hope to ride it to completion.

RiccardoRossi:
Well, if you manage to print also for gid we get rid of the gidpost for good...

matekelemen:
If the GiD team takes a look, requests **small** changes to the HDF5 app, and implements a replacement for `gidpost`, I have no problem with the idea.

However, I wouldn't want to do it myself because
- then we (@suneth and I) would have to maintain it as well. That's potentially a lot of work since we don't have any say in GiD development.
- we'd have to get familiar with GiD's format, possibly redesign the HDF5 app's interface, and implement everything ourselves
- I'm unfamiliar with GiD, have no access to it, and cannot test it

applications/HDF5Application/python_scripts/core/utils.py
```diff
@@ -12,6 +12,17 @@
...
 import KratosMultiphysics
 
+def EvaluatePattern(pattern: str, model_part: KratosMultiphysics.ModelPart, time_format='') -> str:
+    time = model_part.ProcessInfo[KratosMultiphysics.TIME]
+    pattern = pattern.replace("<time>", format(time, time_format))
+
+    if KratosMultiphysics.STEP in model_part.ProcessInfo:
+        step = model_part.ProcessInfo[KratosMultiphysics.STEP]
+    else:
+        step = 0
```
matekelemen: why does `TIME` not get the same treatment?

applications/HDF5Application/python_scripts/core/operations/model_part.py
```diff
@@ -141,223 +127,118 @@ class ProcessInfoInput(ProcessInfoIOOperation):
...
-    '''Writes non-historical nodal flag values to a file.'''
+class VariableOutputOperation(IOOperation):
+    def _MakeIO(self) -> typing.Any:
+        class_name = self.__class__.__name__
+        if class_name.endswith("Output"):
+            class_type_name = f"HDF5{class_name[:-6]}IO"
+            if hasattr(KratosHDF5, class_type_name):
+                return getattr(KratosHDF5, class_type_name)(self.parameters, self.file)
+            else:
+                raise RuntimeError(f"The {class_type_name} not found in HDF5Application.")
```
matekelemen: ok this needs some explanation in the form of some comments

applications/HDF5Application/python_scripts/core/operations/model_part.py
```diff
@@ -141,223 +127,118 @@ class ProcessInfoInput(ProcessInfoIOOperation):
...
+class VariableInputOperation(IOOperation):
+    def _MakeIO(self) -> typing.Any:
+        class_name = self.__class__.__name__
+        if class_name.endswith("Input"):
+            class_type_name = f"HDF5{class_name[:-5]}IO"
+            if hasattr(KratosHDF5, class_type_name):
+                return getattr(KratosHDF5, class_type_name)(self.parameters, self.file)
+            else:
+                raise RuntimeError(f"The {class_type_name} not found in HDF5Application.")
+        else:
```
matekelemen: same; needs explanation what's going on

applications/HDF5Application/python_scripts/core/operations/model_part.py
```diff
@@ -141,223 +127,118 @@ class ProcessInfoInput(ProcessInfoIOOperation):
...
     '''Reads nodal solution step data from a file.
 
     This is used by the transient adjoint solvers.
     '''
-
-    def Execute(self) -> None:
-        KratosHDF5.HDF5NodalSolutionStepBossakIO(
-            self.parameters,
-            self.file).ReadNodalResults(self.model_part)
+    def _MakeIO(self) -> typing.Any:
```
matekelemen, line 244: don't these IO return types have a common base class you could provide rather than `typing.Any`?

applications/HDF5Application/python_scripts/core/utils.py
sunethwarna: Actually, both does not need this check since `ProcessInfo` is a data value container, and when `ProcessInfo::GetValue` called in serial, if the variable is not found, it adds the zero value of the variable and returns it. So, I removed this check as well.

applications/HDF5Application/python_scripts/core/operations/model_part.py
sunethwarna: done :)

applications/HDF5Application/python_scripts/core/operations/model_part.py
sunethwarna: done :)

applications/HDF5Application/python_scripts/core/operations/model_part.py
sunethwarna: Unfortunately no, since some IO classes does not need to have an instance, they only have method. But some need instances, then have classes... :/

applications/HDF5Application/python_scripts/core/operations/model_part.py
sunethwarna: done :)

applications/HDF5Application/python_scripts/core/operations/model_part.py
sunethwarna: done :)

applications/HDF5Application/python_scripts/core/utils.py
matekelemen: That's not great, I'd rather just do the check both for `TIME` and `STEP` because I wouldn't expect an output process to mutate anything in my data structures.

matekelemen (approved):

PR #11509, opened 2023-08-27
[Core] Controllers - Add Output Controller

sunethwarna:
**📝 Description**
Currently, the `Controller::Evaluate` method is made non-const which makes it to have multiple responsibilities such as evaluating a given controller and returning a bool, and changing the state of the controller. This becomes problematic if someone wants to evaluate multiple times in the same state which is not possible without hackish ways. In this PR, I would like to seperate these two responsibilities, so one can evaluate multiple times in the same state, and if required to change state, they can do it seperately.

1. The **`Controller::Evaluate` method is made `const`** so that, now one can only do evaluation, not changing the state of the controller.
2. `Controller::Update` method is introduced which allows changing the state of the `Controller`.

I introduced controllers:
1. `OutputController` -> This has the exact same logic which is used in `VtkOutputProcess` and `VtuOutputProcess` (Which will be used in `HDF5OutputProcess` as well.

**TODO**
- [x]  Add tests if agreed upon.

**🆕 Changelog**
- Add `OutputController`
- Add tests
- Updated `VtuOutputProcess`, `VtkOutputProcess`, `GidOutputProcess`, `UnvOutputProcess`


rubenzorrilla (commented):
I think that thes changes in the base controller really make sense. My unique comment is the new `DefaultController`, why is it required?

EDIT: I'd take the chance to also update the GiD output process accordingly.

sunethwarna:
@rubenzorrilla I did not want to gave if blocks in the rest of the Kratos if someone wishes to have always true controller (I would be using it in the HDF5).

I will change the gid_output process as well.

ddiezrod:
I agree with the changes in the base class, it makes complete sense to separate it into two methods 👍

kratos/controllers/temporal_controller.cpp
```diff
@@ -0,0 +1,158 @@
...
+    : mpModel(&rModel)
+{
+    KRATOS_TRY
+
+    Settings.AddMissingParameters(GetDefaultParameters());
+
+    mModelPartName = Settings["model_part_name"].GetString();
+
+    const auto& output_control_type = Settings["output_control_type"].GetString();
+    if (output_control_type == "step") {
```
roigcarlo: Agree with the changes, but tbh, if we are going to derive different controllers from the base, I would move this to a different one (Time for time, StepController for steps, etc...).

But I would understand leaving this for historical reasons...

sunethwarna: It is for the historical reasons :/

rubenzorrilla: I thought the same. Maybe we can call it OutputController to make clear that this is for the output and save the Time keyword for an eventual time-based cotroller (same argument applies to the step keyword).

sunethwarna: Make sense. I will do it :)

sunethwarna: done :) can somebody approve this?

sunethwarna:
Can someone approve this? @rubenzorrilla @roigcarlo @ddiezrod ?

rubenzorrilla (approved):
Thanks for the implementation.

PR #11518, opened 2023-08-28
[HDF5] Update - 10 - Update python scripts

sunethwarna:
**📝 Description**
This PR updates the following:
1. `Controllers` are now derrived from the Kratos core.
2. Core based `TemporalController` is used, which removes the internal tracking of `time` and `step`
3. New `DefaultController` and `SingleTimeController` is introduced for always active and one time active controller respectively.
4. `ControlledOperation` is introduced which attach a `Controller` to an `Operation`.
5. `AggregatedControlledOperation` is introduced which attach multiple `ControlledOperation` to one single `H5` file.
6. `HDF5Process` and `HDF5OutputProcess` is introduced.
7. Pattern matching with regex for Path search is introduced.

Requires #11509 

**🆕 Changelog**
- Added new controllers
- Added new operations
- Added tests.


sunethwarna:
@matekelemen this is ready for your review :)

applications/HDF5Application/python_scripts/core/controllers.py
```diff
@@ -11,120 +11,34 @@
...
-    @abc.abstractmethod
-    def __call__(self) -> None:
-        """!Execute assigned operations if a check is passed."""
-        pass
-
-    def ExecuteOperation(self) -> None:
-        """!Execute all assigned operations, bypassing the controller's checks."""
-        self.__operation.Execute()
 
+class DefaultController(KratosMultiphysics.Controller):
```
matekelemen, line 20: `Scheduler` would probably be a better name for these classes, but I'm not going to nitcpick about it at this point.

applications/HDF5Application/python_scripts/core/controllers.py
```diff
@@ -11,120 +11,34 @@
...
-
-    def ExecuteOperation(self) -> None:
-        super().ExecuteOperation()
-        self.__last_output_time = self._model_part.ProcessInfo[KratosMultiphysics.TIME]
-        self.__last_output_step = self._model_part.ProcessInfo[KratosMultiphysics.STEP]
+class SingleTimeController(KratosMultiphysics.Controller):
+    def __init__(self, temporal_controller: KratosMultiphysics.OutputController) -> None:
+        KratosMultiphysics.Controller.__init__(self)
+        self.__temporal_controller = temporal_controller
+        self.__is_evaluated = temporal_controller.GetNextPossibleEvaluateControlValue() > temporal_controller.GetInterval()
```
matekelemen, line 31: I've never seen `Controller` in core before and I'm worried about its interface.

applications/HDF5Application/python_scripts/core/pattern.py
```diff
@@ -0,0 +1,163 @@
...
+            "|": r"\|",
+            "?": r"\?",
+            "*": r"\*",
+            "(": r"\(",
+            ")": r"\)",
+            "[": r"\[",
+            "]": r"\]",
+            "{": r"\{",
+            "}": r"\}"
+        }
```
matekelemen: it's almost as if there was a simple pattern in this map. Something like prepending whatever character you have with a backslash.

What I'm trying to say is that you don't need this map. Also you're missing the backslash `\` from this list.

applications/HDF5Application/python_scripts/core/pattern.py
```diff
@@ -0,0 +1,163 @@
...
+            if tag in tag_type_dict.keys():
+                self.__converters.append(tag_type_dict[tag])
+
+        for tag, value_type in tag_type_dict.items():
+            if value_type == int:
+                regex_pattern = regex_pattern.replace(tag, r"([0-9]+)")
+            elif value_type == float:
+                regex_pattern = regex_pattern.replace(tag, r"([0-9]*\.?[0-9]+(?:[eE][-+]?[0-9]+)?)")
+            else:
+                raise RuntimeError(f"Unsupported tag type for tag = \"{tag}\" [ type = {value_type} ].")
```
matekelemen, line 66: What are you trying to do here? I have a terrible feeling that you're missing **lots** of corner cases.

applications/HDF5Application/python_scripts/core/pattern.py
```diff
@@ -0,0 +1,163 @@
...
+        if len(values) >= 1:
+            converted_values = [converter(v) for v, converter in zip(self.__value_getter_method(values), self.__converters)]
+            return True,  converted_values
+        else:
+            return False, []
+
+    def GetNumberOfDataItems(self) -> int:
+        return len(self.__converters)
+
+def __GetMachingEntities(starting_entity: PatternEntity, patterns: 'list[str]', tag_type_dict: 'dict[str, typing.Any]', common_data: 'list[typing.Any]') -> 'typing.Generator[tuple[PatternEntity, typing.Any]]':
```
matekelemen:
```suggestion
def __GetMatchingEntities(starting_entity: PatternEntity, patterns: 'list[str]', tag_type_dict: 'dict[str, typing.Any]', common_data: 'list[typing.Any]') -> 'typing.Generator[tuple[PatternEntity, typing.Any]]':
```

applications/HDF5Application/python_scripts/core/pattern.py
```diff
@@ -0,0 +1,163 @@
+import re
+import abc
+import typing
+from pathlib import Path
+
+import KratosMultiphysics as Kratos
+
+class PatternEntity(abc.ABC):
```
matekelemen, line 8: What does `Entity` refer to in this context? I have a hunch that you probably don't need it in this class' name.

applications/HDF5Application/python_scripts/core/pattern.py
```diff
@@ -0,0 +1,163 @@
...
+                    yield itr, *common_data, *file_pattern_data
+    else:
+        # this is a dir pattern
+        pattern = Pattern(patterns[0], tag_type_dict)
+        for itr in starting_entity.Iterate():
+            if not itr.IsLeaf():
+                is_valid, dir_pattern_data = pattern.GetData(itr.Name())
+                if is_valid:
+                    for data in __GetMachingEntities(itr, patterns[1:], tag_type_dict, [*common_data, *dir_pattern_data]):
+                        yield data
```
matekelemen, line 102: Yes indeed I implemented this in #10793. Please take a look at it.

applications/HDF5Application/python_scripts/core/pattern.py
```diff
@@ -0,0 +1,163 @@
...
+                initial_value_index += 1
+                found_tag = True
+                break
+            current_value_index += 1
+
+        if not found_tag:
+            pattern += entity_name[current_pos]
+            current_pos += 1
+
+    return pattern, tags_dict
```
matekelemen, line 165: This function is even more mysterious to me. I hope you're not trying to reconstruct a regex from an actual string.

Anyway please add docs because I can't continue with the review.

matekelemen (changes requested):

applications/HDF5Application/python_scripts/core/controllers.py
sunethwarna: Hmm... I like the `Controller` also, anyways this is a core thingy....

applications/HDF5Application/python_scripts/core/controllers.py
sunethwarna: Discussion for a different day. I would say, `Controllers` have a fair interface, `OutputControllers` may need some adaptation. Anyways, those are fairly new and most people doesn't even know their existance to use them. So hardly used -> hardly updated.

applications/HDF5Application/python_scripts/core/pattern.py
sunethwarna: to replace tags such as `<time>`, `<step>` with correct regex for identification of the files.

sunethwarna: That is perfectly valid concern @matekelemen . May be i am missing corner cases, but i don't know how to do it better. Any suggestions? At least it works for the current cases, may be we have to improve it when it is used heavily. But, cannot say or do anything about it now i think. I am open for suggestions if you have any....

applications/HDF5Application/python_scripts/core/pattern.py
sunethwarna: backslash is there in the below line.
We need the map because, I don't expect some one to write the proper regex in the json file. This is used to translate the tag based json input to a regex string.

applications/HDF5Application/python_scripts/core/pattern.py
sunethwarna: Ok, i see, unlucky me... Would have saved my time... anyways, I think we have to go with this.

TBH: I think, with the new developments I did to vtu, we can do a better job in hdf5 also to get rid of the whole pattern matching and be more consistent and correct. But that needs to wait and will not be a part of this PR. I will do it once the VtuOutput is finally done.

applications/HDF5Application/python_scripts/core/pattern.py
sunethwarna: That is what exactly i am doing her unfortunately. I will add docs.

sunethwarna: done :)

applications/HDF5Application/python_scripts/core/pattern.py
sunethwarna: The problem is I have below the `Pattern` class, so this needs to be different. I am open for suggestions.

sunethwarna:
@matekelemen The ball is in your court again :D

matekelemen:
Looks like some of my comments were not posted? Idk what's going on with github. Anyway I was asking for docs and comments, especially for `operations/pattern.py`.

applications/HDF5Application/python_scripts/core/pattern.py
matekelemen: It's hard to come up with suggestions because I don't know what the class is meant to do. This class is in dire need of docs.

applications/HDF5Application/python_scripts/core/pattern.py
matekelemen: I mean
```diff
- map = {"." : r"\.", ...}
- for k, v in map.items():
-     pattern = pattern.replace(k, v)
+ characters = ["c", ...]
+ for k in characters:
+     pattern = pattern.replace(k, r"\" + k)
```

applications/HDF5Application/python_scripts/core/pattern.py
matekelemen: well, if you're only looking for `<time>` and `<step>` then don't do regex and only match `<time>` and `<step>` directly. Also the numeric patterns are broken. For example they don't match negative values (even though we don't forbid negative steps and times in kratos).

These nuances are part of the reason why I recommended my old PR. I did quite a bit of testing for them.

applications/HDF5Application/python_scripts/core/pattern.py
matekelemen: I know we're short on time but if we improperly rush-implement stuff, it's only going to generate more headaches for us down the line. Surely you don't think your workload will be lighter any time soon. And if that's the case, you won't spend time on fixing something that "*kinda works*".

applications/HDF5Application/python_scripts/core/pattern.py
matekelemen: > That is what exactly i am doing her unfortunately. I will add docs.

why do you need to do that? It's an impossible task, and will likely generate an endless stream of bugs.

applications/HDF5Application/python_scripts/core/pattern.py
```diff
@@ -0,0 +1,179 @@
...
+    It also constructs a dictionary mapping each placeholder tag to the `float` type.
+
+    Args:
+        entity_name (str): The input string potentially containing floating-point numbers.
+
+    Returns:
+        tuple[str, dict[str, typing.Any]]:
+            - pattern (str): The input string with all float values replaced by their corresponding tags.
+            - tags_dict (dict[str, typing.Any]): A dictionary mapping each tag to the `float` type.
+    """
```
matekelemen, line 136: We're using doxygen to generate python docs as well, so annotations need to be in doxygen format.

applications/HDF5Application/python_scripts/core/pattern.py
matekelemen: It seems to me that this function is only used in testing. If that really is the case, can you please remove it? I'm very uncomfortable with this feature.

applications/HDF5Application/tests/test_pattern.py
```diff
@@ -0,0 +1,133 @@
+from typing import Any
+
+import KratosMultiphysics.KratosUnittest as KratosUnittest
+from KratosMultiphysics.HDF5Application.core.pattern import PatternEntity
+from KratosMultiphysics.HDF5Application.core.pattern import GetMatchingEntities
+from KratosMultiphysics.HDF5Application.core.pattern import IdentifyPattern
```
matekelemen:
```suggestion
from KratosMultiphysics.HDF5Application.core.pattern import PatternEntity, GetMatchingEntities, IdentifyPattern
```

applications/HDF5Application/python_scripts/core/controllers.py
matekelemen: `Controller` seems to be designed exclusively with its **current** behavior in mind (regular time/step intervals) but it could be so much more useful.

Imagine that you want to trigger something (like an output process) when some event happens, but you don't know exactly when (e.g.: contact, buckling, onset of plasticity, etc). The current interface makes these impossible.

applications/HDF5Application/python_scripts/core/pattern.py
sunethwarna: Unfortunately, this is used in the #11534 , Therefore I would not like to remove it.

applications/HDF5Application/python_scripts/core/controllers.py
sunethwarna: Sure, as you said... The `Controller` was exclusively designed to work with step/time/mesh update points.... Making it generic as you suggest would require a lot of work, and may have a lot of corner cases which it fails... Just an opinion though.....

applications/HDF5Application/python_scripts/core/pattern.py
sunethwarna: Two reasons why we cannot support negative values at all in any future....
1. If someone has in the json the following file name format `<model_part_full_name>-<time>`, no regex can detect this time as positive if we include the negative sign support (This is the most used string for the file name and for step also)
2. `TIME` and `STEP` need to be always positive. If someone do negative values, it does not have a meaning. Why support it when it does not have any meaning at all.
3. `TIME` and `STEP` are `int` only because we don't have `Variable<unsigned int>`.

So negative regex matching will not be possible, and will break everything.

I understand your implementation is superior to mine, but I see you support negative values as well. That will be a big blocker for this since we simply cannot support identifying values as negative values, because this use case is different. Anyhow, if we are to use your system, I would bring it to the HDF5App, and would not put it in the core (as you know the reasons).

So, what do you think? Could you work on your #10793 to have an option to ignore negative numbers? Then I could integrate that into HDF5App....

applications/HDF5Application/python_scripts/core/pattern.py
sunethwarna: I fully understand where this is coming from.... hehe... As you know, these PRs were thoguht and discussed by both of us 2 years back, and we have learnt a lot within this time period. So, you have my gurantee once the `VtuOutput` method is finalized, we can use same mechanism to keep track of the file names, or something similar. But for the time being, this is tested for the most common use cases well engough (may be not to your PRs extent, but your PR is more generic than this use case anyways). 

So as i suggested in the previous comment, there are few blockers to use your PR. What do you think?

PS: When I was going through this part of the code with the pattern maching, I am starting to hate pattern matching which always can have element of surprise failures with crazy filename patterns (too complicated implementation for sure)... So, as i suggested, best is to get rid of this whole pattern matching.

applications/HDF5Application/python_scripts/core/pattern.py
sunethwarna: sure, my bad

applications/HDF5Application/python_scripts/core/pattern.py
sunethwarna: good point.... I will add documentation.

applications/HDF5Application/python_scripts/core/pattern.py
sunethwarna: done :)

sunethwarna:
@matekelemen I addressed your comments. :)

applications/HDF5Application/python_scripts/core/pattern.py
sunethwarna: done :)

applications/HDF5Application/python_scripts/core/pattern.py
matekelemen: Since we decided on dropping further developments to XDMF, can you please remove this?

applications/HDF5Application/python_scripts/core/pattern.py
matekelemen: Can you please check whether you still need the pattern matching if we don't mess around with XDMF?

PR #11526, opened 2023-08-30
[Testing] Split Checks

roigcarlo:
**📝 Description**
This splits the check.h file into two. Expect and Check. Incode checks reamin the same, and tests assertions will use the expect interface. This change does nothing by its own and its needed for #11506 as requested by @loumalouomega.

String checks may require further changes in the GTest PR.

**🆕 Changelog**
- Split check.h in check.h and expect.h


loumalouomega (dismissed):
If only these changes then Okay. But maybe should a warning be added in Cmake, otherwise many branches of developers may not compile or change behaviour

roigcarlo:
I would add the warning in the other PR, this at most will cause a couple of tests from branches to fail, but if they keep using checks it will break nothing

loumalouomega (dismissed):

loumalouomega (dismissed):

loumalouomega:
Some Ci are failing due segmentation fault

roigcarlo:
@loumalouomega I am aware. I get mails from the CI

loumalouomega (dismissed):

loumalouomega (approved):

PR #11533, opened 2023-09-01
[FastPR][Core] Adding HasPrototypeEntity to EntityIdentifier

rubenzorrilla:
**📝 Description**
On top of being handy, this is required to avoid the error thrown of the getter when using the EntityIdentifier in a context that mixes elements and conditions (see #11370 as an example).

### DIFF
--- a/kratos/utilities/entities_utilities.cpp
+++ b/kratos/utilities/entities_utilities.cpp
@@ -63,15 +63,6 @@ bool EntitityIdentifier<TEntity>::IsInitialized() const
 /***********************************************************************************/
 
 template<class TEntity>
-bool EntitityIdentifier<TEntity>::HasPrototypeEntity(typename GeometryType::Pointer pGeometry) const
-{
-    return mTypes[static_cast<std::size_t>(pGeometry->GetGeometryType())] != nullptr;
-}
-
-/***********************************************************************************/
-/***********************************************************************************/
-
-template<class TEntity>
 bool EntitityIdentifier<TEntity>::HasPrototypeEntity(const GeometryType& rGeometry) const
 {
     return mTypes[static_cast<std::size_t>(rGeometry.GetGeometryType())] != nullptr;
--- a/kratos/utilities/entities_utilities.cpp
+++ b/kratos/utilities/entities_utilities.cpp
@@ -63,7 +63,7 @@ bool EntitityIdentifier<TEntity>::IsInitialized() const
 /***********************************************************************************/
 
 template<class TEntity>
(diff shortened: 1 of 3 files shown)


loumalouomega (approved):
I guess this is so simple we can live without test

loumalouomega (dismissed):

kratos/utilities/entities_utilities.h
```diff
@@ -103,6 +103,22 @@ namespace EntitiesUtilities
          */
         bool IsInitialized() const;
 
+        /**
+         * @brief Get the prototype entity.
+         * @param pGeometry The pointer to the geometry.
+         * @return true there is a prototype for the provided entity.
+         * @return false there is no prototype for the provided entity.
+         */
+        bool HasPrototypeEntity(typename GeometryType::Pointer pGeometry) const;
```
philbucher: this creates another shared pointer on every fct call
You sure this is necessary? I would not add it, or at least make it a const ref

kratos/utilities/entities_utilities.cpp
```diff
@@ -62,6 +62,24 @@ bool EntitityIdentifier<TEntity>::IsInitialized() const
 /***********************************************************************************/
 /***********************************************************************************/
 
+template<class TEntity>
+bool EntitityIdentifier<TEntity>::HasPrototypeEntity(typename GeometryType::Pointer pGeometry) const
+{
+    return mTypes[static_cast<std::size_t>(pGeometry->GetGeometryType())] != nullptr;
```
philbucher: this will segfault if `pGeometry` is a `nullptr`

Another reason IMO for not adding this version, the other one puts the responsibility of the nu--check on the caller (which is where I believe it should be

philbucher (commented):
some design comments

Also test is missing

kratos/utilities/entities_utilities.h
rubenzorrilla: I added the pointer and reference version of the method to be consistent with the getter. Nevertheless, I've no objection on removing the pointer version.

kratos/utilities/entities_utilities.cpp
rubenzorrilla: Makes sense. I'll remove the pointer version then.

rubenzorrilla:
> some design comments
> 
> Also test is missing

I didn't found any test for this utility. I'll create a new one.

rubenzorrilla:
@philbucher I removed the pointer version and added a test in the last commit.

philbucher (approved):
:heart:

PR #11540, opened 2023-09-05
[Core] make remove installation directories optional

jrubiogonzalez:
**📝 Description**
This PR makes removing KratosMultiphysics and libs directory optional. This might be dangerous if some applications get stuck, but on our case as we have plenty of external dll's and several applications this reduces the build time for small changes significantly.


roigcarlo (dismissed):

CMakeLists.txt
```diff
@@ -46,6 +46,12 @@ if(NOT DEFINED INSTALL_RUNKRATOS)
   set(INSTALL_RUNKRATOS ON)
 endif(NOT DEFINED INSTALL_RUNKRATOS)
 
+if(NOT DEFINED REMOVE_INSTALL_DIRECTORIES)
+  message("-- No REMOVE_INSTALL_DIRECTORIES is defined, setting to ON")
+  set(REMOVE_INSTALL_DIRECTORIES ON)
+endif(NOT DEFINED REMOVE_INSTALL_DIRECTORIES)
```
philbucher, line 52: option(REMOVE_INSTALL_DIRECTORIES "Remove the install directories to achieve a clean compilation" ON)

philbucher (changes requested):
please use more modern CMake syntax

CMakeLists.txt
philbucher:
```suggestion
option(REMOVE_INSTALL_DIRECTORIES "Remove the install directories to achieve a clean compilation" ON)
```

please use modern Cmake Syntax

matekelemen:
@jrubiogonzalez just out of curiosity:
- why does it take so much time to copy the external libs back into the install dir?
   - are you building them directly into the install dir?
   - or do you have to manually copy them there?
- how many/large external libs and apps are we talking about?
- can you give me a rough estimate of how much faster the build process is without clearing the install dir first?

jrubiogonzalez:
> 

We are compiling around 17 applications total between public and private and adding few extra libraries, including mkl. That sums up about 700Mb in win and far more in linux.

If we add our environment anti-virus that is quite intrusive (specially in windows) we go in a 0 build (no changes at all) in my workstation from 15s to 5s.

philbucher (approved):
Nice, thx!

CMakeLists.txt
roigcarlo: We should definitely change that all over kratos...

PR #11541, opened 2023-09-06
[DAMApplication] Fix Python Wheel

roigcarlo:
**📝 Description**
Fix wrong filename preventing the dam(n) wheel from being generated.
This is what was preventing the wheel from being generated since 9.2....

**🆕 Changelog**
- Change README.txt to README.md.


joaquinirazabal (approved):
Thanks!

PR #11547, opened 2023-09-08
[Core] Adding clearing of processes to `Clear` in `AnalysisStage`

loumalouomega:
**📝 Description**

This PR makes improvements to the `analysis_stage.py` file. 

Mainly it adds the clearing of processes to the `Clear` method, where right now only the  solver is cleared. This is not only an enhancement, but also it is more consistent with the rest of `AnalysisStage`, where for example the `Check` method not only checks the solver, but the processes too.

Additionally, it refactors the `Check` and `Clear` methods documentation, adding additonal comments and definitions.

@KratosMultiphysics/altair FYI, we already do this in our custom base analysis class, once this is merged we can remove our local implementation.

**🆕 Changelog**

- [Adding clearing of processes to Clear in AnalysisStage](https://github.com/KratosMultiphysics/Kratos/commit/8569daa7356dddef7f26023be7491f127689e531)

### DIFF
--- a/kratos/python_scripts/analysis_stage.py
+++ b/kratos/python_scripts/analysis_stage.py
@@ -177,15 +177,29 @@ class AnalysisStage(object):
 
     def Check(self):
         """This function checks the AnalysisStage
+
+        Keyword arguments:
+        self -- It signifies an instance of a class.
         """
+        # Checking solver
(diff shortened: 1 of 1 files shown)


rubenzorrilla (approved):
Makes sense.

PR #11555, opened 2023-09-12
[GeoMechanicsApplication] Added a process factory to be used by custom workflows (in C++)

### LINKED ISSUE #11556: [GeoMechanicsApplication] Create a process factory for custom workflows (in C++)
The purpose of the factory is to produce process objects with a common interface (i.e. `Process`), where each produced process has a concrete implementation behind it that is targeted at the intended process (e.g. to carry out the K0 procedure). We will need this in our custom workflows, e.g. for settlement analysis.

avdg81:
**📝 Description**
The factory first needs to be populated with creators for concrete process types that need to be produced. On request, it will produce a smart pointer to a `Process` instance, thereby supplying a uniform and consistent interface to the processes.

This commit includes the implementation of the factory and a set of unit tests. The actual usage by a custom workflow will be part of another commit.

### DIFF
--- a/applications/GeoMechanicsApplication/custom_utilities/process_factory.cpp
+++ b/applications/GeoMechanicsApplication/custom_utilities/process_factory.cpp
@@ -1,0 +1,38 @@
+// KRATOS___
+//     //   ) )
+//    //         ___      ___
+//   //  ____  //___) ) //   ) )
+//  //    / / //       //   / /
+// ((____/ / ((____   ((___/ /  MECHANICS
+//
+//  License:         geo_mechanics_application/license.txt
+//
+//  Main authors:    Wijtze Pieter Kikstra
+//                   Anne van de Graaf
+//
+
+#include "custom_utilities/process_factory.hpp"
+#include "includes/kratos_parameters.h"
+
+
+namespace Kratos
+{
+
+ProcessFactory::ProductType ProcessFactory::Create(const std::string& rProcessClassName,
+                                                   const Parameters&  rProcessSettings) const
+{
+    auto pos = mCreatorMap.find(rProcessClassName);
+    if (pos == mCreatorMap.end()) {
+        return nullptr;
+    }
+
+    return pos->second ? pos->second(rProcessSettings) : nullptr;
+}
+
+void ProcessFactory::AddCreator(const std::string&                            rProcessClassName,
+                                std::function<ProductType(const Parameters&)> Creator)
+{
+    mCreatorMap[rProcessClassName] = std::move(Creator);
+}
+
+}
--- a/applications/GeoMechanicsApplication/tests/cpp_tests/test_process_factory.cpp
+++ b/applications/GeoMechanicsApplication/tests/cpp_tests/test_process_factory.cpp
@@ -1,0 +1,123 @@
+// KRATOS___
+//     //   ) )
+//    //         ___      ___
+//   //  ____  //___) ) //   ) )
+//  //    / / //       //   / /
+// ((____/ / ((____   ((___/ /  MECHANICS
+//
+//  License:         geo_mechanics_application/license.txt
+//
+//  Main authors:    Wijtze Pieter Kikstra
+//                   Anne van de Graaf
+//
+
+#include "testing/testing.h"
+#include "custom_utilities/process_factory.hpp"
+
+using namespace Kratos;
+
+
+namespace
+{
+
+class FooProcess : public Process
+{
+};
+
+class BarProcess : public Process
+{
+};
+
+}
+
+
+namespace Kratos::Testing
+{
+
+KRATOS_TEST_CASE_IN_SUITE(CreateNothingWhenNoCreatorWasAddedForRequestedProcess, KratosGeoMechanicsFastSuite)
+{
+    ProcessFactory factory;
+
+    const Parameters process_settings;
+    const auto process = factory.Create("UnknownProcess", process_settings);
+
+    KRATOS_EXPECT_EQ(process.get(), nullptr);
+}
+
+KRATOS_TEST_CASE_IN_SUITE(CreateNothingWhenTheAddedCreatorIsEmpty, KratosGeoMechanicsFastSuite)
+{
+    ProcessFactory factory;
+    factory.AddCreator("TestProcess", {});
+
+    const Parameters process_settings;
+    const auto process = factory.Create("TestProcess", process_settings);
+
+    KRATOS_EXPECT_EQ(process.get(), nullptr);
+}
+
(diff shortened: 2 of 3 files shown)


applications/GeoMechanicsApplication/custom_utilities/process_factory.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include "includes/kratos_parameters.h"
+
+
+namespace Kratos
+{
+
+ProcessFactory::ProductType ProcessFactory::Create(const std::string& rProcessClassName,
+                                                   const Parameters&  rProcessSettings) const
+{
+    auto pos = mCreatorMap.find(rProcessClassName);
```
rfaasse, line 24: This one can be const auto

applications/GeoMechanicsApplication/tests/cpp_tests/test_process_factory.cpp
```diff
@@ -0,0 +1,123 @@
...
+{
+
+KRATOS_TEST_CASE_IN_SUITE(CreateNothingWhenNoCreatorWasAddedForRequestedProcess, KratosGeoMechanicsFastSuite)
+{
+    ProcessFactory factory;
+
+    const Parameters process_settings;
+    const auto process = factory.Create("UnknownProcess", process_settings);
+
+    KRATOS_EXPECT_FALSE(process)
```
rfaasse: Minor suggestion: for readability, I think it would be clearer to explicitly say  KRATOS_EXPECT_TRUE(process == nullptr), because from the const auto, it's not directly clear a pointer is returned from 'Create')

applications/GeoMechanicsApplication/custom_utilities/process_factory.cpp
```diff
@@ -0,0 +1,38 @@
...
+        return nullptr;
+    }
+
+    return pos->second ? pos->second(rProcessSettings) : nullptr;
+}
+
+void ProcessFactory::AddCreator(const std::string&                            rProcessClassName,
+                                std::function<ProductType(const Parameters&)> Creator)
+{
+    mCreatorMap[rProcessClassName] = std::move(Creator);
```
rfaasse, line 35: I'm not sure if it's feasible, but for the key, we might be able to use the Process::Info() function, such that the key is always in accordance with the process (unless we need the flexibility to have custom names/keys)

applications/GeoMechanicsApplication/custom_utilities/process_factory.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+ProcessFactory::ProductType ProcessFactory::Create(const std::string& rProcessClassName,
+                                                   const Parameters&  rProcessSettings) const
+{
+    auto pos = mCreatorMap.find(rProcessClassName);
+    if (pos == mCreatorMap.end()) {
+        return nullptr;
+    }
+
+    return pos->second ? pos->second(rProcessSettings) : nullptr;
```
rfaasse, line 29: Just to verify my understanding: only here this unique_ptr is created, by calling the function that is defined when adding the creator right? I have not used functions in a map before, that's really interesting!

applications/GeoMechanicsApplication/tests/cpp_tests/test_process_factory.cpp
```diff
@@ -0,0 +1,123 @@
...
+    const auto bar_process = factory.Create("BarProcess", bar_process_settings);
+
+    KRATOS_EXPECT_TRUE(dynamic_cast<FooProcess*>(foo_process.get()))
+    KRATOS_EXPECT_TRUE(dynamic_cast<BarProcess*>(bar_process.get()))
+}
+
+KRATOS_TEST_CASE_IN_SUITE(CreatorReceivesGivenSettingsWhenMakingNewProcess, KratosGeoMechanicsFastSuite)
+{
+    ProcessFactory factory;
+    const Parameters* received_settings = nullptr;
```
rfaasse, line 88: In my opinion, this const is a bit misleading, since the pointer itself is changed later on. It's fine like this as well (especially since the scope of the test is nice and small).

rfaasse (dismissed):
Nice and clean way of implementing the factory (and clear tests)! I have a few minor suggestions/questions, but none of them blocking.

applications/GeoMechanicsApplication/custom_utilities/process_factory.cpp
avdg81: To the best of my knowledge it's rather unusual to use `const auto` for iterators (just `auto` is fine). The `const`ness is normally expressed by the kind of iterator: either an `iterator` or a `const_iterator` (they are different classes). Which one `find` returns depends on the mutability of the underlying map. For this reason, I feel it's better to not follow your suggestion.

applications/GeoMechanicsApplication/tests/cpp_tests/test_process_factory.cpp
avdg81: I see your point. I would, however, slightly modify your suggestion and write it as `KRATOS_EXPECT_EQ(process.get(), nullptr);`. For consistency, I will also propagate this change to some of the other unit tests in this file.

applications/GeoMechanicsApplication/custom_utilities/process_factory.cpp
avdg81: I have opted to give the client of the factory some flexibility with respect to the adopted names and to not enforce the result returned by `Process::Info()`.

applications/GeoMechanicsApplication/custom_utilities/process_factory.cpp
avdg81: Yes, only when the given process name is in the map and when the associated creator is not empty, we will create a `unique_ptr` that owns a `Process` instance. Otherwise, we get a `unique_ptr` that doesn't own anything.

What I like about this approach with the map (which I learned from a former colleague) is that it there are no changes to factory needed when you want to support other kinds of processes (even ones that don't exist yet). Just add them through `AddCreator` and you're ready to use it. A good example of the Open/Closed Principle if you'd ask me.

applications/GeoMechanicsApplication/tests/cpp_tests/test_process_factory.cpp
avdg81: My intention is to express that I just want to observe the object pointed to, not to modify it through this pointer. I feel that every C++ programmer should understand the difference between `const X*`, `X* const` and `const X* const`.

rfaasse: Agreed

rfaasse (approved):

PR #11557, opened 2023-09-13
[OptimizationApp] Marking test a expected failure until fixed

roigcarlo:
**📝 Description**
Marking a problematic test as expected fail until fixed.

@avdg81 @SADPR  This should get rid of the failing CI tests.

@sunethwarna I marked it as expected failure until fixed as is holding down several PR.

PR #11558, opened 2023-09-13
[GeoMechanicsApplication] Modified pressure points location for different order element.

### LINKED ISSUE #11484: [GeoMechanicsApplication] 15 Noded element leads to irregular stresses
**Description**
When solving deformation-pressure cases, the stress field will be irregular.
By removing the pressure load, this irregularity disappear.
In my opinion, it comes from the link between deformation and pressure elements when different order elements are applied. The arrangment of points for the pressure seems to be incorrect in 
"Kratos\applications\GeoMechanicsApplication\custom_elements\small_strain_U_Pw_diff_order_element.cpp" 
lines 219 and 222

**Scope**
- GeoMechanicsApplication in SmallStrainUPwDiffOrderElement2D15N

**To Reproduce**
[Kratos_Plaxis.zip](https://github.com/KratosMultiphysics/Kratos/files/12357517/Kratos_Plaxis.zip)
Can't visualize by GiD+ as it is not supported. Here we visualize it by Plaxis.

**Expected behavior**
MicrosoftTeams-image
The field has to be horizontally uniform as a uniform vertical load is applied. 

**Environment**
- OS: Windows 10
- Branch: master
- Python 3.10

mnabideltares:
Done for 10 and 15 noded geometries to match the 6 and 10 noded reduced element for pressure.

**📝 Description**
The pressure for 15 noded element is solved on 10 noded element because for different order elements the pressure must be solved on an element with one order lower. 
for 6 noded element, the pressure is solved on 3 noded, which is geometrically compitable, because the 3 noded elemet is a subset of 6 noded element. 
However, it is not the case for 15 and 10 noded elements. The location of the points on 10 noded element does not match the 15 noded ones.
For 10 and 6 noded elements, the same issue exist as 6 noded element is not a subset of 10 node element.
Here an interpolation precedure is added to generate new set of points

**🆕 Changelog**
- Use interpolation for the location of points for pressure.

### DIFF
--- a/applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
+++ b/applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
@@ -17,6 +17,10 @@
 #include "geometries/quadrilateral_2d_4.h"
 #include "geometries/tetrahedra_3d_4.h"
 #include "geometries/hexahedra_3d_8.h"
+#include "geometries/point.h"
+#include "geometries/geometry.h"
+#include "integration/line_gauss_legendre_integration_points.h"
+#include "utilities/integration_utilities.h"
 
 // Application includes
 #include "custom_elements/small_strain_U_Pw_diff_order_element.hpp"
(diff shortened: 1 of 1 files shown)


roigcarlo:
Seems this Pr is also affected by #11557. Merge it once its in the master pls

mnabideltares:
Modified pressure points with interpolations for pressure, as they are not at the same locations for 10 and 15 nodes elements. 
However, we still see incorrect visualization in Plaxis interface.
Then we found the visualization is well correct when we used 15 nodes elements (with proper numbering for integration points), but not as different order element. 
Whether the Plaxis interface visualization for different order element, needs to be investigated. 
It is more probably that the interpolation of pressure in Kratos (for difforder) needs to be adjusted, because the pressure finally gets interpolated back to the 15 noded element,

indigocoral:
We'll close this pull request for now but the work remains in the branch.

PR #11559, opened 2023-09-13
Revert "Bump actions/checkout@v3 to @v4"

roigcarlo:
Reverts KratosMultiphysics/Kratos#11552

Seems that Fulldebug/Gcc workflows are randomly failing. The reason is not clear to me but it seems to happen since the bumped checkout/actions, so I will revert the change until I can assure its safe.


roigcarlo:
Disclaimer, not sure this is the problem, but just in case.

philbucher (approved):

roigcarlo:
Disclaimer, not sure this is the problem, but just in case.

roigcarlo:
Seems that Pr relying on v3 have the same problem

PR #11562, opened 2023-09-14
[GeoMechanicsApplication] Improved reading of input files from settlement workflow

### LINKED ISSUE #11561: [GeoMechanicsApplication] Read `mdpa` file only once when running a settlement workflow
So far, when a stage was run using the settlement workflow (in C++), it always attempted to read the given `mdpa` file. Although the set of input files worked using the Python workflow, it raised an error when the custom workflow was used (about already existing elements). To prevent that, we now follow the same approach as in our Python workflow. That is, when the model part corresponding to the given model part name already exists, don't attempt to create a new one and also don't attempt to read the `mdpa` file. With these changes, the Python workflow and the custom (C++) workflow are more consistent.

avdg81:
**📝 Description**
So far, when a stage was run using the settlement workflow (in C++), it always attempted to read the given `mdpa` file. Although the set of input files worked using the Python workflow, it raised an error when the custom workflow was used (about already existing elements). To prevent that, we now follow the same approach as in our Python workflow. That is, when the model part corresponding to the given model part name already exists, don't attempt to create a new one and also don't attempt to read the `mdpa` file. With these changes, the Python workflow and the custom (C++) workflow are more consistent.

**🆕 Changelog**
- Use `std::filesystem::path`s rather than strings when handling file paths. This implicitly addresses problems related to path separators on different platforms. Use "generic strings" when the API expects a string rather than a path object.
- Only attempt to read from a material file when one is specified in the `ProjectParameters.json` file.
- The approach for reading of `.mdpa` files in `KratosGeoSettlement` is now more in line with how this is handled by the Python code. We first check whether the given model part name corresponds to an existing model part. If not, we create a new one and populate it using the data from the `.mdpa` file.

### DIFF
--- a/applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
+++ b/applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
@@ -31,8 +31,8 @@ KratosGeoSettlement::KratosGeoSettlement()
     }
 }
 
-int KratosGeoSettlement::RunStage(const std::string&                      rWorkingDirectory,
-                                  const std::string&                      rProjectParametersFileName,
+int KratosGeoSettlement::RunStage(const std::filesystem::path&            rWorkingDirectory,
+                                  const std::filesystem::path&            rProjectParametersFile,
                                   const std::function<void(const char*)>& ,
                                   const std::function<void(double)>&      ,
                                   const std::function<void(const char*)>& ,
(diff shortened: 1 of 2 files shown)


applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
```diff
@@ -31,44 +31,60 @@ KratosGeoSettlement::KratosGeoSettlement()
...
-    model_part.SetBufferSize(2);
-    KRATOS_INFO("KratosGeoSettlement") << "Created a model part" << std::endl;
+    if (const auto model_part_name = project_parameters["solver_settings"]["model_part_name"].GetString();
+        !mModel.HasModelPart(model_part_name)) {
+        auto& model_part = AddNewModelPart(model_part_name);
+        const auto mesh_file_name = project_parameters["solver_settings"]["model_import_settings"]["input_filename"].GetString();
+        ReadModelFromFile(rWorkingDirectory / mesh_file_name, model_part);
+    }
+
+    if (project_parameters["solver_settings"].Has("material_import_settings")) {
```
rfaasse, line 54: Just to double check: could it be that the material_import_settings are initialized empty, so there is no materials_filename (or that's empty) and could that lead to issues further down the line?

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
```diff
@@ -31,44 +31,60 @@ KratosGeoSettlement::KratosGeoSettlement()
...
+        InputUtilities::AddMaterialsFrom(material_file_path.generic_string(), mModel);
+        KRATOS_INFO("KratosGeoSettlement") << "Read the materials from " << material_file_path << std::endl;
+    }
+
+    return 0;
+}
+
+ModelPart& KratosGeoSettlement::AddNewModelPart(const std::string& rModelPartName)
+{
+    auto& result = mModel.CreateModelPart(rModelPartName);
```
rfaasse, line 66: Are we sure that none of these steps need to be 'redone' (i.e. are there any steps here that change something that's not persistent in between steps?)

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
```diff
@@ -31,44 +31,60 @@ KratosGeoSettlement::KratosGeoSettlement()
     }
 }
 
-int KratosGeoSettlement::RunStage(const std::string&                      rWorkingDirectory,
-                                  const std::string&                      rProjectParametersFileName,
+int KratosGeoSettlement::RunStage(const std::filesystem::path&            rWorkingDirectory,
```
rfaasse, line 34: Would it be doable to add a UT for this behavior?

rfaasse (dismissed):
Looks good to me! I have a couple of questions, but they are mainly to verify my understanding of the flow

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
avdg81: You are right: in this custom workflow we barely do any input checking. That is a fair point that we may need to investigate further. (If we feel we should improve in that area, we should also revisit `dgeoflow`, since it is similar in its set up.)

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
avdg81: Yes, I think we can be rather confident that we need to carry out these actions only once per model part. If we come across a case where this doesn't apply, we probably also need to modify the corresponding Python workflow.

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
avdg81: A legitimate question, but I'm afraid that the answer is that it will be rather difficult to add a unit test for this behavior. We might have a look together and investigate the possibilities that we see.

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
rfaasse: As discussed, it's out of scope to tackle this now and the extra checks added in this PR suffice for the current purposes.

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
rfaasse: As discussed, we'll move adding the UT to a separate PR, since it requires refactoring some of the implementation first.

rfaasse (approved):

PR #11567, opened 2023-09-15
[GeoMechanicsApplication] dgeosettlement refactoring for testability

### LINKED ISSUE #11572: Refactor KratosGeoSettlement to be able to add unit tests
The RunStage function in the new DGeoSettlement workflow is not unit-tested due to dependencies on IO operations (needing the MaterialParameters.json, ProjectParameters.json etc to execute the function). 

This task aims to extract the IO functionality via an interface, so they can be stubbed for the unit-test. This improves testability and ensures unit-tests only test one class, without depending on other classes.

rfaasse:
**📝 Description**
The RunStage function in the new DGeoSettlement workflow is not unit-tested due to dependencies on IO operations (needing the MaterialParameters.json, ProjectParameters.json etc to execute the function).

This PR extracts the IO functionality via an interface, so they can be stubbed for the unit-test. This improves testability and ensures unit-tests only test one class, without depending on other classes.

**🆕 Changelog**
- An InputUtilityInterface is created, which has functions to read project parameters, MaterialSettings and mesh info. This interface is implemented in custom_workflows with the actual IO functionality and passed on in the KratosGeoSettlement constructor.
- Unit tests are created for the KratosGeoSettlement, which is now easier to do, due to the interface extraction
- A stub is created for the InputUtilityInterface, which is used in the new KratosGeoSettlement unit tests.


applications/GeoMechanicsApplication/custom_utilities/input_utility_stub.cpp
```diff
@@ -0,0 +1,46 @@
+// KRATOS___
```
rfaasse: I'll move this one to the test folder (reminder to self)

applications/GeoMechanicsApplication/tests/cpp_tests/test_dgeosettlement.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include "testing/testing.h"
+#include "custom_workflows/dgeosettlement.h"
+#include "custom_utilities/input_utility_stub.h"
+
+using namespace Kratos;
+
+namespace Kratos::Testing {
+
+KRATOS_TEST_CASE_IN_SUITE(CreatingKratosGeoSettlementDoesNotThrow, WorkInProgress) {
```
rfaasse: Reminder to change to the correct suite before merge

applications/GeoMechanicsApplication/custom_utilities/input_utility.h
```diff
@@ -0,0 +1,27 @@
...
+
+#include "interface_input_utility.h"
+
+namespace Kratos {
+
+class InputUtility : public InterfaceInputUtility {
+public:
+    [[nodiscard]] Parameters ProjectParametersFrom(const std::string& rProjectFilePath) const override;
+    void ReadModelFromFile(const std::filesystem::path& rModelPartFilePath, ModelPart& rModelPart) const override;
+    void AddMaterialsFrom(const std::string& rMaterialFilePath, Model& rModel) const override;
```
avdg81: I would suggest to make the member function names more consistent. Two out of three end with `From`, whereas one ends with `FromFile`. Shall we make them all end with `FromFile`?

Also, I noticed that `ReadModelFromFile` receives a `const std::filesystem::path&`, whereas the other two member functions use `const std::string&` to pass file paths. I would prefer to use `const std::filesystem::path&` in all of these signatures, since it is clearer what to expect.

applications/GeoMechanicsApplication/custom_utilities/input_utility.cpp
```diff
@@ -0,0 +1,44 @@
...
+}
+
+void InputUtility::ReadModelFromFile(const std::filesystem::path& rModelPartFilePath,
+                                            Kratos::ModelPart& rModelPart) const
+{
+    // Note that the file extension of the model part file must be excluded, since that is automatically appended by the
+    // constructor of class ModelPartIO
+    ModelPartIO reader{rModelPartFilePath.generic_string()};
+    reader.ReadModelPart(rModelPart);
+    KRATOS_INFO("KratosGeoSettlement") << "Read the mesh data from " << rModelPartFilePath << std::endl;
```
avdg81: I would either remove this line or pass a different string to `KRATOS_INFO`, e.g. `"InputUtility::ReadModelFromFile"`.

applications/GeoMechanicsApplication/custom_utilities/input_utility.cpp
```diff
@@ -0,0 +1,44 @@
...
+Parameters InputUtility::ProjectParametersFrom(const std::string& rProjectFilePath) const {
+    std::ifstream t(rProjectFilePath);
+    std::stringstream buffer;
+    buffer << t.rdbuf();
+    Parameters projFile{buffer.str()};
+    return projFile;
+}
+
+void InputUtility::ReadModelFromFile(const std::filesystem::path& rModelPartFilePath,
+                                            Kratos::ModelPart& rModelPart) const
```
avdg81: I prefer to align the type names and parameter names in a column-wise fashion, i.e.:
```suggestion
void InputUtility::ReadModelFromFile(const std::filesystem::path& rModelPartFilePath,
                                     Kratos::ModelPart&           rModelPart) const
```

(Perhaps is this something that we should add to our own coding conventions which supplement the Kratos Style Guide.)

applications/GeoMechanicsApplication/custom_workflows/custom_workflow_factory.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include "custom_utilities/input_utility.h"
+#include <memory>
+
+namespace Kratos
+{
+
+KratosGeoSettlement* CustomWorkflowFactory::CreateKratosGeoSettlement()
+{
+    auto inputUtility = std::make_unique<InputUtility>();
+    return new KratosGeoSettlement(std::move(inputUtility));
```
avdg81: There is no need to use `std::move` if you would create a temporary:
```suggestion
    return new KratosGeoSettlement{std::make_unique<InputUtility>()};
```

Also, I prefer to use curly braces when constructing a new object.

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
```diff
@@ -11,15 +11,14 @@
...
-#include "includes/model_part_io.h"
-
+#include "custom_utilities/interface_input_utility.h"
 
 namespace Kratos
 {
 
-KratosGeoSettlement::KratosGeoSettlement()
+KratosGeoSettlement::KratosGeoSettlement(std::unique_ptr<InterfaceInputUtility> pInputUtility) :
+        mpInputUtility{std::move(pInputUtility)}
```
avdg81, line 21: You may want to check in the constructor's body whether `mpInputUtility` is not empty. If it is (e.g. when default-constructing a `std::unique_ptr`), the class won't function properly (because there is no input utility object pointed to). As such, throwing an exception in that case might be a good thing to do.

applications/GeoMechanicsApplication/tests/cpp_tests/input_utility_stub.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+
+#include "input_utility_stub.h"
+
+namespace Kratos {
+
+InputUtilityStub::InputUtilityStub(const std::string& rParameterJsonString) :
+    mParameterJsonString{rParameterJsonString}
+{
+
```
avdg81: I would delete this blank line.

applications/GeoMechanicsApplication/tests/cpp_tests/input_utility_stub.cpp
```diff
@@ -0,0 +1,49 @@
...
+}
+
+Parameters InputUtilityStub::ProjectParametersFrom(const std::string& rProjectFilePath) const
+{
+    Parameters result(mParameterJsonString);
+
+    return result;
```
avdg81: I would change this to the following one-liner:
```suggestion
    return Parameters{mParameterJsonString};
```

applications/GeoMechanicsApplication/tests/cpp_tests/input_utility_stub.cpp
```diff
@@ -0,0 +1,49 @@
...
+}
+
+void InputUtilityStub::ReadModelFromFile(const std::filesystem::path& rModelPartFilePath, ModelPart& rModelPart) const
+{
+    mNumberOfReadCalls++;
```
avdg81: Prefer pre-increment operators over post-increment operators (the latter ones are more efficient), so:
```suggestion
    ++mNumberOfReadCalls;
```

applications/GeoMechanicsApplication/tests/cpp_tests/input_utility_stub.cpp
```diff
@@ -0,0 +1,49 @@
...
+}
+
+void InputUtilityStub::AddMaterialsFrom(const std::string& rMaterialFilePath, Model& rModel) const
+{
+    mNumberOfMaterialCalls++;
```
avdg81: My previous comment also applies here:
```suggestion
    ++mNumberOfMaterialCalls;
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_dgeosettlement.cpp
```diff
@@ -0,0 +1,118 @@
...
+                                            "{"
+                                            "\"input_type\": \"mdpa\","
+                                            "\"input_filename\": \"mesh_stage1\""
+                                            "},"
+                                            "\"material_import_settings\": "
+                                            "{"
+                                            "\"materials_filename\": \"MaterialParameters1.json\""
+                                            "}"
+                                            "}"
+                                            "}"; // these have material_import settings
```
avdg81: I would use a [raw string literal](https://en.cppreference.com/w/cpp/language/string_literal) here to avoid the need to escape double quotes.

applications/GeoMechanicsApplication/tests/cpp_tests/test_dgeosettlement.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+
+void RunStage(KratosGeoSettlement& rSettlement) {
+    rSettlement.RunStage("",
+                         "",
+                         [](const char *) {}, // kept empty as a stub method
+                        [](const double) {}, // kept empty as a stub method
+                        [](const char *) {}, // kept empty as a stub method
+                        []() { return true; });
```
avdg81: I think `return false;` is more appropriate here (you don't want to cancel the operation).

applications/GeoMechanicsApplication/tests/cpp_tests/test_dgeosettlement.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+void ExpectNumberOfReadCallsIsEqualToOne(const KratosGeoSettlement &rSettlement) {
+    const auto input_utility_from_settlement = dynamic_cast<const InputUtilityStub*>(rSettlement.GetInterfaceInputUtility());
+    KRATOS_EXPECT_NE(input_utility_from_settlement, nullptr);
+    KRATOS_EXPECT_EQ(input_utility_from_settlement->NumberOfReadCalls(), 1);
+}
+
+void ExpectNumberOfMaterialCallsEqualTo(const int expectedNumberOfMaterialCalls, const KratosGeoSettlement& rSettlement) {
```
avdg81: I would remove `const` from the first parameter:
```suggestion
void ExpectNumberOfMaterialCallsEqualTo(int expectedNumberOfMaterialCalls, const KratosGeoSettlement& rSettlement) {
```

As explained in Section [Con.1: By default, make objects immutable](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#con1-by-default-make-objects-immutable) of the C++ Core Guidelines, it is a bit pedantic to declare parameters that are passed by value `const`:

> Function parameters passed by value are rarely mutated, but also rarely declared `const`. To avoid confusion and lots of false positives, don’t enforce this rule for function parameters.

applications/GeoMechanicsApplication/tests/cpp_tests/test_dgeosettlement.cpp
```diff
@@ -0,0 +1,118 @@
...
+    KRATOS_EXPECT_EQ(input_utility_from_settlement->NumberOfMaterialCalls(), expectedNumberOfMaterialCalls);
+}
+
+KRATOS_TEST_CASE_IN_SUITE(CreatingKratosGeoSettlementDoesNotThrow, KratosGeoMechanicsFastSuite) {
+    auto input_utility = std::make_unique<InputUtilityStub>(parameter_json_settings);
+
+    bool has_thrown = false;
+    try
+    {
+        KratosGeoSettlement settlement(std::move(input_utility));
```
avdg81: To avoid the `std::move` here, creating a temporary object would be sufficient:
```suggestion
        KratosGeoSettlement settlement{std::make_unique<InputUtilityStub>(parameter_json_settings)};
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_dgeosettlement.cpp
```diff
@@ -0,0 +1,118 @@
...
+    catch (...)
+    {
+        has_thrown = true;
+    }
+
+    KRATOS_EXPECT_FALSE(has_thrown); // No other way to check that the constructor does not throw
+}
+
+KRATOS_TEST_CASE_IN_SUITE(RunStageMakesRelevantCallsOnce, KratosGeoMechanicsFastSuite) {
+    auto input_utility = std::make_unique<InputUtilityStub>(parameter_json_settings); // invalid after the constructor
```
avdg81: The comment provides a good reason to use a temporary here :-) (i.e. avoid the `std::move`).

applications/GeoMechanicsApplication/tests/cpp_tests/test_dgeosettlement.cpp
```diff
@@ -0,0 +1,118 @@
...
+                                                                               "\"solver_settings\":"
+                                                                               "{"
+                                                                               "\"model_part_name\":\"test\","
+                                                                               "\"model_import_settings\":"
+                                                                               "{"
+                                                                               "\"input_type\": \"mdpa\","
+                                                                               "\"input_filename\": \"mesh_stage1\""
+                                                                               "}"
+                                                                               "}"
+                                                                               "}";
```
avdg81: Also here, please use a raw string literal.

applications/GeoMechanicsApplication/tests/cpp_tests/test_dgeosettlement.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+    auto input_utility = std::make_unique<InputUtilityStub>(parameter_json_string_without_material_import_settings); // invalid after the constructor
+
+    KratosGeoSettlement settlement(std::move(input_utility));
```
avdg81: ... and use a temporary object to avoid `std::move` here.

avdg81 (commented):
Very nice refactoring. I have mostly minor remarks and a few suggestions.

applications/GeoMechanicsApplication/tests/cpp_tests/test_dgeosettlement.cpp
rfaasse: Agreed, I started using it in the next PR and it works way simpler, so I'll also rewrite this one

applications/GeoMechanicsApplication/tests/cpp_tests/test_dgeosettlement.cpp
rfaasse: Good point, changed

applications/GeoMechanicsApplication/tests/cpp_tests/test_dgeosettlement.cpp
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_dgeosettlement.cpp
rfaasse: Agreed, done

applications/GeoMechanicsApplication/tests/cpp_tests/test_dgeosettlement.cpp
rfaasse: It's indeed not necessary. I like to do it to show the intentions with this function, but since it only adds confusion, I removed it

applications/GeoMechanicsApplication/tests/cpp_tests/test_dgeosettlement.cpp
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_dgeosettlement.cpp
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_dgeosettlement.cpp
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/input_utility_stub.cpp
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/input_utility_stub.cpp
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/input_utility_stub.cpp
rfaasse: Agreed, forgot to clean that up after simplification, done!

applications/GeoMechanicsApplication/tests/cpp_tests/input_utility_stub.cpp
rfaasse: Done

applications/GeoMechanicsApplication/custom_workflows/custom_workflow_factory.cpp
rfaasse: Done

applications/GeoMechanicsApplication/custom_utilities/input_utility.h
rfaasse: Agreed, good catch, made it consistent now

applications/GeoMechanicsApplication/custom_utilities/input_utility.cpp
rfaasse: I don't know how useful this logging is. Until I know more, I would propose to keep it (unless you know it would never help anyone), so I changed the logging like you suggested and didn't remove it

applications/GeoMechanicsApplication/custom_utilities/input_utility.cpp
rfaasse: I agree it looks better, but when renaming anything, the alignment will be wrong again, so that's why I generally don't spend extra time for the 2nd column alignment, but keep it like it is after this commit. Let me know what you think 👍

rfaasse:
For documentation purposes: the 5 newly added tests run within 40 ms, since they're not dependent on any IO

image

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
rfaasse: Done, good point

applications/GeoMechanicsApplication/custom_utilities/file_input_utility.h
```diff
@@ -0,0 +1,27 @@
...
+//                   Richard Faasse
+//
+
+#pragma once
+
+#include "input_utility.h"
+
+namespace Kratos {
+
+class FileInputUtility : public InputUtility {
```
rfaasse, line 20: I renamed it like this, let me know what you think!

applications/GeoMechanicsApplication/custom_utilities/input_utility.h
```diff
@@ -0,0 +1,30 @@
...
+// ((____/ / ((____   ((___/ /  MECHANICS
+//
+//  License:         geo_mechanics_application/license.txt
+//
+//  Main authors:    Richard Faasse
+//
+
+#pragma once
+
+#include <string>
```
avdg81: Shouldn't we now include `<filesystem>` rather than `<string>`?
```suggestion
#include <filesystem>
```

applications/GeoMechanicsApplication/custom_utilities/input_utility.cpp
avdg81: I guess that these were just debug messages, to see which parts of the code are executed and when. Your change will do. Thanks.

applications/GeoMechanicsApplication/custom_utilities/input_utility.cpp
avdg81: I suggest we should first come up with coding conventions in addition to the Kratos Style Guide before making this kind of stylistic changes.

applications/GeoMechanicsApplication/custom_utilities/file_input_utility.h
avdg81: Yes, I'm with you. Nice improvement!

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
```diff
@@ -11,17 +11,18 @@
 //
 #include "dgeosettlement.h"
 #include "input_output/logger.h"
-#include "custom_utilities/input_utilities.h"
 #include "utilities/variable_utils.h"
-#include "includes/model_part_io.h"
-
+#include "custom_utilities/input_utility.h"
+#include <iostream>
```
avdg81: I don't think we need this include, do we?

avdg81 (dismissed):

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
rfaasse: Good catch, removed

applications/GeoMechanicsApplication/custom_utilities/input_utility.h
rfaasse: Removed

avdg81 (approved):
Nice piece of work!

PR #11571, opened 2023-09-18
[Hotfix][Core] Quad element size fix

rubenzorrilla:
**📝 Description**
Quadrilateral `AverageElementSize` formula in `ElementSizeCalculator`  is only valid for rectangular quadrilaterals. Changes in here make it valid for any quadrilateral.


loumalouomega:
BTW, this means that the dreivative is also wrong

rubenzorrilla:
> BTW, this means that the dreivative is also wrong

It is.

sunethwarna:
Ok, i fixed the derivatives @rubenzorrilla :)

RiccardoRossi (approved):

PR #11574, opened 2023-09-19
[Core] Change Registry GetValue

roigcarlo:
**📝 Description**
- Registry now expects for items to be registered with their bases
- `GetValue<T>` now expects you to obtain the base value (Like current KratosComponents)
- `GetValueAs<T,C>` now makes the cast so you can retrieve the actual downcasted object, if possible

Example:

```C++
class Process {};
class MyProcess : public Process {};
class MyBadProcess {};

ADD_REGISTRY_PROTOTYPE("Key", Process, MyProcess)

auto p = Registry::GetValue<Process>("Key.MyProcess.Prototype")  // Correct, p is of type Process
auto p = Registry::GetValue<MyProcess>("Key.MyProcess.Prototype")  // Compilation Error
auto p = Registry::GetValueAs<Process, MyProcess>("Key.MyProcess.Prototype")  // Correct, p is of type MyProcess
auto p = Registry::GetValueAs<Process, MyBadProcess>("Key.MyProcess.Prototype")  // Compilation Error (Process is not a base of MyBadProcess)
```

Test to check that GetValueAs with non derived classes cannot be compiled is super ugly and involves dark templates, so left it behind.

### DIFF
--- a/kratos/tests/cpp_tests/sources/test_registry_access.cpp
+++ b/kratos/tests/cpp_tests/sources/test_registry_access.cpp
@@ -1,0 +1,50 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
+//
+//  Main authors:    Pooyan Dadvand
+//
+//
+
+// System includes
+#include <numeric>
+
+// External includes
+
+// Project includes
+#include "testing/testing.h"
+#include "includes/expect.h"
+#include "includes/registry.h"
+#include "processes/process.h"
+#include "processes/output_process.h"
+
+namespace Kratos::Testing {
+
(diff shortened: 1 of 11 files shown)


kratos/tests/cpp_tests/sources/test_registry_access.cpp
```diff
@@ -0,0 +1,50 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
+//
+//  Main authors:    Pooyan Dadvand
```
rubenzorrilla:
```suggestion
//  Main authors:    Carlos Roig
```

rubenzorrilla (commented):
Makes perfect sense for me. However I'm not sure how this propagates to the Python layer (shouldn't we edit the template arguments of the  `GetProcess` and `GetOperation` methods?).

As a minor comment, I'd put the access test together with the already existent ones (i.e. within `test_registry.cpp`.

rubenzorrilla (approved):

PR #11575, opened 2023-09-19
[CompressiblePotentialFlowApp] Adding Line Search Strategy to settings

Marco1410:
**📝 Description**
This PR adds the use of the line search strategy in the application.

The use of this method provides better stability in nonlinear problems even in those with strong nonlinearities as occurs in a transonic flow with strongs shocks.

This also allows obtaining better results in cases in which it would not converge using only Newton Raphson.

**🆕 Changelog**

- Changed "potential_flow_solver.py"

### DIFF
--- a/applications/CompressiblePotentialFlowApplication/python_scripts/potential_flow_solver.py
+++ b/applications/CompressiblePotentialFlowApplication/python_scripts/potential_flow_solver.py
@@ -160,7 +160,6 @@ class PotentialFlowSolver(FluidSolver):
             "absolute_tolerance": 1e-12,
             "compute_reactions": false,
             "reform_dofs_at_each_step": false,
-            "calculate_solution_norm": false,
             "linear_solver_settings": {
                 "solver_type": "amgcl"
             },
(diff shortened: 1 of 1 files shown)


marcnunezc (commented):
Looks good to me!

I would just suggest to check if any of the functions
 `_CreatePotentialFlowLinearStrategy`
 `_CreatePotentialFlowNewtonRaphsonStrategy`
 `_CreatePotentialFlowLineSearchStrategy`

can be derived from fluid_solver.py to avoid code duplication, for instance:

https://github.com/KratosMultiphysics/Kratos/blob/5a16407fe4dc6ca18ac8bf9f428ef25ebd112198/applications/FluidDynamicsApplication/python_scripts/fluid_solver.py#L409-L422

Marco1410:
Hi @marcnunezc, thanks for the comment.

For the Newton Raphson I see that it would not be a problem, the line search I think is not included in the fluid_solver.py and for the linear strategy I see that has the option to calculate the norm set to false by default.

Screenshot from 2023-09-20 09-50-17

I could use the Newton Raphson and the linear strategy  but I should also remove that option from the settings. I don't know if we need to be able to configure this option.

Screenshot from 2023-09-20 09-53-28

I think that would be the only problem.

marcnunezc:
> Hi @marcnunezc, thanks for the comment.
> 
> For the Newton Raphson I see that it would not be a problem, the line search I think is not included in the fluid_solver.py and for the linear strategy I see that has the option to calculate the norm set to false by default.
> 
> 
> I could use the Newton Raphson and the linear strategy but I should also remove that option from the settings. I don't know if we need to be able to configure this option.
> 
> 
> I think that would be the only problem.

Ok! Then I would try to use those functions from the fluid_solver.py, so only the line_search should be on the potential solver. 

Regarding the `calculate_solution_norm ` setting, I think we are always using it as `false`, so it will not affect us.  Actually it seems only the potential solver had a setting for it (it was there from the first commit of the file!). The rest of solvers that use it have this option hardcoded.

Marco1410:
Perfect, I already uploaded the changes. 

So I think it's ready.

marcnunezc (approved):
Great! thanks

PR #11577, opened 2023-09-19
[RomApp] Adding candidate elements (and/or conditions) to HROM

Rbravo555:
**📝 Description**
This PR will allow to make the element selector algorithm choose from a specific candidate set of elements or conditions. In case choosing the provided candidates the algorithm cannot converge, then the candidate set is expanded to include the rest of the elements and conditions in the computing model part.

It is also possible that by using the provided candidate elements, no solution can be achived that results in all weights positive. In such a case, the algorithm will run again from scratch (ignoring the provided candidates).

**🆕 Changelog**
- Added two function to the RomAuxiliaryUtilities to retrieve the ids from elements and conditions in a modelpart
- Added candidate set to Empirical Cubature Method
- Modified the HromTraining utility to account for a list of model parts to include as candidates for the ECM
- Modified the RomManager to account for a list of model parts to include as candidates for the ECM
- Expanded ECM test

### DIFF
--- a/applications/RomApplication/custom_utilities/rom_auxiliary_utilities.cpp
+++ b/applications/RomApplication/custom_utilities/rom_auxiliary_utilities.cpp
@@ -342,7 +342,7 @@ std::vector<IndexType> RomAuxiliaryUtilities::GetNodalNeighbouringElementIdsNotI
 {
     std::vector<IndexType> new_element_ids;
     const auto& r_elem_weights = rHRomWeights.at("Elements");
-    
+
     FindGlobalNodalEntityNeighboursProcess<ModelPart::ElementsContainerType> find_nodal_elements_neighbours_process(rModelPart);
     find_nodal_elements_neighbours_process.Execute();
 
@@ -373,7 +373,7 @@ std::vector<IndexType> RomAuxiliaryUtilities::GetElementIdsNotInHRomModelPart(
 
     for (const auto& r_elem : rModelPartWithElementsToInclude.Elements()) {
         IndexType element_id = r_elem.Id();
-        
+
         // Check if the element is already added
         if (r_elem_weights.find(element_id - 1) == r_elem_weights.end()) {
             new_element_ids.push_back(element_id - 1);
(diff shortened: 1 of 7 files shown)


applications/RomApplication/python_scripts/hrom_training_utility.py
```diff
@@ -172,9 +201,11 @@ def __GetHRomTrainingDefaultSettings(cls):
             "projection_strategy": "galerkin",
             "include_conditions_model_parts_list": [],
             "include_elements_model_parts_list": [],
+            "initial_candidate_elements_model_part_list" : [],
+            "initial_candidate_conditions_model_part_list" : [],
             "include_nodal_neighbouring_elements_model_parts_list":[],
             "include_minimum_condition": false,
-            "include_condition_parents": true
+            "include_condition_parents": false
```
SADPR, line 209: In the worst case, it won't use the info of the parent element. Still, it is generally more likely to fail because we need an associated parent element (e.g. Convection-Diffusion, Fluid-Dynamics, CompressiblePotenital).

SADPR: image

SADPR:
Tested (evidence):
ECM without passing the list of candidates (right wall):
No_Candidates

ECM passing the list of candidates (right wall):
Candidates

SADPR (approved):
LGTM.

applications/RomApplication/custom_utilities/rom_auxiliary_utilities.cpp
```diff
@@ -403,6 +404,28 @@ std::vector<IndexType> RomAuxiliaryUtilities::GetConditionIdsNotInHRomModelPart(
...
+std::vector<IndexType> RomAuxiliaryUtilities::GetConditionIdsInModelPart(
+    const ModelPart& rModelPart)
+{
+    std::vector<IndexType> condition_ids;
+
+    for (const auto& r_cond : rModelPart.Conditions()) {
+        condition_ids.push_back(r_cond.Id());
+    }
+    return condition_ids;
+}
```
rubenzorrilla, line 427: Just pointing that these two can be easily done in parallel with an index partition.

PR #11579, opened 2023-09-19
Geo/11578 cpp tests copy

### LINKED ISSUE #11578: [GeoMechanicsApplication] Cpp_tests not copied to application in bin folder
This is just for completeness (no computational issue)

mcgicjn2:
**📝 Description**
Updated the install build step to package *.cpp and *.h in the test folder of the bin\release\applications\GeoMechanicApplication

This means that cpp_tests are packaged with the test suite artifact on Deltares buidl server.

**🆕 Changelog**
removed exclude from *.cpp and *.h in test folder for the packaging of kratos in the install step of building.

No computational running issue, it is for completeness to package all tests so they are readable to the user


avdg81 (commented):
I'm not sure whether this is something we would like to do. However, if we would, I'd suggest to also remove the exclusion for `*.hpp` files and update the block's comment, which now reads "Install everything except sources ..." (since then it's no longer accurate).

But let's take a bit of time to think this over first. Thanks.

mcgicjn2:
I agree with the comments of @avdg81 above thus will close the PR

PR #11580, opened 2023-09-20
[MEDApp] Minor Fixes

matekelemen:
- remove the dummy definitions of `MPI_Comm` and `MPI_Info` that lead to multiple definitions. Forward declaration also fails because `med.h` indirectly includes `mpi.h` that defines these as `typedef struct`s that are incompatible with `struct`s. I haven't tried this on non-MPI builds.
- update testing macros
- delete a try-catch-rethrow block from a destructor (destructors are implicitly `noexcept`)


philbucher:
I will review tonight, sorry for the delays

applications/MedApplication/med_inc.h
```diff
@@ -14,8 +14,8 @@
 
 // Those dummy structs are necessary when the MED-library is compiled with MPI support.
 // Then the header also contains functions which use those MPI-types
-struct MPI_Comm{};
-struct MPI_Info{};
+//struct MPI_Comm;
+//struct MPI_Info;
```
philbucher: the issue here is when you compile without MPI

I am not sure how we can go about this. Maybe `ifdef`? Might be the only option we have for now, can you give it a try?

Also, how did you install the med-lib? By hand or via the package manager?

philbucher (dismissed):
We need to figure out a better solution for the MPI-issue, but thx otherwise

applications/MedApplication/med_inc.h
matekelemen: > Maybe ifdef?

Yep, I think we'll have to resort to the preprocessor.

> Also, how did you install the med-lib? By hand or via the package manager?

I installed [salome-medcoupling](https://aur.archlinux.org/packages/salome-medcoupling) (should include both the serial and OpenMPI med lib) using the system package manager.

philbucher: Ok ifdef it is (at least for now, until someone complains :D), can you please add it?

Also, can you please add how you installed the lib to the readme? I did it already for ubuntu

matekelemen:
I found that the created geometries were automatically assigned IDs (**really** high ones) that eventually lead to exceptions when trying to output the resulting `ModelPart` to MDPA via `ModelPartIO::WriteModelPart`, so I added explicit IDs for the created geometries.

The current impl obviously won't work with `ModelPart`s that are already populated, but we can worry about those later because the `Node`s are created in the same manner.

philbucher:
> I found that the created geometries were automatically assigned IDs (**really** high ones) that eventually lead to exceptions when trying to output the resulting `ModelPart` to MDPA via `ModelPartIO::WriteModelPart`, so I added explicit IDs for the created geometries.
> 
> The current impl obviously won't work with `ModelPart`s that are already populated, but we can worry about those later because the `Node`s are created in the same manner.

ok thanks that makes sense
Lets add a check in read mode if nodes already exist (I thought I had added that one already). Can you please take care of this one?

Ids of entities and mypy, my programming arch-enemies :D

philbucher (approved):

PR #11583, opened 2023-09-21
[CI] Store Nightly Artifacts

roigcarlo:
**📝 Description**
This will store nightly artifacts for ubuntu and windows. The retention time is set to 1 day, so only the last one will be available.

@mcgicjn2

**🆕 Changelog**
- Added action to upload artifacts for nightly builds


philbucher:
what is the intended usage?

roigcarlo:
@KratosMultiphysics/geomechanics wanted to take a look at the artifacts generated by the nightly pipeline. I am trying to fit them inside the 500mb that we have available.

This should upload the result into the artifacts cache, so we can download them afterwards. The problem is that I need to fit everything in just 500mb so it needs to have some kind of control on what do we upload (would be easier if I can just push all the compilation matrix)

Seems that nowadays all problems are related to disk quotas :/

mcgicjn2:
@roigcarlo From our perspective it is ok to close this PR and issue, we have made alternative arrangements to build our own nightly

PR #11591, opened 2023-09-22
[GeoMechanicsApplication] Add process factory to settlement workflow

### LINKED ISSUE #11576: [GeoMechanicsApplication] Add necessary processes to the Settlement workflow
As a developer in the Kratos core, I would like to create and instantiate processes from the project parameters, such that I can use them in the Settlement Workflow.

#### Background

This issue aims to implement the earlier created process factory (see #11556 for more details) into the KratosGeoSettlement class, to create all necessary processes, **except ApplyScalarConstraintTableProcess**, for the settlement workflow. This includes the following processes and will instantiate them to make them available for the next part of the workflow:

1. ApplyVectorConstraintTableProcess
2. SetParameterFieldProcess
3. ApplyExcavationProcess
4. ApplyK0ProcedureProcess

**_However_** for implementing the before mentioned ApplyScalarConstraintTableProcess (needed for the initial hydrostatic water pressure and pressure boundary condition), some more work needs to be done to move logic currently present in the python script to a c++ class, which is picked up in #11581 

#### Acceptance Criteria

**Given** the developer has processes defined in the ProjectParameters.json file
**When** the developer runs the settlement stage (using the `KratosGeoSettlement::RunStage` function)
**Then** the processes associated to the settlement workflow are instantiated and ready for further use in the workflow

**Given** the developer/user has processes defined which are not in the list of expected processes.
**When** the developer/user runs the settlement stage (using the `KratosGeoSettlement::RunStage` function)
**Then** these non-relevant processes raises an error message "Unexpected process (name of the process), calculation is aborted" and stops the calculation.

rfaasse:
**📝 Description**
In this PR, the functionality is added to create and instantiate processes from the project parameters in the KratosGeoSettlement class, such that they can be used in the next steps of the settlement workflow.

**🆕 Changelog**
- Added functionality to read project parameters to instantiate the relevant processes
- Added the process factory to the settlement workflow
- Added cpp unit tests for all newly added functionality


applications/GeoMechanicsApplication/custom_utilities/process_parameters.h
```diff
@@ -0,0 +1,37 @@
...
+#pragma once
+
+#include "includes/kratos_parameters.h"
+
+#include <string>
+
+namespace Kratos
+{
+
+struct ProcessParameters
```
rfaasse, line 21: @avdg81 For this one I got this SQ issue, which I don't completely understand, because I didn't implement a move or assignment operator here:
https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&types=CODE_SMELL&branch=geo%2Fadd-process-factory-to-settlement&id=KratosGeomechanics&open=AYqz1Y3GQOYbbtPOkeNJ

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
```diff
@@ -59,6 +106,22 @@ int KratosGeoSettlement::RunStage(const std::filesystem::path&            rWorki
         KRATOS_INFO("KratosGeoSettlement") << "Read the materials from " << material_file_path << std::endl;
     }
 
+    std::vector<std::reference_wrapper<Process>> process_references;
+    std::vector<std::unique_ptr<Process>> process_unique_ptrs;
```
rfaasse: @avdg81 The reference wrappers seem a bit cumbersome here. Maybe we can just pass an std::vector<std::unique_ptr<Process>>& to the TimeLoopExecutor?

applications/GeoMechanicsApplication/custom_utilities/process_parameters.h
rfaasse: As discussed, this is because Parameter is not 'no exept', which is out of scope to fix in this PR

avdg81: I believe that this issue is caused by the fact that the move operator of class `Parameters` is not `noexcept`. (Which is rather unfortunate and not in line with the C++ Core Guidelines.) Since you have a data member of type `Parameters`, the compiler cannot generate a `noexcept` move operator for your new class. If we want to resolve this, we need to correct it in the Kratos core.

applications/GeoMechanicsApplication/custom_utilities/json_process_info_parser.cpp
```diff
@@ -0,0 +1,51 @@
...
+//  Main authors:    Richard Faasse
+//
+#include "json_process_info_parser.h"
+
+namespace Kratos
+{
+
+std::vector<ProcessParameters> JsonProcessInfoParser::GetProcessList(const Parameters& rProcessParameters)
+{
+    mProcessParameters = rProcessParameters;
```
avdg81: By using data members to share data between member functions, the overall process became less clear to me. You may try to pass `rProcessParameters` as an argument to `AddProcessesForList`. Also, I would let `AddProcessesForList` return a vector of `ProcessParameters` objects (which could be empty if the given list does not exist). That would add the benefit that you can make all of these member function `static`, since you no longer need any data members. My understanding is that the aim of this class is to produce a flat list of process instances by going over three kinds of lists in the user input.

If you make these changes, you may also need to rename a few things.

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
```diff
@@ -59,6 +106,22 @@ int KratosGeoSettlement::RunStage(const std::filesystem::path&            rWorki
...
+    if (project_parameters.Has("processes"))
+    {
+        const auto processes = mpProcessInfoParser->GetProcessList(project_parameters["processes"]);
+        for (const auto& process : processes)
+        {
+            process_unique_ptrs.emplace_back(mProcessFactory->Create(process.name, process.parameters));
+        }
```
avdg81: I would extract this to a separate member function, e.g. `CreateProcesses`, which returns a `std::vector<std::unique_ptr<Process>>`.

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
```diff
@@ -59,6 +106,22 @@ int KratosGeoSettlement::RunStage(const std::filesystem::path&            rWorki
...
+        for (const auto& process_unique_ptr : process_unique_ptrs)
+        {
+            process_references.emplace_back(*process_unique_ptr);
+        }
```
avdg81: This is in essence a transformation. So we could extract that to a separate member function, too. And perhaps use `std::transform` (if that makes the code more readable). Feel free to experiment a bit first and then decide on the simplest/most readable implementation.

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
avdg81: I still prefer the reference wrappers, since I would typically expect smart pointers to be used in the API only when there is a transfer of ownership (which is not the case here). So I would suggest to keep it the way it is.

applications/GeoMechanicsApplication/custom_utilities/process_parameters.h
```diff
@@ -0,0 +1,37 @@
...
+{
+    Parameters parameters;
+    std::string name;
+
+    ProcessParameters(const Parameters& rParameters, const std::string& rName) :
```
avdg81: To make reading easier, I would put the name as the first parameter and the settings (parameters) second.

applications/GeoMechanicsApplication/tests/cpp_tests/stub_time_loop_executor.h
```diff
@@ -0,0 +1,29 @@
...
+#pragma once
+
+#include "custom_workflows/time_loop_executor.h"
+
+namespace Kratos {
+
+class StubTimeLoopExecutor : public TimeLoopExecutor
+{
+public:
+    explicit StubTimeLoopExecutor(size_t NumberOfExpectedProcesses = 0);
```
avdg81, line 22: Shouldn't the parameter type be `std::size_t` rather than just `size_t`?

applications/GeoMechanicsApplication/tests/cpp_tests/stub_time_loop_executor.h
```diff
@@ -0,0 +1,29 @@
...
+    void SetProcessReferences(const std::vector<std::reference_wrapper<Process>>& rProcessRefs) override;
+
+private:
+    size_t mNumberOfExpectedProcesses;
```
avdg81: Also here I would expect the type to be `std::size_t`.

applications/GeoMechanicsApplication/tests/cpp_tests/stub_time_loop_executor.cpp
```diff
@@ -0,0 +1,32 @@
...
+//  Main authors:    Richard Faasse
+//
+#include "stub_time_loop_executor.h"
+#include "processes/process.h"
+#include "testing/testing.h"
+
+namespace Kratos {
+
+StubTimeLoopExecutor::StubTimeLoopExecutor(size_t NumberOfExpectedProcesses) :
+        mNumberOfExpectedProcesses{NumberOfExpectedProcesses}
```
avdg81: I'm used to repeating the base class here, so:
```suggestion
StubTimeLoopExecutor::StubTimeLoopExecutor(size_t NumberOfExpectedProcesses) :
        TimeLoopExecutor(),
        mNumberOfExpectedProcesses{NumberOfExpectedProcesses}
```

But perhaps you have a good reason not to? If so, could you please share your thoughts on this?

applications/GeoMechanicsApplication/tests/cpp_tests/test_process_info_parser.cpp
```diff
@@ -0,0 +1,136 @@
...
+      ]
+    }
+    )";
+
+    // When
+    JsonProcessInfoParser parser;
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(parser.GetProcessList(process_without_parameters), "Getting a value that does not exist. entry string : Parameters")
+}
+
+KRATOS_TEST_CASE_IN_SUITE(GetProcessList_ReturnsEmptyString_WhenNoProcessesAreDefined, KratosGeoMechanicsFastSuite)
```
avdg81: I would change the phrase `EmptyString` to e.g. `EmptyList` (or `EmptyVector`) to make it consistent with the actual return type.

applications/GeoMechanicsApplication/tests/cpp_tests/test_process_info_parser.cpp
```diff
@@ -0,0 +1,136 @@
...
+{
+    JsonProcessInfoParser parser;
+    KRATOS_EXPECT_EQ(parser.GetProcessList(Parameters{}).size(), 0);
```
avdg81: I would prefer to use the `empty()` member function, i.e.:
```suggestion
    KRATOS_EXPECT_TRUE(parser.GetProcessList(Parameters{}).empty());
```

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.h
```diff
@@ -24,12 +24,17 @@
...
+class ProcessFactory;
 class InputUtility;
+class ProcessInfoParser;
+class TimeLoopExecutor;
 
 class KRATOS_API(GEO_MECHANICS_APPLICATION) KratosGeoSettlement
 {
 public:
-    explicit KratosGeoSettlement(std::unique_ptr<InputUtility> pInputUtility);
+    explicit KratosGeoSettlement(std::unique_ptr<InputUtility> pInputUtility,
```
rfaasse: Does not need explicit anymore

applications/GeoMechanicsApplication/custom_utilities/json_process_info_parser.cpp
rfaasse: I changed it, since I'm fine with either version (although we get more parameters in the function calls, but that's up for discussion). However, it cannot become a static method due to the inheritance structure

applications/GeoMechanicsApplication/custom_utilities/process_parameters.h
rfaasse: Agreed, good suggestion!

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
rfaasse: Agreed, done!

applications/GeoMechanicsApplication/tests/cpp_tests/stub_time_loop_executor.cpp
rfaasse: No specific reason. Since only the default constructor is used here from the base-class, do you still think it's necessary?

applications/GeoMechanicsApplication/tests/cpp_tests/stub_time_loop_executor.h
rfaasse: Good point

applications/GeoMechanicsApplication/tests/cpp_tests/stub_time_loop_executor.h
rfaasse: Good point

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
rfaasse: Made it simpler (since shared_ptrs can be implicitly converted into weak_ptrs

applications/GeoMechanicsApplication/tests/cpp_tests/test_process_info_parser.cpp
rfaasse: Agreed, that's more readable

applications/GeoMechanicsApplication/tests/cpp_tests/test_process_info_parser.cpp
rfaasse: Makes sense, changed

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
rfaasse: Changed it to the weak_ptr alternative we discussed. Let me know what you think!

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
```diff
@@ -59,6 +106,15 @@ int KratosGeoSettlement::RunStage(const std::filesystem::path&            rWorki
         KRATOS_INFO("KratosGeoSettlement") << "Read the materials from " << material_file_path << std::endl;
     }
 
+    std::vector<std::shared_ptr<Process>> processes = GetProcesses(project_parameters);
+    std::vector<std::weak_ptr<Process>> process_observables;
+    process_observables.insert(process_observables.end(), processes.begin(), processes.end());
+
+    if (mpTimeLoopExecutor != nullptr)
```
rfaasse: @avdg81 I added this nullptr check, because when calling the API from python now, there is no implementation yet for the time loop executor, so it will be a nullptr. If we want to test this a bit through python, we don't want it to segfault here, since it's only temporary anyway. Later on, we should move this check to the constructor, same as with the InputUtility

applications/GeoMechanicsApplication/custom_utilities/json_process_info_parser.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+std::vector<ProcessParameters> JsonProcessInfoParser::AddProcessesForList(const std::string& rProcessListName, const Parameters& rProcessParameters) const {
+    if (!rProcessParameters.Has(rProcessListName))
+    {
+        return {};
+    }
+
+    std::vector<ProcessParameters> result;
+    const auto processes_in_list = rProcessParameters[rProcessListName];
+    for (Parameters process : processes_in_list)
```
avdg81:
```suggestion
    for (Parameters process : rProcessParameters[rProcessListName])
```

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
avdg81: The check is fine with me. I would only express it slightly differently:
```suggestion
    if (mpTimeLoopExecutor)
```

applications/GeoMechanicsApplication/custom_utilities/process_factory.cpp
```diff
@@ -22,9 +22,7 @@ ProcessFactory::ProductType ProcessFactory::Create(const std::string& rProcessCl
                                                    const Parameters&  rProcessSettings) const
 {
     auto pos = mCreatorMap.find(rProcessClassName);
-    if (pos == mCreatorMap.end()) {
-        return nullptr;
-    }
+    KRATOS_ERROR_IF(pos == mCreatorMap.end()) << "Unexpected process (" << rProcessClassName << "), calculation is aborted";
```
avdg81: We have already discussed how we can make this more generic by using a callback mechanism.

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
```diff
@@ -59,6 +106,15 @@ int KratosGeoSettlement::RunStage(const std::filesystem::path&            rWorki
         KRATOS_INFO("KratosGeoSettlement") << "Read the materials from " << material_file_path << std::endl;
     }
 
+    std::vector<std::shared_ptr<Process>> processes = GetProcesses(project_parameters);
+    std::vector<std::weak_ptr<Process>> process_observables;
+    process_observables.insert(process_observables.end(), processes.begin(), processes.end());
```
avdg81: I think you can do this also in one go, like this:
```suggestion
    std::vector<std::weak_ptr<Process>> process_observables{processes.begin(), processes.end()};
```

avdg81 (commented):
Looks good to me. Just a few minor comments.

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
rfaasse: Even shorter, nice!

applications/GeoMechanicsApplication/custom_utilities/process_factory.cpp
rfaasse: Done

applications/GeoMechanicsApplication/custom_utilities/json_process_info_parser.cpp
rfaasse: Changed

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
rfaasse: Changed

avdg81 (dismissed):
Looks good to me. Thanks!

applications/GeoMechanicsApplication/custom_utilities/process_parameters.h
avdg81: Agreed, this issue is outside the scope of this PR.

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
avdg81: The solution looks good to me.

applications/GeoMechanicsApplication/tests/cpp_tests/stub_time_loop_executor.cpp
avdg81: Let's leave it as it is now. We may want to revisit this point when we will write our coding conventions that supplement the Kratos Style Guide.

avdg81 (approved):
Thanks for fixing these SonarQube issues.

PR #11601, opened 2023-09-23
[MedApp] Extend Features

matekelemen:
## Description

*@philbucher This PR is a work in progress, and will include support for new features reading MED files. I'll keep the changelog updated as the PR advances.*

## Changelog

1) add support for reading to `ModelPart`s that already have nodes or geometries

### DIFF
--- a/applications/MedApplication/custom_io/med_model_part_io.cpp
+++ b/applications/MedApplication/custom_io/med_model_part_io.cpp
@@ -12,6 +12,7 @@
 
 // System includes
 #include <optional>
+#include <limits>
 
 // External includes
 
@@ -20,6 +21,7 @@
 #include "med_model_part_io.h"
 #include "includes/model_part_io.h"
 #include "utilities/parallel_utilities.h"
+#include "utilities/reduction_utilities.h"
 #include "utilities/variable_utils.h"
 
 namespace Kratos {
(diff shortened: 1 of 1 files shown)


philbucher:
Are you sure we should allow to fill non-empty ModelParts?

I am afraid the corner cases we need to consider is worth this feature?

Do you have a usecase?

philbucher:
Also, you could work on the integration of this into the reading workflow. This can be done independently of the reading of the submodelparts

Or how did you plan on using it?

matekelemen:
> Are you sure we should allow to fill non-empty ModelParts?

I don't see why we shouldn't support it; the `ModelPartIO` in core doesn't expect an empty `ModelPart` either, so I think we should match its behavior.

> I am afraid the corner cases we need to consider is worth this feature?

I don't see any immediate corner cases apart from index overflows, but hopefully I managed to handle those - I'll add some tests for them. Do you have some other cases in mind?

> Also, you could work on the integration of this into the reading workflow. This can be done independently of the reading of the submodelparts

I didn't quite get this; can you elaborate plz?

As I said, I'm not familiar with the MED lib yet, but I'm hoping we'll be able to call `MedModelPartIO::ReadModelPart` for sub model parts with different offsets within the file.

philbucher:
> > Are you sure we should allow to fill non-empty ModelParts?
> 
> I don't see why we shouldn't support it; the `ModelPartIO` in core doesn't expect an empty `ModelPart` either, so I think we should match its behavior.
> 
> > I am afraid the corner cases we need to consider is worth this feature?
> 
> I don't see any immediate corner cases apart from index overflows (but hopefully I managed to handle those - I'll add some tests for them later). Do you have some other cases in mind?

Not atm :thinking: 

> 
> > Also, you could work on the integration of this into the reading workflow. This can be done independently of the reading of the submodelparts
> 
> I didn't quite get this; can you elaborate plz?

I mean you can already add the `MedModelPartIO` e.g. in the `PythonSolver` etc. Whereever it is required to enable simulation with MED-files. Or how do you plan on using it? It is still some effort to go from Geometries to a simulatable domain

> 
> As I said, I'm not familiar with the MED lib yet, but my hope is that we'll be able to call `MedModelPartIO::ReadModelPart` recursively for sub model parts with different offsets within the file.

Ah no thats not what we will do, this would complicate things. We will use the submeshes from salome to create submodelparts.

PR #11603, opened 2023-09-25
install med in docker

philbucher:
I would like to enable the MedApplication in the CI due to the growing interest from the community.
For this we need to install the Med-library itself. It is not very large so I think it should not be an issue to add it


roigcarlo:
From my side go ahead :+1: 

p.s. still trying to figure out how to trim the artifact generation in the FullDebug-gcc CI.

roigcarlo (approved):

PR #11605, opened 2023-09-25
[MedApplication] Add support for reading SubModelParts

philbucher:
This PR adds support for reading SubMeshes from med into SubModelPart.


philbucher:
@matekelemen this is a draft/skeleton so far. I have not tested it, but this is basically how it works.

Feel free to give it a try with a simple case, otherwise I will continue in the next days

matekelemen:
thanks, I'm a bit short on time atm, but I'll definitely have a look next week!

philbucher:
> I'm a bit short on time atm,

well, you are a phd student after all xD


jokes aside, I did some more testing, and pushed a basic working version. Probably there are still some bugs to iron out, but in general it is working. Lets see if I have time in the next days to continue testing

philbucher:
Requires #11641 to work with SubSub...ModelParts

philbucher:
Ping @matekelemen @sunethwarna

applications/MedApplication/custom_io/med_model_part_io.cpp
```diff
@@ -440,10 +567,12 @@ void MedModelPartIO::ReadModelPart(ModelPart& rThisModelPart)
...
         &coordinatechangement, &geotransformation); // TODO error if smaller zero, holds probably for the other functions too that return med_int
 
-    int num_geometries_total = 0;
+    IndexType num_geometries_total = 0;
+
+    std::unordered_map<std::string, std::vector<IndexType>> smp_geoms;
 
     // looping geometry types
-    for (int it_geo=1; it_geo<=num_geometry_types; ++it_geo) {
+    for (int it_geo=1; it_geo<=num_geometry_types; ++it_geo) { //TODO +1?
```
matekelemen: soo
is it 1-based or not?

applications/MedApplication/custom_io/med_model_part_io.cpp
```diff
@@ -498,20 +638,51 @@ void MedModelPartIO::ReadModelPart(ModelPart& rThisModelPart)
             rThisModelPart.CreateNewGeometry(kratos_geo_name,
                                              ++num_geometries_total,
                                              geom_node_ids);
+
+            if (groups_by_fam.empty()) {continue;} // file does not contain families
+            const int fam_num = geom_family_numbers[i];
+            if (fam_num == 0) {continue;} // geometry does not belong to a SubModelPart
```
matekelemen, line 638: how can an object not belong to a family? Is it in the root family in this case? Or is it not part of the mesh at all?

applications/MedApplication/tests/med_files/cube_sub_subgroups/create_mesh.py
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env python
+
+###
+### This file is generated automatically by SALOME v9.6.0 with dump python functionality
+###
+
+import salome
```
matekelemen: Are these Salome scripts leftovers? Or maybe just some examples? You aren't planning to add Salome as a dependency, right?

matekelemen:
I'll take a closer look this Saturday.

I promise!

applications/MedApplication/tests/med_files/cube_sub_subgroups/create_mesh.py
philbucher: those are the batch scripts that can be used to recreate the mesh.

I guess I remove them, dont think they will be of much user later ...

And no, I will never add Salome as dependency

applications/MedApplication/custom_io/med_model_part_io.cpp
philbucher: good catch, I think the current version works correct, but I will check it again

applications/MedApplication/custom_io/med_model_part_io.cpp
philbucher: they all belong to a family. Family zero is where everything is that is not part of a group (aka SubModelPart).

The check in line 642 is if no family information was written to the file at all, the one in 644 if the entity only belongs to the main mesh but not any submodelparts

does that answer the question?

matekelemen: yup, got it. Thanks!

applications/MedApplication/tests/med_files/cube_sub_subgroups/create_mesh.py
philbucher: removed the scripts

applications/MedApplication/custom_io/med_model_part_io.cpp
philbucher: works correct as is, removed the comment

applications/MedApplication/custom_io/med_model_part_io.cpp
```diff
@@ -265,6 +273,74 @@ auto GetNodeCoordinates(
...
+
+        c_group_names.resize(MED_LNAME_SIZE * num_groups + 1);
+
+        const med_err err = MEDfamilyInfo(FileHandle, pMeshName, i, family_name.data(), &family_number, c_group_names.data());
+        CheckMEDErrorCode(err, "MEDfamilyInfo");
+
+        std::vector<std::string> group_names(num_groups);
+        // split the goup names
+        for (int i = 0; i < num_groups; i++) {
+            group_names[i] = std::string(c_group_names.substr(i * MED_LNAME_SIZE, MED_LNAME_SIZE));
```
matekelemen:
```suggestion
            group_names[i] = c_group_names.substr(i * MED_LNAME_SIZE, MED_LNAME_SIZE);
```

applications/MedApplication/custom_io/med_model_part_io.cpp
```diff
@@ -265,6 +273,74 @@ auto GetNodeCoordinates(
...
+            RemoveTrailingNullChars(group_names[i]);
+        }
+
+        groups_by_family[family_number] = group_names;
```
matekelemen:
```suggestion
        groups_by_family[family_number] = std::move(group_names);
```

applications/MedApplication/custom_io/med_model_part_io.cpp
```diff
@@ -265,6 +273,74 @@ auto GetNodeCoordinates(
...
+    KRATOS_CATCH("")
+}
+
+auto GetGroupsByFamily(
+    const med_idt FileHandle,
+	const char* pMeshName)
+{
+    KRATOS_TRY
+
+    std::unordered_map<int, std::vector<std::string>> groups_by_family;
```
matekelemen, line 306: How are family numbers generated in MED? If they're sequentially assigned (meaning that the highest family number is lower than the number of families) it might make sense to use an `std::vector` instead of an `std::unordered_map`, with the indices of entries representing the family number.

The reason why I'm bringing this up is because this map is in the hottest loop of reading: it's accessed for each node and geometry, so it would make sense to make it as efficient as possible.

applications/MedApplication/custom_io/med_model_part_io.cpp
```diff
@@ -410,6 +492,32 @@ void MedModelPartIO::ReadModelPart(ModelPart& rThisModelPart)
...
+    std::vector<med_int> node_family_numbers;
+    if (!groups_by_fam.empty()) {
+        node_family_numbers = GetFamilyNumbers(
+            mpFileHandler->GetFileHandle(),
+            mpFileHandler->GetMeshName(),
+            num_nodes,
+            med_entity_type::MED_NODE);
+    }
+
+    std::unordered_map<std::string, std::vector<IndexType>> smp_nodes;
```
matekelemen, line 519: Following along the lines of efficiency, it might make sense to refer to sub model parts by locally assigned IDs rather than their names, to avoid hashing a string for every node/geometry read.

This would involve a bit more work tho (the code below is just an example, I haven't tested it)
```diff
+    std::set<std::string> smp_names;
    // create SubModelPart hierarchy
    for (const auto& r_map : groups_by_fam) {
        for (const auto& r_smp_name : r_map.second) {
            if (!rThisModelPart.HasSubModelPart(r_smp_name)) {
                rThisModelPart.CreateSubModelPart(r_smp_name);
+                smp_names.insert(r_smp_name);
            }
        }
    }
```
```cpp
std::vector<std::pair<
    std::string,    // <== smp name
    std::size_t     // <== locally assigned ID to the smp
>> smp_name_id_map;
std::size_t smp_counter = 0ul;
std::transform(smp_names.begin(),
               smp_names.end(),
               std::back_inserter(smp_name_id_map),
               [&smp_counter](const std::string& r_smp_name){
                    return std::make_pair(r_smp_name, smp_counter++);
                });
smp_names.clear(); // <== no longer need smp names

std::vector<std::vector<std::size_t>> group_ids_by_fam(highest_family_num + 1); // <== assuming that family numbers are small (assigned sequentially)
for (int i_fam=0; i_fam<group_ids_by_fam.size(); ++i_fam) {
    const auto it_group_by_fam = groups_by_fam.find(i_fam);
    if (it_group_by_fam != groups_by_fam.end()) {
        const std::string& r_group_name = it_group_by_fam->second;

        // Find the local id of the smp whose name is r_group_name
        auto it_group_id_by_fam = std::find_if(group_ids_by_fam.begin(),
                                               group_ids_by_fam.end(),
                                               [&r_group_name](const auto& r_name_id_pair){
                                                    return r_name_id_pair.first == r_group_name;
                                               });
        if (it_group_id_by_fam != group_ids_by_fam.end()) [[likely]] {
            group_ids_by_fam[i_fam].push_back(it_group_id_by_fam->second);
        } else {
            KRATOS_ERROR << "Missing local ID for sub model part << r_group_name;
        }
    }
}

std::vector<std::vector<IndexType>> smp_nodes(smp_counter);
```
And then use the local IDs in the hot loop:
```diff
        if (groups_by_fam.empty()) {continue;} // file does not contain families

        const int fam_num = node_family_numbers[i];
        if (fam_num == 0) {continue;} // node does not belong to a SubModelPart

-         const auto it_groups = groups_by_fam.find(fam_num);
-         KRATOS_ERROR_IF(it_groups == groups_by_fam.end()) << "Missing node family with number " << fam_num << "!" << std::endl;
-         for (const auto& r_smp_name : it_groups->second) {
-             smp_nodes[r_smp_name].push_back(new_node_id);
-         }
+         KRATOS_ERROR_IF_NOT(fam_num < group_ids_by_fam.size()) << "Missing node family with number " << fam_num << "!" << std::endl;
+         const auto& r_group_ids = group_ids_by_fam[fam_num];
+         for (std::size_t smp_id : group_ids_by_fam[fam_num]) {
+             smp_nodes[smp_id].push_back(new_node_id);
+         }
```

applications/MedApplication/custom_io/med_model_part_io.cpp
```diff
@@ -440,7 +561,9 @@ void MedModelPartIO::ReadModelPart(ModelPart& rThisModelPart)
         MED_CONNECTIVITY, MED_NODAL,
         &coordinatechangement, &geotransformation); // TODO error if smaller zero, holds probably for the other functions too that return med_int
 
-    int num_geometries_total = 0;
+    IndexType num_geometries_total = 0;
+
+    std::unordered_map<std::string, std::vector<IndexType>> smp_geoms;
```
matekelemen, line 566: same as `smp_nodes`

philbucher:
thanks for the review @matekelemen 

Regarding the optimizations, lets do them in a follow-up PR so that this feature can already be used. I would like to spend some time on it and not rush it
What do you think?

matekelemen (approved):
Sure, thanks for the hard work!

PR #11611, opened 2023-09-26
[RomApp] Fixing bug in ECM

Rbravo555:
This PR fixes a bug found in the Empirical Cubature Method which was introduced in a recent addition (#11577)

**📝 Description**
An iteration counter inside the ECM was used to monitor whether the initial candidate set is not converging. However this was done incorrectly. The idea was that after a "large" number of unsuccessful iterations (this is still hard-coded to 1000), the algorithm should be considered not succesful. The counter now works correctly

Another change was performed to the RomManager in case a candidate set was passed and, it was not successful. Then the algorithm should be invoked again without any candidate set. This was done incorrectly in the RomManger due to a typo. 

**🆕 Changelog**
- Added the correct unsuccessful iteration counter to the ECM
- Added the correct invocation to the ECM to the RomManager, if the algorithm did not converge

### DIFF
--- a/applications/RomApplication/python_scripts/empirical_cubature_method.py
+++ b/applications/RomApplication/python_scripts/empirical_cubature_method.py
@@ -218,7 +218,7 @@ class EmpiricalCubatureMethod():
             MaximumLengthZ = max(MaximumLengthZ, np.size(self.z))
             k = k+1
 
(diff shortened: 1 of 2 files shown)


SADPR (approved):
Fast PR.

PR #11612, opened 2023-09-26
[RomApp] BugFix RomManager on a multiparametric simulation

Rbravo555:
**📝 Description**
This PR fixes a problem with the RomManager whenever a multiparametric simulation is launched. Once the parameters are passed to the first simulation, the object might be modified, and some simulations fail due to the presence of "new" fields in the parameters object, which make the validation process of the particular physics to fail.

**🆕 Changelog**
- Added a Parameters.Clone() to the different simulations in order not to add "new" fields


SADPR (approved):
LGTM. Fast PR.

PR #11625, opened 2023-09-27
[GeoMechanicsApplication] Add the missing conditions to Geomechanics Application for 2D

### LINKED ISSUE #11616: [GeoMechanicsApplication] Add the missing conditions to Geomechanics Application for 2D
**Description**
The conditions are not completely added to the GeoMechanics Application. For example, PwCondition is defined on 2 noded element (in 2D) but it is not extended to 3 noded element. There are also missing higher order condition elements in 3D.
Here we fix the 2D conditions. The 3D conditions will be done in a seperate issue in order to keep this commit managable.

**TODO**
- [ ] Add the missing 2D conditions
- [ ] Modify the functions to make them generic for n-th order line elements/conditions
- [ ] Test the implementation
- [ ] If possible, add test cases for the missing conditions.

mnabideltares:
**📝 Description**
The conditions are not completely added to the GeoMechanics Application. For example, PwCondition is defined on 2 noded element (in 2D) but it is not extended to 3 noded element. There are also missing higher order condition elements in 3D.
Here we fix the 2D conditions. The 3D conditions will be done in a seperate issue in order to keep this commit managable

**🆕 Changelog**
- Added missing 2D conditions
- Modified the functions to make them generic for n-th order line elements/conditions
- Cleaned up some parts

** Acceptance
- All existing test cases have to pass


WPK4FEM (changes requested):
Codewise this is quite an improvement, the formulas have been written down using vector multiplications i.s.o. writing down component multiplication and summation.

However, new conditions have been added without adding tests for the new conditions. These have to be made.

Further, run clang format on the changed code.

applications/GeoMechanicsApplication/tests/test_conditions/test_AxisymmetricUPwNormalFaceLoadCondition2D2N/ProjectParameters.json
```diff
@@ -0,0 +1,171 @@
...
+                        "file_label":          "step",
+                        "output_control_type": "step",
+                        "output_interval":     1,
+                        "body_output":         true,
+                        "node_output":         true,
+                        "skin_output":         false,
+                        "plane_output":        [],
+                        "nodal_results":       ["DISPLACEMENT","TOTAL_DISPLACEMENT","WATER_PRESSURE","NORMAL_FLUID_FLUX","VOLUME_ACCELERATION","HYDRAULIC_DISCHARGE"],
+                        "gauss_point_results": ["FLUID_FLUX_VECTOR","LOCAL_FLUID_FLUX_VECTOR","HYDRAULIC_HEAD"]
+                    },
```
rfaasse: Since we only check DISPLACEMENT, we don't need to output off all these variables. Could you remove the redundant ones (also in the other tests?)

applications/GeoMechanicsApplication/tests/test_GeoMechanicsApplication.py
```diff
@@ -80,7 +81,8 @@ def AssembleTestSuites():
                         KratosGeoMechanicsChangingWaterLevelTests,
                         KratosGeoMechanicsStrainMeasureTests,
                         KratosGeoMechanicsSetMultipleMovingLoadProcessTests,
-                        KratosGeoMechanicsRotationWithMovingLoadTests
+                        KratosGeoMechanicsRotationWithMovingLoadTests,
+                        KratosGeoMechanicsConditionTests
```
rfaasse, line 85: Just checking: how long do these tests take approximately (i.e. are they short enough to run in the small tests)?

applications/GeoMechanicsApplication/custom_conditions/Pw_normal_flux_condition.cpp
```diff
@@ -88,66 +88,32 @@ void PwNormalFluxCondition<TDim,TNumNodes>::
...
-    return ds * Weight;
-}
-
-
-//----------------------------------------------------------------------------------------
-template< >
-double PwNormalFluxCondition<3, 3>::
-CalculateIntegrationCoefficient(const Matrix& Jacobian,
+// ============================================================================================
+// ============================================================================================
```
rfaasse: Could you please remove these comment lines (here and at the other conditions)? They take away space and in my opinion do not clarify the code

applications/GeoMechanicsApplication/custom_conditions/Pw_normal_flux_condition.cpp
```diff
@@ -88,66 +88,32 @@ void PwNormalFluxCondition<TDim,TNumNodes>::
...
-    return dA * Weight;
+    if constexpr (TDim == 2)
+    {
+        NormalVector = column(Jacobian, 0);
+    }
+    else if constexpr (TDim == 3)
+    {
+        MathUtils<double>::CrossProduct(NormalVector, column(Jacobian, 0), column(Jacobian, 1));
+    }
+    return MathUtils<double>::Norm(NormalVector) * Weight;
```
rfaasse: This removes a lot of redundant duplications and error-prone, hard to read code, very nice!

applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
```diff
@@ -298,21 +298,14 @@ double UPwFaceLoadInterfaceCondition<2,2>::CalculateIntegrationCoefficient(const
 //----------------------------------------------------------------------------------------
 
 template< >
-double UPwFaceLoadInterfaceCondition<3,4>::CalculateIntegrationCoefficient(const Matrix& Jacobian, const double& Weight, const double& JointWidth)
+double UPwFaceLoadInterfaceCondition<3, 4>::CalculateIntegrationCoefficient(const Matrix& Jacobian, const double& Weight, const double& JointWidth)
```
rfaasse, line 302: It might be a good idea to also generalize this function (in the same way you did for the earlier ones), such that if we ever want to add versions with differen TDim/TNumNodes, we don't need to change the class. I'm not sure if that's realistic for this type of condition, I leave that up to you.

applications/GeoMechanicsApplication/custom_conditions/U_Pw_normal_face_load_condition.cpp
```diff
@@ -72,134 +72,65 @@ void UPwNormalFaceLoadCondition<TDim,TNumNodes>::
...
-
-//----------------------------------------------------------------------------------------
-template< >
-void UPwNormalFaceLoadCondition<3,4>::
-    InitializeConditionVariables(NormalFaceLoadVariables& rVariables,
-                                 const GeometryType& rGeom)
-{
-    for (unsigned int i=0; i<4; ++i) {
-        rVariables.NormalStressVector[i] = rGeom[i].FastGetSolutionStepValue(NORMAL_CONTACT_STRESS);
+    //
```
rfaasse: Could you remove this empty comment?

applications/GeoMechanicsApplication/custom_conditions/U_Pw_normal_face_load_condition.cpp
```diff
@@ -72,134 +72,65 @@ void UPwNormalFaceLoadCondition<TDim,TNumNodes>::
...
-//----------------------------------------------------------------------------------------
-template< >
-void UPwNormalFaceLoadCondition<3,3>::
-    InitializeConditionVariables(NormalFaceLoadVariables& rVariables,
-                                 const GeometryType& rGeom)
-{
-    for (unsigned int i=0; i<3; ++i) {
+    for (unsigned int i = 0; i < TNumNodes; ++i) {
         rVariables.NormalStressVector[i] = rGeom[i].FastGetSolutionStepValue(NORMAL_CONTACT_STRESS);
     }
```
rfaasse: I think this loop could be replaced with an std::transform (and similarly line 87-89):
```suggestion
    std::transform(rGeom.begin(), rGeom.end(), rVariables.NormalStressVector.begin(),
                   [](const auto& node) {
                       return node.FastGetSolutionStepValue(NORMAL_CONTACT_STRESS);
                   });
```

applications/GeoMechanicsApplication/custom_conditions/U_Pw_normal_face_load_condition.cpp
```diff
@@ -72,134 +72,65 @@ void UPwNormalFaceLoadCondition<TDim,TNumNodes>::
...
-//----------------------------------------------------------------------------------------
-template< >
-void UPwNormalFaceLoadCondition<3,4>::
-    InitializeConditionVariables(NormalFaceLoadVariables& rVariables,
-                                 const GeometryType& rGeom)
-{
-    for (unsigned int i=0; i<4; ++i) {
-        rVariables.NormalStressVector[i] = rGeom[i].FastGetSolutionStepValue(NORMAL_CONTACT_STRESS);
+    //
+    if (TDim == 2) {
```
rfaasse: This can be constexpr, such that it's checked at compile time
```suggestion
    constexpr if (TDim == 2) {
```

applications/GeoMechanicsApplication/custom_conditions/U_Pw_normal_face_load_condition.cpp
```diff
@@ -72,134 +72,65 @@ void UPwNormalFaceLoadCondition<TDim,TNumNodes>::
...
+    const unsigned int& GPoint)
 {
-    double NormalStress = 0.0;
-    double TangentialStress = 0.0;
-
-    for (unsigned int i=0; i<2; ++i) {
-        NormalStress     += NContainer(GPoint,i)*Variables.NormalStressVector[i];
-        TangentialStress += NContainer(GPoint,i)*Variables.TangentialStressVector[i];
+    Vector NormalVector = ZeroVector(TDim);
+    double NormalStress = MathUtils<>::Dot(row(NContainer, GPoint), Variables.NormalStressVector);
```
rfaasse: Due to your improvement, this can be const  (very nice improvement btw, way better than the old loop)
```suggestion
    const double NormalStress = MathUtils<>::Dot(row(NContainer, GPoint), Variables.NormalStressVector);
```

applications/GeoMechanicsApplication/custom_conditions/U_Pw_normal_face_load_condition.cpp
```diff
@@ -72,134 +72,65 @@ void UPwNormalFaceLoadCondition<TDim,TNumNodes>::
...
+    //
+    if (TDim == 2) {
+        double TangentialStress = MathUtils<>::Dot(row(NContainer, GPoint), Variables.TangentialStressVector);
```
rfaasse: Also this one can be const 
```suggestion
        const double TangentialStress = MathUtils<>::Dot(row(NContainer, GPoint), Variables.TangentialStressVector);
```

applications/GeoMechanicsApplication/custom_conditions/U_Pw_normal_face_load_condition.cpp
```diff
@@ -72,134 +72,65 @@ void UPwNormalFaceLoadCondition<TDim,TNumNodes>::
...
-    double NormalStress = 0.0;
-    double TangentialStress = 0.0;
-
-    for (unsigned int i=0; i<2; ++i) {
-        NormalStress     += NContainer(GPoint,i)*Variables.NormalStressVector[i];
-        TangentialStress += NContainer(GPoint,i)*Variables.TangentialStressVector[i];
+    Vector NormalVector = ZeroVector(TDim);
+    double NormalStress = MathUtils<>::Dot(row(NContainer, GPoint), Variables.NormalStressVector);
+    //
+    if (TDim == 2) {
```
rfaasse:
```suggestion
    constexpr if (TDim == 2) {
```

applications/GeoMechanicsApplication/custom_conditions/axisymmetric_U_Pw_normal_face_load_condition.cpp
```diff
@@ -46,11 +46,31 @@ template< unsigned int TDim, unsigned int TNumNodes >
 GeometryData::IntegrationMethod 
     AxisymmetricUPwNormalFaceLoadCondition<TDim,TNumNodes>::GetIntegrationMethod() const
 {
-    return GeometryData::IntegrationMethod::GI_GAUSS_2;
+    GeometryData::IntegrationMethod GI_GAUSS;
+    //
```
rfaasse: Could you remove this empty comment?

applications/GeoMechanicsApplication/custom_conditions/axisymmetric_U_Pw_normal_face_load_condition.cpp
```diff
@@ -46,11 +46,31 @@ template< unsigned int TDim, unsigned int TNumNodes >
...
+    case 3:
+        GI_GAUSS = GeometryData::IntegrationMethod::GI_GAUSS_2;
+        break;
+    case 4:
+        GI_GAUSS = GeometryData::IntegrationMethod::GI_GAUSS_3;
+        break;
+    default:
+        GI_GAUSS = GeometryData::IntegrationMethod::GI_GAUSS_2;
+        break;
+    }
```
rfaasse: This is probably my lack of knowledge, but this switch is not completely clear to me. We added <2,3>, <2,4> and <2,5> elements, which now result in `GI_GAUSS_2`, `GI_GAUSS_3` and `GI_GAUSS_2` respectively. Why is that?

applications/GeoMechanicsApplication/geo_mechanics_application.cpp
```diff
@@ -256,8 +264,6 @@ void KratosGeoMechanicsApplication::Register() {
     KRATOS_REGISTER_CONDITION("LineNormalLoadDiffOrderCondition2D4N", mLineNormalLoadDiffOrderCondition2D4N)
     KRATOS_REGISTER_CONDITION("LineNormalLoadDiffOrderCondition2D5N", mLineNormalLoadDiffOrderCondition2D5N)
     KRATOS_REGISTER_CONDITION("LineNormalFluidFluxDiffOrderCondition2D3N", mLineNormalFluidFluxDiffOrderCondition2D3N)
-    KRATOS_REGISTER_CONDITION("LineNormalFluidFluxDiffOrderCondition2D4N", mLineNormalFluidFluxDiffOrderCondition2D4N)
-    KRATOS_REGISTER_CONDITION("LineNormalFluidFluxDiffOrderCondition2D5N", mLineNormalFluidFluxDiffOrderCondition2D5N)
```
rfaasse, line 260: Should these 2 be removed? Do they also not work correctly for these already existing diff order conditions?

applications/GeoMechanicsApplication/geo_mechanics_application.h
```diff
@@ -503,8 +511,6 @@ class KRATOS_API(GEO_MECHANICS_APPLICATION) KratosGeoMechanicsApplication : publ
     const LineNormalLoad2DDiffOrderCondition mLineNormalLoadDiffOrderCondition2D5N{ 0, Kratos::make_shared< Line2D5 <NodeType> >(Condition::GeometryType::PointsArrayType(5)) };
 
     const LineNormalFluidFlux2DDiffOrderCondition mLineNormalFluidFluxDiffOrderCondition2D3N{ 0, Kratos::make_shared< Line2D3 <NodeType> >(Condition::GeometryType::PointsArrayType(3)) };
-    const LineNormalFluidFlux2DDiffOrderCondition mLineNormalFluidFluxDiffOrderCondition2D4N{ 0, Kratos::make_shared< Line2D4 <NodeType> >(Condition::GeometryType::PointsArrayType(4)) };
-    const LineNormalFluidFlux2DDiffOrderCondition mLineNormalFluidFluxDiffOrderCondition2D5N{ 0, Kratos::make_shared< Line2D5 <NodeType> >(Condition::GeometryType::PointsArrayType(5)) };
```
rfaasse, line 503: Same here, why were they removed?

applications/GeoMechanicsApplication/tests/test_conditions.py
```diff
@@ -0,0 +1,246 @@
+import os
```
rfaasse, line 1: Could we add a README.md for these tests (a single one for all of these is probably enough)? That also helps me understand the test a bit better (e.g. which node are we testing at which time-step, which effects are taken into account etc)

rfaasse (commented):
Thanks a lot @mnabideltares for all the work in this PR (both the code improvements and the added tests), it cleans up a lot of duplication and improves the readability significantly! 

I have a few suggestions for improvements, let me know what you think!

applications/GeoMechanicsApplication/custom_conditions/axisymmetric_U_Pw_normal_face_load_condition.cpp
```diff
@@ -46,11 +46,31 @@ template< unsigned int TDim, unsigned int TNumNodes >
 GeometryData::IntegrationMethod 
     AxisymmetricUPwNormalFaceLoadCondition<TDim,TNumNodes>::GetIntegrationMethod() const
 {
-    return GeometryData::IntegrationMethod::GI_GAUSS_2;
+    GeometryData::IntegrationMethod GI_GAUSS;
+    //
+    switch (TNumNodes) {
+    case 2:
+        GI_GAUSS = GeometryData::IntegrationMethod::GI_GAUSS_2;
+        break;
```
markelov208: It is possible to remove these lines.

applications/GeoMechanicsApplication/tests/test_conditions.py
```diff
@@ -0,0 +1,246 @@
...
+        # Code here will be placed AFTER every test in this TestCase.
+        pass
+
+    def test_PwNormalFluxCondition2D2N(self):
+        test_name = 'test_PwNormalFluxCondition2D2N'
+        file_path = test_helper.get_file_path(os.path.join('test_conditions', test_name))
+        simulation = test_helper.run_kratos(file_path)
+        pressure = test_helper.get_water_pressure(simulation)
+        pres = pressure[2]
+        self.assertAlmostEqual(self.etalon_value1, pres)
```
markelov208: If python files are also checked with sonarcloud, then it is better to create a function for these lines and called it for each case, for example, as check_water_pressure('test_PwNormalFluxCondition2D2N', self.etalon_value1)

applications/GeoMechanicsApplication/tests/test_conditions.py
```diff
@@ -0,0 +1,246 @@
...
+        self.assertAlmostEqual(self.etalon_value3, pres)
+
+        
+    def test_UPwFaceLoadCondition2D2N(self):
+        test_name = 'test_UPwFaceLoadCondition2D2N'
+        file_path = test_helper.get_file_path(os.path.join('test_conditions', test_name))
+        simulation = test_helper.run_kratos(file_path)
+        displacement = test_helper.get_displacement(simulation)
+        disp = displacement[2]
+        self.assertAlmostEqual(self.etalon_value2, disp[1])
```
markelov208: a function for displacement like for water_pressure

applications/GeoMechanicsApplication/tests/test_conditions/test_AxisymmetricUPwNormalFaceLoadCondition2D3N/MaterialParameters.json
markelov208, line 1: A general question, it is easy to keep all required files in a separate folder. However, some files like MaterialParameters,json are identical for many tests. Is it good to have only different MaterialParameters files and refer to a specific file?

applications/GeoMechanicsApplication/tests/test_conditions/test_AxisymmetricUPwNormalFaceLoadCondition2D4N/ProjectParameters.json
```diff
@@ -0,0 +1,171 @@
+{
+    "problem_data": {
+        "problem_name":         "test_AxisymmetricUPwNormalFaceLoadCondition2D4N",
+        "start_time":           0.0,
+        "end_time":             18000.0,
```
markelov208: In the previous test, end_time is only 1200. Why is it larger by a factor of 15 here?

applications/GeoMechanicsApplication/geo_mechanics_application.h
```diff
@@ -518,11 +524,18 @@ class KRATOS_API(GEO_MECHANICS_APPLICATION) KratosGeoMechanicsApplication : publ
...
 
     const AxisymmetricLineNormalLoad2DDiffOrderCondition mAxisymmetricLineNormalLoadDiffOrderCondition2D3N{ 0, Kratos::make_shared< Line2D3<NodeType> >(Condition::GeometryType::PointsArrayType(3)) };
+    const AxisymmetricLineNormalLoad2DDiffOrderCondition mAxisymmetricLineNormalLoadDiffOrderCondition2D4N{ 0, Kratos::make_shared< Line2D4<NodeType> >(Condition::GeometryType::PointsArrayType(4)) };
+    const AxisymmetricLineNormalLoad2DDiffOrderCondition mAxisymmetricLineNormalLoadDiffOrderCondition2D5N{ 0, Kratos::make_shared< Line2D5<NodeType> >(Condition::GeometryType::PointsArrayType(5)) };
 
     const AxisymmetricLineNormalFluidFlux2DDiffOrderCondition mAxisymmetricLineNormalFluidFluxDiffOrderCondition2D3N{ 0, Kratos::make_shared< Line2D3<NodeType> >(Condition::GeometryType::PointsArrayType(3)) };
+    const AxisymmetricLineNormalFluidFlux2DDiffOrderCondition mAxisymmetricLineNormalFluidFluxDiffOrderCondition2D4N{ 0, Kratos::make_shared< Line2D4<NodeType> >(Condition::GeometryType::PointsArrayType(4)) };
```
WPK4FEM: These are added, but the equivalent ones for plane strain are removed. Why?

markelov208 (commented):
Hi Mohamed, great work! So many changes to unify functions and remove code smells. Just a few questions from my side.

applications/GeoMechanicsApplication/tests/test_conditions.py
WPK4FEM: I see no need for variable pres as we can write:
self.asserAlmostEqual(self.etalon_value1, pressure[2])

This repeats many times.

applications/GeoMechanicsApplication/tests/test_conditions.py
```diff
@@ -0,0 +1,246 @@
...
+        pres = pressure[2]
+        self.assertAlmostEqual(self.etalon_value1, pres)
+
+    def test_PwNormalFluxCondition2D3N(self):
+        test_name = 'test_PwNormalFluxCondition2D3N'
+        file_path = test_helper.get_file_path(os.path.join('test_conditions', test_name))
+        simulation = test_helper.run_kratos(file_path)
+        pressure = test_helper.get_water_pressure(simulation)
+        pres = pressure[2]
+        self.assertAlmostEqual(self.etalon_value1, pres)
```
WPK4FEM: The only changing thing in all these bodies is test_name and the etalon value. That urges for writing a procedure that has test_name and etalon as argument. It will save 6 lines per test.

applications/GeoMechanicsApplication/tests/test_conditions/test_AxisymmetricUPwNormalFaceLoadCondition2D2N/ProjectParameters.json
```diff
@@ -0,0 +1,171 @@
...
+        "python_module": "apply_vector_constraint_table_process",
+        "kratos_module": "KratosMultiphysics.GeoMechanicsApplication",
+        "process_name":  "ApplyVectorConstraintTableProcess",
+        "Parameters":    {
+            "model_part_name": "PorousDomain.Gravity",
+            "variable_name":   "VOLUME_ACCELERATION",
+            "active":          [false,true,false],
+            "value":           [0.0,0.0,0.0],
+            "table":           [0,0,0]
+        }
```
WPK4FEM: If no gravity is intended, leave out this ApplyVectorConstraintTableProcess.

applications/GeoMechanicsApplication/tests/test_conditions/test_AxisymmetricUPwNormalFaceLoadCondition2D2N/test_AxisymmetricUPwNormalFaceLoadCondition2D2N.mdpa
```diff
@@ -0,0 +1,189 @@
...
+  7  1  3  8  9
+  8  1  5  9  8
+End Elements
+
+
+Begin Conditions UPwNormalFluxCondition2D2N
+  1  1  4  8
+  2  1  8  5
+End Conditions
+
```
WPK4FEM: The UPwNormalFluxCondition2D2N is not used in this test. Please remove.

applications/GeoMechanicsApplication/tests/test_conditions/test_AxisymmetricUPwNormalFaceLoadCondition2D2N/test_AxisymmetricUPwNormalFaceLoadCondition2D2N.mdpa
```diff
@@ -0,0 +1,189 @@
...
+    5
+    8
+  End SubModelPartNodes
+  Begin SubModelPartElements
+  End SubModelPartElements
+  Begin SubModelPartConditions
+    1
+    2
+  End SubModelPartConditions
+End SubModelPart
```
WPK4FEM: Not used SubModelPart.

applications/GeoMechanicsApplication/tests/test_conditions/test_AxisymmetricUPwNormalFaceLoadCondition2D2N/test_AxisymmetricUPwNormalFaceLoadCondition2D2N.mdpa
```diff
@@ -0,0 +1,189 @@
...
+    4
+    5
+    6
+    7
+    8
+  End SubModelPartElements
+  Begin SubModelPartConditions
+  End SubModelPartConditions
+End SubModelPart
+
```
WPK4FEM: When the process for gravity is not used, this also can be removed.

applications/GeoMechanicsApplication/tests/test_conditions/test_AxisymmetricUPwNormalFaceLoadCondition2D2N/MaterialParameters.json
```diff
@@ -0,0 +1,35 @@
...
+              "POROSITY"                 :  0.0,
+              "BULK_MODULUS_SOLID"       :  1.0e9,
+              "BULK_MODULUS_FLUID"       :  2.0e6,
+              "PERMEABILITY_XX"          :  1.157e-12,
+              "PERMEABILITY_YY"          :  1.157e-12,
+              "PERMEABILITY_XY"          :  0.0,
+              "DYNAMIC_VISCOSITY"        :  1e-3,
+              "THICKNESS"                :  1.0,
+              "BIOT_COEFFICIENT"         :  1.0,
+              "RETENTION_LAW"                    : "SaturatedLaw",
```
WPK4FEM: The combination of SaturatedLaw and porosity 0 will still lead to floatation ( buoyancy ) i.c.w. gravityload.

applications/GeoMechanicsApplication/tests/test_conditions/test_AxisymmetricUPwNormalFaceLoadCondition2D2N/ProjectParameters.json
```diff
@@ -0,0 +1,171 @@
+{
```
WPK4FEM, line 1: Remarks for this test apply to the other tests as well.

applications/GeoMechanicsApplication/tests/test_conditions/test_AxisymmetricUPwNormalFaceLoadCondition2D2N/ProjectParameters.json
```diff
@@ -0,0 +1,171 @@
...
+                            "MultiFileFlag":         "SingleFile"
+                        },
+                        "file_label":          "step",
+                        "output_control_type": "step",
+                        "output_interval":     1,
+                        "body_output":         true,
+                        "node_output":         true,
+                        "skin_output":         false,
+                        "plane_output":        [],
+                        "nodal_results":       ["DISPLACEMENT","TOTAL_DISPLACEMENT","WATER_PRESSURE","NORMAL_FLUID_FLUX","VOLUME_ACCELERATION","HYDRAULIC_DISCHARGE"],
```
WPK4FEM: REACTION or REACTION_Y may be a way to check for the presence of the full applied load.

applications/GeoMechanicsApplication/tests/test_conditions/test_LineNormalFluidFluxDiffOrderCondition2D3N/ProjectParameters.json
```diff
@@ -0,0 +1,160 @@
...
+        "python_module": "apply_vector_constraint_table_process",
+        "kratos_module": "KratosMultiphysics.GeoMechanicsApplication",
+        "process_name":  "ApplyVectorConstraintTableProcess",
+        "Parameters":    {
+            "model_part_name": "PorousDomain.Body_Acceleration-auto-1",
+            "variable_name":   "VOLUME_ACCELERATION",
+            "active":          [false,true,false],
+            "value":           [0.0,-10,0.0],
+            "table":           [0,0,0]
+        }
```
WPK4FEM: Why use gravity if all nodal displacements are fixed/supported. This load will disappear into the supports immediately.

applications/GeoMechanicsApplication/tests/test_conditions/test_LineNormalFluidFluxDiffOrderCondition2D3N/ProjectParameters.json
```diff
@@ -0,0 +1,160 @@
...
+        "python_module": "apply_scalar_constraint_table_process",
+        "kratos_module": "KratosMultiphysics.GeoMechanicsApplication",
+        "process_name":  "ApplyScalarConstraintTableProcess",
+        "Parameters":    {
+            "model_part_name":      "PorousDomain.PressureBottom",
+            "variable_name":        "WATER_PRESSURE",
+            "is_fixed":             true,
+            "fluid_pressure_type": "Hydrostatic",
+            "gravity_direction":    1,
+            "reference_coordinate": -2,
```
WPK4FEM: This says that the phreatic line is at the bottom of your model. The saturated law tells that there is water above the phreatic line. For the flux not to blow up the model like filling a balloon a prescribed pressure somewhere is needed to create an exit flux. Any constant value on that Dirichlet boundary will do, so I would use a uniform pressure value i.s.o. of using a hydrostatic field.

WPK4FEM (commented):
Dear Mohammed,
Thank you for the massive work. To get it really complete, please add documentation of the tests and a short documentation of the functionality. I think the formulas are available from the document https://deltares.sharepoint.com/:p:/r/teams/d730-KratosTeamChannel/_layouts/15/Doc2.aspx?action=edit&sourcedoc=%7Bd26c925e-fced-4911-8f80-a691e9ef8a5a%7D&wdOrigin=TEAMS-MAGLEV.teamsSdk_ns.rwc&wdExp=TEAMS-TREATMENT&wdhostclicktime=1710233895908&web=1
Or similar to the document you wrote for the T conditions.

Regards, Wijtze Pieter

applications/GeoMechanicsApplication/custom_conditions/Pw_normal_flux_condition.cpp
rfaasse: Due to another review, I came across the function `static double CalculateIntegrationCoefficient(const Matrix& rJacobian, double Weight)` in ConditionUtilities, that does exactly this, we can probably re-use that one?

applications/GeoMechanicsApplication/tests/test_conditions.py
markelov208: This function is created.

applications/GeoMechanicsApplication/tests/test_conditions/test_AxisymmetricUPwNormalFaceLoadCondition2D3N/MaterialParameters.json
markelov208: MaterialParameters.json files are moved to Common folder

applications/GeoMechanicsApplication/tests/test_conditions.py
markelov208: The functions is created.

applications/GeoMechanicsApplication/geo_mechanics_application.cpp
rfaasse: @mnabideltares Could you please add the answer to this question here, such that it's documented?

mnabideltares: I will handle these two conditions (and two other from Axisymmetric which are removed in revision), in a coming issue related to diff_order conditions. 

Because the locations of the nodes do not match for 15 and 10, as well as, for 10 and 6 elements, special treatment need to be taken into account. For the conditions, it is also same. The diff-order conditions do not match for 5 and 4 line conditions. I think the element and condition need to be fixed together so that they use the same ordering and same interpolation schemes.

Here, if the conditions are registed in Kratos, it can happern that a user unaware of this issue, tries to use. The danger of using it is that it works but it gives wrong results. So, I decided to remove them temporarily from the list to not give opportunity to be used at the current state. Next sprint, these conditions, together with diff_order elements will be handeled and conditions will be added.

Will be fixed in:
issue: #11484
PR: https://github.com/KratosMultiphysics/Kratos/pull/11558

applications/GeoMechanicsApplication/tests/test_conditions/README.md
```diff
@@ -0,0 +1,54 @@
...
+- PwNormalFluxCondition2D5N
+- UPwNormalFluxCondition2D2N
+- UPwNormalFluxCondition2D3N
+- UPwNormalFluxCondition2D4N
+- UPwNormalFluxCondition2D5N
+
+
+## Setup
+
+This tests consist of 3, 6, 10 and 15-noded elements wuth 2, 3, 4, 5 noded line conditions, respectively. For all cases, a rectangular domain of $1 \mathrm{m} \cdot 2 \mathrm{m}$, with 8 triangular elements is used. A schematic representation can be found in the figure below:
```
markelov208: It is nicer $1 \mathrm{m} \times 2 \mathrm{m}$

markelov208 (commented):
Mohamed, just a small comment on READMe file but it is not blocking.

applications/GeoMechanicsApplication/custom_conditions/axisymmetric_U_Pw_normal_face_load_condition.cpp
mnabideltares: Gauss points. 2 noded element, needs 1 point. However, we use usually 2 points practically we do not get accurate solution wih 1 point.
For 3 noded => 2 points
4 noded => 3 points
5 noded => proposed to give by 4 or 5 points. However, 2 points gave accurate solution, why 4 or 5 points gave a difference of 10^-8. It needs to be investigated, but I think it has no priority at moment.

applications/GeoMechanicsApplication/tests/test_GeoMechanicsApplication.py
mnabideltares: It takes less than one minute in debug mode. We probably can shift them to nightly if we exceed the 90 seconds on Git?

applications/GeoMechanicsApplication/tests/test_conditions/documentation_data/load_flux_domain.svg
```diff
@@ -0,0 +1 @@
...9.167 69.3667 322.5 82.7001 315.833 69.3667Z" fill="#4472C4"/><path d="M371.833 19.5001 371.833 70.7001 369.167 70.7001 369.167 19.5001ZM377.167 69.3667 370.5 82.7001 363.833 69.3667Z" fill="#4472C4"/><path d="M419.833 19.5001 419.833 70.7001 417.167 70.7001 417.167 19.5001ZM425.167 69.3667 418.5 82.7001 411.833 69.3667Z" fill="#4472C4"/><path d="M467.833 19.5001 467.833 70.7001 465.167 70.7001 465.167 19.5001ZM473.167 69.3667 466.5 82.7001 459.833 69.3667Z" fill="#4472C4"/><path d="M515.833 19.5001 515.833 70.7001 513.167 70.7001 513.167 19.5001ZM521.167 69.3667 514.5 82.7001 507.833 69.3667Z" fill="#4472C4"/><path d="M563.833 19.5001 563.833 70.7001 561.167 70.7001 561.167 19.5001ZM569.167 69.3667 562.5 82.7001 555.833 69.3667Z" fill="#4472C4"/><path d="M611.833 19.5001 611.833 70.7001 609.167 70.7001 609.167 19.5001ZM617.167 69.3667 610.5 82.7001 603.833 69.3667Z" fill="#4472C4"/><text font-family="Calibri,Calibri_MSFontService,sans-serif" font-style="italic" font-weight="700" font-size="24" transform="translate(261.847 46)">q</text><text font-family="Calibri,Calibri_MSFontService,sans-serif" font-weight="400" font-size="24" transform="translate(280.013 46)">[N]</text></g></svg>
```
WPK4FEM: Nice start of a picture, some remarks:
- q loads or distributed line loads are usually schematized with a connecting line over the starts of the arrows. See the copyable thing in our team channel Documents/Schematics.svg ( only needs to be rotated )
- The support indication with skew lines usually connects to the line for which they apply and stay within the width of the line. See  the copyable thing in our team channel Documents/Schematics.svg

applications/GeoMechanicsApplication/tests/test_conditions/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+The material is described using:
+-   A linear elastic plane strain model (LinearElasticPlaneStrain2DLaw)
+-   A Young's modulus $E = 2.0E6 \mathrm{[Pa]}$ with Poisson's ratio $\nu = 0.0 \mathrm{[-]}$.
+-   The soil and water density are specified, but irrelevant due to the absence of gravity. The porosity is 0.
+-   The dynamic viscosity of water is given as $\mu = 1.0E-3 \mathrm{[Pa \cdot s]}$ and the intrinsic permeability of the soil as $\kappa = 1.157E-12 \mathrm{[m^2]}$
+-   The bulk modulus of water $K_w = 2.0E6 \mathrm{[Pa]}$
+
+## Assertions
+The test asserts that the vertical distribution of displacement and pressure matches the linear distribution which matches the analytical solution.
```
WPK4FEM: Proposed:
The tests assert that the vertical distribution of displacement and pressure matches the analytical distribution, which is a linear distribution over the height of the column.

applications/GeoMechanicsApplication/tests/test_conditions/test_AxisymmetricUPwNormalFaceLoadCondition2D4N/ProjectParameters.json
```diff
@@ -0,0 +1,160 @@
...
+        "python_module": "apply_normal_load_table_process",
+        "kratos_module": "KratosMultiphysics.GeoMechanicsApplication",
+        "process_name":  "ApplyNormalLoadTableProcess",
+        "Parameters":    {
+            "model_part_name": "PorousDomain.LoadTop",
+            "variable_name":   "NORMAL_CONTACT_STRESS",
+            "active":          [true,false],
+            "table":           [0,0],
+            "fluid_pressure_type": "Uniform",
+            "value":           [1000,0.0]
```
WPK4FEM, line 155: The fluid_pressure_type line is probably a copying mistake. I think it should not be here.

applications/GeoMechanicsApplication/tests/test_conditions/test_AxisymmetricUPwNormalFaceLoadCondition2D5N/ProjectParameters.json
```diff
@@ -0,0 +1,160 @@
...
+    "loads_process_list": [{
+        "python_module": "apply_normal_load_table_process",
+        "kratos_module": "KratosMultiphysics.GeoMechanicsApplication",
+        "process_name":  "ApplyNormalLoadTableProcess",
+        "Parameters":    {
+            "model_part_name": "PorousDomain.LoadTop",
+            "variable_name":   "NORMAL_CONTACT_STRESS",
+            "active":          [true,false],
+            "table":           [0,0],
+            "fluid_pressure_type": "Uniform",
```
WPK4FEM, line 154: The fluid_pressure_type line does not belong here for a normal load with given value. 

Likely this repeats in the other NormalFaceLoadCondition tests.

applications/GeoMechanicsApplication/tests/test_conditions/test_LineNormalLoadDiffOrderCondition2D3N/ProjectParameters.json
```diff
@@ -0,0 +1,160 @@
...
+    "loads_process_list": [{
+        "python_module": "apply_normal_load_table_process",
+        "kratos_module": "KratosMultiphysics.GeoMechanicsApplication",
+        "process_name":  "ApplyNormalLoadTableProcess",
+        "Parameters":    {
+            "model_part_name": "PorousDomain.LoadTop",
+            "variable_name":   "NORMAL_CONTACT_STRESS",
+            "active":          [true,false],
+            "table":           [0,0],
+            "fluid_pressure_type": "Uniform",
```
WPK4FEM, line 154: Probably doesn't belong here.

applications/GeoMechanicsApplication/tests/test_conditions/test_LineNormalLoadDiffOrderCondition2D4N/ProjectParameters.json
```diff
@@ -0,0 +1,160 @@
...
+    "loads_process_list": [{
+        "python_module": "apply_normal_load_table_process",
+        "kratos_module": "KratosMultiphysics.GeoMechanicsApplication",
+        "process_name":  "ApplyNormalLoadTableProcess",
+        "Parameters":    {
+            "model_part_name": "PorousDomain.LoadTop",
+            "variable_name":   "NORMAL_CONTACT_STRESS",
+            "active":          [true,false],
+            "table":           [0,0],
+            "fluid_pressure_type": "Uniform",
```
WPK4FEM, line 154: Doesn't belong here?

applications/GeoMechanicsApplication/tests/test_conditions/test_LineNormalLoadDiffOrderCondition2D5N/ProjectParameters.json
```diff
@@ -0,0 +1,160 @@
...
+    "loads_process_list": [{
+        "python_module": "apply_normal_load_table_process",
+        "kratos_module": "KratosMultiphysics.GeoMechanicsApplication",
+        "process_name":  "ApplyNormalLoadTableProcess",
+        "Parameters":    {
+            "model_part_name": "PorousDomain.LoadTop",
+            "variable_name":   "NORMAL_CONTACT_STRESS",
+            "active":          [true,false],
+            "table":           [0,0],
+            "fluid_pressure_type": "Uniform",
```
WPK4FEM, line 154: remove?

applications/GeoMechanicsApplication/tests/test_conditions/test_UPwNormalFaceLoadCondition2D4N/ProjectParameters.json
```diff
@@ -0,0 +1,160 @@
...
+    "loads_process_list": [{
+        "python_module": "apply_normal_load_table_process",
+        "kratos_module": "KratosMultiphysics.GeoMechanicsApplication",
+        "process_name":  "ApplyNormalLoadTableProcess",
+        "Parameters":    {
+            "model_part_name": "PorousDomain.LoadTop",
+            "variable_name":   "NORMAL_CONTACT_STRESS",
+            "active":          [true,false],
+            "table":           [0,0],
+            "fluid_pressure_type": "Uniform",
```
WPK4FEM, line 154: remove?

applications/GeoMechanicsApplication/tests/test_conditions/test_UPwNormalFaceLoadCondition2D3N/ProjectParameters.json
```diff
@@ -0,0 +1,160 @@
...
+    "loads_process_list": [{
+        "python_module": "apply_normal_load_table_process",
+        "kratos_module": "KratosMultiphysics.GeoMechanicsApplication",
+        "process_name":  "ApplyNormalLoadTableProcess",
+        "Parameters":    {
+            "model_part_name": "PorousDomain.LoadTop",
+            "variable_name":   "NORMAL_CONTACT_STRESS",
+            "active":          [true,false],
+            "table":           [0,0],
+            "fluid_pressure_type": "Uniform",
```
WPK4FEM, line 154: Remove?

applications/GeoMechanicsApplication/tests/test_conditions/test_UPwNormalFaceLoadCondition2D2N/ProjectParameters.json
```diff
@@ -0,0 +1,160 @@
...
+    "loads_process_list": [{
+        "python_module": "apply_normal_load_table_process",
+        "kratos_module": "KratosMultiphysics.GeoMechanicsApplication",
+        "process_name":  "ApplyNormalLoadTableProcess",
+        "Parameters":    {
+            "model_part_name": "PorousDomain.LoadTop",
+            "variable_name":   "NORMAL_CONTACT_STRESS",
+            "active":          [true,false],
+            "table":           [0,0],
+            "fluid_pressure_type": "Uniform",
```
WPK4FEM, line 154: remove?

WPK4FEM (commented):
Thank you Mohamed and Gennady for all improvements.
Some small things left.

applications/GeoMechanicsApplication/tests/test_conditions.py
```diff
@@ -7,9 +7,9 @@ class KratosGeoMechanicsConditionTests(KratosUnittest.TestCase):
     """
     This class contains benchmark tests which are checked with the regression on a previously obtained value.
     """
-    etalon_value1 = -76430.423509075
+    etalon_value1 = -86430.42350907
     etalon_value2 = -0.0005
-    etalon_value3 = 1356.95764908908
+    etalon_value3 = -8643.04235086
```
rfaasse, line 12: What is the reason these values changed? It also seems etalon_value1 = 10*etalon_value3, while before there was no such relation, just double checking this is an intended change?

applications/GeoMechanicsApplication/tests/test_conditions.py
```diff
@@ -19,228 +19,89 @@ def tearDown(self):
...
-        disp = displacement[2]
-        self.assertAlmostEqual(self.etalon_value2, disp[1])
-        
-    def test_AxisymmetricUPwNormalFaceLoadCondition2D5N(self):
-        test_name = 'test_AxisymmetricUPwNormalFaceLoadCondition2D5N'
-        file_path = test_helper.get_file_path(os.path.join('test_conditions', test_name))
-        simulation = test_helper.run_kratos(file_path)
-        displacement = test_helper.get_displacement(simulation)
-        disp = displacement[2]
-        self.assertAlmostEqual(self.etalon_value2, disp[1])
```
rfaasse: Is it correct that these two tests are removed? They are not diff order, so shouldn't they remain?

applications/GeoMechanicsApplication/tests/test_conditions/README.md
```diff
@@ -0,0 +1,54 @@
+# Load conditions
+
+This test verifies the integrity of the load and fluid flux conditions for low and high order elements.
+The following conditions are tested.
+
+- AxisymmetricUPwNormalFaceLoadCondition2D2N
+- AxisymmetricUPwNormalFaceLoadCondition2D3N
+- AxisymmetricUPwNormalFaceLoadCondition2D4N
+- AxisymmetricUPwNormalFaceLoadCondition2D5N
```
rfaasse, line 9: These two tests seem to be removed, could you check if that's correct and if the rest of the list is up to date?

applications/GeoMechanicsApplication/tests/test_conditions/README.md
```diff
@@ -0,0 +1,54 @@
...
+- PwNormalFluxCondition2D5N
+- UPwNormalFluxCondition2D2N
+- UPwNormalFluxCondition2D3N
+- UPwNormalFluxCondition2D4N
+- UPwNormalFluxCondition2D5N
+
+
+## Setup
+
+This tests consist of 3, 6, 10 and 15-noded elements wuth 2, 3, 4, 5 noded line conditions, respectively. For all cases, a rectangular domain of $1 \mathrm{m} \times 2 \mathrm{m}$, with 8 triangular elements is used. A schematic representation can be found in the figure below:
```
rfaasse: Small typo
```suggestion
This tests consist of 3, 6, 10 and 15-noded elements with 2, 3, 4, 5 noded line conditions, respectively. For all cases, a rectangular domain of $1 \mathrm{m} \times 2 \mathrm{m}$, with 8 triangular elements is used. A schematic representation can be found in the figure below:
```

applications/GeoMechanicsApplication/tests/test_conditions/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+The material is described using:
+-   A linear elastic plane strain model (LinearElasticPlaneStrain2DLaw)
+-   A Young's modulus $E = 2 \times 10^{6} \mathrm{[Pa]}$ with Poisson's ratio $\nu = 0.0 \mathrm{[-]}$.
+-   The soil and water density are specified, but irrelevant due to the absence of gravity. The porosity is 0.
+-   The dynamic viscosity of water is given as $\mu = 10^{-3} \mathrm{[Pa \cdot s]}$ and the intrinsic permeability of the soil as $\kappa = 1.157 \times 10^{-12} \mathrm{[m^2]}$
+-   The bulk modulus of water $K_w = 2 \times 10^6 \mathrm{[Pa]}$
+
+## Assertions
+The tests assert that the vertical distribution of displacement and pressure matches the analytical distribution, which is a linear distribution over the height of the column.
```
rfaasse: We only check one displacement value if I'm correct, so I'm not sure if we assert the vertical distribution here? Also, I might've missed it, but I didn't see an analytical distribution in the test, so we might need to remove that from the readme.

applications/GeoMechanicsApplication/tests/test_conditions/documentation_data/load_flux_domain.svg
```diff
@@ -0,0 +1 @@
...001 417.167 70.7001 417.167 19.5001ZM425.167 69.3667 418.5 82.7001 411.833 69.3667Z" fill="#4472C4"/><path d="M467.833 19.5001 467.833 70.7001 465.167 70.7001 465.167 19.5001ZM473.167 69.3667 466.5 82.7001 459.833 69.3667Z" fill="#4472C4"/><path d="M515.833 19.5001 515.833 70.7001 513.167 70.7001 513.167 19.5001ZM521.167 69.3667 514.5 82.7001 507.833 69.3667Z" fill="#4472C4"/><path d="M563.833 19.5001 563.833 70.7001 561.167 70.7001 561.167 19.5001ZM569.167 69.3667 562.5 82.7001 555.833 69.3667Z" fill="#4472C4"/><path d="M611.833 19.5001 611.833 70.7001 609.167 70.7001 609.167 19.5001ZM617.167 69.3667 610.5 82.7001 603.833 69.3667Z" fill="#4472C4"/><text font-family="Calibri,Calibri_MSFontService,sans-serif" font-style="italic" font-weight="700" font-size="24" transform="translate(261.847 46)">q</text><text font-family="Calibri,Calibri_MSFontService,sans-serif" font-weight="400" font-size="24" transform="translate(280.013 46)">[N]</text><path d="M322 19 608.839 19.0001" stroke="#4472C4" stroke-width="2" stroke-miterlimit="8" fill="none" fill-rule="evenodd"/><path d="M321 666 609 666" stroke="#000000" stroke-width="2" stroke-miterlimit="8" fill="none" fill-rule="evenodd"/></g></svg>
```
WPK4FEM: The unit of a line load is force / length, for a face load force / area. I suggest to remove the units from the picture and leave those to the readme document.

applications/GeoMechanicsApplication/tests/test_GeoMechanicsApplication.py
rfaasse: I'm not sure, how close we already are to the limit, do you know @avdg81?

applications/GeoMechanicsApplication/custom_conditions/U_Pw_face_load_interface_condition.cpp
rfaasse: As discussed, this will be done in a separate PR

applications/GeoMechanicsApplication/custom_conditions/axisymmetric_U_Pw_normal_face_load_condition.cpp
rfaasse: I do not have enough knowledge about this topic to know whether 5 noded elements with a 2 point gauss integration method are common/feasible for now. Do you know @WPK4FEM?

applications/GeoMechanicsApplication/geo_mechanics_application.h
rfaasse: See other comment

applications/GeoMechanicsApplication/tests/test_conditions.py
rfaasse: Thank you @mnabideltares for adding the readme, it helps a lot in understanding the tests!

applications/GeoMechanicsApplication/tests/test_conditions/test_AxisymmetricUPwNormalFaceLoadCondition2D4N/ProjectParameters.json
WPK4FEM: Sorry, this is my mistake. The process needs the line with fluid_pressure_type. All the repeated remarks about this should be ignored.

rfaasse (commented):
Thank you for your hard work in incorporating these changes @mnabideltares and @markelov208! This PR improves the code quality a lot!

There are a couple of minor suggestions and questions left from my side and there is a (very minor) merge conflict with master that still needs to be fixed (it's about added includes that should just both be included), but I think we're quite close now to getting this done!

applications/GeoMechanicsApplication/tests/test_conditions.py
mnabideltares: In order to test a single process, we decided to remove the gravity, to measure only the effect of the flux. 
Previously, the pressure was superpositioned by two compenet, the hydrostatic pressure and the pressure which gets imposed by flux from the boundary. After removing the gravity, the pressure is purely the pressure by the boundary.
I imposed two fluxes 1e-4 and 1e-5, related to test case, so the pressure of one is 10 times of the other.

applications/GeoMechanicsApplication/tests/test_conditions.py
mnabideltares: Not intended to be removed. Probably were removed by accident. I brought them back

applications/GeoMechanicsApplication/tests/test_conditions/test_LineNormalFluidFluxDiffOrderCondition2D3N/ProjectParameters.json
WPK4FEM: Was changed to uniform Pw in the tests.

applications/GeoMechanicsApplication/custom_conditions/axisymmetric_U_Pw_normal_face_load_condition.cpp
mnabideltares: It is fixed, I found a node ordering mistake in the input file. The code is fixed:

    switch (TNumNodes) {
    case 2:
    case 3:
        GI_GAUSS = GeometryData::IntegrationMethod::GI_GAUSS_2;
        break;
    case 4:
        GI_GAUSS = GeometryData::IntegrationMethod::GI_GAUSS_3;
        break;
    case 5:
        GI_GAUSS = GeometryData::IntegrationMethod::GI_GAUSS_5;
        break;
    default:
        GI_GAUSS = GeometryData::IntegrationMethod::GI_GAUSS_2;
        break;
    }

applications/GeoMechanicsApplication/tests/test_GeoMechanicsApplication.py
avdg81: I'm not sure either. Perhaps we can see that if we look at the results of the GitHub pipelines?

WPK4FEM (approved):
My stack of remarks has been processed.

rfaasse (approved):
Thank you for your hard work, looks good to me!

PR #11639, opened 2023-10-01
[GeoMechanicsApplication] Add an adaptive time incrementor

### LINKED ISSUE #11619: [GeoMechanicsApplication] Add an abstract interface for incrementing time and a concrete adaptive time incrementor
As a developer, I would like to have an abstract interface for incrementing time that can be used by other components. Furthermore, I would like to have a concrete implementation of the currently used adaptive time incrementing scheme.

#### Background

This issue aims to provide a general interface for incrementing time, that will be used by a time loop executor. This interface shall be abstract. In addition, we aim to provide a concrete implementation that mimics the current implementation of adaptive time incrementing.

#### Acceptance Criteria

**Given** a time reduction factor that is not in the range <0, 1]
**When** the developer attempts to construct an adaptive time incrementor
**Then** its constructor must throw an exception.

**Given** a time increase factor that is not greater than or equal to 1.0
**When** the developer attempts to construct an adaptive time incrementor
**Then** its constructor must throw an exception.

**Given** a maximum number of iterations that equals 0
**When** the developer attempts to construct an adaptive time incrementor
**Then** its constructor must throw an exception.

**Given** a minimum number of iterations that is not in the range [0, maximum_num_of_iterations>
**When** the developer attempts to construct an adaptive time incrementor
**Then** its constructor must throw an exception.

**Given** a maximum number of cycles that equals 0
**When** the developer attempts to construct an adaptive time incrementor
**Then** its constructor must throw an exception.

**Given** a start increment size that is non-positive (i.e. in the range <,0])
**When** the developer attempts to construct an adaptive time incrementor
**Then** its constructor must throw an exception.

**Given** an adaptive time incrementor
**When** the developer queries whether a next step should be carried out (by calling member function `WantNextStep(...)` and the end time has not been attained yet
**Then** the query must return true.

**Given** an adaptive time incrementor
**When** the developer queries whether a next step should be carried out (by calling member function `WantNextStep(...)` and the end time has been exceeded
**Then** the query must return false.

**Given** an adaptive time incrementor
**When** the developer queries whether a next step should be carried out (by calling member function `WantNextStep(...)` and we are exactly at the end time
**Then** the query must return false.

**Given** an adaptive time incrementor
**When** the developer queries whether the step should be retried (by calling member function `WantRetryStep(...)` and the cycle number equals zero (i.e. no attempt has been made yet)
**Then** the query must return true.

**Given** an adaptive time incrementor and a previous step state that indicates convergence has not been achieved yet
**When** the developer queries whether the step should be retried (by calling member function `WantRetryStep(...)` and there is room for at least one more cycle
**Then** the query must return true.

**Given** an adaptive time incrementor and a previous step state that indicates convergence has not been achieved yet
**When** the developer queries whether the step should be retried (by calling member function `WantRetryStep(...)` and there are no more cycles left
**Then** the query must return false.

**Given** an adaptive time incrementor and a previous step state that indicates convergence has been achieved
**When** the developer queries whether the step should be retried (by calling member function `WantRetryStep(...)`
**Then** the query must return false.

**Given** an adaptive time incrementor with a start increment that does not result in exceeding the end time
**When** the developer queries the time increment to be adopted (by calling member function `GetIncrement(...)`
**Then** the query must return the start increment.

**Given** an adaptive time incrementor with a start increment that does result in exceeding the end time
**When** the developer queries the time increment to be adopted (by calling member function `GetIncrement(...)`
**Then** the query must return the increment such that we exactly attain the end time.

**Given** an adaptive time incrementor with a start increment and a previous step state that indicates that it did not converge
**When** the developer queries the time increment to be adopted (by calling member function `GetIncrement(...)`
**Then** the query must return a reduced increment corresponding to the given reduction factor.

**Given** an adaptive time incrementor with a start increment and N failed cycles (i.e. no convergence was found)
**When** the developer queries the time increment to be adopted (by calling member function `GetIncrement(...)`
**Then** the query must return an increment that has been reduced N times by the given reduction factor.

**Given** an adaptive time incrementor with a start increment and a converged step state that required the maximum number of iterations
**When** the developer queries the time increment to be adopted (by calling member function `GetIncrement(...)`
**Then** the query must return an increment that has been reduced by the given reduction factor.

**Given** an adaptive time incrementor and a converged step state that required less than the minimum number of iterations
**When** the developer queries the time increment to be adopted (by calling member function `GetIncrement(...)` and the "distance" to the end time is larger than last increment size multiplied by the increase factor
**Then** the query must return an increment that has been increased by the given increase factor.

**Given** an adaptive time incrementor and a converged step state that required less than the minimum number of iterations
**When** the developer queries the time increment to be adopted (by calling member function `GetIncrement(...)` and the "distance" to the end time is less than the last increment size multiplied by the increase factor
**Then** the query must return an increment that results in attaining the end time.

avdg81:
**📝 Description**
This PR adds an adaptive time incrementor to be used by the C++ custom workflows for geomechanics. It is a time incrementor that starts with a prescribed increment size, which may be adjusted as needed. That is, when no convergence is achieved, the applied time increment will be reduced on each failed attempt. (The maximum number of attempt, aka cycles, is limited by the user.) Conversely, when convergence took only a few iterations, the time increment will be increased.

### DIFF
--- a/applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.cpp
+++ b/applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.cpp
@@ -1,0 +1,87 @@
+// KRATOS___
+//     //   ) )
+//    //         ___      ___
+//   //  ____  //___) ) //   ) )
+//  //    / / //       //   / /
+// ((____/ / ((____   ((___/ /  MECHANICS
+//
+//  License:         geo_mechanics_application/license.txt
+//
+//  Main authors:    Wijtze Pieter Kikstra
+//                   Anne van de Graaf
+//
+
+#include "adaptive_time_incrementor.h"
+#include "time_step_end_state.hpp"
+
+#include "includes/exception.h"
+
+
+namespace Kratos
+{
+
+AdaptiveTimeIncrementor::AdaptiveTimeIncrementor(double      StartTime,
+                                                 double      EndTime,
+                                                 double      StartIncrement,
+                                                 std::size_t MaxNumOfCycles,
+                                                 double      ReductionFactor,
+                                                 double      IncreaseFactor,
+                                                 std::size_t MinNumOfIterations,
+                                                 std::size_t MaxNumOfIterations) :
+    TimeIncrementor(),
+    mEndTime(EndTime),
+    mDeltaTime(std::min(StartIncrement, EndTime - StartTime)), // avoid exceeding the end time
+    mMaxNumOfCycles(MaxNumOfCycles),
+    mReductionFactor(ReductionFactor),
+    mIncreaseFactor(IncreaseFactor),
+    mMinNumOfIterations(MinNumOfIterations),
+    mMaxNumOfIterations(MaxNumOfIterations)
+{
+    KRATOS_ERROR_IF(StartTime >= mEndTime) << "Start time (" << StartTime << ") must be smaller than end time (" << mEndTime << ")";
+    KRATOS_ERROR_IF(mDeltaTime <= 0.0) << "Start increment must be positive, but got " << mDeltaTime;
+    KRATOS_ERROR_IF(mMaxNumOfCycles == std::size_t(0)) << "Maximum number of cycles must be positive";
+    KRATOS_ERROR_IF(mMinNumOfIterations >= mMaxNumOfIterations) << "Minimum number of iterations ("
+                                                                << mMinNumOfIterations
+                                                                << ") is not less than maximum number of iterations ("
+                                                                << mMaxNumOfIterations << ")";
+    KRATOS_ERROR_IF(mReductionFactor >  1.0) << "Reduction factor must not be greater than 1, but got " << mReductionFactor;
+    KRATOS_ERROR_IF(mReductionFactor <= 0.0) << "Reduction factor must be positive, but got " << mReductionFactor;
+    KRATOS_ERROR_IF(mIncreaseFactor  <  1.0) << "Increase factor must be greater than or equal to 1, but got " << mIncreaseFactor;
+}
+
+bool AdaptiveTimeIncrementor::WantNextStep(const TimeStepEndState& rPreviousState) const
+{
+    return rPreviousState.time < mEndTime;
+}
+
+bool AdaptiveTimeIncrementor::WantRetryStep(std::size_t             CycleNumber,
+                                            const TimeStepEndState& rPreviousState) const
+{
+    if (CycleNumber == 0) return true;  // always carry out a first attempt
+
+    if (rPreviousState.Converged()) return false;  // the time step is done
+
+    return CycleNumber < mMaxNumOfCycles;  // stopping criterion
+}
+
+double AdaptiveTimeIncrementor::GetIncrement() const
+{
+    return mDeltaTime;
+}
+
+void AdaptiveTimeIncrementor::PostTimeStepExecution(const TimeStepEndState& rResultantState)
+{
+    if (rResultantState.NonConverged() ||
+        (rResultantState.Converged() && (rResultantState.num_of_iterations == mMaxNumOfIterations))) {
+        mDeltaTime *= mReductionFactor;
+    }
+    else if (rResultantState.Converged() &&
+             (rResultantState.num_of_iterations < mMinNumOfIterations)) {
+        mDeltaTime *= mIncreaseFactor;
+    }
+
+    // Avoid incrementing the time beyond the end time
+    mDeltaTime = std::min(mDeltaTime, mEndTime - rResultantState.time);
+}
+
+}
--- a/applications/GeoMechanicsApplication/tests/cpp_tests/test_time_incrementor.cpp
+++ b/applications/GeoMechanicsApplication/tests/cpp_tests/test_time_incrementor.cpp
@@ -1,0 +1,389 @@
+// KRATOS___
+//     //   ) )
+//    //         ___      ___
+//   //  ____  //___) ) //   ) )
+//  //    / / //       //   / /
+// ((____/ / ((____   ((___/ /  MECHANICS
+//
+//  License:         geo_mechanics_application/license.txt
+//
+//  Main authors:    Wijtze Pieter Kikstra
+//                   Anne van de Graaf
+//
+
+#include "testing/testing.h"
+#include "custom_workflows/adaptive_time_incrementor.h"
+#include "custom_workflows/time_step_end_state.hpp"
+
+using namespace Kratos;
+
+
+namespace
+{
+
+struct AdaptiveTimeIncrementorSettings
+{
+    double      StartTime{0.0};
+    double      EndTime{8.0};
+    double      StartIncrement{0.5};
+    std::size_t MaxNumOfCycles{8};
+    double      ReductionFactor{0.5};
+    double      IncreaseFactor{2.0};
+    std::size_t MinNumOfIterations{3};
+    std::size_t MaxNumOfIterations{15};
+};
+
+AdaptiveTimeIncrementor MakeAdaptiveTimeIncrementor(const AdaptiveTimeIncrementorSettings& rSettings)
+{
+    return AdaptiveTimeIncrementor{rSettings.StartTime,
+                                   rSettings.EndTime,
+                                   rSettings.StartIncrement,
+                                   rSettings.MaxNumOfCycles,
+                                   rSettings.ReductionFactor,
+                                   rSettings.IncreaseFactor,
+                                   rSettings.MinNumOfIterations,
+                                   rSettings.MaxNumOfIterations};
+}
+
+}
+
+
+namespace Kratos::Testing
+{
+
+KRATOS_TEST_CASE_IN_SUITE(AdaptiveTimeIncrementorThrowsWhenStartTimeExceedsEndTime, KratosGeoMechanicsFastSuite)
+{
+    AdaptiveTimeIncrementorSettings settings;
+    settings.EndTime = settings.StartTime - 2.0;
+
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(MakeAdaptiveTimeIncrementor(settings),
+                                      "Start time (0) must be smaller than end time (-2)")
+}
+
+KRATOS_TEST_CASE_IN_SUITE(AdaptiveTimeIncrementorThrowsWhenStartTimeEqualsEndTime, KratosGeoMechanicsFastSuite)
+{
+    AdaptiveTimeIncrementorSettings settings;
+    settings.EndTime = settings.StartTime;
+
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(MakeAdaptiveTimeIncrementor(settings),
+                                      "Start time (0) must be smaller than end time (0)")
+}
+
+KRATOS_TEST_CASE_IN_SUITE(AdaptiveTimeIncrementorThrowsWhenStartIncrementIsNegative, KratosGeoMechanicsFastSuite)
+{
+    AdaptiveTimeIncrementorSettings settings;
+    settings.StartIncrement = -0.5;
+
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(MakeAdaptiveTimeIncrementor(settings),
+                                      "Start increment must be positive, but got -0.5")
+}
+
+KRATOS_TEST_CASE_IN_SUITE(AdaptiveTimeIncrementorThrowsWhenStartIncrementEqualsZero, KratosGeoMechanicsFastSuite)
+{
+    AdaptiveTimeIncrementorSettings settings;
+    settings.StartIncrement = 0.0;
+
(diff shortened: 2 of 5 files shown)


applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.hpp
```diff
@@ -0,0 +1,50 @@
...
+//  //    / / //       //   / /
+// ((____/ / ((____   ((___/ /  MECHANICS
+//
+//  License:         geo_mechanics_application/license.txt
+//
+//  Main authors:    Wijtze Pieter Kikstra
+//                   Anne van de Graaf
+//
+
+#pragma once
```
rfaasse: Just for my information: When do we use .hpp and when do we use .h in this project?

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_incrementor.cpp
```diff
@@ -0,0 +1,439 @@
...
+KRATOS_TEST_CASE_IN_SUITE(AdaptiveTimeIncrementorThrowsWhenMinNumberOfIterationsIsGreaterThanMaxNumberOfIterations, KratosGeoMechanicsFastSuite)
+{
+    const auto start_time            = 0.0;
+    const auto end_time              = start_time + 8.0;
+    const auto start_increment       = 0.5;
+    const auto max_num_of_cycles     = std::size_t{8};
+    const auto reduction_factor      = 0.5;
+    const auto increase_factor       = 2.0;
+    const auto max_num_of_iterations = std::size_t{10};
+    const auto min_num_of_iterations = max_num_of_iterations + 1;
```
rfaasse: I think it would be nice to have valid defaults for the inputs, such that you don't have to specify all the valid inputs every test (and it's clear in the test which variable is specifically tested by which test).

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_incrementor.cpp
```diff
@@ -0,0 +1,439 @@
...
+    const auto start_increment   = 0.5;
+    const auto time_incrementor  = AdaptiveTimeIncrementor{start_time, end_time, start_increment};
+    auto       previous_state    = TimeStepEndState{};
+    previous_state.time          = end_time + 1.0;
+
+    KRATOS_EXPECT_FALSE(time_incrementor.WantNextStep(previous_state))
+}
+
+KRATOS_TEST_CASE_IN_SUITE(NoNextTimeStepWhenAtEndTime, KratosGeoMechanicsFastSuite)
+{
```
rfaasse, line 240: I think we just miss one test case, where the previous_state is before the end time and previous_state + start_increment is after the end time (i.e. the last step would push it over the end time)

rfaasse (changes requested):
Very clear class and test cases! The only real suggestion I have is that I think the test cases could be a bit more readable by having default valid values for the inputs (so you don't have to specify them each test and only focus on the ones that are specificall tested).

applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.cpp
```diff
@@ -0,0 +1,87 @@
...
+    mDeltaTime(std::min(StartIncrement, EndTime - StartTime)), // avoid exceeding the end time
+    mMaxNumOfCycles(MaxNumOfCycles),
+    mReductionFactor(ReductionFactor),
+    mIncreaseFactor(IncreaseFactor),
+    mMinNumOfIterations(MinNumOfIterations),
+    mMaxNumOfIterations(MaxNumOfIterations)
+{
+    KRATOS_ERROR_IF(StartTime >= mEndTime) << "Start time (" << StartTime << ") must be smaller than end time (" << mEndTime << ")";
+    KRATOS_ERROR_IF(mDeltaTime <= 0.0) << "Start increment must be positive, but got " << mDeltaTime;
+    KRATOS_ERROR_IF(mMaxNumOfCycles == std::size_t(0)) << "Maximum number of cycles must not be equal to 0";
```
WPK4FEM: This is more programmers than users thinking. I would expect a text: Maximum number of cycles must be positive.

applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.cpp
```diff
@@ -0,0 +1,87 @@
...
+}
+
+bool AdaptiveTimeIncrementor::WantRetryStep(std::size_t             CycleNumber,
+                                            const TimeStepEndState& rPreviousState) const
+{
+    if (CycleNumber == 0) return true;
+
+    if (rPreviousState.Converged()) return false;
+
+    return CycleNumber < mMaxNumOfCycles;
```
WPK4FEM: This may not be self explaining enough, please add some comment.
ie. CycleNumber == 0 starts the cycling process
rPreviousState.Converged() means the step is done.
CycleNumber < mMaxNumOfCycles is the stopping criterion.

applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+double AdaptiveTimeIncrementor::GetIncrement() const
+{
+    return mDeltaTime;
+}
+
+void AdaptiveTimeIncrementor::PostTimeStepExecution(const TimeStepEndState& rResultantState)
+{
+    if (rResultantState.NonConverged() ||
+        (rResultantState.Converged() && (rResultantState.num_of_iterations == mMaxNumOfIterations))) {
```
WPK4FEM, line 75: If only 2 convergence states ( Converged, NonConverged ) are possible, the second condition only needs to be rResultantState.num_of_iterations == mMaxNumOfIterations.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_incrementor.cpp
```diff
@@ -0,0 +1,439 @@
...
+
+KRATOS_TEST_CASE_IN_SUITE(AdaptiveTimeIncrementorThrowsWhenMaxNumberOfCyclesEqualsZero, KratosGeoMechanicsFastSuite)
+{
+    const auto start_time        = 0.0;
+    const auto end_time          = start_time + 8.0;
+    const auto start_increment   = 0.5;
+    const auto max_num_of_cycles = std::size_t{0};
+
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(AdaptiveTimeIncrementor(start_time, end_time, start_increment, max_num_of_cycles),
+                                      "Maximum number of cycles must not be equal to 0")
```
WPK4FEM: See comment about failure text in adaptive_time_incrementor

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_incrementor.cpp
```diff
@@ -0,0 +1,439 @@
...
+KRATOS_TEST_CASE_IN_SUITE(DontRetryWhenPreviousAttemptConverged, KratosGeoMechanicsFastSuite)
+{
+    const auto start_time        = 0.0;
+    const auto end_time          = start_time + 8.0;
+    const auto start_increment   = 0.5;
+    const auto max_num_of_cycles = std::size_t{4};
+    const auto time_incrementor  = AdaptiveTimeIncrementor{start_time, end_time, start_increment, max_num_of_cycles};
+    const auto cycle_number      = std::size_t{1};
+    auto previous_state          = TimeStepEndState{};
+    previous_state.convergence_state = TimeStepEndState::ConvergenceState::converged;
```
WPK4FEM: Test is correct, but the practical fill in is unexpected. cycle_Number == 1, a previous cycle does is sort of a starting state which I expect to have convergence_state NonConverged. Here it is set to Converged. Maybe set cycle_number > 1 to avoid this confusion.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_incrementor.cpp
```diff
@@ -0,0 +1,439 @@
...
+    const auto start_time        = 0.0;
+    const auto end_time          = start_time + 8.0;
+    const auto start_increment   = 0.5;
+    const auto max_num_of_cycles = std::size_t{4};
+    const auto reduction_factor  = 0.6;
+    auto time_incrementor        = AdaptiveTimeIncrementor{start_time, end_time, start_increment, max_num_of_cycles, reduction_factor};
+    auto previous_state          = TimeStepEndState{};
+    previous_state.convergence_state = TimeStepEndState::ConvergenceState::non_converged;
+
+    time_incrementor.PostTimeStepExecution(previous_state); // previous attempt did not converge
```
WPK4FEM: unexpected comment. PostTimeStepExecution processes the end state and sets up for the new cycle or increment.

WPK4FEM (commented):
Not much to comment about from my side.

applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.hpp
avdg81: To the best of my knowledge, there is no clear policy about the file extension to be used for header files. Let's first double-check the Kratos Style guide, just to make sure that we didn't overlook it. If it is not stated there, we may want to define a policy in our supplementary coding conventions.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_incrementor.cpp
avdg81: Agreed. I have added a `struct` to hold the time incrementor's settings. A default-constructed instance of that `struct` constitutes a valid collection of settings. Furthermore, I have added a convenience function to build an adaptive time incrementor given a settings object. Most tests are now more compact and to the point. Thanks for the suggestion!

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_incrementor.cpp
avdg81: I may misunderstand you, but I believe this is test case `ReduceIncrementToAvoidExceedingEndTime`. Could you please verify whether that one covers what you had in mind? Otherwise, let's discuss what you meant and add a corresponding unit test. Thanks!

rfaasse: So in that case, the 'WantNextStep' function should return true? Maybe we should add that assert to the `ReduceIncrementToAvoidExceedingEndTime` test then? Just checking if this specific function returns the correct value in that case

applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.hpp
rfaasse: I don't see anything in the Kratos Style guide, so let's indeed discuss later

applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.cpp
avdg81: Agreed. I have updated the text according to your suggestion.

applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.cpp
avdg81: Done.

applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.cpp
avdg81: I can imagine that at some point in time we may want to add a third state: `Diverged` (this was also the reason I did not opt for a simple `bool` flag). From that point of view, I'd rather be explicit about the set of the conditions that define a particular state to be handled. Therefore, I'd rather keep it the way it is, unless you can convince me to change it :-)

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_incrementor.cpp
avdg81: Done.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_incrementor.cpp
avdg81: Cycle number 0 is the first attempt, where it may or may not converge. The test reflects a case where convergence was achieved after this first attempt. As discussed, the confusion is due an off-by-one error.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_incrementor.cpp
avdg81: Done.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_incrementor.cpp
rfaasse: Resolving this comment, as the test case was added :)

rfaasse (approved):
Looks good to me!

PR #11647, opened 2023-10-02
Added custom_docker_scripts 

mcgicjn2:
including configuration for Deltares TeamCity build GCC Linux configuration

**📝 Description**
This is a script injected into our process to reduce Deltares GCC builds to only those needed for GeoMechanics application.

i.e.:

- GeoMechanicsAplication
- StructuralMechanicsApplication
- LinearSolverApplication

The teamcity build configuration has been added to copy this script rather than the general one in github actions (https://dpcbuild.deltares.nl/buildConfiguration/GEOFEA_KratosGeo_Docker_GccFullDebug?buildTypeTab=overview&mode=builds)

**🆕 Changelog**
Added custom configuration script

### DIFF
--- a/applications/GeoMechanicsApplication/custom_docker_scripts/linux/configure.sh
+++ b/applications/GeoMechanicsApplication/custom_docker_scripts/linux/configure.sh
@@ -23,6 +23,7 @@ export KRATOS_INSTALL_PYTHON_USING_LINKS=ON
 add_app ${KRATOS_APP_DIR}/LinearSolversApplication;
 add_app ${KRATOS_APP_DIR}/StructuralMechanicsApplication;
 add_app ${KRATOS_APP_DIR}/GeoMechanicsApplication;
+add_app ${KRATOS_APP_DIR}/ConstitutiveLawsApplication;
 
 # Clean
 clear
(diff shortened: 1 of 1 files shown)


sunethwarna:
I am not sure about putting applications specific build configuration to the main repo because then it may end up with endless number of scripts. 

mcgicjn2:
@sunethwarna @roigcarlo This is a reduced copy of the .git workflow for the GCC build. It allows us to run the builds on our server to prevent (and reduce our need for draft PRs) errors at PRs. 

However I think we can inject these from another location on our own system. 

I will try to inject it at our end and close if we achieve this.

roigcarlo:
I think we should distinguish two cases:

1 - Scripts in the main `scripts` folder: There I agree with @sunethwarna 
2 - Scripts in the app:  There I see no problem, we have applications that have full libraries and scripts everywhere so its hard to argue that this is a problem 

Since this falls into the later, for me its ok,

sunethwarna:
Ok, I saw this as in the main scripts folder. I am also ok if this is in the application ( which is the case)

Sorry for the noise.

roigcarlo:
Ok then @mcgicjn2 if you make it a PR I will approve.

mcgicjn2:
Thanks for the input all, we decided to inject the scripts from our own repo, in this way the repo remains cleaner. Cheers for the discussion it actually advanced an internal discussion on how to handle such issues.

PR #11649, opened 2023-10-03
Geo/add time step executor

### LINKED ISSUE #11617: [GeoMechanicsApplication] Add a class for carrying out a single time step
As a developer I would like to be able to carry out a single time step, given the solver strategy and a list of processes to take into account.

**Background**

The Kratos team at Deltares would like to integrate time stepping into their custom workflows in a "modular" fashion. By "modular" we mean that we should be able to customize the time stepping by making various compositions of components rather than by modifying a monolithic block that includes all aspects of time stepping. Carrying out the time step at any given time and for the given solver strategy and a list of processes is one of those components.

**Acceptance criteria**

**Given** the solver strategy does not converge
**When** the developer carries out the time step (by calling the `TimeStepExecuter<...>::Run(...)` member function)
**Then** the returned time step state should indicate that it did not converge.

**Given** the solver strategy converges
**When** the developer carries out the time step (by calling the `TimeStepExecuter<...>::Run(...)` member function)
**Then** the returned time step state should indicate that it converged.

**Given** the solver strategy converges and a list containing a single process
**When** the developer carries out the time step (by calling the `TimeStepExecuter<...>::Run(...)` member function)
**Then** the process's member functions `ExecuteInitializeSolutionStep()` and `ExecuteFinalizeSolutionStep()` each must have been called exactly once.

**Given** the solver strategy converges
**When** the developer carries out the time step (by calling the `TimeStepExecuter<...>::Run(...)` member function)
**Then** the solver strategy's member functions `Initialize()`, `InitializeSolutionStep()`, `Predict()`, `SolveSolutionStep()`, and `FinalizeSolutionStep()` each must have been called exactly once.

**Given** the solver strategy converges in N iterations
**When** the developer carries out the time step (by calling the `TimeStepExecuter<...>::Run(...)` member function)
**Then** the returned time step state should indicate that the convergence took N iterations.

**Given** the solver strategy converges for time T
**When** the developer carries out the time step (by calling the `TimeStepExecuter<...>::Run(...)` member function)
**Then** the returned time step state should indicate that the corresponding time equals T.

**Given** the solver strategy does not converge for time T
**When** the developer carries out the time step (by calling the `TimeStepExecuter<...>::Run(...)` member function)
**Then** the returned time step state should indicate that the corresponding time equals T.

WPK4FEM:
An executor to perform a single time step for GeoMechanical Analysis

It uses:
- a solving strategy
- a list of process references

and gives back 
- a time step end state containing the time, the convergence state and the number of iterations used.

7 unit tests for checking basic behavior are added.


applications/GeoMechanicsApplication/custom_workflows/time_stepping.hpp
```diff
@@ -0,0 +1,77 @@
+// KRATOS___
+//     //   ) )
```
rfaasse: Since this is no longer a template class, We should move the implementations to a cpp file and just have the declarations in the header file

applications/GeoMechanicsApplication/custom_workflows/time_stepping.hpp
```diff
@@ -0,0 +1,77 @@
...
+        result.time = Time;
+        result.convergence_state = mStrategyWrapper->IsConverged() ? TimeStepEndState::ConvergenceState::converged :
+                                                                     TimeStepEndState::ConvergenceState::non_converged;
+        result.number_of_nonlinear_iterations = mStrategyWrapper->GetNumberOfIterations();
+        return result;
+    }
+
+private:
+    std::shared_ptr<StrategyWrapper> mStrategyWrapper;
+    ProcessRefVec                 mProcessRefs;
```
rfaasse: Very minor, but the alignment here is not correct anymore

applications/GeoMechanicsApplication/custom_workflows/time_stepping.hpp
```diff
@@ -0,0 +1,77 @@
...
+        for (const auto& process : mProcessRefs)
+        {
+            process.get().ExecuteFinalizeSolutionStep();
+        }
+
+        mStrategyWrapper->FinalizeSolutionStep();
+
+        TimeStepEndState result;
+        result.time = Time;
+        result.convergence_state = mStrategyWrapper->IsConverged() ? TimeStepEndState::ConvergenceState::converged :
```
rfaasse: Just a suggestion, but maybe we can let the wrapper return the convergence state, instead of a bool for IsConverged and then creating the state afterwards like we do now?

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
```diff
@@ -0,0 +1,218 @@
...
+                                            return true;}
+    void FinalizeSolutionStep()   override {mSolverStrategyFinalizeSolutionStepCalls += 1;}
+
+private:
+    unsigned int mSolverStrategyInitializeCalls             = 0;
+    unsigned int mSolverStrategyInitializeSolutionStepCalls = 0;
+    unsigned int mSolverStrategyPredictCalls                = 0;
+    unsigned int mSolverStrategySolveSolutionsStepCalls     = 0;
+    unsigned int mSolverStrategyFinalizeSolutionStepCalls   = 0;
+};
```
rfaasse: I don't think the two solver strategies and the DummySparse and DummyDenseType are used anymore, could you remove them?

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+    unsigned int NumberOfExecuteInitializeSolutionStepCalls() const {return mSolutionStepInitializedCalls;}
+    unsigned int NumberOfExecuteFinalizeSolutionStepCalls()   const {return mSolutionStepFinalizedCalls;}
+
+private:
+    unsigned int mSolutionStepInitializedCalls = 0;
+    unsigned int mSolutionStepFinalizedCalls   = 0;
+};
+
+class DummyAlwaysConvergingStrategyWrapper : public StrategyWrapper
```
rfaasse: Since the `DummyAlwaysConvergingStrategyWrapper` and `DummyNeverConvergingStrategyWrapper` are almost the same, my suggestion would be to just create one DummyStrategyWrapper, where in the constructor we can specify e.g. a boolean to say if it should converge or not. That would save some lines of code.

applications/GeoMechanicsApplication/custom_workflows/time_stepping.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+    void SetSolverStrategy(std::shared_ptr<StrategyWrapper> SolverStrategy)
+    {
+        mStrategyWrapper = std::move(SolverStrategy);
+    }
+
+    void SetProcessReferences(ProcessRefVec ProcessRefs)
+    {
+        mProcessRefs = std::move(ProcessRefs);
+    }
```
rfaasse: In the end we changed a bit how these processes are passed. It will now be passed as a vector of std::weak_ptr. @WPK4FEM @avdg81 Do you think we should fix that as part of this PR, or in the timeloopexecutor?

rfaasse (changes requested):
Looks very clear to me! I have some suggestions for improvements, but in general it looks good to me!

applications/GeoMechanicsApplication/custom_workflows/time_stepping.hpp
WPK4FEM: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/custom_workflows/time_stepping.hpp
WPK4FEM: done

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/custom_workflows/time_stepping.hpp
WPK4FEM: We will tie things together when implementing the time loop executor.

applications/GeoMechanicsApplication/custom_workflows/time_stepping.hpp
WPK4FEM: Done

rfaasse (dismissed):

applications/GeoMechanicsApplication/custom_workflows/strategy_wrapper.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include "time_step_end_state.hpp"
+
+namespace Kratos
+{
+    class StrategyWrapper
+    {
+    public:
+        virtual ~StrategyWrapper() = default;
+
+        [[nodiscard]] virtual TimeStepEndState::ConvergenceState GetConvergenceState(const TimeStepEndState& rEndState)       = 0;
```
avdg81: I'm not sure why we need the `const TimeStepEndState&` parameter here. Perhaps this is a leftover from an earlier iteration? I mean, shouldn't the strategy wrapper itself know whether the strategy converged or not? Also, class `DummyStrategyWrapper` doesn't seem to use this parameter. If you agree, we may remove the parameter from the function's signature.

applications/GeoMechanicsApplication/custom_workflows/time_step_executor.cpp
```diff
@@ -0,0 +1,56 @@
...
+    for (const auto& process : mProcessRefs)
+    {
+         process.get().ExecuteFinalizeSolutionStep();
+    }
+
+    mStrategyWrapper->FinalizeSolutionStep();
+
+    TimeStepEndState result;
+    result.time              = Time;
+    result.convergence_state = mStrategyWrapper->GetConvergenceState(result);
```
avdg81: When you agree with my previous suggestion, you no longer need to pass `result` to this member function (which makes sense, since `result` is actually being build up here).

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+
+namespace
+{
+
+class ProcessSpy : public Process
+{
+public:
+    void ExecuteInitializeSolutionStep() override {mSolutionStepInitializedCalls += 1;}
+    void ExecuteFinalizeSolutionStep()   override {mSolutionStepFinalizedCalls   += 1;}
```
avdg81: Although correct, I think it's more common to use the pre-increment operator (`++`) if you want to increase a variable's value by one:
```suggestion
    void ExecuteInitializeSolutionStep() override {++mSolutionStepInitializedCalls;}
    void ExecuteFinalizeSolutionStep()   override {++mSolutionStepFinalizedCalls;}
```
(What you wrote here is common in Python though, since Python doesn't have increment operators.)

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
```diff
@@ -0,0 +1,154 @@
...
+private:
+    unsigned int mSolutionStepInitializedCalls = 0;
+    unsigned int mSolutionStepFinalizedCalls   = 0;
+};
+
+class DummyStrategyWrapper : public StrategyWrapper
+{
+public:
+    explicit DummyStrategyWrapper::DummyStrategyWrapper(bool convergence_needed) :
+        mConvergenceNeeded ( convergence_needed ? TimeStepEndState::ConvergenceState::converged : TimeStepEndState::ConvergenceState::non_converged) {}
```
avdg81: Instead of using a `bool`ean  flag, we could also have passed the desired convergence state directly (i.e. `TimeStepEndState::ConvergenceState::converged` or `TimeStepEndState::ConvergenceState::non_converged`). This may result in slightly more readable constructor calls: `DummyStrategyWrapper{true}` vs `DummyStrategyWrapper{TimeStepEndState::ConvergenceState::converged}` (where we may want to define an alias to reduce the latter to `DummyStrategyWrapper{converged}`. What do you think?

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
```diff
@@ -0,0 +1,154 @@
...
+    [[nodiscard]] TimeStepEndState::ConvergenceState GetConvergenceState(const TimeStepEndState& rEndState) override { return mConvergenceNeeded;};
+    [[nodiscard]] std::size_t                        GetNumberOfIterations()                                const override { return 4;};
+    [[nodiscard]] double                             GetEndTime()                                           const override { return 10.;};
+    void Initialize()             override {mSolverStrategyInitializeCalls += 1;}
+    void InitializeSolutionStep() override {mSolverStrategyInitializeSolutionStepCalls += 1;}
+    void Predict()                override {mSolverStrategyPredictCalls += 1;}
+    bool SolveSolutionStep()      override {mSolverStrategySolveSolutionsStepCalls += 1;
+                                            return true;}
+    void FinalizeSolutionStep()   override {mSolverStrategyFinalizeSolutionStepCalls += 1;}
```
avdg81: Also here I would prefer to use the pre-increment operator.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
```diff
@@ -0,0 +1,154 @@
...
+namespace Kratos::Testing
+{
+
+KRATOS_TEST_CASE_IN_SUITE(RunReturnsNonConvergedWhenStrategyDoesNotConverge, KratosGeoMechanicsFastSuite)
+{
+    TimeStepExecutor executor;
+    auto nonconverging_strategy = std::make_shared<DummyStrategyWrapper>(false);
+    executor.SetSolverStrategy(nonconverging_strategy);
+    const auto time = 0.0;
+    KRATOS_EXPECT_EQ(TimeStepEndState::ConvergenceState::non_converged, executor.Run(time).convergence_state);
```
avdg81: I think we can now simply this to `KRATOS_EXPECT_TRUE(executor.Run(time).NonConverged())` since the `TimeStepEndState` struct now is equipped with `Converged()` and `NonConverged()` queries.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
```diff
@@ -0,0 +1,154 @@
...
+}
+
+KRATOS_TEST_CASE_IN_SUITE(RunReturnsConvergedWhenStrategyConverged, KratosGeoMechanicsFastSuite)
+{
+    TimeStepExecutor executor;
+    auto converging_strategy = std::make_shared<DummyStrategyWrapper>(true);
+    executor.SetSolverStrategy(converging_strategy);
+    const auto time = 0.0;
+    KRATOS_EXPECT_EQ(TimeStepEndState::ConvergenceState::converged, executor.Run(time).convergence_state);
```
avdg81: ... and in the same way we could simplify this to `KRATOS_EXPECT_TRUE(executor.Run(time).Converged())`.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
```diff
@@ -0,0 +1,154 @@
...
+    executor.SetSolverStrategy(converging_strategy);
+    ProcessSpy spy;
+    TimeStepExecutor::ProcessRefVec process_refs{spy};
+    executor.SetProcessReferences(process_refs);
+    const auto time = 0.0;
+
+    executor.Run(time);
+
+    KRATOS_EXPECT_EQ(1,spy.NumberOfExecuteInitializeSolutionStepCalls());
+    KRATOS_EXPECT_EQ(1,spy.NumberOfExecuteFinalizeSolutionStepCalls());
```
avdg81: Nit-picking from my side:
```suggestion
    KRATOS_EXPECT_EQ(1, spy.NumberOfExecuteInitializeSolutionStepCalls());
    KRATOS_EXPECT_EQ(1, spy.NumberOfExecuteFinalizeSolutionStepCalls());
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
```diff
@@ -0,0 +1,154 @@
...
+}
+
+KRATOS_TEST_CASE_IN_SUITE(SolverStrategyMemberFunctionsAllCalledOnce, KratosGeoMechanicsFastSuite)
+{
+    TimeStepExecutor executor;
+    auto converging_strategy = std::make_shared<DummyStrategyWrapper>(true);
+    executor.SetSolverStrategy(converging_strategy);
+    ProcessSpy spy;
+    TimeStepExecutor::ProcessRefVec process_refs{spy};
+    executor.SetProcessReferences(process_refs);
```
avdg81: I think that this test doesn't need the spy process to fulfill its purpose. If that is correct, I would leave it out.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
```diff
@@ -0,0 +1,154 @@
...
+    const auto time = 0.0;
+
+    executor.Run(time);
+
+    KRATOS_EXPECT_EQ(1,converging_strategy->NumberOfSolverStrategyInitializeCalls());
+    KRATOS_EXPECT_EQ(1,converging_strategy->NumberOfSolverStrategyInitializeSolutionStepCalls());
+    KRATOS_EXPECT_EQ(1,converging_strategy->NumberOfSolverStrategyPredictCalls());
+    KRATOS_EXPECT_EQ(1,converging_strategy->NumberOfSolverStrategySolveSolutionStepCalls());
+    KRATOS_EXPECT_EQ(1,converging_strategy->NumberOfSolverStrategyFinalizeSolutionStepCalls());
```
avdg81: More nit-picking:
```suggestion
    KRATOS_EXPECT_EQ(1, converging_strategy->NumberOfSolverStrategyInitializeCalls());
    KRATOS_EXPECT_EQ(1, converging_strategy->NumberOfSolverStrategyInitializeSolutionStepCalls());
    KRATOS_EXPECT_EQ(1, converging_strategy->NumberOfSolverStrategyPredictCalls());
    KRATOS_EXPECT_EQ(1, converging_strategy->NumberOfSolverStrategySolveSolutionStepCalls());
    KRATOS_EXPECT_EQ(1, converging_strategy->NumberOfSolverStrategyFinalizeSolutionStepCalls());
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
```diff
@@ -0,0 +1,154 @@
...
+}
+
+KRATOS_TEST_CASE_IN_SUITE(ConvergingTimeStepExecutionReturnsGivenTime, KratosGeoMechanicsFastSuite)
+{
+    TimeStepExecutor executor;
+    auto converging_strategy = std::make_shared<DummyStrategyWrapper>(true);
+    executor.SetSolverStrategy(converging_strategy);
+    const auto time = 2.0;
+    executor.Run(time);
```
avdg81: I don't see why we need this call. Just calling `Run` once (at the line below) should be sufficient, right?

avdg81 (commented):
This looks really nice Wijtze Pieter! I have several minor remarks which may help to make the code even more elegant. See for yourself whether my comments make sense. Thanks!

applications/GeoMechanicsApplication/custom_workflows/strategy_wrapper.hpp
WPK4FEM: I agree, done.

applications/GeoMechanicsApplication/custom_workflows/time_step_executor.cpp
WPK4FEM: done

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
WPK4FEM: Learned something, done.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
WPK4FEM: done

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
WPK4FEM: done with a smile

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
WPK4FEM: Tried and seems to work without. Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
WPK4FEM: Done :)

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
WPK4FEM: indeed, done

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
WPK4FEM: Done ( I hope as intended ... )

avdg81 (approved):
I think this one is ready to go.

PR #11650, opened 2023-10-03
[MPM] Cleanup set and fill buffer size

ncrescenzio:
Buffer size is now assigned to model parts using the function `SetAndFillBuffer` from `auxiliary_solver_utilities` (similarly to other applications).

This also solves a problem related to the vtk output of the background grid: variables `TIME` and `STEP` where not found in `ModelPart.ProcessInfo` when calling `VtkOutput.Check()`.

### DIFF
--- a/applications/ParticleMechanicsApplication/python_scripts/mpm_solver.py
+++ b/applications/ParticleMechanicsApplication/python_scripts/mpm_solver.py
@@ -24,9 +24,6 @@ class MPMSolver(PythonSolver):
         # Add model part containers
         self._AddModelPartContainers()
 
-        # Default settings
-        self.min_buffer_size = 2
-
         KratosMultiphysics.Logger.PrintInfo("::[MPMSolver]:: ", "Solver is constructed correctly.")
 
     @classmethod
@@ -96,6 +93,9 @@ class MPMSolver(PythonSolver):
 
     ### Solver public functions
 
+    def GetMinimumBufferSize(self):
+        return 2
+
     def AddVariables(self):
         # Add variables to background grid model part
         self._AddVariablesToModelPart(self.grid_model_part)
@@ -110,7 +110,7 @@ class MPMSolver(PythonSolver):
 
     def PrepareModelPart(self):
         # Set buffer size
-        self._SetBufferSize()
+        self._SetAndFillBuffer()
 
         # Executes the check and prepare model process
         self.__ExecuteCheckAndPrepare()
@@ -229,7 +229,6 @@ class MPMSolver(PythonSolver):
         self.material_point_model_part.SetNodes(self.grid_model_part.GetNodes())
 
         if not self.is_restarted():
-            self.material_point_model_part.SetBufferSize(self.grid_model_part.GetBufferSize())
             self.material_point_model_part.ProcessInfo = self.grid_model_part.ProcessInfo
 
             # Generate MP Element and Condition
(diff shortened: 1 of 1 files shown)


applications/ParticleMechanicsApplication/python_scripts/mpm_solver.py
```diff
@@ -441,25 +440,15 @@ def _CreateLinearStrategy(self):
...
+            auxiliary_solver_utilities.SetAndFillBuffer(self.material_point_model_part, required_buffer_size, delta_time)
+            auxiliary_solver_utilities.SetAndFillBuffer(self.grid_model_part, required_buffer_size, delta_time)
         else:
-            current_buffer_size = self.material_point_model_part.GetBufferSize()
-            if self.min_buffer_size > current_buffer_size:
-                self.material_point_model_part.SetBufferSize(self.min_buffer_size)
-            else:
-                self.material_point_model_part.SetBufferSize(current_buffer_size)
+            required_buffer_size = self.material_point_model_part.GetBufferSize()
+            auxiliary_solver_utilities.SetAndFillBuffer(self.grid_model_part, required_buffer_size, delta_time)
```
rubenzorrilla, line 451: Maybe I'm saying something stupid, but shouldn't the database of `self.grid_model_part` be already filled if you're restarting the `Model` container?

ncrescenzio: Only `self.material_point_model_part` is imported using the `RestartUtility`, while `self.grid_model_part` is imported as usual using the `mdpa` file. So it is necessary to set again the buffer size of the grid model part, no?

rubenzorrilla: But, as far as I understand, in a standard case (no restart) the `self.grid_model_part` is imported once at the beginning of the simulation from its corresponding mdpa. Assuming this, the `self.grid_model_part` will be part of the `Model` container meaning that it will be serialized, together with all its data when doing the save of the restart. Hence, when loading from the restart data file the `self.grid_model_part` will be there with its corresponding data.  With your current implementation you are wiping the data of `self.grid_model_part`.

In any case, I'm not aware of the particularities of your application so this is maybe the behavior you intend to achieve.

ncrescenzio: I'm not the one who implemented the restart in the `ParticleMechanicsApplication`, so I'm not really familiar with it. But I think that in the current implementation only the `self.material_point_model_part` is serialized with all its data when doing the save of the restart through the `SaveRestartProcess`. The `self.grid_model_part` and `self.initial_mesh_model_part` (both part of the `Model` container) are not serialized and data is not saved. And I think it makes sense because the `self.initial_mesh_model_part` is used only at the very beginning of the simulation (not needed and not loaded in the restart) while the `self.grid_model_part` is only used for the computations (and thus it is loaded using the `mdpa` file as in the standard case).

ncrescenzio: @rubenzorrilla was my last comment clear? Can we merge this?

I am not changing the behavior of the application, I only need to define the variables `TIME` and `STEP` otherwise the `VtkOutput` process is not working and I am doing it by using `auxiliary_solver_utilities.SetAndFillBuffer` like it is done in other applications.

If something is not clear about how the restart is implemented in our application I think we can discuss it separately

rubenzorrilla: Please go ahead. I was just pointing that the good way to go would be to serialize the `Model`. If you already have a working implementation I've no objections about.

lauramoremar (approved):

PR #11653, opened 2023-10-03
[Core][Model] remove flatmap

philbucher:
this was deprecated via an error (i.e. rendered unusable) in Nov 2019. I think it is time to go :)


loumalouomega (approved):
Bye. Do we have to remove the message in CMake?

philbucher:
> Bye. Do we have to remove the message in CMake?

there is none, and using the flat map already issued an error since 4 years

PR #11656, opened 2023-10-04
[Core][MPI] Minor fix in STL IO in MPI write

loumalouomega:
**📝 Description**

In this PR , changes have been made to the `stl_io.cpp`. The PR introduces fixes in MPI communication. Specifically, it improves the handling of the number of entities and geometries in the code.

Here's a breakdown of the changes made in the PR:

1. A new variable "number_of_entities" has been introduced to store the count of entities.
2. The count of entities is calculated and summed across all processes using "rDataCommunicator.SumAll(number_of_entities)".
3. Similar changes have been made for the count of geometries, with the introduction of the variable "number_of_geometries."
4. The condition for checking if there are entities or geometries has been updated to use these new count variables, improving efficiency in MPI communication.

Until now, due to the previous check, if a partition was empty the communication could get stuck.

**🆕 Changelog**

- [Minor fix in STL IO in MPI write](https://github.com/KratosMultiphysics/Kratos/commit/47f0d0147e2d5055a31a22b22b6c902234a94b18)

### DIFF
--- a/kratos/input_output/stl_io.cpp
+++ b/kratos/input_output/stl_io.cpp
@@ -135,6 +135,12 @@ void StlIO::WriteModelPart(const ModelPart& rThisModelPart)
     // Get data communicator
     const auto& r_data_communicator = rThisModelPart.GetCommunicator().GetDataCommunicator();
 
+    // If rank is not zeo we remove the stream to avoid conflict
+    if (r_data_communicator.Rank() != 0) {
+        mpInputStream = nullptr;
+    }
+    r_data_communicator.Barrier();
+
     /* Write the solid block */
     // Write header of the file
     if (r_data_communicator.Rank() == 0) {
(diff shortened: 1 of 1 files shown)


philbucher (dismissed):
classic :smile:

matekelemen:
Good catch! We're in desperate need of a better way of finding these bugs other than stumbling into them by chance at runtime.

Does anyone have experience with other MPI software, and how they avoid these problems?

loumalouomega:
> Good catch! We're in desperate need of a better way of finding these bugs other than stumbling into them by chance at runtime.
> 
> Does anyone have experience with other MPI software, and how they avoid these problems?

@matekelemen reapproval?

philbucher:
> Good catch! We're in desperate need of a better way of finding these bugs other than stumbling into them by chance at runtime.
> 
> 
> 
> Does anyone have experience with other MPI software, and how they avoid these problems?

One simple way would be to create a test modelpart that has an empty partition, and use that in the tests. It is always the same problem.

philbucher (approved):

PR #11658, opened 2023-10-04
[GeoMechanicsApplication] Add support for non-uniform line loads

### LINKED ISSUE #11532: [GeoMechanicsApplication] Add support for non-uniform line loads
As a user, I would like to be able to apply a non-uniform load in Kratos, such that I can model an embankment with it's actual shape.

**Description** 

To model an embankment in simulations with D-Settlement, we will use a non-uniform line load that represents the embankment's dead weight imposed on the subsoil. So far, we can only handle uniform line loads. However, due the embankment's geometry (with slopes) we actually need non-uniform line loads. The idea is to define a line load with a table, where the table describes the relationship between a coordinate on the subsoil (almost always X) and the magnitude of the line load. Since this is a different kind of dependency than what we have done so far (time dependencies only), we first need a way to retrieve the variables involved. This is part of another issue (see [PR #11528](https://github.com/KratosMultiphysics/Kratos/pull/11528)), which needs to be finished first.

[We have intentionally separated the changes needed in the Kratos core from the changes in the GeoMechanicsApplication.]

The following branch contains a first working implementation for this issue (no PR opened yet): https://github.com/KratosMultiphysics/Kratos/tree/geo/non-uniform-line-load-test

Identified implementation steps:
1. Change the reading of the table in GeoMechanics, such that it initializes the modelpart correctly from the mdpa file
2. Perform a simple test by directly integrating a simple non-uniform line load and comparing the results.

**Acceptance Criteria**

**Given** the user has modelled an embankment with its actual shape in Kratos,
**When** the user defines a 1D non-uniform line load for the simulation,
**Then** the application should correctly interpret and use this table for non-uniform load distribution.

**Given** the user has created a 1D non-uniform load,
**When** the user specifies the relationship between coordinates and line load magnitudestable as input in a table,
**Then** the load table generated in the mdpa file, should be containing the relationship between a coordinate (almost always X) and the magnitude of the line load.

avdg81:
**📝 Description**
This PR adds support for applying non-uniform line loads. It does so by attaching a table to a uniform line load, where the table defines the relationship between a set of X coordinates and the corresponding "pressure" values. The pressures applied to nodes will be linearly interpolated (and/or linearly extrapolated, if needed) using this table.


applications/GeoMechanicsApplication/custom_processes/apply_component_table_process.hpp
```diff
@@ -79,13 +81,22 @@ class ApplyComponentTableProcess : public Process
...
+            block_for_each(mrModelPart.Nodes(), [&var, this](Node& rNode) {
+                if (mIsFixed) rNode.Fix(var);
+                else if (mIsFixedProvided) rNode.Free(var);
+                rNode.FastGetSolutionStepValue(var) = mInitialValue;
+            });
+        }
+        else if (variable_name_1 == "X") {
+            block_for_each(mrModelPart.Nodes(), [&var, this](auto& node){
+                node.FastGetSolutionStepValue(var) = mpTable->GetValue(node.X());
+            });
```
rfaasse, line 98: Are we sure we don't need to support also "Y" (or "Z", whatever the other direction is in the horizontal plane)? I know it's only one dimension at a time, but maybe Embankment does not always use the same axis?

applications/GeoMechanicsApplication/tests/test_line_loads.py
```diff
@@ -1,8 +1,11 @@
 import os
 
+import KratosMultiphysics as Kratos
 import KratosMultiphysics.KratosUnittest as KratosUnittest
+import KratosMultiphysics.StructuralMechanicsApplication as KratosStructural
 import test_helper
 
+
```
rfaasse: Nitpicking, but this blank line can be removed

applications/GeoMechanicsApplication/tests/test_line_loads.py
```diff
@@ -69,5 +72,32 @@ def test_line_load_3D3N_tet(self):
...
+        parent_name = 'line_load_tests'
+        file_path = test_helper.get_file_path(os.path.join(parent_name, test_name))
+        simulation = test_helper.run_kratos(file_path)
+
+        bottom_node_ids = [1, 2, 6, 11, 17, 25, 34, 46, 59, 75, 90]
+        reactions_y = [reaction[1] for reaction in test_helper.get_nodal_variable(simulation, Kratos.REACTION, bottom_node_ids)]
+
+        self.assertAlmostEqual(20.0, sum(reactions_y), 4)
+
+        line_load_y_by_node_id = [( 91,   0.0),
```
rfaasse, line 85: When looking at the mdpa file, it is quite clear that the load is set using the table and interpolated for nodes in between, but it might be nice to make that explicit and add a comment here (or a README) of why these results are the way they are .

applications/GeoMechanicsApplication/tests/test_line_loads.py
```diff
@@ -69,5 +72,32 @@ def test_line_load_3D3N_tet(self):
...
+                                  ( 93,   0.0),
+                                  ( 98,   0.0),
+                                  (102,  -5.0),
+                                  (106, -10.0),
+                                  (111, -10.0),
+                                  (114, -10.0),
+                                  (118,  -5.0),
+                                  (122,   0.0),
```
rfaasse: It might be nice to shift one of the line-loads in the input a bit, so we don't get exactly the middle between -10 and 0 here, so we know the interpolation is also correct in slightly less trivial cases

applications/GeoMechanicsApplication/tests/test_line_loads.py
```diff
@@ -69,5 +72,32 @@ def test_line_load_3D3N_tet(self):
...
+    def test_nonuniform_line_load(self):
+        test_name = 'non-uniform_line_load'
+        parent_name = 'line_load_tests'
+        file_path = test_helper.get_file_path(os.path.join(parent_name, test_name))
+        simulation = test_helper.run_kratos(file_path)
+
+        bottom_node_ids = [1, 2, 6, 11, 17, 25, 34, 46, 59, 75, 90]
+        reactions_y = [reaction[1] for reaction in test_helper.get_nodal_variable(simulation, Kratos.REACTION, bottom_node_ids)]
+
+        self.assertAlmostEqual(20.0, sum(reactions_y), 4)
```
rfaasse, line 83: This is the check where the integral at the bottom should be the same as the load at the top right? It might be clearer to add a comment with the simple calculation to show that it should indeed be 20 here.

rfaasse (dismissed):
Looks like a clean local change with a clear test! 

I have some suggestions/questions, mainly to clarify the test a bit more, non of which are really blocking.

applications/GeoMechanicsApplication/custom_processes/apply_component_table_process.hpp
avdg81: I have only added what we need now. As far as I can see, for the Embank team only "X" will suffice (since their models assume a plane strain situation, with the gravity direction pointing downwards). It is rather straightforward to add the other ones (i.e. "Y" and "Z") on request. What we could do now is to print a message when the function has no effect instead of silently returning.

applications/GeoMechanicsApplication/tests/test_line_loads.py
avdg81: I'm not entirely sure. I believe that when you would run `black` to enforce the prescribed style as described by PEP-8, it would ensure to have two blank lines between the `import`s and the actual code. But I have removed the blank line as you suggested, since I'm not entirely sure about which style to follow.

applications/GeoMechanicsApplication/tests/test_line_loads.py
avdg81: I have added a `README` file that gives a brief test description.

applications/GeoMechanicsApplication/tests/test_line_loads.py
avdg81: Good suggestion! I have given the applied line load a small offset in order to arrive at an asymmetric model.

applications/GeoMechanicsApplication/tests/test_line_loads.py
avdg81: I have added that to the new `README` file.

applications/GeoMechanicsApplication/tests/line_load_tests/non-uniform_line_load/README.md
```diff
@@ -0,0 +1,34 @@
...
+```
+
+For all points on the top edge where x < 0.6 m, there is no vertical line load
+applied.  This also holds true for all points on that edge where x > 3.6 m.
+For x in the range [0.6, 1.6> m, the line load increases linearly from 0.0 to
+10 N/m.  From x = 1.6 m to x = 2.6 m, the vertical line load remains constant
+at 10 N/m.  And for x in the range [2.6, 3.6> m the line load decreases linearly
+from 10 N/m to 0.0 N/m.
+
+The total applied load equals (2.6 - 0.6) m * 10 N/m = 20 N.  The sum of the
```
rfaasse: I don't know if it's necessary to elaborate this further (and it's a bit nitpicky), but if I understand correctly, this is actually  (1.6 - 0.6) * 5 + (2.6-1.6) * 10 + (3.6 - 2.6) * 5 right? I leave it up to you if the current one is clear enough.

applications/GeoMechanicsApplication/tests/test_line_loads.py
rfaasse: Thanks, the `README` is very clear!

applications/GeoMechanicsApplication/custom_processes/apply_component_table_process.hpp
rfaasse: Okay, clear. I leave it up to you if you want to add the message/error if something else than TIME or X is used.

rfaasse (dismissed):

applications/GeoMechanicsApplication/tests/test_line_loads.py
rfaasse: Very clear, thanks for adding it!

applications/GeoMechanicsApplication/tests/line_load_tests/non-uniform_line_load/README.md
avdg81: Done.

applications/GeoMechanicsApplication/custom_processes/apply_component_table_process.hpp
avdg81: Done.

rfaasse (approved):
Looks good to me!

PR #11659, opened 2023-10-04
[ConvDiff] ConvectionDiffusionApplication testing utilities

rubenzorrilla:
**📝 Description**
This PR creates a new `ConvectionDiffusionTestingUtilities` static class to be used in the cpp tests, allowing to remove the auxiliary copied&pasted function that prepares the model part in all tests.

PR #11660, opened 2023-10-04
[ConvDiff] Axisymmetric convection-diffusion element and thermal face

rubenzorrilla:
**📝 Description**
This adds an axisymmetric convection-diffusion element together with its corresponding thermal face condition. Element unittest have been added accordingly.

#11659 needs to be merged first.


RiccardoRossi (approved):
this looks ok to me

PR #11662, opened 2023-10-05
[GeoMechanicsApplication] Add a time incrementor that models a collection of prescribed time increments

### LINKED ISSUE #11622: [GeoMechanicsApplication] Add a time incrementor that models a prescribed set of time increments
As a developer, I want to be able to input a set of prescribed time increments to be used by the time loop executor.

#### Background

For testing, it is convenient to have the simplest way of prescribing time increments available. We feel that the simplest way is to input a list of fixed time increments (so no adaptive scheme whatsoever).

#### Acceptance Criteria

**Given** a time incrementor and an empty list of time increments
**When** the developer attempts to construct a time incrementor
**Then** it must throw an exception.

**Given** a time incrementor and a list that contains at least one negative time increment
**When** the developer attempts to construct a time incrementor
**Then** it must throw an exception.

**Given** a time incrementor that contains a set of N time increments and a fixed step state that indicates a converged state
**When** the developer processes the converged step state and subsequently queries whether a next step is needed
**Then** the first N times it must return true. The N+1 and N+2 times it must return false.

**Given** a time incrementor that contains a set of N time increments and a fixed step state that indicates a converged state
**When** the developer processes the converged step state and subsequently queries whether a step should be retried
**Then** the response must always be false (note that we always carry out the first cycle).

**Given** a time incrementor that contains a set of N time increments and a fixed step state that indicates a converged state
**When** the developer processes the converged step state and subsequently asks for the increment sizes (N times in a row)
**Then** the returned increments must match the prescribed increments.

**Given** a time incrementor that contains a set of N time increments and a fixed step state that indicates a converged state
**When** the developer processes the converged step state and subsequently asks for the increment sizes (N+1 times in a row)
**Then** at the N+1 query an exception must be thrown.

**Given** a time incrementor that contains a set of N time increments and a cycle number of zero
**When** the developer queries whether a next cycle is needed
**Then** the query must return true.

**Given** a time incrementor that contains a set of N time increments and a positive cycle number
**When** the developer queries whether a next cycle is needed
**Then** the query must return false.

**Given** a time incrementor that contains a set of N time increments and a fixed step state that indicates a non-converged state
**When** the developer processes the converged step state and subsequently asks whether a next step is needed
**Then** the query must return false.

avdg81:
**📝 Description**
Added a time incrementor that implements the time incrementor interface for the case where we have a collection of prescribed time increments. This class is probably one of the simplest (and most fundamental) ways to step through time. We will use it when developing the time loop executor. (But we may consider to also make it accessible through the `ProjectParameters.json` file.)

### DIFF
--- a/applications/GeoMechanicsApplication/custom_workflows/prescribed_time_incrementor.cpp
+++ b/applications/GeoMechanicsApplication/custom_workflows/prescribed_time_incrementor.cpp
@@ -18,12 +18,13 @@ namespace Kratos
 {
 
 PrescribedTimeIncrementor::PrescribedTimeIncrementor(const std::vector<double>& increments)
-    : mIncrements(increments)
+    : mIncrements(increments),
+      mPos(mIncrements.begin())
 {}
 
 bool PrescribedTimeIncrementor::WantNextStep(const Kratos::TimeStepEndState& rPreviousState) const
 {
-    return !mIncrements.empty();
+    return mPos != mIncrements.end();
 }
 
 bool PrescribedTimeIncrementor::WantRetryStep(std::size_t             CycleNumber,
@@ -39,6 +40,7 @@ double PrescribedTimeIncrementor::GetIncrement() const
 
 void PrescribedTimeIncrementor::PostTimeStepExecution(const Kratos::TimeStepEndState& rResultantState)
 {
+    ++mPos;
 }
 
 }
--- a/applications/GeoMechanicsApplication/custom_workflows/prescribed_time_incrementor.cpp
+++ b/applications/GeoMechanicsApplication/custom_workflows/prescribed_time_incrementor.cpp
@@ -1,0 +1,43 @@
+// KRATOS___
+//     //   ) )
+//    //         ___      ___
+//   //  ____  //___) ) //   ) )
+//  //    / / //       //   / /
+// ((____/ / ((____   ((___/ /  MECHANICS
+//
+//  License:         geo_mechanics_application/license.txt
+//
+//  Main authors:    Wijtze Pieter Kikstra
+//                   Anne van de Graaf
+//
+
+#include "prescribed_time_incrementor.h"
+
+
+namespace Kratos
+{
+
+PrescribedTimeIncrementor::PrescribedTimeIncrementor(const std::vector<double>& increments)
+{}
+
+bool PrescribedTimeIncrementor::WantNextStep(const Kratos::TimeStepEndState& rPreviousState) const
+{
+    return false;
+}
+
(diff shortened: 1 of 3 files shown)


applications/GeoMechanicsApplication/tests/cpp_tests/test_prescribed_time_incrementor.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+}
+
+
+namespace Kratos::Testing
+{
+
+KRATOS_TEST_CASE_IN_SUITE(NoNextTimeStepWhenPrescribedTimeIncrementorHasEmptyList, KratosGeoMechanicsFastSuite)
+{
+    std::vector<double> increments;
```
rfaasse: In all tests, the increments vector can be const std::vector<double> :)

applications/GeoMechanicsApplication/custom_workflows/prescribed_time_incrementor.cpp
```diff
@@ -0,0 +1,58 @@
...
+}
+
+double PrescribedTimeIncrementor::GetIncrement() const
+{
+    KRATOS_ERROR_IF(mPos == mIncrements.end()) << "Out of increment range";
+
+    return *mPos;
+}
+
+void PrescribedTimeIncrementor::PostTimeStepExecution(const Kratos::TimeStepEndState& rResultantState)
```
rfaasse, line 53: Just for my understanding: This is split from the GetIncrement() function in the interface, such that the adaptive time incrementor can re-try a time-step right?

rfaasse (dismissed):
Looks very clear and complete! Nice to see TDD happening more and more 😄 

Just flagged a a couple of missing 'consts' but nothing major.

applications/GeoMechanicsApplication/tests/cpp_tests/test_prescribed_time_incrementor.cpp
avdg81: Good catch. In fact, there were a few more variables that could also be made `const`. I have changed those as well.

applications/GeoMechanicsApplication/custom_workflows/prescribed_time_incrementor.cpp
avdg81: Yes, this API is there to give the incrementor the opportunity to adjust the next increment if needed. `GetIncrement()` is supposed to do nothing more than fetching the increment to be applied.

rfaasse (approved):
Looks good to me!

PR #11665, opened 2023-10-05
[GeoMechanicsApplication] Geo/11626 initialize solving strategies settlement

### LINKED ISSUE #11626: [GeoMechanicsApplication] Initialize the solving strategy for the Settlement Workflow based on project parameters
As a user, I would like to be able to specify a solving strategy in the Settlement workflow, such that it is used to solve the equations during settlement computations.

#### Background
The following solving strategy is supported in the settlement workflow:
1. GeoMechanicsNewtonRaphsonStrategy

The following convergence criteria are supported in the settlement workflow:
1. DisplacementCriteria

The following linear solvers are supported in the settlement workflow:

1. amgcl
2. amgcl_ns
3. bicgstab
4. cg
5. deflated_cg
6. monotonicity_preserving
7. scaling
8. skyline_lu_factorization
9. sparse_cg
10. sparse_lu
11. sparse_qr
12. tfqmr

The following schemes are supported in the settlement workflow:
1. Backward_Euler with a Quasi-Static solution type.

The following 'BuilderAndSolvers' are supported in the settlement workflow:
1. ResidualBasedBlockBuilderAndSolver
2. ResidualBasedEliminationBuilderAndSolver

#### Acceptance Criteria
**Given** The user specifies a supported (see Background) solving strategy in the ProjectParameters.json
**When** The user runs the settlement workflow
**Then** The specified solving strategy and its parameters are available for use

**Given** The user specifies a non-supported (see Background) solving strategy in the ProjectParameters.json
**When** The user runs the settlement workflow
**Then** The workflow aborts with the error message: "The solving strategy <name> is not supported for the settlement workflow."

**Given** The user does not specify a solving strategy in the ProjectParameters.json
**When** The user runs the settlement workflow
**Then** The workflow aborts with the error message: "Please specify a solving strategy as it is mandatory to run the settlement workflow."

rfaasse:
**📝 Description**
This PR aims to create the factory for the solving strategy and the necessary subcomponents (Scheme, BuilderAndSolver, ConvergenceCriteria and LinearSolver) and test them using unit tests.

**🆕 Changelog**
- Added factories for the SolvingStrategy, Scheme, BuilderAndSolver, ConvergenceCriteria and LinearSolver
- Added unit tests for all the factories


applications/GeoMechanicsApplication/tests/run_cpp_unit_tests.py
```diff
@@ -1,6 +1,8 @@
 import sys
 import KratosMultiphysics
 from KratosMultiphysics.GeoMechanicsApplication import *
+from KratosMultiphysics.LinearSolversApplication import *
```
rfaasse: This import is necessary to initialize all solvers (including e.g. sparse_lu). I get a SonarCloud code smell about this import (https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&types=CODE_SMELL&branch=geo%2F11626_InitializeSolvingStrategies_Settlement&id=KratosGeomechanics&open=AYsBGTSPtEYIt1WF2mqJ), so let's discuss if that's acceptable.

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
rfaasse, line 1: Here we would also need to ImportApplication for the linear solver application when we hook it all up, but since we add an explicit dependency then, I'd like to discuss it (although we have the same dependency already in e.g. our python tests).

applications/GeoMechanicsApplication/tests/cpp_tests/test_builder_and_solver_factory.cpp
```diff
@@ -0,0 +1,69 @@
...
+//  Main authors:    Richard Faasse
+//
+
+#include "testing/testing.h"
+#include "custom_utilities/builder_and_solver_factory.hpp"
+#include "spaces/ublas_space.h"
+#include "linear_solvers/linear_solver.h"
+
+using namespace Kratos;
+using SparseSpaceType = UblasSpace<double, CompressedMatrix, Vector>;
```
rfaasse, line 19: I'll check if it's feasible to create dummy's for these, but it's not trivial

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
```diff
@@ -0,0 +1,96 @@
...
+                    parameters);
+        }
+
+        return nullptr;
+    }
+
+    static Parameters ExtractStrategyParameters(const Parameters &rSolverSettings)
+    {
+        Parameters result;
+        std::vector<std::string> strategy_entries = {"min_iteration",
```
rfaasse, line 65: I think this might be a typo in the actual solving strategy (where I pulled this list from), so it should be min_iterations.

applications/GeoMechanicsApplication/custom_utilities/builder_and_solver_factory.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Kratos
+{
+
+template<class TSparseSpace, class TDenseSpace, class TLinearSolver>
+class BuilderAndSolverFactory
+{
+
+public:
+    static std::shared_ptr<BuilderAndSolver<TSparseSpace, TDenseSpace, TLinearSolver>> Create(const Parameters& rSolverSettings,
```
avdg81: Just like you did for the `ConvergenceCriteriaFactory`, you could define an alias that simplifies the return type of `Create`.
```suggestion
    using BuilderAndSolverType = BuilderAndSolver<TSparseSpace, TDenseSpace, TLinearSolver>;
    static std::shared_ptr<BuilderAndSolverType> Create(const Parameters& rSolverSettings,
```

applications/GeoMechanicsApplication/custom_utilities/scheme_factory.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include <memory>
+
+namespace Kratos
+{
+
+template<class TSparseSpace, class TDenseSpace>
+class SchemeFactory
+{
+public:
+    static std::shared_ptr<Scheme<TSparseSpace, TDenseSpace>> Create(const Parameters& rSolverSettings)
```
avdg81: Similar to my previous comment, I think that adding an alias improves readability here:
```suggestion
    using SchemeType = Scheme<TSparseSpace, TDenseSpace>;
    static std::shared_ptr<SchemeType> Create(const Parameters& rSolverSettings)
```

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
```diff
@@ -0,0 +1,96 @@
...
+            GeoMechanicsNewtonRaphsonStrategy<TSparseSpace, TDenseSpace, TLinearSolver>::Name())
+        {
+            return std::make_unique<GeoMechanicsNewtonRaphsonStrategy<TSparseSpace,
+                    TDenseSpace,
+                    TLinearSolver>>(rModelPart,
+                    scheme,
+                    solver,
+                    criteria,
+                    builder_and_solver,
+                    parameters);
```
avdg81: I would suggest to limit the scope of `parameters` to this if-clause. In fact, a temporary will do:
```suggestion
                    ExtractStrategyParameters(rSolverSettings));
```
(And remove the definition `parameters` above.)

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
```diff
@@ -0,0 +1,96 @@
...
+        {
+            if (rSolverSettings.Has(entry))
+            {
+                result.AddValue(entry, rSolverSettings[entry]);
+            }
+        }
+
+        return result;
+    }
+
```
avdg81: This blank line is redundant.

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
```diff
@@ -0,0 +1,96 @@
...
+#include "builder_and_solver_factory.hpp"
+
+namespace Kratos
+{
+
+template<class TSparseSpace, class TDenseSpace, class TLinearSolver>
+class SolvingStrategyFactory
+{
+public:
+    [[nodiscard]] static std::unique_ptr<SolvingStrategy<TSparseSpace, TDenseSpace>> Create(Parameters& rSolverSettings, ModelPart& rModelPart)
```
avdg81: If I'm not mistaken, I think we can make the first parameter a reference to const:
```suggestion
    [[nodiscard]] static std::unique_ptr<SolvingStrategy<TSparseSpace, TDenseSpace>> Create(const Parameters& rSolverSettings, ModelPart& rModelPart)
```

applications/GeoMechanicsApplication/custom_utilities/scheme_factory.hpp
```diff
@@ -0,0 +1,41 @@
...
+    {
+        KRATOS_ERROR_IF_NOT(rSolverSettings.Has("scheme_type")) << "scheme_type is not defined, aborting";
+        KRATOS_ERROR_IF_NOT(rSolverSettings.Has("solution_type")) << "solution_type is not defined, aborting";
+
+        if (rSolverSettings["scheme_type"].GetString() == "Backward_Euler" &&
+            rSolverSettings["solution_type"].GetString() == "Quasi-Static")
```
avdg81, line 34: The Python counterpart of this code is case-insensitive. Shouldn't we propagate that to the C++ side too?

applications/GeoMechanicsApplication/tests/cpp_tests/test_builder_and_solver_factory.cpp
```diff
@@ -0,0 +1,69 @@
...
+using LocalSpaceType = UblasSpace<double, Matrix, Vector>;
+
+namespace Kratos::Testing
+{
+
+template<class TSparseSpaceType, class TDenseSpaceType>
+class DummyLinearSolver : public LinearSolver<TSparseSpaceType, TDenseSpaceType>
+{};
```
avdg81: I'm not sure why we need a new derived class here. It doesn't seem to add anything. Wouldn't it be better to use a template alias?

applications/GeoMechanicsApplication/tests/cpp_tests/test_builder_and_solver_factory.cpp
```diff
@@ -0,0 +1,69 @@
...
+{
+    const std::string validBlockParameters = R"(
+    {
+        "block_builder": true
+    }
+    )";
+
+    auto solver = std::make_shared<DummyLinearSolver<SparseSpaceType, LocalSpaceType>>();
+
+    auto builder_and_solver = BuilderAndSolverFactory<SparseSpaceType, LocalSpaceType, LinearSolver<SparseSpaceType, LocalSpaceType>>::Create(Parameters{validBlockParameters}, solver);
```
avdg81: I'm just wandering whether the template parameters of `LinearSolver` can be independent of the first two template parameters of `BuilderAndSolverFactory`...? If the answer is yes, then this is the right way to go. If, however, their types must match, then we may consider to remove the third template parameter from `BuilderAndSolverFactory` (since in that case it is fully determined by the first two template parameters).

applications/GeoMechanicsApplication/tests/cpp_tests/test_builder_and_solver_factory.cpp
```diff
@@ -0,0 +1,69 @@
...
+KRATOS_TEST_CASE_IN_SUITE(CreateBuilderAndSolver_Throws_WhenBlockBuilderIsUndefined, KratosGeoMechanicsFastSuite)
+{
+    auto solver = std::make_shared<DummyLinearSolver<SparseSpaceType, LocalSpaceType>>();
+
+    auto factory = BuilderAndSolverFactory<SparseSpaceType, LocalSpaceType, LinearSolver<SparseSpaceType, LocalSpaceType>>();
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(factory.Create(Parameters{}, solver),
+                                      "the block_builder parameter is not defined, aborting BuilderAndSolverCreation")
+}
+
+
```
avdg81: Nitpicking: I guess this blank line is redundant.

applications/GeoMechanicsApplication/tests/cpp_tests/test_convergence_criteria_factory.cpp
```diff
@@ -0,0 +1,59 @@
...
+KRATOS_TEST_CASE_IN_SUITE(Create_Throws_WhenConvergenceCriterionIsUnknown, KratosGeoMechanicsFastSuite)
+{
+    ConvergenceCriteriaFactory<SparseSpaceType, LocalSpaceType> factory;
+
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(factory.Create(Parameters{R"({"convergence_criterion" : "something_unknown" })"}),
+                                      "The convergence_criterion (something_unknown) is unknown, supported criteria are: 'displacement_criterion'")
+}
+
+
+
```
avdg81: More nitpicking: two redundant blank lines.

applications/GeoMechanicsApplication/tests/cpp_tests/test_scheme_factory.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+using namespace Kratos;
+
+namespace Kratos::Testing
+{
+
+using SparseSpaceType = UblasSpace<double, CompressedMatrix, Vector>;
+using LocalSpaceType = UblasSpace<double, Matrix, Vector>;
+
+KRATOS_TEST_CASE_IN_SUITE(CreateScheme_Throws_WhenSolutionTypeIsUndefined, KratosGeoMechanicsFastSuite)
```
avdg81: I think the test case name is incorrect. You probably meant `CreateScheme_Throws_WhenSchemeTypeIsUndefined` (because in this test case `solution_type` _is_ defined).

applications/GeoMechanicsApplication/tests/cpp_tests/test_scheme_factory.cpp
```diff
@@ -0,0 +1,89 @@
...
+    {
+        "solution_type" : "Quasi-Static"
+    }
+    )";
+
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(factory.Create(Parameters{parameters}),
+                                      "scheme_type is not defined, aborting")
+}
+
+KRATOS_TEST_CASE_IN_SUITE(CreateScheme_Throws_WhenSchemeTypeIsUndefined, KratosGeoMechanicsFastSuite)
```
avdg81: And a similar mistake here in the test case name.

applications/GeoMechanicsApplication/tests/cpp_tests/test_convergence_criteria_factory.cpp
```diff
@@ -0,0 +1,59 @@
...
+using SparseSpaceType = UblasSpace<double, CompressedMatrix, Vector>;
+using LocalSpaceType = UblasSpace<double, Matrix, Vector>;
+
+namespace Kratos::Testing
+{
+
+KRATOS_TEST_CASE_IN_SUITE(Create_ReturnsCorrectConvergenceCriteria_ForDisplacement, KratosGeoMechanicsFastSuite)
+{
+    const auto convergence_criteria = ConvergenceCriteriaFactory<SparseSpaceType, LocalSpaceType>::Create(Parameters{validParameters});
+    const auto displacementCriterion = dynamic_cast<DisplacementCriteria<SparseSpaceType, LocalSpaceType>*>(convergence_criteria.get());
```
avdg81: You may consider to add `const` here:
```suggestion
    const auto displacementCriterion = dynamic_cast<const DisplacementCriteria<SparseSpaceType, LocalSpaceType>*>(convergence_criteria.get());
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_solving_strategy_factory.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+KRATOS_TEST_CASE_IN_SUITE(CreateSolvingStrategy_Throws_WhenNoStrategyTypeIsDefined, KratosGeoMechanicsFastSuite)
+{
+    Model model;
+    model.CreateModelPart("dummy");
+
+    Parameters parameters{};
+
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(auto test = SolvingStrategyFactoryType::Create(
+            parameters, model.GetModelPart("dummy")), "The parameter strategy_type is undefined, aborting.")
```
avdg81: I would assign the return value and use that when you try to create the solving strategy:
```suggestion
    auto& dummy_model_part = model.CreateModelPart("dummy");

    Parameters parameters{};

    KRATOS_EXPECT_EXCEPTION_IS_THROWN(auto test = SolvingStrategyFactoryType::Create(
            parameters, dummy_model_part), "The parameter strategy_type is undefined, aborting.")
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_solving_strategy_factory.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+KRATOS_TEST_CASE_IN_SUITE(Create_ReturnsSolvingStrategy_ForNewtonRhapsonStrategy, KratosGeoMechanicsFastSuite)
+{
+    Model model;
+    model.CreateModelPart("dummy");
+    model.GetModelPart("dummy").SetBufferSize(3);
+    Parameters parameters{testParameters};
+
+    const auto created_strategy = SolvingStrategyFactoryType::Create(
+            parameters, model.GetModelPart("dummy"));
```
avdg81: I think we can simplify a bit here:
```suggestion
    auto& dummy_model_part = model.CreateModelPart("dummy", 3);
    Parameters parameters{testParameters};

    const auto created_strategy = SolvingStrategyFactoryType::Create(
            parameters, dummy_model_part);
```

If you feel that the argument value `3` reads too much like a magic number, you're free to give it a name, e.g.:
```
const auto buffer_size = 3;
auto& dummy_model_part = model.CreateModelPart("dummy", buffer_size);
```

applications/GeoMechanicsApplication/tests/run_cpp_unit_tests.py
avdg81: Just a question, wouldn't an `import` like the following also work?
```suggestion
import KratosMultiphysics.LinearSolversApplication
```

avdg81 (commented):
This looks good to me. Nice to have factories that we can extend as we see fit. I just had several minor comments, nothing major.

applications/GeoMechanicsApplication/tests/run_cpp_unit_tests.py
rfaasse: Seems to work, changed

applications/GeoMechanicsApplication/custom_utilities/builder_and_solver_factory.hpp
rfaasse: Agreed, done

applications/GeoMechanicsApplication/custom_utilities/scheme_factory.hpp
rfaasse: Agreed, done

applications/GeoMechanicsApplication/custom_utilities/scheme_factory.hpp
rfaasse: It also checked for both '_'and '-' for separators between words. I'm not too sure how flexible we should be here, what do you think?

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
rfaasse: Yes, makes it a bit more readable, agreed

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
rfaasse: Removed

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
rfaasse: As discussed, this is for the next PR

applications/GeoMechanicsApplication/tests/cpp_tests/test_builder_and_solver_factory.cpp
rfaasse: Indeed, I thought some functions would need to be implemented when I started this 'Dummy', but it turned out it was unnecessary, good catch

applications/GeoMechanicsApplication/tests/cpp_tests/test_builder_and_solver_factory.cpp
rfaasse: Removed

applications/GeoMechanicsApplication/tests/cpp_tests/test_convergence_criteria_factory.cpp
rfaasse: Agreed, done

applications/GeoMechanicsApplication/tests/cpp_tests/test_convergence_criteria_factory.cpp
rfaasse: Removed

applications/GeoMechanicsApplication/tests/cpp_tests/test_scheme_factory.cpp
rfaasse: Good catch, I swapped the names

applications/GeoMechanicsApplication/tests/cpp_tests/test_scheme_factory.cpp
rfaasse: Swapped

applications/GeoMechanicsApplication/tests/cpp_tests/test_solving_strategy_factory.cpp
rfaasse: Good point, simplified

applications/GeoMechanicsApplication/tests/cpp_tests/test_solving_strategy_factory.cpp
rfaasse: Ah, I missed that the create function returned a reference to the model part, good suggestion

applications/GeoMechanicsApplication/custom_utilities/scheme_factory.hpp
rfaasse: As discussed, let's leave it like this for now

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_builder_and_solver_factory.cpp
rfaasse: Unfortunately this is not feasible

applications/GeoMechanicsApplication/tests/cpp_tests/test_builder_and_solver_factory.cpp
rfaasse: As discussed, I took the template arguments from the respective types for all factories, so we can leave it like it is for now.

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
rfaasse: As discussed we'll investigate this later

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
```diff
@@ -0,0 +1,94 @@
...
+#pragma once
+
+#include <memory>
+#include <string>
+
+#include "solving_strategies/strategies/solving_strategy.h"
+#include "spaces/ublas_space.h"
+#include "linear_solvers/linear_solver.h"
+#include "solving_strategies/convergencecriterias/mixed_generic_criteria.h"
+#include "solving_strategy_factory.hpp"
```
avdg81: This seems like a self-`#include` to me...?

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+#pragma once
+
+#include <memory>
+#include <string>
+
+#include "solving_strategies/strategies/solving_strategy.h"
+#include "spaces/ublas_space.h"
+#include "linear_solvers/linear_solver.h"
+#include "solving_strategies/convergencecriterias/mixed_generic_criteria.h"
```
avdg81: I believe this `#include` is redundant, since we don't support mixed generic criteria yet, right?

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
```diff
@@ -0,0 +1,94 @@
...
+                    solver,
+                    criteria,
+                    builder_and_solver,
+                    strategy_parameters);
+        }
+
+        return nullptr;
+    }
+
+    static Parameters ExtractStrategyParameters(const Parameters &rSolverSettings)
```
avdg81, line 62: I think this member function can be made `private` (and therefore should be).

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
```diff
@@ -0,0 +1,94 @@
...
+//  Main authors:    Richard Faasse
+//
+
+#pragma once
+
+#include <memory>
+#include <string>
+
+#include "solving_strategies/strategies/solving_strategy.h"
+#include "spaces/ublas_space.h"
```
avdg81: Are we sure that we need this `#include`?

applications/GeoMechanicsApplication/tests/cpp_tests/test_builder_and_solver_factory.cpp
```diff
@@ -0,0 +1,63 @@
...
+    const auto elimination_builder = dynamic_cast<const ResidualBasedEliminationBuilderAndSolver<SparseSpaceType, LocalSpaceType, LinearSolverType>*>(builder_and_solver.get());
+    KRATOS_EXPECT_NE(elimination_builder, nullptr);
+}
+
+KRATOS_TEST_CASE_IN_SUITE(CreateBuilderAndSolver_Throws_WhenBlockBuilderIsUndefined, KratosGeoMechanicsFastSuite)
+{
+    const auto solver = std::make_shared<LinearSolverType>();
+    const auto factory = BuilderAndSolverFactory<SparseSpaceType, LocalSpaceType, LinearSolverType>();
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(factory.Create(Parameters{}, solver),
+                                      "the block_builder parameter is not defined, aborting BuilderAndSolverCreation")
```
avdg81: Since `Create` is a static member function, you'd better write it like this (just as for the two above test cases):
```suggestion
    KRATOS_EXPECT_EXCEPTION_IS_THROWN(BuilderAndSolverFactory<SparseSpaceType, LocalSpaceType, LinearSolverType>::Create(Parameters{}, solver),
                                      "the block_builder parameter is not defined, aborting BuilderAndSolverCreation")
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_convergence_criteria_factory.cpp
```diff
@@ -0,0 +1,57 @@
...
+//
+//  License:         geo_mechanics_application/license.txt
+//
+//  Main authors:    Richard Faasse
+//
+
+#include "testing/testing.h"
+#include "custom_utilities/convergence_criteria_factory.hpp"
+#include "spaces/ublas_space.h"
+#include "solving_strategies/convergencecriterias/mixed_generic_criteria.h"
```
avdg81: I believe we don't need this `#include`.

applications/GeoMechanicsApplication/tests/cpp_tests/test_convergence_criteria_factory.cpp
```diff
@@ -0,0 +1,57 @@
...
+    const auto convergence_criteria = ConvergenceCriteriaFactory<SparseSpaceType, LocalSpaceType>::Create(Parameters{validParameters});
+    const auto displacementCriterion = dynamic_cast<const DisplacementCriteria<SparseSpaceType, LocalSpaceType>*>(convergence_criteria.get());
+    KRATOS_EXPECT_NE(displacementCriterion, nullptr);
+}
+
+KRATOS_TEST_CASE_IN_SUITE(Create_Throws_WhenConvergenceCriterionDoesNotExist, KratosGeoMechanicsFastSuite)
+{
+    const ConvergenceCriteriaFactory<SparseSpaceType, LocalSpaceType> factory;
+
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(factory.Create(Parameters{"{}"}), "No convergence_criterion is defined, aborting.")
```
avdg81: Also here `Create` is a static member function:
```suggestion
    KRATOS_EXPECT_EXCEPTION_IS_THROWN(ConvergenceCriteriaFactory<SparseSpaceType, LocalSpaceType>::Create(Parameters{"{}"}), "No convergence_criterion is defined, aborting.")
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_convergence_criteria_factory.cpp
```diff
@@ -0,0 +1,57 @@
...
+}
+
+KRATOS_TEST_CASE_IN_SUITE(Create_Throws_WhenConvergenceCriterionIsUnknown, KratosGeoMechanicsFastSuite)
+{
+    const ConvergenceCriteriaFactory<SparseSpaceType, LocalSpaceType> factory;
+
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(factory.Create(Parameters{R"({"convergence_criterion" : "something_unknown" })"}),
```
avdg81:
```suggestion
    KRATOS_EXPECT_EXCEPTION_IS_THROWN(ConvergenceCriteriaFactory<SparseSpaceType, LocalSpaceType>::Create(Parameters{R"({"convergence_criterion" : "something_unknown" })"}),
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_convergence_criteria_factory.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include "spaces/ublas_space.h"
+#include "solving_strategies/convergencecriterias/mixed_generic_criteria.h"
+
+const std::string validParameters = R"(
+    {
+        "convergence_criterion":              "displacement_criterion",
+        "displacement_relative_tolerance":    1.0E-4,
+        "displacement_absolute_tolerance":    1.0E-9
+    }
+)";
```
avdg81: I would move this to the body of the single test case where these settings are being used.

applications/GeoMechanicsApplication/tests/cpp_tests/test_scheme_factory.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+    const auto parameters =
+    R"(
+    {
+        "solution_type" : "Quasi-Static"
+    }
+    )";
+
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(factory.Create(Parameters{parameters}),
+                                      "scheme_type is not defined, aborting")
```
avdg81:
```suggestion
    KRATOS_EXPECT_EXCEPTION_IS_THROWN(SchemeFactory<SparseSpaceType, LocalSpaceType>::Create(Parameters{parameters}),
                                      "scheme_type is not defined, aborting")
```
(And then variable `factory` is no longer needed.)

applications/GeoMechanicsApplication/tests/cpp_tests/test_scheme_factory.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+    const auto parameters =
+            R"(
+    {
+        "scheme_type" : "Backward_Euler"
+    }
+    )";
+
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(factory.Create(Parameters{parameters}),
+                                      "solution_type is not defined, aborting")
```
avdg81: Same thing here...

applications/GeoMechanicsApplication/tests/cpp_tests/test_scheme_factory.cpp
```diff
@@ -0,0 +1,88 @@
...
+    const auto parameters =
+            R"(
+    {
+        "scheme_type" : "NotSupported",
+        "solution_type" : "NotSupported"
+    }
+    )";
+
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(factory.Create(Parameters{parameters}),
+                                      "Specified solution_type/scheme_type is not supported, aborting")
```
avdg81: ... and here

applications/GeoMechanicsApplication/tests/cpp_tests/test_solving_strategy_factory.cpp
```diff
@@ -0,0 +1,118 @@
...
+    const Parameters parameters{};
+
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(auto test = SolvingStrategyFactoryType::Create(
+            parameters, dummy_model_part), "The parameter strategy_type is undefined, aborting.")
+}
+
+KRATOS_TEST_CASE_IN_SUITE(Create_ReturnsSolvingStrategy_ForNewtonRhapsonStrategy, KratosGeoMechanicsFastSuite)
+{
+    Model model;
+    const int buffer_size = 3;
```
avdg81: As Wijtze Pieter suggested, let's change this value to `2` :-)

applications/GeoMechanicsApplication/tests/cpp_tests/test_solving_strategy_factory.cpp
```diff
@@ -0,0 +1,118 @@
...
+    auto& dummy_model_part = model.CreateModelPart("dummy", buffer_size);
+    const Parameters parameters{testParameters};
+
+    const auto created_strategy = SolvingStrategyFactoryType::Create(
+            parameters, dummy_model_part);
+
+    KRATOS_EXPECT_NE(created_strategy, nullptr);
+    created_strategy->Check();
```
avdg81: Was this function call added intentionally? If yes, then I would have expected some kind of checking. Perhaps its return value. Or whether it throws (or not). The way it is written now confuses me a bit.

avdg81 (commented):
Thanks for the great work Richard! I still found several minor things that you may want to address.

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
rfaasse: Good point, I missed cleaning up the includes, which were still there from trying it the 'dgeoflow' hardcoded way

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
rfaasse: cleaned up the includes

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
rfaasse: cleaned up the includes

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
rfaasse: Agreed, done

applications/GeoMechanicsApplication/tests/cpp_tests/test_builder_and_solver_factory.cpp
rfaasse: I agree and I tried this before, but for some reason the macro doesn't support it
image

applications/GeoMechanicsApplication/tests/cpp_tests/test_convergence_criteria_factory.cpp
rfaasse: Indeed, cleaned up (although Wijtze Pieter mentioned we probably need to add the mixed criteria for settlement, but that will be a separate issue)

applications/GeoMechanicsApplication/tests/cpp_tests/test_convergence_criteria_factory.cpp
rfaasse: Agreed, I thought I would need to re-use it, but that's not the case

applications/GeoMechanicsApplication/tests/cpp_tests/test_convergence_criteria_factory.cpp
rfaasse: Same answer as before, the macro doesn't seem to accept it

applications/GeoMechanicsApplication/tests/cpp_tests/test_convergence_criteria_factory.cpp
rfaasse: Same answer as before, the macro doesn't seem to accept it

applications/GeoMechanicsApplication/tests/cpp_tests/test_scheme_factory.cpp
rfaasse: Same answer as before, the macro doesn't seem to accept it

applications/GeoMechanicsApplication/tests/cpp_tests/test_scheme_factory.cpp
rfaasse: Same answer as before, the macro doesn't seem to accept it

applications/GeoMechanicsApplication/tests/cpp_tests/test_scheme_factory.cpp
rfaasse: Same answer as before, the macro doesn't seem to accept it

applications/GeoMechanicsApplication/tests/cpp_tests/test_solving_strategy_factory.cpp
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_solving_strategy_factory.cpp
rfaasse: Agreed, I implicitly check if it throws by just having it there. I added the check for the return value.

applications/GeoMechanicsApplication/tests/cpp_tests/test_builder_and_solver_factory.cpp
avdg81: Yes, you're right. However, you may overcome the above problem by wrapping the first argument of the macro in a pair of parenthesis, i.e.:
```suggestion
    KRATOS_EXPECT_EXCEPTION_IS_THROWN((BuilderAndSolverFactory<SparseSpaceType, LocalSpaceType, LinearSolverType>::Create(Parameters{}, solver)),
                                      "the block_builder parameter is not defined, aborting BuilderAndSolverCreation")
```

rfaasse: This works indeed, and what also does is creating an alias (which is better for readability anyway), since in this case the problem of the macro seems to be related to having the template arguments there.

applications/GeoMechanicsApplication/tests/cpp_tests/test_convergence_criteria_factory.cpp
rfaasse: Done using your suggestion

applications/GeoMechanicsApplication/tests/cpp_tests/test_convergence_criteria_factory.cpp
rfaasse: Done using your suggestion

applications/GeoMechanicsApplication/tests/cpp_tests/test_scheme_factory.cpp
rfaasse: Done using your suggestion

applications/GeoMechanicsApplication/tests/cpp_tests/test_scheme_factory.cpp
rfaasse: Done using your suggestion

applications/GeoMechanicsApplication/tests/cpp_tests/test_scheme_factory.cpp
rfaasse: Done using your suggestion

rfaasse (commented):
Got the static calls in, feel free to re-review

avdg81 (approved):
This looks good and clear to me. Please merge.

PR #11666, opened 2023-10-05
[Core] Adding `Id` methods to `SpatialSearchResult`

loumalouomega:
**📝 Description**

Alternative to https://github.com/KratosMultiphysics/Kratos/pull/11414

Add ID retrieval methods to `SpatialSearchResult`. This PR adds two methods to the `SpatialSearchResult` class. These methods allow for the retrieval of the ID of the object stored in the result. If the object is derived from `IndexedObject`, these methods return the object's ID; otherwise, they return 0. Additionally, some code formatting and comments have been improved for clarity.

**🆕 Changelog**

- [Adding `Id` methods to `SpatialSearchResult`](https://github.com/KratosMultiphysics/Kratos/commit/56ab97cf0d5697b06cc23afb2314a5fd226513d6)


RiccardoRossi (commented):
what if you want to search points? (without id)

loumalouomega:
> what if you want to search points? (without id)

It can be solved if I use the pointer as Id, wait

RiccardoRossi:
I would say you are not right, in the sense that you can only retrieve global pointers via the function you passed me yesterday if they are objects with Id. In case they are not, you will need another strategy to retrieve the global pointer.

loumalouomega:
> I would say you are not right, in the sense that you can only retrieve global pointers via the function you passed me yesterday if they are objects with Id. In case they are not, you will need another strategy to retrieve the global pointer.

In that case a priori it will not work, but the Id is hardcoded in the code I sent you yesterday. I would need to modify the global pointer utilities, which require a lot of work.

loumalouomega:
After reviewing my code, I noticed that all implementations are based on the assumption that the search should be carried out in a GeometricalObject or Node. Upon discussing with @pooyan_dadvand, I recalled that we had indeed decided to prioritize these classes some time ago. Hence, the implementation is accurate and will address our problem. Nonetheless, I'll incorporate a warning to notify users when they use the Id method with incompatible classes.

kratos/spatial_containers/spatial_search_result.h
```diff
@@ -143,6 +144,7 @@ class SpatialSearchResult
         if constexpr (std::is_base_of_v<IndexedObject, TObjectType>) {
             return mpObject->Id();
         } else { // Otherwise 0
+            KRATOS_WARNING("SpatialSearchResult") << "Object does not provide an ID" << std::endl;
```
philbucher: what does this mean? I.e. what should I do as user?

I feel like if some object (i.e. objects without IDs) should not be allowed to be used if they anyway wont work

loumalouomega: Do you prefer an error?, I have doubts about it

philbucher: ideally an error where this `SpatialSearchResult` is used, at construction time

loumalouomega: The Id it is a new method, I would add the error only when calling the Id method, not everytime the class is constructed, because may not be used at all

loumalouomega: Replaced

kratos/spatial_containers/spatial_search_result.h
```diff
@@ -109,6 +116,39 @@ class SpatialSearchResult
...
+    /**
+    * @brief Get the ID of the object.
+    * @details This function essentially performs the same operation as the Id() function. If the object is derived from IndexedObject, retrieves its ID. Otherwise, returns 0.
+    * @note Consider refactoring to avoid redundancy with Id() function.
+    * @return IndexType - The ID of the object or 0 if not derived from IndexedObject.
+    */
+    IndexType GetId() const
+    {
+        // Get Id if the object is derived from IndexedObject
+        if constexpr (std::is_base_of_v<IndexedObject, TObjectType>) {
```
RiccardoRossi, line 144: sorry but I would say that this is simply not correct ... the spatial search result is a global_pointer, so you cannot call directly a method on it as you don't know if it is local

RiccardoRossi: aside, if you need to call a function that requires an id onto an object without Id, i would say it is better not to overload the operator and to call it directly Id onto the result. If the function Id is not there it will give you a compile time error...

loumalouomega: I will close this until I have figure out what to do

PR #11669, opened 2023-10-08
[GeoMechanicsApplication] Moved logic of `ApplyScalarConstraintsTableProcess` to C++

avdg81:
**📝 Description**
This class used to be defined in a Python module only. Now that we also need it for the custom settlement workflow, we had to port it from Python to C++. Note that the Python interface now uses the new C++ class.

Also added a creator for the `ApplyScalarConstraintsTableProcess` to the process factory used by the custom workflow. The factory should now be able to create all kinds of processes that are relevant for settlement analysis.


applications/GeoMechanicsApplication/custom_processes/apply_scalar_constraints_table_process.cpp
```diff
@@ -0,0 +1,227 @@
...
+        KRATOS_ERROR << "Unknown fluid_pressure_type: " << fluid_pressure_type << std::endl;
+    }
+}
+
+void ApplyScalarConstraintsTableProcess::MakeProcessForNonFluidPressureType(const Parameters&        rProcessSettings,
+                                                                            std::vector<std::string> NamesOfSettingsToCopy)
+{
+    NamesOfSettingsToCopy.emplace_back("value");
+
+    if (rProcessSettings["table"].GetInt() == 0) {
```
rfaasse: `rProcessSettings["table"].GetInt() == 0` is used a lot. We could consider making it a function with a descriptive name (then it's also clearer what this check means, because to me that's not clear and if the hardcoded "table" would ever change, we only have to change it in one place)

applications/GeoMechanicsApplication/custom_processes/apply_scalar_constraints_table_process.cpp
```diff
@@ -0,0 +1,227 @@
...
+        mProcess = std::make_unique<ApplyConstantInterpolateLinePressureProcess>(mrModelPart,
+                                                                                 ExtractParameters(rProcessSettings,
+                                                                                                   NamesOfSettingsToCopy));
+    } else {
+        KRATOS_ERROR << "No time dependent interpolate line pressure process available" << std::endl;
+    }
+}
+
+void ApplyScalarConstraintsTableProcess::MakeProcessForPhreaticSurface(const Parameters&        rProcessSettings,
```
WPK4FEM, line 169: This one belongs together with or close to the line and hydrostatic for functional reasons. The interpolate line process is a deviating functionality.

applications/GeoMechanicsApplication/custom_processes/apply_scalar_constraints_table_process.cpp
```diff
@@ -0,0 +1,227 @@
...
+                                                                                       NamesOfSettingsToCopy));
+    } else {
+        NamesOfSettingsToCopy.emplace_back("table");
+        mProcess = std::make_unique<ApplyComponentTableProcess>(mrModelPart,
+                                                                ExtractParameters(rProcessSettings,
+                                                                                  NamesOfSettingsToCopy));
+    }
+}
+
+void ApplyScalarConstraintsTableProcess::MakeProcessForUniformFluidPressure(const Parameters&        rProcessSettings,
```
WPK4FEM: this to me looks exactly the same as the implementations of MakeProcessForNonFluidPressureType. Why have 2.

applications/GeoMechanicsApplication/custom_processes/apply_scalar_constraints_table_process.cpp
```diff
@@ -0,0 +1,227 @@
...
+void ApplyScalarConstraintsTableProcess::MakeProcessForHydrostaticFluidPressure(const Parameters&        rProcessSettings,
+                                                                                std::vector<std::string> NamesOfSettingsToCopy)
+{
+    NamesOfSettingsToCopy.insert(NamesOfSettingsToCopy.end(), {"gravity_direction",
+                                                               "reference_coordinate",
+                                                               "specific_weight"});
+    AppendParameterNameIfExists("pressure_tension_cut_off", rProcessSettings, NamesOfSettingsToCopy);
+    AppendParameterNameIfExists("is_seepage", rProcessSettings, NamesOfSettingsToCopy);
+
+    if (rProcessSettings["table"].GetInt() == 0) {
```
WPK4FEM: The test for a table on hydrostatic, phreatic line and phreatic surface looks like any of an array that has length 1, 2 or 3 is unequal to zero. The test of using a table is more natural to me than not using a table. Can we put the test if table functionality should be used in a function?

applications/GeoMechanicsApplication/custom_processes/apply_scalar_constraints_table_process.cpp
```diff
@@ -0,0 +1,227 @@
...
+    } else if (fluid_pressure_type == "Interpolate_Line") {
+        MakeProcessForInterpolatedLine(rProcessSettings, std::move(NamesOfSettingsToCopy));
+    } else if (fluid_pressure_type == "Phreatic_Surface") {
+        MakeProcessForPhreaticSurface(rProcessSettings, std::move(NamesOfSettingsToCopy));
+    } else {
+        KRATOS_ERROR << "Unknown fluid_pressure_type: " << fluid_pressure_type << std::endl;
+    }
+}
+
+void ApplyScalarConstraintsTableProcess::MakeProcessForNonFluidPressureType(const Parameters&        rProcessSettings,
```
WPK4FEM: This is the general thing, the fluid pressures the specialized thing. I would prefer if the general thing has a general name and does not refer to not being the special thing.

applications/GeoMechanicsApplication/custom_processes/apply_scalar_constraints_table_process.cpp
```diff
@@ -0,0 +1,227 @@
...
+ApplyScalarConstraintsTableProcess::~ApplyScalarConstraintsTableProcess() = default;
+
+void ApplyScalarConstraintsTableProcess::MakeInternalProcess(const Parameters& rProcessSettings)
+{
+    auto names_of_settings_to_copy = std::vector<std::string>{"model_part_name",
+                                                              "variable_name"};
+    AppendParameterNameIfExists("is_fixed", rProcessSettings, names_of_settings_to_copy);
+
+    if (rProcessSettings.Has("fluid_pressure_type")) {
+        MakeProcessForFluidPressureType(rProcessSettings, std::move(names_of_settings_to_copy));
```
rfaasse, line 82: I'm not sure if I agree with the necessity of using std::move for a vector with only two strings. I think passing it as a ref would be fine as well and a bit less confusing, but I leave it up to you, both ways are acceptable to me.

WPK4FEM (commented):
Thank you, this is more reusable than the python interface.

applications/GeoMechanicsApplication/custom_processes/apply_scalar_constraints_table_process.h
```diff
@@ -0,0 +1,63 @@
...
+// ((____/ / ((____   ((___/ /  MECHANICS
+//
+//  License:         geo_mechanics_application/license.txt
+//
+//  Main authors:    Anne van de Graaf,
+//                   Marjan Fathian
+//
+#pragma once
+
+#include "processes/process.h"
```
rfaasse, line 15: I think here Process can also be forward declared (like ModelPart and Parameters)

rfaasse (dismissed):
Clear and clean code! I have some minor suggestions, nothing blocking. 

On the point of UTs: looking at the class it's not soo trivial to unit-test and since it's already tested by the python tests, I don't think it is worth the effort at this point in time

applications/GeoMechanicsApplication/custom_processes/apply_scalar_constraints_table_process.cpp
avdg81: Done together :-)

applications/GeoMechanicsApplication/custom_processes/apply_scalar_constraints_table_process.cpp
avdg81: Done.

applications/GeoMechanicsApplication/custom_processes/apply_scalar_constraints_table_process.cpp
avdg81: Done.

applications/GeoMechanicsApplication/custom_processes/apply_scalar_constraints_table_process.cpp
avdg81: Done.

applications/GeoMechanicsApplication/custom_processes/apply_scalar_constraints_table_process.h
avdg81: We tried, but this class derives from `Process`, so no ;-)

applications/GeoMechanicsApplication/custom_processes/apply_scalar_constraints_table_process.cpp
avdg81: I am stubborn and insist on using `std::move` here ;-). My primary motivation is to show that I want to change the scope of the list of setting names.

applications/GeoMechanicsApplication/custom_processes/apply_scalar_constraints_table_process.cpp
avdg81: We have renamed the member function, so done.

rfaasse (approved):
Looks good, thanks for incorporating the review comments!

PR #11674, opened 2023-10-11
[GeoMechanicsApplication] Renamed a few files and classes as suggested by @carloslubbers

### LINKED ISSUE #11673: [GeoMechanicsApplication] Rename scalar and vector constraint table processes
As pointed out by @carloslubbers, the names of the process classes that apply scalar and vector constraints with an optional table have over time changed from a singular form ("constraint") to a plural form ("constraints"). To ensure consistency with the D-GeoFlow workflow, the names need to be changed back to their original singular form.

avdg81:
**📝 Description**
The header and implementation files that contain the process classes for applying a scalar or vector constraint with an optional table have been renamed. We now use the singular form ("constraint") rather than the plural one ("constraints").

The class names now use singular forms, i.e.
`ApplyScalarConstraintTableProcess` and
`ApplyVectorConstraintTableProcess`. All references to the old class names have been updated.

### DIFF
--- a/applications/GeoMechanicsApplication/custom_processes/apply_scalar_constraint_table_process.cpp
+++ b/applications/GeoMechanicsApplication/custom_processes/apply_scalar_constraint_table_process.cpp
@@ -10,7 +10,7 @@
 //  Main authors:    Anne van de Graaf,
 //                   Marjan Fathian
 //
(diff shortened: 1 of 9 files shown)


rfaasse (approved):
Looks good to me!

PR #11677, opened 2023-10-12
[GeoMechanicsApplication] Geo/time step loop

### LINKED ISSUE #11620: [GeoMechanicsApplication] Add a class for running a time loop
As a developer, I would like to be able to run a full time loop given a solver strategy and a time incrementor.

#### Background

The Kratos team at Deltares would like to be able to carry out a time loop in custom workflows. Being able to execute a time loop is one of those components. The time loop itself must be independent of the selected solver strategy and time incrementor. In that way, we can create various compositions that satisfy our requirements of the workflow.

#### Acceptance criteria

**Given** a time loop executor, a time incrementor that generates four constant time increments, a solver strategy that always converges, and no start state
**When** the developer requests to run the time loop (by calling member function `TimeLoopExecutor::Run(...)`)
**Then** it must return a list of five step states (start state + four increments in time), each one indicating that it converged.

**Given** a time loop executor, a time incrementor that generates four constant time increments, a solver strategy that always converges, and a start state
**When** the developer requests to run the time loop (by calling member function `TimeLoopExecutor::Run(...)`)
**Then** it must return a list of four step states (one for each increment in time), each one indicating that it converged.

WPK4FEM:
TIme loop executor that performs the stepping loop and for each step the cycle loop using TimeStepExecutor and StrategyWrapper..

Contains:
- Unit tests for all developed components
- CloneTimeStep for claiming variable space at start of step
- RestorePosistionsAndDOFVectorToStartOfStep for enabeling the cycling mechanism
- SaveTotalDsplacementFieldAtStartOfStage and AccumulateTotalDisplacementField for keeping track of displacement accumulation over stages
- OutputProces to enable output at the end of each step


applications/GeoMechanicsApplication/CMakeLists.txt
```diff
@@ -39,7 +39,7 @@ endif(${KRATOS_BUILD_TESTING} MATCHES ON)
 ## GeoMechanics python interface sources
 file(GLOB_RECURSE KRATOS_GEO_MECHANICS_APPLICATION_PYTHON_INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/custom_python/*.cpp)
 
-add_library(KratosGeoMechanicsCore SHARED ${KRATOS_GEO_MECHANICS_APPLICATION_CORE} ${KRATOS_GEO_MECHANICS_APPLICATION_TESTING_SOURCES})
+add_library(KratosGeoMechanicsCore SHARED ${KRATOS_GEO_MECHANICS_APPLICATION_CORE} ${KRATOS_GEO_MECHANICS_APPLICATION_TESTING_SOURCES} tests/cpp_tests/test_time_loop_executor.cpp custom_workflows/time_step_end_state.hpp)
```
rfaasse: I think these were added by mistake and can be removed again.

applications/GeoMechanicsApplication/custom_workflows/custom_workflow_factory.cpp
```diff
@@ -12,7 +12,7 @@
 
 #include "custom_workflow_factory.h"
 #include "dgeosettlement.h"
-#include "time_loop_executor.h"
+#include "time_loop_executor.hpp"
```
rfaasse, line 15: Let's discuss, I think since the 'real' strategy wrapper will become a template class, this still might need to be .hpp but most probably it can be split into .h and .cpp

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.h
```diff
@@ -27,7 +28,6 @@ namespace Kratos
 class ProcessFactory;
 class InputUtility;
 class ProcessInfoParser;
-class TimeLoopExecutor;
```
rfaasse, line 30: I think we can keep this forward declaration and remove the include from dgeosettlement.h

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+    virtual ~TimeLoopExecutor() = default;
+    virtual void SetProcessReferences(const std::vector<std::weak_ptr<Process>>& rProcessRefs) {}
+
+    void SetTimeIncrementor(std::unique_ptr<TimeIncrementor> ATimeIncrementor)
+    {
+        mTimeIncrementor = std::move(ATimeIncrementor);
+    }
+
+    void SetSolverStrategyTimeIncrementor(std::shared_ptr<StrategyWrapper> AStrategyWrapper)
```
rfaasse: I think this function should be renamed to SetSolverStrategyWrapper

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
```diff
@@ -0,0 +1,102 @@
...
+{
+
+class Process;
+
+class TimeLoopExecutor{
+public :
+    TimeLoopExecutor() : mTimeStepExecutor{std::make_unique<TimeStepExecutor>()} {}
+
+    virtual ~TimeLoopExecutor() = default;
+    virtual void SetProcessReferences(const std::vector<std::weak_ptr<Process>>& rProcessRefs) {}
```
rfaasse: This function still misses an implementation

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+    void SetTimeIncrementor(std::unique_ptr<TimeIncrementor> ATimeIncrementor)
```
rfaasse: When looking at the [code style guide](https://github.com/KratosMultiphysics/Kratos/wiki/Style-Guide#arguments), this argument should be called pTimeIncrementor.

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
rfaasse: When looking at the [code style guide](https://github.com/KratosMultiphysics/Kratos/wiki/Style-Guide#arguments), this argument should be called pStrategyWrapper.

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
```diff
@@ -0,0 +1,102 @@
...
+        mTimeIncrementor = std::move(ATimeIncrementor);
+    }
+
+    void SetSolverStrategyTimeIncrementor(std::shared_ptr<StrategyWrapper> AStrategyWrapper)
+    {
+        mStrategyWrapper = std::move(AStrategyWrapper);
+        mTimeStepExecutor->SetSolverStrategy(mStrategyWrapper);
+    }
+
+    std::vector<TimeStepEndState> Run(TimeStepEndState end_state)
```
rfaasse: Since this is an argument, end_state should be EndState

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    TimeStepEndState RunCycleLoop(const TimeStepEndState& previous_state)
+    {
+        auto cycle_number = 0;
+        auto end_state = previous_state;
+        while (mTimeIncrementor->WantRetryStep(cycle_number, end_state)) {
+            if (cycle_number > 0)
+            {
+                // restore the nodal positions and D.O.F. to start of step state
```
rfaasse: Since the function name is really descriptive, I think this comment is redundant

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
```diff
@@ -0,0 +1,359 @@
...
+    void Initialize()             override {}
+    void InitializeSolutionStep() override {}
+    void Predict()                override {}
+    bool SolveSolutionStep()      override { return false; }
+    void FinalizeSolutionStep()   override {}
+
+private:
+    TimeStepEndState::ConvergenceState mConvergenceState{TimeStepEndState::ConvergenceState::converged};
+    Model mModel;
+    ModelPart * mpModelPart = nullptr;
```
rfaasse, line 148: Since this is a dummy and this functionality will be implemented in the 'real' strategy wrapper, maybe we should remove the model and modelpart from this dummy class.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
```diff
@@ -44,6 +44,17 @@ class DummyStrategyWrapper : public StrategyWrapper
     [[nodiscard]] TimeStepEndState::ConvergenceState GetConvergenceState()         override { return mConvergenceState;};
     [[nodiscard]] std::size_t                        GetNumberOfIterations() const override { return 4;};
     [[nodiscard]] double                             GetEndTime()            const override { return 10.;};
+    void SetEndTime(double EndTime) override {}
```
rfaasse: I think we should have a look at the possibility of merging these two dummy's, since it poses extra work every time the wrapper interface changes

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
```diff
@@ -0,0 +1,359 @@
...
+}
+
+KRATOS_TEST_CASE_IN_SUITE(ExpectOutputProcessCalledForEveryStep, KratosGeoMechanicsFastSuite)
+{
+    TimeLoopExecutor executor;
+    const auto wanted_num_of_cycles_per_step = 1;
+    executor.SetTimeIncrementor(std::make_unique<FixedCyclesTimeIncrementor>(wanted_num_of_cycles_per_step));
+    auto solver_strategy = std::make_shared<DummySolverStrategy>(TimeStepEndState::ConvergenceState::converged);
+    executor.SetSolverStrategyTimeIncrementor(solver_strategy);
+    const auto step_states = executor.Run(TimeStepEndState{});
```
rfaasse: I think we should make a function out of these 5 lines. If we add the wanted_num_of_cycles_per_step as an input parameter, we can re-use it in the last 7 tests, i.e.

```
std::shared_ptr<DummySolverStrategy> RunFixedCycleTimeLoop(int wanted_num_of_cycles_per_step)
{
    TimeLoopExecutor executor;
    executor.SetTimeIncrementor(std::make_unique<FixedCyclesTimeIncrementor>(wanted_num_of_cycles_per_step));
    auto solver_strategy = std::make_shared<DummySolverStrategy>(TimeStepEndState::ConvergenceState::converged);
    executor.SetSolverStrategyTimeIncrementor(solver_strategy);
    const auto step_states = executor.Run(TimeStepEndState{});

    return solver_strategy;
}
```

applications/GeoMechanicsApplication/tests/cpp_tests/stub_time_loop_executor.h
```diff
@@ -12,7 +12,7 @@
 
 #pragma once
 
-#include "custom_workflows/time_loop_executor.h"
+#include "custom_workflows/time_loop_executor.hpp"
```
rfaasse: I think this is not working correctly now. This stub/dummy should not derive from the 'real' time loop executor, we're missing an interface above it.

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.h
```diff
@@ -1,29 +0,0 @@
-// KRATOS___
```
rfaasse, line 1: This class was meant as the purely virtual interface, such that we can make a stub in one of the tests. We need to extract an interface from which both the real and dummy time_loop_executors derive

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
```diff
@@ -0,0 +1,359 @@
...
+    executor.SetSolverStrategyTimeIncrementor(std::make_unique<DummySolverStrategy>(TimeStepEndState::ConvergenceState::converged));
+    const auto step_states = executor.Run(TimeStepEndState{});
+
+    // The test time incrementor assumes fixed time increments of 0.5 and an end time of 1.0
+    KRATOS_EXPECT_EQ(2, step_states.size());
+    KRATOS_EXPECT_DOUBLE_EQ(0.5, step_states[0].time);
+    KRATOS_EXPECT_EQ(wanted_num_of_cycles_per_step, step_states[0].num_of_cycles);
+    KRATOS_EXPECT_DOUBLE_EQ(1.0, step_states[1].time);
+    KRATOS_EXPECT_EQ(wanted_num_of_cycles_per_step, step_states[1].num_of_cycles);
+}
```
rfaasse: I don't really see why we need an extra test for checking 5 specifically. Is there a reason the 3 would work and the 5 wouldn't?

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
```diff
@@ -0,0 +1,359 @@
...
+    KRATOS_EXPECT_TRUE(std::all_of(step_states.begin(), step_states.end(),
+                       [](const auto& state) {return state.num_of_cycles == 1;}))
+    auto time = 0.0;
+    for (auto index = std::size_t{0}; index < increments.size(); ++index) {
+        time += increments[index];
+        KRATOS_EXPECT_DOUBLE_EQ(time, step_states[index].time);
+    }
+}
+
+KRATOS_TEST_CASE_IN_SUITE(ExpectEndTimeToBeSetAfterRunningAStep, KratosGeoMechanicsFastSuite)
```
rfaasse: I think it would be clearer to rename ExpectEndTimeToBeSetAfterRunningAStep to ExpectEndTimeToBeSetAfterRunningATimeLoop

rfaasse (changes requested):
Very nice readable and clean code (and of course well-tested)! 

I think there is the possibility for some more clean-up of the tests and there are some discussion points about the structure.

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
```diff
@@ -0,0 +1,102 @@
...
+//                   Wijtze Pieter Kikstra
+//
+
+#pragma once
+
+#include <vector>
+#include <memory>
+
+#include "time_step_end_state.hpp"
+#include "time_incrementor.h."
```
rfaasse: There's a typo here that breaks some of the builds (I have no idea how this does build on windows btw)

applications/GeoMechanicsApplication/CMakeLists.txt
WPK4FEM: Done

applications/GeoMechanicsApplication/custom_workflows/custom_workflow_factory.cpp
WPK4FEM: Postponed to the pull request that binds together all previous work.

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.h
WPK4FEM: Done by trial, no error.

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
WPK4FEM: Rename to SetSolverStrategyTimeStepExecutor instead.

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
WPK4FEM: Done

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
WPK4FEM: Done

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
WPK4FEM: Done

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
WPK4FEM: Done and reduced to 1 line.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
WPK4FEM: The TimeStepExecutor needs the model part to store its information. So left as is.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
WPK4FEM: Implementations are different, therefore left as is for the time being.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
WPK4FEM: Kept 5 as the two tests otherwise are identical.

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
WPK4FEM: Oeps ... Done.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
WPK4FEM: Test name adapted.

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
WPK4FEM: Done ( and changed the vector of processRefs into the desired weak_pointers.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.h
WPK4FEM: Done

applications/GeoMechanicsApplication/tests/cpp_tests/stub_time_loop_executor.h
WPK4FEM: Done

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor_interface.h
```diff
@@ -0,0 +1,38 @@
...
+#include "time_step_end_state.hpp"
+#include "time_incrementor.h"
+#include "strategy_wrapper.hpp"
+
+namespace Kratos
+{
+
+class Process;
+
+class TimeLoopExecutorInterface{
```
rfaasse, line 29: @avdg81 We discussed we didn't want Interface in the name, but @WPK4FEM and I could not think of a suitable pair of names for the interface + implementation of the time loop executor. If you have a suggestion, please let us know!

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
```diff
@@ -127,10 +127,10 @@ int KratosGeoSettlement::RunStage(const std::filesystem::path&            rWorki
...
         std::vector<std::weak_ptr<Process>> process_observables(processes.begin(), processes.end());
 
-        if (mpTimeLoopExecutor)
-        {
-            mpTimeLoopExecutor->SetProcessReferences(process_observables);
-        }
+    if (mpTimeLoopExecutor)
+    {
+        mpTimeLoopExecutor->SetProcessObservables(process_observables);
+    }
```
avdg81: Nitpicking: by looking at the context, the indentation doesn't seem to be correct.
```suggestion
        if (mpTimeLoopExecutor)
        {
            mpTimeLoopExecutor->SetProcessObservables(process_observables);
        }
```

Also (for later, when we start to compose things) I feel that we must have a time loop executor, since without it there is nothing useful that we can do. Therefore, the constructor of this class should check whether the smart pointer is not null. Our class invariant then becomes we can safely assume that the pointer has been set (no further checks needed).

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
```diff
@@ -0,0 +1,99 @@
...
+    {
+        mTimeStepExecutor->SetProcessObservables(rProcessObservables);
+    }
+
+    void SetTimeIncrementor(std::unique_ptr<TimeIncrementor> pTimeIncrementor) override
+    {
+        mTimeIncrementor = std::move(pTimeIncrementor);
+    }
+
+    void SetSolverStrategyTimeStepExecutor(std::shared_ptr<StrategyWrapper> pStrategyWrapper) override
```
avdg81: I have a renaming suggestion: what about `SetStrategyWrapper`? It is (much) shorter, but it still makes clear what it does. What do you think?

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
```diff
@@ -0,0 +1,99 @@
...
+    {
+        mStrategyWrapper = std::move(pStrategyWrapper);
+        mTimeStepExecutor->SetSolverStrategy(mStrategyWrapper);
+    }
+
+    std::vector<TimeStepEndState> Run(TimeStepEndState EndState) override
+    {
+        mStrategyWrapper->SaveTotalDisplacementFieldAtStartOfStage();
```
avdg81: Note that this function name suggests that the start of a time loop always coincides with the start of a stage. I'm not sure whether that applies to all use cases of the time loop executor. Perhaps it does. Still, we may consider to rename it to `SaveTotalDisplacementFieldAtStartOfTimeLoop`.

**By the way, what about saving the pore pressure field at the start of the time loop??**

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
```diff
@@ -0,0 +1,99 @@
...
+        std::vector<TimeStepEndState> result;
+        while (mTimeIncrementor->WantNextStep(EndState)) {
+            mStrategyWrapper->IncrementStepNumber();
+            // clone without end time, the end time is overwritten anyway
+            mStrategyWrapper->CloneTimeStep();
+            EndState = RunCycleLoop(EndState);
+            mStrategyWrapper->AccumulateTotalDisplacementField();
+            mStrategyWrapper->FinalizeSolutionStep();
+            result.emplace_back(EndState);
+            mStrategyWrapper->OutputProcess();
```
avdg81: It may read more logically if we swap these two statements (that shouldn't have any side effects, rights?).
```suggestion
            mStrategyWrapper->OutputProcess();
            result.emplace_back(EndState);
```

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor_interface.h
avdg81: No, I don't have a better suggestion either. Let's keep the class names as they are now.

applications/GeoMechanicsApplication/custom_workflows/time_step_executor.cpp
```diff
@@ -21,31 +21,31 @@ void TimeStepExecutor::SetSolverStrategy(std::shared_ptr<StrategyWrapper> Solver
...
 
     mStrategyWrapper->Predict();
     mStrategyWrapper->SolveSolutionStep();
 
-    for (const auto& process : mProcessRefs)
+    for (const auto& process_observable : mProcessObservables)
     {
-         process.get().ExecuteFinalizeSolutionStep();
+        auto process = process_observable.lock();
+        process->ExecuteFinalizeSolutionStep();
```
avdg81: I think we need to check whether `process` is not empty. Note that `lock` returns a `shared_ptr` object that temporarily partakes in ownership of the managed object (see [this page](https://en.cppreference.com/w/cpp/memory/weak_ptr/lock)). Now when the managed object has been destructed, we get an empty `shared_ptr`, resulting in undefined behavior when attempting to access it.
```suggestion
        if (process) {
            process->ExecuteFinalizeSolutionStep();
        }
```
We may consider to issue a warning when the managed object no longer exists, since that was probably not intended.

This comment also applies to the other loop over the process observables.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
```diff
@@ -93,24 +120,26 @@ KRATOS_TEST_CASE_IN_SUITE(ProcessMemberFunctionsAllCalledOnce, KratosGeoMechanic
     TimeStepExecutor executor;
     auto converging_strategy = std::make_shared<DummyStrategyWrapper>(TimeStepEndState::ConvergenceState::converged);
     executor.SetSolverStrategy(converging_strategy);
-    ProcessSpy spy;
-    TimeStepExecutor::ProcessRefVec process_refs{spy};
-    executor.SetProcessReferences(process_refs);
+    auto spy = std::make_shared<ProcessSpy>();
+
+    std::vector<std::shared_ptr<Process>> processes{spy};
+    std::vector<std::weak_ptr<Process>> process_observables(processes.begin(), processes.end());
```
avdg81: I believe we can simplify this to:
```suggestion
    std::vector<std::weak_ptr<Process>> process_observables{spy};
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
```diff
@@ -93,24 +120,26 @@ KRATOS_TEST_CASE_IN_SUITE(ProcessMemberFunctionsAllCalledOnce, KratosGeoMechanic
...
-KRATOS_TEST_CASE_IN_SUITE(SolverStrategyMemberFunctionsAllCalledOnce, KratosGeoMechanicsFastSuite)
+KRATOS_TEST_CASE_IN_SUITE(SolverStrategyMemberFunctionsAllExceptFinalizeCalledOnce, KratosGeoMechanicsFastSuite)
 {
     TimeStepExecutor executor;
     auto converging_strategy = std::make_shared<DummyStrategyWrapper>(TimeStepEndState::ConvergenceState::converged);
     executor.SetSolverStrategy(converging_strategy);
-    TimeStepExecutor::ProcessRefVec process_refs;
-    executor.SetProcessReferences(process_refs);
+    std::vector<std::weak_ptr<Process>> process_observables;
+    executor.SetProcessObservables(process_observables);
```
avdg81: I think these lines are redundant. On construction, the list of process observables that is kept by the time step executor is empty. Explicitly passing it an empty vector of process observables won't change that. Better to keep the test as simple (and short) as possible.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
```diff
@@ -0,0 +1,356 @@
...
+public:
+    explicit DummySolverStrategy(TimeStepEndState::ConvergenceState AConvergenceState)
+    : mConvergenceState{AConvergenceState}
+    {
+        mModel.Reset();
+        mpModelPart = &mModel.CreateModelPart("TestModelPart");
+        mpModelPart->Reset();
+    }
+
+    [[nodiscard]] TimeStepEndState::ConvergenceState GetConvergenceState() override
```
avdg81: I'm wandering why this member function is not `const`? (If we agree it should be, then we clearly need to change it in the interface class as well.)

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    [[nodiscard]] std::size_t GetStepNumber() const override
+    {
+        return static_cast<std::size_t>(mpModelPart->GetProcessInfo()[STEP]);
+    }
+
+    void IncrementStepNumber() override
+    {
+        mpModelPart->GetProcessInfo()[STEP] += 1;
```
avdg81: Blaming myself here. I think we'd better use the pre-increment operator.
```suggestion
        ++mpModelPart->GetProcessInfo()[STEP];
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    [[nodiscard]] bool IsSaveFieldCalled() const
+    {
+        return mIsSaveTotalDisplacementFieldCalled;
+    }
+
+    void AccumulateTotalDisplacementField() override
+    {
+        mCountAccumulateTotalDisplacementFieldCalled += 1;
```
avdg81: I prefer to use `++` (`+= 1` is more the Python way of doing things):
```suggestion
        ++mCountAccumulateTotalDisplacementFieldCalled;
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    [[nodiscard]] std::size_t GetCountAccumulateTotalDisplacementFieldCalled() const
+    {
+        return mCountAccumulateTotalDisplacementFieldCalled;
+    }
+
+    void OutputProcess() override
+    {
+        mCountOutputProcessCalled += 1;
```
avdg81:
```suggestion
        ++mCountOutputProcessCalled;
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
```diff
@@ -0,0 +1,356 @@
...
+        // intentionally empty
+    }
+    void InitializeSolutionStep() override {
+        // intentionally empty
+    }
+    void Predict()                override {
+        // intentionally empty
+    }
+    bool SolveSolutionStep()      override { return false; }
+    void FinalizeSolutionStep()   override {mCountFinalizeSolutionStepCalled += 1;}
```
avdg81:
```suggestion
    void FinalizeSolutionStep()   override {++mCountFinalizeSolutionStepCalled;}
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
```diff
@@ -0,0 +1,356 @@
...
+KRATOS_TEST_CASE_IN_SUITE(TimeLoopReturnsPerformedStatesAfterRunningAnAlwaysConvergingSolverStrategy, KratosGeoMechanicsFastSuite)
+{
+    TimeLoopExecutor executor;
+    const auto increments = std::vector<double>{1.1, 1.2, 1.3, 1.4};
+    executor.SetTimeIncrementor(std::make_unique<PrescribedTimeIncrementor>(increments));
+    executor.SetSolverStrategyTimeStepExecutor(
+            std::make_unique<DummySolverStrategy>(TimeStepEndState::ConvergenceState::converged));
+    TimeStepEndState start_state;
+    start_state.convergence_state = TimeStepEndState::ConvergenceState::converged;
+    const auto step_states = executor.Run(start_state);
```
avdg81: We can simplify this to:
```suggestion
    const auto step_states = executor.Run(MakeConvergedStepState());
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+    KRATOS_EXPECT_EQ(increments.size(), step_states.size());
+    KRATOS_EXPECT_TRUE(std::all_of(step_states.begin(), step_states.end(), [](const auto& step_state) { return step_state.Converged(); }))
+}
+
+KRATOS_TEST_CASE_IN_SUITE(TimeLoopReturnsOneNonConvergedPerformedStatesAfterRunningAnNeverConvergingSolverStrategy, KratosGeoMechanicsFastSuite)
```
avdg81: Minor thing (use singular form when there is only one state):
```suggestion
KRATOS_TEST_CASE_IN_SUITE(TimeLoopReturnsOneNonConvergedPerformedStateAfterRunningAnNeverConvergingSolverStrategy, KratosGeoMechanicsFastSuite)
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
```diff
@@ -0,0 +1,356 @@
...
+{
+    TimeLoopExecutor executor;
+    const auto increments = std::vector<double>{1.1, 1.2, 1.3, 1.4};
+    executor.SetTimeIncrementor(std::make_unique<PrescribedTimeIncrementor>(increments));
+    executor.SetSolverStrategyTimeStepExecutor(
+            std::make_unique<DummySolverStrategy>(TimeStepEndState::ConvergenceState::non_converged));
+    TimeStepEndState start_state;
+    start_state.convergence_state = TimeStepEndState::ConvergenceState::converged;
+    const auto step_states = executor.Run(start_state);
```
avdg81:
```suggestion
    const auto step_states = executor.Run(MakeConvergedStepState());
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
```diff
@@ -0,0 +1,356 @@
...
+KRATOS_TEST_CASE_IN_SUITE(TimeLoopReturnsEndTimesAfterRunningAnAlwaysConvergingSolverStrategy, KratosGeoMechanicsFastSuite)
+{
+    TimeLoopExecutor executor;
+    const auto increments = std::vector<double>{1.1, 1.2, 1.3, 1.4};
+    executor.SetTimeIncrementor(std::make_unique<PrescribedTimeIncrementor>(increments));
+    executor.SetSolverStrategyTimeStepExecutor(
+            std::make_unique<DummySolverStrategy>(TimeStepEndState::ConvergenceState::converged));
+    TimeStepEndState start_state;
+    start_state.convergence_state = TimeStepEndState::ConvergenceState::converged;
+    const auto step_states = executor.Run(start_state);
```
avdg81:
```suggestion
    TimeStepEndState start_state;
    start_state.convergence_state = TimeStepEndState::ConvergenceState::converged;
    const auto step_states = executor.Run(MakeConvergedStepState());
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+KRATOS_TEST_CASE_IN_SUITE(ExpectOutputProcessCalledForEveryStep, KratosGeoMechanicsFastSuite)
+{
+    TimeLoopExecutor executor;
+    const auto wanted_num_of_cycles_per_step = 1;
+    executor.SetTimeIncrementor(std::make_unique<FixedCyclesTimeIncrementor>(wanted_num_of_cycles_per_step));
+    auto solver_strategy = std::make_shared<DummySolverStrategy>(TimeStepEndState::ConvergenceState::converged);
+    executor.SetSolverStrategyTimeStepExecutor(solver_strategy);
+    const auto step_states = executor.Run(TimeStepEndState{});
+    KRATOS_EXPECT_EQ(step_states.size(), solver_strategy->GetCountOutputProcessCalled());
```
avdg81, line 336: I would add a check for the number of expected step states. Suppose that for whatever reason it no longer returns two states, but only one, and that consequently the output generation process is only called once, this test would not fail. Something along the lines of:
```
KRATOS_EXPECT_EQ(2, step_states.size());
```

avdg81 (commented):
Clear and clean code. Good job! I have several comments, most of which are very minor.

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
WPK4FEM: Done

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
WPK4FEM: As this only aims at being able to construct total displacements accumulating over stages, we do not need to save the water pressures here.
The name change I do here, but these will have impact on the work that Richard is doing.

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
WPK4FEM: I do not expect effects of it so done.

applications/GeoMechanicsApplication/custom_workflows/time_step_executor.cpp
WPK4FEM: Done.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
WPK4FEM: Apparently so.
Done.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_stepping.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
WPK4FEM: I think it should be.
Done.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
WPK4FEM: Done, but then also the A Never i.s.o An Never

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
WPK4FEM: Done ( with removing the start_state too.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
WPK4FEM: Done ( the indentation )

rfaasse (approved):
Thanks for implementing the changes and the collaboration on this one! It turned out nicely, so from my side, it's ready to be merged

PR #11678, opened 2023-10-12
[Core] Transition changes geometrical object bins, add `PointIsInsideBoundingBox` functionality

loumalouomega:
**📝 Description**

This PR introduces new functionality to the `GeometricalObjectsBins` class. Here are the key changes made:

1. Added a new function `bool GeometricalObjectsBins::PointIsInsideBoundingBox(const array_1d<double,3>& rCoords)` that checks if a point is inside the bounding box of geometrical objects.
2. Added another function `bool GeometricalObjectsBins::PointIsInsideBoundingBoxWithTolerance(const array_1d<double,3>& rCoords)` that checks if a point is inside the bounding box with tolerance.
3. Modified the `GeometricalObjectsBins::CalculateCellSize` function to include additional logic.
4. Updated test cases in `test_geometrical_objects_bins.cpp`.

These changes seem to be aimed at improving the functionality of the `GeometricalObjectsBins` class by adding methods to check if a point lies within the bounding box of geometrical objects, with and without tolerance. The changes also include updates to test cases to ensure the new functionality works as expected.

**🆕 Changelog**

- [Transition changes geometrical object bins](https://github.com/KratosMultiphysics/Kratos/commit/a0ae26f9f486e90ae8887cdd7d19220687a96a86)
- [Update test](https://github.com/KratosMultiphysics/Kratos/commit/962b5ffc3c4dc7b5be8cf9d62bcec8dd6a9c024c)

### DIFF
--- a/kratos/spatial_containers/geometrical_objects_bins.h
+++ b/kratos/spatial_containers/geometrical_objects_bins.h
@@ -79,8 +79,8 @@ public:
             for (TIteratorType i_object = GeometricalObjectsBegin ; i_object != GeometricalObjectsEnd ; i_object++){
                 mBoundingBox.Extend(i_object->GetGeometry().begin() , i_object->GetGeometry().end());
             }
+            mBoundingBox.Extend(Tolerance);
         }
-        mBoundingBox.Extend(mTolerance);
         CalculateCellSize(number_of_objects);
         mCells.resize(GetTotalNumberOfCells());
         AddObjectsToCells(GeometricalObjectsBegin, GeometricalObjectsEnd);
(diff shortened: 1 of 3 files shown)


kratos/spatial_containers/geometrical_objects_bins.h
```diff
@@ -79,8 +79,8 @@ class KRATOS_API(KRATOS_CORE) GeometricalObjectsBins
             for (TIteratorType i_object = GeometricalObjectsBegin ; i_object != GeometricalObjectsEnd ; i_object++){
                 mBoundingBox.Extend(i_object->GetGeometry().begin() , i_object->GetGeometry().end());
             }
+            mBoundingBox.Extend(Tolerance);
```
roigcarlo, line 82: Shouldn't the default tolerance of this function be `std::numeric_limits<double>::epsilon()` instead of `1e-12` then?

Edit: The function on which this line is being called, github didnt let me ping the header.

Aside from that all ok. I approve because its just a question.

roigcarlo (approved):

kratos/spatial_containers/geometrical_objects_bins.h
loumalouomega: It is the current tolerance, better don't touch to avoid break things

roigcarlo: :+1:

PR #11680, opened 2023-10-13
[GeoMechanicsApplication] Moved logic of ApplyExcavationProcess to C++

### LINKED ISSUE #11672: [GeoMechanicsApplication] Migrating the ApplyExcavationProcess from Python to C++
As a developer in the Kratos core, I would like to create and instantiate excavation processes from the project parameters, such that I can use them in the Settlement Workflow.

**Background**

In order to be able to use ApplyExcavationProcess in the dgeosettlement.cpp for the Settlement workflow, the logic in the python should be migrated to C++.

**Acceptance Criteria**

**Given** the developer has excavation processes defined in the ProjectParameters.json file
**When** the developer runs the settlement stage (using the KratosGeoSettlement::RunStage function)
**Then** the processes associated with excavation in the settlement workflow are instantiated and ready for further use in the workflow.

indigocoral:
**📝 Description**
This class used to be defined partialy in Python module and in C++. In order to be able to use it in the settlement workflow, the logic is migrated from Python to C++.

🆕 Changelog

- The "ApplyExcavationProcess" class is migrated from Python to C++
- The logic existing in the "apply_excavation_process.hpp" file is split into .cpp and .hpp files

### DIFF
--- a/applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.cpp
+++ b/applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.cpp
@@ -1,0 +1,61 @@
+// KRATOS___
+//     //   ) )
+//    //         ___      ___
+//   //  ____  //___) ) //   ) )
+//  //    / / //       //   / /
+// ((____/ / ((____   ((___/ /  MECHANICS
+//
+//  License:         geo_mechanics_application/license.txt
+//
+//  Main authors:    Lorenzo Gracia,
+//                   Aron Noordam,
+//                   Vahid Galavi
+//                   Marjan Fathian
+//
+#include "apply_excavation_process.hpp"
+#include "includes/model_part.h"
+#include "includes/element.h"
+#include "utilities/variable_utils.h"
+#include "includes/kratos_flags.h"
+#include "includes/kratos_parameters.h"
+
+namespace Kratos
+{
+
+ApplyExcavationProcess::ApplyExcavationProcess(ModelPart&        rModelPart,
+                                               const Parameters& rProcessSettings)
+           : Process(Flags()),
+             mrModelPart{rModelPart}
+{
+    Parameters params = Parameters("{}");
+    params.AddValue("model_part_name", rProcessSettings["model_part_name"]);
+    params.AddValue("variable_name", rProcessSettings["variable_name"]);
+
+    if (rProcessSettings.Has("deactivate_soil_part")) {
+        params.AddValue("deactivate_soil_part", rProcessSettings["deactivate_soil_part"]);
+        KRATOS_TRY
+        mDeactivateSoilPart = rProcessSettings["deactivate_soil_part"].GetBool();
+        KRATOS_CATCH("")
+    }
+}
+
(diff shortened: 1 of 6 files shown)


applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.hpp
avdg81, line 1: Now that you have split the implementation from the interface, I would suggest to change the file extension from `.hpp` to `.h`. Recently, Richard and I have discovered that one of the static code analyzers (I think it was Codacy, but I'm not sure) assumes that `.hpp` files are files that contain the interface plus the implementation (typically used for class templates and function templates). To silence any warnings that stem from this assumption, please change the file extension as suggested.

applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.hpp
```diff
@@ -10,61 +10,32 @@
...
-class ApplyExcavationProcess : public Process
+
+class KRATOS_API(GEO_MECHANICS_APPLICATION) ApplyExcavationProcess : public Process
 {
-  public:
+public:
     KRATOS_CLASS_POINTER_DEFINITION(ApplyExcavationProcess);
 
+    ApplyExcavationProcess(ModelPart&        rModelPart,
+                           const Parameters& rSettings);
```
avdg81: Since you have updated the signature of the constructor, please also update it in file `add_custom_processes_to_python.cpp` (add `const` to the `Parameters&` parameter). Search for `ApplyExcavationProcess` in that file.

applications/GeoMechanicsApplication/custom_processes/apply_excavation_process.cpp
```diff
@@ -0,0 +1,61 @@
...
+    Parameters params = Parameters("{}");
+    params.AddValue("model_part_name", rProcessSettings["model_part_name"]);
+    params.AddValue("variable_name", rProcessSettings["variable_name"]);
+
+    if (rProcessSettings.Has("deactivate_soil_part")) {
+        params.AddValue("deactivate_soil_part", rProcessSettings["deactivate_soil_part"]);
+        KRATOS_TRY
+        mDeactivateSoilPart = rProcessSettings["deactivate_soil_part"].GetBool();
+        KRATOS_CATCH("")
+    }
```
avdg81: I feel that this part of the _original Python code_ was very misleading. The reason is that when "deactivate_soil_part" is not found in the settings, it did not create an internal process (i.e. `self.process` would not be defined). However, the other methods of the class assumed it was there. In other words, this class becomes useless when "deactivate_soil_part" is not found and it should result in an error immediately.

Furthermore, I noticed that `params` is built up, but it is never actually used. So we can remove that.

The only thing that is left is to initialize the data member `mDeactivateSoilPart`, which can be done using the class's initializer list. As a result, the body of the constructor becomes empty. Nice and clean :-)

avdg81 (commented):
This looks nice and clean to me. I only have a few suggestions to make the code even better. Well done Marjan!

avdg81 (approved):
Feel free to merge this one, Marjan!

PR #11684, opened 2023-10-16
[MedApp] more geom types and other minor improvements

philbucher:
basically support for more geometry types including lots of testing

Hexa20 & Hexa27 are still missing


matekelemen:
how about tracking the mesh files with `git lfs` instead of `git` directly?
@roigcarlo ?

applications/MedApplication/tests/test_med_model_part_io.py
```diff
@@ -4,38 +4,31 @@
...
 
         with self.subTest("check_model_part"):
             check_fct(self.mp_read_1)
 
         with self.subTest("read_write_read"):
             med_temp_path = GetMedPath(med_path, "temp.med")
-            DeleteFileIfExisting(med_temp_path) # make sure there are no leftovers from previous tests
-            self.addCleanup(DeleteFileIfExisting, med_temp_path) # clean up after test
+            DeleteFileIfExisting(med_temp_path)  # make sure there are no leftovers from previous tests
+            self.addCleanup(DeleteFileIfExisting, med_temp_path)  # clean up after test
```
matekelemen, line 31: why not put it in `tearDown`? Idk, which one's better in this case, just asking why you chose `addCleanup`.

applications/MedApplication/tests/test_med_model_part_io.py
```diff
@@ -110,13 +101,174 @@ def test_quadrilateral_4N_linear_mesh(self):
...
+            self.assertAlmostEqual(KratosMed.MedTestingUtilities.ComputeArea(model_part), 340000)
+            self.assertAlmostEqual(KratosMed.MedTestingUtilities.ComputeVolume(model_part), 10000000)
+            self.assertAlmostEqual(KratosMed.MedTestingUtilities.ComputeDomainSize(model_part), 10343200)
+
+            for node in model_part.Nodes:
+                self.assertTrue(0.0 <= node.X <= 500.0)
+                self.assertTrue(0.0 <= node.X0 <= 500.0)
+
+                self.assertTrue(0.0 <= node.Y <= 100.0)
+                self.assertTrue(0.0 <= node.Y0 <= 100.0)
```
matekelemen, line 123: oh no python, why do chained comparison operators work like this? :'(

applications/MedApplication/tests/test_med_model_part_io.py
```diff
@@ -110,13 +101,174 @@ def test_quadrilateral_4N_linear_mesh(self):
...
+
+
+def write_vtk(model_part, name):
+    # using the modeler to create elements for visualization,
+    # until the vtk-output supports geometries directly
+    modeler_parameters = KM.Parameters(
+        """{
+        "elements_list" : [{
+            "model_part_name" : "read_1",
+            "element_name" : "Element2D3N;Element2D4N;Element3D4N;Element3D8N;Element3D10N;Element3D20N;Element3D27N"
```
matekelemen, line 250: I know it's is unrelated to this PR, but why isn't this a list of strings instead of a `;` delimited one??

applications/MedApplication/tests/test_med_model_part_io.py
philbucher: I like it :sweat_smile:

loumalouomega (approved):
Okay

applications/MedApplication/tests/test_med_model_part_io.py
philbucher: @rubenzorrilla

applications/MedApplication/tests/test_med_model_part_io.py
philbucher: if the test crashes mid-execution, then `tearDown` is not called. `addCleanup` on the other hand is still called, making sure that the file is removed. (Pretty sure it is like this, but its been years since I looked this up)

philbucher:
> how about tracking the mesh files with `git lfs` instead of `git` directly? @roigcarlo ?

oh boy, I think the size-of-repo battle is lost for Kratos :see_no_evil: 
I keep the meshes very small, and will not add many more, thus I think it is not worth the additional struggle

roigcarlo:
> > how about tracking the mesh files with `git lfs` instead of `git` directly? @roigcarlo ?
> 
> oh boy, I think the size-of-repo battle is lost for Kratos 🙈 I keep the meshes very small, and will not add many more, thus I think it is not worth the additional struggle

Not against using lfs, but for the existing repo that would require a history rewrite. I've been simulating the process in my private copy, but it still gives problems. I promise one day we will do it.

Also lfs is only 5GB of sotrage so we cannot go wild.

loumalouomega:
> Also lfs is only 5GB of sotrage so we cannot go wild.

shrek_movie.mdpa

PR #11686, opened 2023-10-17
[Core] Clean up `mesh.h` file

loumalouomega:
**📝 Description**

Minor clean up of `mesh.h` file, update header. Replace `typedef` by `using`, among other minor clean ups.

**🆕 Changelog**

- [Clean up mesh.h file](https://github.com/KratosMultiphysics/Kratos/commit/0ece4c35c891c981c7bdcd124bed2c0c2139af1f)

### DIFF
--- a/kratos/includes/mesh.h
+++ b/kratos/includes/mesh.h
@@ -4,18 +4,14 @@
 //   _|\_\_|  \__,_|\__|\___/ ____/
 //                   Multi-Physics
 //
-//  License:		 BSD License
-//					 Kratos default license: kratos/license.txt
+//  License:         BSD License
+//                   Kratos default license: kratos/license.txt
 //
 //  Main authors:    Pooyan Dadvand
 //
 //
 
-
-#if !defined(KRATOS_MESH_H_INCLUDED )
-#define  KRATOS_MESH_H_INCLUDED
-
-
+#pragma once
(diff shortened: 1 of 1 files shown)


kratos/includes/mesh.h
```diff
@@ -60,143 +53,116 @@ namespace Kratos
...
-
-    typedef Mesh<TNodeType, TPropertiesType, TElementType, TConditionType> MeshType;
+    // Alias for representing indices, typically used for indexing arrays or collections.
+    using IndexType = std::size_t;
+
+    // Alias for representing sizes or counts, commonly used to indicate the size of data structures.
+    using SizeType = std::size_t;
+
+    // Alias for representing node types in a mesh data structure.
+    using NodeType = TNodeType;
```
roigcarlo, line 84: We should delete this one an use Node directly.

roigcarlo (approved):
Ok, but we should consider starting to remove the `NodeType` from Kratos

kratos/includes/mesh.h
loumalouomega: It is a template argument, IMHO we could remove all those templates or directly inegrate the mesh into the model part

roigcarlo: Agree

PR #11687, opened 2023-10-17
[Core] AMGCL NS - Allow using different var for pressure

ddiezrod:
**📝 Description**
AMGCL_NS already allows using any variable for pressure (not just PRESSURE) from parameters. However, this setting is not being used later. Here I make sure that when pmask is built, the correct variable is used.
FYI @pablobecker @pooyan-dadvand

### DIFF
--- a/kratos/linear_solvers/amgcl_ns_solver.h
+++ b/kratos/linear_solvers/amgcl_ns_solver.h
@@ -28,6 +28,7 @@
 
 // Project includes
 #include "includes/define.h"
+#include "includes/kratos_components.h"
 #include "linear_solvers/iterative_solver.h"
 
 #include <boost/property_tree/json_parser.hpp>
@@ -118,7 +119,8 @@ public:
         //kratos specific settings
         mTol = rParameters["inner_settings"]["solver"]["tol"].GetDouble();
         mVerbosity=rParameters["verbosity"].GetInt();
-        mPressureVarName = rParameters["schur_variable"].GetString();
+        const std::string pressure_var_name = rParameters["schur_variable"].GetString();
+        mpPressureVar = &KratosComponents<Variable<double>>::Get(pressure_var_name);
         mndof = 1; 
 
         //pass settings to amgcl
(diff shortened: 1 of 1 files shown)


ddemidov (dismissed):
Looks good to me

kratos/linear_solvers/amgcl_ns_solver.h
```diff
@@ -419,6 +421,7 @@ class AMGCL_NS_Solver : public LinearSolver< TSparseSpaceType,
 
     double mTol;
     int mVerbosity;
+    std::string mPressureVarName;
```
loumalouomega: I would use a variable pointer, avoids to retreive the variable each time and is less memory than a string

loumalouomega (approved):

kratos/linear_solvers/amgcl_ns_solver.h
ddiezrod: done

kratos/linear_solvers/amgcl_ns_solver.h
```diff
@@ -408,7 +411,7 @@ class AMGCL_NS_Solver : public LinearSolver< TSparseSpaceType,
             const unsigned int eq_id = it->EquationId();
             if( eq_id < rA.size1() )
             {
-                mp[eq_id]  = (it->GetVariable().Key() == PRESSURE);
+                mp[eq_id]  = (it->GetVariable().Key() == *mpPressureVar);
```
roigcarlo, line 414: Could you use `it->GetVariable().Name()` instead of key and leave the setting as a string?. This way you don't need to access the components and you can use any variable type.


If you did this to force `Variable<double>`ignore the comment, I have not much experience with this part of Kratos.

ddiezrod: yes, this code only makes sense if the variable is `Variable<double>` as this variable has to be somehow equivalent to `PRESSURE` (so for example we use `SCALED_PRESSURE`) , so maybe we can leave it as this...

PR #11688, opened 2023-10-17
[Core] Mixed Generic Criterion - fixing error when dofs and input variables mismatch

ddiezrod:
**📝 Description**
There was a bug in our code that took some time to debug. We were sending the wrong variables to the constructor, so as the dofset included different variables, new entries were being added to the mLocalKeyMap as it was using the brackets operator (added new entry if it does not exist). This caused me errors in omp as two threads might try to create this new entry at the same time. Im skipping now the dofs that do not match any of the input variables


kratos/solving_strategies/convergencecriterias/mixed_generic_criteria.h
```diff
@@ -493,7 +493,8 @@ class MixedGenericCriteria : public ConvergenceCriteria< TSparseSpace, TDenseSpa
                     dof_dx = TSparseSpace::GetValue(rDx, dof_id);
 
                     const auto &r_current_variable = it_dof->GetVariable();
-                    int var_local_key = mLocalKeyMap[r_current_variable.IsComponent() ? r_current_variable.GetSourceVariable().Key() : r_current_variable.Key()];
+                    const KeyType key = r_current_variable.IsComponent() ? r_current_variable.GetSourceVariable().Key() : r_current_variable.Key();
+                    const int var_local_key = mLocalKeyMap.at(key);
```
philbucher: please use `find` with a corresponding proper error. My issue with `at` is that it only throws the cryptic standard lib error

philbucher (changes requested):
nice catch!

ddiezrod:
FYI @pooyan-dadvand  @pablobecker

kratos/solving_strategies/convergencecriterias/mixed_generic_criteria.h
ddiezrod: cool, that makes sense.

ddiezrod: changed, feel free to suggest another error message, Im very bad at those :)

kratos/solving_strategies/convergencecriterias/mixed_generic_criteria.h
```diff
@@ -493,7 +492,13 @@ class MixedGenericCriteria : public ConvergenceCriteria< TSparseSpace, TDenseSpa
                     dof_dx = TSparseSpace::GetValue(rDx, dof_id);
 
                     const auto &r_current_variable = it_dof->GetVariable();
-                    int var_local_key = mLocalKeyMap[r_current_variable.IsComponent() ? r_current_variable.GetSourceVariable().Key() : r_current_variable.Key()];
+                    const KeyType key = r_current_variable.IsComponent() ? r_current_variable.GetSourceVariable().Key() : r_current_variable.Key();
+                    auto key_find = mLocalKeyMap.find(key);
+                    if (key_find == mLocalKeyMap.end()) {
+                        KRATOS_ERROR << "MixedGenericCriteria: The key " << key << " was not found in mLocalKeyMap." <<
+                          "Make sure the input variables in constructor and the dofset variables match.\n";
+                    }
```
philbucher:
```suggestion
                    KRATOS_ERROR_IF(key_find == mLocalKeyMap.end()) << "MixedGenericCriteria: The key " << key << " was not found in mLocalKeyMap for variable " << r_current_variable.Name() << ". Make sure the input variables in constructor and the dofset variables match." << std::endl;
```
I would flush here (`std::endl` == `\n` + flush), since this anyway crashes the execution and thus is not performance critical

philbucher (dismissed):

ddiezrod:
Hmm there are some tests failing in @KratosMultiphysics/geomechanics  because in some cases they are only checking some variables (WATER_PRESSURE) even if they also have dofs for DISPLACEMENT. I think that is actually a fair use, so maybe instead of throwing an error we should just skip the dof if the variable was not added to `mLocalKeyMap`. @KratosMultiphysics/technical-committee

matekelemen:
Why is `GetNormValues` not `const`? I would not expect this function to mutate its state, so the error is the correct way to go in my opinion.

ddiezrod:
> Why is `GetNormValues` not `const`? I would not expect this function to mutate the class' state, so the error is the correct way to go in my opinion.

I agree that the function should be const. But I think its also a fair use that for example in the context of a CFD problem where you have VELOCITY and PRESSURE dofs, for whatever reason you only want to check VELOCITY (that'd throw an error with current implementation in this MR)

ddiezrod:
I just pushed my suggestion @philbucher @KratosMultiphysics/implementation-committee @KratosMultiphysics/technical-committee

philbucher:
> > Why is `GetNormValues` not `const`? I would not expect this function to mutate the class' state, so the error is the correct way to go in my opinion.
> 
> I agree that the function should be const. But I think its also a fair use that for example in the context of a CFD problem where you have VELOCITY and PRESSURE dofs, for whatever reason you only want to check VELOCITY (that'd throw an error with current implementation in this MR)

perhaps there should be a setting to explicitly ignore these dofs? I feel like just neglecting them silently can cause problems too

ddiezrod:
> perhaps there should be a setting to explicitly ignore these dofs? I feel like just neglecting them silently can cause problems too

Hmm honestly I do not see it. What do you want to do with dofs that do not belong to any of the input variables? You do not even have a target error for those, relative and absolute errors are only defined for input vars...

philbucher:
> > perhaps there should be a setting to explicitly ignore these dofs? I feel like just neglecting them silently can cause problems too
> 
> Hmm honestly I do not see it. What do you want to do with dofs that do not belong to any of the input variables? You do not even have a target error for those, relative and absolute errors are only defined for input vars...

then you can/should ignore those. Only a suggestion, IMO one should be fully aware of what dofs are used in a simulation, and how the convergence is checked

applications/TrilinosApplication/custom_strategies/convergencecriterias/trilinos_mixed_generic_criteria.h
```diff
@@ -183,8 +183,14 @@ class TrilinosMixedGenericCriteria : public MixedGenericCriteria< TSparseSpace,
...
-                KeyType var_key = r_current_variable.IsComponent() ? r_current_variable.GetSourceVariable().Key() : r_current_variable.Key();
-                int var_local_key = r_local_key_map[var_key];
+                const KeyType var_key = r_current_variable.IsComponent() ? r_current_variable.GetSourceVariable().Key() : r_current_variable.Key();
+                auto key_find = r_local_key_map.find(var_key);
+                if (key_find == r_local_key_map.end()) {
+                    // the dof does not belong to the list of variables
+                    // we are checking for convergence, so we skip it
+                    continue;
+                }
+                const int var_local_key = key_find->second;
```
philbucher: why not using the baseclass? Refactor to a function? :innocent:

ddiezrod: I put this in a function. Looking at the rest of the code I do no think there is much more you can do without overcomplicating it.

philbucher (dismissed):
nice!

ddiezrod:
It seems its passing now :) @philbucher

philbucher (dismissed):

kratos/solving_strategies/convergencecriterias/mixed_generic_criteria.h
```diff
@@ -405,6 +404,31 @@ class MixedGenericCriteria : public ConvergenceCriteria< TSparseSpace, TDenseSpa
         }
     }
 
+    /**
+     * @brief Finds the var local key in the mLocalKeyMap for 
+     * a gifen dof. If the variable does not exist in mLocalKeyMap
+     * this function returns false
+     * @param iDof the dof
```
rubenzorrilla:
```suggestion
     * @param itDof the DOF iterator
```

kratos/solving_strategies/convergencecriterias/mixed_generic_criteria.h
```diff
@@ -405,6 +404,31 @@ class MixedGenericCriteria : public ConvergenceCriteria< TSparseSpace, TDenseSpa
...
+     * @param rVarLocalKey variable local key 
+     * @return dof variable is found or not
+     */
+    bool FindVarLocalKey(
+        typename DofsArrayType::const_iterator iDof,
+        int& rVarLocalKey) const
+    {
+        const auto &r_current_variable = iDof->GetVariable();
```
rubenzorrilla:
```suggestion
        const auto &r_current_variable = itDof->GetVariable();
```

kratos/solving_strategies/convergencecriterias/mixed_generic_criteria.h
```diff
@@ -405,6 +404,31 @@ class MixedGenericCriteria : public ConvergenceCriteria< TSparseSpace, TDenseSpa
         }
     }
 
+    /**
+     * @brief Finds the var local key in the mLocalKeyMap for 
+     * a gifen dof. If the variable does not exist in mLocalKeyMap
```
rubenzorrilla:
```suggestion
     * a gifen DOF. If the variable does not exist in mLocalKeyMap
```

rubenzorrilla (dismissed):
Makes perfect sense for me. It is customary to have more DOFs than the ones you might want to actually check (think for instance in any volume coupling problem). My suggestions are purely aesthetic and typos in documentation.

ddiezrod:
comments addressed @rubenzorrilla  ;)

kratos/solving_strategies/convergencecriterias/mixed_generic_criteria.h
```diff
@@ -406,17 +406,17 @@ class MixedGenericCriteria : public ConvergenceCriteria< TSparseSpace, TDenseSpa
...
+     * a gifen DOF. If the variable does not exist in mLocalKeyMap
      * this function returns false
-     * @param iDof the dof
+     * @param itDof the DOF iterator
      * @param rVarLocalKey variable local key 
      * @return dof variable is found or not
      */
     bool FindVarLocalKey(
-        typename DofsArrayType::const_iterator iDof,
+        typename DofsArrayType::const_iterator itDof,
```
philbucher, line 416: how about passing the variable? The dof is only used to access the variable

ddiezrod: well I used the dof to minimize the duplicated code in the trilinos side, what is the advantage on passing the variable instead?

philbucher: More of a philosophical question :)

I try to always pass the least amount of information possible

philbucher (approved):

PR #11689, opened 2023-10-17
[Core] Some refactor in `MPISearchUtilities`

loumalouomega:
**📝 Description**

In the current implementation (which is not currently used, will be used in  a future PR), the points are checked if all the same directly checking its coordinates, this change just does a check when working with nodes and check that the node is local. 

**🆕 Changelog**

- [Some refactor in MPISearchUtilities](https://github.com/KratosMultiphysics/Kratos/commit/3d9f6178ddd8fd7d68ef759021b52798c6bf053d)
- [Update tests for MPISearchUtilities](https://github.com/KratosMultiphysics/Kratos/commit/63774855280170918acfde01b02dd4e4a7410047)

### DIFF
--- a/kratos/mpi/utilities/mpi_search_utilities.h
+++ b/kratos/mpi/utilities/mpi_search_utilities.h
@@ -20,6 +20,8 @@
 
 // Project includes
 #include "includes/data_communicator.h"
+#include "includes/variables.h"
+#include "includes/model_part.h"
 
 namespace Kratos
 {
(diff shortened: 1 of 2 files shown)


kratos/mpi/utilities/mpi_search_utilities.h
```diff
@@ -20,6 +20,8 @@
 
 // Project includes
 #include "includes/data_communicator.h"
+#include "includes/variables.h"
+#include "includes/model_part.h"
```
philbucher, line 24: dont think these are used

loumalouomega: INDEX_PARTITION + ModelPart::NodeIterator

PR #11692, opened 2023-10-17
[GeoMechanicsApplication] Fixed the model part used by the process creators for settlement

avdg81:
**📝 Description**
The name of the model part that is associated with a process, is kept by the process parameters. Look up the model part by name and pass that to the constructor of a process rather than the complete model part that is created by the settlement workflow.


rfaasse (approved):
Nice fix, no comments!

WPK4FEM (commented):
Looks consistent to me.

PR #11693, opened 2023-10-18
[GeoMechanicsApplication] Added reading missing parameters for the solving strategy

rfaasse:
This PR aims to add reading more parameters for the solving strategy from the parameters json format including: 
1. The move_mesh_flag
2. The reform_dofs_at_each_step flag
3. The compute_reactions flag
4. The max_iterations

### DIFF
--- a/applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
+++ b/applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
@@ -44,10 +44,10 @@ public:
         if (rSolverSettings[strategy_type].GetString() ==
             GeoMechanicsNewtonRaphsonStrategy<TSparseSpace, TDenseSpace, TLinearSolver>::Name())
         {
-            const int max_iterations = rSolverSettings["max_iterations"].GetInt();
-            const bool calculate_reactions = rSolverSettings["calculate_reactions"].GetBool();
-            const bool reform_dof_set_at_each_step = rSolverSettings["reform_dofs_at_each_step"].GetBool();
-            const bool move_mesh_flag = rSolverSettings["move_mesh_flag"].GetBool();
+            const auto max_iterations = rSolverSettings["max_iterations"].GetInt();
+            const auto calculate_reactions = rSolverSettings["calculate_reactions"].GetBool();
+            const auto reform_dof_set_at_each_step = rSolverSettings["reform_dofs_at_each_step"].GetBool();
+            const auto move_mesh_flag = rSolverSettings["move_mesh_flag"].GetBool();
 
             auto strategy_parameters = ExtractStrategyParameters(rSolverSettings);
             return std::make_unique<GeoMechanicsNewtonRaphsonStrategy<TSparseSpace,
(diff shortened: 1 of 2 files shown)


applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
```diff
@@ -44,6 +44,11 @@ class SolvingStrategyFactory
         if (rSolverSettings[strategy_type].GetString() ==
             GeoMechanicsNewtonRaphsonStrategy<TSparseSpace, TDenseSpace, TLinearSolver>::Name())
         {
+            const int max_iterations = rSolverSettings["max_iterations"].GetInt();
```
rfaasse: @avdg81 Do you think we need to check this with 'Has' or is it fine if it throws when it doesn't have these parameters

applications/GeoMechanicsApplication/tests/cpp_tests/test_solving_strategy_factory.cpp
```diff
@@ -111,6 +111,7 @@ KRATOS_TEST_CASE_IN_SUITE(Create_ReturnsSolvingStrategy_ForNewtonRhapsonStrategy
     const auto created_strategy = SolvingStrategyFactoryType::Create(
             parameters, dummy_model_part);
 
+    KRATOS_EXPECT_TRUE(created_strategy->GetMoveMeshFlag()) // since it is set to true in the parameters
```
rfaasse, line 114: Unfortunately, this is the only flag we can check

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
```diff
@@ -44,6 +44,11 @@ class SolvingStrategyFactory
...
+            const bool calculate_reactions = rSolverSettings["calculate_reactions"].GetBool();
+            const bool reform_dof_set_at_each_step = rSolverSettings["reform_dofs_at_each_step"].GetBool();
+            const bool move_mesh_flag = rSolverSettings["move_mesh_flag"].GetBool();
```
avdg81: I would use `const auto` here for all cases, since the types are clear from the expressions that initialize the variables:
```suggestion
            const auto max_iterations = rSolverSettings["max_iterations"].GetInt();
            const auto calculate_reactions = rSolverSettings["calculate_reactions"].GetBool();
            const auto reform_dof_set_at_each_step = rSolverSettings["reform_dofs_at_each_step"].GetBool();
            const auto move_mesh_flag = rSolverSettings["move_mesh_flag"].GetBool();
```

applications/GeoMechanicsApplication/custom_utilities/solving_strategy_factory.hpp
avdg81: Since it seems that these settings are optional rather than required I agree with you that checking whether any particular setting exists is probably the better way to go. However, you could have any combination of optional settings (some may be present, others are not), which complicates things. In that case, we would need to know about the default values for the undefined settings, which is not straightforward. So, for now, I would suggest to keep it the way it is.

applications/GeoMechanicsApplication/tests/cpp_tests/test_solving_strategy_factory.cpp
avdg81: Then this is the best you can do now :-)

avdg81 (commented):
Looks good to me. I only have one minor suggestion and two replies to your own comments.

avdg81 (approved):
Nice improvement.

PR #11696, opened 2023-10-18
Geo/tests strain measures

### LINKED ISSUE #11642: [GeoMechanicsApplication] Validate "updated_mesh" parameter
Related issue: [#11643](https://github.com/KratosMultiphysics/Kratos/issues/11643)

Validate "updated_mesh" parameter for NEN-Bjerrum soil model and investigate the effect of it on strains.
Possibly the term is "move_mesh_flag" in the project parameters.

After getting more information from Raymond van der Meij:
The Plaxis jargon updated_mesh refers to geometrical nonlinear implementation of elements. In Kratos GeoMechanics the UpdatedLagrangian elements are available for that. 
The move_mesh_flag in ProjectParameters.jsonhas something to do with what we need here, According to a communication from Vahid to Jonathan and Carlos it should be used together with USE_HENCKY_STRAIN in MaterialParameters.json. Both should be false or both should be true.

WPK4FEM:
Tests for linear and Hencky strain measures for small strain and Updated Lagrangian elements and the different order variants of these UPw element classes ( using the quadratic quadrilaterals ).


applications/GeoMechanicsApplication/tests/test_strain_measures.py
```diff
@@ -0,0 +1,120 @@
+import os
+
+import KratosMultiphysics.KratosUnittest as KratosUnittest
+import test_helper
+
+class KratosGeoMechanicsStrainMeasureTests(KratosUnittest.TestCase):
+    """
+    This class contains tests which check the displacement result of a column with linear elastic and linear or Hencky ( = natural = logarithmic ) strains
```
indigocoral, line 8: Nice work; just a small suggestion. In my opinion, at the beginning of the test, the analytical solution can be better described as follows:

```suggestion
    This class contains tests which check the displacement result of a column with linear elastic and linear or Hencky ( = natural = logarithmic ) strains
The analytical solution for these tests is:

For linear strain: ∆H = H0 * (σ / E)
For logarithmic strain: ∆H = H0 * (exp (σ / E) -1)

∆H : top total displacement (m)
H_0 : Original height of column (m)
σ : Compressive stress (N/m2)
E : Young's Modulus (N/m2)'
```

applications/GeoMechanicsApplication/tests/test_strain_measures/diff_order_elements/column_small_deformation_logarithmic_strain/ProjectParameters.json
```diff
@@ -0,0 +1,159 @@
...
+        "convergence_criterion":              "displacement_criterion",
+        "displacement_relative_tolerance":    1.0E-4,
+        "displacement_absolute_tolerance":    1.0E-9,
+        "residual_relative_tolerance":        1.0E-4,
+        "residual_absolute_tolerance":        1.0E-9,
+        "water_pressure_relative_tolerance":  1.0E-4,
+        "water_pressure_absolute_tolerance":  1.0E-9,
+        "min_iterations":                     6,
+        "max_iterations":                     15,
+        "number_cycles":                      1,
```
indigocoral, line 51: This would be more of a question from my side. I see changes in the "number_cycles", "reduction_factor" and "increase_factor" parameters in different tests. It probably doesn't matter, but I will discuss it with you tomorrow :)

applications/GeoMechanicsApplication/tests/test_strain_measures/same_order_elements/column_small_deformation_logarithmic_strain/README.md
indigocoral, line 1: Thanks for providing the README.md files. My comment on the test file regarding the description of the analytical solution can also be added to these README.md files.

indigocoral (commented):
Thank you Wijtze Pieter for making these tests.

applications/GeoMechanicsApplication/tests/test_strain_measures.py
WPK4FEM: Very good suggestion, I have committed it.

applications/GeoMechanicsApplication/tests/test_strain_measures/diff_order_elements/column_small_deformation_logarithmic_strain/ProjectParameters.json
WPK4FEM: The changes to number_cycles, reduction_factor and increase_factor are for the following reasons:
1 bringing number_cycles down reduced waiting time when I was not doing correct things yet. Otherwise I would have waited for more cycles not bringing a solution. Now that I have correct and converging results, those happen in 1 cycle, so more are not necessary.
2 that changes to the reduction_factor and increase_factor are such that the time stepping scheme just does equal size steps.

applications/GeoMechanicsApplication/tests/test_strain_measures/same_order_elements/column_small_deformation_logarithmic_strain/README.md
WPK4FEM: After discussing, we left them in both.

applications/GeoMechanicsApplication/tests/test_strain_measures.py
```diff
@@ -0,0 +1,129 @@
...
+The analytical solution for these tests is:
+
+For linear strain: ∆H = H0 * (q / E)
+For logarithmic strain: ∆H = H0 * (exp (q / E) -1)
```
mcgicjn2, line 12: Checked and confirmed

mcgicjn2 (approved):
@WPK4FEM Are we satisfied that this will also work for 2D6N triangular elements and the other various versions of elements, as the testing is only on 2D8N. I suspect its OK.

Also as this is a linear elastic test can we not possibly automate the test further using random youngs modulus and then provide analytical values for the calculations. Just a discussion point.

applications/GeoMechanicsApplication/tests/test_strain_measures.py
WPK4FEM: Thank you.

PR #11708, opened 2023-10-22
[MedApp] add python registry

philbucher:
Adding the python registry lists for the MedApp

@rubenzorrilla I am not fully sure what this does, my imports in the projParams also worked without this :thinking:

### DIFF
--- a/applications/MedApplication/python_registry_lists.py
+++ b/applications/MedApplication/python_registry_lists.py
@@ -1,0 +1,11 @@
+from typing import List
+
+python_modelers_to_be_registered: List[str] = ["modelers.import_med_modeler.ImportMedModeler"]
+
+python_operations_to_be_registered: List[str] = []
+
+python_processes_to_be_registered: List[str] = []
+
+python_stages_to_be_registered: List[str] = []
+
+python_orchestrators_to_be_registered: List[str] = []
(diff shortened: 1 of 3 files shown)


applications/MedApplication/python_registry_lists.py
```diff
@@ -0,0 +1,11 @@
+from typing import List
+
+python_modelers_to_be_registered: List[str] = ["modelers.import_med_modeler.ImportMedModeler"]
```
rubenzorrilla, line 3:
```suggestion
python_modelers_to_be_registered: List[str] = ["import_med_modeler.ImportMedModeler"]
```
As it is in the modelers list it will automatically add the corresponding `modeler` keyword. Long story short, in the list it is only required to add `python_module.ClassName`.

rubenzorrilla:
> Adding the python registry lists for the MedApp
> 
> @rubenzorrilla I am not fully sure what this does, my imports in the projParams also worked without this 🤔

Basically it will add all the Python entries in lists to the registry at the time the application is imported. Most probably you didn't notice any change as you're not using a registry-based factory.

applications/MedApplication/python_registry_lists.py
philbucher: I copied it from the core where it is done like I do
https://github.com/KratosMultiphysics/Kratos/blob/5bf5e40b7fc6e061e5e2f3056efda3ed3b59d963/kratos/python_interface/python_registry_lists.py#L3

I prefer to keep it consistent

rubenzorrilla: The point is that in the core we have a submodule `modeler` coming from the corresponding subfolder within `python_scripts`. I didn't note that this is also the case in the `MedApplication` so you're right.  Forget about my comment (I'd leave it as unresolved as "documentation").

rubenzorrilla (approved):

PR #11711, opened 2023-10-23
[GeoMechanicsApplication]  Move normal load logic to cpp

### LINKED ISSUE #13036: [GeoMechanicsApplication] Migrating the ApplyNormalLoadTableProcess From Python to C++
As a developer in the Kratos core, I would like to create and instantiate normal load processes from the project parameters, such that I can use them in the Settlement Workflow.

Background

In order to be able to use ApplyNormalLoadTableProcess in the dgeosettlement.cpp for the Settlement workflow, the logic in the python should be migrated to C++.

Acceptance Criteria

Given the developer has excavation processes defined in the ProjectParameters.json file
When the developer runs the settlement stage (using the KratosGeoSettlement::RunStage function)
Then the processes associated with normal load in the settlement workflow are instantiated and ready for further use in the workflow.

Given the developer/user has processes defined which are not in the list of expected processes.
When the developer/user runs the settlement stage (using the KratosGeoSettlement::RunStage function)
Then these non-defined processes raise an error message "Unknown Process: (name of the process), the calculation is then aborted" and stops.

**We forgot to create a new issue on GitHub, and we believe there is no point in creating one afterwards. Therefore, we include the link to the merged PR for future reference:** https://github.com/KratosMultiphysics/Kratos/pull/11711

markelov208:
**📝 Description**
- moved python class apply_normal_load_table_process to cpp

**🆕 Changelog**

- created cpp class ApplyNormalLoadTableProcess
- cleaned python script apply_normal_load_table_process .py
- moved free functions from apply_scalar_constraint_table_process.cpp to a new class FunctionsForParameters
- added ApplyNormalLoadTableProcess in dgeosettlement
- added ApplyNormalLoadTableProcess in add_custom_process_to_python


applications/GeoMechanicsApplication/custom_utilities/functions_for_parameters.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+    Parameters FunctonsForParameters::ExtractParameters(const Parameters& rSourceParameters,
+                                                        const std::vector<std::string>& rNamesOfParametersToCopy)
+    {
+        auto result = Parameters{};
+        result.CopyValuesFromExistingParameters(rSourceParameters, rNamesOfParametersToCopy);
+        return result;
+    }
+
+    void FunctonsForParameters::AppendParameterNameIfExists(const std::string& rParameterName,
```
rfaasse: Minor comment about the styling: The functions in this file don't need to be indented here, inside namespaces we don't use indentation

applications/GeoMechanicsApplication/custom_utilities/functions_for_parameters.h
```diff
@@ -0,0 +1,37 @@
...
+            static void AppendParameterNameIfExists(const std::string& rParameterName,
+                const Parameters& rSourceParameters,
+                std::vector<std::string>& rResult);
+
+            static bool HasTableAttached(const Parameters& rSettings);
+
+            static bool HasTableAttached(const Parameters& rSettings, int component);
+    };
+
+}
```
rfaasse: Also here, no indentation is needed after namespace (only for functions within the class itself)

rfaasse (dismissed):
Nice clean and readable code! I only have 2 minor style related comments.

applications/GeoMechanicsApplication/custom_utilities/functions_for_parameters.h
avdg81, line 1: To be consistent with the other files in the `custom_utilities` directory, I would suggest to rename this file to `parameters_utilities.h`. What do you think?

applications/GeoMechanicsApplication/custom_utilities/functions_for_parameters.h
```diff
@@ -0,0 +1,37 @@
...
+//                   Gennady Markelov
+//
+
+#pragma once
+
+#include "includes/kratos_parameters.h"
+
+namespace Kratos {
+
+    class FunctonsForParameters
```
avdg81: And perhaps rename the class to `ParametersUtilities`? Otherwise, I would at least correct a typo: `FunctionsForParameters` (note the missing `i` in `Functons`).

applications/GeoMechanicsApplication/custom_processes/apply_normal_load_table_process.cpp
```diff
@@ -0,0 +1,195 @@
...
+        KRATOS_ERROR << "Unknown fluid_pressure_type: " << fluid_pressure_type << std::endl;
+    }
+}
+
+void ApplyNormalLoadTableProcess::MakeProcessForUniformFluidPressureType(const Parameters& rProcessSettings,
+                                                                         std::vector<std::string> NamesOfSettingsToCopy)
+{  
+    auto normal_parameters = FunctonsForParameters::ExtractParameters(rProcessSettings,
+                                                                      NamesOfSettingsToCopy);
+    normal_parameters.AddValue("value", rProcessSettings["value"][0]);
```
avdg81: I was just wandering whether the index here (`0`) also refers to the normal component of `value`...? If yes, then we might consider to define that index in a wider scope (e.g. as a class constant) and then replace those indices by the new constant. You have done that already in member functions `IsNormalComponentActive` and `IsTangentialComponentActive`, but perhaps we can use it here as well. What do you think?

applications/GeoMechanicsApplication/custom_processes/apply_normal_load_table_process.cpp
```diff
@@ -0,0 +1,195 @@
...
+void ApplyNormalLoadTableProcess::MakeProcessForPhreaticLineFluidPressureType(const Parameters& rProcessSettings,
+                                                                              std::vector<std::string> NamesOfSettingsToCopy)
+{
+    NamesOfSettingsToCopy.insert(NamesOfSettingsToCopy.end(), { "gravity_direction",
+                                                               "out_of_plane_direction",
+                                                               "first_reference_coordinate",
+                                                               "second_reference_coordinate",
+                                                               "specific_weight"});
+    if (FunctonsForParameters::HasTableAttached(rProcessSettings)) {
+        NamesOfSettingsToCopy.insert(NamesOfSettingsToCopy.end(), { "table" });
```
avdg81: Although correct, I would simplify this to:
```suggestion
        NamesOfSettingsToCopy.emplace_back("table");
```

applications/GeoMechanicsApplication/custom_processes/apply_normal_load_table_process.cpp
```diff
@@ -0,0 +1,195 @@
...
+void ApplyNormalLoadTableProcess::MakeProcessForPhreaticSurfaceFluidPressureType(const Parameters& rProcessSettings,
+                                                                                 std::vector<std::string> NamesOfSettingsToCopy) 
+{
+    NamesOfSettingsToCopy.insert(NamesOfSettingsToCopy.end(), { "gravity_direction",
+                                                                "first_reference_coordinate",
+                                                                "second_reference_coordinate",
+                                                                "third_reference_coordinate",
+                                                                "specific_weight" });
+    if (FunctonsForParameters::HasTableAttached(rProcessSettings)) {
+        NamesOfSettingsToCopy.insert(NamesOfSettingsToCopy.end(), { "table" });
```
avdg81:
```suggestion
        NamesOfSettingsToCopy.emplace_back("table");
```

applications/GeoMechanicsApplication/custom_processes/apply_normal_load_table_process.h
```diff
@@ -0,0 +1,66 @@
...
+//                   Anne van de Graaf,
+//                   Gennady Markelov
+//
+
+#pragma once
+
+#include "processes/process.h"
+
+//#include <memory>
+//#include <vector>
```
avdg81: Although strictly speaking these `#include`s could be left out (they are indirectly included through `processes/process.h`), I would insist to actually include them, since the class declares a data member (`mProcesses`) that requires them. We don't want to rely on accidental includes from other includes.

applications/GeoMechanicsApplication/custom_processes/apply_normal_load_table_process.h
```diff
@@ -0,0 +1,66 @@
...
+                                const Parameters& rProcessSettings);
+
+    ~ApplyNormalLoadTableProcess() override;
+
+    ApplyNormalLoadTableProcess(const ApplyNormalLoadTableProcess&) = delete;
+    ApplyNormalLoadTableProcess& operator=(const ApplyNormalLoadTableProcess&) = delete;
+
+    void ExecuteInitialize() override;
+    void ExecuteInitializeSolutionStep() override;
+
```
avdg81, line 45: Recently, we have added an override for member function `Info()` to class `ApplyScalarConstraintTableProcess`, which comes in really handy during debugging. I would suggest you to add a similar override here too.

applications/GeoMechanicsApplication/custom_utilities/functions_for_parameters.cpp
```diff
@@ -0,0 +1,62 @@
...
+        if (rSettings["table"][component].IsNumber()) {
+            return rSettings["table"][component].GetInt() != 0;
+        }
+
+        KRATOS_ERROR_IF_NOT(rSettings["table"][component].IsArray()) << "'table' is neither a single number nor an array of numbers";
+
+        const auto& table = rSettings["table"][component];
+        return std::any_of(table.begin(), table.end(),
+            [](const auto& value) {return value.GetInt() != 0; });
+    }
```
avdg81: I doubt whether this is correct. My understanding is that `table` either has a single number or an array of numbers. Now this new overload checks one particular element from the array variant of `table`. By inspecting the original Python code, I don't have the impression that we might have a `table` which is an array, where each element contains an array in turn (which would require two indices to access a number). Therefore, I believe we can simplify this function body to:
```suggestion
        return rSettings["table"][component].GetInt() != 0;
```
Could you please verify whether my understanding is correct? And if yes, please correct the implementation as suggested above. Thanks!

avdg81 (changes requested):
Overall, your changes look nice and clean. I have several minor suggestions and a few things that need to be checked. Could you please take care of that? Thanks!

applications/GeoMechanicsApplication/custom_processes/apply_normal_load_table_process.h
markelov208: Thank you. Info() is added now.

applications/GeoMechanicsApplication/custom_utilities/functions_for_parameters.h
markelov208: Fully agreed, so the files are renamed etc.

applications/GeoMechanicsApplication/custom_utilities/functions_for_parameters.cpp
markelov208: Thank you. It is removed now.

applications/GeoMechanicsApplication/custom_utilities/functions_for_parameters.h
markelov208: It is done.

applications/GeoMechanicsApplication/custom_utilities/functions_for_parameters.h
markelov208: the class is renamed as yo usuggested.

applications/GeoMechanicsApplication/custom_processes/apply_normal_load_table_process.cpp
markelov208: Indeed 0 and 1 are used in normal and tangential components. The constants are used now instead of the numbers.

applications/GeoMechanicsApplication/custom_processes/apply_normal_load_table_process.cpp
markelov208: Thank you. It is modified now.

applications/GeoMechanicsApplication/custom_processes/apply_normal_load_table_process.cpp
markelov208: As above.

applications/GeoMechanicsApplication/custom_processes/apply_normal_load_table_process.h
markelov208: restored.

applications/GeoMechanicsApplication/custom_utilities/functions_for_parameters.cpp
markelov208: Thank you for pointing. I also found only [ "table"].GetInt() and ["table"][0/1/2].GetInt(). The check is removed now.

avdg81 (approved):

rfaasse (approved):
Thanks for making the improvements, looks good!

PR #11712, opened 2023-10-23
[Core][MPI] Adding `MinLocAll` and `MaxLocAll` to `DataCommunicator`

loumalouomega:
**📝 Description**

Adding `MinLocAll` and `MaxLocAll` to `DataCommunicator`. This function not only computes the maximum/minimum but the partition where this values is local.

I also changed the suite in the serial  `DataCommunicator` tests, as it was wrong,

**🆕 Changelog**

- [Adding MinLocAll and MaxLocAll to DataCommunicator](https://github.com/KratosMultiphysics/Kratos/commit/218acb5a1c90ffbcf319b3ca6802c11d67a79d6f)
- Tests added
- Corrected the suite for the data communicator (serial)


kratos/mpi/sources/mpi_data_communicator.cpp
```diff
@@ -694,6 +709,33 @@ std::vector<TDataType> MPIDataCommunicator::AllReduceDetailVector(
...
+    } local_reduce, global_reduce;
+    local_reduce.value = rLocalValues.first;
+    local_reduce.rank = rLocalValues.second;
+    MPI_Datatype data_type;
+    if constexpr (std::is_same_v<TDataType, double>) {
+        data_type = MPI_DOUBLE_INT;
+    } else if constexpr (std::is_same_v<TDataType, long int>) {
+        data_type = MPI_LONG_INT;
+    } else if constexpr (std::is_same_v<TDataType, int>) {
+        data_type = MPI_2INT;
```
jcotela: I would prefer registering `MPIDataType`s for the various std::pair types in mpi_message.h instead of the if constexpr block here, so that the type deduction logic is all in one place.

jcotela (commented):
I am not convinced by the change of suite of the data_communicator tests but fair enough if people thinks it makes sense.

I would suggest to put all the logic for MPI datatype deduction on one place (mpi_message).

Other than that, looks good to me

kratos/mpi/sources/mpi_data_communicator.cpp
loumalouomega: Okay

loumalouomega: Done, I still needed the constexpr because there are many templated specializations not defined for MPIMessage and othwerwise is not possible to even compile

jcotela:
I changed it a bit so that MinLoc and MaxLoc are not defined for array/vector types, as I think they would not make sense with the current implementation (min/max by component, so that the result can be a composite of the values from different partitions). If this is not desirable, my preference would be to revert [cde37b7](https://github.com/KratosMultiphysics/Kratos/pull/11712/commits/cde37b73d0ce3a4cfbbc1beaa75b21c8ad574ecc) and [bb95bac](https://github.com/KratosMultiphysics/Kratos/pull/11712/commits/bb95bac49862cb2fea0b2c13077501d24d685cad) and keep @loumalouomega's original implementation.

RiccardoRossi:
could u document what the funciton returns possibly in its signature? 

is it (value, rank)  or something different?

loumalouomega:
> could u document what the funciton returns possibly in its signature?
> 
> is it (value, rank) or something different?

It is an std::pair of value, int

loumalouomega:
> could u document what the funciton returns possibly in its signature?
> 
> is it (value, rank) or something different?

Documentation added

kratos/mpi/includes/mpi_data_communicator.h
```diff
@@ -380,6 +393,20 @@ class KRATOS_API(KRATOS_MPI_CORE) MPIDataCommunicator: public DataCommunicator
...
+    * @brief Performs an AllReduce operation with location information.
+    * @details This function performs an AllReduce operation on a pair of data and an integer location using the specified MPI operation. The AllReduce operation combines the data from all processes and stores the result in the pair's first element. The location information (integer) is not modified during the AllReduce operation.
+    * @tparam TDataType The data type of the pair's first element.
+    * @param rLocalValues A pair containing the local data and location information to be reduced.
+    * @param Operation The MPI operation to use for the reduction.
+    * @return A pair where the first element contains the result of the AllReduce operation, and the second element is the unchanged location information.
+    */
+    template<class TDataType>
+    std::pair<TDataType, int> AllReduceDetailWithLocation(
+        const std::pair<TDataType, int>& rLocalValues,
```
philbucher, line 406: can you please explain what the location is?

loumalouomega: The partititon where the reduced value was found

loumalouomega: Wait

loumalouomega: Now

loumalouomega:
Feel free to approve @jcotela

loumalouomega:
> Feel free to approve @jcotela

?

philbucher (commented):
please add a python test as well then I approve

loumalouomega:
> please add a python test as well then I approve

Done

philbucher (approved):

PR #11714, opened 2023-10-23
[CI] improve intel compiler handling

philbucher:
Instead of installing the intel compiler in the CI, this PR does the following:
- Remove the intel compilers from the docker file
- Remove the legacy intel build (icc and icpc are deprecated)
- Manually install the intel compiler only for the intel build. I will experiment with this more in the future
- closes #11566


philbucher:
hm doesnt really work that easy, will try sth else first

PR #11715, opened 2023-10-23
[Docker] only new intel compiler

philbucher:
Only install the new intel compiler to reduce the size. I will remove the legacy intel build in a follow-up PR

EDIT remove valgrind, and install sudo to support installations in CI (need this for some experimentation)

hopefully helps with #11566


roigcarlo (approved):
If intel keeps running all good to me. I guess the only way to see is merging it.

I'll be watching for PR today to se how it behaves

PR #11716, opened 2023-10-24
[Core][MPI] Partially adapt `CppTestsUtilities` for MPI

loumalouomega:
**📝 Description**

This PR adapts to functions in  `CppTestsUtilities`  in order to be MPI ready.

**🆕 Changelog**

- [Partially adapt CppTestsUtilities for MPI](https://github.com/KratosMultiphysics/Kratos/commit/ea5dc26a227242477a49ed93d6c47084f1a2ce5f)

### DIFF
--- a/kratos/utilities/cpp_tests_utilities.cpp
+++ b/kratos/utilities/cpp_tests_utilities.cpp
@@ -403,32 +403,106 @@ ModelPart& CreateCubeSkinModelPart(
     Model& rCurrentModel,
     const double HalfX,
     const double HalfY,
-    const double HalfZ
+    const double HalfZ,
+    const DataCommunicator& rDataCommunicator
     )
 {
     // Generate the cube skin
     ModelPart& r_skin_part = rCurrentModel.CreateModelPart("Skin");
-    r_skin_part.CreateNewNode(1, -HalfX, -HalfY, -HalfZ);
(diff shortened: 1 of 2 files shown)


roigcarlo:
Could you explain a bit why you have made some of the decisions here?
- Why serial / distributed is different?
- In parallel, why nodes 1 to 4 are created both in rank 0 and 1 but not in the rest? and why the 5 to 8 are created in rank 1? Does it mean this will only work with world size 2?
- Why the elements are created in rank 1 and none of them in rank 0 in parallel but in rank 0 in serial?

Maybe an example on how this will be used in parallel would help?

loumalouomega:
> Could you explain a bit why you have made some of the decisions here?
> 
>     * Why serial / distributed is different?
> 
>     * In parallel, why nodes 1 to 4 are created both in rank 0 and 1 but not in the rest? and why the 5 to 8 are created in rank 1? Does it mean this will only work with world size 2?
> 
>     * Why the elements are created in rank 1 and none of them in rank 0 in parallel but in rank 0 in serial?
> 
> 
> Maybe an example on how this will be used in parallel would help?

It is a synthetic model part I create for certain Cpp tests. I only do 0 partitions when worl size = 1, and 2 partitions otherwise (more partititions would be complicated). Just that. It is used in the same places these methods are used in serial but for MPI tests.

roigcarlo (approved):

PR #11717, opened 2023-10-24
[CI] manually install intel compiler

philbucher:
this way we can remove it from the docker to save space


.github/workflows/ci.yml
```diff
@@ -408,14 +408,18 @@ jobs:
 
     container:
       image: kratosmultiphysics/kratos-image-ci-ubuntu-22-04:latest
-      options: --user 1001
+      # options: --user 1001 # must run as root to install the intel compiler
```
philbucher, line 413: @roigcarlo probably not the best solution, but for not I can live with this. The legacy intel compiler will be eventually removed anyway ...

roigcarlo: OK to me, but we will have to merge with the master for the other Ci, seems that a couple of precison errors (already solved) slip thrug

roigcarlo (approved):
:tada:

PR #11720, opened 2023-10-25
[Core] Update `GlobalPointerUtilities` with some changes related with `SpatialSearchResult`

loumalouomega:
**📝 Description**

Update `GlobalPointerUtilities` with some changes related with `SpatialSearchResult`.

**🆕 Changelog**

- [Update GlobalPointerUtilities with some changes related with `SpatialSearchResult`](https://github.com/KratosMultiphysics/Kratos/commit/e480a13e1a808604a9b04a3fbebc5a1521846dbe)

### DIFF
--- a/kratos/utilities/global_pointer_utilities.h
+++ b/kratos/utilities/global_pointer_utilities.h
@@ -27,6 +27,7 @@
 #include "containers/global_pointers_vector.h"
 #include "containers/global_pointers_unordered_map.h"
 #include "utilities/communication_coloring_utilities.h"
+#include "spatial_containers/spatial_search_result.h"
 
 namespace Kratos
 {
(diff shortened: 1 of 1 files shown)


kratos/utilities/global_pointer_utilities.h
```diff
@@ -27,6 +27,7 @@
 #include "containers/global_pointers_vector.h"
 #include "containers/global_pointers_unordered_map.h"
 #include "utilities/communication_coloring_utilities.h"
+#include "spatial_containers/spatial_search_result.h"
```
RiccardoRossi, line 30: i don't think you should add this include here...global_pointer_utilities are designed to be added anywhere in the code...

loumalouomega: Look why it is included

loumalouomega: I noticed that after your suggestions it is not needed anymore!

PR #11722, opened 2023-10-25
[GeoMechanicsApplication] Updated multiple moving load tests to work and to run

mcgicjn2:
**📝 Description**
Updated geomechanics testing to include the running of test set multiple moving loads automatically.
Updated the logic to include changes made to moving loads (structural mechanics)
Reduced duplication and code smells


applications/GeoMechanicsApplication/tests/test_set_multiple_moving_load_process.py
```diff
@@ -355,27 +384,21 @@ def test_SetMultipleMovingLoadsReverseGeomConfigurationNegative(self):
...
-        # create condition
-        self.mp.CreateNewCondition("MovingLoadCondition2D2N", 1, [2, 1], self.mp.GetProperties()[1])
+        self.create_two_nodes_condition(reverse_nodes=True)
 
         parameters = self.base_parameters
         parameters.AddVector("configuration", [-0.25])
 
         self.mp.ProcessInfo.SetValue(KratosMultiphysics.TIME,
-                                0)
+                                     0)
```
rfaasse: Minor comment, but could you move this one also to one line (as well as the DELTA_TIME lines), like you already did with the other ones?

applications/GeoMechanicsApplication/tests/test_set_multiple_moving_load_process.py
```diff
@@ -646,37 +624,25 @@ def test_SetMultipleMovingLoadsMultipleConditionsConfigurationNegative(self):
...
-
-        # create condition
-        self.mp.CreateNewCondition("MovingLoadCondition2D2N", 1, [1, 2], self.mp.GetProperties()[1])
-        self.mp.CreateNewCondition("MovingLoadCondition2D2N", 2, [2, 3], self.mp.GetProperties()[1])
+        self.create_three_nodes_condition()
 
         parameters = self.base_parameters
         parameters.AddVector("configuration", [-0.5])
 
         self.mp.ProcessInfo.SetValue(KratosMultiphysics.TIME,
```
rfaasse: Same here, could you please move this to a single line to make it consistent with the other changes?

applications/GeoMechanicsApplication/tests/test_set_multiple_moving_load_process.py
```diff
@@ -1,12 +1,54 @@
...
+        builder_and_solver = KratosMultiphysics.ResidualBasedBlockBuilderAndSolver(linear_solver)
+        scheme = KratosMultiphysics.ResidualBasedIncrementalUpdateStaticScheme()
+        convergence_criterion = KratosMultiphysics.ResidualCriteria(1e-1, 1e-1)
+        convergence_criterion.SetEchoLevel(0)
+
+        max_iters = 10
+        compute_reactions = False
+        reform_step_dofs = False
+        move_mesh_flag = False
+        strategy = KratosMultiphysics.ResidualBasedNewtonRaphsonStrategy(mp,
```
rfaasse: Do you think we really need a 'real' solving strategy here, or could we get away with creating a 'dummy' one, so we do not depend on all these objects in the test? You probably added the strategy for a reason, so let's discuss if this is a possibility here.

applications/GeoMechanicsApplication/tests/test_set_multiple_moving_load_process.py
```diff
@@ -24,11 +66,51 @@ def setUp(self):
...
+            self.mp.CreateNewCondition("MovingLoadCondition2D2N", 1, [1, 2], self.mp.GetProperties()[1])
+            self.mp.CreateNewCondition("MovingLoadCondition2D2N", 2, [2, 3], self.mp.GetProperties()[1])
+
+    def initialize_test(self, process):
+        process.ExecuteInitialize()
+        process.ExecuteInitializeSolutionStep()
+        self.strategy.InitializeSolutionStep()
+
+    def next_solution_step(self, process):
+        process.ExecuteFinalizeSolutionStep()
```
rfaasse, line 106: I see it was already the case before, but why do we first execute finalize and afterwards the initialize?

applications/GeoMechanicsApplication/tests/test_set_multiple_moving_load_process.py
```diff
@@ -1,12 +1,54 @@
...
+                                                                         compute_reactions,
+                                                                         reform_step_dofs,
+                                                                         move_mesh_flag)
+
+        strategy.SetEchoLevel(0)
+        return strategy
+
     def setUp(self):
         self.current_model = KratosMultiphysics.Model()
         self.rmp = self.current_model.CreateModelPart("root_part")
```
rfaasse: Since we're improving these tests a lot anyway, I would propose to rename these members: rmp -> root_model_part, mp -> model_part etc.

rfaasse (changes requested):
Very nice refactoring and code duplication reduction, looks a lot cleaner than before! I have a few minor comments/suggestions and one question about the newly added solving strategy.

applications/GeoMechanicsApplication/tests/test_set_multiple_moving_load_process.py
mcgicjn2: Unfortunately we need it to be able to initialize the movingloadcondition at each solution step as it appears the InitializeSolutionStep method although public in c++ is not exposed in python

applications/GeoMechanicsApplication/tests/test_set_multiple_moving_load_process.py
mcgicjn2: Basically because next_solution_step, is between steps, whereas initialize_test is the first.

applications/GeoMechanicsApplication/tests/test_set_multiple_moving_load_process.py
mcgicjn2: Although we could write some standard dummys in c++ to instantiate in python ...... but is it worth it ?

rfaasse: Indeed, creating a dummy in C++, which only has this one override which initializes the movingloadcondition at each step would be the way to get there. If this was a C++ test, it would be a no-brainer, because you can create this dummy locally in your test file, and just add the functionality you need. 

In this case, it is a bit more tricky, because it could happen that people start adding to this separate C++ 'dummy' more and more, until it becomes almost as complex as the real thing. 

My proposal would be to think a bit more about these kinds of re-usable test-utilities, because if it's done right, it will help a lot in making our tests more independent and test only the thing they need to test, meaning they also have a single reason for failure.

For now, this broader discussion should not block this any further, so I approved the PR.

FYI @avdg81

rfaasse (approved):
Thanks for making the changes and improving (and running) this test case!

PR #11723, opened 2023-10-26
[GeoMechanicsApplication] Enable line load in settlement workflow

carloslubbers:
**📝 Description**
Enable the line load variable from structural mechanics in the settlement workflow and write the values to the output file


avdg81 (approved):
Nice improvements. Thanks for the suggestions Carlos!

PR #11727, opened 2023-10-26
[GeoMechanicsApplication] Add method to time loop executor to cancel the loop

carloslubbers:
**📝 Description**
Implement the unused rShouldCancel delegate, by setting it on the time loop executor we can cancel runs while they are ongoing


rubenzorrilla:
Just pointing you to `Controller` class (see [here](https://github.com/KratosMultiphysics/Kratos/blob/master/kratos/controllers/controller.h)). I think it might be useful for this kind of purpose.

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
```diff
@@ -96,7 +102,14 @@ public :
         return end_state;
     }
 
+    bool IsCancelled() const
+    {
+        return mCancelDelegateAvailable && mCancelDelegate();
```
rfaasse: I was just thinking and discussing about this extra mCancelDelegateAvailable, and we don't need it. What we can do here is:
```suggestion
        return mCancelDelegate && mCancelDelegate();
```
if you call mCancelDelegate, it will return false when it is not set, so the order is important here.

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
```diff
@@ -96,7 +102,14 @@ public :
...
+    }
+
     std::unique_ptr<TimeIncrementor>  mTimeIncrementor;
+    bool                              mCancelDelegateAvailable = false;
```
rfaasse: This mCancelDelegateAvailable can then be removed

rfaasse (commented):
Looks clear, thanks for adding this! I just had an idea to make it a little bit cleaner (very minor change) after that, we're good to go!

rfaasse:
> Just pointing you to `Controller` class (see [here](https://github.com/KratosMultiphysics/Kratos/blob/master/kratos/controllers/controller.h)). I think it might be useful for this kind of purpose.

Thanks @rubenzorrilla for pointing to this interface, that's a good suggestion! The only issue here is that it required to be able to define this function outside of the C++ code through a C-interface, so unfortunately we cannot use the Controller here.

rfaasse (approved):

PR #11737, opened 2023-10-30
[Core] Adding `GetObjectType` and define `ObjectType` in `Tree`

loumalouomega:
**📝 Description**

Adding `GetObjectType` and define `ObjectType` in `Tree`. The `ObjectType` is taken from `PointType` when this is defined, otherwise `void` is defined. An auxiliary static method is defined in order to determine when `ObjectType` is defined.

**🆕 Changelog**

- [Adding `GetObjectType` and define `ObjectType` in `Tree`](https://github.com/KratosMultiphysics/Kratos/commit/fd8961f0cb9fc40447d97ca0be2eef9c0a69fbff)

### DIFF
--- a/kratos/spatial_containers/tree.h
+++ b/kratos/spatial_containers/tree.h
@@ -16,6 +16,7 @@
 #include <string>
 #include <iostream>
 #include <cmath>
+#include <type_traits>
 
 // External includes
 
@@ -40,6 +41,17 @@ namespace Kratos
 ///@}
 ///@name  Functions
 ///@{
+
+// Helper template to extract ObjectType or default to void
+template <typename T, typename = void>
+struct GetObjectType {
+    using type = void;
+};
+
+template <typename T>
+struct GetObjectType<T, std::void_t<typename T::ObjectType>> {
+    using type = typename T::ObjectType;
+};
 
 ///@}
 ///@name Kratos Classes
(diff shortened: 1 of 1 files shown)


roigcarlo (approved):
Hey didn't know that trick. That's actually super useful :+1:

loumalouomega:
> Hey didn't know that trick. That's actually super useful 👍

It is the result of hree hours of Google + Stackoverflow and GPT4 (the idea was from GPT4 but the promt was the result of hours of iterations with Google + Stackoverflow)

kratos/spatial_containers/tree.h
```diff
@@ -41,6 +42,17 @@ namespace Kratos
 ///@name  Functions
 ///@{
 
+// Helper template to extract ObjectType or default to void
+template <typename T, typename = void>
+struct GetObjectType {
```
pooyan-dadvand, line 47: I think this is too generic to be in the global namespace. Shall we protect it with Internals namespace or give it a Tree prefix to have it tied to it?

loumalouomega: Okay

loumalouomega: https://github.com/KratosMultiphysics/Kratos/pull/11739 here

PR #11742, opened 2023-10-31
[Core] Extend Expression for Flattening and Abs

sunethwarna:
**📝 Description**
This PR adds two features to `Expression` and `ContainerExpression`.
1. `Flatten()` method will flatten the given lazy expression to values while conserving the shape.
2. `Abs()` will give an expression which will give the component wise absolute values while conserving the shape.

**🆕 Changelog**
- Adds `UnaryAbsExpression`
- Adds `Expression::Flatten` and `ContainerExpression::Flatten`
- Adds `ContainerExpression::Abs` method


kratos/expression/expression.h
```diff
@@ -187,6 +187,19 @@ class KRATOS_API(KRATOS_CORE) Expression {
...
+     *
+     * Returns an expression containing a flattened expression. This will always
+     * be a LiteralFlatExpression. This keeps the item shape as it is.
+     *
+     * The flattened expression does not include any other expression combinations
+     * except the LiteralFlatExpression.
+     *
+     * @return Expression::ConstPointer     Flattened expression
+     */
+    ConstPointer Flatten() const;
```
matekelemen, line 201: The `Flatten` name is very confusing because it is already used for converting nested vectors/matrices into contiguous arrays. I'd rather call it `Collapse`, but I'm open to other suggestions too.

matekelemen: On a more general note, I don't think these kinds of functions should be members of `Expression`. They behave more like operators (produce a new `Expression` and don't mutate the host object), so they should be free-floating functions (or moved to some utility class).

kratos/expression/container_expression.h
```diff
@@ -240,6 +240,19 @@ class KRATOS_API(KRATOS_CORE) ContainerExpression {
...
+     *
+     * This returns a flattened container expression (which does not include any
+     * combination of any Expression except for a LiteralFlatExpression). A
+     * LiteralFlatExpression is created. The data represented by the underlying Expression
+     * in this ContainerExpression is put on that LiteralFlatExpression. Shape of the
+     * original ContainerExpression is preserved.
+     *
+     * @return ContainerExpression          Returns a flattened container expression.
+     */
+    ContainerExpression Flatten() const;
```
matekelemen, line 254: same remarks as for `Expression::Flatten`

kratos/python/add_expression_io_to_python.cpp
```diff
@@ -209,6 +209,7 @@ void AddExpressionIOToPython(pybind11::module& rModule)
         .def("GetItemShape", &Expression::GetItemShape)
         .def("NumberOfEntities", &Expression::NumberOfEntities)
         .def("GetItemComponentCount", &Expression::GetItemComponentCount)
+        .def("Flatten", &Expression::Flatten)
```
matekelemen, line 212: what about `Abs` for `Expression`?

kratos/tests/test_container_expression.py
```diff
@@ -665,6 +665,21 @@ def test_VectorWrite(self):
             velocity = Kratos.Array3([vector[i * 3], vector[i * 3 + 1], vector[i * 3 + 2]])
             self.assertVectorAlmostEqual(velocity, self._GetValue(entity, Kratos.VELOCITY))
 
+    def test_Abs(self):
```
matekelemen, line 668: I'm not sure how to do it, but you need to add a test for `Flatten` as well.

kratos/expression/expression.h
sunethwarna: Agree... Will move now :)

sunethwarna:
Closing since this is addressed in #11973. But keeping the branch untill that is merged.

PR #11746, opened 2023-11-01
[Core] Add SetResponseFunction method to ResidualBasedAdjointStaticScheme

sunethwarna:
**📝 Description**
This PR adds `ResidualBasedAdjointStaticScheme::SetResponseFunction` method. In the case where LHS of adjoint problem does not change (such as same primal problem for multiple response adjoints) and only RHS change due to change in the response function, this allows to reuse the already constructed LHS along with different RHS from the different responses.

**🆕 Changelog**
- Added `ResidualBasedAdjointStaticScheme::SetResponseFunction`


kratos/solving_strategies/schemes/residual_based_adjoint_static_scheme.h
```diff
@@ -166,6 +182,34 @@ class ResidualBasedAdjointStaticScheme : public Scheme<TSparseSpace, TDenseSpace
         KRATOS_CATCH("");
     }
 
+    void CalculateRHSContribution(
+        Element& rCurrentElement,
+        LocalSystemVectorType& rRHS_Contribution,
```
matekelemen:
```suggestion
        LocalSystemVectorType& rRHSContribution,
```

kratos/solving_strategies/schemes/residual_based_adjoint_static_scheme.h
```diff
@@ -237,6 +309,7 @@ class ResidualBasedAdjointStaticScheme : public Scheme<TSparseSpace, TDenseSpace
 
     AdjointResponseFunction::Pointer mpResponseFunction;
     std::vector<LocalSystemVectorType> mAdjointValues;
+    std::vector<LocalSystemMatrixType> mLHS;
```
matekelemen, line 312: I don't get it. Why do you need to store the LHS? The LHS is always computed in the function it's used in, and the stored value is never read outside those functions.

In my opinion, it would make more sense to declare local variables for the left hand sides, and dispose of them after the contribution is computed. Not only would it make the code cleaner, but it would also get rid of all `OpenMPUtils::ThisThread` calls.

sunethwarna: This will be addressed in the upcoming method to build the system matrix. Unfortunately, until then we have to rely on this mechanism.

kratos/solving_strategies/schemes/residual_based_adjoint_static_scheme.h
sunethwarna: done :)

matekelemen (approved):
:+1:

PR #11749, opened 2023-11-02
[Core] std::result_of => std::invoke_result

matekelemen:
`std::result_of` [is deprecated](https://en.cppreference.com/w/cpp/types/result_of#Notes) and replaced by `std::invoke_result`.


loumalouomega (dismissed):
Okay from my side

RiccardoRossi (dismissed):

RiccardoRossi:
ouch, i approved because it looks correct...but it looks like it is not that easy :-D

matekelemen:
I missed the syntax change :/

loumalouomega (approved):

PR #11750, opened 2023-11-02
[GeoMechanicsApplication] Fixed a bug and removed some duplicated code

avdg81:
**📝 Description**
- Pass the correct model part to `ApplyNormalLoadTableProcess`. In the settlement work flow, when creating an `ApplyNormalLoadTableProcess` object, we mistakenly passed it the main model part rather than the model part that is defined in the process's settings. That has been corrected now.
- Extracted a function that makes process creators. By extracting a function that produces process creators, we have removed some duplicated code.


rfaasse (approved):
Looks good, nicely consistent and less duplication! No comments from my side!

WPK4FEM (approved):
Consistent. However, I think that the new MakeCreatorFor is so generic that it will find a place outside dgeosettlement for reuse in other workflows when those will be refactored or created.

PR #11753, opened 2023-11-02
[GeoMechanicsApplication] Add new element class for heat

### LINKED ISSUE #11734: [GeoMechanicsApplication] Add new element class for heat
As a developer, I would like to add a new element class for heat such that I can use later.

### Description 

**This feature is fully implemented.**

Branch: https://github.com/KratosMultiphysics/Kratos/tree/geo/10372-HeatTransfer-temp

A new element class is added for heat, TransientThermalElement: includes creat, check, intialize, RHS vectors, LHS matrices, etc.
applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.hpp
applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.cpp
Whole files

### Acceptance Criteria

**Given** the geomechanics application
**When** the application checks the dimensions, availability of material parameters, DOFs and if valid,
**Then** there is no exception 

**Given** the geomechanics application
**When** the checks of the dimensions, availability of material parameters, DOFs and if fails,
**Then** the application throws an error and stops

------------------------------------------------------------------------------

As a developer, I would like to register the thermal element in GeoMechanicsApplication in order to make a functional element for this application.

Description
This feature is fully implemented.

Branch: https://github.com/KratosMultiphysics/Kratos/tree/geo/10372-HeatTransfer-temp

A new block of thermal elements are added to GeoMechanicsApplications (One element but with different geometries, eg. Triangle2D3N, 6N, etc ...).
applications/GeoMechanicsApplication/geo_mechanics_application.h
Lines 450 to 460

The thermal elements are registered in GeoMechanicsApplication
applications/GeoMechanicsApplication/geo_mechanics_application.cpp
Lines 211 to 220

Acceptance Criteria
Given the geomechanics application
When check for the following templeted elements
Then they must exist abd to be registered

TransientThermalElement2D3N
TransientThermalElement2D4N
TransientThermalElement3D4N
TransientThermalElement3D8N
TransientThermalElement2D6N
TransientThermalElement2D8N
TransientThermalElement2D9N
TransientThermalElement3D10N
TransientThermalElement3D20N
TransientThermalElement3D27N

markelov208:
**📝 Description**
A brief description of the PR.

- added transient thermal element

**🆕 Changelog**
E.g.
- cleaned and refactored transient_thermal_element.h/cpp file
- registered instantiated classes of transient_thermal_element in geo_mechanics_application.h/cpp
- added unit tests for all instantiated classes


applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
```diff
@@ -0,0 +1,155 @@
...
+//  Main authors:    Mohamed Nabi
+//                   John van Esch
+//                   Gennady Markelov
+//
+
+#pragma once
+
+#include "custom_utilities/element_utilities.hpp"
+#include "geo_mechanics_application_variables.h"
+#include "includes/serializer.h"
```
avdg81: I was a bit surprised to not see `#include "includes/element.h" in this list`, since `Element` is the base class of `TransientThermalElement`. So I expect it to be included indirectly, which I don't really like. If the indirect includes change (and I believe they will), that may break the building of this class. I would suggest to add `#include "includes/element.h"` to this list.

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.cpp
```diff
@@ -0,0 +1,32 @@
...
+//  Main authors:    Mohamed Nabi
+//                   John van Esch
+//                   Gennady Markelov
+//
+
+#include "custom_elements/transient_thermal_element.h"
+
+namespace Kratos {
+
+template class TransientThermalElement<2, 3>;
```
rfaasse, line 19: It's not completely clear to me how this works. We specify certain instances of the template?

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
```diff
@@ -0,0 +1,303 @@
...
+        CheckProperty(SOLID_COMPRESSIBILITY);
+
+        if constexpr(TDim == 3) {
+            CheckProperty(THERMAL_CONDUCTIVITY_SOLID_ZZ);
+            CheckProperty(THERMAL_CONDUCTIVITY_SOLID_YZ);
+            CheckProperty(THERMAL_CONDUCTIVITY_SOLID_XZ);
+        }
+    }
+
+    void CheckDomainSize() const
```
rfaasse: For a template class, is it also not possible to have these kinds of functions in the .cpp, which don't use the template arguments?

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
```diff
@@ -0,0 +1,303 @@
...
+    TransientThermalElement(IndexType               NewId,
+                            GeometryType::Pointer   pGeometry,
+                            PropertiesType::Pointer pProperties)
+            : Element(NewId, pGeometry, pProperties)
+    {
+    }
+
+    ~TransientThermalElement() override = default;
+    TransientThermalElement(const TransientThermalElement&) = delete;
+    TransientThermalElement& operator=(const TransientThermalElement&) = delete;
```
rfaasse: Not sure, but SQ might complain about the move assignment operator and move constructor (rule of 5).

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
```diff
@@ -0,0 +1,303 @@
...
+
+        return 0;
+    }
+
+    void GetDofList(DofsVectorType& rElementalDofList,
+                    const ProcessInfo& ) const override
+    {
+        KRATOS_TRY
+
+        rElementalDofList.resize(TNumNodes);
```
rfaasse, line 63: If the size is different, should we just continue, or do we want to log a warning/throw an error.

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
```diff
@@ -0,0 +1,303 @@
...
+        KRATOS_CATCH("")
+    }
+
+    void CalculateLocalSystem(MatrixType&        rLeftHandSideMatrix,
+                              VectorType&        rRightHandSideVector,
+                              const ProcessInfo& rCurrentProcessInfo) override
+    {
+        KRATOS_TRY
+
+        GeometryType::ShapeFunctionsGradientsType DN_DXContainer;
```
rfaasse: The name here does not adhere to the kratos style guide (same for detJContainer in the next line)

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
```diff
@@ -0,0 +1,303 @@
...
+                            PropertiesType::Pointer pProperties)
+        : Element(NewId, pGeometry, pProperties)
+    {
+    }
+
+    ~TransientThermalElement() override = default;
+    TransientThermalElement(const TransientThermalElement&) = delete;
+    TransientThermalElement& operator=(const TransientThermalElement&) = delete;
+
+    Element::Pointer Create(IndexType               NewId,
```
rfaasse, line 44: I thought all ::Pointers were shared pointer types, is it different for elements?

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
```diff
@@ -0,0 +1,303 @@
...
+        CheckHasSolutionStepsDataFor(TEMPERATURE);
+        CheckHasSolutionStepsDataFor(DT_TEMPERATURE);
+        CheckHasDofsFor(TEMPERATURE);
+        CheckProperties();
+        CheckForNonZeroZCoordinateIn2D();
+
+        KRATOS_CATCH("")
+
+        return 0;
+    }
```
rfaasse: In my opinion it would be nicer if the functions that are called here are placed directly under the check function

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
```diff
@@ -0,0 +1,303 @@
...
+        auto result = array_1d<double, TNumNodes>{};
+        const auto& r_geometry = GetGeometry();
+        std::transform(r_geometry.begin(), r_geometry.end(), result.begin(),
+                       [&rNodalVariable](const auto& node) {
+                           return node.FastGetSolutionStepValue(rNodalVariable);
+                       });
+        return result;
+    }
+
+    void CheckProperty(const Kratos::Variable<double>& rVariable) const
```
rfaasse: I would swap the order and place CheckProperty under CheckProperties

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
```diff
@@ -0,0 +1,303 @@
...
+    {
+        KRATOS_ERROR_IF_NOT(GetProperties().Has(rVariable)) << rVariable.Name()
+                                                            << " does not exist in the material properties" << std::endl;
```
rfaasse: is 'material properties' in this case the right wording, maybe 'thermal element' is more descriptive?

applications/GeoMechanicsApplication/tests/cpp_tests/test_transient_thermal_element.cpp
```diff
@@ -0,0 +1,428 @@
...
+void TestThermalElement(ModelPart& rModelPart)
+{
+    Element::Pointer p_element = rModelPart.pGetElement(1);
+
+    for (unsigned int i = 0; i < rModelPart.NumberOfNodes(); i++) {
+        p_element->GetGeometry()[i].AddDof(TEMPERATURE);
+    }
+
+    const ProcessInfo& r_current_process_info = rModelPart.GetProcessInfo();
+    // p_element->Check(r_current_process_info);
```
rfaasse: This is still commented, what do we want to do with this?

applications/GeoMechanicsApplication/tests/cpp_tests/test_transient_thermal_element.cpp
```diff
@@ -0,0 +1,428 @@
...
+    }
+}
+
+void GenerateTransientThermalElement2D3N(ModelPart& rModelPart)
+{
+    // Geometry creation
+    rModelPart.CreateNewNode(1, 0.0, 0.0, 0.0);
+    rModelPart.CreateNewNode(2, 1.0, 0.0, 0.0);
+    rModelPart.CreateNewNode(3, 1.0, 1.0, 0.0);
+    std::vector<ModelPart::IndexType> elemNodes{1, 2, 3};
```
rfaasse: To adhere to the style guide: elemNodes -> elem_nodes (also in all other generate functions)

applications/GeoMechanicsApplication/tests/cpp_tests/test_transient_thermal_element.cpp
```diff
@@ -0,0 +1,428 @@
...
+
+    for (unsigned int i = 0; i < rModelPart.NumberOfNodes(); i++) {
+        p_element->GetGeometry()[i].AddDof(TEMPERATURE);
+    }
+
+    const ProcessInfo& r_current_process_info = rModelPart.GetProcessInfo();
+    // p_element->Check(r_current_process_info);
+    // Check requires a solution?
+
+    Element::DofsVectorType ElementalDofList;
```
rfaasse: To adhere to the style guide: ElementalDofList-> elemental_dof_list

applications/GeoMechanicsApplication/tests/cpp_tests/test_transient_thermal_element.cpp
```diff
@@ -0,0 +1,428 @@
...
+    p_element->GetDofList(ElementalDofList, r_current_process_info);
+
+    for (unsigned int i = 0; i < ElementalDofList.size(); i++) {
+        ElementalDofList[i]->SetEquationId(i);
+    }
+
+    Element::EquationIdVectorType EquationIdVector;
```
rfaasse: To adhere to the style guide: ElementalDofList->equation_id_vector

applications/GeoMechanicsApplication/tests/cpp_tests/test_transient_thermal_element.cpp
```diff
@@ -0,0 +1,428 @@
...
+    p_element->EquationIdVector(EquationIdVector, r_current_process_info);
+
+    // Check the EquationIdVector values
+    for (unsigned int i = 0; i < EquationIdVector.size(); i++) {
+        KRATOS_EXPECT_TRUE(EquationIdVector[i] == i);
```
rfaasse: We are missing some more asserts and checks on the element, but since we want to start testing with the integration tests as soon as possible, that could be a separate PR

rfaasse (commented):
The code looks very clean, I had no issues in reading and understanding the functionality. Thanks for all the effort @markelov208 @mnabideltares and @avdg81. I have some comments, but they are mainly related to order, formatting etc.

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.cpp
avdg81: This technique is called "template instantiation" and it is used for all (or nearly all) elements in the `GeoMechanicsApplication`. Basically, it forces the compiler to instantiate the class template given a set of template parameters. I have personally little experience with this technique (it was already used in this code base before I even joined the Kratos team). Although it may be useful here, we may not use it in the most optimal way. For instance, see this question on SO: https://stackoverflow.com/questions/2351148/explicit-template-instantiation-when-is-it-used. My suggestion is to investigate this in more detail later (e.g. by consulting some reference books), since it is a good question.

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
avdg81: I believe this question is partially related to your previous question. In general, member function definitions of class templates go in the header file. But there are situations where you may want to move them to a separate implementation file, as suggested by the question on SO: https://stackoverflow.com/questions/2351148/explicit-template-instantiation-when-is-it-used. Long story short: let's investigate the template instantiation mechanism in more detail first, and then we can decide about the best way forward. Do you agree?

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
avdg81: Good point. I have checked the copy operations of the base class (`Element`) and I've found those are actually implemented. As such, it is unclear to my why the copy operations are `delete`d in this derived class in the first place. Considering this, I would propose to apply the rule of zero, and let the compiler generate the needed special member functions.

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
avdg81: I believe we should just continue. The point is that this member function returns its result through the argument list (`rElementalDofList`), and not using its return type. Although I would be in favor of changing the function's signature (to return the result using the return type rather than via an "out" parameter), this is a bit involved, since this member function is inherited from the Kratos Core class `Element`. As far as I can see, the size of `rElementalDofList` on entry has no meaning, so I believe we can safely resize it without further notice.

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
avdg81: Thanks for pointing out these inconsistencies. Done.

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
avdg81: Apparently, class `Element` uses macro `KRATOS_CLASS_INTRUSIVE_POINTER_DEFINITION`, which results in `Element::Pointer` being an alias for `Kratos::intrusive_ptr<Element>`. Perhaps at some point the `intrusive_ptr` class template may be replaced by a smart pointer from the STL? However, that is a discussion to be held with the Kratos Technical Committee. As such, it is out of scope for this PR.

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
avdg81: Fair point. Done.

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
avdg81: Done.

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
avdg81: Agreed.

applications/GeoMechanicsApplication/tests/cpp_tests/test_transient_thermal_element.cpp
avdg81: I have removed it.

applications/GeoMechanicsApplication/tests/cpp_tests/test_transient_thermal_element.cpp
avdg81: Changed it to `node_ids`.

applications/GeoMechanicsApplication/tests/cpp_tests/test_transient_thermal_element.cpp
avdg81: Done.

applications/GeoMechanicsApplication/tests/cpp_tests/test_transient_thermal_element.cpp
avdg81: Renamed to `equation_ids`.

applications/GeoMechanicsApplication/tests/cpp_tests/test_transient_thermal_element.cpp
avdg81: Agreed. I will create a new issue for this one.

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
rfaasse: Agreed, let's keep it in mind for improvement possibilities

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
```diff
@@ -0,0 +1,299 @@
...
+            result += prod(rShapeFunctionGradients[GPoint], Temp) * rIntegrationCoefficients[GPoint];
+        }
+
+        return result;
+    }
+
+    BoundedMatrix<double, TNumNodes, TNumNodes> CalculateCapacityMatrix(const Vector& rIntegrationCoefficients) const
+    {
+        const auto& r_properties = GetProperties();
+        const auto  cWater = r_properties[POROSITY] * r_properties[SATURATION] *
```
rfaasse: couple more inconsistencies in this function (c_water, c_solid, n_container, g_point, n)

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
```diff
@@ -0,0 +1,299 @@
...
+    BoundedMatrix<double, TNumNodes, TNumNodes> CalculateConductivityMatrix(const GeometryType::ShapeFunctionsGradientsType& rShapeFunctionGradients,
+                                                                            const Vector& rIntegrationCoefficients,
+                                                                            const ProcessInfo& rCurrentProcessInfo) const
+    {
+        GeoThermalDispersionLaw law{TDim};
+        const auto constitutive_matrix =
+                law.CalculateThermalDispersionMatrix(GetProperties(), rCurrentProcessInfo);
+
+        auto result = BoundedMatrix<double, TNumNodes, TNumNodes>{ZeroMatrix{TNumNodes, TNumNodes}};
+        for (unsigned int GPoint = 0; GPoint < GetGeometry().IntegrationPointsNumber(GetIntegrationMethod()); ++GPoint) {
```
rfaasse: GPoint->g_point

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
```diff
@@ -0,0 +1,299 @@
...
+                array_1d<double, TNumNodes>{-prod(rConductivityMatrix, GetNodalValuesOf(TEMPERATURE))};
+        rRightHandSideVector += conductivity_vector;
+    }
+
+    Vector CalculateIntegrationCoefficients(const Vector& rDetJContainer) const
+    {
+        const auto& r_integration_points = GetGeometry().IntegrationPoints(GetIntegrationMethod());
+
+        auto result = Vector{r_integration_points.size()};
+        for (unsigned int GPoint = 0; GPoint < r_integration_points.size(); ++GPoint) {
```
rfaasse: GPoint -> g_point

rfaasse (commented):
Just a final few style inconsistencies, otherwise, good to go!

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
avdg81: This has been fixed.

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
avdg81: Done.

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
avdg81: Done.

applications/GeoMechanicsApplication/custom_elements/transient_thermal_element.h
avdg81: Done.

rfaasse (approved):
Looks good!

PR #11719, opened 2023-10-25
[Core][MPI] Adding `SpatialSearchResultContainer`/`SpatialSearchResultContainerVector`  and `SearchWrapper` (v.4.5)

loumalouomega:
**📝 Description**

#11416 + addressed comments

This PR introduces a new class named `SpatialSearchResultContainer`, `SpatialSearchResultContainerVector` as well as `SearchWrapper`. The two first classes provides storage and operations related to spatial search results. In addition `SearchWrapper` class implements an interface for serial/MPI search for different search objects,  in current PR adding `GeometricalObjectBins` and the available trees.

The `SpatialSearchResultContainer`  class stores:

1. Local pointers of results (in a `PointerVectorSet`)
2. Global pointers (empty if not synchronized)
3. GlobalPointerCommunicator pointer (to synchronize the global pointers)

The class includes functions such as, `SynchronizeAll`, `HasResult`, `GetResultIndices`, `AddResult`... The interesting part is that integrates a global pointer communicator, so it is possible to calculate anything passing a functor. The most common operations required are already implemented:  `GetResultShapeFunctions`,  `GetResultIndices` and  `GetResultCoordinates`,

The `SynchronizeAll` method synchronizes the container between partitions. The `AddResult` method adds a result to the local pointers. There is a `push_back` which does the same.

The `SpatialSearchResultContainerVector`  class stores a std::vector of `SpatialSearchResultContainer` pointers. In the `SearchWrapper` class the index of the vector corresponds with the node id when using nodes and just a counter when simply searching.

Considering general data communicator instead of sub data communicator

**TODO**: Using coloring utilities instead of manual coloring

**🆕 Changelog**

- The `spatial_search_result_container.cpp` , `spatial_search_result_container.h`m  `spatial_search_result_container_vector.cpp` and `spatial_search_result_container_vector.h` files have been added, representing the implementation and header for the `SpatialSearchResultContainer` and `SpatialSearchResultContainerVector` classes.
- Expose to python `SpatialSearchResultContainer` and `SpatialSearchResultContainerVector` 
- Adding tests for `SpatialSearchResultContainer` and `SpatialSearchResultContainerVector` 
- Update python interface and redefine tests with new interface
- Adding `SearchWrapper` class, currently  just implemented `GeometricalObjectBins`.
- Adding tests for  `SearchWrapper`.
- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11710
- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11712
[...]

PR #11762, opened 2023-11-03
[Core] Add `SetRank` method to `GlobalPointer` +  documentation

loumalouomega:
**📝 Description**

Add `SetRank` method to `GlobalPointer`, plus adding  documentation

**🆕 Changelog**

- [Add `SetRank` method to `GlobalPointer` +  documentation](https://github.com/KratosMultiphysics/Kratos/commit/bd8951cfb18dd09adcaeaed7a9f980e5b594f23b)

### DIFF
--- a/kratos/includes/global_pointer.h
+++ b/kratos/includes/global_pointer.h
@@ -22,272 +22,409 @@
 #include "includes/serializer.h"
 #include "includes/key_hash.h"
 
-namespace Kratos {
+namespace Kratos
+{
+///@addtogroup KratosCore
+///@{
 
+///@name Kratos Classes
+///@{
+
+/**
+* @class GlobalPointer
+* @ingroup KratosCore
+* @brief This class is a wrapper for a pointer to a data that is located in a different rank
+* @details This class is a wrapper for a pointer to a data that is located in a different rank
+* @tparam TDataType The type of the data pointed by the GlobalPointer
+* @see GlobalPointerHasher
+* @author Carlos A. Roig
+*/
 template<class TDataType>
-class GlobalPointer {
-private:
-
-  /// Pointer to the data
-  TDataType * mDataPointer;
-#ifdef KRATOS_USING_MPI
-  int mRank;
-#endif
-
+class GlobalPointer
+{
 public:
+    ///@name Type Definitions
+    ///@{
 
-  typedef TDataType element_type;
+    /// Definition of element type
+    using element_type = TDataType;
 
- /** Default constructor
-	* Default constructor
-	* This should never be called as we need a local pointer to exists
-	*/
-	GlobalPointer() {
-    mDataPointer = nullptr;
+    ///@}
+    ///@name Life Cycle
+    ///@{
+
+   /**
+    * @brief Default constructor
+    * @details This should never be called as we need a local pointer to exists
+    */
+    GlobalPointer() {
+        mDataPointer = nullptr;
 #ifdef KRATOS_USING_MPI
-    this->mRank = 0;
+        this->mRank = 0;
 #endif
-  };
+    };
 
-  /** Constructor by Data
-   * Constructor by Data
-   * @param Data the data.
-   */
-  GlobalPointer(TDataType Data) = delete;
+    /**
+     * @brief Constructor by Data
+     * @param Data the data.
+     */
+    GlobalPointer(TDataType Data) = delete;
 
-  /** Constructor by DataPointer
-   * Constructor by DataPointer
-   * @param DataPointer Pointer to the data.
-   */
-  GlobalPointer(TDataType * DataPointer, int Rank = 0)
-    : mDataPointer(DataPointer)
+    /**
+     * @brief Constructor by DataPointer
+     * @param DataPointer Pointer to the data.
+     */
+    GlobalPointer(
+        TDataType* DataPointer,
+        int Rank = 0
+        ) : mDataPointer(DataPointer)
 #ifdef KRATOS_USING_MPI
-    , mRank(Rank)
+          , mRank(Rank)
 #endif
-     {
+    {
 #ifndef KRATOS_USING_MPI
-     KRATOS_DEBUG_ERROR_IF(Rank != 0) << "trying to construct a global pointer with rank different from zero when kratos is not in MPI mode " << std::endl;
+        KRATOS_DEBUG_ERROR_IF(Rank != 0) << "Trying to construct a global pointer with rank different from zero when kratos is not in MPI mode " << std::endl;
 #endif
-     }
+    }
 
-  /** Constructor by Kratos::shared_ptr
-   * Constructor by Kratos::shared_ptr
-   * @param DataPointer Boost Shared Pointer to the Data.
-   */
-  GlobalPointer(Kratos::shared_ptr<TDataType> DataPointer, int Rank = 0)
-    : mDataPointer(DataPointer.get())
+    /**
+     * @brief Constructor by Kratos::shared_ptr
+     * @param DataPointer Std Shared Pointer to the Data.
+     */
+    GlobalPointer(
+        Kratos::shared_ptr<TDataType> DataPointer,
+        int Rank = 0
+        ) : mDataPointer(DataPointer.get())
 #ifdef KRATOS_USING_MPI
-    , mRank(Rank) 
+          , mRank(Rank)
 #endif
-  {
+    {
 #ifndef KRATOS_USING_MPI
-     KRATOS_DEBUG_ERROR_IF(Rank != 0) << "trying to construct a global pointer with rank different from zero when kratos is not in MPI mode " << std::endl;
+        KRATOS_DEBUG_ERROR_IF(Rank != 0) << "Trying to construct a global pointer with rank different from zero when kratos is not in MPI mode " << std::endl;
 #endif
-  }
+    }
 
-  /** Constructor by Kratos::shared_ptr
-   * Constructor by Kratos::shared_ptr
-   * @param DataPointer Boost Shared Pointer to the Data.
-   */
-  GlobalPointer(Kratos::intrusive_ptr<TDataType>& DataPointer, int Rank = 0)
-    : mDataPointer(DataPointer.get())
-#ifdef KRATOS_USING_MPI
-    , mRank(Rank) 
-#endif
-  {
+    /**
+     * @brief Constructor by Kratos::intrusive_ptr
+     * @param DataPointer Boost intrusive Pointer to the Data.
+     */
+    GlobalPointer(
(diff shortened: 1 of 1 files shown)


roigcarlo:
Why exactly do you need to change the rank of a global pointer? 

GP were designed to be created in the partition they belong, its hard for me to find a use case in which this construction is valid and does not lead to problems.

loumalouomega:
> Why exactly do you need to change the rank of a global pointer?
> 
> GP were designed to be created in the partition they belong, its hard for me to find a use case in which this construction is valid and does not lead to problems.

Because somethimes the rank is not provided in construction time

roigcarlo:
Well yes, but can you elaborate? how can you not have the rank in the processor its mean to create the GP?

loumalouomega:
> Well yes, but can you elaborate? how can you not have the rank in the processor its mean to create the GP?

Because the `GeometricalObjectBins` generates `SpatialSearchResult`, which contains a `GlobalPointer`, but does not provide the rank, so it is assumed to be 0 always.

See: https://github.com/KratosMultiphysics/Kratos/pull/11719/commits/d6873a9b10050387427775abd765304f94b6697f

roigcarlo:
if this is the problem I would personally change the `global_pointer.h` default to something like:

```C++
GlobalPointer(TDataType data, int rank = ParallelEnvironment::GetDefaultRank())
```

But maybe I am biased here, ping @KratosMultiphysics/mpi

loumalouomega:
> if this is the problem I would personally change the `global_pointer.h` default to something like:
> 
> ```c++
> GlobalPointer(TDataType data, int rank = ParallelEnvironment::GetDefaultRank())
> ```
> 
> But maybe I am biased here, ping @KratosMultiphysics/mpi

That will not work, because creates a MPI dependency, will fail in linking time

roigcarlo:
It does not work like that

When you start Kratos with mpi, the first time the kernel is loaded you will call:

```Python
# __init__.py:67
mpi.InitializeMPIParallelRun()
```

This code lives in the MPI extension of the core, and among other things does:

```C++
// Define the World DataCommunicator as a wrapper for MPI_COMM_WORLD and make it the default.
ParallelEnvironment::RegisterDataCommunicator("World", MPIDataCommunicator::Create(MPI_COMM_WORLD), ParallelEnvironment::MakeDefault);
```

This will invoke the code from:

```c++
void ParallelEnvironment::SetAsDefault(
    std::unordered_map<std::string, DataCommunicator::UniquePointer>::iterator& rThisCommunicator)
{
    mDefaultCommunicator = rThisCommunicator;
    const auto& r_comm = *(rThisCommunicator->second);
    mDefaultRank = r_comm.Rank();
    mDefaultSize = r_comm.Size();
}
```
Which will set the `mDefaultcommunicator`, `mDefaultRank` and `mDefaultSize` be the ones from the communicator you just created.

It does not matter that the implementation this part is MPI dependent, because it does not live in the same shared lib as the rest of KratosCore, but it will change the default registered communicator. 

When you call `ParallelEnvironment::GetDefaultRank()`, (which does not depend on mpi) you will access proper process rank, size, or mpi communicator, even if they are mpi dependent.

roigcarlo:
Btw, I will admit that `SetRank` would be usefull in one particular situation, which is trying to transfer GP which belong to more than one communicator (Kratos default one + other custom registered comm) with different rank ids. And even in those i am not sure if its better to create another copy of the pointer when in the context of the correct communicator.

loumalouomega:
> It does not work like that
> 
> When you start Kratos with mpi, the first time the kernel is loaded you will call:
> 
> ```python
> # __init__.py:67
> mpi.InitializeMPIParallelRun()
> ```
> 
> This code lives in the MPI extension of the core, and among other things does:
> 
> ```c++
> // Define the World DataCommunicator as a wrapper for MPI_COMM_WORLD and make it the default.
> ParallelEnvironment::RegisterDataCommunicator("World", MPIDataCommunicator::Create(MPI_COMM_WORLD), ParallelEnvironment::MakeDefault);
> ```
> 
> This will invoke the code from:
> 
> ```c++
> void ParallelEnvironment::SetAsDefault(
>     std::unordered_map<std::string, DataCommunicator::UniquePointer>::iterator& rThisCommunicator)
> {
>     mDefaultCommunicator = rThisCommunicator;
>     const auto& r_comm = *(rThisCommunicator->second);
>     mDefaultRank = r_comm.Rank();
>     mDefaultSize = r_comm.Size();
> }
> ```
> 
> Which will set the `mDefaultcommunicator`, `mDefaultRank` and `mDefaultSize` be the ones from the communicator you just created.
> 
> It does not matter that the implementation this part is MPI dependent, because it does not live in the same shared lib as the rest of KratosCore, but it will change the default registered communicator.
> 
> When you call `ParallelEnvironment::GetDefaultRank()`, (which does not depend on mpi) you will access proper process rank, size, or mpi communicator, even if they are mpi dependent.

I will try

roigcarlo:
Well let see what the rest of people say, maybe I am being paranoid and setrank is fine for everyone...

loumalouomega:
> > It does not work like that
> > When you start Kratos with mpi, the first time the kernel is loaded you will call:
> > ```python
> > # __init__.py:67
> > mpi.InitializeMPIParallelRun()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This code lives in the MPI extension of the core, and among other things does:
> > ```c++
> > // Define the World DataCommunicator as a wrapper for MPI_COMM_WORLD and make it the default.
> > ParallelEnvironment::RegisterDataCommunicator("World", MPIDataCommunicator::Create(MPI_COMM_WORLD), ParallelEnvironment::MakeDefault);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This will invoke the code from:
> > ```c++
> > void ParallelEnvironment::SetAsDefault(
> >     std::unordered_map<std::string, DataCommunicator::UniquePointer>::iterator& rThisCommunicator)
> > {
> >     mDefaultCommunicator = rThisCommunicator;
> >     const auto& r_comm = *(rThisCommunicator->second);
> >     mDefaultRank = r_comm.Rank();
> >     mDefaultSize = r_comm.Size();
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Which will set the `mDefaultcommunicator`, `mDefaultRank` and `mDefaultSize` be the ones from the communicator you just created.
> > It does not matter that the implementation this part is MPI dependent, because it does not live in the same shared lib as the rest of KratosCore, but it will change the default registered communicator.
> > When you call `ParallelEnvironment::GetDefaultRank()`, (which does not depend on mpi) you will access proper process rank, size, or mpi communicator, even if they are mpi dependent.
> 
> I will try

It is impossible, I need many forward declarations in many different classes (20 files I have right now). It is easier to just simply keet the SetRank method

roigcarlo:
Actually, the problem is the `FillCommunicator`. Let me take a look

loumalouomega:
> Actually, the problem is the `FillCommunicator`. Let me take a look

yes, but I needed to change many files

roigcarlo (approved):
Ok to many things are wrong in the core. Lets accept this for now and I will take care of unraveling the kernel

loumalouomega:
> Ok to many things are wrong in the core. Lets accept this for now and I will take care of unraveling the kernel

Indeed..., many many cross dependencies. I think we should reconsider the forward declaration policy to begin with

kratos/includes/global_pointer.h
```diff
@@ -22,272 +22,409 @@
...
+
+   /**
+    * @brief Default constructor
+    * @details This should never be called as we need a local pointer to exists
+    */
+    GlobalPointer() {
+        mDataPointer = nullptr;
 #ifdef KRATOS_USING_MPI
-  int mRank;
+        this->mRank = 0;
```
RiccardoRossi, line 63: why do you want to initialize it? if it is not a correct value better leave it undefined no?

kratos/includes/global_pointer.h
```diff
@@ -22,272 +22,409 @@
...
+    }
+
+    /**
+     * @brief Sets the rank of the global pointer
+     * @details Sets the rank of the global pointer where data is located
+     * @param Rank The rank to set
+     */
+    void SetRank(const int Rank)
+    {
 #ifdef KRATOS_USING_MPI
```
RiccardoRossi, line 278: well, i guess we should have a debug error here in case we are not in mpi mode

RiccardoRossi:
> > Ok to many things are wrong in the core. Lets accept this for now and I will take care of unraveling the kernel
> 
> Indeed..., many many cross dependencies. I think we should reconsider the forward declaration policy to begin with

i am fundamentally against it!

kratos/includes/global_pointer.h
loumalouomega: The diff is terrible, the code was already like that , see line 48 in previous implementation

kratos/includes/global_pointer.h
loumalouomega: And rank is different than 0, because if you implement a generic code in serial it will set always to 0, otherwise you will create many false errors

loumalouomega:
> > > Ok to many things are wrong in the core. Lets accept this for now and I will take care of unraveling the kernel
> > 
> > 
> > Indeed..., many many cross dependencies. I think we should reconsider the forward declaration policy to begin with
> 
> i am fundamentally against it!

In that case we will have longer compilation times and corss dependencies issues all along the core

kratos/includes/global_pointer.h
RiccardoRossi: well, imagine you try to set something to rank 5 in serial... that would be silently ignored

RiccardoRossi:
well, we will need to solve those... forward declaration is really a dangerous thing ...

loumalouomega:
> well, we will need to solve those... forward declaration is really a dangerous thing ...

When it is dangerous?, if a same class is defined diferently in different files?, a priori you will have linking problems

RiccardoRossi:
for example the dependency check mechanism (includes) gets broken by forward declarations

loumalouomega:
> for example the dependency check mechanism (includes) gets broken by forward declarations

I think if you what are you doing should not be problematic, and we approve changes little by little

RiccardoRossi:
well, i ping @KratosMultiphysics/technical-committee ... but i anticpate you that i am completely agains this

loumalouomega:
> well, i ping @KratosMultiphysics/technical-committee ... but i anticpate you that i am completely agains this

I understand, but @roigcarlo also found due to this PR problems related with cross dependencies difficult to solve without this technique

PR #9117, opened 2021-09-10
[Core] adding vars needed from processes

ddiezrod:
**Description**
New attempt to try to add variables in the solver that are required from the processes. For the moment it's basically the same that @philbucher implemented here: https://github.com/KratosMultiphysics/Kratos/pull/3561

I added an example of how we could use it in apply_inlet_process.py

I thought about the possibility of also specifying the model part as it was discussed in the mentioned PR. It worries me a little if the process will always know the correct model part where they should be added and also if this model part will be already created at that moment... At the end the solver will add these variables to its computing model part so I guess its not so bad...

PR #11764, opened 2023-11-03
[Core] Enabling the possibility of adding variables from processes

ddiezrod:
**📝 Description**
This is more or less the same as https://github.com/KratosMultiphysics/Kratos/pull/9117 but using the registry (so it works in cpp and python)

The idea is to use the registry to obtain the variables that a process needs to be added to a model part at the very beginning before model parts are even imported. This improves a lot the modularity as you will not need to always add this variable in the solver.

I adapted apply_inlet_process from FluidDynApp to use this new capability.


kratos/processes/process.h
```diff
@@ -182,6 +184,13 @@ class Process : public Flags
         return default_parameters;
     }
 
+    static HistoricalVarsMap GetHistoricalVariables(
```
loumalouomega, line 194: Doc

kratos/processes/process.h
```diff
@@ -51,6 +51,8 @@ class Process : public Flags
     ///@name Type Definitions
     ///@{
 
+    using HistoricalVarsMap = std::unordered_map<std::string,std::vector<std::string>>;
```
loumalouomega, line 55: Doc

loumalouomega:
I will wait for test foor final review

rubenzorrilla:
Great. As it involves changing a few crucial classes I'll add it for the next @KratosMultiphysics/technical-committee discussion. Nevertheless, at first glance everything makes sense for me.

kratos/processes/process.h
ddiezrod: done

kratos/processes/process.h
ddiezrod: this is just a definition, what documentation you want here?

loumalouomega: For example: /// The historical variables map. Hash is a variable defined as a string, and it contains a list of variables (vector of strings)
```suggestion
    /// The historical variables map. Hash is a variable defined as a string, and it contains a list of variables (vector of strings)
```

ddiezrod: you love comments too much... : P

ddiezrod:
Test added.

loumalouomega (dismissed):
Looks Okay to me

ddiezrod:
@loumalouomega lucky me, there was already a test with exactly same name in ContactStructApp, I have changed the name there because I think it is more clear that way

loumalouomega (commented):
Okay form my side

kratos/python_scripts/process_factory.py
```diff
@@ -18,24 +19,7 @@ def ConstructListOfProcesses( self, process_list ):
...
-                            #TODO: We should enforce this. Now an error happens but as we populate we should throw a warning and search for a ClassName item
-                            err_msg = f"The '{class_name}' class name cannot be found within the '{module_name}' module."
-                            raise Exception(err_msg)
-                    else:
-                        err_msg = f"Registry process '{registry_entry}' cannot be constructed."
-                        raise Exception(err_msg)
+                    prototype = self._GetPrototype(registry_entry)
                     # Construct the process from the obtained prototype and append it to the list
                     instance = prototype(self.Model, item["Parameters"])
                     constructed_processes.append(instance)
```
RiccardoRossi, line 25: i would say there is a problem with this: 

a process may need a variable to be in the modelpart already at construction time (i am afraid there were some examples of this).

for this reason one cannot use a "real" constructor of the process to get a prototype on the top of which we could get the function, without having constructed the real process.
What i am trying to say that the GetVariable should be a function similar to the Create one

I am cc-ing  @KratosMultiphysics/technical-committee since this was a thing we discussed.

RiccardoRossi (changes requested):
i must block since i believe this is incorrect in the case a process uses a historical variable already in its construction process.

kratos/python_scripts/process_factory.py
rubenzorrilla: > i would say there is a problem with this:
> 
> a process may need a variable to be in the modelpart already at construction time (i am afraid there were some examples of this).
> 
> for this reason one cannot use a "real" constructor of the process to get a prototype on the top of which we could get the function, without having constructed the real process. What i am trying to say that the GetVariable should be a function similar to the Create one
> 
> I am cc-ing @KratosMultiphysics/technical-committee since this was a thing we discussed.

Though you're probably right, it would be helpful to locate those. The point is that maybe these "special" processes can be refactored to avoid the variable requirement at construction time, something that I think is very unnatural.

ddiezrod: Any advance on this, or do I need to find another solution on our side?

rubenzorrilla: Unfortunately not... It is still in our todo list for the next KTC meeting. Sorry about this.

pooyan-dadvand:
@KratosMultiphysics/technical-committee had a discussion about this and the main question is if we should go for the static member our virtual one. Here are pros and cons:
Static:
- Prototype would be optional
- No need to implement the default constructor in all Processes
- It needs registry for dispatching the method (That can check if is with prototype or direct use of static function)
- Having the common implementation in the base class is not possible (To be done in a `template<ClassType, BaseType>` utility)
- `GetDefaultVariables` is virtual and we cannot call it. The idea is to have `Specification` as static member and `GetDefaultVariable` uses that as a default implementation

Virtual:
- More flexible in hierarchy
- No need for registry involvement
- It needs default constructor for all processes (Several processes has reference to the modepart which should be changed to pointer)

rubenzorrilla:
Note for myself: after (re-)discussing this in the @KratosMultiphysics/technical-committee we decided to basically mimic what is in the `GetSpecifications` of the element but making it `static`.

philbucher:
> Note for myself: after (re-)discussing this in the @KratosMultiphysics/technical-committee we decided to basically mimic what is in the `GetSpecifications` of the element but making it `static`.

static methods cannot be virtual :/

rubenzorrilla:
> > Note for myself: after (re-)discussing this in the @KratosMultiphysics/technical-committee we decided to basically mimic what is in the `GetSpecifications` of the element but making it `static`.
> 
> static methods cannot be virtual :/

Yes... this is the price we will be paying.

philbucher:
@rubenzorrilla I dont understand the technical solution
What you propose does not work in C++

RiccardoRossi:
this is currently being discussed in  #12506, @ddiezrod can you please comment there?

ddiezrod:
> this is currently being discussed in #12506, @ddiezrod can you please comment there?

ok, Ill close this

PR #11765, opened 2023-11-03
[GeoMechanicsApplication] Fixed a bug related to handling of conditions

avdg81:
**📝 Description**
We have found that conditions were only added to the computational model part, but never removed from it. In multi-stage analysis, that could result in adding inactive loads from previous stages. That has been corrected.

**🆕 Changelog**
Other changes include:
- Added a simple regression test with two uniform line loads, where each one is active in its own stage.
- Added a static method to Python class `GiDOutputFileReader`, which returns a list of nodal result values at the given time. Optionally, it filters its result given a sequence of node IDs.
- Use `math.isclose` to test whether two given times are nearly equal rather than using `==`.
- Removed some duplicated conditions that had been created by GiD.
- Updated the expected results of the settlement work flow test.
- Added a few nodal results that can be written by class `GeoOutputWriter`: reaction forces, normal contact stresses, and tangential contact stresses.


applications/GeoMechanicsApplication/tests/line_load_tests/line_loads_in_stages/README.md
```diff
@@ -0,0 +1,19 @@
...
+which have displacement degrees of freedom as well as pore water pressure
+degrees of freedom.  The left and right sides of the domain cannot move
+horizontally, whereas the bottom edge has been completely fixed.  Along the top
+edge, uniform vertical line loads are applied.  In the first stage, the line
+load (applied to model part `First_line_load`) has a magnitude of -10.0 N/m.
+In the second stage, the line load (applied to model part `Second_line_load`)
+has a magnitude of -20.0 N/m.  Even though the line loads are associated with
+different model parts, both of them reference the same set of nodes.  Since the
+top edge is fully loaded, the total vertical reactions are expected to be equal
+to 5.0 m * -10.0 N/m = -50.0 N and 5.0 m * -20.0 N/m = -100.0, respectively,
```
WPK4FEM: -100.0 N ( missing unit )

applications/GeoMechanicsApplication/tests/test_settlement_workflow/test_model.mdpa
```diff
@@ -3144,15 +3144,6 @@ Begin Conditions LineNormalLoadDiffOrderCondition2D3N
   27  1  153 183 166
   28  1  55 88 70
   29  1  132 153 142
-  30  2  119 88 102
```
WPK4FEM, line 3147: Good find. Load only the remaining part.

WPK4FEM (dismissed):
Tiny improvement is adding the forgotten unit in the README file

applications/GeoMechanicsApplication/tests/line_load_tests/line_loads_in_stages/README.md
avdg81: Good catch. Done.

applications/GeoMechanicsApplication/tests/test_settlement_workflow/test_model.mdpa
avdg81: Thanks.

WPK4FEM (dismissed):

applications/GeoMechanicsApplication/custom_workflows/write_output.cpp
rfaasse, line 1: (I also forgot in my last PR) Is this a good moment to rename the file to something like geo_output_writer.cpp/h?

applications/GeoMechanicsApplication/tests/test_helper.py
```diff
@@ -529,11 +530,32 @@ def _strip_off_quotes(self, quoted_string):
...
                 return time_results["values"]
 
     @staticmethod
     def get_value_at_node(node, time_results):
         for node_results in time_results:
             if node_results["node"] == node:
                 return node_results["value"]
+
+    @staticmethod
+    def nodal_values_at_time(result_item_name, time, output_data, node_ids=None):
```
rfaasse, line 542: It seems that this function has some overlap with the get_values_at_time and get_value_at_node. Do you think we could replace the older functions with this more elaborate one?

rfaasse (dismissed):
Clear PR, good find! Just two (non-blocking) comments.

applications/GeoMechanicsApplication/tests/test_helper.py
avdg81: Good point! Yes, I think we can. However, I would propose to do that in a separate PR.

applications/GeoMechanicsApplication/custom_workflows/write_output.cpp
avdg81: Good suggestion. Done.

rfaasse (approved):

PR #11767, opened 2023-11-03
[Core] Define and use `PointLocalCoordinatesTetrahedra3D4N` in `Tetrahedra3D10` when planar face

loumalouomega:
**📝 Description**

Define and use `PointLocalCoordinatesTetrahedra3D4N` in `Tetrahedra3D10` when planar face, for faster computation (explicit formula instead of iterative) and avoid non covergence.

**🆕 Changelog**

- [Define the PointLocalCoordinatesTetrahedra3D4N](https://github.com/KratosMultiphysics/Kratos/commit/00ae3649eb1debd8a34723a19672a748fce71a5c)
- [Replace PointLocalCoordinatesTetrahedra3D4N in Tetrahedra3D4N](https://github.com/KratosMultiphysics/Kratos/commit/fcc66ef8fdb214962eca58776c5cf0e2684d586e)
- [Use `PointLocalCoordinatesTetrahedra3D4N` in `Tetrahedra3D10` when planar face](https://github.com/KratosMultiphysics/Kratos/commit/35a65207717d58be1ccaaa3a1e8958a75faa035b)

### DIFF
--- a/kratos/utilities/geometry_utilities.h
+++ b/kratos/utilities/geometry_utilities.h
@@ -56,6 +56,114 @@ public:
     static std::string GetGeometryName(const GeometryData::KratosGeometryType TypeOfGeometry);
 
     /**
+     * @brief Returns the local coordinates of a given arbitrary point for a given linear tetrahedra
+     * @details Based on https://www.colorado.edu/engineering/CAS/courses.d/AFEM.d/AFEM.Ch09.d/AFEM.Ch09.pdf. Section 9.1.6
+     * @param rGeometry The geometry to be considered
+     * @param rResult The vector containing the local coordinates of the point
+     * @param rPoint The point in global coordinates
+     * @return The vector containing the local coordinates of the point
+     */
+    template<class TGeometryType>
+    static inline typename TGeometryType::CoordinatesArrayType& PointLocalCoordinatesTetrahedra3D4N(
+        const TGeometryType& rGeometry,
+        typename TGeometryType::CoordinatesArrayType& rResult,
+        const typename TGeometryType::CoordinatesArrayType& rPoint
+        )
+    {
+        // Compute RHS
+        array_1d<double,4> X;
+        X[0] = 1.0;
+        X[1] = rPoint[0];
+        X[2] = rPoint[1];
+        X[3] = rPoint[2];
+
+        // Auxiliary coordinates
+        const auto& r_coordinates_0 = rGeometry[0].Coordinates();
+        const auto& r_coordinates_1 = rGeometry[1].Coordinates();
+        const auto& r_coordinates_2 = rGeometry[2].Coordinates();
+        const auto& r_coordinates_3 = rGeometry[3].Coordinates();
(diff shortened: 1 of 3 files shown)


ddiezrod (dismissed):
thanks!

kratos/utilities/geometry_utilities.h
```diff
@@ -55,6 +55,114 @@ class KRATOS_API(KRATOS_CORE) GeometryUtils
...
+    /**
+     * @brief Returns the local coordinates of a given arbitrary point for a given linear tetrahedra
+     * @details Based on https://www.colorado.edu/engineering/CAS/courses.d/AFEM.d/AFEM.Ch09.d/AFEM.Ch09.pdf. Section 9.1.6
+     * @param rGeometry The geometry to be considered
+     * @param rResult The vector containing the local coordinates of the point
+     * @param rPoint The point in global coordinates
+     * @return The vector containing the local coordinates of the point
+     */
+    template<class TGeometryType>
+    static inline typename TGeometryType::CoordinatesArrayType& PointLocalCoordinatesTetrahedra3D4N(
```
rubenzorrilla: As far as I understand, you aim to use the simplex formula for any tetrahedron featuring planar faces. If so, I think the name of the method is misleading as you make explicit `Tetrahedra3D4N` while the argument accepts any tetrahedron.

Having said this, I'd suggest something like `PointLocalCoordinatesPlanarFaceTetrahedra` or similar.

kratos/utilities/geometry_utilities.h
```diff
@@ -55,6 +55,114 @@ class KRATOS_API(KRATOS_CORE) GeometryUtils
...
+        const TGeometryType& rGeometry,
+        typename TGeometryType::CoordinatesArrayType& rResult,
+        const typename TGeometryType::CoordinatesArrayType& rPoint
+        )
+    {
+        // Compute RHS
```
rubenzorrilla, line 76: At least in debug, I'd check that `rGeometry` is of tetrahedra type.

kratos/utilities/geometry_utilities.h
loumalouomega: Precisely I changed the name specifying that is just for the simplex, when using in higher order tetrahedra it will work only when the higher order element is geometrically a simplex

kratos/utilities/geometry_utilities.h
loumalouomega: Okay

rubenzorrilla (commented):
Two minor comments. The rest looks good to me.

loumalouomega:
> Two minor comments. The rest looks good to me.

Done

kratos/utilities/geometry_utilities.h
rubenzorrilla: Fair enough. Indeed, usages of this must be through the geometry API so the naming shouldn't be a problem at all. In any case, thanks for taking my suggestion into consideration.

rubenzorrilla (approved):
Approving again.

PR #11777, opened 2023-11-08
[Core] Remove deprecated warning

loumalouomega:
**📝 Description**

Remove deprecated warning

**🆕 Changelog**

- [Remove deprecated warning](https://github.com/KratosMultiphysics/Kratos/commit/28c2312007c3110d1acdf2be32e7cdced01bd392)


philbucher (changes requested):
That is even worse than using the default, why not using Testing::GetDefaultDataComm?

loumalouomega:
> That is even worse than using the default, why not using Testing::GetDefaultDataComm?

Because I din't know thatw as the alternative

philbucher (approved):

PR #5187, opened 2019-07-05
[CoSim] minor updates

philbucher:
This PR does some cleanup in the CoSimApp:
- Adding class-descriptions for the base-classes
- Making interface of the ConvAcc consistent with the FSIApp (not yet fully compatible, bcs the FSIApp-acc do the update in-place, whereas we currently return the update. This will be changed in a future PR)
- Other minor things

related to #5140

PR #11780, opened 2023-11-09
[CoSim] Adding a strong block solver and block MVQN acceleration

azzeddinetiba:
**📝 Description**
If I understand correctly, doing a block QN acceleration for CoSim coupling (using the PYTHON interface) is not possible as things are at the moment, since the acceleration is only applied after a full _F_(_S_(**f**)) ( Or _S_(_F_(**u**)) ) computation with the current `coupled_solvers`.
This is a suggestion to do that, with a block MVQN for example.
The idea is that the `ConvergenceAcceleratorWrapper` handles two `CouplingInterfaceData` instead of just one.

I realised ( probably too late :sweat_smile: ) that you aim to unify the accelerator interface [with the ones in the Core](https://github.com/KratosMultiphysics/Kratos/projects/32#card-47954065) or [the ones in FSI](https://github.com/KratosMultiphysics/Kratos/blob/master/applications/FSIApplication/custom_utilities/mvqn_convergence_accelerator.hpp#L57) (#5187) .. ? In which case this PR wouldn't be of help (?)

**🆕 Changelog**
- Added block coupled solver
- Added MVQN convergence acceleration in CoSim


azzeddinetiba:
An example would be to change the config file on [fsi_mok](https://github.com/KratosMultiphysics/Examples/tree/master/co_simulation/validation/fsi_mok) to

```json
...

    "solver_settings" :
    {
        "type" : "coupled_solvers.block_solver",
        "echo_level" : 1,
        "num_coupling_iterations" : 12,
        "convergence_accelerators" : [
            {
                "type"      : "blockMvqn",
                "solver_sequence" :
                [
                    {
                        "solver": "structure",
                        "data_name"  : "disp",
                        "coupled_data_name" : "load"
                    },
                    {
                        "solver": "fluid",
                        "data_name"  : "load",
                        "coupled_data_name" : "disp"
                    }
                ]
                // "type"      : "mvqn",
                // "solver"    : "fluid",
                // "data_name" : "disp"
            }
        ],
        "convergence_criteria" : [
            {
                "type"          : "relative_norm_previous_residual",
                "solver"        : "fluid",
                "data_name"     : "load",
                "abs_tolerance" : 1e-8,
                "rel_tolerance" : 1e-8
            }
        ],

...

        "coupling_sequence":
        [
            {
                "name": "structure",
                "input_data_list": [
                    {
                        "data"             : "load",
                        "from_solver"      : "fluid",
                        "from_solver_data" : "load",
                        "data_transfer_operator" : "mapper",
                        "data_transfer_operator_options" : ["swap_sign", "use_transpose"]
                    }
                ],
                "output_data_list": []
            },
            {
                "name": "fluid",
                "input_data_list"  : [
                    {
                        "data"             : "disp",
                        "from_solver"      : "structure",
                        "from_solver_data" : "disp",
                        "data_transfer_operator" : "mapper"
                    }
                ],
                "output_data_list" : [
                    {
                        "data"           : "load",
                        "to_solver"      : "structure",
                        "to_solver_data" : "load",
                        "data_transfer_operator" : "mapper",
                        "data_transfer_operator_options" : ["swap_sign", "use_transpose"]
                    }
                ]
            }
        ],

...
```

.gitignore
```diff
@@ -86,3 +86,6 @@ remote_*.md
 
 # External downloaded libs (TPL downloaded libs, etc...)
 external_libraries/tetgen/*
+
+# Mac folder info files
+*.DS_Store
```
philbucher: this should be done in a separate PR

applications/CoSimulationApplication/python_scripts/convergence_accelerators/blockMvqn.py
```diff
@@ -0,0 +1,128 @@
...
+
+        blockJacobian = (np.eye(row) - self.J_hat[data_name] @ self.J_hat[coupled_data_name])
+        b = r - self.J_hat[data_name] @ yResidual
+
+        return np.linalg.solve( blockJacobian, b )
+
+    def FinalizeSolutionStep( self ):
+
+        ## Assign J=J_hat
+        for _, data_name in enumerate(self.J.keys()):
```
philbucher:
```suggestion
        for data_name in self.J:
```

applications/CoSimulationApplication/python_scripts/convergence_accelerators/convergence_accelerator_wrapper.py
```diff
@@ -20,7 +20,20 @@ def __init__(self,
                  settings: KratosMultiphysics.Parameters,
                  interface_data_dict: "dict[str,CouplingInterfaceData]",
                  parent_coupled_solver_data_communicator: KratosMultiphysics.DataCommunicator):
-        self.interface_data = interface_data_dict[settings["data_name"].GetString()]
+        if settings.Has("solver_sequence"):
+            self.interface_data_dict = interface_data_dict
+            self.solver_names_sequence = []
+            self.coupl_data_names = {}
+            self.input_data = {}
+            for _, data_name in enumerate(self.interface_data_dict.keys()):
```
philbucher:
```suggestion
            for data_name in self.interface_data_dict:
```

applications/CoSimulationApplication/python_scripts/convergence_accelerators/convergence_accelerator_wrapper.py
```diff
@@ -31,16 +44,36 @@ def __init__(self,
...
+                self.gather_scatter_required = self.interface_data.IsDistributed() and not conv_acc_supports_dist_data
+                if self.gather_scatter_required:
+                    self.data_comm = self.interface_data.GetModelPart().GetCommunicator().GetDataCommunicator()
+                    self.sizes_from_ranks = np.cumsum(self.data_comm.GatherInts([self.interface_data.Size()], 0))
+        else:
+            self.executing_rank = {}
+            self.gather_scatter_required = {}
+            self.data_comm = {}
+            self.sizes_from_ranks = {}
+            for _, data_name in enumerate(interface_data_dict.keys()):
```
philbucher:
```suggestion
            for data_name in interface_data_dict:
```

applications/CoSimulationApplication/python_scripts/convergence_accelerators/convergence_accelerator_wrapper.py
```diff
@@ -31,16 +44,36 @@ def __init__(self,
...
+                    if self.gather_scatter_required[data_name]:
+                        self.data_comm[data_name] = self.interface_data_dict[data_name].GetModelPart().GetCommunicator().GetDataCommunicator()
+                        self.sizes_from_ranks[data_name] = np.cumsum(self.data_comm[data_name].GatherInts([self.interface_data_dict[data_name].Size()], 0))
 
     def Initialize(self):
         self.conv_acc.Initialize()
+        if self.isBlockAccelerator:
+            self.prev_input_data = {}
+            self.output_data = {}
+            for _, data_name in enumerate(self.interface_data_dict.keys()):
```
philbucher: same

applications/CoSimulationApplication/python_scripts/convergence_accelerators/convergence_accelerator_wrapper.py
```diff
@@ -52,17 +85,23 @@ def FinalizeSolutionStep(self):
...
 
     def InitializeNonLinearIteration(self):
-        if self.interface_data.IsDefinedOnThisRank():
-            # Saving the previous data for the computation of the residual
-            # and the computation of the solution update
-            self.input_data = self.interface_data.GetData()
+        # Saving the previous data for the computation of the residual
+        # and the computation of the solution update
+        if self.isBlockAccelerator:
+            for _, data_name in enumerate(self.interface_data_dict.keys()):
```
philbucher: same

applications/CoSimulationApplication/python_scripts/convergence_accelerators/convergence_accelerator_wrapper.py
philbucher, line 1: conceptual question, do you intend on adding more accelerators of this type?
If not, can you more the modifications of this file to `block_mvqn.py`?

applications/CoSimulationApplication/python_scripts/convergence_accelerators/blockMvqn.py
philbucher, line 1: file name should be `block_mvqn.py`

philbucher (commented):
overall I am fine with this addition, I dont think it would block the unification of the interfaces with the Core/FSIApp

But I have some conceptual questions

applications/CoSimulationApplication/python_scripts/convergence_accelerators/convergence_accelerator_wrapper.py
azzeddinetiba: Yes, I think [the IBQNLS](https://doi.org/10.1016/j.compstruc.2006.11.006) could be very helpful too. It's coming, do you think it should be in another PR or this one is fine ?

applications/CoSimulationApplication/python_scripts/convergence_accelerators/block_mvqn.py
```diff
@@ -0,0 +1,128 @@
...
+        self.alpha = self.settings["alpha"].GetDouble()
+        self.epsilon = self.settings["epsilon"].GetDouble()
+
+        self.X_tilde = {}
+        self.X = {}
+        self.J = {}
+        self.J_hat = {}
+        self.coupl_data_names = {}
+
+        for solverData in settings["solver_sequence"].values():
```
philbucher:
```suggestion
        for solver_data in settings["solver_sequence"].values():
```

applications/CoSimulationApplication/python_scripts/convergence_accelerators/convergence_accelerator_wrapper.py
```diff
@@ -20,7 +20,20 @@ def __init__(self,
...
                  parent_coupled_solver_data_communicator: KratosMultiphysics.DataCommunicator):
-        self.interface_data = interface_data_dict[settings["data_name"].GetString()]
+        if settings.Has("solver_sequence"):
+            self.interface_data_dict = interface_data_dict
+            self.solver_names_sequence = []
+            self.coupl_data_names = {}
+            self.input_data = {}
+            for data_name in self.interface_data_dict:
+                self.solver_names_sequence.append(self.interface_data_dict[data_name].solver_name)
+            self.isBlockAccelerator = True
```
philbucher:
```suggestion
            self.is_block_accelerator = True
```

philbucher: also please initialize this to `False`, or `None`, depending on what you think is more appropriate

applications/CoSimulationApplication/python_scripts/convergence_accelerators/ibqnls.py
```diff
@@ -0,0 +1,198 @@
...
+        self.previousR = R.copy()
+        if isFirstiter: # V only needs to be saved completely on the first iteration
+            self.previousV = V.copy()
+        else: # Otherwise it can be retrieved from V_new and V_old and memory is freed
+            self.previousV = None
+
+        return delta_x
+
+    def _augmentedMatrix(self, mat, oldMat, isFirstDt):
+        if isFirstDt:
```
philbucher:
```suggestion
        if is_first_dt:
```

applications/CoSimulationApplication/python_scripts/convergence_accelerators/convergence_accelerator_wrapper.py
philbucher: It seems like there is an `if` in basically every member function. Therefore, can you please make this a separate class? => e.g. `BlockConvergenceAcceleratorWrapper`

=> then you can add a check in `applications/CoSimulationApplication/python_scripts/factories/helpers.py::CreateConvergenceAccelerators`, sth like `if name.startwith("block_")` to construct the correct wrapper

applications/CoSimulationApplication/python_scripts/coupled_solvers/block_solver.py
philbucher, line 1: how about making clear that this is a strongly coupled? => `BlockGaussSeidelStrongCoupledSolver`?

applications/CoSimulationApplication/python_scripts/factories/helpers.py
```diff
@@ -34,8 +34,15 @@ def CreateConvergenceAccelerators(convergence_accelerator_settings_list: KM.Para
...
-        interface_data_dict = solver.data_dict
+        if conv_acc_settings.Has("solver"):
+            solver = solvers[conv_acc_settings["solver"].GetString()]
+            interface_data_dict = solver.data_dict
+        else:
+            interface_data_dict = {}
+            for sequenceData in conv_acc_settings["solver_sequence"].values():
+                solver = solvers[sequenceData["solver"].GetString()]
+                data_name = sequenceData["data_name"].GetString()
+                interface_data_dict[data_name] = solver.data_dict[data_name]
```
philbucher: this becomes problematic if more things are added, thus my suggestion above to make the check more explicit

philbucher (commented):
Apologies for the delay. Here are some more comments, style as well as conceptual:
- Please use python-style for variable names whenever possible => `snake_case`
- Please separate the wrapper for the block-accelerators, the code becomes quite messy otherwise
- Generally I think this is a valuable addition, thanks!
- Can you please add some tests?
    - For the wrapper I recommend doing similar tests to what I did for the current wrapper
    - For the accelerators ideally you would add a [test like this one in the FSI-App](https://github.com/KratosMultiphysics/Kratos/blob/master/applications/FSIApplication/tests/convergence_accelerator_spring_test.py). This tests the accelerators independently. I know that I never got to add it for the other accelerators, so this is optional. Maybe @rubenzorrilla can comment/recommend sth.
    - What about adding the mok-fsi case also for both of the new accelerators? This si a pretty small test case that prevent regressions, it runs an entire strongly coupled FSI simulation super fast

applications/CoSimulationApplication/python_scripts/convergence_accelerators/convergence_accelerator_wrapper.py
```diff
@@ -115,6 +247,17 @@ def __init__(self,
...
+def CreateBlockResidualComputation(settings: KratosMultiphysics.Parameters,
+                                interface_data_dict: "dict[str,CouplingInterfaceData]"):
+    residual_computation_type = "data_difference"
+    if settings.Has("residual_computation"):
+        residual_computation_type = settings["residual_computation"]["type"].GetString()
+
+    if residual_computation_type == "data_difference":
+        return DataDifferenceBlockResidual(settings, interface_data_dict)
+    else:
+        raise Exception('The specified residual computation "{}" is not available!'.format(residual_computation_type))
```
philbucher:
```suggestion
        raise Exception(f'The specified residual computation "{residual_computation_type}" is not available!'))
```

please always use f-strings, modern python has no place for `format` any more

philbucher (commented):
in terms of code it looks good now, thanks for the effort!

Will you add tests?

azzeddinetiba:
> in terms of code it looks good now, thanks for the effort!
> 
> Will you add tests?

I still didn't take a look at the convergence wrapper tests.
For the fsi_mok test I think I can add one in a few moments. I'm thinking smth similar to [the aitken one](https://github.com/KratosMultiphysics/Kratos/blob/cf0f69a91c3ff10f900e14add49354bd25d648d4/applications/CoSimulationApplication/tests/test_mok_fsi.py#L34) is sufficient ? without the external solver part ?

applications/CoSimulationApplication/python_scripts/coupled_solvers/block_solver.py
azzeddinetiba: Actually, I think I'll rename it to `BlockIterativeStrongCoupledSolver` since it's not really a "Gauss-Seidel" per se

applications/CoSimulationApplication/python_scripts/coupled_solvers/block_gauss_seidel_strong.py
philbucher, line 1: the name Gauss Seidel is more consistent with the rest of the application, why is your reasoning for changing it?

azzeddinetiba: There is no "technical" reason, it's just consistent with how I understand the two solvers mathematically. Since -for me- a Gauss Seidel solver solves the $\mathcal{S}(\mathcal{F}(\pmb{u}))$ so the unknown (displacement or force) passes unchanged to the other (sub)solver. Whether with a block solver solves

$$ \begin{bmatrix} 
   \mathcal{S}\circ\mathcal{F} & 0 \\
   0 & \mathcal{F}\circ\mathcal{S} \\
\end{bmatrix}  \begin{bmatrix} 
  u \\
   f \\
\end{bmatrix} =  \begin{bmatrix} 
  0 \\
   0 \\
\end{bmatrix} $$

so it allows accelerating the two (or more) (sub)solver outputs. So I think that might be confusing for users familiar with these notations. What do you think ?

azzeddinetiba:
@philbucher is it possible to run a CoSim problem on MPI while the mapper is on serial compilation ?
Otherwise I can't test on my local machine, I can't compile Trilinos .. #11900

applications/CoSimulationApplication/tests/test_mok_fsi.py
```diff
@@ -42,6 +43,32 @@ def test_mok_fsi_aitken(self):
...
+            self.__ManipulateCFDSettings()
+            self.__RemoveOutputFromCFD() # comment to get output
+            self.__AddTestingToCFD()
+            self.__DumpUpdatedCFDSettings()
+            self._runTest()
+
+
+    def test_mok_fsi_block_ibqnls(self):
+        self.accelerator_type = "block_ibqnls"
+        self.err_tol = "6e-5"
```
azzeddinetiba, line 60: When I test on multi threads, the error is a little larger than 1e-6, the results differ each time the the test is run. It's probably linked to the scipy operations (since it's larger than the error with `block_mvqn`) in the accelerator but I can't figure out where exactly. The results are the same when `OMP_NUM_THREADS=1`

philbucher: interesting. I think the tolerance is still small enough, maybe you will find the problem in the future. For me no reason to block

philbucher:
> @philbucher is it possible to run a CoSim problem on MPI while the mapper is on serial compilation ? Otherwise I can't test on my local machine, I can't compile Trilinos .. #11900

Unfortunately not. I tried to implement this at some point, but decided it was not worth the effort, esp since we have Kratos-native matrices which can be used to completely replace trilinos

applications/CoSimulationApplication/python_scripts/coupled_solvers/block_gauss_seidel_strong.py
philbucher: ok I (think I) understand

In that case, can you please rename it to `block_strong.py`? `iterative` and `strong` are the same and thus redundant.

applications/CoSimulationApplication/tests/test_mok_fsi.py
philbucher: if you are sure which operation it is, then we could dynamically set the omp-threads to 1 for just this function, what do you think?

azzeddinetiba: I can't figure out which operation specifically, but my guess it's the combination of `sp.linalg.solve_triangular` and `sp.sparse.linalg.gmres`. But it's the "pseudo-inverse" operation itself and not just the functions, because I tested with a brute inversion of matrices and the error still varies with the same magnitude.
Setting omp-threads to 1 is fine too. But the error tolerance should still stay bigger than the other accelerators, because of the difference of results between different OS.

philbucher: Ok

FYI if you want to limit this accelerator to always use one thread, you could use sth like I did here:
https://github.com/KratosMultiphysics/Kratos/blob/c252d518b6f7bb441322e46daded6708a9263260/applications/CoSimulationApplication/python_scripts/solver_wrappers/kratos/kratos_base_wrapper.py#L16-L29

philbucher (approved):
looks good to me, thanks for the good work (and the patience)

azzeddinetiba:
Hi, can this be merged ?

philbucher:
sure
I approved months ago :)

matekelemen:
@philbucher, seems like @azzeddinetiba is not a member of KratosMultiphysics and so cannot merge the PR, even if you approved it. Can you please merge it?

rubenzorrilla:
@philbucher I took the freedom so we don't miss it again.

philbucher:
Thanks

Dont wait for me pls, I might or might not see things 😉

azzeddinetiba:
Thank you ❤️

PR #11782, opened 2023-11-09
[GeoMechanicsApplication] Add new schemes for thermal calculations

### LINKED ISSUE #11769: [GeoMechanicsApplication] Add new scheme for the thermal process
Branch: https://github.com/KratosMultiphysics/Kratos/tree/geo/10372-HeatTransfer-temp

**Investigate whether this can be unified with U-Pw and Pw schemes.**

A new scheme for the thermal process is added
applications/GeoMechanicsApplication/custom_strategies/schemes/backward_euler_quasistatic_T_scheme.hpp
applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_T_scheme.hpp
Whole files

Registered the srategies for python.
applications\GeoMechanicsApplication\custom_python\add_custom_strategies_to_python.cpp
Lines: 62, 66, 90 to 92, 102 to 104

#### Acceptance Criteria
**Given** the BackwardEulerQuasistaticTScheme
**When** the UpdateVariablesDerivatives function is called
**Then** the DT_TEMPERATURE solution step value is updated as delta T / delta time

rfaasse:
**📝 Description**
This PR aims to implement two new schemes for thermal calculations. In doing so, new unit tests are created and some refactoring is done to avoid unwanted inheritance between classes.

**🆕 Changelog**
- Two new thermal schemes are added
- A base class is introduced, which contains all variable-independent functionality of the UPW scheme
- Unit tests for all involved classes are added


applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+        return 0;
+
+        KRATOS_CATCH("")
+    }
+
+    void CheckAllocatedVariables(const ModelPart& rModelPart) const
+    {
+        for (const auto& r_node : rModelPart.Nodes()) {
+            KRATOS_ERROR_IF(!r_node.SolutionStepsDataHas(TEMPERATURE))
```
markelov208: The variable has name() method to provide the variable name. From my viewpoint, it helps to generalize such a check. Would you like to create a method for checking? I introduced a method in my PR #11753

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,349 @@
...
+                    Element::DofsVectorType& rDofList,
+                    const ProcessInfo& rCurrentProcessInfo) override
+    {
+        const bool isActive = (rElement.IsDefined(ACTIVE)) ? rElement.Is(ACTIVE) : true;
+        if (isActive)
+            rElement.GetDofList(rDofList, rCurrentProcessInfo);
+    }
+
+    void GetDofList(const Condition& rCondition,
+                    Element::DofsVectorType& rDofList,
```
markelov208: There is Condition::DofsVector which is the same as Element::DofsVector. I think it is better to use only Conditions or only Elements. From my viewpoint the best way is to keep DofsVector separately, not in Element nor Condition. What do you think?

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,349 @@
...
+                    const ProcessInfo& rCurrentProcessInfo) override
+    {
+        const bool isActive = (rCondition.IsDefined(ACTIVE)) ? rCondition.Is(ACTIVE) : true;
```
markelov208: This line is repeated many times. Would it be better to create a method instead of the line?

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,349 @@
...
+        if (isActive)
+            rCondition.GetDofList(rDofList, rCurrentProcessInfo);
+    }
+
+    void EquationId(const Element& rElement,
+                    Element::EquationIdVectorType& rEquationId,
+                    const ProcessInfo& rCurrentProcessInfo) override
+    {
+        const bool isActive = (rElement.IsDefined(ACTIVE)) ? rElement.Is(ACTIVE) : true;
```
markelov208: The same comment as above for Condition.

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,349 @@
...
+
+        this->UpdateVariablesDerivatives(rModelPart);
+
+        KRATOS_CATCH("")
+    }
+
+protected:
+    void CheckBufferSize(const ModelPart& rModelPart) const
+    {
+        KRATOS_ERROR_IF(rModelPart.GetBufferSize() < 2)
```
markelov208: I would create constexpr with a descriptive name instead of 2.

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_damped_U_Pw_scheme.hpp
```diff
@@ -118,7 +117,7 @@ class NewmarkQuasistaticDampedUPwScheme : public NewmarkQuasistaticUPwScheme<TSp
                          const ProcessInfo& CurrentProcessInfo)
     {
         // adding damping contribution
-        if (C.size1() != 0) noalias(LHS_Contribution) += (mGamma/(mBeta*mDeltaTime))*C;
+        if (C.size1() != 0) noalias(LHS_Contribution) += (mGamma/(mBeta*this->GetDeltaTime()))*C;
```
markelov208, line 120: This looks as VELOCITY_COEFFICIENT defined in the parent class, newmark_quasistatic_U_Ps_scheme.hpp. Can this coefficient be used here?

markelov208, line 120: C is too short name, would it be  better to use rDampingMatrix?

applications/GeoMechanicsApplication/tests/cpp_tests/test_backward_euler_T_scheme.cpp
```diff
@@ -0,0 +1,56 @@
...
+                          KratosGeoMechanicsFastSuite)
+{
+    BackwardEulerTScheme<SparseSpaceType, LocalSpaceType> scheme;
+    Model model;
+    auto& model_part = model.CreateModelPart("dummy", 2);
+
+    model_part.AddNodalSolutionStepVariable(TEMPERATURE);
+    model_part.AddNodalSolutionStepVariable(DT_TEMPERATURE);
+
+    const double current_temperature = 1.0;
```
markelov208: It is a matter of taste;  I prefer to use constexpr instead of const.

markelov208 (commented):
Richard, well done. I am very interested in learning about your tests.

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Agreed, I have some ideas to improve these functions more, up till now I just moved them, but this is a very good suggestion, done!

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Done!

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Good point, done!

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
rfaasse: Good suggestion, I added it in the base class, so we can start using it in other schemes as well 

(Using this function in the other schemes, I'm planning to do in a separate PR)

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_damped_U_Pw_scheme.hpp
rfaasse: I really want to refactor the schemes, so they re-use these kinds of functions where possible. I started in this [branch](https://github.com/KratosMultiphysics/Kratos/tree/geo/re-use-functionality-between-schemes)

However, since this is not related to the newly added temperature scheme, I'd propose to not do this as part of this PR.

rfaasse: Same as the previous comment, I'd like to keep this PR focused on the new scheme as much as possible, but will keep it in mind, because I agree this is not clearly named.

applications/GeoMechanicsApplication/tests/cpp_tests/test_backward_euler_T_scheme.cpp
rfaasse: It's not really in my system yet to work with constexpr, but I like the idea of compile time expressions, so I'll try to get it into my system and make the changes, thanks for the suggestion!

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Good catch, I at least made it consistent (so Condition& also uses Condition::DofsVectorType&), but at some point we should just have a header somewhere that defines these things once and once only.

rfaasse (commented):
Thanks for the review @markelov208! I implemented most suggestions (exceptions are two comments which I feel are out of scope for this PR).

markelov208 (dismissed):

applications/GeoMechanicsApplication/custom_strategies/schemes/backward_euler_T_scheme.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+protected:
+    inline void UpdateVariablesDerivatives(ModelPart& rModelPart) override
+    {
+        KRATOS_TRY
+
+        block_for_each(rModelPart.Nodes(), [this](Node& rNode) {
+            const double DeltaTemperature =
+                rNode.FastGetSolutionStepValue(TEMPERATURE) -
+                rNode.FastGetSolutionStepValue(TEMPERATURE, 1);
```
avdg81, line 40: I was a bit surprised to not see an index in the first term, but we do use an index of 1 in the second term. For clarity, it may be better to also use an index in the first term (0?).

applications/GeoMechanicsApplication/custom_strategies/schemes/backward_euler_T_scheme.hpp
```diff
@@ -0,0 +1,48 @@
...
+            rNode.FastGetSolutionStepValue(DT_TEMPERATURE) = DeltaTemperature / this->GetDeltaTime();
```
avdg81, line 41: Just a question: are we sure that we will never encounter a time increment of 0.0? (If no, we may risk division by zero.)

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
```diff
@@ -0,0 +1,104 @@
...
+    {
+    }
+
+    ~GeneralizedNewmarkTScheme() override = default;
+
+    int Check(const ModelPart& rModelPart) const override
+    {
+        KRATOS_TRY
+
+        Scheme<TSparseSpace, TDenseSpace>::Check(rModelPart);
```
avdg81, line 52: I would simplify this to (using one of the previously defined aliases):
```suggestion
        BaseType::Check(rModelPart);
```

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
```diff
@@ -0,0 +1,104 @@
...
+        CheckAllocatedVariables(rModelPart);
+        this->CheckBufferSize(rModelPart);
+
+        KRATOS_ERROR_IF(mTheta <= 0) << "Theta has an invalid value\n";
```
avdg81, line 56: What about a slightly better message like `"Theta must be positive, but got " << mTheta << std::endl`?

Furthermore, I would move this check to the constructor, to avoid that we can create objects with invalid `mTheta` values in the first place.

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+        KRATOS_CATCH("")
+    }
+
+    void CheckAllocatedVariables(const ModelPart& rModelPart) const
+    {
+        for (const auto& r_node : rModelPart.Nodes()) {
+            this->CheckSolutionStepsData(r_node, TEMPERATURE);
+            this->CheckSolutionStepsData(r_node, DT_TEMPERATURE);
+            this->CheckDof(r_node, TEMPERATURE);
```
avdg81, line 68: These checks appear to be very similar to the checks that have been implemented for the thermal element (see `TransientThermalElement::{CheckHasSolutionStepsDataFor, CheckHasDofsFor}`). We may want to reconsider these to avoid duplication of checks.

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+        this->SetDeltaTime(rModelPart.GetProcessInfo()[DELTA_TIME]);
+        rModelPart.GetProcessInfo()[DT_TEMPERATURE_COEFFICIENT] =
+            1.0 / (mTheta * this->GetDeltaTime());
+
+        KRATOS_CATCH("")
+    }
+
+private:
+    double mTheta = 0.0;
```
avdg81: The default value appears to be invalid (see one of the above checks). Let's try to think of a more sensible default value.

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+protected:
+    inline void UpdateVariablesDerivatives(ModelPart& rModelPart) override
+    {
+        KRATOS_TRY
+        block_for_each(rModelPart.Nodes(), [this](Node& rNode) {
+            const double delta_temperature =
+                rNode.FastGetSolutionStepValue(TEMPERATURE) -
+                rNode.FastGetSolutionStepValue(TEMPERATURE, 1);
+            const auto& previous_dt_temperature =
```
avdg81, line 80: It is very uncommon to use references to `const` for built-in types (e.g. `double`):
```suggestion
            const auto previous_dt_temperature =
```

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
```diff
@@ -0,0 +1,104 @@
...
+            this->CheckDof(r_node, TEMPERATURE);
+        }
+    }
+
+protected:
+    inline void UpdateVariablesDerivatives(ModelPart& rModelPart) override
+    {
+        KRATOS_TRY
+        block_for_each(rModelPart.Nodes(), [this](Node& rNode) {
+            const double delta_temperature =
```
avdg81, line 77: For consistency with the next variable, I would use `auto` here as well:
```suggestion
            const auto delta_temperature =
```

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
```diff
@@ -0,0 +1,104 @@
...
+template <class TSparseSpace, class TDenseSpace>
+class GeneralizedNewmarkTScheme
+    : public GeoMechanicsTimeIntegrationScheme<TSparseSpace, TDenseSpace> {
+public:
+    using BaseType = Scheme<TSparseSpace, TDenseSpace>;
+    using DofsArrayType = typename BaseType::DofsArrayType;
+    using TSystemMatrixType = typename BaseType::TSystemMatrixType;
+    using TSystemVectorType = typename BaseType::TSystemVectorType;
+    using LocalSystemVectorType = typename BaseType::LocalSystemVectorType;
+    using LocalSystemMatrixType = typename BaseType::LocalSystemMatrixType;
```
avdg81, line 37: I may have overlooked something, but to me it seems that these aliases are not used. Could you please double check that? If unused, please remove them.

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
avdg81, line 1: As indicated by @WPK4FEM, we should rename this file and the contained class, since generalized Newmark has a different meaning in the literature. Perhaps something like `NewmarkTSchemeBase`?

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+
+namespace Kratos {
+
+template <class TSparseSpace, class TDenseSpace>
+class GeoMechanicsTimeIntegrationScheme : public Scheme<TSparseSpace, TDenseSpace> {
+public:
+    using BaseType = Scheme<TSparseSpace, TDenseSpace>;
+    using DofsArrayType = typename BaseType::DofsArrayType;
+    using TSystemMatrixType = typename BaseType::TSystemMatrixType;
+    using TSystemVectorType = typename BaseType::TSystemVectorType;
```
avdg81: I would interchange these two alias definitions, such that the simpler data structure (a vector) comes before the more complex one (a matrix).

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+            rElement.GetDofList(rDofList, rCurrentProcessInfo);
+    }
+
+    void GetDofList(const Condition& rCondition,
+                    Condition::DofsVectorType& rDofList,
+                    const ProcessInfo& rCurrentProcessInfo) override
+    {
+        if (IsActive(rCondition))
+            rCondition.GetDofList(rDofList, rCurrentProcessInfo);
+    }
```
avdg81: Since the implementations are virtually identical, I can imagine that we may want to forward these request to a common implementation (which needs to be `private`). E.g.:
```
    template <typename T>
    void GetDofListImpl(const T&                    rElementOrCondition,
                        typename T::DofsVectorType& rDofList,
                        const ProcessInfo&          rCurrentProcessInfo)
    {
        if (IsActive(rElementOrCondition))
            rElementOrCondition.GetDofList(rDofList, rCurrentProcessInfo);
    }
```
And then call it from here:
```suggestion
    void GetDofList(const Element& rElement,
                    Element::DofsVectorType& rDofList,
                    const ProcessInfo& rCurrentProcessInfo) override
    {
        GetDofListImpl(rElement, rDofList, rCurrentProcessInfo);
    }

    void GetDofList(const Condition& rCondition,
                    Condition::DofsVectorType& rDofList,
                    const ProcessInfo& rCurrentProcessInfo) override
    {
        GetDofListImpl(rCondition, rDofList, rCurrentProcessInfo);
    }
```

This comment could also be applied to the overloaded member functions `EquationId` (see below).

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+            rCondition.EquationIdVector(rEquationId, rCurrentProcessInfo);
+    }
+
+    void Initialize(ModelPart& rModelPart) override
+    {
+        KRATOS_TRY
+
+        SetTimeFactors(rModelPart);
+
+        Scheme<TSparseSpace, TDenseSpace>::mSchemeIsInitialized = true;
```
avdg81: This is what the `BaseType` implementation does, so we could change it to:
```suggestion
        BaseType::Initialize();
```
And then perhaps also make this the first statement of the function body, to demonstrate that we extend the `BaseType`'s implementation.

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    void InitializeSolutionStep(ModelPart& rModelPart,
+                                TSystemMatrixType& A,
+                                TSystemVectorType& Dx,
+                                TSystemVectorType& b) override
+    {
+        KRATOS_TRY
+
+        SetTimeFactors(rModelPart);
```
avdg81, line 99: It is unclear to me why we need to `SetTimeFactors` here as well as in `Initialize()`. I would expect calling it from one of these member functions would be sufficient, but apparently it's not. Do you know why?

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+
+        Scheme<TSparseSpace, TDenseSpace>::mSchemeIsInitialized = true;
+
+        KRATOS_CATCH("")
+    }
+
+    void InitializeSolutionStep(ModelPart& rModelPart,
+                                TSystemMatrixType& A,
+                                TSystemVectorType& Dx,
+                                TSystemVectorType& b) override
```
avdg81: Since these parameters are not used in the function's body, I would remove their names (to avoid messages about unused parameters).
```suggestion
                                TSystemMatrixType&,
                                TSystemVectorType&,
                                TSystemVectorType&) override
```

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+        });
+
+        KRATOS_CATCH("")
+    }
+
+    void Predict(ModelPart& rModelPart,
+                 DofsArrayType& rDofSet,
+                 TSystemMatrixType& A,
+                 TSystemVectorType& Dx,
+                 TSystemVectorType& b) override
```
avdg81: Some more unused parameters here:
```suggestion
                 DofsArrayType&,
                 TSystemMatrixType&,
                 TSystemVectorType&,
                 TSystemVectorType&) override
```

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+    {
+        this->UpdateVariablesDerivatives(rModelPart);
+    }
+
+    void InitializeNonLinIteration(ModelPart& rModelPart,
+                                   TSystemMatrixType& A,
+                                   TSystemVectorType& Dx,
+                                   TSystemVectorType& b) override
```
avdg81: And the same thing here:
```suggestion
                                   TSystemMatrixType&,
                                   TSystemVectorType&,
                                   TSystemVectorType&) override
```

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+                rCondition.InitializeNonLinearIteration(r_current_process_info);
+        });
+
+        KRATOS_CATCH("")
+    }
+
+    void FinalizeNonLinIteration(ModelPart& rModelPart,
+                                 TSystemMatrixType& A,
+                                 TSystemVectorType& Dx,
+                                 TSystemVectorType& b) override
```
avdg81:
```suggestion
                                 TSystemMatrixType&,
                                 TSystemVectorType&,
                                 TSystemVectorType&) override
```

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+    {
+        KRATOS_TRY
+
+        const ProcessInfo& r_current_process_info = rModelPart.GetProcessInfo();
+
+        block_for_each(rModelPart.Elements(), [&r_current_process_info, this](Element& rElement) {
+            if (IsActive(rElement))
+                rElement.FinalizeNonLinearIteration(r_current_process_info);
+        });
```
avdg81: As discussed, there are a few patterns in this file that keep coming back. One of them is to call a given member function for each active element. We could define a (private) helper member function as follows:
```
    template <typename MemFuncPtr>
    void BlockForEachActiveElement(ModelPart& rModelPart,
                                   MemFuncPtr pMemberFunction)
    {
        const auto& r_current_process_info = rModelPart.GetProcessInfo();
        block_for_each(rModelPart.Elements(), [&r_current_process_info, pMemberFunction](auto& rElement) {
           if (IsActive(rElement)) {
               (rElement.*pMemberFunction)(r_current_process_info);
           }
        });
    }
```
(Note that here I have changed `IsActive` from a member function to a free function (i.e. a nonmember function).)

With the above helper in place, we can simplify the original code to:
```suggestion
        BlockForEachActiveElement(rModelPart, &Element::FinalizeNonLinearIteration);
```

In a similar way we can define a helper function for calling a given member function for each active condition. Do you feel that this suggestion helps to improve the readability of the code?

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+                rCondition.FinalizeNonLinearIteration(r_current_process_info);
+        });
+
+        KRATOS_CATCH("")
+    }
+
+    void FinalizeSolutionStepActiveEntities(ModelPart& rModelPart,
+                                            TSystemMatrixType& A,
+                                            TSystemVectorType& Dx,
+                                            TSystemVectorType& b)
```
avdg81:
```suggestion
                                            TSystemMatrixType&,
                                            TSystemVectorType&,
                                            TSystemVectorType&)
```

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+                       });
+
+        KRATOS_CATCH("")
+    }
+
+    template <class T>
+    bool IsActive(const T& rComponent) const
+    {
+        return !(rComponent.IsDefined(ACTIVE)) || rComponent.Is(ACTIVE);
+    }
```
avdg81: Since this function does not reference any member, we could either make it a `static` member function or a nonmember function. The latter may be preferred if you agree with my previous suggestion to introduce a helper function like `BlockForEachActiveElement`.

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+    {
+        KRATOS_TRY
+
+        rCurrentCondition.CalculateLocalSystem(
+            LHS_Contribution, RHS_Contribution, CurrentProcessInfo);
+
+        rCurrentCondition.EquationIdVector(EquationId, CurrentProcessInfo);
+
+        KRATOS_CATCH("")
+    }
```
avdg81: Also here, we could opt to extract a template member function that covers both elements and conditions. The same applies to member functions `CalculateRHSContribution` and `CalculateLHSContribution`.

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+        rCurrentCondition.CalculateLeftHandSide(LHS_Contribution, CurrentProcessInfo);
+
+        rCurrentCondition.EquationIdVector(EquationId, CurrentProcessInfo);
+
+        KRATOS_CATCH("")
+    }
+
+    void Update(ModelPart& rModelPart,
+                DofsArrayType& rDofSet,
+                TSystemMatrixType& A,
```
avdg81:
```suggestion
                TSystemMatrixType&,
```

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+                TSystemVectorType& Dx,
+                TSystemVectorType& b) override
```
avdg81:
```suggestion
                TSystemVectorType&) override
```

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+
+        this->UpdateVariablesDerivatives(rModelPart);
+
+        KRATOS_CATCH("")
+    }
+
+protected:
+    void CheckBufferSize(const ModelPart& rModelPart) const
+    {
+        constexpr int minimum_buffer_size = 2;
```
avdg81: In the check below, we compare an `int` with a `ModelPart::IndexType`. Perhaps it's better to use the same type:
```suggestion
        constexpr auto minimum_buffer_size = ModelPart::IndexType{2};
```

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+        KRATOS_ERROR_IF(rModelPart.GetBufferSize() < minimum_buffer_size)
+            << "insufficient buffer size. Buffer size should be "
+               "greater or equal to "
```
avdg81:
```suggestion
               "greater than or equal to "
```

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+            << minimum_buffer_size << ". Current size is "
+            << rModelPart.GetBufferSize() << std::endl;
+    }
+
+    void CheckSolutionStepsData(const Node& r_node, const Variable<double>& variable) const
+    {
+        KRATOS_ERROR_IF(!r_node.SolutionStepsDataHas(variable))
```
avdg81: Perhaps it's slightly better to use the `..._NOT` macro here, since that is harder to overlook than a `!`:
```suggestion
        KRATOS_ERROR_IF_NOT(r_node.SolutionStepsDataHas(variable))
```

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+            << variable.Name() << " variable is not allocated for node "
+            << r_node.Id() << std::endl;
+    }
+
+    void CheckDof(const Node& r_node, const Variable<double>& variable) const
+    {
+        KRATOS_ERROR_IF(!r_node.HasDofFor(variable))
```
avdg81:
```suggestion
        KRATOS_ERROR_IF_NOT(r_node.HasDofFor(variable))
```

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+    virtual inline void SetTimeFactors(ModelPart& rModelPart)
+    {
+        // intentionally empty
+    }
+    virtual inline void UpdateVariablesDerivatives(ModelPart& rModelPart)
+    {
+        // intentionally empty
+    }
+
+    double GetDeltaTime() const
```
avdg81: Clang-tidy suggests to mark this member function `[[nodiscard]]`.

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+                TSystemMatrixType& A,
+                TSystemVectorType& Dx,
+                TSystemVectorType& b) override
+    {
+        KRATOS_TRY
+
+        int num_threads = ParallelUtilities::GetNumThreads();
+        OpenMPUtils::PartitionVector dof_set_partition;
+        OpenMPUtils::DivideInPartitions(static_cast<int>(rDofSet.size()),
+                                        num_threads, dof_set_partition);
```
avdg81, line 298: Just a friendly reminder that we need to replace this deprecated function call by whatever we are supposed to use now (waiting for Charlie's reply).

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -0,0 +1,360 @@
...
+
+    double GetDeltaTime() const
+    {
+        return mDeltaTime;
+    }
+
+    void SetDeltaTime(double DeltaTime)
+    {
+        mDeltaTime = DeltaTime;
+    }
```
avdg81, line 365: I was wondering whether we should add some checks here to ensure that we accept valid time increments only. For instance, a time increment of 0.0 will result in division-by-zero problems when we try to update the time derivatives. (As such, I'm not sure whether a default value of 0.0 for `mDeltaTime` is a good idea, since it will result in a crash when it's not updated.)

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_element.h
```diff
@@ -0,0 +1,45 @@
...
+#pragma once
+
+#include "includes/element.h"
+
+namespace Kratos::Testing {
+
+class SpyElement : public Element {
+public:
+    void FinalizeSolutionStep(const ProcessInfo& rCurrentProcessInfo) override;
+    void InitializeSolutionStep(const ProcessInfo& rCurrentProcessInfo) override;
```
avdg81: Perhaps it's more logical to first declare `InitializeSolutionStep` and only then mention `FinalizeSolutionStep` (to have them in the order they should be used).

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_element.h
```diff
@@ -0,0 +1,45 @@
...
+    void InitializeNonLinearIteration(const ProcessInfo& rCurrentProcessInfo) override;
+    void FinalizeNonLinearIteration(const ProcessInfo& rCurrentProcessInfo) override;
+    void EquationIdVector(EquationIdVectorType& rResult,
+                          const ProcessInfo& rCurrentProcessInfo) const override;
+    void GetDofList(DofsVectorType& rElementalDofList,
+                    const ProcessInfo& rCurrentProcessInfo) const override;
+
+    bool IsSolutionStepFinalized() const;
+    bool IsSolutionStepInitialized() const;
```
avdg81: I would also swap these two member functions, since that feels more logical to me.

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_element.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+void SpyElement::FinalizeSolutionStep(const ProcessInfo& rCurrentProcessInfo)
+{
+    mSolutionStepFinalized = true;
+}
+
+void SpyElement::InitializeSolutionStep(const ProcessInfo& rCurrentProcessInfo)
+{
+    mSolutionStepInitialized = true;
+}
```
avdg81: I would swap these two implementations (initialize first, then finalize).

Also, I would remove the function parameter names, since they are not used:
```suggestion
void SpyElement::FinalizeSolutionStep(const ProcessInfo&)
{
    mSolutionStepFinalized = true;
}

void SpyElement::InitializeSolutionStep(const ProcessInfo&)
{
    mSolutionStepInitialized = true;
}
```
(This suggestion also applies to the other member functions.)

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_element.cpp
```diff
@@ -0,0 +1,78 @@
...
+void SpyElement::FinalizeNonLinearIteration(const ProcessInfo& rCurrentProcessInfo)
+{
+    mNonLinIterationFinalized = true;
+}
+
+void SpyElement::EquationIdVector(Element::EquationIdVectorType& rResult,
+                                  const ProcessInfo& rCurrentProcessInfo) const
+{
+    mIsEquationIdSet = true;
+}
```
avdg81: Is this not a getter? If it is, then perhaps `mIsEquationIdRetrieved` is a more suitable name.

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_element.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+bool SpyElement::IsSolutionStepFinalized() const
+{
+    return mSolutionStepFinalized;
+}
+
+bool SpyElement::IsSolutionStepInitialized() const
+{
+    return mSolutionStepInitialized;
+}
```
avdg81: I would also swap these two functions.

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_condition.h
```diff
@@ -0,0 +1,53 @@
...
+#pragma once
+
+#include "includes/condition.h"
+
+namespace Kratos::Testing {
+
+class SpyCondition : public Condition {
+public:
+    void FinalizeSolutionStep(const ProcessInfo& rCurrentProcessInfo) override;
+    void InitializeSolutionStep(const ProcessInfo& rCurrentProcessInfo) override;
```
avdg81: I would swap these two member functions.

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_condition.h
```diff
@@ -0,0 +1,53 @@
...
+
+    void EquationIdVector(EquationIdVectorType& rResult,
+                          const ProcessInfo& rCurrentProcessInfo) const override;
+
+    void GetDofList(DofsVectorType& rElementalDofList,
+                    const ProcessInfo& rCurrentProcessInfo) const override;
+
+    bool IsSolutionStepFinalized() const;
+
+    bool IsSolutionStepInitialized() const;
```
avdg81: And the same thing here:
```suggestion
    bool IsSolutionStepInitialized() const;
    bool IsSolutionStepFinalized() const;
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_condition.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+void SpyCondition::FinalizeSolutionStep(const ProcessInfo& rCurrentProcessInfo)
+{
+    mSolutionStepFinalized = true;
+}
+
+void SpyCondition::InitializeSolutionStep(const ProcessInfo& rCurrentProcessInfo)
+{
+    mSolutionStepInitialized = true;
+}
```
avdg81: I would swap these two function definitions, and remove the unused parameter names:
```suggestion
void SpyCondition::InitializeSolutionStep(const ProcessInfo&)
{
    mSolutionStepInitialized = true;
}

void SpyCondition::FinalizeSolutionStep(const ProcessInfo&)
{
    mSolutionStepFinalized = true;
}
```
The latter suggestion also applies to the other member functions.

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_condition.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+void SpyCondition::FinalizeNonLinearIteration(const ProcessInfo& rCurrentProcessInfo)
+{
+    mNonLinIterationFinalized = true;
+}
+
+void SpyCondition::EquationIdVector(Condition::EquationIdVectorType& rResult,
+                                    const ProcessInfo& rCurrentProcessInfo) const
+{
+    mIsEquationIdSet = true;
```
avdg81: If this function is actually a getter, a better name for the data member might be `mIsEquationIdRetrieved`.

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_condition.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+bool SpyCondition::IsSolutionStepFinalized() const
+{
+    return mSolutionStepFinalized;
+}
+
+bool SpyCondition::IsSolutionStepInitialized() const
+{
+    return mSolutionStepInitialized;
+}
```
avdg81: I would also swap these.

applications/GeoMechanicsApplication/tests/cpp_tests/test_generalized_newmark_T_scheme.cpp
```diff
@@ -0,0 +1,180 @@
...
+using namespace Kratos;
+using SparseSpaceType = UblasSpace<double, CompressedMatrix, Vector>;
+using LocalSpaceType = UblasSpace<double, Matrix, Vector>;
+
+namespace Kratos::Testing {
+
+GeneralizedNewmarkTScheme<SparseSpaceType, LocalSpaceType> CreateValidScheme()
+{
+    GeneralizedNewmarkTScheme<SparseSpaceType, LocalSpaceType> result(0.75);
+    return result;
```
avdg81: I would merge these two lines:
```suggestion
    return GeneralizedNewmarkTScheme<SparseSpaceType, LocalSpaceType> result(0.75);
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_generalized_newmark_T_scheme.cpp
```diff
@@ -0,0 +1,180 @@
...
+    auto scheme = CreateValidScheme();
+    const auto& model_part = CreateValidModelPart(model);
+
+    KRATOS_EXPECT_EQ(scheme.Check(model_part), 0);
+}
+
+KRATOS_TEST_CASE_IN_SUITE(ForInvalidTheta_CheckBackwardEulerQuasistaticTScheme_Throws,
+                          KratosGeoMechanicsFastSuite)
+{
+    constexpr int invalid_theta = -2;
```
avdg81: Since `theta` is a `double`, let's declare it as such:
```suggestion
    constexpr auto invalid_theta = -2.0;
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_generalized_newmark_T_scheme.cpp
```diff
@@ -0,0 +1,180 @@
...
+                          KratosGeoMechanicsFastSuite)
+{
+    Model model;
+    auto scheme = CreateValidScheme();
+    const auto& model_part = CreateValidModelPart(model);
+
+    KRATOS_EXPECT_EQ(scheme.Check(model_part), 0);
+}
+
+KRATOS_TEST_CASE_IN_SUITE(ForInvalidTheta_CheckBackwardEulerQuasistaticTScheme_Throws,
```
avdg81, line 50: Why would we allow to construct objects with an invalid value for `theta`? Wouldn't it be better to catch such errors on construction?

applications/GeoMechanicsApplication/tests/cpp_tests/test_generalized_newmark_T_scheme.cpp
```diff
@@ -0,0 +1,180 @@
...
+{
+    constexpr int invalid_theta = -2;
+    GeneralizedNewmarkTScheme<SparseSpaceType, LocalSpaceType> scheme(invalid_theta);
+
+    Model model;
+    auto& model_part = model.CreateModelPart("dummy", 2);
+    model_part.AddNodalSolutionStepVariable(TEMPERATURE);
+    model_part.AddNodalSolutionStepVariable(DT_TEMPERATURE);
+    auto p_node = model_part.CreateNewNode(0, 0.0, 0.0, 0.0);
+    p_node->AddDof(TEMPERATURE);
```
avdg81: I guess that you introduced the helper `CreateValidModelPart` later...? You can reuse it here :-)

applications/GeoMechanicsApplication/tests/cpp_tests/test_generalized_newmark_T_scheme.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(scheme.Check(model_part),
+                                      "Theta has an invalid value")
+}
+
+KRATOS_TEST_CASE_IN_SUITE(ForInvalidBufferSize_CheckBackwardEulerQuasistaticTScheme_Throws,
+                          KratosGeoMechanicsFastSuite)
+{
+    GeneralizedNewmarkTScheme<SparseSpaceType, LocalSpaceType> scheme(0.75);
```
avdg81: You could replace this by a call to `CreateValidScheme()`.

applications/GeoMechanicsApplication/tests/cpp_tests/test_generalized_newmark_T_scheme.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+    Model model;
+    constexpr int invalid_buffer_size = 1;
```
avdg81:
```suggestion
    constexpr auto invalid_buffer_size = ModelPart::IndexType{1};
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_generalized_newmark_T_scheme.cpp
```diff
@@ -0,0 +1,180 @@
...
+    auto& model_part = model.CreateModelPart("dummy", invalid_buffer_size);
+    model_part.AddNodalSolutionStepVariable(TEMPERATURE);
+    model_part.AddNodalSolutionStepVariable(DT_TEMPERATURE);
+    auto p_node = model_part.CreateNewNode(0, 0.0, 0.0, 0.0);
+    p_node->AddDof(TEMPERATURE);
```
avdg81: Alternatively, you could create a valid model part first (by calling `CreateValidModelPart`), followed by changing the buffer size to the invalid size of 1 (by calling `model_part.SetBufferSize`). This may make it clearer that the only invalid thing here is the buffer size.

applications/GeoMechanicsApplication/tests/cpp_tests/test_geomechanics_time_integration_scheme.cpp
```diff
@@ -0,0 +1,220 @@
...
+    std::vector<std::pair<std::function<void(const Kratos::intrusive_ptr<T> Component)>, std::function<bool(const Kratos::intrusive_ptr<T> rElement)>>> CreateFunctionsAndChecksCalledOnASingleComponent()
+    {
+        std::vector<std::pair<std::function<void(const Kratos::intrusive_ptr<T> Component)>,
+                              std::function<bool(const Kratos::intrusive_ptr<T> Component)>>>
+            functions_and_checks;
+
+        Element::EquationIdVectorType equation_id_vector_type;
+        ProcessInfo info;
+
+        auto equation_id = [this, &equation_id_vector_type,
```
avdg81: I have serious doubts whether this is correct. In the lambda expression, you capture a _reference_ to a local variable (`equation_id_vector_type`), which will go out of scope as soon as function `CreateFunctionsAndChecksCalledOnASingleComponent` returns. However, the reference is still there (in the closure). So I'm afraid this becomes a _dangling_ reference. Or am I overlooking something?

If you agree that this is incorrect, then you may fix it by creating a data member that you reference from the capture list. Or, alternatively (perhaps even better), create a container in the test's function body and pass a reference to it through the parameter list of `CreateFunctionsAndChecksCalledOnASingleComponent`. Then capture that reference. The lifetimes should then match.

applications/GeoMechanicsApplication/tests/cpp_tests/test_geomechanics_time_integration_scheme.cpp
```diff
@@ -0,0 +1,220 @@
...
+                            &info](const Kratos::intrusive_ptr<T> Component) {
+            mScheme.EquationId(*Component.get(), equation_id_vector_type, info);
+        };
+
+        auto equation_id_check = [](const Kratos::intrusive_ptr<T> Component) {
+            return Component->IsEquationIdSet();
+        };
+
+        Element::DofsVectorType dofs;
+        auto get_dofs_list = [this, &dofs, &info](const Kratos::intrusive_ptr<T> Component) {
```
avdg81: Also here I believe we will have a dangling reference. You may fix it in the same way as previously suggested.

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_quasistatic_U_Pw_scheme.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+    NewmarkQuasistaticUPwSchemeTester()
+    {
+        CreateValidModelPart();
+    }
+
+    NewmarkQuasistaticUPwScheme<SparseSpaceType, LocalSpaceType> CreateValidScheme() const
+    {
+        NewmarkQuasistaticUPwScheme<SparseSpaceType, LocalSpaceType> result(0.25, 0.5, 0.75);
+        return result;
```
avdg81: I would eliminate the redundant variable:
```suggestion
        return NewmarkQuasistaticUPwScheme<SparseSpaceType, LocalSpaceType> result{0.25, 0.5, 0.75};
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_quasistatic_U_Pw_scheme.cpp
```diff
@@ -0,0 +1,126 @@
...
+KRATOS_TEST_CASE_IN_SUITE(CheckScheme_ReturnsZeroForValidScheme, KratosGeoMechanicsFastSuite)
+{
+    NewmarkQuasistaticUPwSchemeTester tester;
+    KRATOS_EXPECT_EQ(tester.mScheme.Check(tester.GetModelPart()), 0);
+}
+
+KRATOS_TEST_CASE_IN_SUITE(InitializeUPWScheme_SetsTimeFactors, KratosGeoMechanicsFastSuite)
+{
+    NewmarkQuasistaticUPwSchemeTester tester;
+    tester.GetModelPart().GetProcessInfo()[DELTA_TIME] = 4.0;
```
avdg81: I think this statement is redundant. (This is already done by `CreateValidModelPart`, which is called when constructing a `NewmarkQuasistaticUPwSchemeTester` object.)

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_quasistatic_U_Pw_scheme.cpp
```diff
@@ -0,0 +1,126 @@
...
+    }
+};
+
+KRATOS_TEST_CASE_IN_SUITE(CheckScheme_ReturnsZeroForValidScheme, KratosGeoMechanicsFastSuite)
+{
+    NewmarkQuasistaticUPwSchemeTester tester;
+    KRATOS_EXPECT_EQ(tester.mScheme.Check(tester.GetModelPart()), 0);
+}
+
+KRATOS_TEST_CASE_IN_SUITE(InitializeUPWScheme_SetsTimeFactors, KratosGeoMechanicsFastSuite)
```
avdg81: Nitpicking:
```suggestion
KRATOS_TEST_CASE_IN_SUITE(InitializeUPwScheme_SetsTimeFactors, KratosGeoMechanicsFastSuite)
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_quasistatic_U_Pw_scheme.cpp
```diff
@@ -0,0 +1,126 @@
...
+    constexpr double expected_dt_pressure_coefficient = 1.0 / 3.0;
+    constexpr double expected_velocity_coefficient = 0.5;
+    KRATOS_EXPECT_TRUE(tester.mScheme.SchemeIsInitialized())
+    KRATOS_EXPECT_DOUBLE_EQ(tester.GetModelPart().GetProcessInfo()[DT_PRESSURE_COEFFICIENT],
+                            expected_dt_pressure_coefficient);
+    KRATOS_EXPECT_DOUBLE_EQ(tester.GetModelPart().GetProcessInfo()[VELOCITY_COEFFICIENT],
+                            expected_velocity_coefficient);
+}
+
+KRATOS_TEST_CASE_IN_SUITE(UPWSchemePredict_UpdatesVariablesDerivatives, KratosGeoMechanicsFastSuite)
```
avdg81: More nitpicking:
```suggestion
KRATOS_TEST_CASE_IN_SUITE(UPwSchemePredict_UpdatesVariablesDerivatives, KratosGeoMechanicsFastSuite)
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_quasistatic_U_Pw_scheme.cpp
```diff
@@ -0,0 +1,126 @@
...
+    Vector b;
+
+    tester.mScheme.Predict(tester.GetModelPart(), dof_set, A, Dx, b);
+
+    // These expected numbers result from the calculations in UpdateVariablesDerivatives
+    const auto expected_acceleration = Kratos::array_1d<double, 3>{-6.75, -9.0, -11.25};
+    const auto expected_velocity = Kratos::array_1d<double, 3>{-4.5, -6.0, -7.5};
+    constexpr auto expected_dt_water_pressure = 1.0 / 3.0;
+
+    KRATOS_EXPECT_EQ(tester.GetModelPart().Nodes()[0].FastGetSolutionStepValue(ACCELERATION),
```
avdg81: I'm not sure whether this macro takes into account a tolerance when comparing the `double` values contained in the arrays. If not, then there is a risk that the test is/becomes fragile (due to comparing (computed) floating-point number for equality).

avdg81 (changes requested):
I think you have done a great job here! Nicely restructured, clean code, and you've added a considerable amount of unit tests. Many thanks for that.

Most of my comments are minor in nature. However, I believe I may have spotted a few dangling references (see the unit tests). If my observations are correct, then those need to be addressed before approval.

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
rfaasse: Just discussed with @WPK4FEM and Generalized Newmark is a known term in literature for this kind of scheme (it was the 'generalized backward Euler' which was incorrectly named, so I reverted that)

applications/GeoMechanicsApplication/custom_strategies/schemes/backward_euler_T_scheme.hpp
rfaasse: We could get delta T = 0 when the SetTimeFactors is not called before this functions, so it's a good idea to double check here and throw a kratos error if that would be the case. Would it be okay with you to do that in the next PR, since this would result in more merge conflicts with the next branch

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
rfaasse: I actually prefer to not use this alias so it is at one glance clear which base function is called

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
rfaasse: I indeed removed these unused usings in the refactoring branch

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
rfaasse: Agreed, noted for the refactor PR

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
rfaasse: Let's discuss if we could remove them from the element

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
rfaasse: Noted for refactor PR

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
rfaasse: Noted for refactor PR

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_T_scheme.hpp
rfaasse: Agreed, I think 1.0 would be better. When theta = 1.0, newmark becomes identical to backward euler

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Agreed, done!

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Done, personally I prefer to be explicit about the base type, so I kept it as the Scheme<...>::Initialize(rModelPart)

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Nice suggestion, done!

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Done!

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Done

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Done

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Done

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Very nice suggestion, implemented!

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Done

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Done

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Done

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Agreed, only issue is that I don't have a UT covering this functionality yet. It's a bit more involved to add one and since in this PR I only moved the function, I'm not too concerned. When making changes, I'd like to have a unit test in place. Would you agree with doing this in the refactor PR? Since this was already there in the UPw scheme, I don't think it's necessary to block this PR for it.

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Good point, changed

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Done!

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Agreed, done

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Agreed, done

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Done

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Agreed, it's not fully robust. I made it 1.0 by default. I tried to throw when the Set got an input <= 0, but that caused a lot of tests to break, so unfortunately we would need some other changes before we add that check

applications/GeoMechanicsApplication/custom_strategies/schemes/backward_euler_T_scheme.hpp
rfaasse: Good point, I noted it down to fix in the refactor PR

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_generalized_newmark_T_scheme.cpp
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_generalized_newmark_T_scheme.cpp
rfaasse: Agreed, but as discussed, let's keep that for the refactor PR

applications/GeoMechanicsApplication/tests/cpp_tests/test_generalized_newmark_T_scheme.cpp
rfaasse: Whoops, good catch, changed!

applications/GeoMechanicsApplication/tests/cpp_tests/test_generalized_newmark_T_scheme.cpp
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_generalized_newmark_T_scheme.cpp
rfaasse: Good suggestion. I first tried to do that with the others as well, but unfortunately there is no 'RemoveNodalSolutionStepVariable'

applications/GeoMechanicsApplication/tests/cpp_tests/test_geomechanics_time_integration_scheme.cpp
rfaasse: I completely agree! Good catch, my mistake, this is why we need reviews 👍

applications/GeoMechanicsApplication/tests/cpp_tests/test_geomechanics_time_integration_scheme.cpp
rfaasse: Fixed

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_quasistatic_U_Pw_scheme.cpp
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_quasistatic_U_Pw_scheme.cpp
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_quasistatic_U_Pw_scheme.cpp
rfaasse: Good catch, removed!

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_quasistatic_U_Pw_scheme.cpp
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_quasistatic_U_Pw_scheme.cpp
rfaasse: Good point, changed it since EXPECT_EQ checks a==b, which I expect does not look at tolerances.

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_condition.cpp
rfaasse: Agreed, changed

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_condition.cpp
rfaasse: Yes, it's indeed a getter (it's also const) so changed to 'retrieved'

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_condition.cpp
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_condition.h
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_condition.h
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_element.cpp
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_element.cpp
rfaasse: Agreed, done

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_element.cpp
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_element.h
rfaasse: Done

applications/GeoMechanicsApplication/tests/cpp_tests/test_utilities/spy_element.h
rfaasse: Done

rfaasse (commented):
Thanks for the thorough review! I implemented most of your comments. Some of them, I'd propose to move to the next PR

avdg81 (approved):
This looks ready to be merged :-)

PR #11784, opened 2023-11-09
mStrainSize and mSpaceDimension by nature are positive integer types

WPK4FEM:
Constitutive laws generally should use kratos/includes/constitutive_law.h . Its member variables mStrainSize and mSpaceDimension are currently double, this PR changes them to SizeType.
That should avoid unnatural type conversions in the places where constitutive_law.h is used. E.g. see applications\StructuralMechanicsApplication\custom_response_functions\adjoint_elements\adjoint_finite_difference_small_displacement_element.cpp at line 168.


RiccardoRossi (approved):

avdg81 (approved):
Nice improvement Wijtze Pieter!

PR #11785, opened 2023-11-09
[GeoMechanics] GeoMechanics with Reduced Order Models using Kratos RomApplication

Rbravo555:
This PR implements the required changes in two python scripts of the GeoMechanics application to conform with the format expected by the RomApplication.

**📝 Description**
The RomApplication works with other applications in Kratos by keeping the behaviour of the specific app, while changing only the builder and solver. To do this, the format inside the scripts should match the expectations by the RomApp. For the GeoMechanics application, changes to two files were required:

- In **geo_mechanics_solver.py**  the required functions should be named `_CreateBuilderAndSolver` rather than `_ConstructBuilderAndSolver`, and there should be a function returning the linear solver.

- In **geomechanics_solvers_wrapper.py**, the mechanism inside the RomApp for setting the RomBuilderAndSolver requires the existance of the `CreateSolver` and `CreateSolverByParameters` functions which did not exist as separate entitied but combined into a single function `CreateSolver(model, custom_settings)`.

**🆕 Changelog**
In **geo_mechanics_solver.py**
- changed `_ConstructBuilderAndSolver(self, param1)` ==> `_CreateBuilderAndSolver(self)`
- added method _GetLinearSolver(self)

In **geomechanics_solvers_wrapper.py**
- splitted operations ` CreateSolver(model, custom_settings)` ==>` CreateSolver(model, custom_settings)` + `CreateSolverByParameters(model, custom_settings, parallelism`)

### DIFF
--- a/applications/GeoMechanicsApplication/python_scripts/geomechanics_solvers_wrapper.py
+++ b/applications/GeoMechanicsApplication/python_scripts/geomechanics_solvers_wrapper.py
@@ -40,5 +40,5 @@ def CreateSolverByParameters(model, custom_settings, parallelism):
 
 
 def CreateSolver(model, custom_settings):
-    parallelism =  custom_settings["problem_data"]["parallel_type"].GetString()
+    parallelism = custom_settings["problem_data"]["parallel_type"].GetString()
     return CreateSolverByParameters(model, custom_settings,parallelism)
--- a/applications/GeoMechanicsApplication/python_scripts/geomechanics_solvers_wrapper.py
+++ b/applications/GeoMechanicsApplication/python_scripts/geomechanics_solvers_wrapper.py
@@ -41,4 +41,4 @@ def CreateSolverByParameters(model, custom_settings, parallelism):
 
 def CreateSolver(model, custom_settings):
     parallelism = custom_settings["problem_data"]["parallel_type"].GetString()
-    return CreateSolverByParameters(model, custom_settings,parallelism)
+    return CreateSolverByParameters(model, custom_settings["solver_settings"], parallelism)
--- a/applications/GeoMechanicsApplication/python_scripts/geomechanics_solvers_wrapper.py
+++ b/applications/GeoMechanicsApplication/python_scripts/geomechanics_solvers_wrapper.py
@@ -9,13 +9,7 @@ def CreateSolverByParameters(model, custom_settings, parallelism):
     if (type(custom_settings) != KratosMultiphysics.Parameters):
         raise Exception("input is expected to be provided as a Kratos Parameters object")
 
-    if custom_settings.Has("solver_settings"):
-        params = custom_settings["solver_settings"]
-        #To conform with other apps where only the "solver_settings" are passed
-    else:
-        params = custom_settings
-
-    solver_type_raw = params["solver_type"].GetString()
+    solver_type_raw = custom_settings["solver_type"].GetString()
 
     # Solvers for OpenMP parallelism
     if (parallelism == "OpenMP"):
@@ -35,7 +29,7 @@ def CreateSolverByParameters(model, custom_settings, parallelism):
         raise Exception(err_msg)
 
     module_full_name = 'KratosMultiphysics.GeoMechanicsApplication.' + solver_module_name
-    solver = import_module(module_full_name).CreateSolver(model, params)
+    solver = import_module(module_full_name).CreateSolver(model, custom_settings)
     return solver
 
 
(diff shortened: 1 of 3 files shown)


applications/GeoMechanicsApplication/python_scripts/geomechanics_solvers_wrapper.py
```diff
@@ -30,6 +35,10 @@ def CreateSolver(model, custom_settings):
...
 
     module_full_name = 'KratosMultiphysics.GeoMechanicsApplication.' + solver_module_name
-    solver = import_module(module_full_name).CreateSolver(model, custom_settings["solver_settings"])
-
+    solver = import_module(module_full_name).CreateSolver(model, params)
     return solver
+
+
+def CreateSolver(model, custom_settings):
+    parallelism =  custom_settings["problem_data"]["parallel_type"].GetString()
```
rfaasse: Very minor formatting suggestion
```suggestion
    parallelism = custom_settings["problem_data"]["parallel_type"].GetString()
```

applications/GeoMechanicsApplication/python_scripts/geomechanics_solvers_wrapper.py
```diff
@@ -30,6 +35,10 @@ def CreateSolver(model, custom_settings):
...
+    return CreateSolverByParameters(model, custom_settings,parallelism)
```
rfaasse: @avdg81 and I discussed this part and we believe that if we pass the custom_settings["solver_settings"] here instead of custom_settings, the if-statement in line 12-16 is not needed.
Since in this function we assume anyway we're in the 'top-level' of the parameters (since we are able to get the "problem_data" part out in line 43), we should be able to make the following change
```suggestion
    return CreateSolverByParameters(model, custom_settings["solver_settings"], parallelism)
```

applications/GeoMechanicsApplication/python_scripts/geomechanics_solvers_wrapper.py
```diff
@@ -1,16 +1,21 @@
...
     if (type(custom_settings) != KratosMultiphysics.Parameters):
         raise Exception("input is expected to be provided as a Kratos Parameters object")
 
-    parallelism = custom_settings["problem_data"]["parallel_type"].GetString()
-    solver_type_raw = custom_settings["solver_settings"]["solver_type"].GetString()
+    if custom_settings.Has("solver_settings"):
+        params = custom_settings["solver_settings"]
+        #To conform with other apps where only the "solver_settings" are passed
+    else:
+        params = custom_settings
```
rfaasse: As mentioned in the comment for line 44, we believe we don't need to check if we're dealing with 'general' parameters or 'solver_settings' if we pass on the solver_settings from the CreateSolver function

applications/GeoMechanicsApplication/python_scripts/geomechanics_solvers_wrapper.py
```diff
@@ -1,16 +1,21 @@
 import KratosMultiphysics
 from importlib import import_module
 
-def CreateSolver(model, custom_settings):
 
+def CreateSolverByParameters(model, custom_settings, parallelism):
```
rfaasse, line 5: If possible within the limits of the framework, I would propose to rename this function to CreateSolverBySolverSettings to make it even more clear that the "solver_settings" should be passed to this function

rfaasse (commented):
@Rbravo555 Since I replaced @huhaas in the Kratos GeoMechanics team, I picked up this review while discussing the PR with @avdg81 and @markelov208 

Thanks for making the geo solver wrapper consistent with other applications and doing a bit more clean-up! The PR looks clean, we just have a few suggestions to make the changes a bit simpler.

applications/GeoMechanicsApplication/python_scripts/geomechanics_solvers_wrapper.py
Rbravo555: I agree that your proposed name is more descriptive, as it reflects reliance on solver_settings. However, this function is consistently named CreateSolverByParameters across multiple applications in Kratos, this consistency is crucial for the RomApplication. 

Given the implications of renaming a function that is widely used, this change would require consideration and approval by the Technical Committee. I will certainly forward your proposal to them for review.

For the time being, we will maintain the current naming convention to ensure compatibility and consistency across applications. We appreciate your understanding and value your input in this matter.

rfaasse: Thank you for taking it into consideration and explaining the situation, then indeed it is out of scope to change the name in this PR. Also, thank you for implementing the other suggestions, the PR looks ready for merging now from my point of view!

rfaasse (approved):
As mentioned, thank you for processing the suggestions, this PR is ready for merge from my point of view

PR #11790, opened 2023-11-10
[CORE]  New compilation options

roigcarlo:
**📝 Description**

This adds some new options for Kratos compilation and changes how the CI performs the compilation. Some of these set of changes should ease the process to convert Kratos and apps to discoverable CMake "modules" (@mpentek).

This also allows to fix the disk problems in the CI/FullDebug

List of new compilation options:

- `EXCLUDE_KRATOS_CORE=[ON/OFF]`: If set to `ON` allows the core to be excluded from a compilation. Default `OFF`
- `EXCLUDE_AUTOMATIC_DEPENDENCIES=[ON/OFF]`: If set to `ON` allows cmake to ignore the application dependencies specified in the cmake file of the applications. Default `OFF` 

List of changes made to the cmake files:

- Kratos core can now be included as a pre-compiled library which will be searched in the default install directory.
- Trilinos can now be find even if trilinos application is not selected
- `MappingMPIExtension` was renamed to `MeshTrilinosExtension`, as was mainly as it relies on the `EPETRA_EV` vector. Note that with this change CoSimApplication is the only a application with a pure non trilinos mpi extension
- Added explicit include trilinos dir to the extension projects
- Rom Application now correctly adds includes from Eigen/MKL/LinearSolvers

List of changes in the CI:

- Compilation has been layered into stages. At the end of each stage, C++ artifacts are cleaned. With the new compilation options is possible to maintain the libraries 
    - First stage compiles the core and direct core components (Linear Solvers, Metis and Trilinos)
    - Second stage compiles utilities ( No physics apps like: Mapping, Meshing, HDF5, MED, etc... ) 
    - Third stage compiles core applications ( List is not final )
    - Fourth stage compiles the rest.


loumalouomega:
> **📝 Description** Live testing for the new compilation options for linux
> 
> @loumalouomega Do not approve or merge this, its only for testing against the real CI.

It is not fair :(

roigcarlo:
Close as I finally managed to test the CI locally :D

roigcarlo:
@philbucher This PR may have implications about your recent PR with the selective compilation

roigcarlo:
Na, just making sure everything works, when we have the final list I'll put it here.

I would say that once we decide the apps we will merge, if I manage to solve the issues before, it can be merged and the move the apps accordingly latter.

TBH, I wanted to have it working today, but mpi problems ....

philbucher:
> Na, just making sure everything works, when we have the final list I'll put it here.
> 
> I would say that once we decide the apps we will merge, if I manage to solve the issues before, it can be merged and the move the apps accordingly latter.

Then probably better if I wait. Feel free to review it though ;) 
#11925

> 
> TBH, I wanted to have it working today, but mpi problems ....

Is trilinos not compiling? Check [this commit](https://github.com/KratosMultiphysics/Kratos/pull/11925/commits/2518c2969473778af0c55c4fbb26399a3e05be53) which fixed it for me, there is some manual parsing (which is IMO not very generic/robust) going on which can mess things up:
https://github.com/KratosMultiphysics/Kratos/blob/dcf62e1a5d4dd6bc8f96988b1f7c36b317ab982d/cmake_modules/FindTRILINOS.cmake#L16

roigcarlo:
Nono, is not trilinos, now is PARMMG with gcc

applications/MappingApplication/tests/blade_mapping_test.py
```diff
@@ -3,7 +3,7 @@
 default_data_comm = KM.Testing.GetDefaultDataCommunicator()
 if default_data_comm.IsDistributed():
     from KratosMultiphysics import mpi as KratosMPI
-    from KratosMultiphysics.MappingApplication import MPIExtension as MappingTrilinosExtension
+    from KratosMultiphysics.MappingApplication import TrilinosExtension as MappingTrilinosExtension
```
philbucher: nooo, please dont change this :D

I know it technically still depends on trilinos, but the plan is to replace with the Kratos-CSR matrix to drop the dependency entirely!

applications/MeshMovingApplication/trilinos_extension/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ file(
 ## Python module
 pybind11_add_module( KratosMeshMovingTrilinosExtension MODULE THIN_LTO ${KRATOS_MESH_MOVING_TRILINOS_EXTENSION_PYTHON_INTERFACE_SOURCES} )
 
+target_include_directories(KratosMeshMovingTrilinosExtension SYSTEM PUBLIC ${TRILINOS_INCLUDE_DIR})
```
philbucher, line 18: Are you sure this is necessary? The `KratosTrilinosCore` target already does this, IMO it should propagate. Since it worked so far I think the problem is somewhere else

applications/RomApplication/CMakeLists.txt
philbucher, line 1: basically same comment, those changes dont seem right. This should also work without, as in the past

applications/MappingApplication/tests/blade_mapping_test.py
roigcarlo: Ok... but then we should change the others to MPI extension instead of Trilinos, otherwise is super confusing

ddiezrod:
pinging @aaunnam (@altair devops) just for  you to be aware this, We might need to check this works in  our CI

applications/MeshMovingApplication/trilinos_extension/CMakeLists.txt
roigcarlo: Yep it is necessary,

As fas as I understand, when CMake resolves the include and linking directories, it takes into account the list of include and linking dirs added with `target_include_directories` from all the projects it has configured during an execution.

Before this changes, we were compiling trilinos along with all other major apps, hence, when including KratosTrillinosCore as a target dependency of, for instance, KratosFluidDynamicsTrilinosExtension, the include was propagated.

After the changes, the compilation of applications can be isolated in different runs, provided that the library exists. This would allow you do to something like:

```
-DKRATOS_ROOT="/already/compiled/kratos"
-DKRATOS_FLUID_DYNAMICS_ROOT="/already/compiled/fluidDynamics"
etc...
```

Which is similar to what is happening with the tiering (the only difference is that the linking directory is currently hardcoded to the install dir, but that could change in the future). With this situation the problem comes the moment the library is added as a linking dependency without having information about how it was created. Then the `target_include_directories` was never executed and cmake has no idea on where to find the headers from trilinos (which are used mainly in the python interface)

applications/MappingApplication/tests/blade_mapping_test.py
roigcarlo: Reverted the changes :+1:

applications/MeshMovingApplication/trilinos_extension/CMakeLists.txt
roigcarlo: To be honest, now that I wrote the response I realize that mostly the TrilinosExtensions are dependent on Trilinos includes alone, not TrilinosApplication. Probably linking the whole application into the extension is not the optimal solution....

loumalouomega:
> pinging @aaunnam (@KratosMultiphysics/altair devops) just for you to be aware this, We might need to check this works in our CI

I  modified your comment because you were mentioning a random guy which username is altair

roigcarlo:
ouch sorry

roigcarlo:
@philbucher Its alive! finally.

@ddiezrod If you give me green light from altair and there is no other problem we can merge :+1:

ddiezrod:
@aaunnam Can you please check if our CI passes with these changes?

philbucher:
cool!
I will take a close look, I would appreciate if you can hold of with merging

.github/workflows/ci-dummy.yml
philbucher, line 1: Why did you delete this one?
I think conceptually it is still needed

Or should we do this similarly to the selective compilation that I am working on?

.github/workflows/ci.yml
```diff
@@ -271,24 +326,16 @@ jobs:
...
-        key: ${{ runner.os }}-no-unity-ccache-${{ github.sha }}
-        restore-keys: ${{ runner.os }}-no-unity-ccache-
-
     - name: Build
       shell: bash
       run: |
-        export CC=/usr/lib/ccache/clang-14
-        export CXX=/usr/lib/ccache/clang++-14
+        export CC=/usr/bin/clang-14
+        export CXX=/usr/bin/clang++-14
```
philbucher, line 327: can you please also remove it from the docker image right away? We are running low on space with it all the time anyway

roigcarlo:
> cool! I will take a close look, I would appreciate if you can hold of with merging

sure np

roigcarlo:
@philbucher, @@aaunnam did you have time to look over the changes?

loumalouomega:
> @philbucher, @aaunnam did you have time to look over the changes?

A priori does not break our CI

.github/workflows/ci.yml
```diff
@@ -56,8 +44,9 @@ jobs:
       KRATOS_BUILD_TYPE: ${{ matrix.build-type }}
       KRATOS_CI_CHANGED_FILES: ${{needs.changed-files.outputs.files}}
       KRATOS_CI_APPLICATIONS: ".github/workflows/ci_apps_linux.json"
-      OMPI_MCA_rmaps_base_oversubscribe: 1 # Allow oversubscription for MPI (needed for OpenMPI >= 3.0)
-      OMPI_MCA_btl_vader_single_copy_mechanism: none # suppressing some annoying OpenMPI messages
+      KRATOS_CI_CORES: 4                              # Current limits are 4 CPU and 16 GB Ram
```
philbucher, line 47: so far we used 2, can you briefly explain why we can use 4 now?

philbucher (commented):
only preliminary comments, still need a bit more time

.github/workflows/ci.yml
roigcarlo: It is what they offer now:

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

.github/workflows/ci-dummy.yml
roigcarlo: It no longer works as we used it, on + pull_request now marks all files as changed:
https://github.com/actions/runner/issues/2324

.github/workflows/ci.yml
roigcarlo: Done

.github/workflows/configure_core.sh
```diff
@@ -0,0 +1,46 @@
...
+export KRATOS_SOURCE="${KRATOS_SOURCE:-${PWD}}"
+export KRATOS_BUILD="${KRATOS_SOURCE}/build"
+export KRATOS_APP_DIR="${KRATOS_SOURCE}/applications"
+export PYTHON_EXECUTABLE="/usr/bin/python3.10"
+export KRATOS_INSTALL_PYTHON_USING_LINKS=ON
+
+# Set applications to compile
+add_app ${KRATOS_APP_DIR}/LinearSolversApplication;
+add_app ${KRATOS_APP_DIR}/MetisApplication;
+add_app ${KRATOS_APP_DIR}/TrilinosApplication;
```
philbucher, line 25: shouldnt this be in dependencies? 
`configure_core.sh` is a bit misleading otherwise

.github/workflows/ci.yml
```diff
@@ -79,21 +68,88 @@ jobs:
...
         bash configure.sh
+        rm -r ${GITHUB_WORKSPACE}/build
+
+    - name: Build Dependencies
+      shell: bash
+      run: |
+        if [ ${{ matrix.compiler }} = gcc ]; then
+          export CC=/usr/bin/gcc-12
+          export CXX=/usr/bin/g++-12
+          export KRATOS_CMAKE_OPTIONS_FLAGS="-DUSE_EIGEN_MKL=ON -DUSE_EIGEN_FEAST=ON -DMMG_ROOT=/external_libraries/mmg/mmg_5_5_1/ -DPMMG_ROOT=/external_libraries/ParMmg_5ffc6ad -DINCLUDE_PMMG=ON"
```
philbucher: `DUSE_EIGEN_MKL=ON -DUSE_EIGEN_FEAST=ON` are required for the `LinearSolversApp`. As is, it wont build pardiso!
Please move it to the corresponding step (see comment above)

philbucher (commented):
nice! Great Job
Only one small correction required

I dont like the code duplication that is required (gitlab ci is soooo much nicer for this) but thats not for this PR. Unless you have an idea/solution for it

.github/workflows/configure_core.sh
roigcarlo: Yes.. I agree, but for reasons I need those three to be compiled in the first stage no matter what, otherwise the list of dependencies and when to activate them becomes a mess. I may change the name of the script to relfect that :S

(bascially moving trilinos and linear solvers retriggers the compilation of the core no matter what, so ....)

.github/workflows/ci.yml
roigcarlo: Ouch, good catch

philbucher (approved):
Lets gooo
Clap 👏🏻 

Btw is this mainly to solve the space issue?
Is it faster or slower now?

And will you also do this for the other apps?

roigcarlo:
It solves the space issue, no change in speed I am afraid, but since we use 4 cores by default maybe its faster.

As for other apps, we can split further if we run out of space again, we can even create a single job for every app and compile selectively is that's what you are asking :)

PR #11791, opened 2023-11-10
[GeoMechanicsApplication] Fixed the factory function that creates builder-and-solver objects

avdg81:
**📝 Description**
The factory function used to be able to create builder-and-solver objects of type `ResidualBasedBlockBuilderAndSolverWithMassAndDamping` when solver setting "prebuild_dynamics" was set to `true`. During some refactoring work (see PR 11127), it was accidentally removed. This commit restores that functionality. Also added a regression test.


aronnoordam (approved):

WPK4FEM (approved):
Indeed good to have a test that will fail when the ResidualBasedBlockBuilderAndSolverWithMassAndDamping would disappear

rfaasse (approved):

PR #11796, opened 2023-11-13
[Hotfix][Core] Monotonicity preserving with LinearSolversApplication inner solver

rubenzorrilla:
Current module import in the linear solvers Python factory imports the corresponding module from the `solver_type` value. This means that in the particular case of the monotonicity preserving solver used in combination with an application inner solver it might happen that the corresponding module is not importing when creating the linear solver instance. 

Considering that in the vast majority of cases we use a linear solver either form the core or from the `LinearSolversApplication` I suggest trying to always import the latter in the Python factory to avoid situations as the one I'm describing above. I note that this could be considered a temporary solution as this would no longer be needed if we register the linear solvers to then take the prototypes from the registry.


philbucher (changes requested):
I think this should be done differently.
Instead of importing this always, I recommend to parse the parameters for the `inner_solver_settings` and import the app if necessary. Reasons:
- More generic, would also work with solvers from other apps
- Would also work with the trilinos solvers (required once the Kratos-Matrices are fully supported and thus the Monotonicity preserving  solver can be moved to the core)
- Is not much effort to implement
- IMO very important: Avoids the import on the cluster, where it can be problematic (intel ...)

rubenzorrilla:
> I think this should be done differently. Instead of importing this always, I recommend to parse the parameters for the `inner_solver_settings` and import the app if necessary. Reasons:
> 
>     * More generic, would also work with solvers from other apps
> 
>     * Would also work with the trilinos solvers (required once the Kratos-Matrices are fully supported and thus the Monotonicity preserving  solver can be moved to the core)
> 
>     * Is not much effort to implement
> 
>     * IMO very important: Avoids the import on the cluster, where it can be problematic (intel ...)

Fair enough. I'll do the changes accordingly.

rubenzorrilla:
@philbucher I think it is ready now :wink:

philbucher (approved):
great!

I am wondering, would this also work with more nesting levels?

rubenzorrilla:
> great!
> 
> I am wondering, would this also work with more nesting levels?

Nope, I assumed one nesting level. Nonetheless, I don't think the recursive nesting is actually a possible use case so I'd leave it as it is. Also note that I also assumed the `inner_solver_settings` keyword (just for the record).

PR #11799, opened 2023-11-14
Geo/incremental linear

WPK4FEM:
**📝 Description**
To facilitate continuous elasticity during staged construction the plane strain elastic model has changed to an incremental formulation ( it was a total formulation before )


applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
```diff
@@ -42,7 +42,7 @@ SmallStrainUPwDiffOrderElement::
 
 //----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 //Destructor
```
rfaasse: Let's remove this redundant comment

applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
```diff
@@ -42,7 +42,7 @@ SmallStrainUPwDiffOrderElement::
 
 //----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
rfaasse: There is a lot of clean-up going on in this file, very nice! Could we maybe also get rid of these //---------- comments while we are making changes here anyway?

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
```diff
@@ -89,12 +101,84 @@ void GeoLinearElasticPlaneStrain2DLaw::CalculatePK2Stress(const Vector& rStrainV
...
+    noalias(mStressVector) = mStressVectorFinalized + prod(C, mDeltaStrainVector);
+
+    SetExternalStressVector(rStressVector);
 
     KRATOS_CATCH("")
 }
 
+void GeoLinearElasticPlaneStrain2DLaw::UpdateInternalDeltaStrainVector(ConstitutiveLaw::Parameters &rValues){
+    const Vector& rStrainVector = rValues.GetStrainVector();
+    std::transform(rStrainVector.begin(), rStrainVector.begin() + mDeltaStrainVector.size(), mStrainVectorFinalized.begin(), mDeltaStrainVector.begin(), std::minus<>{} );
```
rfaasse: If there is a 'minus' operator for vectors, that might make this function more readable (or could they be different sizes meaning you can't use it?)

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
```diff
@@ -89,12 +101,84 @@ void GeoLinearElasticPlaneStrain2DLaw::CalculatePK2Stress(const Vector& rStrainV
...
+}
+
+void GeoLinearElasticPlaneStrain2DLaw::UpdateInternalStrainVectorFinalized(ConstitutiveLaw::Parameters &rValues) {
```
rfaasse: Here, I think it would be cleaner if we have const Vector& as argument and pass rValues.GetStrainVector() (same with the UpdateInternalDeltaStrainVector function).

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
```diff
@@ -89,12 +101,84 @@ void GeoLinearElasticPlaneStrain2DLaw::CalculatePK2Stress(const Vector& rStrainV
...
+    this->SetInternalStrainVector(rValues.GetStrainVector());
+}
+
+void GeoLinearElasticPlaneStrain2DLaw::SetExternalStressVector(Vector& rStressVector){
+    std::copy(mStressVector.begin(), mStressVector.begin() + rStressVector.size(), rStressVector.begin());
+}
+
+void GeoLinearElasticPlaneStrain2DLaw::SetInternalStressVector(const Vector& rStressVector){
+    std::copy(rStressVector.begin(), rStressVector.begin() + mStressVectorFinalized.size(), mStressVectorFinalized.begin());
```
rfaasse: In these functions, I'm a bit confused why we use begin iterators of one and the size of another vector. Could they be different sizes, or is there a reason we would like to only copy e.g. a part of rStressVector here?

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
```diff
@@ -89,12 +101,84 @@ void GeoLinearElasticPlaneStrain2DLaw::CalculatePK2Stress(const Vector& rStrainV
...
+    std::copy(rStrainVector.begin(), rStrainVector.begin() + mStrainVectorFinalized.size(), mStrainVectorFinalized.begin());
+}
+void GeoLinearElasticPlaneStrain2DLaw::InitializeMaterialResponseCauchy(ConstitutiveLaw::Parameters& rValues)
+{
+    KRATOS_TRY
+    if (!mIsModelInitialized)
+    {
+        // stress and strain vectors must be initialized:
+        const Vector& rStressVector = rValues.GetStressVector();
+        const Vector& rStrainVector = rValues.GetStrainVector();
```
rfaasse: Since these are local variables, they should be named r_stress_vector and r_strain_vector respectively.

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
```diff
@@ -89,12 +101,84 @@ void GeoLinearElasticPlaneStrain2DLaw::CalculatePK2Stress(const Vector& rStrainV
...
+
+        SetInternalStressVector(rStressVector);
+        SetInternalStrainVector(rStrainVector);
+
+        mIsModelInitialized = true;
+    }
+    KRATOS_CATCH("")
+}
+
+void GeoLinearElasticPlaneStrain2DLaw::FinalizeMaterialResponseCauchy(ConstitutiveLaw::Parameters & rValues)
```
rfaasse: Could you remove this extra space?
```suggestion
void GeoLinearElasticPlaneStrain2DLaw::FinalizeMaterialResponseCauchy(ConstitutiveLaw::Parameters& rValues)
```

applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
```diff
@@ -335,41 +331,33 @@ void SmallStrainUPwDiffOrderElement::
...
-    // if (!IgnoreUndrained) {
-        for (SizeType i=0; i < NumPNodes; ++i)
-            rElementalDofList[Index++] = GetGeometry()[i].pGetDof( WATER_PRESSURE );
-    // }
-
-    KRATOS_CATCH( "" )
+    for (SizeType i=0; i < NumPNodes; ++i) {
+        rElementalDofList[Index] = GetGeometry()[i].pGetDof(WATER_PRESSURE);
+        ++Index;
+    }
```
rfaasse: Nice cleanup here, are there tests to cover this by any chance? Especially with these 'Index++', things can be overlooked quite quickly

applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
```diff
@@ -622,34 +594,29 @@ void SmallStrainUPwDiffOrderElement::
...
+        ++Index;
+        if (Dim > 2) {
+            rResult[Index] = rGeom[i].GetDof( DISPLACEMENT_Z ).EquationId();
+            ++Index;
         }
     }
+    for ( SizeType i = 0; i < NumPNodes; ++i ){
+        rResult[Index] = rGeom[i].GetDof( WATER_PRESSURE ).EquationId();
+        ++Index;
+    }
```
rfaasse: Same question here, I like the changes, but is this tested?

applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
```diff
@@ -669,22 +636,17 @@ void SmallStrainUPwDiffOrderElement::
...
+        ++Index;
+        if ( Dim > 2 ) {
+            rValues[Index] = rGeom[i].FastGetSolutionStepValue( VELOCITY_Z, Step );
+            ++Index;
         }
     }
-
-    for ( SizeType i = 0; i < NumPNodes; ++i )
-        rValues[Index++] = 0.0;
+    std::fill_n(rValues.begin()+Index, NumPNodes, 0.0);
```
rfaasse: Same question here, I like the changes, but is this tested?

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
```diff
@@ -89,12 +101,84 @@ void GeoLinearElasticPlaneStrain2DLaw::CalculatePK2Stress(const Vector& rStrainV
...
 
+    UpdateInternalDeltaStrainVector(rValues);
+
     Matrix C;
     this->CalculateElasticMatrix(C, rValues);
-    // Total formulation
-    noalias(rStressVector) = prod(C, rStrainVector);
+
+    // Incremental formulation
+    noalias(mStressVector) = mStressVectorFinalized + prod(C, mDeltaStrainVector);
```
rfaasse, line 110: Maybe we can have a short discussion on if we need the two separate members (normal + Finalized). To me this makes it a bit more confusing what is really happening.

rfaasse, line 110: Do you think we can add a unit test for this functionality (of course that also serves as documentation, that makes it easier to understand)

rfaasse (commented):
I really like the clean-up and the fix on the incremental stresses. I have a few improvement suggestions and questions still.

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
WPK4FEM: The rStrainVector is const Vector& the mDeltaStrainVector  is array_1d<double, VoigtSize> Hence I think we can't use the operators for vectors ( which I think would have been far more straightforward ).
The use of the .size reflects what the previous written loop was doing.

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
WPK4FEM: done.

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
WPK4FEM: done

applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
WPK4FEM: This is tested with integration tests. About 30 fail if I put a KRATOS_ERROR here.

applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
WPK4FEM: Yes, this is tested in integration tests. 30 failures when I hardcode an error here.

applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
WPK4FEM: No, not tested. When I code an error here, no evidence of it shows in the integrationtests

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
WPK4FEM: A Newton Raphson like process iterates in the way that every iteration tries something that starts from the beginning of the step ( the Finalized values here ) tries an increment ( e.g. mDeltaStrainVector ) and produces a new value for the end of the step ( e.g. mStressVector ). After acceptance of the step result ( i.e. the step has converged) the finalized values are overwritten by the predicted end of step values.

WPK4FEM: That would indeed be nice, I would appreciate help with it.

applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
WPK4FEM: This reflects the setup of the original loop.

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.h
```diff
@@ -69,22 +45,11 @@ class KRATOS_API(GEO_MECHANICS_APPLICATION) GeoLinearElasticPlaneStrain2DLaw
     /// Counted pointer of LinearPlaneStrainK0Law
     KRATOS_CLASS_POINTER_DEFINITION( GeoLinearElasticPlaneStrain2DLaw );
 
-    ///@name Life Cycle
-    ///@{
-
     /**
      * @brief The clone operation
      */
```
rfaasse, line 50: Only if you have to push other changes anyway (because it's really not important and not worth the 2 hour pipeline wait): but could you remove this trivial comment?

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.h
```diff
@@ -169,45 +96,27 @@ class KRATOS_API(GEO_MECHANICS_APPLICATION) GeoLinearElasticPlaneStrain2DLaw
...
                             ConstitutiveLaw::Parameters& rValues) override;
 
-    ///@}
-
-private:
-
-    ///@name Static Member Variables
-    ///@{
+    void UpdateInternalDeltaStrainVector(const Vector& r_strain_vector);
+    void UpdateInternalStrainVectorFinalized(const Vector& r_strain_vector);
```
rfaasse: Minor style guide thing: these should be rStrainVector

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
rfaasse: Did this work out in the end?

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
```diff
@@ -89,12 +101,83 @@ void GeoLinearElasticPlaneStrain2DLaw::CalculatePK2Stress(const Vector& rStrainV
...
+void GeoLinearElasticPlaneStrain2DLaw::UpdateInternalDeltaStrainVector(const Vector& r_strain_vector){
+    mDeltaStrainVector = r_strain_vector - mStrainVectorFinalized;
+}
+
+void GeoLinearElasticPlaneStrain2DLaw::UpdateInternalStrainVectorFinalized(const Vector& r_strain_vector) {
+    this->SetInternalStrainVector(r_strain_vector);
+}
+
+void GeoLinearElasticPlaneStrain2DLaw::SetExternalStressVector(Vector& r_stress_vector) const{
+    r_stress_vector = mStressVector;
```
rfaasse: I'm not sure assignment like this will work like intended. I think in this way, the r_stress_vector will actually become the (non-const) ref to the member, is that the intention, or do we want to set the values of r_stress_vector using the ones of mStressvector? Same holds for the other assignment functions.

Edit: Starting to doubt, since the mStressVector is not a reference itself, so it might correctly copy mStressvector data to the r_stress_vector

@avdg81 Would you know the answer here?

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
```diff
@@ -89,12 +101,83 @@ void GeoLinearElasticPlaneStrain2DLaw::CalculatePK2Stress(const Vector& rStrainV
...
+    // Incremental formulation
+    noalias(mStressVector) = mStressVectorFinalized + prod(C, mDeltaStrainVector);
+
+    SetExternalStressVector(rStressVector);
 
     KRATOS_CATCH("")
 }
 
+void GeoLinearElasticPlaneStrain2DLaw::UpdateInternalDeltaStrainVector(const Vector& r_strain_vector){
+    mDeltaStrainVector = r_strain_vector - mStrainVectorFinalized;
```
rfaasse: This is way clearer than before, thanks for making the change!

rfaasse (commented):
Looks good, making the members of type Vector helped a lot in the readability, thanks for making these changes. I have a few questions left, one of which will need @avdg81 for an answer 😄

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
```diff
@@ -51,13 +51,25 @@ void GeoLinearElasticPlaneStrain2DLaw::GetLawFeatures(Features& rFeatures)
...
+
 void GeoLinearElasticPlaneStrain2DLaw::CalculateElasticMatrix(Matrix& C, ConstitutiveLaw::Parameters& rValues)
 {
     KRATOS_TRY
 
     const Properties &r_material_properties = rValues.GetMaterialProperties();
-    const double &  E = r_material_properties[YOUNG_MODULUS];
-    const double & NU = r_material_properties[POISSON_RATIO];
+    const double &  E  = r_material_properties[YOUNG_MODULUS];
+    const double &  NU = r_material_properties[POISSON_RATIO];
```
avdg81: Nitpicking: it is rather uncommon to use references to const objects for built-in types such as `double`. In these cases I would prefer values over references. Furthermore, I would use `auto` here and rename `NU` such that is follows the Kratos Style Guide:
```suggestion
    const auto E  = r_material_properties[YOUNG_MODULUS];
    const auto nu = r_material_properties[POISSON_RATIO];
```

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
avdg81, line 1: Now that we make a clear distinction between total stress/strain and incremental stress/strain, we may want to rename a few variables and data members. For instance, the member function `CalculatePK2Stress` receives a _total_ strain vector, from which we compute the _incremental_ strain vector. The parameter name, however, doesn't reflect this explicitly. We might consider to rename it from `rStrainVector` to `rTotalStrainVector`. Would this be helpful? Or do we adopt a convention, where we use `Delta` in variable names to indicate incremental entities and otherwise we assume it is a total entity?

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
avdg81: I totally agree. In fact, I feel that this statement is so expressive that we no longer need the function itself. (In other words, we can inline this statement and eliminate the member function.)

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
avdg81: I believe this statement is copy assignment. On initialization of a reference, you make it point to a particular object. Here, we don't initialize `r_stress_vector`, it already points to an object. There is no way to make a reference point to another object later. All you can do is change the referenced object (assuming it is a reference to non-const).

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
```diff
@@ -89,12 +101,83 @@ void GeoLinearElasticPlaneStrain2DLaw::CalculatePK2Stress(const Vector& rStrainV
...
+{
+    KRATOS_TRY
+    if (!mIsModelInitialized)
+    {
+        // stress and strain vectors must be initialized:
+        const Vector& r_stress_vector = rValues.GetStressVector();
+        const Vector& r_strain_vector = rValues.GetStrainVector();
+
+        SetInternalStressVector(r_stress_vector);
+        SetInternalStrainVector(r_strain_vector);
```
avdg81: What about eliminating the temporary references?
```suggestion
        SetInternalStressVector(rValues.GetStressVector());
        SetInternalStrainVector(rValues.GetStrainVector());
```

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
```diff
@@ -89,12 +101,83 @@ void GeoLinearElasticPlaneStrain2DLaw::CalculatePK2Stress(const Vector& rStrainV
...
+    FinalizeMaterialResponseCauchy(rValues);
+}
+
+void GeoLinearElasticPlaneStrain2DLaw::save(Serializer &rSerializer) const {
+    KRATOS_SERIALIZE_SAVE_BASE_CLASS( rSerializer, LinearPlaneStrainK0Law)
+    rSerializer.save("StressVector",mStressVector);
+    rSerializer.save("StressVectorFinalized",mStressVectorFinalized);
+    rSerializer.save("DeltaStrainVector",mDeltaStrainVector);
+    rSerializer.save("StrainVectorFinalized",mStrainVectorFinalized);
+    rSerializer.save("mIsModelInitialized",mIsModelInitialized);
```
avdg81, line 150: A very minor style thing (sorry for the nitpicking ;-)):
```suggestion
    rSerializer.save("StressVector", mStressVector);
    rSerializer.save("StressVectorFinalized", mStressVectorFinalized);
    rSerializer.save("DeltaStrainVector", mDeltaStrainVector);
    rSerializer.save("StrainVectorFinalized", mStrainVectorFinalized);
    rSerializer.save("mIsModelInitialized", mIsModelInitialized);
```

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
```diff
@@ -89,12 +101,83 @@ void GeoLinearElasticPlaneStrain2DLaw::CalculatePK2Stress(const Vector& rStrainV
...
+}
+
+void GeoLinearElasticPlaneStrain2DLaw::load(Serializer &rSerializer) {
+    KRATOS_SERIALIZE_LOAD_BASE_CLASS( rSerializer, LinearPlaneStrainK0Law)
+    rSerializer.load("StressVector",mStressVector);
+    rSerializer.load("StressVectorFinalized",mStressVectorFinalized);
+    rSerializer.load("DeltaStrainVector",mDeltaStrainVector);
+    rSerializer.load("StrainVectorFinalized",mStrainVectorFinalized);
+    rSerializer.load("mIsModelInitialized",mIsModelInitialized);
```
avdg81, line 159: And the same nitpicking here:
```suggestion
    rSerializer.load("StressVector", mStressVector);
    rSerializer.load("StressVectorFinalized", mStressVectorFinalized);
    rSerializer.load("DeltaStrainVector", mDeltaStrainVector);
    rSerializer.load("StrainVectorFinalized", mStrainVectorFinalized);
    rSerializer.load("mIsModelInitialized", mIsModelInitialized);
```

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.h
```diff
@@ -169,45 +96,27 @@ class KRATOS_API(GEO_MECHANICS_APPLICATION) GeoLinearElasticPlaneStrain2DLaw
...
+    void SetExternalStressVector(Vector& rStressVector) const;
+    void SetInternalStressVector(const Vector& rStressVector);
+    void SetInternalStrainVector(const Vector& rStrainVector);
+    void InitializeMaterialResponseCauchy(ConstitutiveLaw::Parameters& rValues) final;
+    void FinalizeMaterialResponseCauchy(ConstitutiveLaw::Parameters & rValues) final;
```
avdg81, line 101: More nitpicking:
```suggestion
    void FinalizeMaterialResponseCauchy(ConstitutiveLaw::Parameters& rValues) final;
```

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.h
```diff
@@ -95,63 +60,25 @@ class KRATOS_API(GEO_MECHANICS_APPLICATION) GeoLinearElasticPlaneStrain2DLaw
...
+    bool IsIncremental() override;
 
     /**
-     * @brief  Itreturns the value of a specified variable
+     * @brief  It returns the value of a specified variable
      * @param rThisVariable the variable to be returned
      * @param rValue a reference to the returned value
      * @param rValue output: the value of the specified variable
      */
     bool& GetValue(const Variable<bool>& rThisVariable, bool& rValue) override;
```
avdg81, line 79: To avoid hiding of member functions inherited from the `ConstitutiveLaw` class, I would add the following statement (to make them visible):
```
using ConstitutiveLaw::GetValue;
```
(This will resolve a problem found by SonarQube as well as by clang-tidy.)

avdg81 (commented):
This is again a nice step forwards. Thanks for your effort to make this code more general and broadening its application, as well as cleaning up and restructuring the code. Very nice achievement!

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
WPK4FEM: Changed member variables to vectors.

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
WPK4FEM: Made the argument of ClculatePK2Stress rTotalStrainVector.

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
WPK4FEM: Inlined the statement.

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
WPK4FEM: done

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.cpp
WPK4FEM: done

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.h
WPK4FEM: done

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.h
WPK4FEM: done

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.h
WPK4FEM: done

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_plane_strain_2D_law.h
WPK4FEM: done

avdg81 (approved):
Very nice work :-) Let's merge it.

rfaasse (approved):
Looks clear, thanks for incorporating the changes!

PR #11800, opened 2023-11-14
Geo/optimise solver linear systems

aronnoordam:
- add solver without neccesity stiffness matrix calculation for the RHS. As internal forces do not have to be calculated for completely linear systems 
- check if factorized LHS can be maintained rather than the regular LHS

### DIFF
--- a/applications/GeoMechanicsApplication/custom_python/add_custom_solvers_to_python.cpp
+++ b/applications/GeoMechanicsApplication/custom_python/add_custom_solvers_to_python.cpp
@@ -1,0 +1,54 @@
+// KRATOS___
+//     //   ) )
+//    //         ___      ___
+//   //  ____  //___) ) //   ) )
+//  //    / / //       //   / /
+// ((____/ / ((____   ((___/ /  MECHANICS
+//
+//  License:         geo_mechanics_application/license.txt
+//
+//  Main authors:    Ignasi de Pouplana,
+//                   Aron Noordam
+//
+
+// External includes
+#include "spaces/ublas_space.h"
+
+// Project includes
+#include "custom_python/add_custom_strategies_to_python.h"
+#include "includes/kratos_parameters.h"
+
+//linear solvers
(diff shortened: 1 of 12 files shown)


aronnoordam:
same as #11792 with name which can be detected by build server

applications/GeoMechanicsApplication/custom_python/add_custom_solvers_to_python.cpp
```diff
@@ -0,0 +1,54 @@
...
+    using ReordererType = Reorderer<SparseSpaceType, LocalSpaceType >;
+    using DirectSolverType = DirectSolver<SparseSpaceType, LocalSpaceType, ReordererType >;
+    using PreFactorizedSkylineLUFactorizationSolverType = PreFactorizedSkylineLUFactorizationSolver<SparseSpaceType, LocalSpaceType, ReordererType >;
+
+
+//----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+
+
+    py::class_<PreFactorizedSkylineLUFactorizationSolverType, PreFactorizedSkylineLUFactorizationSolverType::Pointer, DirectSolverType>(m, "PreFactorizedSkylineLUFactorizationSolver")
+        .def(py::init< >())
```
RiccardoRossi, line 45: just beware because the skyline LU solver is a legacy "native" solver in Kratos. It works...but it is not high performence. You may want to do something similar in the  LinearSolversApplication using either SuperLI or Pardiso solvers

RiccardoRossi (commented):
this is self contained within the geomechanics application, so i have no problem with it. 

my only comment is thay skyline LI is not hight performance ...

applications/GeoMechanicsApplication/custom_python/add_custom_solvers_to_python.cpp
aronnoordam: okay thank you, I will look into it

PR #11807, opened 2023-11-15
[GeoMechanics] New apply excavation operation

rubenzorrilla:
**📝 Description**
This PR adds a new `ApplyExcavationOperation` created from the existent `ApplyExcavationProcess`. This is required for the upgrade of current multistage capabilities of the `GeoMechanicsApplication` to use `Orchestrator`-based from the core.


applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
```diff
@@ -0,0 +1,74 @@
...
+{}
+
+Operation::Pointer ApplyExcavationOperation::Create(
+    Model &rModel,
+    Parameters Parameters) const
+{
+    return Kratos::make_shared<ApplyExcavationOperation>(rModel, Parameters);
+}
+
+void ApplyExcavationOperation::Execute()
```
rfaasse: It would be a good idea to add some unit tests for this public function to have quick tests that ensure the behavior stays the same.

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.h
```diff
@@ -0,0 +1,113 @@
...
+// System includes
+
+
+// External includes
+
+
+// Project includes
+#include "operations/operation.h"
+
+// Application includes
```
rfaasse: Since there are no system, external or application includes here, we generally don't add these comments within the geomechanics application.

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include "includes/kratos_parameters.h"
+#include "utilities/variable_utils.h"
+
+// Application includes
+#include "apply_excavation_operation.h"
+
+namespace Kratos
+{
+
+ApplyExcavationOperation::ApplyExcavationOperation(
```
rfaasse: Since this will be a new operation, we would like to change the structure a bit and create one 'ActivateModelPart' and one 'DeactivateModelPart' class deriving from Operation (while of course re-using any shared functionality they both need access to).

We had a team discussion and we believe this makes it more clear to the users what this operation actually does (rather than having a single class which can activate and deactivate based on a flag).

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
```diff
@@ -0,0 +1,74 @@
...
+    Model& rModel,
+    const Parameters rSettings)
+    : Operation()
+    , mDeactivateSoilPart(rSettings["deactivate_soil_part"].GetBool())
```
rfaasse: When splitting this into the two new operations, this boolean member would no longer be necessary.

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
```diff
@@ -0,0 +1,74 @@
...
+//                   Aron Noordam,
+//                   Vahid Galavi,
+//                   Marjan Fathian
+//                   Ruben Zorrilla
+//
+
+// System includes
+
+
+// External includes
```
rfaasse: Since there are no system or external includes here, we generally don't add these comments within the geomechanics application.

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.h
```diff
@@ -0,0 +1,113 @@
...
+    /// Pointer definition of ApplyExcavationOperation
+    KRATOS_CLASS_POINTER_DEFINITION(ApplyExcavationOperation);
+
+    /// Registry current operation
+    KRATOS_REGISTRY_ADD_PROTOTYPE("Operations.KratosMultiphysics.GeoMechanicsApplication", Operation, ApplyExcavationOperation)
+    KRATOS_REGISTRY_ADD_PROTOTYPE("Operations.All", Operation, ApplyExcavationOperation)
+
+    ///@}
+    ///@name Life Cycle
+    ///@{
```
rfaasse: Where are the names of these sections (Type Definitions, Life Cycle etc) used? Is it for autogenerated documentation? And if so, who are the users of this documentation?

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.h
```diff
@@ -0,0 +1,113 @@
...
+
+    /// Default constructor.
```
rfaasse: It seems this comment is added for autogenerated documentation, however since it is trivial that this is the default constructor (same with the destructor, copy constructor and assignment operator) we would propose to remove these comments as it does not add a lot of clarification.

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
```diff
@@ -0,0 +1,74 @@
...
+    // Activate/deactivate the elements of the model part according to the excavation status
+    VariableUtils().SetFlag(ACTIVE, !mDeactivateSoilPart, mpModelPart->Elements());
+
+    // Reset the elements' constitutive law in case of excavation
+    if (mDeactivateSoilPart) {
+        block_for_each(mpModelPart->Elements(), [](Element &rElement) {
+            rElement.ResetConstitutiveLaw();
+        });
+    } else {
+        //TODO: Ask Deltares team why this is needed
```
rfaasse: This is a good question to which I don't have a quick answer. I'll discuss it within the team.

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
```diff
@@ -0,0 +1,74 @@
...
+    Model &rModel,
+    Parameters Parameters) const
+{
+    return Kratos::make_shared<ApplyExcavationOperation>(rModel, Parameters);
+}
+
+void ApplyExcavationOperation::Execute()
+{
+    KRATOS_TRY
+
```
rfaasse: Since this function is now duplicated it would be a good idea to add comments that link the two (both in this function and its counterpart in the ApplyExcavationProcess). It probably also helps to state the intent of removing the process and using this operation instead.

rfaasse (commented):
Thank you @rubenzorrilla for taking the initiative and time to move the excavation process logic to the more suitable structure of an operation (and in turn taking a step towards being able to use the orchestrator-based multi-stage analysis for geomechanics)!
 
Since this is a new operation which is not used yet, we would like to use that freedom to make some changes (as we also briefly discussed during our meeting about the orchestrator). I placed our suggestions in the comments. However, since these are quite significant changes and the scope is (for now at least) limited to the geomechanics application, let us know if it would be more suitable for the geomechanics team to pick it up. Of course, if you would prefer to process these comments, that is also fine with us!

In any case, I have also posed a question about the 'doxygen-like' comments that are added to the operation, which we are interested in learning more about.

Please let me know if there are any unclarities, or if you think a meeting with (part of) the team would be more convenient!

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
rubenzorrilla: Sure (I just wanted to make sure of the implementation before doing the test).

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.h
rubenzorrilla: OK. Fine for me. I'll remove them.

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
rubenzorrilla: OK. Makes sense for me. My unique concern is the naming, do you want them to be called `ActivateModelPartOperation` / `DeactivateModelPartOperation` or `ExcavateOperation` / `FillOperation` (or of this sort).

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.h
rubenzorrilla: We normally take them from the corresponding base classes. There is also a `templates` folder in the Kratos core. Currently, we don't generate the Doxygen (yet) but the idea is to use it as documentation of the code and to implement it properly to ease the eventual generation of it.

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.h
rubenzorrilla: We follow this as a standard for the header documentation, so my suggestion would be to keep it. Note that this might be trivial for you, but not for a newbie or a student which is not as acquainted with c++.

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
rubenzorrilla: OK. Just let me know. For the moment I'm keeping it like this with the `TODO` comment.

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
rubenzorrilla: Makes pretty much sense. Indeed, we can even add a deprecation warning to the `ApplyExcavationProcess` if you want.

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
rfaasse: We see more applications for this functionality than just excavation/filling operations (tunneling, soil structure interactions, adding/removing certain structures on top of the soil), so we would prefer the more general names of `ActivateModelPartOperation` / `DeactivateModelPartOperation`.

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
rfaasse: That's indeed a very good idea! I would propose to do this when we start supporting the orchestrator-based workflow for the geomechanics users.

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.h
rfaasse: Thank you!

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.h
rfaasse: Good to know, thank you for that perspective and pointing to the templates (I didn't know about these yet)!

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.h
rfaasse: Okay clear, we can leave it like it is for this PR!

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
rubenzorrilla: Very good. I'll do so then.

rfaasse:
@rubenzorrilla Thanks for your quick answers to my questions and suggestions! Please let us know if you need/want any assistance with making the changes or if there are still things unclear!

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.h
rubenzorrilla: Done.

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
rubenzorrilla: Done.

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
rubenzorrilla: Done.

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
rubenzorrilla: Just in case, I'm keeping the activation / deactivation of nodes (though I think it is not needed at all).

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
rubenzorrilla: Removed.

applications/GeoMechanicsApplication/custom_operations/apply_excavation_operation.cpp
rubenzorrilla: Done.

applications/GeoMechanicsApplication/tests/cpp_tests/test_activate_model_part_operation.cpp
```diff
@@ -0,0 +1,76 @@
...
+    auto p_point_4 = Kratos::make_intrusive<Node>(4, 1.0, 0.0, 0.0);
+    Quadrilateral2D4<Node> domain_geometry(p_point_1, p_point_2, p_point_3, p_point_4);
+
+    Parameters mesher_parameters(R"({
+        "number_of_divisions": 2,
+        "element_name": "Element2D3N",
+        "condition_name": "LineCondition2D2N",
+        "create_skin_sub_model_part": false
+    })");
+    StructuredMeshGeneratorProcess(domain_geometry, r_test_model_part, mesher_parameters).Execute();
```
rfaasse, line 44: Its interesting to see how you use this process, thanks for showing this! I have one question about it: I don't see that the process adds conditions to the model part, so it might be that this test loops over the conditions, but there might not be any (so then it also doesn't test them). Could we add a check to the test (same for the de-activate) for the number of nodes/elements/conditions (i.e. KRATOS_EXPECT_TRUE(r_test_model_part.Conditions().size() > 0))? Then we're sure we're testing everything.

applications/GeoMechanicsApplication/custom_operations/activate_model_part_operation.cpp
```diff
@@ -0,0 +1,60 @@
...
+    Model &rModel,
+    Parameters Parameters) const
+{
+    return Kratos::make_shared<ActivateModelPartOperation>(rModel, Parameters);
+}
+
+void ActivateModelPartOperation::Execute()
+{
+    KRATOS_TRY
+
```
rfaasse, line 47: In theory the mpModelPart could be a nullptr here (e.g. when there is a typo in the project parameters 'model_part_name'). It might be a good idea to throw a KRATOS_ERROR when that's the case, otherwise we'll segfault.

rfaasse (commented):
@rubenzorrilla Thanks a lot for incorporating the review comments and adding the unit tests! There are two questions I have left, but other than that, It looks good to go and ready for us to start experimenting with the orchestrator!

applications/GeoMechanicsApplication/custom_operations/activate_model_part_operation.cpp
rubenzorrilla: ~Fair enough. Though a wrong model part name value would be handled by the `Model` the wrong key value (e.g. `modle_port_name` is not properly handled).~

Edit: After having checked the implementation, I think that all the possible exceptions are already handled, either by the `Parameters` or by the `Model`. I mean,
- if wrong `model_part_name` key is provided the `Parameters` will throw an error.
- if wrong `model_part_name` value is provided the `Model` will throw an error when trying to retrieve the model part.

applications/GeoMechanicsApplication/tests/cpp_tests/test_activate_model_part_operation.cpp
rubenzorrilla: Yes, the `StructuredMeshGeneratorProcess` is quite handy for building unittests of this sort without the need of having an mdpa. About the test modification, sure.

rubenzorrilla: Actually I switched on the creation of the skin (see the `"create_skin_sub_model_part": true`) in order to test the conditions activation/deactivation as well.

applications/GeoMechanicsApplication/custom_operations/activate_model_part_operation.cpp
rfaasse: Good point, then indeed no extra check is necessary here.

applications/GeoMechanicsApplication/tests/cpp_tests/test_activate_model_part_operation.cpp
rfaasse: Thanks, then I think this test is okay!

rfaasse (commented):
@rubenzorrilla Sorry for not re-reviewing earlier, I missed the notifications of your replies. The code looks good, there is just the matter of SonarCloud left. 

On this branch there are 14 new code smells (see [SonarCloud](https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&sinceLeakPeriod=true&branch=geo%2Fapply-excavation-operation&id=KratosGeomechanics), let me know if you can't access it). They are all very small changes (there are some redundant semi-colons and some variables that could be const&). Could you have a look at these? Please also let me know if you want me to make the changes and apologies for not bringing up SonarCloud earlier.

rubenzorrilla:
@rfaasse I applied the SonarCloud suggestions (the ones referring to the settings cannot be applied as the `Parameters` necessarily need to be passed by copy). I think that now this is ready to be merged.

rfaasse (approved):
Thanks for taking the time to fix the code smells! It looks all good now!

PR #11808, opened 2023-11-15
[GeoMechanicsApplication] Reduce duplicated functionality between schemes and refactor the inheritance structure

### LINKED ISSUE #11805: [GeoMechanicsApplication] Improve inheritance structure of thermal and related schemes
This issue aims to make sure functionality between the very similar T and PW schemes is shared instead of duplicated. This makes it a lot more convenient to add schemes with new parameters and This also includes making sure this could be extended to the more complex family of UPw schemes. _However, refactoring the UPw schemes is out of scope for this issue._

rfaasse:
**📝 Description**
This PR aims to simplify the class structure of the schemes in the geomechanics application. This reduces the number of duplicated functionality and makes the structure more intuitive. In an earlier PR, the common functionality was extracted from the UPw scheme to the geomechanics time integration scheme, which is still the baseclass of our hierarchy. In this PR, base classes for BackwardEuler and GeneralizedNewmark schemes were created which contains al common functionality in a generalized way. This makes it a lot easier to add a new scheme with different scalar variables.

Note that not all UPw functionality is cleaned up in the same way (the backward euler upw scheme still incorrectly derives from the newmark scheme). However, when changing that part as well, the overview of what is happening in this refactoring effort might be lost. Hence, it's not included in this PR but will be part of future efforts.

Before refactoring, unit tests were put in place to make sure the functionality didn't change along the way (see the test folder in this PR)

Old class hierarchy depiction:
image

New class hierarchy depiction:
image


applications/GeoMechanicsApplication/custom_strategies/schemes/backward_euler_scheme.hpp
```diff
@@ -0,0 +1,81 @@
...
+            delta_variable / this->GetDeltaTime();
+    }
+
+    inline void SetTimeFactors(ModelPart& rModelPart) override
+    {
+        KRATOS_TRY
+
+        this->SetDeltaTime(rModelPart.GetProcessInfo()[DELTA_TIME]);
+        rModelPart.GetProcessInfo()[mDeltaTimeVariableCoefficient] =
+            1.0 / (this->GetDeltaTime());
```
WPK4FEM: I do not doubt correctness here. To emphasize that the same delta t is used, it seems logical to do the following:
auto delta_time = rModelPart.GetProcessInfo()[DELTA_TIME];
this->SetDeltaTime(delta_time);
rModelPart.GetProcessInfo()[mDeltaTimeVariableCoefficient] = 1 / delta_time:

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_scheme.hpp
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    inline void SetTimeFactors(ModelPart& rModelPart) override
+    {
+        KRATOS_TRY
+
+        this->SetDeltaTime(rModelPart.GetProcessInfo()[DELTA_TIME]);
+        rModelPart.GetProcessInfo()[mDeltaTimeVariableCoefficient] =
+            1.0 / (mTheta * this->GetDeltaTime());
```
WPK4FEM: In essence the same remark as earlier, make clear that the delta time is one and the same by using a named variable i.s.o. copying to a new place and then getting from the new place.

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -354,6 +369,18 @@ class GeoMechanicsTimeIntegrationScheme : public Scheme<TSparseSpace, TDenseSpac
         // intentionally empty
```
WPK4FEM: Not empty, filled with KRATOS_ERROR_IF_NOT.

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -354,6 +369,18 @@ class GeoMechanicsTimeIntegrationScheme : public Scheme<TSparseSpace, TDenseSpac
...
+    virtual void CheckAllocatedVariables(const ModelPart& rModelPart) const
+    {
+        // intentionally empty
+    }
+
+    virtual void UpdateScalarTimeDerivative(Node& rNode,
+                                            const Variable<double>& variable,
+                                            const Variable<double>& dt_variable) const
+    {
+        // intentionally empty
```
WPK4FEM: Derived classes must provide an override of this, therefore a fatal errormessage is appropriate here.

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_dynamic_U_Pw_scheme.hpp
```diff
@@ -177,13 +176,7 @@ class NewmarkDynamicUPwScheme : public NewmarkQuasistaticUPwScheme<TSparseSpace,
...
                                                                  * rNode.FastGetSolutionStepValue(ACCELERATION);
 
-            const double DeltaPressure =  rNode.FastGetSolutionStepValue(WATER_PRESSURE)
-                                        - rNode.FastGetSolutionStepValue(WATER_PRESSURE, 1);
-
-            rNode.FastGetSolutionStepValue(DT_WATER_PRESSURE) =  (1.0/(mTheta*this->GetDeltaTime()))
-                                                                * ( DeltaPressure
-                                                                   - (1.0-mTheta) * this->GetDeltaTime()
-                                                                   * rNode.FastGetSolutionStepValue(DT_WATER_PRESSURE,1));
+            this->UpdateScalarTimeDerivative(rNode, WATER_PRESSURE, DT_WATER_PRESSURE);
```
WPK4FEM, line 179: nice!

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_dynamic_U_Pw_scheme.hpp
```diff
@@ -36,7 +36,6 @@ class NewmarkDynamicUPwScheme : public NewmarkQuasistaticUPwScheme<TSparseSpace,
     using LocalSystemMatrixType = typename BaseType::LocalSystemMatrixType;
     using NewmarkQuasistaticUPwScheme<TSparseSpace,TDenseSpace>::mBeta;
     using NewmarkQuasistaticUPwScheme<TSparseSpace,TDenseSpace>::mGamma;
-    using NewmarkQuasistaticUPwScheme<TSparseSpace,TDenseSpace>::mTheta;
```
WPK4FEM, line 39: I do not understand why theta should be removed here.

Now I do. Its storage is handled in the base class generalized Newmark. In future, I expect Beta and Gamma to end up there too. That is the nature of Generalized Newmark. Beta and Gamma are used for the equations with first and second order time derivatives. Theta is used for the equations with only first order time derivatives.

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
```diff
@@ -17,167 +17,135 @@
...
+    KRATOS_CLASS_POINTER_DEFINITION(NewmarkQuasistaticUPwScheme);
 
-    using BaseType              = Scheme<TSparseSpace,TDenseSpace>;
-    using DofsArrayType         = typename BaseType::DofsArrayType;
-    using TSystemMatrixType     = typename BaseType::TSystemMatrixType;
-    using TSystemVectorType     = typename BaseType::TSystemVectorType;
+    using BaseType = Scheme<TSparseSpace, TDenseSpace>;
+    using DofsArrayType = typename BaseType::DofsArrayType;
+    using TSystemMatrixType = typename BaseType::TSystemMatrixType;
+    using TSystemVectorType = typename BaseType::TSystemVectorType;
```
WPK4FEM, line 34: somehow, I like the previous layout better. Did we agree on a style that dictates which one to use.

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
```diff
@@ -17,167 +17,135 @@
...
         }
 
         this->CheckBufferSize(rModelPart);
 
         // Check beta, gamma and theta
-        KRATOS_ERROR_IF(mBeta <= 0.0 || mGamma<= 0.0 || mTheta <= 0.0)
-            << "Some of the scheme variables: beta, gamma or theta has an invalid value"
-            << std::endl;
+        KRATOS_ERROR_IF(mBeta <= 0.0 || mGamma <= 0.0) << "Some of the scheme variables: beta or gamma has an invalid value"
+                                                                              << std::endl;
```
WPK4FEM: It would be friendly to the user to split this into 2 and be specific on which variable has the invalid value.

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
```diff
@@ -186,32 +154,27 @@ class NewmarkQuasistaticUPwScheme : public GeoMechanicsTimeIntegrationScheme<TSp
...
+                 this->GetDeltaTime() * rNode.FastGetSolutionStepValue(VELOCITY, 1) -
+                 (0.5 - mBeta) * this->GetDeltaTime() * this->GetDeltaTime() *
+                     rNode.FastGetSolutionStepValue(ACCELERATION, 1)) /
+                (mBeta * this->GetDeltaTime() * this->GetDeltaTime());
+
+            noalias(rNode.FastGetSolutionStepValue(VELOCITY)) =
+                rNode.FastGetSolutionStepValue(VELOCITY, 1) +
+                (1.0 - mGamma) * this->GetDeltaTime() *
+                    rNode.FastGetSolutionStepValue(ACCELERATION, 1) +
+                mGamma * this->GetDeltaTime() * rNode.FastGetSolutionStepValue(ACCELERATION);
```
WPK4FEM: Could be factored out as UpdateVectorFirstTimeDerivative and UpdateVectorSecondTimeDerivative.

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_damped_U_Pw_scheme.hpp
```diff
@@ -36,9 +36,9 @@ class NewmarkQuasistaticDampedUPwScheme : public NewmarkQuasistaticUPwScheme<TSp
         : NewmarkQuasistaticUPwScheme<TSparseSpace,TDenseSpace>(beta, gamma, theta)
     {
         //Allocate auxiliary memory
-        int NumThreads = ParallelUtilities::GetNumThreads();
-        mDampingMatrix.resize(NumThreads);
-        mVelocityVector.resize(NumThreads);
+        int num_threads = ParallelUtilities::GetNumThreads();
```
WPK4FEM: unsigned int or SizeType? a negative number of Threads would be strange.

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
```diff
@@ -17,167 +17,135 @@
...
-                               TSystemVectorType& Dx,
-                               TSystemVectorType& b) override
+    void FinalizeSolutionStep(ModelPart& rModelPart,
+                              TSystemMatrixType& A,
+                              TSystemVectorType& Dx,
+                              TSystemVectorType& b) override
     {
         KRATOS_TRY
 
         if (rModelPart.GetProcessInfo()[NODAL_SMOOTHING]) {
```
WPK4FEM, line 57: It is a mystery to me why this should be part of the scheme. I would expect it in a post processing process, and only for items that are requested by the user.

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
```diff
@@ -17,167 +17,135 @@
...
-            unsigned int Dim = rModelPart.GetProcessInfo()[DOMAIN_SIZE];
+            unsigned int dim = rModelPart.GetProcessInfo()[DOMAIN_SIZE];
 
-            SizeType StressTensorSize = STRESS_TENSOR_SIZE_2D;
-            if (Dim == N_DIM_3D) StressTensorSize = STRESS_TENSOR_SIZE_3D; 
+            SizeType stress_tensor_size = STRESS_TENSOR_SIZE_2D;
+            if (dim == N_DIM_3D)
+                stress_tensor_size = STRESS_TENSOR_SIZE_3D;
```
WPK4FEM, line 62: stress_tensor_size = ( dim == N_DIM_3D ? STRESS_TENSOR_SIZE_3D : STRESS_TENSOR_SIZE_2D );
not sure about the correct syntax, but I would expect STRESS_TENSOR_SIZE_PLANE_STRAIN ( maybe that does not exist, but at least the Voigt vector size differs between plane stress, plane strain and 3D ( 3, 4 and 6 as values ).

applications/GeoMechanicsApplication/tests/cpp_tests/test_backward_euler_Pw_scheme.cpp
```diff
@@ -0,0 +1,121 @@
...
+    model_part.AddNodalSolutionStepVariable(WATER_PRESSURE);
+    model_part.AddNodalSolutionStepVariable(DT_WATER_PRESSURE);
+
+    constexpr double current_temperature = 1.0;
+    constexpr double previous_temperature = 0.0;
+    constexpr double delta_time = 4.0;
+
+    model_part.GetProcessInfo()[DELTA_TIME] = delta_time;
+    auto p_node = model_part.CreateNewNode(0, 0.0, 0.0, 0.0);
+    p_node->FastGetSolutionStepValue(WATER_PRESSURE) = current_temperature;
```
WPK4FEM: equal, but even clearer to me when comparing to the next line:
p_node->FastGetSolutionStepValue(WATER_PRESSURE, 0) = current_temperature;

applications/GeoMechanicsApplication/tests/cpp_tests/test_backward_euler_Pw_scheme.cpp
```diff
@@ -0,0 +1,121 @@
...
+    model_part.AddNodalSolutionStepVariable(DT_WATER_PRESSURE);
+
+    constexpr double delta_time = 3.0;
+    model_part.GetProcessInfo()[DELTA_TIME] = delta_time;
+
+    scheme.Initialize(model_part);
+
+    KRATOS_EXPECT_TRUE(scheme.SchemeIsInitialized())
+    KRATOS_EXPECT_DOUBLE_EQ(model_part.GetProcessInfo()[DT_PRESSURE_COEFFICIENT],
+                            1.0 / (delta_time));
```
WPK4FEM: extra ( )?

applications/GeoMechanicsApplication/tests/cpp_tests/test_backward_euler_T_scheme.cpp
```diff
@@ -53,4 +53,69 @@ KRATOS_TEST_CASE_IN_SUITE(BackwardEulerScheme_UpdatesVariablesDerivatives_WhenPr
...
+}
+
+KRATOS_TEST_CASE_IN_SUITE(ForMissingTemperatureSolutionStepVariable_CheckBackwardEulerTScheme_Throws,
+                          KratosGeoMechanicsFastSuite)
+{
+    BackwardEulerTScheme<SparseSpaceType, LocalSpaceType> scheme;
+
+    Model model;
+    auto& model_part = model.CreateModelPart("dummy", 2);
+    model_part.AddNodalSolutionStepVariable(DT_TEMPERATURE);
```
WPK4FEM: What is the purpose of this line for this test?

applications/GeoMechanicsApplication/tests/cpp_tests/test_generalized_newmark_T_scheme.cpp
```diff
@@ -91,11 +54,10 @@ KRATOS_TEST_CASE_IN_SUITE(ForMissingNodalDof_CheckBackwardEulerQuasistaticTSchem
...
-KRATOS_TEST_CASE_IN_SUITE(ForMissingDtTemperatureSolutionStepVariable_CheckBackwardEulerQuasistaticTScheme_Throws,
+KRATOS_TEST_CASE_IN_SUITE(ForMissingDtTemperatureSolutionStepVariable_CheckNewmarkTScheme_Throws,
                           KratosGeoMechanicsFastSuite)
 {
-    auto scheme = CreateValidScheme();
-
+    GeneralizedNewmarkTScheme<SparseSpaceType, LocalSpaceType> scheme(0.75);
     Model model;
     auto& model_part = model.CreateModelPart("dummy", 2);
     model_part.AddNodalSolutionStepVariable(TEMPERATURE);
```
WPK4FEM: What is the use of this line for this test?

applications/GeoMechanicsApplication/tests/cpp_tests/test_generalized_newmark_T_scheme.cpp
```diff
@@ -140,7 +101,7 @@ KRATOS_TEST_CASE_IN_SUITE(ThermalSchemeUpdate_SetsDtTemperature, KratosGeoMechan
     node.FastGetSolutionStepValue(DT_TEMPERATURE, 1) = previous_dt_temperature;
 
     // This is the expected value as calculated by the UpdateVariablesDerivatives
-    constexpr double expected_value = 7.0/3.0;
+    constexpr double expected_value = 7.0 / 3.0;
```
WPK4FEM: This far from the place where this variable is used. Further, writing 7.0 / 3.0 in the KRATOS_EXPECT_DOUBLE_EQ statement is equally clear.

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_Pw_scheme.cpp
```diff
@@ -0,0 +1,145 @@
...
+    constexpr double delta_time = 2.0;
+
+    model_part.GetProcessInfo()[DELTA_TIME] = delta_time;
+    Node& node = model_part.Nodes()[0];
+    node.FastGetSolutionStepValue(WATER_PRESSURE) = current_pressure;
+    node.FastGetSolutionStepValue(WATER_PRESSURE, 1) = previous_temperature;
+    node.FastGetSolutionStepValue(DT_WATER_PRESSURE, 1) = previous_dt_temperature;
+
+    // This is the expected value as calculated by the UpdateVariablesDerivatives
+    constexpr double expected_value = 7.0/3.0;
```
WPK4FEM: Far away from its use.

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_Pw_scheme.cpp
```diff
@@ -0,0 +1,145 @@
...
+    ModelPart& model_part = CreateValidModelPart(model);
+
+    constexpr double current_pressure = 10.0;
+    constexpr double previous_temperature = 5.0;
+    constexpr double previous_dt_temperature = 3.0;
+    constexpr double delta_time = 2.0;
+
+    model_part.GetProcessInfo()[DELTA_TIME] = delta_time;
+    Node& node = model_part.Nodes()[0];
+    node.FastGetSolutionStepValue(WATER_PRESSURE) = current_pressure;
```
WPK4FEM: (WATER_RPESSURE, 0) for my faster understanding.

applications/GeoMechanicsApplication/tests/test_generalized_newmark_scheme.cpp
```diff
@@ -0,0 +1,34 @@
...
+namespace Kratos::Testing {
+
+KRATOS_TEST_CASE_IN_SUITE(ForInvalidTheta_CheckNewmarkScheme_Throws, KratosGeoMechanicsFastSuite)
+{
+    constexpr double invalid_theta = -2.0;
+    using SchemeType = GeneralizedNewmarkScheme<SparseSpaceType, LocalSpaceType>;
+
+    KRATOS_EXPECT_EXCEPTION_IS_THROWN(
+        SchemeType scheme(invalid_theta, TEMPERATURE, DT_TEMPERATURE, DT_TEMPERATURE_COEFFICIENT),
+        "Theta must be larger than zero, but got -2")
```
WPK4FEM: 0<= Theta <= 1 ?

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_dynamic_U_Pw_scheme.hpp
rfaasse: Agreed, I would really like to do something similar for the second order part, but that's indeed too much to do in this PR

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
rfaasse: What I did here, is format the file using the clang-format settings that are found in the kratos wiki. However, since it's not in the style guide, I think we can still change this if we would want to

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
avdg81: I agree that when a derived class must provide an implementation for a particular member function, the base class shouldn't provide an empty implementation. However, I would propose to make such member functions pure virtual. In that way, a compiler error will be raised when we don't comply with this requirement. By raising a runtime error, we would put the burden on the end-user rather than on the programmer.

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -25,6 +25,19 @@ class GeoMechanicsTimeIntegrationScheme : public Scheme<TSparseSpace, TDenseSpac
     using LocalSystemVectorType = typename BaseType::LocalSystemVectorType;
     using LocalSystemMatrixType = typename BaseType::LocalSystemMatrixType;
 
+    int Check(const ModelPart& rModelPart) const override
+    {
+        KRATOS_TRY
+
+        Scheme<TSparseSpace, TDenseSpace>::Check(rModelPart);
```
avdg81, line 32: We can make this slightly more readable:
```suggestion
        BaseType::Check(rModelPart);
```

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -331,14 +325,16 @@ class GeoMechanicsTimeIntegrationScheme : public Scheme<TSparseSpace, TDenseSpac
             << rModelPart.GetBufferSize() << std::endl;
     }
 
-    void CheckSolutionStepsData(const Node& r_node, const Variable<double>& variable) const
+    template <class T>
+    void CheckSolutionStepsData(const Node& r_node, const Variable<T>& variable) const
```
avdg81: A minor style issue:
```suggestion
    void CheckSolutionStepsData(const Node& rNode, const Variable<T>& rVariable) const
```

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -331,14 +325,16 @@ class GeoMechanicsTimeIntegrationScheme : public Scheme<TSparseSpace, TDenseSpac
...
     {
         KRATOS_ERROR_IF_NOT(r_node.SolutionStepsDataHas(variable))
             << variable.Name() << " variable is not allocated for node "
             << r_node.Id() << std::endl;
     }
 
-    void CheckDof(const Node& r_node, const Variable<double>& variable) const
+    template <class T>
+    void CheckDof(const Node& r_node, const Variable<T>& variable) const
```
avdg81: And the same style issue is found here:
```suggestion
    void CheckDof(const Node& rNode, const Variable<T>& rVariable) const
```

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -292,28 +302,12 @@ class GeoMechanicsTimeIntegrationScheme : public Scheme<TSparseSpace, TDenseSpac
...
-            // Update Displacement and Pressure (DOFs)
-            for (typename DofsArrayType::iterator it_dof = dofs_begin;
-                 it_dof != dofs_end; ++it_dof) {
-                if (it_dof->IsFree())
-                    it_dof->GetSolutionStepValue() +=
-                        TSparseSpace::GetValue(Dx, it_dof->EquationId());
+        block_for_each(rDofSet, [&Dx](auto& dof) {
+            if (dof.IsFree()) {
+                dof.GetSolutionStepValue() +=
+                    TSparseSpace::GetValue(Dx, dof.EquationId());
```
avdg81, line 308: I like this cleanup a lot. Well done!

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
```diff
@@ -354,6 +350,17 @@ class GeoMechanicsTimeIntegrationScheme : public Scheme<TSparseSpace, TDenseSpac
...
     }
 
+    virtual void CheckAllocatedVariables(const ModelPart& rModelPart) const
+    {
+        // intentionally empty
+    }
+
+    virtual void UpdateScalarTimeDerivative(Node& rNode,
+                                            const Variable<double>& variable,
+                                            const Variable<double>& dt_variable) const {
```
avdg81: A minor style issue:
```suggestion
                                            const Variable<double>& rVariable,
                                            const Variable<double>& rDtVariable) const {
```

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_scheme.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+#include "geomechanics_time_integration_scheme.hpp"
+
+namespace Kratos {
+
+template <class TSparseSpace, class TDenseSpace>
+class GeneralizedNewmarkScheme
+    : public GeoMechanicsTimeIntegrationScheme<TSparseSpace, TDenseSpace> {
+public:
+    explicit GeneralizedNewmarkScheme(double theta,
```
avdg81: I feel that marking this constructor `explicit` is not really needed here, since we can't call it with a single argument (i.e. it is not possible to perform implicit conversions).

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
rfaasse: I agree we should find a more logical place for it in the future

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
rfaasse: In this part I only changed formatting. Since there is no unit test for the nodal smoothing specifically (that's quite involved to produce), I would propose to have a look at that in a separate issue

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_damped_U_Pw_scheme.hpp
rfaasse: Agreed, but the function returns an int (and is located in core), so unless we want to convert it (implicitly or explicitly) to a SizeType, we can't change this.

applications/GeoMechanicsApplication/tests/cpp_tests/test_backward_euler_T_scheme.cpp
rfaasse: This one can be removed, good catch

applications/GeoMechanicsApplication/tests/cpp_tests/test_generalized_newmark_T_scheme.cpp
rfaasse: If temperature is not added, the check is going to throw with the message that temperature is missing instead of what I want to test for (DT_TEMPERATURE)

applications/GeoMechanicsApplication/tests/test_generalized_newmark_scheme.cpp
rfaasse: I took the original restrictions here, since if theta = 0.0, we will end up with a division by 0. Is there a good reason to have 1.0 as the upper limit (same would hold then for beta and/or gamma)?

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_Pw_scheme.cpp
rfaasse: Agreed, added the ', 0' everywhere I could find it in the tests

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_Pw_scheme.cpp
rfaasse: Moved

applications/GeoMechanicsApplication/tests/cpp_tests/test_generalized_newmark_T_scheme.cpp
rfaasse: Agreed, moved

applications/GeoMechanicsApplication/tests/cpp_tests/test_backward_euler_Pw_scheme.cpp
rfaasse: Yes, removed

applications/GeoMechanicsApplication/tests/cpp_tests/test_backward_euler_Pw_scheme.cpp
rfaasse: Done (and everywhere else I could find it)

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_scheme.hpp
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+protected:
+    inline void UpdateVariablesDerivatives(ModelPart& rModelPart) override
+    {
+        KRATOS_TRY
+
+        block_for_each(rModelPart.Nodes(), [&](Node& rNode) {
+            this->UpdateScalarTimeDerivative(rNode, mVariable, mDeltaTimeVariable);
+        });
```
avdg81: I have compared this implementation with the one in `BackwardEulerScheme::UpdateVariablesDerivatives` and I would make them identical:
```suggestion
        block_for_each(rModelPart.Nodes(), [this](Node& rNode) {
            this->UpdateScalarTimeDerivative(rNode, mVariable, mDeltaTimeVariable);
        });
```

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
rfaasse: Nice suggestion!

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Discussed with Anne, we'll probably make it purely virtual, such that it is checked compile time if the derived classes have implemented this function

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Discussed with Anne, we'll probably make it purely virtual, such that it is checked compile time if the derived classes have implemented this function

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_scheme.hpp
rfaasse: Agreed, done

applications/GeoMechanicsApplication/custom_strategies/schemes/backward_euler_scheme.hpp
rfaasse: Agreed, changed!

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
avdg81, line 1: A remark regarding the ordering of some member functions: `SetTimeFactors`, `UpdateVariablesDerivatives`, `CheckAllocatedVariables`, and `UpdateScalarTimeDerivative` appear in different orders here and in files `generalized_newmark_scheme.hpp` and `backward_euler_scheme.hpp`. To compare their implementations, it may be convenient to apply the same ordering in all of these files.

rfaasse (commented):
Thanks for the review @WPK4FEM, implemented most of the suggestions (some were a bit out of scope in my opinion, so let me know what you think)

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_dynamic_U_Pw_scheme.hpp
avdg81: I totally agree. Great work!

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
```diff
@@ -186,32 +144,36 @@ class NewmarkQuasistaticUPwScheme : public GeoMechanicsTimeIntegrationScheme<TSp
     {
         KRATOS_TRY
 
-        //Update Acceleration, Velocity and DtPressure
+        // Update Acceleration, Velocity and DtPressure
         block_for_each(rModelPart.Nodes(), [&](Node& rNode) {
```
avdg81: I would only capture `this` rather than "everything".

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_damped_U_Pw_scheme.hpp
avdg81: We could, however, commit to the type returned by `ParallelUtilities::GetNumThreads` (which is now an `int`, but it may be improved in the future). I would also make it `const`:
```suggestion
        const auto num_threads = ParallelUtilities::GetNumThreads();
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_backward_euler_Pw_scheme.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+    scheme.Initialize(model_part);
+    ModelPart::DofsArrayType dof_set;
+    CompressedMatrix A;
+    Vector Dx;
+    Vector b;
+    scheme.Predict(model_part, dof_set, A, Dx, b);
+
+    constexpr double expected_dt_temperature = 0.25;
+    KRATOS_EXPECT_DOUBLE_EQ(p_node->FastGetSolutionStepValue(DT_WATER_PRESSURE),
```
avdg81: Perhaps you may want to add the (optional) buffer index here as well? So:
```suggestion
    KRATOS_EXPECT_DOUBLE_EQ(p_node->FastGetSolutionStepValue(DT_WATER_PRESSURE, 0),
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_backward_euler_Pw_scheme.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+    scheme.Initialize(model_part);
+
+    KRATOS_EXPECT_TRUE(scheme.SchemeIsInitialized())
+    KRATOS_EXPECT_DOUBLE_EQ(model_part.GetProcessInfo()[DT_PRESSURE_COEFFICIENT],
+                            1.0 / delta_time);
+}
+
+KRATOS_TEST_CASE_IN_SUITE(ForMissingNodalDof_CheckBackwardEulerPwScheme_Throws,
+                          KratosGeoMechanicsFastSuite)
```
avdg81, line 76: Just a remark: apparently, the other tests don't suffer from the fact that no water pressure degree of freedom was added to the node. It's just that when `Check` is called and the DoF is missing, it will throw an exception. That surprises me a bit.

applications/GeoMechanicsApplication/tests/cpp_tests/test_backward_euler_T_scheme.cpp
```diff
@@ -49,8 +49,71 @@ KRATOS_TEST_CASE_IN_SUITE(BackwardEulerScheme_UpdatesVariablesDerivatives_WhenPr
...
+}
+
+KRATOS_TEST_CASE_IN_SUITE(ForMissingTemperatureSolutionStepVariable_CheckBackwardEulerTScheme_Throws,
+                          KratosGeoMechanicsFastSuite)
+{
+    BackwardEulerTScheme<SparseSpaceType, LocalSpaceType> scheme;
+
+    Model model;
+    auto& model_part = model.CreateModelPart("dummy", 2);
+    auto p_node = model_part.CreateNewNode(0, 0.0, 0.0, 0.0);
```
avdg81, line 112: You may want to add variable `DT_TEMPERATURE` (to be consistent with test case `ForMissingWaterPressureSolutionStepVariable_CheckBackwardEulerPwScheme_Throws`):
```suggestion
    model_part.AddNodalSolutionStepVariable(DT_TEMPERATURE);
    auto p_node = model_part.CreateNewNode(0, 0.0, 0.0, 0.0);
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_Pw_scheme.cpp
```diff
@@ -0,0 +1,141 @@
...
+using LocalSpaceType = UblasSpace<double, Matrix, Vector>;
+
+namespace Kratos::Testing {
+
+namespace {
+NewmarkQuasistaticPwScheme<SparseSpaceType, LocalSpaceType> CreateValidScheme()
+{
+    NewmarkQuasistaticPwScheme<SparseSpaceType, LocalSpaceType> result(0.75);
+    return result;
+}
```
avdg81: This function has been inlined in one of the other test files. You may consider to do the same thing here.

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_Pw_scheme.cpp
```diff
@@ -0,0 +1,141 @@
...
+    scheme.Initialize(model_part);
+    scheme.Predict(model_part, dof_set, A, Dx, b);
+
+    // This is the expected value as calculated by the UpdateVariablesDerivatives
+    KRATOS_EXPECT_DOUBLE_EQ(node.FastGetSolutionStepValue(DT_WATER_PRESSURE, 0), 7.0 / 3.0);
+}
+
+KRATOS_TEST_CASE_IN_SUITE(InitializeNewmarkPwScheme_SetsTimeFactors, KratosGeoMechanicsFastSuite)
+{
+    auto scheme = CreateValidScheme();
```
avdg81: When you inline this function, you can move `constexpr double theta = 0.75;` to the top of the test body and use it when creating the scheme. That will make it more obvious that the numbers are correct.

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_quasistatic_U_Pw_scheme.cpp
```diff
@@ -42,7 +42,6 @@ class NewmarkQuasistaticUPwSchemeTester {
     void CreateValidModelPart()
     {
         auto& result = mModel.CreateModelPart("dummy", 2);
-        result.AddNodalSolutionStepVariable(TEMPERATURE);
```
avdg81, line 45: Good catch!

avdg81 (commented):
This is a very nice step forwards again. Thank you very much for the hard work that you have done Richard! The structure has improved a lot and it makes much more sense now.

I only have several minor remarks, nothing blocking.

applications/GeoMechanicsApplication/custom_strategies/schemes/generalized_newmark_scheme.hpp
rfaasse: Agreed, remnant from when the only parameter was theta

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Good suggestion, made it uniform across the schemes that implement these functions

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: I actually prefer to keep it like this: now it is clear at a glance (without having to look at the definition) which function is called. We could discuss (we could also delete the base check, since at this moment it just returns 0 and funnily enough has a KRATOS_TRY and CATCH around it).

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Thanks!

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Done

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Done

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: This is done, so now there is no empty function anymore in the base class

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: Done

applications/GeoMechanicsApplication/custom_strategies/schemes/geomechanics_time_integration_scheme.hpp
rfaasse: This is done, so now there is no empty function anymore in the base class

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
rfaasse: Done!

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_damped_U_Pw_scheme.hpp
rfaasse: Good point, changed

applications/GeoMechanicsApplication/tests/cpp_tests/test_backward_euler_Pw_scheme.cpp
rfaasse: Ah you found one more I missed (I thought I caught them all 😄 )

applications/GeoMechanicsApplication/tests/cpp_tests/test_backward_euler_Pw_scheme.cpp
rfaasse: Indeed, if check isn't called, it won't throw (at least when the DoF is not used in the function that's tested in that particular test). I hope everywhere these schemes are used, the check is actually done, but I don't know. I'm not sure how easy it would be to change this way of checking (apart from just doing the check at the start of every public function, which I don't know if I like)

applications/GeoMechanicsApplication/tests/cpp_tests/test_backward_euler_T_scheme.cpp
rfaasse: I removed this line due to the review comment of WPK, which I agree with. I prefer to remove the same line from the Pw version (and the newmark T/Pw equivalents)

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_Pw_scheme.cpp
rfaasse: Fine with me, done

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_Pw_scheme.cpp
rfaasse: Done

rfaasse (commented):
Thanks both @avdg81 and @WPK4FEM for the thorough reviews. I processed the comments, so should be ready for re-review!

applications/GeoMechanicsApplication/tests/cpp_tests/test_generalized_newmark_T_scheme.cpp
WPK4FEM: Understood.

WPK4FEM (approved):
Read through once more. Looks like a consistent set of things to me.

PR #11809, opened 2023-11-15
[Geo] Orchestrator-based multistage

rubenzorrilla:
**📝 Description**
This is a draft PR for the discussion of the more delicate changes required for the upgrade of the current multistage capabilities of the `GeoMechanicsApplication` to use the `Orchestrator`-based multistage from `KratosCore`.

For the moment, the unique thing would be to rely on the standard import mechanism, that is to say, if the model part is already in the model set
```json
"model_import_settings" : {
    "input_type": "use_input_model_part"
}
```
in the `solver_settings` of the project parameters.


applications/GeoMechanicsApplication/python_scripts/geomechanics_solver.py
```diff
@@ -190,8 +187,7 @@ def GetMinimumBufferSize(self):
     def ImportModelPart(self):
         """This function imports the ModelPart
         """
-        if self.solver_imports_model_part:
-            self._ImportModelPart(self.main_model_part, self.settings["model_import_settings"])
+        self._ImportModelPart(self.main_model_part, self.settings["model_import_settings"])
```
RiccardoRossi, line 193: the idea is to use the "standard" Kratos importing mechanism

RiccardoRossi: actually i think that this would correspond the mechanism used in the base_analysis_stage, so prabably we could directly remove the reimplementation of this function

rubenzorrilla: > actually i think that this would correspond the mechanism used in the base_analysis_stage, so prabably we could directly remove the reimplementation of this function

We can't since the base Python solver throws an error in the  `ImportModelPart`.

What we can eventually consider in here is to substitute it by `pass`. However, this will enforce to always import the mesh at the `AnalysisStage` level, breaking retrocompatibility of already existent (not necessarily multistage) cases. Having said this, I'd keep it as it is right now, at least for a while. In the future we can throw a deprecation warning in the `_ImportModelPart` so we deprecate the mechanism at once for the whole Kratos.

avdg81: Please note that the suggested changes break the way we currently carry out staged analysis. As you may have noticed, this branch has two test failures. We feel that the more gentle approach may be to duplicate one of our multi-stage tests and change it such that is uses your multi-stage approach. In this way, we don't break our production code, but we do create an opportunity to get some hands-on experience with your multi-stage approach. What do you think?

I will try to create a simple test by myself (based on an existing multi-stage test), but I may or may not need a bit of help to get it working. When I do, I will contact you by email. Thanks.

rubenzorrilla: Sure @avdg81 (this is why we make it draft). We created the PR to have something to discuss on top of. I think that the best is to arrange a short meeting to discuss precisely about this, how to do the changes without breaking your current stuff.

I already have a working case so I'd suggest to arrange a meeting before doing any further implementation or test (either from our side or yours).

rubenzorrilla:
Now that #11807 is already merged we can retake this.

applications/GeoMechanicsApplication/python_scripts/geomechanics_solver.py
mcgicjn2: @avdg81 @rubenzorrilla Can we restart this conversation here.

rubenzorrilla: Sure :+1:

rfaasse: @rubenzorrilla As far as I understand, we would need to migrate our multi-stage analysis tests to use the new workflow. This is something we could get started with (looking at @mcgicjn2 for a possible timeline there). We would need some time to communicate the change to users of the geomechanics application and deprecate the old mechanism.

Another question from our side is if the changes in kratos related to the orchestrator could have an influence on a C++-only workflow we have, which uses a similar multi-stage approach as these failing python tests? For this specific PR, there are no risks since it's python-only, but are there any other changes planned in the C++ core?

Possibilities for having (part of) this powerful concept of orchestrator-based analysis available from C++ would be a separate discussion for later, I can imagine.

rubenzorrilla: > @rubenzorrilla As far as I understand, we would need to migrate our multi-stage analysis tests to use the new workflow. This is something we could get started with (looking at @mcgicjn2 for a possible timeline there). We would need some time to communicate the change to users of the geomechanics application and deprecate the old mechanism.
> 

Very good. I'll try to recover the geomechanics use case I used during the development just in case it is of help for this. Besides, I would also consider @avdg81 suggestion of temporarily duplicating the tests (keep current ones and add new ones with the standard mechanism).

> Another question from our side is if the changes in kratos related to the orchestrator could have an influence on a C++-only workflow we have, which uses a similar multi-stage approach as these failing python tests? For this specific PR, there are no risks since it's python-only, but are there any other changes planned in the C++ core?
> 

No. Multistage is conceived to occur at Python level only. Having said this, I think that your current C++ workflow shouldn't be affected by the new one.  

> Possibilities for having (part of) this powerful concept of orchestrator-based analysis available from C++ would be a separate discussion for later, I can imagine.

Yes. At some point we spoke with @mcgicjn2 about this. The idea is that the official multistage must remain at the Python level. Nevertheless, and considering your current needs, we can give you support on creating your custom C++ implementation emulating the Python official one.

rubenzorrilla:
Here it is the example I was referring to in my previous comment [reexamolenofproblemtypeapplication_cimne.zip](https://github.com/KratosMultiphysics/Kratos/files/15217203/reexamolenofproblemtypeapplication_cimne.zip).

I note that the checkpointing and restart mechanism, which was working the last time I tried it, is currently not working as the `SmallStrainUPwDiffOrderElement` serialization is not completely implemented. Note that the `mpStressStatePolicy` pointer is neither serialized nor loaded https://github.com/KratosMultiphysics/Kratos/blob/2bbc87f06265282ae3db84905619ec4a3ba7edd5/applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.hpp#L324-L332.
I tried to add it but the `StressStatePolicy` class lacks the serialization implementation so I guess that this is a problem affecting not only this element in the application.

rfaasse:
> Here it is the example I was referring to in my previous comment [reexamolenofproblemtypeapplication_cimne.zip](https://github.com/KratosMultiphysics/Kratos/files/15217203/reexamolenofproblemtypeapplication_cimne.zip).
> 
> I note that the checkpointing and restart mechanism, which was working the last time I tried it, is currently not working as the `SmallStrainUPwDiffOrderElement` serialization is not completely implemented. Note that the `mpStressStatePolicy` pointer is neither serialized nor loaded.
> 
> https://github.com/KratosMultiphysics/Kratos/blob/2bbc87f06265282ae3db84905619ec4a3ba7edd5/applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.hpp#L324-L332
> 
> 
> I tried to add it but the `StressStatePolicy` class lacks the serialization implementation so I guess that this is a problem affecting not only this element in the application.

Thank you for flagging this! The stress state policy was indeed a newly added component. Since we don't have tests yet for components using serialization, this one slipped through. The issue here is that the mpStressStatePolicy is a pointer to the StressState interface and it seems we don't have a mechanism to instantiate derived classes from serialization (except for using the registry). We'll investigate how to best implement this at our end.

applications/GeoMechanicsApplication/python_scripts/geomechanics_solver.py
rfaasse: Clear, thank you! Indeed it's probably best to temporarily have the two mechanism both working correctly in this transition period (if this is possible). We'll add some issues to our board, which we can pick up after prioritization by @mcgicjn2!

RiccardoRossi:
i have the feeling that this is outdataed. Can you guys please check and eventually close?

rfaasse:
> i have the feeling that this is outdataed. Can you guys please check and eventually close?

Indeed, in the meantime we have created an orchestrator-based test (see https://github.com/KratosMultiphysics/Kratos/pull/13057). We have since worked on the serialization of our classes to be able to use the major benefit of checkpoints and we will have a look at the import mechanisms in more detail, but I'll close this PR for now.

Issue #11804, opened 2023-11-15
[CI] [Custom, Clang] Memory Corruption in MeshMovingApplication


matekelemen:
The `Ubuntu Custom clang` CI job broke a few days ago, and is blocking at least #11796 #11798 #11799. Unfortunately, I cannot reproduce the error locally, but I'm guessing it's a buffer overflow or using a dangling ref/pointer, which will be a pain to find.

An example log: https://github.com/KratosMultiphysics/Kratos/actions/runs/6851062547/job/18653675480?pr=11798

I suspect this is a combination of at least 2 PRs that completed their CI runs and merging without ever pulling the latest master. @roigcarlo Is it possible to invalidate CI runs if master got updated in the meantime?

roigcarlo:
No its imposible, I cannot force all branches to update to the current master unless there is a conflict.

I will take a look

philbucher:
[you can do that :)](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/keeping-your-pull-request-in-sync-with-the-base-branch)
![image](https://github.com/KratosMultiphysics/Kratos/assets/25484702/6fff82a2-bb6b-474c-9d23-f98654fd0bbf)
(Pic is from CoSimIO)

But in the past we decided against it because it will require more CI runs
I think it would make sense though

roigcarlo:
Aaa didn't know. Done :)

roigcarlo:
Respect on the error. I believe I found the problem but I am not sure how to fix or why it startet to appear now.

Seems the problem is located in the `test_parametric`, more specifically while calling the `MoveModelPart` function:

```c++
void MoveModelPart(
    ModelPart& rModelPart,
    const LinearTransform& rTransform)
{
    KRATOS_TRY

    block_for_each(
        rModelPart.Nodes(),
        [&rTransform](Node& rNode){
            const array_1d<double,3>& initial_position = rNode.GetInitialPosition();
            noalias(rNode.GetSolutionStepValue(MESH_DISPLACEMENT)) = rTransform.Apply(initial_position) - initial_position;
                                                                        //      ^^^^^ --- This Call
        });

    KRATOS_CATCH("");
}
```

While rTransform is a `LinearTransform`, there is no problem, but when is an instance of `ParametricLinearTransform`, the apply function is defined as:


```c++
[...]

PR #11811, opened 2023-11-15
[MeshMoving] Fix Race Condition In MoveMesh

matekelemen:
## Main change
Move the `ParametricAffineTransform` in `block_for_each` from a lambda capture to TLS, because it's not thread-safe. The related discussion is in #11804.

## Other changes
- cosmetic changes
- `LinearTransform` was a misnomer (because it also involved a translation) so I renamed it to `AffineTransform`


applications/MeshMovingApplication/custom_utilities/move_mesh_utilities.cpp
```diff
@@ -115,17 +116,18 @@ void MoveModelPart(
...
         rModelPart.Nodes(),
-        [&rTransform, time](Node& rNode){
+        rTransform,
+        [time](Node& rNode, ParametricAffineTransform& rTLSTransform){
             const array_1d<double,3>& initial_position = rNode.GetInitialPosition();
-            noalias(rNode.GetSolutionStepValue(MESH_DISPLACEMENT)) = rTransform.Apply(
+            noalias(rNode.GetSolutionStepValue(MESH_DISPLACEMENT)) = rTLSTransform.Apply(
                 initial_position,
                 time,
                 rNode.X0(),
```
matekelemen, line 133: this is the main change, the rest is style/docs/renaming

applications/MeshMovingApplication/custom_utilities/affine_transform.h
```diff
@@ -0,0 +1,145 @@
...
+//   _|\_\_|  \__,_|\__|\___/ ____/
+//                   Multi-Physics
+//
+//  License:        BSD License
+//	                Kratos default license: kratos/license.txt
+//
+//  Main Authors:   Máté Kelemen
+//
+
+#ifndef KRATOS_MESH_MOVING_LINEAR_TRANSFORMATION_INCLUDED
```
loumalouomega: pragma once

RiccardoRossi (dismissed):
Overall i am ok with the change, however i see you rename methods so this makes it an API changer.

are u sure those methods are not used by other people?
If that is so (no impact on other people) than just merge...otherwise let's discuss the way forward to minimize impact onto other people

matekelemen:
> Overall i am ok with the change, however i see you rename methods so this makes it an API changer.
> 
> are u sure those methods are not used by other people? If that is so (no impact on other people) than just merge...otherwise let's discuss the way forward to minimize impact onto other people

I'm not sure how many people are using it. I originally wrote these ~3 years ago for @mpentek. Maybe he has a better idea of other potential users.

philbucher (dismissed):
I dont think we need the backward compatibility

loumalouomega (approved):

philbucher (approved):

roigcarlo:
Can we merge?

RiccardoRossi:
go ahead

PR #11816, opened 2023-11-16
[GeoMechanicsApplication] max_delta_time_factor taken into account in the c++ adaptive time incrementor

### LINKED ISSUE #11815: [GeoMechanicsApplication] max_delta_time_factor not taken into account in the c++ time incrementor
There is a small discrepancy between the c++ and python behavior for settlement. While testing with Zheng, we found that the time_stepping -> max_delta_time_factor is not taken into account in the c++ routine, but it is used in python to limit the time step size. If not defined in the project parameters, python will set it to 1000.0

WPK4FEM:
**📝 Description**
the max_delta_time_factor for the adaptive time incrementor is available through python, but not in c++

**🆕 Changelog**
adaptive_time_incrementor class extended with the max_delta_time_factor
2 unit tests added in test_time_incrementor.cpp

### DIFF
--- a/applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.cpp
+++ b/applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.cpp
@@ -80,8 +80,7 @@ void AdaptiveTimeIncrementor::PostTimeStepExecution(const TimeStepEndState& rRes
     }
     else if (rResultantState.Converged() &&
              (rResultantState.num_of_iterations < mMinNumOfIterations)) {
-        mDeltaTime *= mIncreaseFactor;
-        mDeltaTime = std::min(mDeltaTime,mMaxDeltaTime);
+        mDeltaTime = std::min(mDeltaTime * mIncreaseFactor, mMaxDeltaTime);
     }
 
     // Avoid incrementing the time beyond the end time
--- a/applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.cpp
+++ b/applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.cpp
@@ -26,6 +26,7 @@ AdaptiveTimeIncrementor::AdaptiveTimeIncrementor(double      StartTime,
                                                  std::size_t MaxNumOfCycles,
                                                  double      ReductionFactor,
                                                  double      IncreaseFactor,
+                                                 double      MaxTimeStepFactor,
                                                  std::size_t MinNumOfIterations,
                                                  std::size_t MaxNumOfIterations) :
     TimeIncrementor(),
@@ -34,6 +35,7 @@ AdaptiveTimeIncrementor::AdaptiveTimeIncrementor(double      StartTime,
     mMaxNumOfCycles(MaxNumOfCycles),
     mReductionFactor(ReductionFactor),
     mIncreaseFactor(IncreaseFactor),
+    mMaxDeltaTime(MaxTimeStepFactor*mDeltaTime),
     mMinNumOfIterations(MinNumOfIterations),
     mMaxNumOfIterations(MaxNumOfIterations)
 {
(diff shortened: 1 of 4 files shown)


applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.cpp
```diff
@@ -78,6 +81,7 @@ void AdaptiveTimeIncrementor::PostTimeStepExecution(const TimeStepEndState& rRes
     else if (rResultantState.Converged() &&
              (rResultantState.num_of_iterations < mMinNumOfIterations)) {
         mDeltaTime *= mIncreaseFactor;
+        mDeltaTime = std::min(mDeltaTime,mMaxDeltaTime);
```
avdg81: We could also express this in a single statement:
```suggestion
        mDeltaTime = std::min(mDeltaTime * mIncreaseFactor, mMaxDeltaTime);
```

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
```diff
@@ -69,6 +69,11 @@ double GetIncreaseFactorFrom(const Parameters& rProjectParameters)
     return rProjectParameters["solver_settings"]["increase_factor"].GetDouble();
 }
 
+double GetMaxDeltaTimeFactorFrom(const Parameters& rProjectParameters)
+{
+    return rProjectParameters["solver_settings"]["max_delta_time_factor"].GetDouble();
```
avdg81: Are you sure this is the correct location in the project parameters `.json` file? When I look at lines 116-117 of file `geomechanics_analysis.py`, it appears to me that we are missing an intermediate field: `["solver_settings"]["time_stepping"]["max_delta_time_factor"]`. If that is correct, then we may also want to add a small test for that, since it was not detected by the test suite.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_incrementor.cpp
```diff
@@ -386,4 +397,17 @@ KRATOS_TEST_CASE_IN_SUITE(ReduceUpscaledIncrementToAvoidExceedingEndTime, Kratos
...
+{
+    AdaptiveTimeIncrementorSettings settings;
+    settings.MaxDeltaTimeFactor = 1.5; // lower than the increase factor, this should lead to truncation
+    auto time_incrementor = MakeAdaptiveTimeIncrementor(settings);
+    auto previous_state   = TimeStepEndState{};
+    previous_state.convergence_state = TimeStepEndState::ConvergenceState::converged;
+    previous_state.num_of_iterations = settings.MinNumOfIterations - 1; // this should normally increase the increment
+
+    time_incrementor.PostTimeStepExecution(previous_state);
+    KRATOS_EXPECT_DOUBLE_EQ(1.5 * settings.StartIncrement, time_incrementor.GetIncrement());
```
avdg81: Perhaps the following is slightly clearer?
```suggestion
    KRATOS_EXPECT_DOUBLE_EQ(settings.MaxDeltaTimeFactor * settings.StartIncrement, time_incrementor.GetIncrement());
```

avdg81 (commented):
Nice and clean extension of the adaptive time incrementor. I have spotted one potential (small) mistake. After it has been fixed (or when it has been proven to be correct), I will approve this PR. Thanks for the good work Wijtze Pieter!

applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.cpp
WPK4FEM: Done

applications/GeoMechanicsApplication/custom_workflows/dgeosettlement.cpp
WPK4FEM: Very good find, thank you! This has to do with the inconsistent setup of the .json file. Within ["time_stepping"] only 2 items are defined of which this is one. All the others ( like the decrease factor, increase factor, min number of iterations ) are one level higher. To get a more logical structure, I would like to reorder this at some point.  For now, I have checked by enabeling the settlement workflow test during developing, without this fix it fails.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_incrementor.cpp
WPK4FEM: done

avdg81 (approved):
Nice and clean extension. Let's bring this back to `master`.

PR #11819, opened 2023-11-17
Geo/11771 fix code smells in dgeflowcpp reported by sonarcloud

### LINKED ISSUE #11771: [GeoMechanicsApplication] Fix code smells in `dgeflow.cpp` reported by SonarCloud
Fix code smells being reported by SonarCloud:
https://sonarcloud.io/project/issues?fileUuids=AYR60YSpHgHwR2jhmMjL&resolved=false&types=CODE_SMELL&id=KratosGeomechanics&open=AYksdjdSWGW4NEBuwuOw

Some of the code smells:
- Refactor functions to reduce its Cognitive Complexity from 85 to the 25 allowed
- Function has 10 parameters, which is greater than the 7 authorized
- Reduce the number of nested "break" statements from 2 to 1 authorized
- Refactor this code to not nest more than 3 if|for|do|while|switch statements
- ...

indigocoral:
**📝 Description**
Fixed some code smells in 'dgeflow.cpp' and 'dgeflow.h' files

**🆕 Changelog**

- broken down "ExecuteFlowAnalysis" into separate functions to reduce the Cognitive Complexity
- added 'CriticalHeadInfo' structure to reduce the parameters given to the functions
- added some parameters as a member 
- made some of the functions 'const' 
- replaced 'new' by 'make_shared'
- removed a 'catch' as it was not doing anything
- replaced 'std::cerr' with 'KRATOS_ERROR_IF_NOT(RiverBoundary)' since it was in the if condition: "if (!RiverBoundary)"
- added a function for logging output
- fixed code smell regarding number of nested "break" statements
- built 'CallBackFunctions' structure to reduce code smells
- replaced 2 const references to "std::string" by a "std::string_view"
- replaced the type aliasing to "using" instead of "typedef" 
- checked usages of aliases and removed those not being used
- reduced number of parameters given to 'ExecuteFlowAnalysis' function in "dgeflow.cpp" and in "kratos_external_bindings.cpp"
- modified two cpp test files because of the change in parameters given to the 'ExecuteFlowAnalysis' function in "test_head_extrapolation_flow_workflow.cpp" and "test_GeoMechanicsNewtonRaphsonErosionProcessStrategy.cpp" test


applications/GeoMechanicsApplication/tests/cpp_tests/test_GeoMechanicsNewtonRaphsonErosionProcessStrategy.cpp
```diff
@@ -34,9 +34,14 @@ KRATOS_TEST_CASE_IN_SUITE(ErosionProcessStrategy, KratosGeoMechanicsIntegrationS
...
 
     auto execute = KratosExecute();
-    int status = execute.ExecuteFlowAnalysis(workingDirectory, projectFile, 3, 4, 0.1, "PorousDomain.Left_head",
-                                             &flow_stubs::emptyLog, &flow_stubs::emptyProgress,
-                                             &flow_stubs::emptyLog, &flow_stubs::emptyCancel);
+    Kratos::KratosExecute::CriticalHeadInfo critical_head_info(3, 4, 0.1);
+    Kratos::KratosExecute::CallBackFunctions call_back_functions(&flow_stubs::emptyLog,
+                                                                 &flow_stubs::emptyProgress,
+                                                                 &flow_stubs::emptyLog,
+                                                                 &flow_stubs::emptyCancel);
```
rfaasse: These two structs can be const (same for the other ones in this file and in the other test file)
```suggestion
    const Kratos::KratosExecute::CriticalHeadInfo critical_head_info(3, 4, 0.1);
    const Kratos::KratosExecute::CallBackFunctions call_back_functions(&flow_stubs::emptyLog,
                                                                 &flow_stubs::emptyProgress,
                                                                 &flow_stubs::emptyLog,
                                                                 &flow_stubs::emptyCancel);
```

applications/GeoMechanicsApplication/tests/cpp_tests/test_GeoMechanicsNewtonRaphsonErosionProcessStrategy.cpp
```diff
@@ -34,9 +34,14 @@ KRATOS_TEST_CASE_IN_SUITE(ErosionProcessStrategy, KratosGeoMechanicsIntegrationS
...
+
+    int status = execute.ExecuteFlowAnalysis(workingDirectory, projectFile, critical_head_info, "PorousDomain.Left_head", call_back_functions);
```
rfaasse: Also this int can be const (same holds for the other tests)
```suggestion
    const int status = execute.ExecuteFlowAnalysis(workingDirectory, projectFile, critical_head_info, "PorousDomain.Left_head", call_back_functions);
```

applications/GeoMechanicsApplication/external_bindings/kratos_external_bindings.cpp
```diff
@@ -24,16 +24,17 @@ extern "C"
                                                void __stdcall         reportTextualProgress(const char*),
                                                bool __stdcall         shouldCancel())
     {
+        Kratos::KratosExecute::CriticalHeadInfo critical_head_info(minCriticalHead, maxCriticalHead, stepCriticalHead);
+        Kratos::KratosExecute::CallBackFunctions call_back_functions(logCallback,
+                                                                     reportProgress,
+                                                                     reportTextualProgress,
+                                                                     shouldCancel);
```
rfaasse: These can be const
```suggestion
        const Kratos::KratosExecute::CriticalHeadInfo critical_head_info(minCriticalHead, maxCriticalHead, stepCriticalHead);
        const Kratos::KratosExecute::CallBackFunctions call_back_functions(logCallback,
                                                                     reportProgress,
                                                                     reportTextualProgress,
                                                                     shouldCancel);
```

applications/GeoMechanicsApplication/external_bindings/kratos_external_bindings.cpp
```diff
@@ -24,16 +24,17 @@ extern "C"
...
+
```
rfaasse, line 32: Here we can directly return instance->ExecuteFlowAnalysis( ... ) instead of saving it in errorCode

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.h
```diff
@@ -46,56 +46,109 @@
...
         // Initial Setup
         Model current_model;
         Kernel kernel;
         KratosGeoMechanicsApplication::Pointer geoApp;
-        
+        std::string mWorkingDirectory;
+        std::string mCriticalHeadBoundaryModelPartName;
+        bool mPipingSuccess = false;
+        double mCriticalHead;
+        double mCurrentHead;
```
rfaasse: Here I would give some default value (not sure if these would be sensible?)
```suggestion
        double mCriticalHead = 0.0;
        double mCurrentHead = 0.0;
```

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.h
```diff
@@ -46,56 +46,109 @@
...
-                                 std::equal_to<result_type>, Dof<double> *>
-            DofsArrayType;
+        using result_type = SetIdentityFunction<Dof<double>>;
+
 
         static ConvergenceCriteriaType::Pointer setup_criteria_dgeoflow();
         static LinearSolverType::Pointer setup_solver_dgeoflow();
         static GeoMechanicsNewtonRaphsonErosionProcessStrategyType::Pointer setup_strategy_dgeoflow(ModelPart &model_part);
-        static std::vector<std::shared_ptr<Process>> parseProcess(ModelPart &model_part, Parameters projFile);
+        void parseProcess(ModelPart &model_part, Parameters projFile);
```
rfaasse: I think it would be better to rename this function to ParseProcesses (since it initializes multiple of them and to be in line with the style guide)

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.h
```diff
@@ -46,56 +46,109 @@
...
+        std::string mCriticalHeadBoundaryModelPartName;
+        bool mPipingSuccess = false;
+        double mCriticalHead;
+        double mCurrentHead;
+        bool mExitLoop = false;
+        std::vector<std::shared_ptr<Process>> mProcesses;
+
         void ResetModelParts();
 
         int echoLevel = 1;
```
rfaasse: Let's move this one to all the other members and call it mEchoLevel

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.h
```diff
@@ -46,56 +46,109 @@
...
+
+        void HandleCancellationAndReset(const std::function<void(const char*)>& rLogCallback,
+                                    LoggerOutput::Pointer p_output);
+
 
     private:
         // Initial Setup
         Model current_model;
         Kernel kernel;
         KratosGeoMechanicsApplication::Pointer geoApp;
```
rfaasse: These members do not yet adhere to the style guide
```suggestion
        Model mCurrentModel;
        Kernel mKernel;
        KratosGeoMechanicsApplication::Pointer mpGeoApp;
```

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
-                }
-                CriticalHeadFile << "\t }\n";
-                CriticalHeadFile << "}\n";
+        KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Finished writing result" << std::endl;
+        return 0;
+    }
 
-                // Close the file
-                CriticalHeadFile.close();
+    void KratosExecute::AddNodalSolutionStepVariables(ModelPart& model_part) const
```
rfaasse: To adhere to the style-guide (which is a bit confusing, I agree), model_part should be renamed to rModelPart

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
 
             KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Added DoF" << std::endl;
 
-            std::vector<std::shared_ptr<Process>> processes = parseProcess(model_part, projectfile);
+            parseProcess(model_part, projectfile);
 
             KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Parsed Process Data" << std::endl;
 
-            bool hasPiping = stepCriticalHead != 0;
+            bool hasPiping = criticalHeadInfo.stepCriticalHead != 0;
```
rfaasse: Minor style guide thing: hasPiping -> has_piping

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -254,106 +250,79 @@ namespace Kratos
...
-                                           double                                   maxCriticalHead,
-                                           double                                   stepCriticalHead,
-                                           const std::string&                       rCriticalHeadBoundaryModelPartName,
-                                           const std::function<void(const char*)>&  rLogCallback,
-                                           const std::function<void(double)>&       rReportProgress,
-                                           const std::function<void(const char*)>&  rReportTextualProgress,
-                                           const std::function<bool()>&             rShouldCancel)
+    int KratosExecute::ExecuteFlowAnalysis(std::string_view rWorkingDirectory,
+                                           const std::string& rProjectParamsFileName,
+                                           const CriticalHeadInfo& criticalHeadInfo,
```
rfaasse: Minor style guide comment: criticalHeadInfo -> rCriticalHeadInfo

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
-            bool hasPiping = stepCriticalHead != 0;
+            bool hasPiping = criticalHeadInfo.stepCriticalHead != 0;
 
-            if (rShouldCancel())
+            if (rCallBackFunctions.rShouldCancel())
             {
-                rLogCallback(kratosLogBuffer.str().c_str());
-                Logger::RemoveOutput(p_output);
-                ResetModelParts();
+                HandleCancellationAndReset(rCallBackFunctions.rLogCallback, p_output);
```
rfaasse: I really like you extracted this HandleCancellationAndReset, which you have re-used in several places!

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
-                shared_ptr<Process> RiverBoundary;
-                if (rCriticalHeadBoundaryModelPartName.empty())
-                {
-                    RiverBoundary = FindRiverBoundaryAutomatically(p_solving_strategy, processes);
-                }
-                else
-                {
-                    RiverBoundary = FindRiverBoundaryByName(rCriticalHeadBoundaryModelPartName, processes);
-                }
+            HandleCancellationAndReset(rCallBackFunctions.rLogCallback, p_output);
```
rfaasse: Since here, there is no cancellation, we might need to make the name of the function a bit more general. Maybe HandleCleanUp()?

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
 
-                auto currentProcess = std::static_pointer_cast<GeoFlowApplyConstantHydrostaticPressureProcess>(RiverBoundary);
-                KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "River boundary name: " << currentProcess->GetName() << std::endl;
+    int KratosExecute::ExecuteWithPiping(ModelPart& model_part,
+                                         const Kratos::Parameters& gid_output_settings,
+                                         const CriticalHeadInfo& criticalHeadInfo,
+                                         LoggerOutput::Pointer p_output,
+                                         const CallBackFunctions& rCallBackFunctions)
+    {
+        std::stringstream kratosLogBuffer;
```
rfaasse: kratosLogBuffer can be kratos_log_buffer, since it's a local variable

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
+        KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Critical head search started." << std::endl;
+        KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Critical head min head: " << criticalHeadInfo.minCriticalHead << std::endl;
+        KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Critical head max head: " << criticalHeadInfo.maxCriticalHead << std::endl;
+        KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Critical head step size: " << criticalHeadInfo.stepCriticalHead << std::endl;
+        const auto p_solving_strategy = setup_strategy_dgeoflow(model_part);
+        p_solving_strategy->SetEchoLevel(0);
+        shared_ptr<Process> RiverBoundary;
```
rfaasse: Can be river_boundary to adhere to the style guide

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
-                double criticalHead;
-                double currentHead;
-                bool pipingSuccess = false;
+        GeoOutputWriter writer{gid_output_settings, mWorkingDirectory, model_part};
+        writer.WriteGiDOutput(model_part, gid_output_settings);
+    }
 
-                auto currentProcess = std::static_pointer_cast<GeoFlowApplyConstantHydrostaticPressureProcess>(RiverBoundary);
-                KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "River boundary name: " << currentProcess->GetName() << std::endl;
+    int KratosExecute::ExecuteWithPiping(ModelPart& model_part,
```
rfaasse: We can change this to rModelPart to adhere to the style guide

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
 
-                std::vector<Element *> pipeElements;
-                pipeElements = p_solving_strategy->GetPipingElements();
-                const auto noPipeElements = pipeElements.size();
+        FindCriticalHead(model_part, gid_output_settings, criticalHeadInfo, p_output, RiverBoundary,
+                         p_solving_strategy, rCallBackFunctions);
 
-                int step = 1;
-                const auto maxSteps = static_cast<int>(std::ceil((maxCriticalHead - minCriticalHead) / stepCriticalHead)) + 2;
+        KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Writing result to: " << mWorkingDirectory << "\\criticalHead.json" << std::endl;
```
rfaasse, line 429: I would extract this part (423-443) with the CriticalHeadFile into a function, since it is a very clear singular thing that's going on, which makes sense to group in a single function.

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
-                    if (rShouldCancel())
-                    {
-                        rLogCallback(kratosLogBuffer.str().c_str());
-                        Logger::RemoveOutput(p_output);
-                        ResetModelParts();
-                        return 0;
-                    }
-                }
+        // output critical head_json
+        std::ofstream CriticalHeadFile(mWorkingDirectory + "\\criticalHead.json");
```
rfaasse: To adhere to the style guide, this should be critical_head_file

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
+        else
+        {
+            CriticalHeadFile << "\t\t \"Success\": \"False\"\n";
+        }
+        CriticalHeadFile << "\t }\n";
+        CriticalHeadFile << "}\n";
 
-                // output critical head_json
-                std::ofstream CriticalHeadFile(rWorkingDirectory + "\\criticalHead.json");
+        // Close the file
```
rfaasse: This comment is not really necessary, since the next line CriticalHeadFile.close() describes very clearly what is going on

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
+                                        const Kratos::Parameters& gid_output_settings,
+                                        const CriticalHeadInfo& criticalHeadInfo,
+                                        LoggerOutput::Pointer p_output,
+                                        const shared_ptr<Process>& RiverBoundary,
+                                        const GeoMechanicsNewtonRaphsonErosionProcessStrategyType::Pointer& p_solving_strategy,
+                                        const CallBackFunctions& rCallBackFunctions)
+    {
 
-                KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Finished writing result" << std::endl;
+        auto currentProcess = std::static_pointer_cast<GeoFlowApplyConstantHydrostaticPressureProcess>(RiverBoundary);
```
rfaasse: to adhere to the style guide, this should be current_process

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
+        model_part.AddNodalSolutionStepVariable(NODAL_JOINT_AREA);
+        model_part.AddNodalSolutionStepVariable(NODAL_JOINT_WIDTH);
+        model_part.AddNodalSolutionStepVariable(NODAL_JOINT_DAMAGE);
+    }
+
+    int KratosExecute::FindCriticalHead(ModelPart& model_part,
+                                        const Kratos::Parameters& gid_output_settings,
+                                        const CriticalHeadInfo& criticalHeadInfo,
+                                        LoggerOutput::Pointer p_output,
+                                        const shared_ptr<Process>& RiverBoundary,
```
rfaasse: Since the river boundary is already a shared_ptr, in my opinion it is not necessary to make it a '&' as well. Do you agree @avdg81?

That means I think it should be const std::shared_ptr<Process> pRiverBoundary (also note the p for pointer)

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
+                                        const GeoMechanicsNewtonRaphsonErosionProcessStrategyType::Pointer& p_solving_strategy,
```
rfaasse: Same here, I don't think the '&' is necessary. For the style guide, this should be pSolvingStrategy

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
-                KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Finished writing result" << std::endl;
+        auto currentProcess = std::static_pointer_cast<GeoFlowApplyConstantHydrostaticPressureProcess>(RiverBoundary);
+        KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "River boundary name: " << currentProcess->GetName() << std::endl;
+
+        currentProcess->SetReferenceCoord(criticalHeadInfo.minCriticalHead);
+        mCurrentHead = criticalHeadInfo.minCriticalHead;
+        mCriticalHead = mCurrentHead;
+
+        std::vector<Element *> pipeElements;
+        pipeElements = p_solving_strategy->GetPipingElements();
```
rfaasse: This could be one line (also note the renaming to pipe_elements): 
```suggestion
        auto pipe_elements = p_solving_strategy->GetPipingElements();
```

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
+        const auto noPipeElements = pipeElements.size();
+
+        int step = 1;
+        const auto maxSteps = static_cast<int>(std::ceil((criticalHeadInfo.maxCriticalHead - criticalHeadInfo.minCriticalHead) / criticalHeadInfo.stepCriticalHead)) + 2;
```
rfaasse: This can be max_steps

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
+
+
+        while (!mExitLoop)
```
rfaasse: Is this mExitLoop used outside of this function? I think for this one might be better to keep it a local variable within this function. Maybe we can also discuss if it's necessary (was one of the SQ issues the 'while(true' statement?

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
+        model_part.AddNodalSolutionStepVariable(NODAL_DAMAGE_VARIABLE);
+        model_part.AddNodalSolutionStepVariable(NODAL_JOINT_AREA);
+        model_part.AddNodalSolutionStepVariable(NODAL_JOINT_WIDTH);
+        model_part.AddNodalSolutionStepVariable(NODAL_JOINT_DAMAGE);
+    }
+
+    int KratosExecute::FindCriticalHead(ModelPart& model_part,
+                                        const Kratos::Parameters& gid_output_settings,
+                                        const CriticalHeadInfo& criticalHeadInfo,
+                                        LoggerOutput::Pointer p_output,
```
rfaasse: Also here it can be rModelPart, rGidOutputSettings, rCriticalHeadInfo and pOutput

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
+                break;
             }
 
-            rLogCallback(kratosLogBuffer.str().c_str());
-            Logger::RemoveOutput(p_output);
+            KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Searching at head: " << mCurrentHead << std::endl;
 
-            ResetModelParts();
-            return 0;
+            std::ostringstream currentHeadStream;
```
rfaasse: This could be current_head_stream

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
+            currentHeadStream << std::setprecision(8) << std::noshowpoint << mCurrentHead;
+            std::string currentHeadString = currentHeadStream.str();
```
rfaasse: This can be current_head_string

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
+                HandleCancellationAndReset(rCallBackFunctions.rLogCallback, p_output);
+
+                return 0;
+            }
         }
-        catch (const std::exception &exc)
+        return 0;
+    }
+
+    void KratosExecute::HandleCriticalHeadFound(const CriticalHeadInfo& criticalHeadInfo)
```
rfaasse: Can be rCriticalHeadInfo

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -566,15 +608,14 @@ namespace Kratos
 
         if (!RiverBoundary)
         {
-            std::cerr << "No boundary found with the model part name " << rCriticalHeadBoundaryModelPartName << "." << std::endl;
+            KRATOS_ERROR_IF_NOT(RiverBoundary) << "No boundary found with the model part name " << rCriticalHeadBoundaryModelPartName << "." << std::endl;
             return nullptr;
         }
```
rfaasse: This can be simplified, since KRATOS_ERROR_IF_NOT throws, so this could be changed to just 
KRATOS_ERROR_IF_NOT(RiverBoundary) << "No boundary found with the model part name " << rCriticalHeadBoundaryModelPartName << "." << std::endl;

And remove the if statement and the return nullptr

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -623,21 +657,15 @@ namespace Kratos
...
                 }
             }
         }
 
         if (!RiverBoundary)
         {
-            std::cerr << "No boundary found on the river side at node " << RiverNode << "." << std::endl;
+            KRATOS_ERROR_IF_NOT(RiverBoundary) << "No boundary found on the river side at node " << RiverNode << "." << std::endl;
             return nullptr;
         }
```
rfaasse: Same as earlier comment

This can be simplified, since KRATOS_ERROR_IF_NOT throws, so this could be changed to just
KRATOS_ERROR_IF_NOT(RiverBoundary) << "No boundary found with the model part name " << rCriticalHeadBoundaryModelPartName << "." << std::endl;

And remove the if statement and the return nullptr

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -216,33 +214,31 @@ namespace Kratos
         std::size_t found = name.find_last_of('.');
         std::string subname = name.substr(found + 1);
         ModelPart &part = model_part.GetSubModelPart(subname);
-        processes.push_back(make_shared<ApplyConstantScalarValueProcess>(ApplyConstantScalarValueProcess(part, VOLUME_ACCELERATION_X,
+        mProcesses.push_back(make_shared<ApplyConstantScalarValueProcess>(ApplyConstantScalarValueProcess(part, VOLUME_ACCELERATION_X,
                                                                                                          0.0, 0, ApplyConstantScalarValueProcess::VARIABLE_IS_FIXED)));
```
avdg81: Since `make_shared` receives a list of arguments that it will pass on to the constructor of class `ApplyConstantScalarValueProcess`, there is no need to create a temporary object of this type. In other words, you could simplify this as follows:
```suggestion
        mProcesses.push_back(make_shared<ApplyConstantScalarValueProcess>(part, VOLUME_ACCELERATION_X,
                                                                                                         0.0, 0, ApplyConstantScalarValueProcess::VARIABLE_IS_FIXED));
```

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -597,24 +638,17 @@ namespace Kratos
...
-                    currentModelPart = &current_process->GetModelPart();
-                    try
-                    {
-                        currentModelPart->GetNode(RiverNode);
-                        RiverBoundary = current_process;
-                    }
-                    catch (...)
-                    {
-                    }
+                    currentModelPart->GetNode(RiverNode);
```
rfaasse, line 649: Here we miss the initialization of currentModelPart, so it will be a nullptr when calling it. We still need the `currentModelPart = &current_process->GetModelPart();` which was there before

rfaasse (commented):
Very well done @indigocoral, this cleaned up the code a lot and made it more readable! 

Don't be discouraged by the number of comments, since only one of them is really important (it has to do with an un-initialized pointer of a model part). 

Most of the others are small style/readability improvements and probably 9 out of 10 are on code that was already there, but are easy fixes to make now we're working on it anyway.

All in all, a big step forward for this class!

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
rfaasse: By the way, @indigocoral did this not result in one or more failing tests? Then we might not have a test related to this specific part

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.h
```diff
@@ -46,56 +46,109 @@
...
+                    minCriticalHead(minCriticalHead), maxCriticalHead(maxCriticalHead), stepCriticalHead(stepCriticalHead)
+            {}
+        };
+
+        struct CallBackFunctions
+        {
+            std::function<void(const char*)> rLogCallback;
+            std::function<void(double)> rReportProgress;
+            std::function<void(const char*)> rReportTextualProgress;
+            std::function<bool()> rShouldCancel;
```
avdg81: Since these are no longer references, we should remove the `r` prefix (yes, this is a very annoying result of the needlessly complex style guide).
```suggestion
            std::function<void(const char*)> LogCallback;
            std::function<void(double)> ReportProgress;
            std::function<void(const char*)> ReportTextualProgress;
            std::function<bool()> ShouldCancel;
```

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.h
```diff
@@ -46,56 +46,109 @@
...
+
+            CallBackFunctions(std::function<void(const char*)> rLogCallback,
+                              std::function<void(double)> rReportProgress,
+                              std::function<void(const char*)> rReportTextualProgress,
+                              std::function<bool()> rShouldCancel) :
```
avdg81: And the same thing here:
```suggestion
            CallBackFunctions(std::function<void(const char*)> LogCallback,
                              std::function<void(double)> ReportProgress,
                              std::function<void(const char*)> ReportTextualProgress,
                              std::function<bool()> ShouldCancel) :
```

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.h
```diff
@@ -46,56 +46,109 @@
...
+                    rLogCallback(std::move(rLogCallback)), rReportProgress(std::move(rReportProgress)), rReportTextualProgress(std::move(rReportTextualProgress)),
+                    rShouldCancel(std::move(rShouldCancel))
+            {}
+        };
+
+        int ExecuteFlowAnalysis(std::string_view rWorkingDirectory,
```
avdg81: Again a minor style thing (the parameter is now passed by value and no longer by reference):
```suggestion
        int ExecuteFlowAnalysis(std::string_view WorkingDirectory,
```

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.h
```diff
@@ -46,56 +46,109 @@
...
+                                const std::string& rProjectParamsFileName,
+                                const CriticalHeadInfo& criticalHeadInfo,
+                                std::string_view rCriticalHeadBoundaryModelPartName,
```
avdg81: And the same thing here:
```suggestion
                                std::string_view CriticalHeadBoundaryModelPartName,
```

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -216,33 +214,31 @@ namespace Kratos
         std::size_t found = name.find_last_of('.');
         std::string subname = name.substr(found + 1);
         ModelPart &part = model_part.GetSubModelPart(subname);
-        processes.push_back(make_shared<ApplyConstantScalarValueProcess>(ApplyConstantScalarValueProcess(part, VOLUME_ACCELERATION_X,
+        mProcesses.push_back(make_shared<ApplyConstantScalarValueProcess>(ApplyConstantScalarValueProcess(part, VOLUME_ACCELERATION_X,
                                                                                                          0.0, 0, ApplyConstantScalarValueProcess::VARIABLE_IS_FIXED)));
 
-        processes.push_back(make_shared<ApplyConstantScalarValueProcess>(ApplyConstantScalarValueProcess(part, VOLUME_ACCELERATION_Y, -9.81,
+        mProcesses.push_back(make_shared<ApplyConstantScalarValueProcess>(ApplyConstantScalarValueProcess(part, VOLUME_ACCELERATION_Y, -9.81,
                                                                                                          0, ApplyConstantScalarValueProcess::VARIABLE_IS_FIXED)));
```
avdg81: And the same simplification here:
```suggestion
        mProcesses.push_back(make_shared<ApplyConstantScalarValueProcess>(part, VOLUME_ACCELERATION_Y, -9.81,
                                                                                                         0, ApplyConstantScalarValueProcess::VARIABLE_IS_FIXED));
```

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -216,33 +214,31 @@ namespace Kratos
...
 
-        processes.push_back(make_shared<Process>(ApplyConstantScalarValueProcess(part, VOLUME_ACCELERATION_Z, 0.0,
+        mProcesses.push_back(make_shared<Process>(ApplyConstantScalarValueProcess(part, VOLUME_ACCELERATION_Z, 0.0,
                                                                                  0, ApplyConstantScalarValueProcess::VARIABLE_IS_FIXED)));
```
avdg81: And also here (note the template parameter now must be updated too):
```suggestion
        mProcesses.push_back(make_shared<ApplyConstantScalarValueProcess>(part, VOLUME_ACCELERATION_Z, 0.0,
                                                                                 0, ApplyConstantScalarValueProcess::VARIABLE_IS_FIXED));
```

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
+                ExecuteWithoutPiping(model_part, gid_output_settings);
             }
             else
             {
-                KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Critical head search started." << std::endl;
-                KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Critical head min head: " << minCriticalHead << std::endl;
-                KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Critical head max head: " << maxCriticalHead << std::endl;
-                KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Critical head step size: " << stepCriticalHead << std::endl;
+                ExecuteWithPiping(model_part, gid_output_settings, criticalHeadInfo, p_output, rCallBackFunctions);
+            }
```
avdg81: Now that you have simplified the bodies of the `if` and `else` clauses, I would invert the logic (because negation is slightly harder to understand than not negating something). I.e. (pseudo-code, to show what I mean):
```
if (hasPiping)
{
    ExecuteWithPiping(...);
}
else
{
    ExecuteWithoutPiping(...);
}
```

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
avdg81: Yes, I agree. We could even consider to make it a raw pointer or a reference, since we don't intend to share ownership here. But perhaps it's better to leave that for another PR. The focus here is removing code smells.

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
+                return 0;
+            }
         }
-        catch (const std::exception &exc)
+        return 0;
+    }
+
+    void KratosExecute::HandleCriticalHeadFound(const CriticalHeadInfo& criticalHeadInfo)
+    {
+        if (abs(mCurrentHead - criticalHeadInfo.minCriticalHead) < 1e-9)
```
avdg81: I always prefer to use the math functions from the `std` namespace (to ensure that we don't accidentally call the C-style variants, which don't support overloading). So:
```suggestion
        if (std::abs(mCurrentHead - criticalHeadInfo.minCriticalHead) < 1e-9)
```

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
+
+            return 1;
+        }
+    }
+
+    void KratosExecute::ExecuteWithoutPiping(ModelPart& model_part,
+                                             const Kratos::Parameters& gid_output_settings) const
+    {
+        const auto p_solving_strategy = setup_strategy_dgeoflow(model_part);
+        p_solving_strategy->SetEchoLevel(0);
```
avdg81: I believe these lines have been accidentally duplicated. They are also present in their original location (lines 312-313), but the strategy created there is now no longer used. I would suggest to pass it to this function through the parameter list (i.e. add an additional function parameter), and to not create it here.

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
avdg81: I would actually remove it, since it is an unused variable in this function.

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
+                                         LoggerOutput::Pointer p_output,
+                                         const CallBackFunctions& rCallBackFunctions)
+    {
+        std::stringstream kratosLogBuffer;
+        KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Critical head search started." << std::endl;
+        KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Critical head min head: " << criticalHeadInfo.minCriticalHead << std::endl;
+        KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Critical head max head: " << criticalHeadInfo.maxCriticalHead << std::endl;
+        KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Critical head step size: " << criticalHeadInfo.stepCriticalHead << std::endl;
+        const auto p_solving_strategy = setup_strategy_dgeoflow(model_part);
+        p_solving_strategy->SetEchoLevel(0);
```
avdg81: As for the member function `ExecuteWithoutPiping`, I believe these lines have been accidentally duplicated. I would suggest to pass the originally created strategy through the parameter list, rather than creating a new one here.

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
```diff
@@ -371,187 +340,260 @@ namespace Kratos
...
+                if (element->GetValue(PIPE_ACTIVE))
+                    count += 1;
+            }
+
+            KRATOS_INFO_IF("GeoFlowKernel", this->GetEchoLevel() > 0) << "Open pipe elements: " << count << std::endl;
+
+            if (count == noPipeElements)
+            {
+                HandleCriticalHeadFound(criticalHeadInfo);
+                mExitLoop = true;
```
avdg81: There used to be a `break` here, which means that we would jump out of the loop. Now we continue until the end of loop body. I'm not sure whether that was your intention, because the behavior is different now.

avdg81 (changes requested):

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
indigocoral: No, none of the tests failed because of this.

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
rfaasse: Nicely done, thanks!

applications/GeoMechanicsApplication/custom_workflows/dgeoflow.cpp
rfaasse: Solved in a different way now, nicely done!

rfaasse (approved):
Thanks for making all the changes and cleaning up the code in general! It was a lot of work, but in the end it improved the code a lot!

avdg81 (approved):

PR #11820, opened 2023-11-17
[GeoMechanicsAnalysis] separated start of code cleaning U Pw diff order element

WPK4FEM:
**📝 Description**
Start of reducing code smells in U Pw diff order element

Has been split of from the incremental linear material where this accidentally was included.
SonarCloud now shows 0 codesmells on this.

### DIFF
--- a/applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
+++ b/applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
@@ -28,23 +28,18 @@ namespace Kratos
 // Default Constructor
 SmallStrainUPwDiffOrderElement::SmallStrainUPwDiffOrderElement() : Element() {}
 
-//----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 //Constructor 1
 SmallStrainUPwDiffOrderElement::
     SmallStrainUPwDiffOrderElement( IndexType NewId, GeometryType::Pointer pGeometry ) : Element( NewId, pGeometry ) {}
 
-//----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 //Constructor 2
 SmallStrainUPwDiffOrderElement::
     SmallStrainUPwDiffOrderElement( IndexType NewId,
                                     GeometryType::Pointer pGeometry,
                                     PropertiesType::Pointer pProperties ) : Element( NewId, pGeometry, pProperties ) {}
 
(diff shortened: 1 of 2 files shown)


applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
```diff
@@ -669,27 +614,21 @@ void SmallStrainUPwDiffOrderElement::
...
-            rValues[Index++] = rGeom[i].FastGetSolutionStepValue( VELOCITY_X, Step );
-            rValues[Index++] = rGeom[i].FastGetSolutionStepValue( VELOCITY_Y, Step );
+    for ( SizeType i = 0; i < NumUNodes; ++i ) {
+        rValues[Index] = rGeom[i].FastGetSolutionStepValue( VELOCITY_X, Step );
+        ++Index;
+        rValues[Index] = rGeom[i].FastGetSolutionStepValue( VELOCITY_Y, Step );
+        ++Index;
+        if ( Dim > 2 ) {
+            rValues[Index] = rGeom[i].FastGetSolutionStepValue( VELOCITY_Z, Step );
+            ++Index;
```
rfaasse, line 624: If I'm correct, this was a part which wasn't tested. Do we still want to make these changes then?

applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
```diff
@@ -622,39 +574,32 @@ void SmallStrainUPwDiffOrderElement::
...
+        ++Index;
+        if (Dim > 2) {
+            rResult[Index] = rGeom[i].GetDof( DISPLACEMENT_Z ).EquationId();
+            ++Index;
         }
     }
+    for ( SizeType i = 0; i < NumPNodes; ++i ){
+        rResult[Index] = rGeom[i].GetDof( WATER_PRESSURE ).EquationId();
+        ++Index;
+    }
```
rfaasse, line 597: I like the improvement in this part, let's keep in mind for a next PR that we can extract a function here, since it seems identical for velocity and acceleration

rfaasse (commented):
Thanks for splitting this part out of the other PR, this makes it easier to see what's going on and thanks for incorporating the comments. I still see room for improvement within the files, but I think this is a very nice first step in cleaning this class up.

I have just one question left about changes to a part which might not be sufficiently tested. Other than that, this is ready to be merged

rfaasse (approved):
Thanks for the clean-up, looks good to me!

PR #11825, opened 2023-11-19
Get files changed in PR

philbucher:
This PR is the first step towards improving the CI by not always compiling everything. E.g. not always all apps need to be compiled, or not Cpp tests need to be compiled if only python files were changes. Stuff like this, ofc TBD once we actually want to introduce it

In a nutshell, this PR uses the [github cli](https://cli.github.com/) to get obtain the files modified in the PR. For now I am only printing them to observe it for a while. Eventually we can use them to refine the CI, see above


roigcarlo:
Were is the list printed? the CI only shows a variable :S

matekelemen:
@philbucher can you plz give a quick rundown on what this does?

philbucher:
Sorry I marked too early as ready. I updated the description of the PR

roigcarlo:
Out of curiosity, how are you planning to get the list of what needs to be run based on the modified files? I remember that I tried that once and I just gave up.

I specially has problems with multilevel dependencies. for example:

I modify process pA

There is a script in the application aX that does not depend on pA that uses process pB. process pB uses process pC from app aY  and this process pC (which is pure python) callas the pA that I've modified.

In this case the CI would need to compile:
- Kernel (that's easy)
- App `aY` ( and maybe others) that directly make use of `pA`at python level. Sort of easy as well parsing all the python interface
- App `aX` because one of its python scripts calls the python script form `aY`. Harder, as I have to parse all the python interface and have the python interface of all other apps parsed as well to know the potential call graph

roigcarlo:
Also, consider that this: https://github.com/KratosMultiphysics/Kratos/tree/core/new-compile-pipeline will be merged soon™

philbucher:
> Also, consider that this: https://github.com/KratosMultiphysics/Kratos/tree/core/new-compile-pipeline will be merged soon™

I know, this will not conflict. In any case, I would adapt the developments to it if necessary

philbucher:
Regarding the dependencies: I had similar thoughts.
In the end I want to add a small file in each app, where the dependencies for the CI are specified :) Then the devs can decide themselves what needs to be run.
E.g. for the StructuralMechanicsApp: LinearSolversApp, ConstitutiveLawsApp
For CoSim: StructuralMechanics (and thus also the LinearSolversApp and the ConstLawApp), FluidDynamics, ...


Given that a core change (which usually also happens daily) compiles everything, same as the nightly build I would say we are on the safe side. But I would discuss the details in the future

roigcarlo (approved):

PR #11827, opened 2023-11-20
[CI] Testing general shared PCH

roigcarlo:
**📝 Description**
Trying to reuse the most parsing intensive part of the core into a shared PCH across `Core` targets (possible extendable to `application` targets as well)

Note that this PR will case the core to compile a second time without tests and without marking the entry functions as export.

I am shooting myself in the feet because will make the compilation split more difficult, but if all goes well should reduce compilation times ( up to which extend is what I want to check with the draft, basically of the overhead of compiling the core twice balances out. )

**🆕 Changelog**
- Adding shared PCH


roigcarlo:
Recompiling the core takes all the time reduction in the CI. Closing :/

PR #11832, opened 2023-11-22
[GeoMechanicsApplication] Add progress delegate to time loop executor…

carloslubbers:
… and update it with the end time of each time step


applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
```diff
@@ -106,8 +112,16 @@ public :
         return mCancelDelegate && mCancelDelegate();
     }
 
+    void UpdateProgress(double Time)
```
rfaasse: This function can be const (see also [SonarCloud](https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&types=CODE_SMELL&branch=geo%2Fsettlement-progress-delegate&id=KratosGeomechanics&open=AYv4hfTGfW0A5I6bHWgc), which is a nice tool to find these kinds of 'code-smells')

rfaasse (commented):
Clean implementation with the check to see if the delegate exists! There is just one minor 'code-smell' to be fixed such that it's ready for merge!

rfaasse (approved):
Looks good, thanks for incorporating the changes!

PR #11833, opened 2023-11-22
[GeoMechanicsApplication] Support residual_criterion in c++ interface

carloslubbers:
This PR adds support for another convergence criterion, including the following changes:
* Support residual_criterion in c++ interface
* Added test for new criterion and extracted some shared functionality
* Slightly changed function in parameters_utilities.cpp such that it is usable in the factory (and other factories)
* Removed duplicate functionality from solving_strategy_factory.hpp


applications/GeoMechanicsApplication/custom_utilities/parameters_utilities.cpp
```diff
@@ -12,15 +12,25 @@
...
+    for (const std::string& entry : rNamesOfParametersToCopy)
+    {
+        if (rSourceParameters.Has(entry))
+        {
+            result.AddValue(entry, rSourceParameters[entry]);
+        }
+    }
+
     return result;
  }
```
WPK4FEM: Good that this is now shared i.s.o. copied.

applications/GeoMechanicsApplication/custom_utilities/convergence_criteria_factory.hpp
```diff
@@ -26,25 +28,32 @@ class ConvergenceCriteriaFactory
...
+            return std::make_shared<DisplacementCriteria<TSparseSpace, TDenseSpace>>(convergence_inputs);
+        }
+        if (rSolverSettings["convergence_criterion"].GetString() ==
+            "residual_criterion")
+        {
+            const std::vector<std::string> entries_to_copy = {
+                "residual_absolute_tolerance", "residual_relative_tolerance"};
+            Parameters convergence_inputs = ParametersUtilities::ExtractParameters(rSolverSettings, entries_to_copy);
+            return std::make_shared<ResidualCriteria<TSparseSpace, TDenseSpace>>(convergence_inputs);
         }
```
WPK4FEM: Correct functionality as far as I can read. Still the 2 branches are very similar, the difference being displacement or residual in the names. I wonder if it is worth to make that generic, such that water_pressure and temperature can be done in future without making another copy.

WPK4FEM (dismissed):
Meets the objective of making residual_criterion available and having a unit test for it.

applications/GeoMechanicsApplication/custom_utilities/convergence_criteria_factory.hpp
rfaasse: Good point, that should be possible, let's keep it in mind when we start adding the rest!

applications/GeoMechanicsApplication/custom_utilities/convergence_criteria_factory.hpp
```diff
@@ -26,25 +28,32 @@ class ConvergenceCriteriaFactory
...
-                if (rSolverSettings.Has(entry))
-                {
-                    convergenceInputs.AddValue(entry, rSolverSettings[entry]);
-                }
-            }
-            return std::make_shared<DisplacementCriteria<TSparseSpace, TDenseSpace>>(convergenceInputs);
+            const std::vector<std::string> entries_to_copy = {
+                "displacement_absolute_tolerance",
+                "displacement_relative_tolerance"};
+            Parameters convergence_inputs = ParametersUtilities::ExtractParameters(rSolverSettings, entries_to_copy);
```
avdg81: I would make this variable `const`:
```suggestion
            const auto convergence_inputs = ParametersUtilities::ExtractParameters(rSolverSettings, entries_to_copy);
```

applications/GeoMechanicsApplication/custom_utilities/convergence_criteria_factory.hpp
```diff
@@ -26,25 +28,32 @@ class ConvergenceCriteriaFactory
...
+            return std::make_shared<DisplacementCriteria<TSparseSpace, TDenseSpace>>(convergence_inputs);
+        }
+        if (rSolverSettings["convergence_criterion"].GetString() ==
+            "residual_criterion")
+        {
+            const std::vector<std::string> entries_to_copy = {
+                "residual_absolute_tolerance", "residual_relative_tolerance"};
+            Parameters convergence_inputs = ParametersUtilities::ExtractParameters(rSolverSettings, entries_to_copy);
```
avdg81: And the same thing here:
```suggestion
            const auto convergence_inputs = ParametersUtilities::ExtractParameters(rSolverSettings, entries_to_copy);
```

applications/GeoMechanicsApplication/custom_utilities/convergence_criteria_factory.hpp
avdg81: I agree with @WPK4FEM's comment. We could consider to follow the same strategy as we did for the process factory, where the factory itself is very generic and creators can be added as needed. But for me, it is fine to do that in a separate PR.

applications/GeoMechanicsApplication/custom_utilities/parameters_utilities.cpp
```diff
@@ -12,15 +12,25 @@
...
-    result.CopyValuesFromExistingParameters(rSourceParameters, rNamesOfParametersToCopy);
+
+    for (const std::string& entry : rNamesOfParametersToCopy)
+    {
+        if (rSourceParameters.Has(entry))
+        {
+            result.AddValue(entry, rSourceParameters[entry]);
+        }
+    }
+
```
avdg81, line 41: Now that we have changed the behavior of this function, I'm pondering whether we should rename it to indicate that it will only copy parameters that exist in `rSourceParameters`. The supplied parameter names don't have to be present there and they are silently skipped if they don't. The only suggestion that comes to my mind now is `CopyOptionalParameters`. (Its counterpart could then be named `CopyRequiredParameters`.) What do you think?

avdg81 (commented):
Looks like a clear and clean extension of existing functionality. I also like it a lot that you have eliminated some duplicated code. Well done! I only have a few minor suggestions, nothing blocking.

applications/GeoMechanicsApplication/custom_utilities/convergence_criteria_factory.hpp
rfaasse: Done, for some other constructors this was not possible, but here it seems okay

applications/GeoMechanicsApplication/custom_utilities/parameters_utilities.cpp
rfaasse: Good suggestion, I brought back the old function so we can use both

avdg81 (approved):
We are very happy with these changes, so please merge this PR.

PR #11836, opened 2023-11-22
[Core][MPI] Extend constraint utilities for MPI

loumalouomega:
**📝 Description**

This PR introduces a new function called `DistributedComputeActiveDofs` in the `kratos/utilities/constraint_utilities.cpp` file. This function is designed to compute active degrees of freedom (Dofs) in a distributed computing environment. It takes several parameters, including a ModelPart, a vector to store the active Dofs, the DofSet, and the InitialDofId. The function calculates which Dofs are active based on certain conditions and updates the `rActiveDofs` vector accordingly.

Additionally, there are some minor updates in the `kratos/utilities/constraint_utilities.cpp` file, such as including a new header file and adding comments for clarity. The `kratos/utilities/constraint_utilities.h` file also receives a new function declaration for `DistributedComputeActiveDofs`.

**🆕 Changelog**

- [Extend constraint utilities for MPI](https://github.com/KratosMultiphysics/Kratos/commit/4c2c6628f256a08bb7bd7c4539861fcfaa4ad4a5)

### DIFF
--- a/kratos/utilities/constraint_utilities.cpp
+++ b/kratos/utilities/constraint_utilities.cpp
@@ -18,6 +18,7 @@
 // Project includes
 #include "utilities/constraint_utilities.h"
 #include "utilities/parallel_utilities.h"
+#include "utilities/reduction_utilities.h"
 #include "utilities/atomic_utilities.h"
 
 namespace Kratos::ConstraintUtilities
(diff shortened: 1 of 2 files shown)


loumalouomega:
Ping @roigcarlo

kratos/utilities/constraint_utilities.cpp
```diff
@@ -66,6 +67,66 @@ void ComputeActiveDofs(
...
+    });
+
+    // Base active dofs
+    rActiveDofs.resize(number_local_dofs);
+
+    block_for_each(
+        rActiveDofs,
+        [](int& r_dof)
+        { r_dof = 1; }
+    );
```
roigcarlo, line 101: Probably its better to use std::fill()as the compiler will likely replace it for a std::memset(), which cannot do if its inside the parallel lambda.

But I am not sure, we should measure it. 

(Same comment)

loumalouomega: The current one is faster, and if I try to use std::fill(std::execution::par, rActiveDofs.begin(), rActiveDofs.end(), 1); I have linklig problems with TBB (strange, I am not using TBB)

roigcarlo: Its used by the residual based builder and solve (as an option) if you compile with intel or MKL (or TBB). Otherwise windows is excruciatingly slow.

loumalouomega: Yes, but I am not invoking him, I am using std::execution::par, should use STL one. Any case I read in google that OMP is faster than std::fill without the execution policy, can we keep it as it is?

loumalouomega: Then we can keep it as it is

loumalouomega:
Hello @roigcarlo

roigcarlo (approved):
OK but I don't know what this code does

PR #11829, opened 2023-11-20
[Core] Toward Trilinos/MPI unification of `DisplacementCriteria`/`ResidualCriteria`

loumalouomega:
**📝 Description**

This PR involves changes in `displacement_criteria.h` and `residual_criteria.h` toward the unification and simplification of the implementation between serial and MPI/Trilinos. The idea is to minimize diferences and code duplication between serial and MPI code, for a future PR. The changes include:

1. In  `displacement_criteria.h`:
   - Refactoring of the `DisplacementCriteria` class, including updates to type definitions and member functions.
   - The introduction of MPI rank handling in some parts of the code.
   - Improved DoF handling and calculations.
   - Adjustments to the class constructor and destructor.

2. In  `residual_criteria.h`:
   - Refactoring of the `ResidualCriteria` class, including updates to type definitions and member functions.
   - Deletion of the assignment operator.
   - Improvements in handling Degree of Freedom (DoF) status.
   - Improved DoF handling and calculations.
   - Adjustments to the class constructor and destructor.
   - Updates to the calculation of residual solution norms toward unification with MPI/Trilinos.

**🆕 Changelog**

- [Toward Trilinos/MPI unification of DisplacementCriteria`/`ResidualCriteria`](https://github.com/KratosMultiphysics/Kratos/commit/d6e267ba267531150dbc529a77eecf4368d715b7)

PR #11837, opened 2023-11-22
[Core][MPI][TrilinosApplication] Support MPC in distributed convergence criteria

loumalouomega:
**📝 Description**

Adding support for displacement and residual criteria for **MPI** in the `TrilinosApplication`. It takes into account changes from https://github.com/KratosMultiphysics/Kratos/pull/11829 in order to derive the trilinos versions with minimal changes, reducing code duplication and unifying behaviour. Displacement criteria in `TrilinosApplication` was not 100% consistent with serial implementation, a priori differences would be small, but this PR also unifies to 100% same behaviour.

Tests added accordingly.

Includes https://github.com/KratosMultiphysics/Kratos/pull/11829 , https://github.com/KratosMultiphysics/Kratos/pull/11835 and https://github.com/KratosMultiphysics/Kratos/pull/11836

Some TODOs added for more consistent changes, that may imply an API change and therefore is postpone for later PR.

**🆕 Changelog**

- [Update TrilinosResidualCriteria](https://github.com/KratosMultiphysics/Kratos/commit/0d35b7762ee0029f247c56b9a27a58c6f43d3762)
- [Adding test for TrilinosResidualCriteria](https://github.com/KratosMultiphysics/Kratos/commit/034940acbd2f575caa05b71711ad5f03691f0867)
- [Update refactor TrilinosDisplacementCriteria](https://github.com/KratosMultiphysics/Kratos/commit/12918a85252c2e0fb2439eba116d2925e1d18c34)
- [Adding test for TrilinosDisplacementCriteria](https://github.com/KratosMultiphysics/Kratos/commit/74468d99edd8017d326d295c1d8ad68fe412e107)
- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11829
- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11835
- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11836
- Changes from https://github.com/KratosMultiphysics/Kratos/pull/11846


kratos/utilities/constraint_utilities.cpp
```diff
@@ -66,6 +67,66 @@ void ComputeActiveDofs(
...
+    });
+
+    // Base active dofs
+    rActiveDofs.resize(number_local_dofs);
+
+    block_for_each(
+        rActiveDofs,
+        [](int& r_dof)
+        { r_dof = 1; }
+    );
```
roigcarlo: Probably its better to use `std::fill()`as the compiler will likely replace it for a `std::memset()`, which cannot do if its inside the parallel lambda.

But I am not sure, we should measure it.

applications/TrilinosApplication/custom_strategies/convergencecriterias/trilinos_displacement_criteria.h
```diff
@@ -4,332 +4,285 @@
...
+     * @param Rank The rank of the Dof.
+     * @return True if the Dof is free, false otherwise.
+     * @todo We should doo as in the residual criteria, and consider the active DoFs (not just free), taking into account the MPC in addition to fixed DoFs
+     */
+    bool IsFreeDof(
+        const DofType& rDof,
+        const int Rank
+        ) override
+    {
+        return (rDof.IsFree() && (rDof.GetSolutionStepValue(PARTITION_INDEX) == Rank));
```
RiccardoRossi: why do you need to check the PARTITION_INDEX here? the fixity status should be consistent on any Cpu

RiccardoRossi: btw, this function is very small, making it virtual makes an important overhead

applications/TrilinosApplication/custom_strategies/convergencecriterias/trilinos_residual_criteria.h
```diff
@@ -73,81 +124,198 @@ class TrilinosResidualCriteria : public ResidualCriteria< TSparseSpace, TDenseSp
...
+    ///@}
+    ///@name Protected LifeCycle
+    ///@{
+
+    ///@}
+private:
+    ///@name Member Variables
+    ///@{
+
+    IndexType mInitialDoFId = 0; /// The initial DoF Id
```
RiccardoRossi: i would start this with an "impossible" is, say -1

loumalouomega: -1 is not possible for unsigned int (IndexType -> std::size_t)

applications/TrilinosApplication/custom_strategies/convergencecriterias/trilinos_displacement_criteria.h
loumalouomega: I can avoid virtual with some static polimorfism

roigcarlo (commented):
I would rethink this `trilinos_displacement_criteria.h` as a whole tbh. I understand this is a continuation of #11829.

After looking at yesterday's Pr and this one. Why did not unify everything in the "serial" version of the  `displacement_criteria.h`  and just export it with the trilinos space in python?

- After the changes in the `IsFree`for the dofs which accepts the rank both in serial and parallel I see no other change in the code base.

- You can easily move the `InitializeSolutionStep` changes to the base class since most of the functions in the base also make use of the rank. This one should not be different.

- `GetDefaultparameters` becomes obsolete because there is only one class.

- `PostCriteria` and `CalculateReferenceNorm` are gone already.


**Edit**:

Also and more importantly, the `Epetra_communicator` is not used anywhere, you can remove the include which is the only think that would justify a separate class inside trilinos app.

applications/TrilinosApplication/custom_strategies/convergencecriterias/trilinos_displacement_criteria.h
loumalouomega: You must check that it is on the partition, otherwise you are summing several times the same value into the norm

loumalouomega: See the previous implementation

kratos/utilities/constraint_utilities.cpp
loumalouomega: Review first https://github.com/KratosMultiphysics/Kratos/pull/11836

loumalouomega:
> I would rethink this `trilinos_displacement_criteria.h` as a whole tbh. I understand this is a continuation of #11829.
> 
> After looking at yesterday's Pr and this one. Why did not unify everything in the "serial" version of the `displacement_criteria.h` and just export it with the trilinos space in python?
> 
>     * After the changes in the `IsFree`for the dofs which accepts the rank both in serial and parallel I see no other change in the code base.
> 
>     * You can easily move the `InitializeSolutionStep` changes to the base class since most of the functions in the base also make use of the rank. This one should not be different.
> 
>     * `GetDefaultparameters` becomes obsolete because there is only one class.
> 
>     * `PostCriteria` and `CalculateReferenceNorm` are gone already.
> 
> 
> **Edit**:
> 
> Also and more importantly, the `Epetra_communicator` is not used anywhere, you can remove the include which is the only think that would justify a separate class inside trilinos app.

The problem is that you need to check the PARTITION_INDEX in Trilinos, I can do some magic to avoid that...

roigcarlo:
You can pass a lambda that checks that once instantiated

loumalouomega:
> You can pass a lambda that checks that once instantiated

I have a better idea without penalty

loumalouomega:
> > You can pass a lambda that checks that once instantiated
> 
> I have a better idea without penalty

Look now

roigcarlo:
mm not very fan tbh, you are assuming that anything that is not the base type will require `PARTITION_INDEX` and you still need two separate classes.

Again not much familiar with the code, but can't the caller handle who is set or unset based on its parittion and leave the criterion to blindly trust what's set in the dof?

example of what I say:

```c++
# include "displacement_criteria.h"

void prepare_whatever_mpi() {
    auto dofs = getDofsArray()

    for (auto &d in dofs) {
        if dofs &= (dofs.GetValue(PARTITION_INDEX) == rank)
    } 

    DisplacementCriteria<TrilinosSparseSpaceType, TrilinosLocalSpaceType>(dofs)
}
```

?

loumalouomega:
@RiccardoRossi can you check this?

kratos/solving_strategies/convergencecriterias/displacement_criteria.h
```diff
@@ -436,7 +441,7 @@ class DisplacementCriteria
         struct TLS {TDataType dof_value{};};
 
         const TDataType reference_disp_norm = block_for_each<SumReduction<TDataType>>(rDofSet, TLS(), [this, &rank](auto& rDof, TLS& rTLS) {
```
RiccardoRossi, line 443: why don't you check for distributed or not distributed outside of this loop? this would allow avoiding the if constexpr in the IsFreeDof

also, you did not answer to my comment about the why the partition index check is needed

and indeed if we can do this without any dependency on trilinos, than i would only have the base class and instantiate it with different template arguments from within the trilinos_applicaiton as suggested by @roigcarlo

kratos/solving_strategies/convergencecriterias/residual_criteria.h
```diff
@@ -404,12 +404,16 @@ class ResidualCriteria
...
      * @return True if the Dof is free, false otherwise.
      */
-    virtual bool IsFreeDof(
+    bool IsFreeDof(
         const DofType& rDof,
         const int Rank
         )
     {
-        return rDof.IsFree();
+        if constexpr (std::is_same<TSparseSpace, UblasSpace<double, CompressedMatrix, Vector>>::value) {
```
RiccardoRossi: same comment as above. I would put the if on the outside of the loops where you use this

kratos/solving_strategies/convergencecriterias/displacement_criteria.h
loumalouomega: I have a better solution in mind. My point is to avoid the penalty of that check and minimize the code duplication

loumalouomega: Fixed with https://github.com/KratosMultiphysics/Kratos/pull/11846, now I defined a static method in the spaces that provide information if distributed or not. If a new space is considered a compilation error will appear and the method is implemented in less than 1 minute, I think is a good solution. This avoid any penalty in performance as it is a static check.

kratos/solving_strategies/convergencecriterias/residual_criteria.h
loumalouomega: See my previous comment.

applications/TrilinosApplication/custom_strategies/convergencecriterias/trilinos_displacement_criteria.h
RiccardoRossi: ok

RiccardoRossi: just it is a wrong naming for the funciton then. we should have a comment in the code to explain that it is done for that reason

applications/TrilinosApplication/custom_strategies/convergencecriterias/trilinos_residual_criteria.h
RiccardoRossi: well, the largest possible number then. In any case it is not really important so let's leave this out

applications/TrilinosApplication/custom_strategies/convergencecriterias/trilinos_displacement_criteria.h
loumalouomega: Do you mean in the space function or in the convergence criteria one?

applications/TrilinosApplication/custom_strategies/convergencecriterias/trilinos_residual_criteria.h
loumalouomega: Okay, I can use the lasrgest number then

RiccardoRossi:
> @RiccardoRossi can you check this?

So the current version is much improved, however i agree with @roigcarlo that at this point we can directly remove the files from the trilinos application and implement everything in the base class.

to be honest at that point i think it would be clearer to do something on the line of 

~~~cpp
if(serial) 
     loop for the serial case
else
    loop for the distributed case
~~~
(which surely could be made nicer defining a lambda in different ways and then using it in a parallel loop

doing so it would be very clear why yo need to use the PARTITION_INDEX (that is, not to sum twice the value of the same dof)

RiccardoRossi:
IsFreeDof.

you do not only check it is free, you also check that you account it only once

loumalouomega:
> > @RiccardoRossi can you check this?
> 
> So the current version is much improved, however i agree with @roigcarlo that at this point we can directly remove the files from the trilinos application and implement everything in the base class.

I would not do it for the following reason. As I say in the header of the PR, I found an inconsistency in the displacement criteria, but I will not fix it in this PR, because it is an actual behaviour change, so I will do it in an independent PR. After that, I will need to add things to the displacement criteria and the differences between the trilinos and serial version would be similar to the the diferences can be found between the residual criteria in serial/trilinos.

> to be honest at that point i think it would be clearer to do something on the line of
> 
> ```c++
> if(serial) 
>      loop for the serial case
> else
>     loop for the distributed case
> ```

Of course it will be clearer, but will imply code duplication with the corresponding problems. The only difference is to check when to take into account the dofs.

> (which surely could be made nicer defining a lambda in different ways and then using it in a parallel loop

Lambdas have a cost, the current check is done in compilation time.

> doing so it would be very clear why yo need to use the PARTITION_INDEX (that is, not to sum twice the value of the same dof)

Added

loumalouomega:
> IsFreeDof.
> 
> you do not only check it is free, you also check that you account it only once

I wil, rename it, IsFreeAndLocalDof

RiccardoRossi:
> > > @RiccardoRossi can you check this?
> > 
> > 
> > So the current version is much improved, however i agree with @roigcarlo that at this point we can directly remove the files from the trilinos application and implement everything in the base class.
> 
> I would not do it for the following reason. As I say in the header of the PR, I found an inconsistency in the displacement criteria, but I will not fix it in this PR, because it is an actual behaviour change, so I will do it in an independent PR. After that, I will need to add things to the displacement criteria and the differences between the trilinos and serial version would be similar to the the diferences can be found between the residual criteria in serial/trilinos.
> 
> > to be honest at that point i think it would be clearer to do something on the line of
> > ```c++
> > if(serial) 
> >      loop for the serial case
> > else
> >     loop for the distributed case
> > ```
> 
> Of course it will be clearer, but will imply code duplication with the corresponding problems. The only difference is to check when to take into account the dofs.
> 
> > (which surely could be made nicer defining a lambda in different ways and then using it in a parallel loop
> 
> Lambdas have a cost, the current check is done in compilation time.
> 
> > doing so it would be very clear why yo need to use the PARTITION_INDEX (that is, not to sum twice the value of the same dof)
> 
> Added

I would say that you are not right on the cost of the lambda, in the sense that when you use for_each YOU ARE using a lambda, so anyhow this is a price you pay

roigcarlo:
The `IsDistributed` solution looks very nice :+1: from my side too. About merging the classes, its difficult for me to judge whether its mergeable or not until we see the changes you want to make, so I would side to leave them split by now, and discuss again if the merging is possible then.

roigcarlo:
To me ti can go In, but I prefer to wait for explicit @RiccardoRossi approval for this one.

loumalouomega:
> To me ti can go In, but I prefer to wait for explicit @RiccardoRossi approval for this one.

Waiting for @RiccardoRossi then

RiccardoRossi (approved):
ok then

PR #11843, opened 2023-11-23
Geo/quickfix precision phreatic law

aronnoordam:
fixed issue in saturated below phreatic level law, where Saturation level is not calculated consistently


avdg81 (changes requested):
I have a few problems with this PR:
1. The adopted precision is defined in two different places. Preferably there should be one and only one place where this value is defined.
2. The fact that we are comparing computed zeros suggests that we may need to relate the applied tolerance value in some way to the convergence criterion. We suspect that if a stricter convergence criterion is applied, the tolerance value here should be stricter too.
3. There is no regression test. If this ever gets broken again, we won't know about it immediately.

WPK4FEM:
The test is for a U_Pw element, but contains no boundary conditions and no initial field for the Pw variable. That may make this an ill posed problem. The objective of the test seems to be to have the phreatic level at the lower boundary of the model. The attached projectparameters file aims at curing the lack of Pw boundary conditions. To make sure this works I have run the attached projectparameters with a Release build.

Further if the water conditions are not supposed to change during the computation, "ignore_undrained": true, may be added to the solver settings. That results in decoupling the U from the Pw variables.
[ProjectParameters.json](https://github.com/KratosMultiphysics/Kratos/files/13486254/ProjectParameters.json)

PR #11845, opened 2023-11-24
Do not attempt to change ROTATION if ROTATION is not present.

WPK4FEM:
**📝 Description**
In geomechanics_analysis.py tests on presence of SolutionStepVariables have been added before attempting to change values of non-existing things.

**🆕 Changelog**
- extra check in geomechanics_analysis.py


avdg81 (approved):
Nice improvement! Thank you for fixing this issue with rotations and for removing some code duplication.

PR #11854, opened 2023-11-26
[Core] Improve Master-Slave Constraint Insertion

matekelemen:
## Problem

Inserting `MasterSlaveConstraint` into `ModelPart` has horrendous performance and complexity due to repeatedly making sure that the `PointerVectorSet` containing the constraints remains sorted and unique. Here's what happens when you call `ModelPart::CreateNewMasterSlaveConstraint`:

1) the called `ModelPart` defers the constraint's constuction to its parent `ModelPart`
2) ... recursive deferring until the root `ModelPart` is reached
3) the root `ModelPart` throws an error if the constraint already exists in the requested `Mesh`
4) the root `ModelPart` constructs the constraint and inserts it into the requested `Mesh`
   -  `PointerVectorSet` sorts itself during the insertion process
5) the root `ModelPart` makes the **constraints of the `Mesh`** unique
   - the already sorted `PointerVectorSet` sorts itself again
   - then does an `std::unique` + `std::erase` combo
6) at this point, the root `ModelPart` is done with the constraint (1 binary search, 1 unique, 2 sorts) and sends the construced constraint back, down the `ModelPart` hierarchy
7) each sub `ModelPart` in the recursive hierarchy repeats the insertion (1 sort) and uniquification (1 sort, 1 unique) disregarding the (very likely) possibility that the constraint is already in the mesh.

This results in atrocious setup performance that takes up 70% of the analysis time (**including the solution of the linear system**) in one of my cases.

## Changes

- defer the responsibility of maintaining a sorted and unique constraint set to the `Mesh` (it's the one owning the set anyway)
- check for existing constraints **BEFORE** insertion, and ignore the request instead of performing the insertion and deleting the (newly added) constraint
- check whether the current `ModelPart` shares its mesh with its parent, and skip the insertion request if it does

With the changes above, the root `ModelPart` performs a single sort (no uniquification), and the rest of the sub `ModelPart`s only sort (and do it only once) if they don't share their meshes.


loumalouomega (approved):
Hyperapproved as in @KratosMultiphysics/altair we noticed this

loumalouomega:
@jcotela and @ddiezrod FYI

AlejandroCornejo:
> Hyperapproved as in @KratosMultiphysics/altair we noticed this

Since when? Maybe we could have solved this before...

loumalouomega:
> > Hyperapproved as in @KratosMultiphysics/altair we noticed this
> 
> Since when? Maybe we could have solved this before...

Some days that @jcotela started to fine tunning the performance of an operation involving MPC

jcotela:
> > > Hyperapproved as in @KratosMultiphysics/altair we noticed this
> > 
> > 
> > Since when? Maybe we could have solved this before...
> 
> Some days that @jcotela started to fine tunning the performance of an operation involving MPC

@loumalouomega the changes in this PR are welcome, but our performance problems come from the search that matches the constrained modelparts, constraint creation does not seem to be a bottleneck for us as in @matekelemen's case.

@AlejandroCornejo we are working on the issue at the moment, we only noticed it last week. #11839 is a small first step in the direction of fixing it, but we are still investigating how to best approach the problem.

kratos/sources/model_part.cpp
```diff
@@ -1229,45 +1229,53 @@ void ModelPart::AddMasterSlaveConstraints(std::vector<IndexType> const& MasterSl
...
+            ConstraintName,
+            Id,
+            rMasterDofsVector,
+            rSlaveDofsVector,
+            RelationMatrix,
+            ConstantVector,
+            ThisIndex);
+
+        // Add the constraint
+        if (&r_mesh != &mpParentModelPart->GetMesh(ThisIndex)) {
```
RiccardoRossi, line 1260: you need to insert also in the current modelpart no?

RiccardoRossi: oh, i saw now the check you do first.

still, the mesh used in submodelparts should be normally different to the one used in the parent...so this should be a cornercase

RiccardoRossi (commented):
ok, my previous comment was stupid, so please disregard it.

however in general meshes should not be sheared between modelparts, so i don't really see why this improves. (i am fine with checking for existence before creating though)

having said this, the good way to achieve performence woudl be to create constraints and add them in a batch. That should allow to preven all performence problems

matekelemen:
> having said this, the good way to achieve performence woudl be to create constraints and add them in a batch. That should allow to preven all performence problems

true, but since we have this function as well, we should make sure there are no unnecessary bottlenecks like this in the implementation

> however in general meshes should not be sheared between modelparts, so i don't really see why this improves. (i am fine with checking for existence before creating though)

checking the sub model parts' meshes is just a minor addition. What truly makes a difference is getting rid of manual calls to `Sort` and `Unique`. Here's a quick benchmark I wrote to test the performance on current master (blue) and this PR (orange) - note that the axes are logarithmic:

```cpp
// --- External Includes ---
#include "benchmark/benchmark.h"

// --- Benchmark Includes ---
#include "BenchmarkApplication/common.hpp"
#include "BenchmarkApplication/cmake_variables.hpp"

// --- Med Includes ---
#include "custom_io/med_model_part_io.h"

// --- Core Includes ---
#include "containers/model.h"
#include "includes/kratos_application.h"


void InsertMasterSlaveConstraints(benchmark::State& rState)
{
    // Initialize Kratos
    Kratos::KratosApplication kratos("core");
    kratos.Register();

    for ([[maybe_unused]] auto dummy : rState) {
        rState.PauseTiming();

        { // <== open a scope to control the model life time
            Kratos::Model model;
            auto& r_model_part = model.CreateModelPart("root");
            r_model_part.AddNodalSolutionStepVariable(Kratos::DISPLACEMENT_X);

            // Load a model
            Kratos::MedModelPartIO((Kratos::GetBenchmarkDataDir() / "cube_quadratic_tets.med").string()).ReadModelPart(r_model_part);
            for (Kratos::Node& r_node : r_model_part.Nodes()) {
                r_node.AddDof(Kratos::DISPLACEMENT_X);
            }

            // Insert 100 linear master slave constraints
            // (or less if the model has fewer nodes)
            const std::size_t number_of_constraints = std::min(static_cast<std::size_t>(rState.range(0)),
                                                               r_model_part.NumberOfNodes() / 2);
            std::size_t i_constraint = 0ul;

            rState.ResumeTiming();

            for (std::size_t i_master_node=0ul; i_master_node<2*number_of_constraints; i_master_node+=2) {
                Kratos::Node& r_master_node = *(r_model_part.Nodes().begin() + i_master_node);
                Kratos::Node& r_slave_node  = *(r_model_part.Nodes().begin() + i_master_node + 1);
                r_model_part.CreateNewMasterSlaveConstraint(
                    "LinearMasterSlaveConstraint",
                    ++i_constraint,
                    r_master_node,
                    Kratos::DISPLACEMENT_X,
                    r_slave_node,
                    Kratos::DISPLACEMENT_X,
                    1.23,
                    4.56
                );
            } // for i_master_node

            // Exclude destructors from the benchmark
            rState.PauseTiming();
        } // <== destroy the model

        rState.ResumeTiming();
    }
}


BENCHMARK(InsertMasterSlaveConstraints)->Arg(1 << 6 )->Iterations(50);
BENCHMARK(InsertMasterSlaveConstraints)->Arg(1 << 8 )->Iterations(25);
BENCHMARK(InsertMasterSlaveConstraints)->Arg(1 << 10)->Iterations(12);
BENCHMARK(InsertMasterSlaveConstraints)->Arg(1 << 12)->Iterations(6 );
BENCHMARK(InsertMasterSlaveConstraints)->Arg(1 << 14)->Iterations(3 );
```

Screenshot_20231204_165706

roigcarlo:
ping

loumalouomega:
> ping

I agree

matekelemen:
Can I merge this, or are we waiting for someone's review?

Issue #11853, opened 2023-11-25
[Core] Incorrect AMGCLSolver::mBlockSize


matekelemen:
Judging by how it's calculated, I'm assuming `AMGCLSolver::mBlockSize` should represent the number of DoFs per variable (or node, I'm not sure), but it seems to me that the current implementation is broken.

Can someone explain what exactly is `mBlockSize` supposed to be equal to on the Kratos side?

##

I'm running a linear structural mechanics case with quadratic elements and some MPCs. For this kind of model, the number of DoFs in a variable should be 3 (x, y, z displacement or rotation components). However, `AMGCLSolver` ends up with an `mBlockSize` of 1, which completely derails the solver to the point it doesn't even converge (hardcoding it to 3 results in convergence).

Here's how `mBlockSize` is calculated:

https://github.com/KratosMultiphysics/Kratos/blob/86d7a1f8fe1ffbc14029a0fb89a94bf7b2ccd2f0/kratos/linear_solvers/amgcl_solver.h#L525-L547

It appears to me that this piece of code is supposed to count how many DoFs share the same ID, and makes 2 assumptions:
1) every ID is referenced by the same number of DoFs, so it's enough to count once (the first ID)
2) DoFs sharing their IDs are consecutive in `rDofSet`

The first assumption seems reasonable to me, but the second does not. `rDofSet` is a `PointerVectorSet<Dof<double>>`, which means that it's sorted by the values of raw `Dof` pointers. I don't see how this would guarantee consecutive IDs.

Am I missing something or is this a bug?

## 

P.S.: the solver does actually converge with a block size of 1, but more than 500% slower (for my specific model).

AlejandroCornejo:
Wow, that's very interesting (as much as concerning...). I had the same idea about what is the Block size and seems that this piece of code is not workling as expected... I pinned @ddemidov and @RiccardoRossi to be sure about this

ddemidov:
I am not familiar with Kratos internals well enough to understand what is the problem here, sorry.

dcagritan:
Do we have documentation regarding solvers in Kratos? Documentation or a wiki page that explains the parameters of the solvers for instance?

RiccardoRossi:
no

Issue #8003, opened 2020-12-18
AMGCL should accept settings in AMGCL format


philbucher:
in my recent discussions with @ddemidov I realized that it would be good for the amgcl solver (meaning the [wrapper in Kratos](https://github.com/KratosMultiphysics/Kratos/blob/master/kratos/linear_solvers/amgcl_solver.h)) to be able to accept settings in AMGCL format. This way if new settings become available in AMGCL we could even use them before adding them to our interface. I think this can be very useful in certain cases.
Also @RiccardoRossi started doing this for the ns solver in #7999

Also we do this already for the Trilinos solves:
https://github.com/KratosMultiphysics/Kratos/blob/c4a6e0ffb155f5aefa4c564eb1bf4d191aacfb3d/applications/TrilinosApplication/external_includes/amesos_solver.h#L67

PR #11855, opened 2023-11-26
[LinearSolvers] Add AMGCL Solver with a Different Interface

matekelemen:
*Disclaimer: I'm assuming that we want to keep the AMGCL solver in core as it is, which is why I've added a new solver instead of modifying the existing one.*

## Changes

Add another wrapper for AMGCL that has the following differences compared to `AMGCLSolver` in core:

1) the original wrapper constructs the AMG hierarchy in `Solve`. This is fine in most cases, but quickly becomes expensive if `Solve` is called multiple times for the same system (for example: because the solver is nested inside another one). The new solver constructs the hierarchy in `ProvideAdditionalData` and stores it until it's destroyed.

2) the original `AMGCLSolver` limits the set of options available in AMGCL, and does a lot of parameter manipulation after construction that the user has no control over. This might be for the better in most cases, but prohibits the exact specification of the solver when the user knows precisely what they want. The new solver passes settings directly to AMGCL without restricting or manipulating them.

3) beside the built-in CPU backend and double precision GPU backend supported by the original solver, the new one adds support for a single precision GPU backend. Here's the reasoning behind adding it:

   - small to medium sized systems with reasonable condition numbers and moderate tolerances will also converge with single precision representations

   - some solution strategies don't actually require high accuracy (for example composite solvers - I'll open a PR for one when I solve some of its quirks)

   - in my experience, the single precision system doesn't necessarily result in better performance, but it can save some video memory. This can be quite important for larger systems, since video memory is far more expensive and is in much shorter supply than main CPU memory

4) support block sizes up to 6 (instead of the original 4). An example use case is 3D structural mechanics with rotational DoFs

### DIFF
--- a/applications/LinearSolversApplication/custom_solvers/amgcl_raw_solver.cpp
+++ b/applications/LinearSolversApplication/custom_solvers/amgcl_raw_solver.cpp
@@ -734,8 +734,8 @@ void AMGCLRawSolver<TSparseSpace,TDenseSpace,TReorderer>::PrintData(std::ostream
 
 
 
-template KRATOS_API(LINEAR_SOLVERS_APPLICATION)
-class AMGCLRawSolver<
+template class KRATOS_API(LINEAR_SOLVERS_APPLICATION)
+AMGCLRawSolver<
     TUblasSparseSpace<double>,
     TUblasDenseSpace<double>,
     Reorderer<
--- a/applications/LinearSolversApplication/custom_solvers/amgcl_raw_solver.cpp
+++ b/applications/LinearSolversApplication/custom_solvers/amgcl_raw_solver.cpp
@@ -734,7 +734,7 @@ void AMGCLRawSolver<TSparseSpace,TDenseSpace,TReorderer>::PrintData(std::ostream
 
 
 
-template
+template KRATOS_API(LINEAR_SOLVERS_APPLICATION)
 class AMGCLRawSolver<
     TUblasSparseSpace<double>,
     TUblasDenseSpace<double>,
--- a/applications/LinearSolversApplication/custom_solvers/amgcl_raw_solver.cpp
+++ b/applications/LinearSolversApplication/custom_solvers/amgcl_raw_solver.cpp
@@ -734,7 +734,7 @@ void AMGCLRawSolver<TSparseSpace,TDenseSpace,TReorderer>::PrintData(std::ostream
 
 
 
-template class KRATOS_API(LINEAR_SOLVERS_APPLICATION)
+template class KRATOS_API(LINEARSOLVERS_APPLICATION)
 AMGCLRawSolver<
     TUblasSparseSpace<double>,
     TUblasDenseSpace<double>,
--- a/applications/LinearSolversApplication/custom_solvers/amgcl_raw_solver.cpp
+++ b/applications/LinearSolversApplication/custom_solvers/amgcl_raw_solver.cpp
@@ -292,7 +292,7 @@ struct AMGCLRawSolver<TSparseSpace,TDenseSpace,TReorderer>::Impl
     // A variant for grouping members related to the wrapped AMGCL solver. It's meant
     // to bundle solvers using different backends as well as its associated matrix wrapper.
     // The wrapped types are composed of the permutations of the following attributes:
-    // - block size [1, 2, 3, 4]
+    // - block size [1, 2, 3, 4, 5, 6]
     // - backend type [bultin, vexcl]
     // - value type [double, float (vexcl only)]
     std::variant<
@@ -617,11 +617,11 @@ void AMGCLRawSolver<TSparseSpace,TDenseSpace,TReorderer>::ProvideAdditionalData(
                 break;                                                                                      \
             }                                                                                               \
             case 5: {                                                                                       \
-                KRATOS_CONSTRUCT_AMGCL_SOLVER_BUNDLE_WITH_BLOCK_SIZE(BACKEND_TEMPLATE, BACKEND_SCALAR, 4);  \
+                KRATOS_CONSTRUCT_AMGCL_SOLVER_BUNDLE_WITH_BLOCK_SIZE(BACKEND_TEMPLATE, BACKEND_SCALAR, 5);  \
                 break;                                                                                      \
             }                                                                                               \
             case 6: {                                                                                       \
-                KRATOS_CONSTRUCT_AMGCL_SOLVER_BUNDLE_WITH_BLOCK_SIZE(BACKEND_TEMPLATE, BACKEND_SCALAR, 4);  \
+                KRATOS_CONSTRUCT_AMGCL_SOLVER_BUNDLE_WITH_BLOCK_SIZE(BACKEND_TEMPLATE, BACKEND_SCALAR, 6);  \
                 break;                                                                                      \
             }                                                                                               \
             default: KRATOS_ERROR << "unsupported block size: " << mpImpl->mDoFCount << "\n";               \
--- a/applications/LinearSolversApplication/custom_solvers/amgcl_raw_solver.cpp
+++ b/applications/LinearSolversApplication/custom_solvers/amgcl_raw_solver.cpp
@@ -1,0 +1,748 @@
+//    |  /           |
+//    ' /   __| _` | __|  _ \   __|
+//    . \  |   (   | |   (   |\__ `
+//   _|\_\_|  \__,_|\__|\___/ ____/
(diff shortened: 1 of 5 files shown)


philbucher:
I did not review the code, but have some comments:

1. This could be solved by adding an additional parameter in the settings no?
2. See #8003. I believe this would solve your problems no?
3. I dont see why this cannot be added to the core?
4. Same, how about adding it to the core?

Overall I am not sure if the reasons you mention justify a duplication of the existing interface. 
BTW, the name `amgcl_raw` reminds me of uncooked meat xD => perhaps a better name would be better (obviously I dont have a suggestion :D)

matekelemen:
@philbucher yes, aptly naming this class wasn't on my priority list and I'm not proud of the outcome :smile: 

1) it's not that easy unfortunately. The AMGCL solver types are all unique for each backend, so making the solver instance persistent in the class requires some type system gymnastics. Allowing two points of construction would make debugging quite difficult in the future.

2) That'd be sweet, but I imagine that the @KratosMultiphysics/technical-committee would want to retain the original parameter manipulating behaviour (which would be extremely complicated for the complete AMGCL parameter set). I support the idea though.

3) This complicates the solver construction and solution process quite a bit. The system matrix must first be copied to a new single precision matrix (meaning that the matrix adapter must also have a different type) at solver construction time. Then at solution time, you have to do the same thing for the RHS and solution vector. After conversion, these containers can be loaded to the GPU, but you have to repeat the process in reverse after solution. My point is that it's a load of extra complications that generally don't get accepted in core :(

4) yes that's simple and I intend to, but
   - first, I'd wait and see what the verdict on #11853 is
   - remember that this adds 2 additional AMGCL template instantiations, which lead to a non-negligible increase in core build times

matekelemen:
P.S.: @philbucher have a look at the code when you have some time, and you'll probably understand why I'm hesitant about opening a PR for Core with it (even though I went out of my way to structure and comment it as nicely as I could).

RiccardoRossi:
@matekelemen personally i am ok with the idea of an alternative solver interface, which is closer to the original interface.
Per regards the name of the class i would discard the "raw" in favor or "Barebone" or "Native" or something of the sort, but admittedly i don't have a very good naming proposal.

per regards the capabilities, it would be definitely very good that one can compute the preconditioner once and then use it many times. Is this possible with this PR?

on the technical side maybe AMGCL's author @ddemidov may want to suggest something?

philbucher:
Just to say that I like the native name
E.g. amgcl_native

loumalouomega:
> Just to say that I like the native name E.g. amgcl_native

[joke] E.g amgcl_pre_hispanic [/joke]

matekelemen:
> amgcl_native

Yep, that sounds a bit better.

> per regards the capabilities, it would be definitely very good that one can compute the preconditioner once and then use it many times. Is this possible with this PR?

Yes, that's one of the main motivations.

> per regards the use with single precision it is not that i dislike it but rather that i am not too sure about how it would work for "difficult" problems.

The primary motivation behind supporting a single precision system is solving problems that specifically don't require accurate solutions. The specific task I'm working on right now is improving the linear solver performance for quadratic elements. In this case, the first level of coarsening can be replaced by reducing the quadratic basis functions to equivalent linear basis functions. An efficient solution strategy for this case is to get a rough estimate of the linear system's solution (hence the single precision support - we only need residuals on the order of magnitude of `1e-1`) and smooth it out on the quadratic system.

> if "amgcl_native_parameters" == null (the default) we would keep the current behaviour.,
if it is different from null than it would completely override current parameters

I think this would lead to an unintuitive interface and a complicated implementation that'd be hard to debug.

applications/LinearSolversApplication/custom_solvers/amgcl_raw_solver.cpp
```diff
@@ -0,0 +1,748 @@
...
+vex::Context& GetVexCLContext() {
+    static vex::Context ctx(vex::Filter::Env);
+    [[maybe_unused]] static bool run_once = [](){
+        return true;
+    }();
+    return ctx;
+}
+
+template <int TBlockSize>
+void RegisterVexCLStaticMatrixType() {
```
ddemidov, line 62: Should not this be also parameterized on value type (`double`/`float`)?

applications/LinearSolversApplication/custom_solvers/amgcl_raw_solver.cpp
```diff
@@ -0,0 +1,748 @@
...
+    if (requested_backend_type.empty()) {
+        // CPU backend by default
+        mpImpl->mBackendType = AMGCLBackendType::CPU;
+    } else if (requested_backend_type == "float") {
+        #ifdef AMGCL_GPGPU
+            RegisterVexCLStaticMatrixType<2>();
+            RegisterVexCLStaticMatrixType<3>();
+            RegisterVexCLStaticMatrixType<4>();
+            RegisterVexCLStaticMatrixType<5>();
+            RegisterVexCLStaticMatrixType<6>();
```
ddemidov, line 395: These declare vexcl_static_matrices for `double` value type, see comment above.

ddemidov (commented):
I can not comment on what would be the best strategy for Kratos (creating a new solver, or branching from existsing one), but I did look at the `amgcl_raw_solver.cpp` implementation. 

The only problem I immediately see is that static matrix declaration is done for double value type only, even when the target value type is float.

The rest looks ok to me, but I only had a quick look.

applications/LinearSolversApplication/custom_solvers/amgcl_raw_solver.cpp
matekelemen: yep I missed this one, thank you! How come this didn't cause an issue at runtime?

ddemidov: could it be you did not test a single-precision block solver?

matekelemen: I definitely tried, but maybe there's another bug we haven't spotted yet.

Looking at the VRAM usage during solution, requesting a single precision run does result in 20-25% less memory usage (visual estimate, I haven't properly measure it yet), which is mostly what I'd expect.

ddemidov: You could run the tests with `export VEXCL_SHOW_KERNELS=1`, then vexcl would dump the generated kernels (the ones it actually uses) to stdout.

matekelemen: you're right, I was running double precision kernels. That's scary because I still got correct results. Does VexCL do an internal copy if the value types are mismatched?

applications/LinearSolversApplication/custom_solvers/amgcl_raw_solver.cpp
```diff
@@ -0,0 +1,748 @@
...
+                KRATOS_CONSTRUCT_AMGCL_SOLVER_BUNDLE_WITH_BLOCK_SIZE(BACKEND_TEMPLATE, BACKEND_SCALAR, 6);  \
+                break;                                                                                      \
+            }                                                                                               \
+            default: KRATOS_ERROR << "unsupported block size: " << mpImpl->mDoFCount << "\n";               \
+        } // switch mpImpl->mDoFCount
+
+    // Construct the solver
+    switch (mpImpl->mBackendType) {
+        #ifdef AMGCL_GPGPU
+            case AMGCLBackendType::SinglePrecisionGPU: {
```
ddemidov, line 633: @matekelemen , I can not see where the selection between float and double could go wrong, but I would add some debug prints here in each of the cases. And then, if it behaves here as expected, I would add some temporary output inside [amgcl::make_solver](https://github.com/KratosMultiphysics/Kratos/blob/217a930723dc6099d18eb426db783935d8dea917/external_libraries/amgcl/make_solver.hpp#L102) (may be with `sizeof(value_type)` in the output).

RiccardoRossi:
@matekelemen as an aside, we could take the occasion of this PR to add interface to the native csr format in kratos. (interface should be identical to that of ublas)

RiccardoRossi:
@matekelemen, To give a followup on the oral discusssion we had a TUM, @pooyan-dadvand and myself were discussing later on the way back.

we reached the agreemeent that even though building in single precision would be effective, making the linear solver to support single precision is a good first step (read needed).
So it definitely make sense that the AMGCL written in single precision works as is (read passing an already constructed single precision matrix)

PR #11860, opened 2023-11-28
[GeoMechanicsApplication] Add .clang-format to GeoMechanicsApplication for automatic formatting

rfaasse:
**📝 Description**
This adds a .clang-format file to the GeoMechanicsApplication folder, allowing for automatic formatting. The .clang-format file in the kratos wiki was used as a starting point. Some values were updated/removed due to newer versions of clang not supporting it anymore and one change was made to reflect the formatting of the current geomechanics code base a bit better.


rfaasse:
@avdg81 @markelov208 We add this .clang-format file in the geomechanics application folder, but right now it would also be able to format e.g. core folders, since there is no other .clang-format file. 

Might there be a reason not to check it in in the 'general' repository but only in e.g. our own FEA-tools repo?

avdg81:
> @avdg81 @markelov208 We add this .clang-format file in the geomechanics application folder, but right now it would also be able to format e.g. core folders, since there is no other .clang-format file.
> 
> Might there be a reason not to check it in in the 'general' repository but only in e.g. our own FEA-tools repo?

I would prefer to have the `.clang-format` file in the `GeoMechanicsApplication` subdirectory, close to the source code to be formatted. If, however, that would interfere with how other Kratos developers work (in particular those who have nothing to do with our application), we need to put it somewhere else, in order to not disturb/annoy them.

rfaasse:
> > @avdg81 @markelov208 We add this .clang-format file in the geomechanics application folder, but right now it would also be able to format e.g. core folders, since there is no other .clang-format file.
> > Might there be a reason not to check it in in the 'general' repository but only in e.g. our own FEA-tools repo?
> 
> I would prefer to have the `.clang-format` file in the `GeoMechanicsApplication` subdirectory, close to the source code to be formatted. If, however, that would interfere with how other Kratos developers work (in particular those who have nothing to do with our application), we need to put it somewhere else, in order to not disturb/annoy them.

@roigcarlo Would there be any worry from your side for checking in a .clang-format file into the GeoMechanicsApplication folder? If so, we can find another place for it, but as @avdg81 mentioned it would be preferred to have it close to the source code, as proposed in this PR.

roigcarlo:
You can add the file, but we will be unable to automatically apply the formating. 
We are also ok on adding a check for it.

avdg81 (approved):
I think it is really nice that we can use clang-format to format our source files consistently. Thanks for adding it.

PR #11871, opened 2023-12-04
[GeoMechanicsApplication] Small refactoring of T_normal_flux_condition

WPK4FEM:
**📝 Description**
Changed naming of local variables to indicate their location ( integration point, DOF ). Factored out a function and a struct.

**🆕 Changelog**
- renamed local variables
- removed the NormalFluxVariables struct
- inlined the functionality of the calculateAndAddRHS function
- removed 2 unnecessary includes

### DIFF
--- a/applications/GeoMechanicsApplication/custom_conditions/T_normal_flux_condition.cpp
+++ b/applications/GeoMechanicsApplication/custom_conditions/T_normal_flux_condition.cpp
@@ -71,16 +71,16 @@ void GeoTNormalFluxCondition<TDim, TNumNodes>::CalculateRHS(Vector& rRightHandSi
         auto N = row(r_N_container, integration_point);
 
         // Interpolation of nodal normal flux to integration point normal flux.
-        auto NormalFluxOnIntegrationPoint = MathUtils<>::Dot(N, normal_flux_vector);
+        auto normal_flux_on_integration_point = MathUtils<>::Dot(N, normal_flux_vector);
 
         // Compute weighting coefficient for integration
-        auto IntegrationCoefficient = CalculateIntegrationCoefficient(j_container[integration_point],
-                                                                             r_integration_points[integration_point].Weight());
+        auto integration_coefficient = CalculateIntegrationCoefficient(j_container[integration_point],
+                                                                              r_integration_points[integration_point].Weight());
 
         // Contributions to the right hand side
-        auto NormalFluxOnDOF = NormalFluxOnIntegrationPoint * N * IntegrationCoefficient;
+        auto normal_flux_on_DOF = normal_flux_on_integration_point * N * integration_coefficient;
         GeoElementUtilities::AssemblePBlockVector<0, TNumNodes>(rRightHandSideVector,
-                                                                NormalFluxOnDOF);
+                                                                normal_flux_on_DOF);
     }
 }
 
--- a/applications/GeoMechanicsApplication/custom_conditions/T_normal_flux_condition.cpp
+++ b/applications/GeoMechanicsApplication/custom_conditions/T_normal_flux_condition.cpp
@@ -73,12 +73,11 @@ void GeoTNormalFluxCondition<TDim, TNumNodes>::CalculateRHS(Vector& rRightHandSi
         // Interpolation of nodal normal flux to integration point normal flux.
         auto normal_flux_on_integration_point = MathUtils<>::Dot(N, normal_flux_vector);
 
-        // Compute weighting coefficient for integration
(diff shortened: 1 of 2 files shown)


applications/GeoMechanicsApplication/custom_conditions/T_normal_flux_condition.cpp
```diff
@@ -43,17 +43,16 @@ template <unsigned int TDim, unsigned int TNumNodes>
 GeoTNormalFluxCondition<TDim, TNumNodes>::~GeoTNormalFluxCondition() = default;
 
 template <unsigned int TDim, unsigned int TNumNodes>
-void GeoTNormalFluxCondition<TDim, TNumNodes>::CalculateRHS(VectorType& rRightHandSideVector,
-                                                         const ProcessInfo& rCurrentProcessInfo)
+void GeoTNormalFluxCondition<TDim, TNumNodes>::CalculateRHS(Vector& rRightHandSideVector,
```
markelov208, line 46: VectorType is used through the entire Kratos and it is equal to Vector anywhere, @avdg81, please could you tell it is possible to simplify it?

applications/GeoMechanicsApplication/custom_conditions/T_normal_flux_condition.cpp
```diff
@@ -66,40 +65,28 @@ void GeoTNormalFluxCondition<TDim, TNumNodes>::CalculateRHS(VectorType& rRightHa
     for (unsigned int i = 0; i < TNumNodes; ++i) {
         normal_flux_vector[i] = r_geom[i].FastGetSolutionStepValue(NORMAL_HEAT_FLUX);
     }
-    NormalFluxVariables variables;
```
markelov208, line 69: nice refactoring

applications/GeoMechanicsApplication/custom_conditions/T_normal_flux_condition.cpp
```diff
@@ -66,40 +65,28 @@ void GeoTNormalFluxCondition<TDim, TNumNodes>::CalculateRHS(VectorType& rRightHa
...
 
         // Interpolation of nodal normal flux to integration point normal flux.
-        variables.NormalFluxOnNode = MathUtils<>::Dot(variables.N, normal_flux_vector);
+        auto normal_flux_on_integration_point = MathUtils<>::Dot(N, normal_flux_vector);
 
         // Compute weighting coefficient for integration
-        this->CalculateIntegrationCoefficient(
-            variables.IntegrationCoefficient, j_container[integration_point],
-            r_integration_points[integration_point].Weight());
+        auto integration_coefficient = CalculateIntegrationCoefficient(j_container[integration_point],
```
markelov208: Would it be better to rename as weighting_integration_coefficient and to remove the comment?

applications/GeoMechanicsApplication/custom_conditions/T_normal_flux_condition.h
```diff
@@ -27,11 +25,9 @@ template <unsigned int TDim, unsigned int TNumNodes>
 class KRATOS_API(GEO_MECHANICS_APPLICATION) GeoTNormalFluxCondition
     : public GeoTCondition<TDim, TNumNodes> {
 public:
-    using NodeType = Node;
```
markelov208, line 30: As for VectorType, is it possible to simplify NodeType? @avdg81 , please what is your opinion?

markelov208 (commented):
Nice refactoring, now it is shorter and more understandable. Just a few comments, two of them are general.

applications/GeoMechanicsApplication/custom_conditions/T_normal_flux_condition.cpp
avdg81: I agree that I don't see the added value of defining `VectorType` inside this class. In fact, the alias name (`VectorType`) is longer than the  name of the type being aliased (`Vector`). For me, this is adding needless complexity to the class, so removing the alias is fine with me.

applications/GeoMechanicsApplication/custom_conditions/T_normal_flux_condition.h
avdg81: Also here, I don't see the added value of the `NodeType` alias. I prefer to have a single type (`Node`) rather than it being aliased to `NodeType` in many classes. So I'm in favor of removing this alias.

avdg81 (approved):
Very nice and clear cleanup. By removing the things that we don't need, we can focus on the things that actually matter. Thanks for the work done!

markelov208 (approved):
Well done

PR #11875, opened 2023-12-05
[Core][MPI] Adding `GetSubDataCommunicator` to `DataCommunicator` and `MPIDataCommunicator`

loumalouomega:
**📝 Description**

This PR makes changes to the `DataCommunicator` and `MPIDataCommunicator`  classes, and some minor cleaning 

- Modifications to the `data_communicator.h`, `mpi_data_communicator.h`, and `mpi_data_communicator.cpp` files. 
	1. In the `DataCommunicator` class:
	 	- Added comments and improved code documentation for several member functions.
		- Refactored the `Rank()`, `Size()`, `IsDistributed()`, `IsDefinedOnThisRank()`, and `IsNullOnThisRank()` functions for better clarity and documentation.
		- Added a new function `GetSubDataCommunicator()` for obtaining a sub-DataCommunicator based on provided ranks and a new communicator name.
	2. In the `MPIDataCommunicator` class (a subclass of `DataCommunicator`):
	 	- Updated and improved the documentation for member functions, including `Rank()`, `Size()`, `IsDistributed()`, `IsDefinedOnThisRank()`, and `IsNullOnThisRank()`.
	 	- Added a new function `GetSubDataCommunicator()` for obtaining a sub-DataCommunicator, either by retrieving an existing one or creating a new one based on provided ranks and a new communicator name.
- Refactoring of the `kratos/mpi/utilities/data_communicator_factory.cpp` file. It includes the following changes:
	1. Removed an unused system includes comment.
	2. Added comments for system, external, and project includes.
	4. Reordered the includes to group system, external, and project includes separately.
- Changes made to the `parallel_environment.h` header file are as follows:
	1. Removed the `#ifndef KRATOS_PARALLEL_ENVIRONMENT_H_INCLUDED` and `#define KRATOS_PARALLEL_ENVIRONMENT_H_INCLUDED` preprocessor directives, and replaced them with `#pragma once`. This change simplifies the header file's include guard.

**🆕 Changelog**

- [Adding `GetSubDataCommunicator` to `DataCommunicator`](https://github.com/KratosMultiphysics/Kratos/commit/4537c85fe2bc5d4473bfe516089a2f292e1dac91)
- [Minor](https://github.com/KratosMultiphysics/Kratos/commit/de8c50b94e58039cfa61a1a951ccc6d16fbabee8)
- [Another minor](https://github.com/KratosMultiphysics/Kratos/commit/c2e7070cfd10f136ba2edf22b3b288db9e41c569)


philbucher:
can you please explain the usage / give an example?

I am pretty curious, since to my knowledge so far I was the only one using sub-communicators

loumalouomega:
> can you please explain the usage / give an example?
> 
> I am pretty curious, since to my knowledge so far I was the only one using sub-communicators

My problem was that currently that `data_communicator_factory.cpp` is only in `MPICore`, therefore I cannot call it from code without linking the `MPICore`. The solution therefore is to add the call in the data communicator, in the same way we do **MPI** calls in the `MPIDataCommunicator` and this avoids hardcode the **MPI** dependency.

The case use is going to be for **MPI** search, current implementation (not merged) defines serach in all partitions, being quite inefficient, so with this I can focus only in the aprtitions actually needed.

kratos/mpi/sources/mpi_data_communicator.cpp
```diff
@@ -474,6 +474,18 @@ bool MPIDataCommunicator::IsNullOnThisRank() const
     return mComm == MPI_COMM_NULL;
 }
 
+const DataCommunicator& MPIDataCommunicator::GetSubDataCommunicator(
+    const std::vector<int>& rRanks,
+    const std::string& rNewCommunicatorName
+    ) const
+{
+    if (ParallelEnvironment::HasDataCommunicator(rNewCommunicatorName)) {
+        return ParallelEnvironment::GetDataCommunicator(rNewCommunicatorName);
```
philbucher: this is problematic because the ranks passed here could be different!

looking at this again I am leaning to make the creation explicit :thinking: 

other opinions? This is important in terms of the concept

philbucher (commented):
thanks for the explanation

Not sure if the current approach is the best in terms of separating the concepts, but I see your point

@jcotela perhaps the factory could be in the regular core to keep things cleaner?

FYI I did [something similar](https://github.com/KratosMultiphysics/Kratos/blob/master/applications/CoSimulationApplication/python_scripts/utilities/data_communicator_utilities.py), but ofc in python this avoids the problem :)

philbucher:
> The case use is going to be for MPI search, current implementation (not merged) defines serach in all partitions, being quite inefficient, so with this I can focus only in the aprtitions actually needed.

BTW in my experience working with subcomms is not that convenient (basically requires to constantly check if the DataComm is defined on the rank). For your usecase excluding the partitions differently (e.g. a size-check) might be easier. But ofc I dont know the details of your implementation

loumalouomega:
> > The case use is going to be for MPI search, current implementation (not merged) defines serach in all partitions, being quite inefficient, so with this I can focus only in the aprtitions actually needed.
> 
> BTW in my experience working with subcomms is not that convenient (basically requires to constantly check if the DataComm is defined on the rank). For your usecase excluding the partitions differently (e.g. a size-check) might be easier. But ofc I dont know the details of your implementation

A priori in the way I am going to implement will not be necessary as the instance will be defined only the partitions were the communicator will work

kratos/mpi/sources/mpi_data_communicator.cpp
loumalouomega: I understand your point...

jcotela:
> thanks for the explanation
> 
> Not sure if the current approach is the best in terms of separating the concepts, but I see your point
> 
> @jcotela perhaps the factory could be in the regular core to keep things cleaner?
> 
> FYI I did [something similar](https://github.com/KratosMultiphysics/Kratos/blob/master/applications/CoSimulationApplication/python_scripts/utilities/data_communicator_utilities.py), but ofc in python this avoids the problem :)

It might be cleaner to access the factory from the ParallelEnvironment rather than through the DataCommunicator, but we would still need a base do-nothing serial implementation to put in the core.

Regarding the point of data communicators with the same name acting on different ranks, at least we should add some explicit checking when they are created, I agree that as is may be very confusing to users when a name collision occurs.

kratos/mpi/sources/mpi_data_communicator.cpp
loumalouomega: Would fix the problem this?

~~~C++
const DataCommunicator& MPIDataCommunicator::GetSubDataCommunicator(
    const std::vector<int>& rRanks,
    const std::string& rNewCommunicatorName
    ) const
{
    int has_data_communicator = ParallelEnvironment::HasDataCommunicator(rNewCommunicatorName);
    has_data_communicator = SumAll(has_data_communicator);
    KRATOS_ERROR_IF_NOT(has_data_communicator == 0 || has_data_communicator == Size()) << "The communicator " << rNewCommunicatorName << " already exists in some partitions but not in some others." << std::endl;
    if (has_data_communicator) {
        return ParallelEnvironment::GetDataCommunicator(rNewCommunicatorName);
    } else {
        return DataCommunicatorFactory::CreateFromRanksAndRegister(*this, rRanks, rNewCommunicatorName);
    }
}
~~~

loumalouomega: or:

~~~cpp
const DataCommunicator& MPIDataCommunicator::GetSubDataCommunicator(
    const std::vector<int>& rRanks,
    const std::string& rNewCommunicatorName
    ) const
{
    if (ParallelEnvironment::HasDataCommunicator(rNewCommunicatorName)) {
        const DataCommunicator& r_data_communicator = ParallelEnvironment::GetDataCommunicator(rNewCommunicatorName);
        const int rank = Rank();
        const auto it_find = std::find(rRanks.begin(), rRanks.end(), rank);
        if (it_find != rRanks.end()) {
            KRATOS_ERROR_IF_NOT(r_data_communicator.IsDefinedOnThisRank()) << "The rank " << rank << " is not defined in the existing data communicator " << rNewCommunicatorName  << " despite being in the provided rank list" << std::endl;
        } else {
            KRATOS_ERROR_IF_NOT(r_data_communicator.IsNullOnThisRank()) << "The rank " << rank << " is defined in the existing data communicator " << rNewCommunicatorName << " despite not being in the provided rank list" << std::endl;
        }
        return r_data_communicator;
    } else {
        return DataCommunicatorFactory::CreateFromRanksAndRegister(*this, rRanks, rNewCommunicatorName);
    }
}
~~~

loumalouomega:
@philbucher , @jcotela is waiting for your approval

philbucher:
I was waiting for the approval of @jcotela  😄

I leave the approval to him, he designed this stuff initially

jcotela:
Thinking some more about the issue with retrieving a wrong existing subcommunicator with the same name, I think it might be a good idea to add a check to the effect of `number_of_active_ranks == found_data_communicator.Size()` in addition to the existing check on active ranks. This would help preventing hard-to-debug issues where a data communicator has the same name and matches the active ranks known by the caller, but is also active on other, unknown ranks (which I realize is a contrived case, but would be otherwise undetectable and could happen if GetSubDataCommunicator is called from an environment that already works with a subcommunicator).

Other than that, I think that the PR is good to go

jcotela:
> I was waiting for the approval of @jcotela 😄
> 
> I leave the approval to him, he designed this stuff initially

but you are the one who actually used subcommunicators...

philbucher:
> > I was waiting for the approval of @jcotela 😄
> > I leave the approval to him, he designed this stuff initially
> 
> but you are the one who actually used subcommunicators...

good point, will have another look then by tmr

loumalouomega:
@jcotela this should be ready

kratos/mpi/sources/mpi_data_communicator.cpp
```diff
@@ -474,6 +474,32 @@ bool MPIDataCommunicator::IsNullOnThisRank() const
...
+{
+    if (ParallelEnvironment::HasDataCommunicator(rNewCommunicatorName)) {
+        const DataCommunicator& r_data_communicator = ParallelEnvironment::GetDataCommunicator(rNewCommunicatorName);
+        const int rank = Rank();
+        const auto it_find = std::find(rRanks.begin(), rRanks.end(), rank);
+        std::size_t number_of_active_ranks = 0;
+        if (it_find != rRanks.end()) {
+            KRATOS_ERROR_IF_NOT(r_data_communicator.IsDefinedOnThisRank()) << "The rank " << rank << " does not participate in the existing data communicator " << rNewCommunicatorName  << " despite being in the provided rank list" << std::endl;
+        } else {
+            number_of_active_ranks = 1;
```
jcotela: Why are you setting it to 1 here? I would think that the rank should not participate in the new communicator if we reach the `else` branch

loumalouomega: I put the one oin the other side, silly me :P

loumalouomega: Now

kratos/mpi/sources/mpi_data_communicator.cpp
```diff
@@ -474,6 +474,31 @@ bool MPIDataCommunicator::IsNullOnThisRank() const
...
+            KRATOS_ERROR_IF_NOT(r_data_communicator.IsDefinedOnThisRank()) << "The rank " << rank << " does not participate in the existing data communicator " << rNewCommunicatorName  << " despite being in the provided rank list" << std::endl;
+            const std::size_t world_size = static_cast<std::size_t >(r_data_communicator.Size());
+            KRATOS_ERROR_IF_NOT(rRanks.size() == world_size) << "Inconsistency between the communicator world size: " << world_size << " and the number of ranks required: " << rRanks.size() << std::endl;
+            std::size_t number_of_active_ranks = 1;
+            number_of_active_ranks = r_data_communicator.SumAll(number_of_active_ranks);
+            KRATOS_ERROR_IF_NOT(number_of_active_ranks == rRanks.size()) << "Inconsistency between the number of active ranks: " << number_of_active_ranks << " and the number of ranks required: " << rRanks.size() << std::endl;
+        } else {
+            KRATOS_ERROR_IF_NOT(r_data_communicator.IsNullOnThisRank()) << "The rank " << rank << " participates in the existing data communicator " << rNewCommunicatorName << " despite not being in the provided rank list" << std::endl;
```
philbucher: I would move this check out of the condition, this should be checked in both cases

kratos/mpi/sources/mpi_data_communicator.cpp
```diff
@@ -474,6 +474,31 @@ bool MPIDataCommunicator::IsNullOnThisRank() const
...
+const DataCommunicator& MPIDataCommunicator::GetSubDataCommunicator(
+    const std::vector<int>& rRanks,
+    const std::string& rNewCommunicatorName
+    ) const
+{
+    if (ParallelEnvironment::HasDataCommunicator(rNewCommunicatorName)) {
+        const DataCommunicator& r_data_communicator = ParallelEnvironment::GetDataCommunicator(rNewCommunicatorName);
+        const int rank = Rank();
+        const auto it_find = std::find(rRanks.begin(), rRanks.end(), rank);
+        if (it_find != rRanks.end()) {
```
philbucher: this should be an error if it is not in the list

kratos/mpi/sources/mpi_data_communicator.cpp
```diff
@@ -474,6 +474,31 @@ bool MPIDataCommunicator::IsNullOnThisRank() const
     return mComm == MPI_COMM_NULL;
 }
 
+const DataCommunicator& MPIDataCommunicator::GetSubDataCommunicator(
+    const std::vector<int>& rRanks,
```
philbucher, line 478: I am thinking that conceptually this should be a `set`. I see that the vector is easier to use, maybe this could be made unique in the DataCommFactory @jcotela ?

kratos/mpi/tests/cpp_tests/utilities/test_pointer_communicator.cpp
```diff
@@ -85,6 +86,71 @@ KRATOS_TEST_CASE_IN_SUITE(PointerCommunicator, KratosMPICoreFastSuite)
...
+    for(int i = current_rank + 1; i < world_size; ++i) {
+        indices.push_back(i);
+    }
+    ranks.reserve(world_size - 1);
+    for(int i = 0; i < world_size - 1; ++i) {
+        ranks.push_back(i);
+    }
+
+    if (current_rank < world_size - 1) {
+        auto& r_partial_data_comm = r_default_comm.GetSubDataCommunicator(ranks, "SubDataComm");
```
philbucher, line 117: can you please a tests also for the error conditions
and one for calling this 2 times, i.e. creating a sub-sub datacomm

philbucher (commented):
some more comments

kratos/mpi/sources/mpi_data_communicator.cpp
loumalouomega: I don't unerstand what do you mean

kratos/mpi/sources/mpi_data_communicator.cpp
loumalouomega: The error is later if the data communicator is not null (it ay be null)

kratos/mpi/sources/mpi_data_communicator.cpp
loumalouomega: In the factory a std::vector is used, the same is used here

kratos/mpi/tests/cpp_tests/utilities/test_pointer_communicator.cpp
loumalouomega: Okay

kratos/mpi/sources/mpi_data_communicator.cpp
philbucher: therefore the question :)

kratos/mpi/sources/mpi_data_communicator.cpp
philbucher: this should be checked at the very top of this function

loumalouomega: No, because it will fail, I already tried

philbucher: then it is being used wrong IMO
This is what I meant [last week](https://github.com/KratosMultiphysics/Kratos/pull/11875#issuecomment-1840464992)

Any MPI-call on a rank where the data-comm does not have a communicator defined fails

philbucher: i would say that any **SUB**DataComm can only contain the ranks of the parent, otherwise it is not "Sub"

loumalouomega: I will try again, maybe I did something stupid

kratos/mpi/tests/cpp_tests/utilities/test_pointer_communicator.cpp
loumalouomega: AT least this is done

kratos/mpi/sources/mpi_data_communicator.cpp
loumalouomega: Now I will try this

kratos/mpi/sources/mpi_data_communicator.cpp
jcotela: If I remember correctly, the factory uses a vector just because it passes its data directly as the C array that the corresponding MPI primitive uses. I went with a vector because it seemed straightforward, but I don't have a strong opinion either way. On the other hand, using a set has the nice feature that we are enforcing uniqueness on the interface level, so it is harder to call it with a wrong input.

kratos/mpi/sources/mpi_data_communicator.cpp
loumalouomega: Okay this is the thing, should this method be invoked from all the ranks always?, because is not actuallya fail, in the test I added it gets stuck because the subdatacommunicator is not invoked in all the ranks

loumalouomega:
@philbucher , @jcotela  is waiting for you

philbucher (approved):
Sorry for the delay

LGTM now

kratos/mpi/sources/mpi_data_communicator.cpp
philbucher: Not 100% sure either

I guess conceptually it makes sense to invoke it from all ranks, but I dont think it is possible 

I suggest you start working with it and potentially modify it in the future. This is a very niche feature, I would not be worried about API or behavior changes in the future

PR #11876, opened 2023-12-05
[Core] ComputeNormalsOnSimplex - fix when there are other geometries

ddiezrod:
Im getting an error in FullDebug when calling ComputeNormalsOnSimplex for a model part that contains lines besides the triangle conditions.

The problem is that line conditions they were being ignored for the calculation 
https://github.com/KratosMultiphysics/Kratos/blob/45ff863aad6687ba95362b558a698eeb317d039d/kratos/utilities/normal_calculation_utils.cpp#L730

but later, we were tring to read the NORMAL value from them to assign them to the nodes
https://github.com/KratosMultiphysics/Kratos/blob/45ff863aad6687ba95362b558a698eeb317d039d/kratos/utilities/normal_calculation_utils.cpp#L747

In this MR I make sure that only simplex conditions are used in the last part of code, where we assign the normal values to the nodes. A test is added to make sure this works correctly.

This MR also includes some KRATOS_TRY/CATCH that I added during the debugging.

### DIFF
--- a/kratos/tests/test_normal_utils.py
+++ b/kratos/tests/test_normal_utils.py
@@ -130,6 +130,18 @@ class TestNormalUtilsCoarseSphere(KratosUnittest.TestCase):
 
             self.assertLess(CalculateNorm(normal - solution_normal), 0.15)
 
+    def test_ComputeSimplexNormalModelPartWithLineCondition(self):
+        #Adding one line, to make sure it is getting ignored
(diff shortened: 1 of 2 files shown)


RiccardoRossi:
one way would be to check for the geometry family

rubenzorrilla (approved):
Makes sense. Thanks for spotting.

PR #11877, opened 2023-12-05
[Core] Adding `SynchronizePointsWithBoundingBox` method

loumalouomega:
**📝 Description**

This method will be used in the future `SearchWrapper` class in order to estimate the points to be computed in the distributed search taking into account a bounding box filtering (only the points insise the BB).

**🆕 Changelog**

- [Adding SynchronizePointsWithBoundingBox](https://github.com/KratosMultiphysics/Kratos/commit/c35879c72df4872e31ad8008276b31948c994c01)
- [Update](https://github.com/KratosMultiphysics/Kratos/commit/e443fcfbbe0e0df92bc192b4d0eb2ed2339639cc)
- [Refactor search utilities in order to filter by partition](https://github.com/KratosMultiphysics/Kratos/commit/1ecd720ff4c908c52bf61dea3be914bd4bcba537)


loumalouomega:
@roigcarlo can you take a look?

kratos/utilities/search_utilities.h
```diff
@@ -543,6 +637,251 @@ class SearchUtilities
...
+        std::vector<int> send_points_per_partition(1, NumberOfPoints);
+        std::vector<int> all_points_ranks(TotalNumberOfPoints);
+        rDataCommunicator.AllGather(send_points_per_partition, points_per_partition);
+        std::size_t initial_id = 0;
+        if constexpr (!std::is_same<TPointIteratorType, ModelPart::NodeIterator>::value || std::is_same<TPointIteratorType, ModelPart::NodeConstantIterator>::value) {
+            initial_id = std::accumulate(points_per_partition.begin(), points_per_partition.begin() + rank, 0);
+        }
+
+        // Initialize and resize vectors
+        if (rAllPointsCoordinates.size() != static_cast<unsigned int>(TotalNumberOfPoints * 3)) {
```
roigcarlo:
```suggestion
        if (rAllPointsCoordinates.size() != static_cast<std::size_t>(TotalNumberOfPoints * 3)) {
```

kratos/utilities/search_utilities.h
```diff
@@ -543,6 +637,251 @@ class SearchUtilities
...
+            rAllPointsCoordinates.resize(TotalNumberOfPoints * 3);
+        }
+        if (rAllPointsIds.size() != static_cast<unsigned int>(TotalNumberOfPoints)) {
```
roigcarlo:
```suggestion
        if (rAllPointsIds.size() != static_cast<std::size_t>(TotalNumberOfPoints)) {
```

kratos/utilities/search_utilities.h
```diff
@@ -543,6 +637,251 @@ class SearchUtilities
...
+            rAllPointsIds.resize(TotalNumberOfPoints);
+        }
+        std::vector<int> recv_sizes(world_size, 0);
+
+        // Auxiliary variables
+        std::size_t counter = 0;
+        array_1d<double, 3> coordinates;
```
roigcarlo: Just a suggestion, I found the declaration of `coordinates` here confusing, because then you are forced to use `noalias(coordinates)` to avoid the copy. In this particular case, and since the only usage is to assign it, you could just

```C++
send_points_coordinates[3 * counter + i_coord] = it_point->Coordinates()[i_coord]
```

or 


```C++
auto & coords = it_point->Coordinates();
send_points_coordinates[3 * counter + i_coord] = coords[i_coord]
```

But both of them would get optimized anyway....

kratos/utilities/search_utilities.h
```diff
@@ -543,6 +637,251 @@ class SearchUtilities
...
+        // Some definitions
+        IndexType i_coord = 0;
+        array_1d<double, 3> coordinates;
+
+        // If distributed
+        if (rDataCommunicator.IsDistributed()) {
+            // Prepare MPI data
+            const int rank = rDataCommunicator.Rank();
+            const int world_size = rDataCommunicator.Size();
+            std::vector<int> ranks(1, rank);
```
roigcarlo, line 821: This does not need to be a vector :S

kratos/utilities/search_utilities.h
```diff
@@ -543,6 +637,251 @@ class SearchUtilities
...
+            for (IndexType i_point = 0 ; i_point < total_number_of_points; ++i_point) {
+                auto it_point = itPointBegin + i_point;
+                if (PointIsInsideBoundingBox(rBoundingBox, *it_point, ThresholdBoundingBox)) {
+                    noalias(coordinates) = it_point->Coordinates();
+                    for (i_coord = 0; i_coord < 3; ++i_coord) {
+                        rSearchInfo.PointCoordinates.push_back(coordinates[i_coord]);
+                    }
+                    rSearchInfo.Indexes.push_back(all_points_ids[i_point]);
+                    std::vector<int> ranks(1, 0);
+                    rSearchInfo.Ranks.push_back(ranks);
```
roigcarlo:
```suggestion
                    rSearchInfo.Ranks.push_back({0});
```

kratos/utilities/search_utilities.h
```diff
@@ -543,6 +637,251 @@ class SearchUtilities
...
+        rSearchInfo.Reserve(TotalNumberOfPoints);
+        std::vector<double> all_points_coordinates(TotalNumberOfPoints * 3);
+        std::vector<IndexType> all_points_ids(TotalNumberOfPoints);
+
+        // Sync all points first
+        std::vector<int> all_points_ranks = SynchronizePointsWithRanks(itPointBegin, itPointEnd, all_points_coordinates, all_points_ids, rDataCommunicator, NumberOfPoints, TotalNumberOfPoints, IndexItIsJustCounter);
+
+        // Some definitions
+        IndexType i_coord = 0;
+        array_1d<double, 3> coordinates;
```
roigcarlo: Similar comment, but in this function you are only using it in the serial condition so even better.

kratos/utilities/search_utilities.h
roigcarlo: Not sure about if this is the exact syntax, but you get the idea

kratos/utilities/search_utilities.h
loumalouomega: why not, it is the dat communictaor interface?

loumalouomega:
Shold be Okay now @roigcarlo

kratos/utilities/search_utilities.h
roigcarlo: Because you don't use it as a vector, that's why I was suggesting the {0}.

Anyway, maybe too picky from my side, it's ok

loumalouomega: ah Okay, ithought you meant in general

roigcarlo (approved):

kratos/utilities/search_utilities.h
loumalouomega: In this particular line I use it with the gather interface from data communicator. Would you like it to refactor it?

roigcarlo: nana its ok

PR #11878, opened 2023-12-06
[GeoMechanicsApplication] Missing rotation and derivative components in schemes

### LINKED ISSUE #11874: [GeomechanicsApplication] missing rotation and derivative components in newmark UPw dynamic scheme
missing ROTATION ANGULAR_ACCELERATION and ANGULAR_VELOCITY in Newmark UPw dynamic scheme

rfaasse:
**📝 Description**
This PR aims to add rotation to the UPw schemes in a generic way.


applications/GeoMechanicsApplication/custom_strategies/schemes/backward_euler_quasistatic_U_Pw_scheme.hpp
```diff
@@ -42,31 +43,47 @@ class BackwardEulerQuasistaticUPwScheme : public NewmarkQuasistaticUPwScheme<TSp
...
-            noalias(rNode.FastGetSolutionStepValue(ACCELERATION)) = (  rNode.FastGetSolutionStepValue(VELOCITY)
-                                                                                   - rNode.FastGetSolutionStepValue(VELOCITY,1) ) / this->GetDeltaTime();
-
-            rNode.FastGetSolutionStepValue(DT_WATER_PRESSURE)        = (  rNode.FastGetSolutionStepValue(WATER_PRESSURE)
-                                                                                  - rNode.FastGetSolutionStepValue(WATER_PRESSURE, 1)) / this->GetDeltaTime();
+        // Update Acceleration, Velocity and DtPressure
+        block_for_each(
+            rModelPart.Nodes(),
+            [this](Node& rNode)
+            {
```
WPK4FEM: This reformatting makes 1 line into 5. For me it does not really become clearer with more lines.

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
```diff
@@ -54,54 +70,67 @@ class NewmarkQuasistaticUPwScheme
...
         KRATOS_TRY
 
-        if (rModelPart.GetProcessInfo()[NODAL_SMOOTHING]) {
+        if (rModelPart.GetProcessInfo()[NODAL_SMOOTHING])
+        {
             unsigned int dim = rModelPart.GetProcessInfo()[DOMAIN_SIZE];
 
             SizeType stress_tensor_size = STRESS_TENSOR_SIZE_2D;
             if (dim == N_DIM_3D)
                 stress_tensor_size = STRESS_TENSOR_SIZE_3D;
```
WPK4FEM: Can this be 2 lines ( or one including line 60 with a ternary operator )

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
```diff
@@ -54,54 +70,67 @@ class NewmarkQuasistaticUPwScheme
...
+                        rNode.FastGetSolutionStepValue(NODAL_CAUCHY_STRESS_TENSOR);
+                    if (r_nodal_stress.size1() != stress_tensor_size)
+                        r_nodal_stress.resize(stress_tensor_size, stress_tensor_size, false);
+                    noalias(r_nodal_stress) =
+                        ZeroMatrix(stress_tensor_size, stress_tensor_size);
+                    rNode.FastGetSolutionStepValue(NODAL_DAMAGE_VARIABLE) = 0.0;
+                    rNode.FastGetSolutionStepValue(NODAL_JOINT_AREA) = 0.0;
+                    rNode.FastGetSolutionStepValue(NODAL_JOINT_WIDTH) = 0.0;
+                    rNode.FastGetSolutionStepValue(NODAL_JOINT_DAMAGE) = 0.0;
+                });
```
WPK4FEM: I'm still wondering why this block or in fact the whole nodal smoothing block 73 to 132 should be part of this scheme. It touches even on variables that have no meaning if interface behavior is not present. I think that in an other issue we should try to move this to a place where it makes more sense.

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_dynamic_U_Pw_scheme.cpp
```diff
@@ -0,0 +1,232 @@
...
+    {
+        auto& result = mModel.CreateModelPart("dummy", 2);
+        result.AddNodalSolutionStepVariable(VELOCITY);
+        result.AddNodalSolutionStepVariable(ACCELERATION);
+        result.AddNodalSolutionStepVariable(DT_WATER_PRESSURE);
+        result.AddNodalSolutionStepVariable(DISPLACEMENT);
+        result.AddNodalSolutionStepVariable(WATER_PRESSURE);
+        result.AddNodalSolutionStepVariable(ROTATION);
+        result.AddNodalSolutionStepVariable(ANGULAR_VELOCITY);
+        result.AddNodalSolutionStepVariable(ANGULAR_ACCELERATION);
```
WPK4FEM: Ordering DISPLACEMENT, VELOCITY ACCELERATION, WATER_PRESSURE, DT_WATER_PRESSURE, ROTATION, ANGULAR_VELOCITY, ANGULAR_ACCELERATION or DISPLACEMENT, ROTATION, WATER_PRESSURE, VELOCITY, ANGULAR_VELOCITY, DT_WATER_PRESSURE, ACCELERATION, ANGULAR_ACCELERATION would be more logical to me.

WPK4FEM (commented):
Good generalization, now not having duplication for rotations as they share the code with the displacements.  Probably good to have a more skilled C++ programmer read through as well.

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_dynamic_U_Pw_scheme.cpp
rfaasse: Good point, done

applications/GeoMechanicsApplication/custom_strategies/schemes/backward_euler_quasistatic_U_Pw_scheme.hpp
rfaasse: Indeed, .clang-format tries to clarify that these are 2 different arguments, the first being the vector over which we loop, second is the body of the loop. It's a bit of a strange construction, I'll spend some time later to find if we can improve this formatting while keeping it okay for the rest.

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
rfaasse: Done

applications/GeoMechanicsApplication/custom_strategies/schemes/backward_euler_quasistatic_U_Pw_scheme.hpp
rfaasse: Found the option we need from clang format to do it the right way from the start

applications/GeoMechanicsApplication/custom_strategies/schemes/backward_euler_quasistatic_U_Pw_scheme.hpp
```diff
@@ -42,31 +43,45 @@ class BackwardEulerQuasistaticUPwScheme : public NewmarkQuasistaticUPwScheme<TSp
...
+                SetDerivative(r_variable_with_derivative.first_time_derivative,
+                              r_variable_with_derivative.instance, rNode);
+
+                // Make sure that setting the second_time_derivative is done
+                // after setting the first_time_derivative.
+                SetDerivative(r_variable_with_derivative.second_time_derivative,
+                              r_variable_with_derivative.first_time_derivative, rNode);
+            }
+
+            SetDerivative(DT_WATER_PRESSURE, WATER_PRESSURE, rNode);
```
markelov208, line 80: What do you think is it wise to have a function to add WATER_PRESSURE to mVariableDerivatives at the constructor? In this case there is no need in this line. Yes, there is a question with a number of components but my comment is below.

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
```diff
@@ -116,14 +138,25 @@ class NewmarkQuasistaticUPwScheme
...
+            for (const auto& variable_derivative : mVariableDerivatives)
+            {
+                if (!rModelPart.HasNodalSolutionStepVariable(variable_derivative.instance))
+                    continue;
+
+                this->CheckSolutionStepsData(r_node, variable_derivative.instance);
+                this->CheckSolutionStepsData(r_node, variable_derivative.first_time_derivative);
+                this->CheckSolutionStepsData(r_node, variable_derivative.second_time_derivative);
+
+                std::vector<std::string> components{"X", "Y", "Z"};
```
markelov208: If VariableWithTimeDerivatives will be extended with a number of components then it is possible to avoid this line. Just a loop over the available components. What do you think?

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
```diff
@@ -143,7 +176,8 @@ class NewmarkQuasistaticUPwScheme
         KRATOS_TRY
 
         // Update Acceleration, Velocity and DtPressure
```
markelov208: I think the comment just takes time. As far as I understand velocity==first derivative, acceleration is the second one. I would remove the comment, specially we have rotation components now. What is you opinion?

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_dynamic_U_Pw_scheme.cpp
```diff
@@ -0,0 +1,232 @@
...
+using LocalSpaceType = UblasSpace<double, Matrix, Vector>;
+
+class NewmarkDynamicUPwSchemeTester
+{
+public:
+    Model mModel;
+    NewmarkDynamicUPwScheme<SparseSpaceType, LocalSpaceType> mScheme =
+        CreateValidScheme();
+
+    explicit NewmarkDynamicUPwSchemeTester(const bool Add3DDofs = true)
```
markelov208: Naming is very difficult task. For me the better name is AddThridDof or ThreeDCase. What is your opinion?

markelov208 (commented):
Richard, well done. Just a few comments but probably the better is enemy of good. ;)

applications/GeoMechanicsApplication/custom_strategies/schemes/backward_euler_quasistatic_U_Pw_scheme.hpp
rfaasse: WATER_PRESSURE at this moment is already stored in the baseclass. However, the problem now is that the BackwardEulerQuasistaticUPwScheme still derives from the GeneralizedNewmarkScheme instead of the BackwardEulerScheme (see snippet right). We have made efforts as you see to improve the structure, but we didn't fix everything yet (to not get gigantic PRs). I would propose to fix this as part of our next refactoring effort (I created a draft issue for it on the backlog).
image

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
rfaasse: Agreed, it's no longer necessary, removed it!

applications/GeoMechanicsApplication/tests/cpp_tests/test_newmark_dynamic_U_Pw_scheme.cpp
rfaasse: I think we should keep in the name that it is indeed 3D, so I'll change it to ThreeDCase. I agree that's a bit more clear

applications/GeoMechanicsApplication/custom_strategies/schemes/newmark_quasistatic_U_Pw_scheme.hpp
rfaasse: Thanks for pointing to this part, there is still a mistake in it: We should not check for "Z" here. since it should be allowed to be a 2D case. 

Since in that case these components with only XY are only used here and it's not really general, I don't think we gain a lot with storing it in every VariableWithTimeDerivatives.

markelov208: Agreed if it is only here then there is no need to make it general.

applications/GeoMechanicsApplication/custom_strategies/schemes/backward_euler_quasistatic_U_Pw_scheme.hpp
markelov208: Thank you for the clarification. Sure it can be done later.

markelov208 (approved):

WPK4FEM (approved):
Nice, even less codesmells.

PR #11887, opened 2023-12-11
[Core][IO] added check for zero size in WriteIntegrationVectorContainerVariable …

### LINKED ISSUE #11885: [IO] In vtk_ouput.cpp non initialized variables are written to file, which crashes the computation
In vtk_ouput.cpp non initialized variables are written to file, which crashes the computation.

for example in vtk_ouput.cpp -> WriteIntegrationVectorContainerVariable, the following code is present which crashes

` // determining size of results
 const auto& r_process_info = mrModelPart.GetProcessInfo();
 std::vector<TVarType> tmp_result;
 rContainer.begin()->CalculateOnIntegrationPoints(rVariable, tmp_result, r_process_info);
 const int res_size = tmp_result[0].size();
`

if rContainer doesnt contain rVariable, tmp_result has size 0, thus tmp_result[0].size() is an invalid command which crashes the computation

Note that the computation does not crash when i request GIDOutput

aronnoordam:
added check to check if intergration point data has size==0 in vtk_output.cpp

### DIFF
--- a/kratos/input_output/vtk_output.cpp
+++ b/kratos/input_output/vtk_output.cpp
@@ -964,6 +964,11 @@ void VtkOutput::WriteIntegrationVectorContainerVariable(
     const auto& r_process_info = mrModelPart.GetProcessInfo();
     std::vector<TVarType> tmp_result;
     rContainer.begin()->CalculateOnIntegrationPoints(rVariable, tmp_result, r_process_info);
+    
+    // if no results, return
+    if (tmp_result.size() == 0) {
+        return;
+    }
     const int res_size = tmp_result[0].size();
 
     rFileStream << rVariable.Name() << " " << res_size << " " << rContainer.size() << "  float\n";


loumalouomega (approved):

kratos/input_output/vtk_output.cpp
```diff
@@ -964,6 +964,11 @@ void VtkOutput::WriteIntegrationVectorContainerVariable(
     const auto& r_process_info = mrModelPart.GetProcessInfo();
     std::vector<TVarType> tmp_result;
     rContainer.begin()->CalculateOnIntegrationPoints(rVariable, tmp_result, r_process_info);
+    
+    // if no results, return
+    if (tmp_result.size() == 0) {
+        return;
+    }
```
philbucher, line 971: in which case would one not have results? If an element does not write output for a variable?

I am mostly concerned how this works in MPI, when some partitions write this and some dont, do you know?

philbucher (changes requested):
Blocking to prevent MPI problems

kratos/input_output/vtk_output.cpp
```diff
@@ -964,6 +964,11 @@ void VtkOutput::WriteIntegrationVectorContainerVariable(
...
     const int res_size = tmp_result[0].size();
 
     rFileStream << rVariable.Name() << " " << res_size << " " << rContainer.size() << "  float\n";
```
philbucher, line 974: my concern is here, this would only be written on some ranks and thus make the files inconsistent

philbucher (approved):
I checked, I think it is ok, thanks for the fix

The handling for MPI with mixed entities is not ideal (already in master, not related to this MR), but should not really be a usecase

aronnoordam:
okay, thanks for the review

PR #11899, opened 2023-12-19
Rename `KRATOS_EXPECT_MATRIX_EQUAL` to `KRATOS_EXPECT_MATRIX_EQ`

avdg81:
**📝 Description**
This PR extracts the changes related to the macro renaming from PR #11506. The aim is to assist in breaking down PR #11506 into smaller pieces that can be reviewed more easily and which can be merged back separately. This commit renames the macro and it updates all of its usages.

### DIFF
--- a/applications/HDF5Application/tests/test_hdf5_file.cpp
+++ b/applications/HDF5Application/tests/test_hdf5_file.cpp
@@ -947,7 +947,7 @@ KRATOS_TEST_CASE_IN_SUITE(HDF5_File_ReadWriteParameterAttribute1, KratosHDF5Test
         mat_values.data()[2] = 2;
         mat_values.data()[3] = 3;
         KRATOS_EXPECT_TRUE(attr_in["matrix"].IsMatrix());
-        KRATOS_EXPECT_MATRIX_EQUAL(attr_in["matrix"].GetMatrix(), mat_values);
+        KRATOS_EXPECT_MATRIX_EQ(attr_in["matrix"].GetMatrix(), mat_values);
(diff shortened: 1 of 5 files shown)


avdg81:
@roigcarlo: May I kindly ask you to review this small PR?

@loumalouomega: Are these simple changes in line with one of your suggestions for PR #11506?

loumalouomega:
I am okay with changes

roigcarlo (approved):
:+1:

philbucher:
does GTest have a matrix comparison?
Does it behave the same as ours?

philbucher:
> does GTest have a matrix comparison? Does it behave the same as ours?

ping @avdg81

avdg81:
> > does GTest have a matrix comparison? Does it behave the same as ours?
> 
> ping @avdg81

@philbucher: As far as I know, GTest doesn't have an `EXPECT` and/or `ASSERT` macro that is targeted at matrices. However, I would expect that we can write a validation function (or macro) for this purpose that is composed of the elementary assertions that GTest offers. Alternatively, I can imagine that we write a [matcher](https://google.github.io/googletest/reference/matchers.html) for this purpose that we then use in conjunction with the [`EXPECT_THAT`](https://google.github.io/googletest/reference/assertions.html#EXPECT_THAT) macro. Does this answer your question?

roigcarlo:
Adding to @avdg81 answer, it does not have a direct matrix comparison, but it can compare any container that is iterable. The problem is that our matrices are not iterable (or at least it does not like the type of the iterator).

For the vectors its already done with iterators:

```c++
EXPECT_THAT(a, Pointwise(DoubleNear(tolerance), b));
```

PR #11901, opened 2023-12-19
[GeoMechanicsApplication] Fixed two warnings reported by the clang compiler

avdg81:
**📝 Description**
Don't use `#pragma once` in implementation files (`*.cpp`).


markelov208 (approved):
Anne, nice to find this.

PR #11902, opened 2023-12-20
[GeoMechanicsApplication] Remove redundant pragma once in cpp file dgeoflow (gives clang warning)

rfaasse:
Found one more pragma once in a cpp file. I did a project wide search after this, with file mask *.cpp and it seems this is the last one.


avdg81 (approved):
Thanks for having performed a project-wide search for `#pragma once` in `*.cpp` files. And thanks for fixing this one. One compiler warning less :-)

PR #11908, opened 2023-12-21
[Core] Minor improvement `search_utilities.h`

loumalouomega:
**📝 Description**

Some minor changes in `search_utilities.h`, instead of manually iterating over the container with a for loop, you can use a range-based for loop which is more idiomatic in modern C++.

Also using std::make_shared instead of `new` is a good practice in modern C++. It's more efficient and safer, as it combines the allocation of memory for the object and the creation of the shared_ptr into one step, which can prevent potential memory leaks.

**🆕 Changelog**

-  [Minor improvement `search_utilities.h`](https://github.com/KratosMultiphysics/Kratos/commit/92ed51e6e4b5c2e67faf929459bc51bb7b2b1e21)

### DIFF
--- a/kratos/utilities/search_utilities.h
+++ b/kratos/utilities/search_utilities.h
@@ -454,7 +454,7 @@ public:
 
         // Creating the points
         PointVector points;
-        point.reserve(rStructure.size());
+        points.reserve(rStructure.size());
         for (const auto& r_element : rStructure) {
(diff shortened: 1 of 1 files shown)


philbucher (approved):

loumalouomega:
Huum, it compiles in my mchiane, well is not that important.

PR #11920, opened 2024-01-02
[GeoMechanicsApplication] Formatted the code in the custom_elements folder using clang-format

### LINKED ISSUE #11919: [GeoMechanicsApplication] Format GeoMechanics code base using clang-format
In this issue, the geomechanics code base will be formatted using clang-format. It will be done folder by folder, which is why there are multiple PRs linked to this issue.

rfaasse:
In this PR clang-format is used to format the custom_elements in geomechanics to make all formatting consistent


applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_FIC_element.cpp
```diff
@@ -292,33 +280,33 @@ void UPwSmallStrainFICElement<2,3>::
...
+            noalias(mNodalConstitutiveTensor[i][j]) =
+                column(AuxNodalConstitutiveTensor, j);
     }
 
     /* INFO:
      *
-     *                            [ ( |D00-0|   |D01-0|   |D02-0| )   ( |D10-0|   |D11-0|   |D12-0| ) ]
-     * mNodalConstitutiveTensor = [ ( |D00-1|   |D01-1|   |D02-1| )   ( |D10-1|   |D11-1|   |D12-1| ) ]
+     *                            [ ( |D00-0|   |D01-0|   |D02-0| )   ( |D10-0| |D11-0|   |D12-0| ) ]
+     * mNodalConstitutiveTensor = [ ( |D00-1|   |D01-1|   |D02-1| )   ( |D10-1| |D11-1|   |D12-1| ) ]
```
rfaasse: Not sure why it's changing this one

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_FIC_element.cpp
```diff
@@ -500,116 +485,118 @@ void UPwSmallStrainFICElement<TDim,TNumNodes>::
...
 {
     // Quadrilateral_2d_4 with GI_GAUSS_2
     KRATOS_TRY
 
-    BoundedMatrix<double,4,8> ShapeFunctionsGradientsContainer; //NumGPoints X TDim*TNumNodes
+    BoundedMatrix<double, 4, 8> ShapeFunctionsGradientsContainer; // NumGPoints X TDim*TNumNodes
     unsigned int index;
 
-    for (unsigned int i = 0; i < 4; ++i) //NumGPoints
+    for (unsigned int i = 0; i < 4; ++i) // NumGPoints
```
rfaasse, line 545: Apparently, when there's a comment the opening accolade { is not put on the same line

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_FIC_element.cpp
```diff
@@ -624,119 +611,135 @@ void UPwSmallStrainFICElement<2,4>::
...
-        rNodalShapeFunctionsGradients[i][17] = AuxNodalShapeFunctionsGradients(5,index+2);
-        rNodalShapeFunctionsGradients[i][18] = AuxNodalShapeFunctionsGradients(6,index);
-        rNodalShapeFunctionsGradients[i][19] = AuxNodalShapeFunctionsGradients(6,index+1);
-        rNodalShapeFunctionsGradients[i][20] = AuxNodalShapeFunctionsGradients(6,index+2);
-        rNodalShapeFunctionsGradients[i][21] = AuxNodalShapeFunctionsGradients(7,index);
-        rNodalShapeFunctionsGradients[i][22] = AuxNodalShapeFunctionsGradients(7,index+1);
-        rNodalShapeFunctionsGradients[i][23] = AuxNodalShapeFunctionsGradients(7,index+2);
+        index = i * 3;
+
+        rNodalShapeFunctionsGradients[i][0] = AuxNodalShapeFunctionsGradients(0, index);
```
rfaasse: This is a bit of a shame due to the 80 column width

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_FIC_element.cpp
```diff
@@ -826,233 +826,240 @@ void UPwSmallStrainFICElement<3,8>::
...
+            rFICVariables.StrainGradients(0, index + 1);
     }
 
     /* INFO:
      *
-     *                                                    ( |N0xx|   |N1xx|   |N2xx|   |N3xx| )
-     * rFICVariables.ShapeFunctionsSecondOrderGradients = ( |N0yy|   |N1yy|   |N2yy|   |N3yy| )
-     *                                                    ( |N0xy| , |N1xy| , |N2xy| , |N3xy| )
+     *                                                    ( |N0xx|   |N1xx|
+     * |N2xx|   |N3xx| ) rFICVariables.ShapeFunctionsSecondOrderGradients = (
```
rfaasse: This one formats in the wrong way. Maybe we can exclude these multi-line comments somehow

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_axisymmetric_element.hpp
```diff
@@ -43,99 +41,102 @@ class KRATOS_API(GEO_MECHANICS_APPLICATION) UPwSmallStrainAxisymmetricElement :
...
 
-///----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+    ///----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 
     /// Default Constructor
-    UPwSmallStrainAxisymmetricElement(IndexType NewId = 0) :
-        UPwSmallStrainElement<TDim,TNumNodes>( NewId ) {}
+    UPwSmallStrainAxisymmetricElement(IndexType NewId = 0)
+        : UPwSmallStrainElement<TDim, TNumNodes>(NewId)
+    {
```
rfaasse, line 51: Apparently we don't allow the {} to be on the same line.

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_element.cpp
```diff
@@ -48,62 +45,70 @@ int UPwSmallStrainElement<TDim, TNumNodes>::Check(const ProcessInfo& rCurrentPro
...
     // Verify specific properties
     KRATOS_ERROR_IF_NOT(rProp.Has(IGNORE_UNDRAINED))
-        << "IGNORE_UNDRAINED does not exist in the parameter list" << this->Id() << std::endl;
+        << "IGNORE_UNDRAINED does not exist in the parameter list" << this->Id()
+        << std::endl;
 
     if (!rProp[IGNORE_UNDRAINED]) {
-        KRATOS_ERROR_IF(!rProp.Has( BULK_MODULUS_FLUID ) || rProp[BULK_MODULUS_FLUID] < 0.0)
-            << "BULK_MODULUS_FLUID has Key zero, is not defined or has an invalid value at element" << this->Id() << std::endl;
+        KRATOS_ERROR_IF(!rProp.Has(BULK_MODULUS_FLUID) || rProp[BULK_MODULUS_FLUID] < 0.0) << "BULK_MODULUS_FLUID has Key zero, is not defined or has an invalid value at element"
```
rfaasse: This does not seem to be improved

rfaasse (commented):
A few weird behaviors, let's see how to fix it

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_FIC_element.cpp
rfaasse: I turned clang-format off for these two comments

RiccardoRossi (commented):
I have no problem with this sort of changes in the GeoMechanics application.

I am writing this comment simply to tell that in the past we preferred not to go for this kind of automatic formatting options as it would effectively destroy (read, make meaningless) the commit history in the affected files.

Still, as i am telling please feel free to go ahead if you consider. I am writing this just to inform you of a side effect

rfaasse:
> I have no problem with this sort of changes in the GeoMechanics application.
> 
> I am writing this comment simply to tell that in the past we preferred not to go for this kind of automatic formatting options as it would effectively destroy (read, make meaningless) the commit history in the affected files.
> 
> Still, as i am telling please feel free to go ahead if you consider. I am writing this just to inform you of a side effect

@RiccardoRossi Thank you for expressing your concern here! It is indeed one of the downsides that we considered before choosing this approach. We still value the formatting consistency above losing clarity in the history. 

The commit history is of course still intact and we are exploring ways to ignore certain commits. Adding a .git-blame-ignore-revs file should help us here. GitHub should also be able to pick this up when checking the history, see also this [entry](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) in the GitHub Docs.

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_FIC_element.cpp
rfaasse: Disabled clang format for this comment by adding the `// clang-format off` and `// clang-format on` comments around the info block.

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_axisymmetric_element.hpp
rfaasse: With our current brace configuration, I don't see how to fix this, since the constructor is already multi-line.

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_element.cpp
rfaasse: However, it looks like this should be abstracted into a function anyway, so I would propose to leave it as is.

rfaasse:
> yep seems to be failing for everyone, they must have changed to sources. I will update it asap.

Clear, thank you!

FYI @avdg81

roigcarlo:
Should work with the changes on #11930

rfaasse:
> Should work with the changes on #11930

Perfect, thanks for the quick fix!

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_element.cpp
```diff
@@ -972,54 +941,49 @@ void UPwSmallStrainElement<TDim, TNumNodes>::CalculateMassMatrix(MatrixType& rMa
...
 
     // Create general parameters of retention law
     RetentionLaw::Parameters RetentionParameters(this->GetProperties(), rCurrentProcessInfo);
 
     // Defining shape functions at all integration points
     // Defining necessary variables
-    BoundedMatrix<double, TDim+1, TNumNodes*(TDim+1)> Nut              = ZeroMatrix(TDim+1, TNumNodes*(TDim+1));
-    BoundedMatrix<double, TDim+1, TNumNodes*(TDim+1)> AuxDensityMatrix = ZeroMatrix(TDim+1, TNumNodes*(TDim+1));
-    BoundedMatrix<double, TDim+1, TDim+1>             DensityMatrix    = ZeroMatrix(TDim+1, TDim+1);
+    BoundedMatrix<double, TDim + 1, TNumNodes*(TDim + 1)> Nut = ZeroMatrix(TDim + 1, TNumNodes * (TDim + 1));
```
WPK4FEM, line 960: Does clang format allow for special formatting of indices? Here it seems to apply the rule of spaces around +, but not around  all *.  The slightly more compact formatting on the unchanged lines here is faster to read ( for me ).

applications/GeoMechanicsApplication/custom_elements/geo_curved_beam_element.hpp
```diff
@@ -30,192 +30,178 @@ namespace Kratos
...
     {
-        KRATOS_SERIALIZE_SAVE_BASE_CLASS( rSerializer, Element )
+        KRATOS_SERIALIZE_SAVE_BASE_CLASS(rSerializer, Element)
     }
 
-    void load(Serializer& rSerializer) override
-    {
-        KRATOS_SERIALIZE_LOAD_BASE_CLASS( rSerializer, Element )
-    }
+    void load(Serializer& rSerializer) override{KRATOS_SERIALIZE_LOAD_BASE_CLASS(rSerializer, Element)}
```
WPK4FEM: save remains on 4 lines, which gives a recognizable function.
The load goes to 1 line, but becomes harder to read for me.

applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
```diff
@@ -63,205 +67,223 @@ int  SmallStrainUPwDiffOrderElement::Check( const ProcessInfo& rCurrentProcessIn
     if (rGeom.DomainSize() < 1.0e-15)
         KRATOS_ERROR << "DomainSize < 1.0e-15 for the element " << this->Id() << std::endl;
 
-    //verify that the variables are correctly initialized
-    // Verify specific properties
+    // verify that the variables are correctly initialized
+    //  Verify specific properties
```
WPK4FEM: What causes the extra space in this comment. Does it try to retain an indentation.

applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
```diff
@@ -1616,60 +1610,60 @@ void SmallStrainUPwDiffOrderElement::InitializeElementVariables(ElementVariables
...
 
-        if (Dim > 2 ) {
-            rVariables.BodyAcceleration[Local_i+2]   = BodyAccelerationAux[2];
-            rVariables.DisplacementVector[Local_i+2] = rGeom[i].FastGetSolutionStepValue(DISPLACEMENT_Z);
-            rVariables.VelocityVector[Local_i+2]     = rGeom[i].FastGetSolutionStepValue(VELOCITY_Z);
+        if (Dim > 2) {
+            rVariables.BodyAcceleration[Local_i + 2] = BodyAccelerationAux[2];
+            rVariables.DisplacementVector[Local_i + 2] = rGeom[i].FastGetSolutionStepValue(DISPLACEMENT_Z);
+            rVariables.VelocityVector[Local_i + 2] = rGeom[i].FastGetSolutionStepValue(VELOCITY_Z);
         }
```
WPK4FEM, line 1643: For some reason, the = signs do no longer align.

applications/GeoMechanicsApplication/custom_elements/steady_state_Pw_piping_element.hpp
```diff
@@ -102,50 +104,46 @@ class KRATOS_API(GEO_MECHANICS_APPLICATION) SteadyStatePwPipingElement :
...
 
-    void load(Serializer& rSerializer) override
-    {
-        KRATOS_SERIALIZE_LOAD_BASE_CLASS( rSerializer, Element )
-    }
+    void load(Serializer& rSerializer) override{KRATOS_SERIALIZE_LOAD_BASE_CLASS(rSerializer, Element)}
 
     /// Assignment operator.
-    SteadyStatePwPipingElement& operator=(SteadyStatePwPipingElement const& rOther);
+    SteadyStatePwPipingElement&
```
WPK4FEM: Why split over 2 lines, is this really longer than 100 characters

WPK4FEM (commented):
Hi Richard,
69 was a lot of files and consequently lines. Formatting goes quite well, I came across some that I did not understand ( or like ).
Wijtze Pieter

applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
rfaasse: Because of the extra spaces, aligning is no longer possible, since it would become more than 100 characters

applications/GeoMechanicsApplication/custom_elements/steady_state_Pw_piping_element.hpp
rfaasse: No idea, it would be 85... I'll have a quick look to see if I can fix it

applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
rfaasse: I think so indeed, I removed the extra space on the second line

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_element.cpp
rfaasse: As far as I can found, it's either spaces around any calculation (so also all 'normal' formulas), or remove the spaces everywhere, which I think makes it less readable. My theory what happens with the * here, is that clang-format does not know what the type is of TNumNodes. Therefore, it might think TNumNodes* is a pointer to something and therefore it does not add the extra space. At least for '-' and '/' it also adds the spaces. '*' seems to be the only exception

applications/GeoMechanicsApplication/custom_elements/geo_curved_beam_element.hpp
rfaasse: I agree, but the only way to fix it, is to never allow single-line functions.

applications/GeoMechanicsApplication/custom_elements/small_strain_U_Pw_diff_order_element.cpp
rfaasse: I think this is actually a bug in clang-format. For some reason, it does not take into account lines which overflow. For example, this
```
    shortname = 0;
    slightlyLongerName = 0;
    veryLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongName = 0;
```

Is nicely formatted to
```
    shortname                                                                                = 0;
    slightlyLongerName                                                                       = 0;
    veryLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongName = 0;
```

However, this:
```
    shortname = 0;
    slightlyLongerName = 0;
    veryLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongName = 20000000; // over the column limit
```

will be formatted into:
```
    shortname          = 0;
    slightlyLongerName = 0;
    veryLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongName = 20000000;
```

It makes sense that the column limit is taken into account, but here it seems its too strict and does not take into account that the rhs of an assignment could have a different length, such that this is done correctly

I found they made some changes here quite recently because there was a bug that the column width was not taken into account at all with assignemnts, but maybe haven't considered everything:
https://github.com/llvm/llvm-project/issues/56444

applications/GeoMechanicsApplication/custom_elements/steady_state_Pw_piping_element.hpp
rfaasse: I really don't know why this happens. For some reason it only happens when the serialization functions are above the assignment ones. There were already some elements where this wasn't a problem. What I did for now is move the operators around, then the formatting is fine

rfaasse:
@WPK4FEM Thank you for your extensive and thorough review. I wasn't able to get everything right, but I think looking into if these unexplained behaviors are really bugs in the latest version of clang-format is out of scope for this PR in my opinion. Let me know if you find the answers to your (very valid) questions satisfactory!

rfaasse:
As a bonus of the last fix, it seems that the load functions are now also written out in the same way as the save instead of single-line (when looking at [the last two commits](https://github.com/KratosMultiphysics/Kratos/pull/11920/files/65aa5458757b0ccd7ffab8a7c7eda86ecf9237da..0013d2f6acfc3dba7f1b13fc1258f338c58ca7e1))

applications/GeoMechanicsApplication/custom_elements/U_Pw_small_strain_FIC_element.hpp
```diff
@@ -226,6 +226,12 @@ class KRATOS_API(GEO_MECHANICS_APPLICATION) UPwSmallStrainFICElement
 
     ///----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 
+    /// Assignment operator.
+    UPwSmallStrainFICElement& operator=(UPwSmallStrainFICElement const& rOther);
+
+    /// Copy constructor.
+    UPwSmallStrainFICElement(UPwSmallStrainFICElement const& rOther);
+
```
WPK4FEM, line 234: Hidden by private. I think the preferred alternative is explicitly deleting them.
This is brought to the attention by the re-ordering that clang format does here.
Out of scope, we will look at those things later.

WPK4FEM (approved):
Difficult to look at formatting only, but this revision should be about formatting only.
Good step towards consistency.

PR #11926, opened 2024-01-05
[GeomechanicsApplication] fixed wrongly defined dirichlet condition in residualbased_block_builder_and_solver_with_mass_and_damping

aronnoordam:
dirichlet boundary condition was applied to the rhs incorectly in the first iteration if there was no damping


WPK4FEM (approved):
Nice correction. Thanks for supplying the accompanying test.

PR #11937, opened 2024-01-09
[TrilinosApplication] Fix master DoFs in block builder and solver

loumalouomega:
**📝 Description**

Until now only local slave DoFs were filtered for the `mMasterIds` vector, in here all partitions slave DoFs are considered. I added two dynamic size C++ test (until now I was only checking up two partitions, therefore I was checking many cornercases), so the size of the test depends in the number of partitions considered.

(I discovered this looking for another bug) 

**🆕 Changelog**

- [Fix master dofs](https://github.com/KratosMultiphysics/Kratos/commit/b9c37f56e27a2c26e9bd4f6a72f0e0a363bceb6c)
- [Adding dynamic size testing in order to properly check fix](https://github.com/KratosMultiphysics/Kratos/commit/3a40ac97cb60f5cdaed6996688a8480f1387e915)

### DIFF
--- a/applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
+++ b/applications/TrilinosApplication/custom_strategies/builder_and_solvers/trilinos_block_builder_and_solver.h
@@ -1310,8 +1310,12 @@ protected:
             for (IndexType i = 0; i < number_of_local_rows; ++i) {
                 temp_master_ids.insert(mFirstMyId + i);
             }
-            for (auto id_slave : mSlaveIds) {
-                temp_master_ids.erase(id_slave);
+
+            // Remove ids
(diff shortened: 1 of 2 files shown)


loumalouomega:
Ci broken?

roigcarlo:
Github api had a dowtime of about 1 min, it failed some of my merges too. Bad luck :(

loumalouomega:
@roigcarlo can you check this?, in this case a test is actually added

roigcarlo:
I'll take a look after lunch

roigcarlo (approved):
Ok,

Just as a curiosity: Why did you change some of the reference values in the tests, I get you want `E_100` as an alias and also `2 * E_100`, but then you can use `E_100 * 0.5` for `1034500000` or `E_100 * 0.25` for `517250000`.

PR #11938, opened 2024-01-10
[GeoMechanicsApplication] time stepping issue, avoid extremely small steps near end_time

### LINKED ISSUE #11921: [GeomechanicsApplication] geomechanics application occasionally hangs after the final step, due to floating errors
In KeepAdvancingSolutionLoop in geomechanics_solver.py, the current time is compared to the end time, but occasionally, the difference between the current time and the end time is too small and not relevant such that it does another time step which cannot be solved by kratos.

solution is to stop the solution loop when the current time is smaller then the end time minus an epsilon

WPK4FEM:
**📝 Description**
The adaptive time stepping scheme was sometimes creating extremely small time steps to try and reach the end_time.
This is prevented by slightly enlarging the step before the problematic small step, on the condition that it otherwise would have stopped at a time point less than 1.E-3 times its stepsize short of the end time.

Adapted geomechanics_analysis.py, dgeosettlement.cpp and added unit test and a python regression test.


applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.cpp
```diff
@@ -85,6 +85,12 @@ void AdaptiveTimeIncrementor::PostTimeStepExecution(const TimeStepEndState& rRes
 
     // Avoid incrementing the time beyond the end time
     mDeltaTime = std::min(mDeltaTime, mEndTime - rResultantState.time);
+
+    // Avoid very small remaining time steps
+    double small_time_increment = 1.E-3 * mDeltaTime;
```
avdg81: I would prefer to make this variable `const` (since it is not supposed to be changed):
```suggestion
    const auto small_time_increment = 1.E-3 * mDeltaTime;
```
(And yes, I prefer `auto`, but I'll leave that up to you.)

avdg81 (commented):
This looks nice and clean to me. And I'm very pleased about the new tests that you've added. I only have one minor remark.

applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
```diff
@@ -168,7 +175,8 @@ def RunSolutionLoop(self):
                 KratosMultiphysics.Logger.PrintInfo(self._GetSimulationName(), "cycle: ", number_cycle)
 
                 # set new_time and delta_time in the nonlinear solver
-                new_time = t + self.delta_time
+                if (number_cycle > 0):
```
rfaasse: I don't think number_cycle will ever be 0. We initialize it with 0, but at the start of the while loop we increment it. I think it would be more clear to initialize with 1 and increment at the end (meaning we don't need this if statement), but I leave that up to you

applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
```diff
@@ -183,11 +191,11 @@ def RunSolutionLoop(self):
...
                         if (self._GetSolver().GetComputingModelPart().ProcessInfo[KratosMultiphysics.NL_ITERATION_NUMBER] < self.min_iterations):
                             # scale up next step
-                            self.delta_time = min( self.increase_factor * self.delta_time, self.max_delta_time )
-                            if ( new_time + self.delta_time <= self.end_time ):
+                            self.delta_time = min(self.increase_factor * self.delta_time, self.max_delta_time)
+                            if (new_time + self.delta_time <= self.end_time):
                                 KratosMultiphysics.Logger.PrintInfo(self._GetSimulationName(), "Up-scaling to delta time: ", self.delta_time)
                             else:
-                                KratosMultiphysics.Logger.PrintInfo(self._GetSimulationName(), "Up-scaling to reach end_time: ", self.end_time)
+                                KratosMultiphysics.Logger.PrintInfo(self._GetSimulationName(), "Up-scaling to reach end_time: ", self.delta_time)
```
rfaasse, line 194: Here I would change the logging a bit as well then, to make it say "Up-scaling to reach end_time, using a delta time of "" , self.delta_time.

applications/GeoMechanicsApplication/tests/test_time_integration/blocks.mdpa
rfaasse, line 1: This seems to be quite a large grid. If it costs very little time (because this is really not that important), it might be worth it to try and make the test case smaller

rfaasse (dismissed):
Nice fix, thank you for the work! I have a couple of questions, but none of them blocking.

applications/GeoMechanicsApplication/custom_workflows/adaptive_time_incrementor.cpp
WPK4FEM: done

applications/GeoMechanicsApplication/tests/test_time_integration/blocks.mdpa
WPK4FEM: Very tricky action. It is indeed large, I forgot to move it to the nightly suite. Will take care of that.

applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
WPK4FEM: good catch. Anne also indicated that the condition did not add clarity. Removed the condition.

applications/GeoMechanicsApplication/python_scripts/geomechanics_analysis.py
WPK4FEM: Looked how that plays out in a transcript of the logging. Kept it short as is.

rfaasse (approved):
Thanks for incorporating the changes! Ready to be merged in my opinion!

PR #11939, opened 2024-01-11
[GeoMechanicsApplication] Split ApplyK0ProcedureProcess into a header and source file

rfaasse:
**📝 Description**
Minor PR to split the ApplyK0Procedure class .hpp file in a .h and .cpp file.


avdg81 (approved):
Thanks for splitting the original `.hpp` file in a header file and an implementation file. This makes the code easier to read, since the class's interface has been nicely separated from the implementation details. Well done Richard!

PR #11944, opened 2024-01-12
[FastPR][Fluid] Fluid auxiliary utils to header

uxuech:
**📝 Description**
Moving the implementation of three functions to the header so I can use them in the hydraulics application.

### DIFF
--- a/applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.h
+++ b/applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.h
@@ -61,7 +61,19 @@ public:
      * @return true The element is split
      * @return false The element is not split
      */
-    static inline bool IsSplit(const Vector& rElementDistancesVector);
+    static inline bool IsSplit(const Vector& rElementDistancesVector)
+    {
+        std::size_t n_pos(0);
+        std::size_t n_neg(0);
+        const std::size_t pts_number = rElementDistancesVector.size();
+        for (std::size_t i_node = 0; i_node < pts_number; ++i_node){
+            if (rElementDistancesVector[i_node] > 0.0)
+                n_pos++;
+            else
+                n_neg++;
+        }
+        return (n_pos > 0 && n_neg > 0) ? true : false;
+    }
 
     /**
      * @brief Checks if an element is positive
@@ -70,7 +82,16 @@ public:
      * @return true The element is positive
      * @return false The element is not positive
      */
-    static inline bool IsPositive(const Vector &rElementDistancesVector);
+    static inline bool IsPositive(const Vector &rElementDistancesVector)
+    {
+        std::size_t n_pos (0);
+        const std::size_t pts_number = rElementDistancesVector.size();
+        for (std::size_t i_node = 0; i_node < pts_number; ++i_node){
+            if (rElementDistancesVector[i_node] > 0.0)
+                n_pos++;
+        }
+        return (n_pos == pts_number) ? true : false;
+    }
 
     /**
      * @brief Checks if an element is negative
@@ -79,7 +100,16 @@ public:
      * @return true The element is negative
      * @return false The element is not negative
      */
-    static inline bool IsNegative(const Vector &rElementDistancesVector);
+    static inline bool IsNegative(const Vector &rElementDistancesVector)
+    {
+        std::size_t n_neg (0);
+        const std::size_t pts_number = rElementDistancesVector.size();
+        for (std::size_t i_node = 0; i_node < pts_number; ++i_node){
+            if (rElementDistancesVector[i_node] < 0.0)
+                n_neg++;
+        }
+        return n_neg == pts_number;
+    }
 
     /**
      * @brief Calculate the fluid volume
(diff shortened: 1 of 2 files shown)


philbucher (changes requested):
If this is not working then something else is wrong :)

If the hydraulics app depends on the fluid app, then you need to add the dependency properly (i.e. use the dependency macro and link against the fluid core)

I.e. you need the equivalent of this
https://github.com/KratosMultiphysics/Kratos/blob/c56cf92433e7947081914ebe4ab386cb1187515d/applications/ContactStructuralMechanicsApplication/CMakeLists.txt#L5

and this
https://github.com/KratosMultiphysics/Kratos/blob/c56cf92433e7947081914ebe4ab386cb1187515d/applications/ContactStructuralMechanicsApplication/CMakeLists.txt#L40

uxuech:
I missed the `kratos_add_dependency` (now it is added in my other branch). In any case, adding it changes nothing (the linking error appears again). Note that these three functions are `static`. After having discussed it with @roigcarlo it seems that this is the issue and there's no other way to do it.

roigcarlo:
@philbucher The problem is more subtle, it has nothing to do with the linkage.

The use of `static inline` in the declaration (not the definition) of the utility functions imply that:
- The compiled must not check the ODR for those functions
- The definition of the function will exists only in the context of the file its defined on (in this case, as is the member of an auxiliary class it will be seen in the whole transactional unit, so the whole FluidDynamics shared lib)

The moment we try to access this definition outside the shared lib it won't be available, producing a linking error.

Not declaring the functions `static inline` and making them just static could work, the ODR then applies, and the compiler is forced to check that all transactional units have the same definiton. Sadly this does not solve the problem of how we pass the function definition to the Hydraulics compilation.

For that we have two alternatives: 
- Move the definition to the header, ensuring it lives in the same context as the declaration. Then we can leave or remove the `inline`
- Compile the definition of the function in both transactional untis (with `inline`) but not being able to guarantee ODR. 

Personally I think the cleanest solution here is just to move the definitions in the header.

philbucher:
if we had the functions  as free functions and not as part of the class, then it would work also no?

roigcarlo:
Nop, if we move them outside the class we will only cause the FluidDynamics to also be unable to see the definitions (it will pass the CI thou, because the unitary build will "fix" the problem).

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.h
```diff
@@ -70,7 +82,16 @@ class KRATOS_API(FLUID_DYNAMICS_APPLICATION) FluidAuxiliaryUtilities
...
-    static inline bool IsPositive(const Vector &rElementDistancesVector);
+    static inline bool IsPositive(const Vector &rElementDistancesVector)
+    {
+        std::size_t n_pos (0);
+        const std::size_t pts_number = rElementDistancesVector.size();
+        for (std::size_t i_node = 0; i_node < pts_number; ++i_node){
+            if (rElementDistancesVector[i_node] > 0.0)
+                n_pos++;
+        }
+        return (n_pos == pts_number) ? true : false;
```
philbucher, line 93:
```suggestion
        return n_pos == pts_number;
```

applications/FluidDynamicsApplication/custom_utilities/fluid_auxiliary_utilities.h
```diff
@@ -61,7 +61,19 @@ class KRATOS_API(FLUID_DYNAMICS_APPLICATION) FluidAuxiliaryUtilities
...
+        std::size_t n_pos(0);
+        std::size_t n_neg(0);
+        const std::size_t pts_number = rElementDistancesVector.size();
+        for (std::size_t i_node = 0; i_node < pts_number; ++i_node){
+            if (rElementDistancesVector[i_node] > 0.0)
+                n_pos++;
+            else
+                n_neg++;
+        }
+        return (n_pos > 0 && n_neg > 0) ? true : false;
```
philbucher, line 75:
```suggestion
        return n_pos > 0 && n_neg > 0;
```

philbucher (approved):
ah man too bad

made some comments to improve the implementation

matekelemen:
Please remove the inline, it doesn't do what the original author thought it'd do, and it's completely unnecessary (or even dangerous) with any solution that @roigcarlo suggested.

Btw why wouldn't just removing the inline and linking to the FluidDynamicsApp work? I guess it's related to KRATOS_API, but then no app would be able to consume another app in C++.

philbucher (changes requested):
I just checked again, the same (am I missing sth?) is working for the cablenetapp and the strucutral app

https://github.com/KratosMultiphysics/Kratos/blob/00c495df3e89a979543c7c312c93fa676eaf6c0a/applications/CableNetApplication/custom_elements/empirical_spring.cpp#L110-L113

roigcarlo:
@philbucher those are not static, are normal functions in a namespace (something that we do not recommend btw)

@matekelemen it has nothing to do with the kratos api. It is true that the inline can be removed, but the problem is that the hydraulic app cannot inherit the definition, only the declaration.

It would work with static without inline if there were variables, but with the static function is not possible to make de definition visible unless is in the header ( or its defined twice by using the inline)

philbucher:
> @philbucher those are not static, are normal functions in a namespace (something that we do not recommend btw)

I know that we discussed this a while ago, but here it seems that class+static functions is the technically inferior solution?

Mostly trying to understand the technical solutions that are possible, maybe time to reconsider free functions :thinking:

philbucher (approved):
sorry @uxuech that you get caught up in this, my comments are not related to your PR

PR #11948, opened 2024-01-12
[GeoMechanicsApplication] Added a .git-blame-ignore-revs file to avoid seeing the auto-format commits in our history

rfaasse:
To use this, you need to configure this once on your machine by running the following command in the root of your repo and bring all your branches up to date with master (such that the .git-blame-ignore-revs file is there).
git config --global blame.ignoreRevsFile applications\GeoMechanicsApplication\.git-blame-ignore-revs


avdg81 (approved):
This will help us in ignoring commits that are irrelevant for "forensic analysis". Thanks for investigating how this works and preparing the file.

PR #11953, opened 2024-01-15
[GeoMechanicsApplication] Adapt linear elastic laws

### LINKED ISSUE #11903: [GeoMechanicsApplication] Compare, and adapt, the Kratos K₀ Procedure with Plaxis
It will be expected from typical users that the K<sub>0</sub> procedure will be comparable with that developed in Plaxis. From our initial investigations it appears that the K<sub>0</sub> procedure is straight forward and correctly implemented. However the development of the initial stresses differ. 

In our version the initial vertical stress field is developed using any material model that the user wishes to apply, while the procedure in Plaxis appears to use self-weight. 

Initial tests show that the use of the linear-elastic model in Kratos is not comparable. This issue will investigate and conclude

**Acceptance Criteria**
**Given** the developed K<sub>0</sub> procedure
**When** used in all circumstances
**Then** all results of the stage should be comparable to those in Plaxis

rfaasse:
**📝 Description**
Added the option to force compatibility with Plaxis's behavior when it comes to the K0 procedure. When compatibility is requested, all coupling terms in the elasticity matrix are set to zero. When the K0 procedure ends, the coupling terms are restored.


applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_law.cpp
```diff
@@ -0,0 +1,202 @@
...
+{
+    // Since `ConstitutiveLaw::SetValue` always throws an exception, an override
+    // is required when that is undesired behavior
+}
+
+void GeoLinearElasticLaw::SetValue(const Variable<Vector>&, const Vector&, const ProcessInfo&)
+{
+    // Since `ConstitutiveLaw::SetValue` always throws an exception, an override
+    // is required when that is undesired behavior
+}
```
rfaasse, line 157: These functions were not there in e.g. the k0 3d law, why are they added here?

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_law.h
```diff
@@ -0,0 +1,75 @@
...
+    virtual void CalculateCauchyGreenStrain(ConstitutiveLaw::Parameters& rValues,
+                                            Vector& rStrainVector) = 0;
+    virtual void CalculatePK2Stress(const Vector& rStrainVector,
+                                    Vector& rStressVector,
+                                    ConstitutiveLaw::Parameters& rValues) = 0;
+
+private:
+    friend class Serializer;
+    void save(Serializer& rSerializer) const override;
+    void load(Serializer& rSerializer) override;
```
rfaasse, line 68: We are missing the save/load for the member variable.

applications/GeoMechanicsApplication/tests/cpp_tests/test_apply_k0_procedure_process.cpp
```diff
@@ -0,0 +1,105 @@
...
+{
+    Model model;
+    auto& r_model_part = PrepareTestModelPart(model);
+
+    Parameters k0_settings{R"({"force_plaxis_compatibility": true})"};
+
+    ApplyK0ProcedureProcess process{r_model_part, k0_settings};
+    process.ExecuteInitialize();
+
+    KRATOS_EXPECT_TRUE(AllElementsHaveSameCouplingOption(r_model_part.Elements(), GeoLinearElasticLaw::Coupling::No))
```
rfaasse: Maybe we can add the finalize of the process here as well and check that coupling is set to yes again

applications/GeoMechanicsApplication/tests/test_k0_procedure_process.py
```diff
@@ -8,6 +8,24 @@ class KratosGeoMechanicsK0ProcedureProcessTests(KratosUnittest.TestCase):
     """
     This class contains tests which check the result of a K0 Procedure Process
     """
+    def assert_stresses_at_integration_point(self, cauchy_stress_tensors, integration_point, expected_horizontal_stress, expected_vertical_stress, rel_tol):
+        """
+        Verifies whether the computed stresses are (nearly) equal to some expected values at the
+        given integration point.  Note that this function assumes there are no shear stresses!
```
rfaasse, line 14: This might be a bit more clear (same with the other tests)
```suggestion
        given integration point.  Note that this test assumes there are no shear stresses!
```

rfaasse (commented):
Looks good, just a few comments from my side.

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_law.cpp
avdg81: I think they were: see lines 216 to 222 of file `applications/GeoMechanicsApplication/custom_constitutive/elastic_isotropic_K0_3d_law.h`. Note that the (empty) function definitions were in the header file, not in the implementation file. Or am I overlooking something?

rfaasse: Nope, I overlooked them, my bad (missed them in the cpp and didn't check for an implementation in the header)

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_law.h
avdg81: After our discussion, we've concluded that there is no need to serialize data member `mIsCouplingWanted`. The reason is this data member is supposed to be changed only during the execution of a stage that includes the K0 procedure. The serialization, however, aims at saving/loading the model state at the begin (or end) of a stage, but not during a stage run. When the need arises to serialize `mIsCouplingWanted`, we will add it.

applications/GeoMechanicsApplication/tests/cpp_tests/test_apply_k0_procedure_process.cpp
avdg81: Good suggestion! I have added a separate unit test to cover that behavior.

applications/GeoMechanicsApplication/tests/test_k0_procedure_process.py
avdg81: What I tried to express here is that function `assert_stresses_at_integration_point` is intended to check the stress state after applying the K0 procedure. In other words, it cannot be used to check _any_ stress state, but only the ones produced by the K0 procedure (which, by definition, will yield shear stresses that are equal to zero). From that point of view, it is not the test which assumes there are no shear stresses, but this check function. The reason for this comment is to avoid that this check function will be accidentally reused to verify stress states that were not produced by the K0 procedure. Therefore, I think we should not replace "function" by "test". We could, however, replace "function" by "checker" or "assertion". Would that be clearer in your opinion?

rfaasse: Okay clear, in that case indeed the current comment is as it should be! I misread and thought this was a test as well, but indeed I now see this is a helper function

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_law.cpp
```diff
@@ -0,0 +1,202 @@
...
+// NOTE: Since we are in the hypothesis of small strains we can use the same function for everything
+void GeoLinearElasticLaw::CalculateMaterialResponsePK1(ConstitutiveLaw::Parameters& rValues)
+{
+    CalculateMaterialResponsePK2(rValues);
+}
+
+void GeoLinearElasticLaw::CalculateMaterialResponsePK2(ConstitutiveLaw::Parameters& rValues)
+{
+    KRATOS_TRY
+    // b.- Get Values to compute the constitutive law:
```
WPK4FEM: What is the intention of b.- here?

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_law.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+    if (r_options.Is(ConstitutiveLaw::COMPUTE_CONSTITUTIVE_TENSOR)) {
+        Matrix& r_constitutive_matrix = rValues.GetConstitutiveMatrix();
+        CalculateElasticMatrix(r_constitutive_matrix, rValues);
+    }
+
+    if (r_options.Is(ConstitutiveLaw::COMPUTE_STRESS)) {
+        Vector& r_stress_vector = rValues.GetStressVector();
+        CalculatePK2Stress(r_strain_vector, r_stress_vector, rValues);
+    }
```
WPK4FEM, line 70: rValues is the output argument here. That raises the expectation that the if conditions should be mutually exclusive, otherwise rValues contains the last computed content.

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_law.cpp
```diff
@@ -0,0 +1,202 @@
...
+        << "YOUNG_MODULUS is not available in material parameters" << std::endl;
+    KRATOS_ERROR_IF(rMaterialProperties[YOUNG_MODULUS] <= 0.0)
+        << "YOUNG_MODULUS is invalid value " << std::endl;
+
+    KRATOS_ERROR_IF_NOT(rMaterialProperties.Has(POISSON_RATIO))
+        << "POISSON_RATIO is not available in material parameters" << std::endl;
+
+    const double& nu = rMaterialProperties[POISSON_RATIO];
+    KRATOS_ERROR_IF((nu > 0.499 && nu < 0.501) || (nu < -0.999 && nu > -1.01))
+        << "POISSON_RATIO has invalid value " << std::endl;
```
WPK4FEM, line 173: strange range, I would expect -1.0 <= nu < 0.5 ( give or take equal signs )

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.cpp
```diff
@@ -22,18 +22,51 @@
...
+using namespace Kratos;
+
+void SetCouplingOption(ModelPart::ElementsContainerType& rElements, GeoLinearElasticLaw::IsCouplingWanted WantCoupling)
+{
+    block_for_each(rElements, [WantCoupling](Element& rElement) {
+        auto pLinearElasticLaw =
+            dynamic_cast<GeoLinearElasticLaw*>(rElement.GetProperties().GetValue(CONSTITUTIVE_LAW).get());
+        if (pLinearElasticLaw) {
+            pLinearElasticLaw->SetCouplingOption(WantCoupling);
+        }
```
WPK4FEM: What does the defined clang format say here? I expected a 1 line if statement, but maybe it exceeds 100 characters.

WPK4FEM (commented):
Nice work including some unraveling of the class structure. 
Good that it mentions coupling. The opposite of it is decoupling. I wonder if we should avoid explicit mentioning of Plaxis throughout everything here.

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_law.cpp
avdg81: I've moved several functions verbatim from one file to this one. To not complicate the code review (i.e. to make clear this code was only moved), I was reluctant to make any other changes. (Those can be done in a separate PR.) For improving comments and addressing code smells, however, we can make an exception and fix them right away. So I will do that here. Thanks for pointing this one out.

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_law.cpp
avdg81: Yes, I agree with your suggestion to make them mutually exclusive. However, it turned out that we have misunderstood this part of the code. Basically, it represents the three steps to calculate the stress state at an integration point: (1) calculate the Green strain from the given strain measure (whatever that is), (2) calculate the constitutive matrix, and (3) calculate the stress tensor from the Green strain and the constitutive matrix. Each of these steps appears to be optional, which implies that some cached value is used when a calculation is omitted. This code can (and probably should) be made clearer, but that goes beyond the scope of this PR.

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_law.cpp
avdg81: Good catch! Definitely something that needs to be addressed, but not in this PR. I will add an issue to our product backlog to make sure we don't forget about it.

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.cpp
avdg81: Done. (It was not done automatically by clang-format; but after modifying it manually, it was also not reverted by clang-format.)

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_law.cpp
```diff
@@ -0,0 +1,202 @@
...
+        Vector& r_stress_vector = rValues.GetStressVector();
+        CalculatePK2Stress(r_strain_vector, r_stress_vector, rValues);
+    }
+
+    KRATOS_CATCH("")
+}
+
+void GeoLinearElasticLaw::CalculateMaterialResponseKirchhoff(ConstitutiveLaw::Parameters& rValues)
+{
+    CalculateMaterialResponsePK2(rValues);
```
markelov208, line 77: A general question, most likely for a separate issue, here CalculateMaterialResponseKirchhoff calls CalculateMaterialResponsePK2. Below CalculateMaterialResponseCauchy calls CalculateMaterialResponsePK2. When SmallStrainUDSM3DLaw::CalculateMaterialResponsePK2 calls CalculateMaterialResponseCauchy. It looks that in both cases function names do not reflect the numerics and here CalculateMaterialResponsePK2 should be split in appropriate functions.

markelov208 (commented):
Anne, sorry only one general question.

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_law.cpp
avdg81: Thanks for pointing out this inconsistency, Gennady! I agree with you that this can be improved. But, as you indicated yourself, this is beyond the scope of this PR. I have created an issue for it on our product backlog.

applications/GeoMechanicsApplication/tests/cpp_tests/test_apply_k0_procedure_process.cpp
```diff
@@ -47,75 +50,72 @@ ModelPart& PrepareTestModelPart(Model& rModel)
...
 
-    Parameters k0_settings{R"({"force_decoupled_elasticity": false})"};
+    Parameters k0_settings{R"({"use_standard_procedure": true})"};
 
     ApplyK0ProcedureProcess process{r_model_part, k0_settings};
     process.ExecuteInitialize();
 
-    KRATOS_EXPECT_TRUE(AllElementsHaveSameCouplingOption(
-        r_model_part.Elements(), GeoLinearElasticLaw::IsCouplingWanted::Yes))
+    KRATOS_EXPECT_TRUE(boost::algorithm::all_of(r_model_part.Elements(), ElementConsidersDiagonalEntriesOnlyAndNoShear))
```
rfaasse, line 90: Is there a specific reason we move to the boost version instead of the stl one with iterators? I like the readability, just wondering about stl vs 'third party' (although boost is of course a special kind of 3rd party in my opinion)

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.cpp
```diff
@@ -52,15 +52,15 @@ ApplyK0ProcedureProcess::ApplyK0ProcedureProcess(ModelPart& model_part, const Pa
 
 void ApplyK0ProcedureProcess::ExecuteInitialize()
 {
-    if (ForceDecoupledElasticity()) {
-        SetCouplingOption(mrModelPart.Elements(), GeoLinearElasticLaw::IsCouplingWanted::No);
+    if (UseStandardProcedure()) {
+        SetConsiderDiagonalEntriesOnlyAndNoShear(mrModelPart.Elements(), true);
```
rfaasse: I think this reads very clearly now

applications/GeoMechanicsApplication/tests/test_k0_procedure_process/README.md
```diff
@@ -1,12 +1,13 @@
...
+-   [K<sub>0</sub> procedure normal consolidation with skew layers and dam](https://github.com/KratosMultiphysics/Kratos/tree/master/applications/GeoMechanicsApplication/tests/test_k0_procedure_process/test_k0_procedure_k0_nc_skew_layers_dam/README.md)
+-   [K<sub>0</sub> procedure with umat](https://github.com/KratosMultiphysics/Kratos/tree/master/applications/GeoMechanicsApplication/tests/test_k0_procedure_process/test_k0_procedure_k0_umat/README.md)
+-   [K<sub>0</sub> procedure for simple dike](https://github.com/KratosMultiphysics/Kratos/tree/master/applications/GeoMechanicsApplication/tests/test_k0_procedure_process/test_k0_procedure_simple_dike/README.md)
+-   [K<sub>0</sub> procedure for horizontal layers](https://github.com/KratosMultiphysics/Kratos/tree/master/applications/GeoMechanicsApplication/tests/test_k0_procedure_process/test_k0_procedure_with_horizontal_layers/README.md)
+-   [K<sub>0</sub> procedure for tilted layers](https://github.com/KratosMultiphysics/Kratos/tree/master/applications/GeoMechanicsApplication/tests/test_k0_procedure_process/test_k0_procedure_with_tilted_layers/README.md)
```
rfaasse, line 13: Shouldn't we make changes to all the k0 tests which were there before to explicitly set it the 'use-standard-procedure' to false? Or do all tests pass now?

rfaasse (commented):
I think the changes look clear, a few final questions.

Thanks for the work and rework on this PR!

applications/GeoMechanicsApplication/tests/cpp_tests/test_apply_k0_procedure_process.cpp
avdg81: The primary reason to use `boost::algorithm::all_of` is it provides an overload that accepts a range (which I feel is easier to read than a pair of iterators; also, you are very explicit that you want to visit all elements in the container, and you can't accidentally pass a mismatching pair of iterators). As of C++20, ranges are available in the standard. Since we are still at C++17, this is a nice alternative. And since it doesn't pull in a new dependency, I feel this is acceptable. We may consider to use it in other places, too.

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.cpp
avdg81: Thank you :-)

applications/GeoMechanicsApplication/tests/test_k0_procedure_process/README.md
avdg81: Since all tests pass, I decided to not add the "use_standard_procedure" flag to the existing test files. But I will double-check with @WPK4FEM to make sure that I did not make any mistake here.

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_law.cpp
WPK4FEM: We checked together and these are not mutually exclusive.

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.cpp
```diff
@@ -22,16 +22,46 @@
...
 #include "geo_mechanics_application_variables.h"
 #include "includes/kratos_flags.h"
-#include "includes/kratos_parameters.h"
+
+namespace
+{
+
+using namespace Kratos;
+
+void SetConsiderDiagonalEntriesOnlyAndNoShear(ModelPart::ElementsContainerType& rElements, bool Whether)
```
WPK4FEM, line 34: I think the AndNoShear referred to the last diagonal entries. Those are now left intact, thus making this a less descriptive name.

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.cpp
```diff
@@ -22,16 +22,46 @@
...
+void ApplyK0ProcedureProcess::ExecuteInitialize()
 {
+    if (UseStandardProcedure()) {
+        SetConsiderDiagonalEntriesOnlyAndNoShear(mrModelPart.Elements(), true);
+    }
+}
+
+void ApplyK0ProcedureProcess::ExecuteFinalize()
+{
+    if (UseStandardProcedure()) {
```
WPK4FEM: Does clang format not take out the { } for this 1 line like statement?

applications/GeoMechanicsApplication/tests/cpp_tests/test_apply_k0_procedure_process.cpp
```diff
@@ -0,0 +1,121 @@
...
+KRATOS_TEST_CASE_IN_SUITE(UseStandardProcedureFlagIsInEffectDuringProcessExecutionOnly, KratosGeoMechanicsFastSuite)
+{
+    Model model;
+    auto& r_model_part = PrepareTestModelPart(model);
+
+    Parameters k0_settings{R"({"use_standard_procedure": true})"};
+
+    ApplyK0ProcedureProcess process{r_model_part, k0_settings};
+    process.ExecuteInitialize(); // start considering diagonal entries only and no shear
+    process.ExecuteFinalize(); // stop considering diagonal entries only and no shear
```
WPK4FEM, line 116: Remaining "and no shear" remark, I think this should be removed.

applications/GeoMechanicsApplication/tests/test_k0_procedure_process/test_k0_procedure_with_tilted_layers/ProjectParameters.json
```diff
@@ -0,0 +1,246 @@
...
+                        "nodal_results":       ["TOTAL_DISPLACEMENT","WATER_PRESSURE"],
+                        "gauss_point_results": ["GREEN_LAGRANGE_STRAIN_TENSOR","CAUCHY_STRESS_TENSOR","TOTAL_STRESS_TENSOR"]
+                    },
+                    "point_data_configuration":  []
+                }
+            }
+        }]
+    },
+    "processes": {
+        "auxilliary_process_list": [
```
WPK4FEM, line 42: I still am confused. Apparently both auxilliary_process_list and auxilliar_process_list both work, but I do not understand how and why.

WPK4FEM (dismissed):

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.cpp
WPK4FEM: Mistake of Wijtze Pieter, checked again in the code and the G term on the (XY, XY) diagonal really becomes 0. Ignore the remark.

applications/GeoMechanicsApplication/tests/cpp_tests/test_apply_k0_procedure_process.cpp
WPK4FEM: See previous remark. The (XY,XY) diagonal term really becomes 0.

applications/GeoMechanicsApplication/tests/test_k0_procedure_process/test_k0_procedure_with_tilted_layers/ProjectParameters.json
WPK4FEM: This is out of scope here, we have made a ticket in the product backlog.

applications/GeoMechanicsApplication/tests/test_k0_procedure_process/README.md
WPK4FEM: Anne has checked today.

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_law.cpp
WPK4FEM: Out of scope here, there now is an issue on the product backlog.

applications/GeoMechanicsApplication/custom_constitutive/linear_elastic_law.cpp
WPK4FEM: done

applications/GeoMechanicsApplication/custom_processes/apply_k0_procedure_process.cpp
avdg81: No, it does not remove the curly braces automatically. But it also does not put them back when I take them out manually. And so I did.

WPK4FEM (approved):
Done.

PR #11955, opened 2024-01-15
[TrilinosApplication] Adding new test for B&S

loumalouomega:
**📝 Description**

I was looking for a bug recreating a failing case. Funny thing is that I cannot reproduce the error, but it help me to find something new to debug. In any case, I keep the test for ... more testing. This test depends in the `StructuralMechanicsApplication` and it is skipped if not found.

**🆕 Changelog**

- [Adding new test](https://github.com/KratosMultiphysics/Kratos/commit/1b6efb7979bf0443363157f88e65c94ac3e32fa3)

### DIFF
--- a/applications/TrilinosApplication/tests/cpp_tests/strategies/builder_and_solvers/test_builder_and_solver.cpp
+++ b/applications/TrilinosApplication/tests/cpp_tests/strategies/builder_and_solvers/test_builder_and_solver.cpp
@@ -611,10 +611,207 @@ namespace Kratos::Testing
                 auto pnode2 = rModelPart.pGetNode(2);
                 rModelPart.CreateNewMasterSlaveConstraint("LinearMasterSlaveConstraint", 1, *pnode1, DISPLACEMENT_Y, *pnode2, DISPLACEMENT_Y, 1.0, 0.0);
             }
-
-            // Compute communication plan and fill communicator meshes correctly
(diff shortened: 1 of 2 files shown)


applications/TrilinosApplication/tests/cpp_tests/strategies/builder_and_solvers/test_builder_and_solver.cpp
```diff
@@ -611,10 +611,207 @@ namespace Kratos::Testing
...
-            // Compute communication plan and fill communicator meshes correctly
-            ParallelFillCommunicator(rModelPart, rDataCommunicator).Execute();
+        // Compute communication plan and fill communicator meshes correctly
+        ParallelFillCommunicator(rModelPart, rDataCommunicator).Execute();
+    }
+
+    /**
+    * @brief It generates a truss structure with an expected solution
+    */
+    static inline bool StructuralTestBuilderAndSolverDisplacement(
```
roigcarlo, line 623: I think it does not need the `inline` qualifier as it is only used in this source file.

roigcarlo (approved):

applications/TrilinosApplication/tests/cpp_tests/strategies/builder_and_solvers/test_builder_and_solver.cpp
loumalouomega: Good point.

PR #11962, opened 2024-01-17
[Expression] Add Entity Domain Size Reading

sunethwarna:
**📝 Description**
This PR adds read mechanism to fill a given `ConditionExpression` or `EntityExpression` with their corresponding domain sizes.

**🆕 Changelog**
- Adds `DomainSizeExpressionIO`
- Expose to python
- Adds tests


kratos/expression/entity_domain_size_expression_io.h
```diff
@@ -0,0 +1,89 @@
...
+#include "expression/container_expression.h"
+#include "expression/traits.h"
+#include "expression_io.h"
+#include "includes/define.h"
+#include "includes/model_part.h"
+
+namespace Kratos {
+
+
+class KRATOS_API(KRATOS_CORE) EntityDomainSizeExpressionIO
```
matekelemen: I think `DomainSizeExpressionIO` (without the `Entity`) is enough.

Also, docs are missing and this one definitely needs some explanation. `DomainSize` is super confusing in Kratos, so add some doxygen annotations describing what this IO class actually fetches.

kratos/expression/entity_domain_size_expression_io.h
```diff
@@ -0,0 +1,89 @@
...
+{
+public:
+    ///@name Type definitions
+    ///@{
+
+    using Configuration = Kratos::Globals::Configuration;
```
matekelemen: is `Configuration` used anywhere in this class?

kratos/expression/entity_domain_size_expression_io.h
```diff
@@ -0,0 +1,89 @@
...
+
+        ///@}
+        ///@name Life cycle
+        ///@{
+
+        EntityDomainSizeExpressionInput(
+            const ModelPart& rModelPart,
+            const ContainerType& rContainerType);
+
+        ~EntityDomainSizeExpressionInput() override = default;
```
matekelemen: you can safely remove this if you're `default`ing it anyway.

kratos/expression/entity_domain_size_expression_io.h
```diff
@@ -0,0 +1,89 @@
...
+
+        Expression::Pointer Execute() const override;
+
+        ///@}
+
+    private:
+        ///@name Private member variables
+        ///@{
+
+        const ModelPart& mrModelPart;
```
matekelemen: this class won't be copy or move assignable if you store a reference. Please consider storing a pointer instead.

roigcarlo (commented):
Sure I am missing something but, why do you need to nested classes?

Isn't it easier to just move the `EntityDomainSizeExpressionIO::Read` function as an auxiliary of the python_interface layer (including the instantiations), remove its static qualifier, avoid the export and leave `EntityDomainSizeExpressionInput` as a  regular class in the core?

kratos/expression/entity_domain_size_expression_io.h
```diff
@@ -0,0 +1,89 @@
...
+
+        KRATOS_CLASS_POINTER_DEFINITION(EntityDomainSizeExpressionInput);
+
+        ///@}
+        ///@name Life cycle
+        ///@{
+
+        EntityDomainSizeExpressionInput(
+            const ModelPart& rModelPart,
+            const ContainerType& rContainerType);
```
matekelemen:
```suggestion
        EntityDomainSizeExpressionInput(
            const ModelPart& rModelPart,
            ContainerType ContainerType);
```

kratos/expression/entity_domain_size_expression_io.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+namespace Kratos {
+
+EntityDomainSizeExpressionIO::EntityDomainSizeExpressionInput::EntityDomainSizeExpressionInput(
+    const ModelPart& rModelPart,
+    const ContainerType& rContainerType)
+    : mrModelPart(rModelPart),
+      mContainerType(rContainerType)
+{
+}
```
matekelemen:
```suggestion
EntityDomainSizeExpressionIO::EntityDomainSizeExpressionInput::EntityDomainSizeExpressionInput(
    const ModelPart& rModelPart,
    Globals::DataLocation rContainerType)
    : mrModelPart(rModelPart),
      mContainerType(ContainerType)
{
}
```

kratos/expression/entity_domain_size_expression_io.h
```diff
@@ -0,0 +1,89 @@
...
+
+        ///@}
+
+    private:
+        ///@name Private member variables
+        ///@{
+
+        const ModelPart& mrModelPart;
+
+        const ContainerType mContainerType;
```
matekelemen: I'd rather stick to [`Globals::DataLocation`](https://github.com/KratosMultiphysics/Kratos/blob/fbc41845e9059b8457c9fa44dd0d2457a1c48931/kratos/includes/global_variables.h#L47-L55) since it's already in core, and get rid of `ContainerType`.

Also, the `const` qualifier prevents assignment operators again.
```suggestion
        Globals::DataLocation mContainerType;
```

kratos/expression/entity_domain_size_expression_io.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+Expression::Pointer EntityDomainSizeExpressionIO::EntityDomainSizeExpressionInput::Execute() const
+{
+    const auto& r_mesh = mrModelPart.GetCommunicator().LocalMesh();
+
+    LiteralFlatExpression<double>::Pointer expression;
```
matekelemen:
```suggestion
    LiteralFlatExpression<double>::Pointer p_expression;
```

kratos/expression/entity_domain_size_expression_io.cpp
```diff
@@ -0,0 +1,84 @@
...
+            expression = LiteralFlatExpression<double>::Create(r_mesh.NumberOfElements(), {});
+            IndexPartition<IndexType>(r_mesh.NumberOfElements()).for_each([&r_mesh, &expression](const IndexType Index) {
+                const auto& r_element = *(r_mesh.ElementsBegin() + Index);
+                *(expression->begin() + Index) = r_element.GetGeometry().DomainSize();
+            });
+            break;
+        default:
+            KRATOS_ERROR << "EntityDomainSizeExpressionInput only supports ConditionNonHistorical and ElementNonHistorical container types.";
+            break;
+    }
```
matekelemen: seems like a job for [proxies](https://github.com/KratosMultiphysics/Kratos/blob/fbc41845e9059b8457c9fa44dd0d2457a1c48931/kratos/utilities/proxies.h)

```cpp
template <Globals::DataLocation TContainerType>
Expression::Pointer ReadDomainSizes(const ModelPart& rModelPart)
{
    LiteralFlatExpression<double>::Pointer p_expression;

    auto container = MakeProxy<TContainerType>(rModelPart);
    IndexPartition<IndexType>(container.size()).for_each([&p_expression, &container] (IndexType iEntity) {
        const auto& r_entity = container[i_entity].GetEntity();
        *(p_expression->begin() + iEntity) = r_entity.DomainSize();
    });

    return p_expression;
}

auto p_expression = ReadDomainSizes<Globals::DataLocation::Element>(rModelPart); // <== gets the domain sizes of elements
auto p_expression = ReadDomainSizes<Globals::DataLocation::Condition>(rModelPart); // <== gets the domain sizes of conditions
```

kratos/expression/entity_domain_size_expression_io.cpp
```diff
@@ -0,0 +1,84 @@
...
+{
+    if constexpr(std::is_same_v<TContainerType, ModelPart::ConditionsContainerType>) {
+        auto p_expression = EntityDomainSizeExpressionInput(rContainerExpression.GetModelPart(), ContainerType::ConditionNonHistorical).Execute();
+        rContainerExpression.SetExpression(p_expression);
+    } else if constexpr(std::is_same_v<TContainerType, ModelPart::ElementsContainerType>) {
+        auto p_expression = EntityDomainSizeExpressionInput(rContainerExpression.GetModelPart(), ContainerType::ElementNonHistorical).Execute();
+        rContainerExpression.SetExpression(p_expression);
+    } else {
+        static_assert(std::is_same_v<TContainerType, TContainerType>, "Unsupported TContainerType");
+    }
```
matekelemen: same, use proxies for cleaner code

kratos/tests/test_container_expression.py
```diff
@@ -758,6 +758,20 @@ def _Read(self, container_expression, variable):
...
+        numpy_condition_exp = condition_exp.Evaluate()
+        for i, condition in enumerate(self.model_part.Conditions):
+            self.assertAlmostEqual(numpy_condition_exp[i], condition.GetGeometry().DomainSize(), 12)
+
+    def testEntityDomainSizeExpressionIOElement(self):
+        element_exp = Kratos.Expression.ElementExpression(self.model_part)
+        Kratos.Expression.EntityDomainSizeExpressionIO.Read(element_exp)
+        numpy_element_exp = element_exp.Evaluate()
+        for i, condition in enumerate(self.model_part.Elements):
+            self.assertAlmostEqual(numpy_element_exp[i], condition.GetGeometry().DomainSize(), 12)
```
matekelemen: please make the elements'/conditions' doman sizes different, otherwise the tests will pass even if the IO gets the wrong entity. Also, add tests for empty containers.

matekelemen: and in line 772, you're looping over elements, not conditions

kratos/expression/entity_domain_size_expression_io.h
sunethwarna: done :)

kratos/expression/entity_domain_size_expression_io.h
sunethwarna: nice catch. Removed it :)

kratos/expression/entity_domain_size_expression_io.h
sunethwarna: done :)

kratos/expression/entity_domain_size_expression_io.h
sunethwarna: done :)

kratos/expression/entity_domain_size_expression_io.h
sunethwarna: done :)

kratos/expression/entity_domain_size_expression_io.cpp
sunethwarna: This won't work since, I need to iterate over only the local entities. This  passes through all the local and ghost entities.

kratos/expression/entity_domain_size_expression_io.cpp
sunethwarna: As mentioned in the above.

kratos/expression/entity_domain_size_expression_io.h
sunethwarna: done and removed.

sunethwarna:
@roigcarlo I agree, but, we would like to have the same interfaces in C++ to be able to them in templated methods. The use cases will be templated from the IO.

@matekelemen I have addressed your comments. Could you have a look?

kratos/tests/test_container_expression.py
sunethwarna: I am iterating in different containers here. I have a naming problem which is fixed.
Added empty mp tests.

kratos/tests/test_container_expression.py
```diff
@@ -852,6 +852,34 @@ def _Read(self, container_expression, variable):
     def _Evaluate(self, container_expression, variable):
         Kratos.Expression.VariableExpressionIO.Write(container_expression, variable)
 
+    def testDomainSizeExpressionIOCondition(self):
+        condition_exp = Kratos.Expression.ConditionExpression(self.model_part)
+        Kratos.Expression.DomainSizeExpressionIO.Read(condition_exp)
+        numpy_condition_exp = condition_exp.Evaluate()
+        for i, condition in enumerate(self.model_part.Conditions):
+            self.assertAlmostEqual(numpy_condition_exp[i], condition.GetGeometry().DomainSize(), 12)
```
matekelemen: just curious: why do you need the `12` tolerance? This expression shouldn't do any arithmetic on the domain sizes it's getting from elements/conditions, so I'd expect it to be exactly identical to `condition.GetGeometry().DomainSize()`.

kratos/expression/domain_size_expression_io.h
```diff
@@ -0,0 +1,96 @@
...
+        ///@}
+
+    };
+
+    ///@}
+    ///@name Public static operations
+    ///@{
+
+    template<class TContainerType, MeshType TMeshType>
+    KRATOS_API(KRATOS_CORE) static void Read(ContainerExpression<TContainerType, TMeshType>& rContainerExpression);
```
matekelemen, line 90:
```suggestion
    /// @tparam TContainerType @ref ModelPart::ElementsContainerType or @ref ModelPart::ConditionsContainerType
    template<class TContainerType, MeshType TMeshType>
    KRATOS_API(KRATOS_CORE) static void Read(ContainerExpression<TContainerType, TMeshType>& rContainerExpression);
```

kratos/expression/domain_size_expression_io.cpp
```diff
@@ -0,0 +1,83 @@
...
+void DomainSizeExpressionIO::Read(ContainerExpression<TContainerType, TMeshType>& rContainerExpression)
+{
+    if constexpr(std::is_same_v<TContainerType, ModelPart::ConditionsContainerType>) {
+        auto p_expression = Input(rContainerExpression.GetModelPart(), Globals::DataLocation::Condition).Execute();
+        rContainerExpression.SetExpression(p_expression);
+    } else if constexpr(std::is_same_v<TContainerType, ModelPart::ElementsContainerType>) {
+        auto p_expression = Input(rContainerExpression.GetModelPart(), Globals::DataLocation::Element).Execute();
+        rContainerExpression.SetExpression(p_expression);
+    } else {
+        static_assert(std::is_same_v<TContainerType, TContainerType>, "Unsupported TContainerType");
```
matekelemen:
```suggestion
        static_assert(!std::is_same_v<TContainerType, TContainerType>, "Unsupported TContainerType");
```

sunethwarna: again a stupid mistake :/

kratos/tests/test_container_expression.py
sunethwarna: The problem is, how the domain size is computed, it has a square root, and squaring which may not produce the same exact values. :/

matekelemen: what? But the expression doesn't compute a square or square root, right? That's done internally in the element/condition, am I correct?

If that's the case, the condition/element should **always** return the same number. Floating point operations may not be **exact**, but they must absolutely be **consistent**.

sunethwarna: > what? But the expression doesn't compute a square or square root, right? That's done internally in the element/condition, am I correct?

You are correct.

> If that's the case, the condition/element should always return the same number. Floating point operations may not be exact, but they must absolutely be consistent.

I will give it a try. :)

sunethwarna: My only worry is, when we do the numpy conversion, whether we will lose something. But lets see :)

matekelemen (approved):

PR #11964, opened 2024-01-18
[OptApp] Add ConnectivityPreservingModelPartDuplicationController

sunethwarna:
**📝 Description**
This PR adds connectivity preserving model part duplication input for the OptimizationApplication

**🆕 Changelog**
- Add `ConnectivityPreservingModelPartDuplicationController`


Igarizza (commented):
Model part controller can be confused with optimization controllers. The code is fine, but I don't understand what it does and why do you need it=(

applications/OptimizationApplication/python_scripts/model_part_controllers/connectivity_preserving_model_part_duplication_controller.py
```diff
@@ -0,0 +1,39 @@
+import KratosMultiphysics as Kratos
+from KratosMultiphysics.OptimizationApplication.model_part_controllers.model_part_controller import ModelPartController
+
+def Factory(model: Kratos.Model, parameters: Kratos.Parameters, _) -> ModelPartController:
+    if not parameters.Has("settings"):
```
matekelemen, line 5: isn't it supposed to be `Settings` (note the capitalization)? I'm not fond of pointless capitalization either, but we should be consistent with core.

applications/OptimizationApplication/python_scripts/model_part_controllers/connectivity_preserving_model_part_duplication_controller.py
```diff
@@ -0,0 +1,39 @@
...
+        raise RuntimeError(f"ConnectivityPreservingModelPartDuplicationController instantiation requires a \"settings\" in parameters [ parameters = {parameters}].")
+    return ConnectivityPreservingModelPartDuplicationController(model, parameters["settings"])
+
+class ConnectivityPreservingModelPartDuplicationController(ModelPartController):
```
matekelemen, line 9: 0 / 10 naming :laughing: 

how about `ElementFactoryController` or `ConditionFactoryController` (with the necessary changes to make it only work for elements/conditions)? Or at least something **much** more concise that explains what the controller does, without encoding the entire source code in the class' name.

sunethwarna:
Closing because, more robust implementation is done in #12294 .

PR #11966, opened 2024-01-18
[GeoMechanicsApplication] Discussion PR: Extract different strain forms by using a strategy pattern

rfaasse:


rfaasse:
Closing this PR, as it is not the intention to merge this, but only as a discussion starting point

PR #11967, opened 2024-01-18
[OptApp] Fix line search

sunethwarna:
**📝 Description**
Fixes a small bug in the line search.

**🆕 Changelog**
- Fixed a bug in line search

### DIFF
--- a/applications/OptimizationApplication/python_scripts/utilities/opt_line_search.py
+++ b/applications/OptimizationApplication/python_scripts/utilities/opt_line_search.py
@@ -55,7 +55,7 @@ class ConstStep():
     @time_decorator()
     def ComputeStep(self) -> float:
         norm = self.ComputeScaleFactor()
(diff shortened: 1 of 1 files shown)


applications/OptimizationApplication/python_scripts/utilities/opt_line_search.py
```diff
@@ -55,7 +55,7 @@ def ComputeScaleFactor(self) -> float:
     @time_decorator()
     def ComputeStep(self) -> float:
         norm = self.ComputeScaleFactor()
-        if not math.isclose(norm, 0.0, abs_tol=1e-16):
+        if not math.isclose(norm, 0.0, abs_tol=1e-22):
```
philbucher, line 58: are you sure about those limits? This seems waaaay smaller than machine accuracy. Might be the problem is elsewhere :thinking:

sunethwarna:
@philbucher make sense. closing :)

PR #11970, opened 2024-01-19
[Exp] Add GetMaxDepth

sunethwarna:
**📝 Description**
This PR adds `Expression::GetMaxDepth` to compute the maximum depth of the lazy expression tree structure.

**🆕 Changelog**
- Added 'Expression::GetMaxDepth'
- Added tests


kratos/expression/unary_combine_expression.h
```diff
@@ -113,6 +113,16 @@ class UnaryCombineExpression : public Expression {
...
 
+    IndexType GetMaxDepth() const override
+    {
+        return (*std::max_element(
+                        mSourceExpressions.begin(),
+                        mSourceExpressions.end(),
+                        [](const auto& rV1, const auto& rV2) {
+                            return rV1->GetMaxDepth() < rV2->GetMaxDepth();
+                        }))->GetMaxDepth() + 1;
+    }
```
matekelemen: that's a lot of repeated depth calculations. It might be worth to consider adding a depth member variable that gets computed at construction time (provided that child `Expression`s are immutable), instead of always recomputing every depth for each child over and over again.

kratos/tests/test_container_expression.py
```diff
@@ -665,6 +665,13 @@ def test_VectorWrite(self):
             velocity = Kratos.Array3([vector[i * 3], vector[i * 3 + 1], vector[i * 3 + 2]])
             self.assertVectorAlmostEqual(velocity, self._GetValue(entity, Kratos.VELOCITY))
 
+    def test_GetMaxDepth(self):
+        a = Kratos.Expression.LiteralExpression.Create(1, 10)
+        b = a + 10
+        c = b * 2
+        d = c ** 2
+        self.assertEqual(d.GetMaxDepth(), 4)
```
matekelemen: please add tests for every expression type, and also for cases where different branches have different depths

sunethwarna: done :)

kratos/expression/unary_combine_expression.h
sunethwarna: I agree, but I don't see it getting called several times. [At the moment, it will only be called when the future `Collapse` is called, but that happens when we `Collapse` at the end. There is no point in `Collapsing` every time]. I would say, we can move it to a member variable if this gets out of hand.

In any case, I simpliifed the recursions, so there are no redundant recursive calculations.

sunethwarna: I added a warning as well for future reference.

sunethwarna:
@matekelemen I have addressed your comments. Could you have a look?

matekelemen (approved):

PR #11972, opened 2024-01-19
[Core][MPI] Adding `RanksList` and `InitializeRanksList` in `DataCommunicator`

loumalouomega:
**📝 Description**

This PR addresses a limitation encountered in the use of SubDataCommunicators within the KratosMultiphysics project. Initially, the SubDataCommunicators were designed with the primary use of `XxxAll` methods, which did not necessitate the knowledge of specific ranks involved in communication after creation. However, practical applications involving `Send`/`Recv` methods highlighted the inability to identify the specific ranks participating in the SubDataCommunicators. This is particularly challenging when dealing with non-consecutive and non-zero starting rank scenarios.

In contrast to global data communicators, where the `Size` method allows for iteration over all ranks, SubDataCommunicators require a more nuanced approach due to their potential to connect non-consecutive ranks. A workaround involving the `AllGather` method was considered, as implemented in `InitializeRanksList`, but this approach may lead to synchronization issues in certain cases.

To mitigate this challenge, this PR introduces functionality for retrieving a list of ranks involved in a SubDataCommunicator right from its creation. This enhancement improves the usability and flexibility of SubDataCommunicators, especially in complex communication scenarios. While this update is primarily for discussion and the current implementation seeks to circumvent this issue, the added functionality is a significant step forward in managing SubDataCommunicators more effectively.

This PR is intended to spark a discussion on the best ways to handle rank identification in SubDataCommunicators and to provide a potential solution for the issue. In my current implementation I will try to avoid this.

**🆕 Changelog**

- [Adding base methods for `RanksList` and `InitializeRanksList`](https://github.com/KratosMultiphysics/Kratos/commit/833e7e05ccd6e37f26cc30495be56a53cda41781)
- [Adding derived MPI methods for `RanksList` and `InitializeRanksList`. Also initialize at subdatacommunicators creation.](https://github.com/KratosMultiphysics/Kratos/commit/580734e28cad70fd9a4a2b4ee193575e632c96c8)
- [Adding test](https://github.com/KratosMultiphysics/Kratos/commit/ce86e6189729cd622ac49b250bc1c1c243c6c837)

### DIFF
--- a/kratos/mpi/includes/mpi_data_communicator.h
+++ b/kratos/mpi/includes/mpi_data_communicator.h
@@ -210,12 +210,6 @@ class KRATOS_API(KRATOS_MPI_CORE) MPIDataCommunicator: public DataCommunicator
     static MPIDataCommunicator::UniquePointer Create(MPI_Comm MPIComm);
 
     /**
-     * @brief Initializes the rank list.
-     * @details This function is responsible for initializing the rank list. You should call this function before using some ranking-related operations with sub-DataCommunicator.
-     */
-    void InitializeRanksList() const override;
-
-    /**
      * @brief Pause program execution until all threads reach this call.
      * @details This function serves as a wrapper for MPI_Barrier, which ensures synchronization among all threads before proceeding further in the program.
      */
@@ -399,13 +393,24 @@ class KRATOS_API(KRATOS_MPI_CORE) MPIDataCommunicator: public DataCommunicator
     ///@name Member Variables
     ///@{
 
-    MPI_Comm mComm;                      /// The MPI communicator
+    /// The MPI communicator
+    MPI_Comm mComm;
+
     /// The list of ranks involved in this DataCommunicator
     mutable std::vector<int> mRanksList;
 
     ///@}
     ///@name Operations
     ///@{
+
+    /**
+     * @brief Generate a rank list using the current rank and size of the system.
+     * @details This function generates a rank list by first obtaining the current rank and size
+     * of the system, and then using MPI's AllGather function to gather the ranks of all
+     * processes into a vector.
+     * @return A std::vector<int> containing the rank list of all processes.
+     */
+    std::vector<int> GenerateRankList() const;
 
     void CheckMPIErrorCode(const int ierr, const std::string& MPICallName) const;
 
--- a/kratos/mpi/includes/mpi_data_communicator.h
+++ b/kratos/mpi/includes/mpi_data_communicator.h
@@ -209,6 +209,16 @@ class KRATOS_API(KRATOS_MPI_CORE) MPIDataCommunicator: public DataCommunicator
      */
     static MPIDataCommunicator::UniquePointer Create(MPI_Comm MPIComm);
 
+    /**
+     * @brief Initializes the rank list.
+     * @details This function is responsible for initializing the rank list. You should call this function before using some ranking-related operations with sub-DataCommunicator.
+     */
+    void InitializeRanksList() const override;
+
+    /**
+     * @brief Pause program execution until all threads reach this call.
+     * @details This function serves as a wrapper for MPI_Barrier, which ensures synchronization among all threads before proceeding further in the program.
+     */
     void Barrier() const override;
 
     KRATOS_MPI_DATA_COMMUNICATOR_DECLARE_PUBLIC_INTERFACE_FOR_TYPE(char)
@@ -273,6 +283,12 @@ class KRATOS_API(KRATOS_MPI_CORE) MPIDataCommunicator: public DataCommunicator
     ///@}
     ///@name Inquiry
     ///@{
+
+    /**
+     * @brief Get the list of ranks involved in this DataCommunicator.
+     * @return A vector containing the ranks involved in this DataCommunicator.
+     */
+    std::vector<int> RanksList() const override;
 
     /**
      * @brief Get the parallel rank for this DataCommunicator.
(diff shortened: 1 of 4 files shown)


matekelemen:
I don't get what the issue is. Can you please explain it through a small example?

loumalouomega:
> I don't get what the issue is. Can you please explain it through a small example?

If you want to send to all ranks inside a subcommunicator (lets say 1,4,6, non consecutive and not starting at 0), you need to know the list of ranks. You can retrieve the ranks with AllGather command, but this may look communication as some partitions may be doing something else. In order to avoid that you can pre-compute the ranks involved in the subdatacommunicator.

matekelemen:
You mean you want to communicate with the ranks of a subcommunicator from outside of it? Makes sense if that's the case.

How about initializing the list of ranks in the communicator's constructor? It would be a lot safer and more straightforward than leaving it to the user.

kratos/mpi/includes/mpi_data_communicator.h
```diff
@@ -383,7 +399,8 @@ class KRATOS_API(KRATOS_MPI_CORE) MPIDataCommunicator: public DataCommunicator
     ///@name Member Variables
     ///@{
 
-    MPI_Comm mComm;
+    MPI_Comm mComm;                      /// The MPI communicator
+    mutable std::vector<int> mRanksList; /// The list of ranks involved in this DataCommunicator
```
philbucher: why `mutable`?

kratos/mpi/includes/mpi_data_communicator.h
```diff
@@ -209,6 +209,16 @@ class KRATOS_API(KRATOS_MPI_CORE) MPIDataCommunicator: public DataCommunicator
      */
     static MPIDataCommunicator::UniquePointer Create(MPI_Comm MPIComm);
 
+    /**
+     * @brief Initializes the rank list.
+     * @details This function is responsible for initializing the rank list. You should call this function before using some ranking-related operations with sub-DataCommunicator.
+     */
+    void InitializeRanksList() const override;
```
philbucher:
```suggestion
    void InitializeRanksList() override;
```

I dont think `Initialize...` methods can/should be const

philbucher (commented):
basically same comment as @matekelemen

kratos/mpi/includes/mpi_data_communicator.h
matekelemen: also, the doxygen annotations should precede the item they refer to:
```suggestion
    /// The list of ranks involved in this DataCommunicator
    mutable std::vector<int> mRanksList;
```

kratos/mpi/includes/mpi_data_communicator.h
loumalouomega: The thing is that the class is usually always const, so the method shpuld be. Anycase, i agree with @matekelemen and add this to the constructor

kratos/mpi/includes/mpi_data_communicator.h
loumalouomega: Mutable as it is modified when const, but with the idea of @matekelemen can be avoided

kratos/mpi/includes/mpi_data_communicator.h
```diff
@@ -383,12 +393,25 @@ class KRATOS_API(KRATOS_MPI_CORE) MPIDataCommunicator: public DataCommunicator
...
     ///@{
 
+    /**
+     * @brief Generate a rank list using the current rank and size of the system.
+     * @details This function generates a rank list by first obtaining the current rank and size
+     * of the system, and then using MPI's AllGather function to gather the ranks of all
+     * processes into a vector.
+     * @return A std::vector<int> containing the rank list of all processes.
+     */
+    std::vector<int> GenerateRankList() const;
```
philbucher:
```suggestion
    std::vector<int> GenerateRankList();
```

kratos/mpi/includes/mpi_data_communicator.h
```diff
@@ -383,12 +393,25 @@ class KRATOS_API(KRATOS_MPI_CORE) MPIDataCommunicator: public DataCommunicator
     ///@name Member Variables
     ///@{
 
+    /// The MPI communicator
     MPI_Comm mComm;
 
+    /// The list of ranks involved in this DataCommunicator
+    mutable std::vector<int> mRanksList;
```
philbucher:
```suggestion
    std::vector<int> mRanksList;
```

kratos/mpi/includes/mpi_data_communicator.h
loumalouomega: i will try, but I think I already tried and I couldn't use it as ususally the data communicators are const

kratos/mpi/includes/mpi_data_communicator.h
loumalouomega: Idem

kratos/mpi/includes/mpi_data_communicator.h
philbucher: but thats used only in the CTor no?

loumalouomega: True, I have change it

loumalouomega:
I just realised that local rank are always from 0 to world size - 1, so this is not needed

matekelemen:
> I just realised that local rank are always from 0 to world size - 1, so this is not needed

but wait, didn't you want to communicate with the subcomm from the **outside**? I haven't checked it, but the subcomm constructor should have access to its parent, right?

loumalouomega:
> > I just realised that local rank are always from 0 to world size - 1, so this is not needed
> 
> but wait, didn't you want to communicate with the subcomm from the **outside**? I haven't checked it, but the subcomm constructor should have access to its parent, right?

Nope, but we could add it...

philbucher:
afaik in MPI there is no relation between groups of processes (what we call subcomms). Therefore not sure if we need it :thinking:

PR #11979, opened 2024-01-25
[Docs] Fix KRATOS_API with kratos_doxyfile

sunethwarna:
**📝 Description**
Current `kratos_dxyfile` params set will ignore everyfile which has used `KRATOS_API` from the doxygen generated documentation (basically almost all the classes from Kratos is ignored unless they are headers only). This fixes it.

Now you can see generated doxygen for all the classes in Kratos such as `ModelPart` and so on :) 

**🆕 Changelog**
- Fix doxygen kratos_doxy file to be compatible with Kratos_API


matekelemen (approved):
Awesome!

PR #11985, opened 2024-01-25
[Core] Adding `CreateSphereTriangularMesh` in `CppTestsUtilities`

loumalouomega:
**📝 Description**

Adding `CreateSphereTriangularMesh` in `CppTestsUtilities`

**🆕 Changelog**

- [Adding `CreateSphereTriangularMesh` in `CppTestsUtilities`](https://github.com/KratosMultiphysics/Kratos/commit/3f57d083e2c372e90baeebb1a26c275d33a4b55f)

### DIFF
--- a/kratos/python/add_other_utilities_to_python.cpp
+++ b/kratos/python/add_other_utilities_to_python.cpp
@@ -68,6 +68,7 @@
 #include "utilities/particles_utilities.h"
 #include "utilities/string_utilities.h"
 #include "utilities/model_part_operation_utilities.h"
+#include "utilities/cpp_tests_utilities.h"
 
 namespace Kratos::Python {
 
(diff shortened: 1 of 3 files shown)


roigcarlo (approved):
Ok, but I would like that eventually this moves out of the core. This is only test related and should be in the tests folder (not in the core)

loumalouomega:
> Ok, but I would like that eventually this moves out of the core. This is only test related and should be in the tests folder (not in the core)

(These are tests utilities and can be used in any app)

roigcarlo:
Yep yep, its just a remainder for myself, when we finish with the gtests we will have a dedicate module for that. What i did not have into account is the possibility of exposing it to python, I have to thing about that :+1:

PR #11990, opened 2024-01-27
[Docs] Add user level documentation for Expressions

sunethwarna:
**📝 Description**
This PR adds user level documentation for the `Expression`.

**🆕 Changelog**
- Adding user level documentation for `Expression`

FYI @adityabodas


docs/scripts/build_site.sh
```diff
@@ -15,13 +15,16 @@ fi
...
     python3 process_pages.py -t local
     cd scripts
 fi
 
 if [ -d "../_data" ]; then
+    if [ -d ../_site/Page_files/_data ]; then
+        rm -r ../_site/Page_files/_data
+    fi
     mv ../_data ../_site/Page_files/_data
 fi
```
matekelemen, line 30: this script needs some comments explaining what its dependencies and assumptions are and what it's doing ...

it might also be useful to mention the part of the docs which explains how to build and deploy the jekyll docs.

docs/pages/Kratos/Expressions/General/Overview.md
```diff
@@ -0,0 +1,50 @@
...
+tags: [overview.md]
+sidebar: kratos_expressions
+summary: 
+---
+## Introduction
+
+An expression represents a mechanism where a given mathematical expression is evaluated when actual value of that expression in required. One can relate these
+expressions to lazy expressions (expression templates) in [Eigen](https://eigen.tuxfamily.org/dox/TopicLazyEvaluation.html) or in [Boost](https://www.boost.org/doc/libs/1_82_0/doc/html/boost_yap/manual.html). The main difference being, Kratos expressions are lazy expressions which is evaluated at runtime, whereas Eigen anb Boost has lazy expressions which are evaluated at compile-time. Few of the main advantages of these lazy expressions are:
+* May reduce memory foot.
+* May reduce computational cost.
```
matekelemen:
```suggestion
An expression represents a mechanism where a given mathematical expression is evaluated when actual value of that expression in required. One can compare these
expressions to lazy expressions (expression templates) in [Eigen](https://eigen.tuxfamily.org/dox/TopicLazyEvaluation.html) or in [Boost](https://www.boost.org/doc/libs/1_82_0/doc/html/boost_yap/manual.html). The main difference being, Kratos expressions are **dynamic** lazy expressions (evaluated on demand, not automatically) whose type is evaluated at runtime, whereas Eigen and Boost has **static** lazy expressions whose type is known at compile-time. Few of the main advantages of these lazy expressions are:
* May reduce memory footprint.
* May reduce computational cost.
```

docs/pages/Kratos/Expressions/General/Overview.md
```diff
@@ -0,0 +1,50 @@
...
+
+Kratos - Multiphysics has various types of data containers to store data. All of those data containers requires a ```Kratos::Variable``` to access its underlying data. This requires defining the variable before hand, and then using it to store and retrieve data. The expressions allows the user and/or developer to use these data containers with minimal to no variable to store/retrieve data from data containers. All the expressions are enabled with **shared memory** and **distributed memory** parallelisms. All of these expressions are implemented in C++, while being exposed to python.
```
matekelemen:
```suggestion
KratosMultiphysics stores its data in different containers depending on what kind of geometry that data is associated with (node, element, condition, etc.) and what that data represents. Each container requires a ```Kratos::Variable``` to access its underlying data. This requires defining the variable beforehand, and then using it to store and retrieve its value. Expressions allow users/developers to access and manipulate the data within these containers regardless of what `Kratos::Variable` they belong to. All expressions support **shared memory** and **distributed memory** parallelism. Though implemented in C++, expressions are meant to be used mainly in python.
```

docs/pages/Kratos/Expressions/General/Overview.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Supported data containers
+
+Following data containers are supported:
```
matekelemen:
```suggestion
The following data containers are supported:
```

docs/pages/Kratos/Expressions/General/Overview.md
```diff
@@ -0,0 +1,50 @@
...
+* NodalExpression - Nodal data value container (Historical / Non historical).
+* ConditionExpression - Condition data value container.
+* ElementExpression - Element data value container.
```
matekelemen:
```suggestion
* NodalExpression - data related to nodes (with historical or non-historical variables).
* ConditionExpression - data related to conditions.
* ElementExpression - data related to elements.
```

docs/pages/Kratos/Expressions/General/Overview.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Supported data types
+
+The expressions support almost all the variable types in Kratos, and many more data types as well. The data storage in the expressions can be assumed as a flat ```int``` or ```double``` array. Followings are few supported data types:
```
matekelemen: > The data storage in the expressions can be assumed as a flat ```int``` or ```double``` array

that's an implementation detail

```suggestion
Expressions are compatible with most main variable types in Kratos:
```

docs/pages/Kratos/Expressions/General/Expression_Arithmetics.md
```diff
@@ -0,0 +1,56 @@
...
+The ```Expressions``` of type ```NodalExpression```, ```ConditionExpression```, ```ElementExpression``` and their underlying ```Expression``` can be used in arithmetic formulations.
+
+Following operations are supported:
+* Addition ```+``` or ```+=```
+* Substraction ```-``` or ```-=```
+* Multiplication ```*``` or ```*=```
+* Division ```/``` or ```/=```
+* Power ```**``` or ```**=```
+
+**All these arithmatics always has ```O(1)``` computational cost because, they are not evaluated at the point these arithmetics are used. They are only evaluated when the final value of the ```Expression``` is required.**
```
Igarizza: Not sure were, but we need to mention that the flattening is needed at some point after arithmetics. Otherwise it can be expansive. And, here is a typo in arithm**e**tics. https://en.wikipedia.org/wiki/Arithmetic

docs/pages/Kratos/Expressions/General/Overview.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Use cases
+
+Expressions are usefull in following scenarios:
+1. Transferring data back and forth from/to Kratos with another libraries - Library may be developed in another language like Fortran as well.
+2. Transferring data back and forth from/to Kratos with numpy/scipy.
+3. Transferring data back and forth from/to Kratos with HDF5.
+4. If it is required to compute some intermediary value at python level, where a variable usage is not required such as in computing intermediate values with efficiency.
+5. Transerring data back and forth within Kratos without variables such as in OptimizationApplication.
+6. Prototyping your research code with ease in python level while utilizing **shared memory** and **distributed memory** paralellism.
```
matekelemen:
```suggestion
Expressions are useful in following scenarios:
1. Transferring data between Kratos and other libraries - third-party libraries may be developed in other languages like Fortran as well.
2. Transferring data between Kratos and numpy/scipy.
3. Storing and fetching Kratos data in HDF5.
4. Non performance-critical intermediate calculations in python.
5. Transerring data back and forth within Kratos without variables such as in OptimizationApplication.
6. Quick and convenient prototyping of research code in python while using **shared memory** and **distributed memory** paralellism.
```

docs/pages/Kratos/Expressions/General/Notes.md
```diff
@@ -0,0 +1,15 @@
...
+title: Notes
+keywords: 
+tags: [notes, warnings, expression]
+sidebar: kratos_expressions
+summary: 
+---
+
+## Introduction
+
+The use of ```Expressions```, ```NodalExpression```, ```ConditionExpression``` or ```ElementExpression``` should be done carefully for the following reasons.
```
matekelemen:
```suggestion
The use of ```Expression```, ```NodalExpression```, ```ConditionExpression``` or ```ElementExpression``` should be done carefully for the following reasons.
```

docs/pages/Kratos/Expressions/General/Notes.md
```diff
@@ -0,0 +1,15 @@
...
+
+1. If it is possible to implement the arithmetics or opertations you do in ```C++``` level, then it is encouraged.
+2. Using arithmetics operators should be done carefully. In the case if, an expression being updated by another vectorial expression using operators such as ```+=```, ```-=```, ```*=```, ```/=```, ```**=```, then it will keep both left hand side and right hand side double vectors in memory. If this is being done for lot of iterations, it can exceed available RAM and crash the execution. Therefore, it is a good practice to use ```Kratos.Expression.Utils.Collpase``` to collapse the lazy expression hierarchy.
```
matekelemen:
```suggestion
2. Use arithmetic operators (`+=`, `-=`, `*=`, `/=`, `**=`, `+`, `-`, `*`, `/`) carefully. Operators keep their operands in memory, which means that if you chain operations for long enough, at some point you will run out of RAM. In such cases, `Kratos.Expression.Utils.Collpase` to forcefully evaluate an expression and discard the hierarchy of operands it stored. An example use case would be updating the same expression with new data in every time step of an analysis.
```

docs/pages/Kratos/Expressions/General/Notes.md
```diff
@@ -0,0 +1,15 @@
...
+3. Use of expressions has an overhead cost due to the use of runtime-polymorphism. So, if it is bottleneck, then transfer the expression to the ```C++```.
```
matekelemen:
```suggestion
3. Expressions have significant memory and performance overhead due to dynamic polymorphism. If it becomes a bottleneck, move your calculations to `C++`.
```

docs/pages/Kratos/Expressions/General/Expression_Types.md
```diff
@@ -0,0 +1,79 @@
...
+sidebar: kratos_expressions
+summary: 
+---
+
+## Introduction
+
+There are three types of expressions.
+* NodalExpression
+* ConditionExpression
+* ElementExpression
```
matekelemen:
```suggestion
There are four types of expressions.
* `Expression`
* `NodalExpression`
* `ConditionExpression`
* `ElementExpression`
```

docs/pages/Kratos/Expressions/General/Working_with_Numpy.md
```diff
@@ -0,0 +1,130 @@
...
+node_2 = model_part.CreateNewNode(2, 0.0, 1.0, 0.0)
+
+# now create the expression:
+nodal_expression = Kratos.Expression.NodalExpression(model_part)
+
+# now read in the data from numpy array to expression
+Kratos.Expression.CArrayExpressionIO.Read(nodal_expression, numpy_array)
+
+# now write the expression data to VELOCITY variable
+Kratos.Expression.VariableExpressionIO.Write(nodal_expression, Kratos.VELOCITY, False)
```
Igarizza, line 87: False isn't self-explaining. Consider add "Variable_name=False"

docs/pages/Kratos/Expressions/Utilities/Comb.md
```diff
@@ -0,0 +1,57 @@
...
+---
+
+## Introduction
+This allows combining multiple expressions in to one expression. **The number of entities on all the combining expressions should have the same.**
+
+Example:
+All provided expressions in the list_of_expression.
+```
+For example, let the list_of_expression contain the following expressions:
+exp_list[0] = data{1, 2, 3} with 3 items, and item shape = []
```
Igarizza, line 15: just curious, if shape [ ] takes 1 item, shape = [3] takes 3 items. What should I aspect with shape = [1] and shape = [2]

docs/pages/Kratos/Expressions/General/Expression_Types.md
```diff
@@ -0,0 +1,79 @@
...
+
```
matekelemen, line 23:
```suggestion

## `Expression`

`Expression` is the lowest level class that everything else is built on top of. It models an array of numeric values (scalars, vectors, matrices, etc.), supports arithmetic operations, but is not related to any container in Kratos. To create an `Expression` or write an existing one, you can use one of the derived classes of `ExpressionIO`.

TODO: add an example (@matekelemen) - requires exposing the `Input` and `Output` nested classes of `VariableExpressionIO` to python.


```

docs/pages/Kratos/Expressions/Utilities/Reshape.md
```diff
@@ -0,0 +1,46 @@
...
+Kratos.Expression.VariableExpressionIO.Read(u_exp, Kratos.VELOCITY, False)
+Kratos.Expression.VariableExpressionIO.Read(p_exp, Kratos.PRESSURE, False)
+
+# now combine
+combined_exp = Kratos.Expression.Utils.Comb([p_exp, u_exp])
+
+# now reshape
+reshaped_exp = Kratos.Expression.Utils.Reshape(combined_exp, [2, 2])
+
+print(reshaped_exp.Evaluate())
```
Igarizza, line 44: To avoid running your code you can add printed output as well.

docs/pages/Kratos/Expressions/General/Expression_Types.md
```diff
@@ -0,0 +1,79 @@
...
+## Nodal Expression
+
+Nodal expressions are used to store data related to nodal quantities. They may be historical or non-historical. Following code snippet illustrtes creating a nodal expression
+and reading ```VELOCITY``` from nodal non-historical data value container.
```
matekelemen: Even though the name of the container class is indeed `DataValueContainer`, it's a stupid name that need not be exposed to the reader.


```suggestion
and reading ```VELOCITY``` from non-historical nodal container.
```

docs/pages/Kratos/Expressions/General/Expression_Types.md
```diff
@@ -0,0 +1,79 @@
...
+nodal_expression = Kratos.Expression.NodalExpression(model_part)
+
+# now read the VELOCITY from the non-historical container
+Kratos.Expression.VariableExpressionIO.Read(nodal_expression, Kratos.VELOCITY, False)
+```
+
+## Condition Expression
+
+Condition expressions are used to store data related to condition quantities. Following code snippet illustrtes creating a condition expression
+and reading ```VELOCITY``` from condition data value container.
```
matekelemen:
```suggestion
and reading ```VELOCITY``` from the conditions' container.
```

docs/pages/Kratos/Expressions/General/Expression_Types.md
```diff
@@ -0,0 +1,79 @@
...
+condition_expression = Kratos.Expression.ConditionExpression(model_part)
+
+# now read the VELOCITY from the condition data value container
+Kratos.Expression.VariableExpressionIO.Read(condition_expression, Kratos.VELOCITY)
+```
+
+## Element Expression
+
+Element expressions are used to store data related to element quantities. Following code snippet illustrtes creating a element expression
+and reading ```VELOCITY``` from element data value container.
```
matekelemen:
```suggestion
and reading ```VELOCITY``` from the elements' container.
```

docs/pages/Kratos/Expressions/General/Expression_Arithmetics.md
sunethwarna: The flattening is mentioned in the Notes section. I will change the spelling error.

docs/pages/Kratos/Expressions/General/Working_with_Numpy.md
sunethwarna: It may or may not work depending on whether Kratos is compiled with hinting support. Hence, we have to live like this untill hinting support is made default.

docs/pages/Kratos/Expressions/General/Expression_Arithmetics.md
matekelemen:
```suggestion
These operations are carried out only when their results are required ([lazy evaluation](https://en.wikipedia.org/wiki/Lazy_evaluation)).
```

docs/pages/Kratos/Expressions/Utilities/Comb.md
sunethwarna: `shape[1]` is not allowed. it will throw an error in the `Expression` if it find `shape[1]`. `shape[2]` cannot be used within Kratos since we don't have 2D vectors. But `shape[2]` can be used with `Expressions`, but won't be able to convert to `Kratos::Variable`

docs/pages/Kratos/Expressions/General/Expression_Arithmetics.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Example usage
+Following code snippet explains few use cases for ```NodalExpression```. All the other ```Expression``` also have a similar usage.
```
matekelemen:
```suggestion
Here is an example for using ```NodalExpression``` (other expression types can be used in a similar fashion).
```

docs/pages/Kratos/Expressions/Utilities/Comb.md
sunethwarna: `shape[1]` and `shape[]` both represent scalar quantities.

docs/pages/Kratos/Expressions/General/Working_with_Numpy.md
```diff
@@ -0,0 +1,130 @@
...
+title: Working with Numpy
+keywords: 
+tags: [numpy, scipy, expressions, variable expression io, carray expression io]
+sidebar: kratos_expressions
+summary: 
+---
+
+## Introduction
+
+Expressions make working with nump/scipy or any other thrid party library easier. It allows reading numpy arrays to expressions, and then assign them to data value containers using variables and writing expressions to numpy arrays so variable data in data value containers of kratos do some calculation using numpy.
```
matekelemen:
```suggestion
Expressions make working with numpy/scipy or any other thrid party library easier. It allows reading numpy arrays to expressions, and then assign them to Kratos containers using variables.
```

docs/pages/Kratos/Expressions/General/Working_with_Numpy.md
```diff
@@ -0,0 +1,130 @@
...
+
+**In the case of shared memory parallelized runs, these expressions will hold only data of corresponding data value containers of the local mesh**.
```
matekelemen:
```suggestion
**When running with MPI, these expressions will hold only data of corresponding containers of the local mesh**.
```

docs/pages/Kratos/Expressions/General/Working_with_Numpy.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Writing to numpy arrays
+
+Writing an expression to a numpy array is super simple. In the case of **shared memory** parallelized runs, this will only generate a numpy array having only the data from the
+```LocalMesh```. Following code snippet shows an example:
```
matekelemen: No need for marketing in the docs

```suggestion
```

docs/pages/Kratos/Expressions/General/Working_with_Numpy.md
```diff
@@ -0,0 +1,130 @@
...
+numpy_nodal_expression = nodal_expression.Evaluate()
+
+# first dimension of the numpy array always represents number of entities in the expression (local mesh entities only)
+# rest of the dimensions represent the dimensions of the entity data. In this case, VELOCITY have only three components,
+# it shows 3 as the last dimension.
+print("Shape of the numpy array = ", numpy_nodal_expression.shape)
+```
+
+## Reading from numpy arrays
+Reading from a numpy array is also simple with Expressions. First you can create the numpy array independent from Kratos if you are aware of the number of local entities (i.e. number of nodes/conditions/elements in the local mesh). Followings are supported numpy data types for the conversion:
```
matekelemen:
```suggestion
First create the numpy array independent from Kratos if the number of local entities are known (i.e. number of nodes/conditions/elements in the local mesh). The following numpy data types are supported when reading numpy arrays:
```

docs/pages/Kratos/Expressions/General/Working_with_Numpy.md
```diff
@@ -0,0 +1,130 @@
...
+1. ```numpy.int32```
+2. ```numpy.float64```
+
+Thereafter, you can make a **copy** of the values to an expression. Following code snippet shows an example:
```
matekelemen:
```suggestion
The numpy array can then be **copied** to the expression. The following code snippet shows an example:
```

docs/pages/Kratos/Expressions/General/Working_with_Numpy.md
```diff
@@ -0,0 +1,130 @@
...
+for node in model_part.Nodes:
+    velocity = node.GetValue(Kratos.VELOCITY)
+    print(f"node id: {node.Id}, velocity: [{velocity[0]}, {velocity[1]}, {velocity[2]}]")
+```
+
+## Moving from numpy arrays
+
+This is similar to the [reading from numpy arrays](#reading-from-numpy-arraystest). The difference is, this does not create a copy within the expression.
+It uses the memory location of the original numpy array with in the expression, hence it is of ```O(1)``` cost. No new meory allocation will be done
+within the expression. So, if someone uses the moving of numpy array to an expression, **they need to make sure that the numpy array is not destroyed untill the expression or any of the dependent expressions are used up**. Otherwise, there will be a **Segmentation fault** error.
```
matekelemen:
```suggestion
This is similar to [reading from numpy arrays](#reading-from-numpy-arraystest) with the difference that this does not create a copy for the expression.
It uses the memory location of the original numpy array within the expression, eliding the copy and its related memory allocation. An important consequence is that **the original numpy array must be kept in memory for entire lifetime of the expression**, otherwise evaluating the expression will result in a segmentation fault.
```

matekelemen (commented):
I'll review this PR in chunks. This chunk covers `docs/pages/Kratos/Expressions/General`

docs/pages/Kratos/Expressions/Utilities/Reshape.md
sunethwarna: Will do :)

docs/scripts/build_site.sh
sunethwarna: It is already explained in here 
https://kratosmultiphysics.github.io/Kratos/pages/Kratos/Documentation_Guide/How_Tos/Testing.html

Would that be enough?

docs/pages/Kratos/Expressions/IOs/C_Array_Expression_IO.md
```diff
@@ -0,0 +1,94 @@
...
+title: C Array Expression IO
+keywords: 
+tags: [c array expression io, numpy, scipy, Vector]
+sidebar: kratos_expressions
+summary: 
+---
+
+## Introduction
+
+```CArrayExpressionIO``` is used to exchange data between `C` type arrays and expressions. This allows reading from `C` arrays as well as writing to `C` arrays. In the python later, it can be used to exchange data between numpy arrays, ```Kratos::Vectors``` as well.
```
matekelemen:
```suggestion
```CArrayExpressionIO``` is used to exchange data between `C` type arrays and expressions. In python, it can be used to exchange data between numpy arrays, ```Kratos::Vectors```, or any other contiguous array of supported types.
```

docs/pages/Kratos/Expressions/IOs/C_Array_Expression_IO.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Reading and moving from numpy array
```
matekelemen: Renaming `Move` to `View` is long overdue, so let's not propagate this misnomer to the docs.

```suggestion
## Reading or viewing a numpy array
```

docs/pages/Kratos/Expressions/IOs/C_Array_Expression_IO.md
```diff
@@ -0,0 +1,94 @@
...
+
+Please refer to [Working with numpy](../General/Working_with_Numpy.html) section.
+
+## Writing to numpy array
+Writing an expression to a numpy array is super simple. In the case of **shared memory** parallelized runs, this will only write only the data from the
+```LocalMesh``` to the numpy array. The size of the numpy array should be compatible with the expression shape. Following code snippet shows an example:
```
matekelemen:
```suggestion
In MPI runs, expressions will only write data from the ```LocalMesh``` to the numpy array. The size of the numpy array should be compatible with the expression shape. The following code snippet shows an example:
```

docs/pages/Kratos/Expressions/IOs/C_Array_Expression_IO.md
```diff
@@ -0,0 +1,94 @@
...
+numpy_array = numpy.zeros((2,3), dtype=numpy.float64)
+Kratos.Expression.CArrayExpressionIO.Write(nodal_expression, numpy_array)
+
+# first dimension of the numpy array always represents number of entities in the expression (local mesh entities only)
+# rest of the dimensions represent the dimensions of the entity data. In this case, VELOCITY have only three components,
+# it shows 3 as the last dimension.
+print("numpy array = ", numpy_array)
+```
+
+## Reading and writing to Kratos::Vector
```
matekelemen:
```suggestion
## Reading and writing to `Kratos::Vector`
```

docs/pages/Kratos/Expressions/IOs/C_Array_Expression_IO.md
```diff
@@ -0,0 +1,94 @@
...
+```CArrayExpressionIO``` also allows reading and writing from ```Kratos::Vector``` data containers. In this case, the size of the ```Kratos::Vector``` should be the flattened size of the shape of interest which the user will be working on.
```
matekelemen:
```suggestion
```CArrayExpressionIO``` also allows reading and writing from ```Kratos::Vector``` containers. In this case, the size of the ```Kratos::Vector``` should be the flattened size of the shape of interest which the user will be working on.
```

> shape of interest

I don't know what you mean by that.

docs/pages/Kratos/Expressions/IOs/C_Array_Expression_IO.md
```diff
@@ -0,0 +1,94 @@
...
+
+The arguments for reading from a ```Kratos::Vector``` are, first the expression should be passed, secondly the ```Kratos::Vector``` should be passed. Finally the shape of the entity data should be passed. Following code snippet illustrates the usage.
```
matekelemen:
```suggestion
The arguments for reading from a ```Kratos::Vector``` are:
- expression to read to
- `Kratos::Vector` to read from
- shape of the entity

Example:
```

docs/pages/Kratos/Expressions/IOs/Domain_Size_Expression_IO.md
```diff
@@ -0,0 +1,64 @@
...
+title: Domain Size Expression IO
+keywords: 
+tags: [domain size, condition, element, expression io]
+sidebar: kratos_expressions
+summary: 
+---
+
+## Introduction
+
+```DomainSizeExpressionIO``` computes the domain sizes of each entitiy (such as ```Condition``` or ```Element```). This only supports ```ConditionExpression``` or ```ElementExression```. The domain sizes are computed by accessing underlying geometry's ```Geometry::DomainSize``` method of the entity.
```
matekelemen:
```suggestion
```DomainSizeExpressionIO``` computes the domain sizes (length, surface or volume, depending of the entity's dimension) of each entitiy (such as ```Condition``` or ```Element```) in the provided `ModelPart`. This only supports ```ConditionExpression``` or ```ElementExression```. Domain sizes are computed by accessing underlying geometry's ```Geometry::DomainSize``` method.
```

docs/pages/Kratos/Expressions/IOs/Domain_Size_Expression_IO.md
```diff
@@ -0,0 +1,64 @@
...
+
+1. If the ```Condition``` or ```Element``` as a geometry representing a line, then this will compute an expression with line length for each entitiy.
+2. If the ```Condition``` or ```Element``` as a geometry representing a surface, then this will compute an expression with surface area for each entitiy.
+3. If the ```Condition``` or ```Element``` as a geometry representing a volume, then this will compute an expression with volume for each entitiy.
+
+The resulting expression will be always a scalar expression.
```
matekelemen:
```suggestion
The resulting expression is always a scalar expression.
```

docs/pages/Kratos/Expressions/IOs/Integration_Point_Expression_IO.md
```diff
@@ -0,0 +1,83 @@
...
+title: Integration Point Expression IO
+keywords: 
+tags: [Integration_Point_Expression_IO.md]
+sidebar: kratos_expressions
+summary: 
+---
+
+## Introduction
+
+```IntegrationPointExpressionIO``` is used to set and get the integration point quantities from ```Kratos::Condition``` or ```Kratos::Element``` containers.
```
matekelemen:
```suggestion
```IntegrationPointExpressionIO``` is used to set or get quantities at integration points from elements or conditions.
```

docs/pages/Kratos/Expressions/IOs/Integration_Point_Expression_IO.md
```diff
@@ -0,0 +1,83 @@
...
+* ```double```
+* ```array_1d<double, 3>```
+* ```array_1d<double, 4>```
+* ```array_1d<double, 6>```
+* ```array_1d<double, 9>```
+* ```Vector```
+* ```Matrix```
+
+### Shape of resulting expressions
+The resulting expressions computed by ```IntegrationPointExpressionIO``` will have two additional dimensionalities apart from the dimensionality of the entity data type. The first dimension will illustrate number of entities in the expression. Second dimension will be number of gauss points computed in each entitiy. Rest of the dimensions will be representing the dimensionality of the entity data.
```
matekelemen:
```suggestion
The resulting expressions computed by ```IntegrationPointExpressionIO``` will have two additional dimensions on top of the data type's dimensions. The first one indicates the number of entities in the expression, while the second dimension is the number of gauss points computed in each entitiy. The rest of the dimensions are carried over from the integration point quantity.
```

docs/pages/Kratos/Expressions/IOs/Integration_Point_Expression_IO.md
```diff
@@ -0,0 +1,83 @@
...
+element_expression *= 2.0
+
+# now read the ACCELERATION from the container
+Kratos.Expression.IntegrationPointExpressionIO.Write(element_expression, Kratos.ACCELERATION)
+```
+
+## Using expressions without the model parts
+The ```NodalExpression```, ```ConditionExpression``` and ```ElementExpression``` has an expression which can be directly used if required. The advantage of working
+with the ```Expression``` directely is, then it is not bound to a model part of a ```DataValueContainer```. Hence, these expressions can be interchanged if required in
+advanced use cases. Following code snippet shows how to use bare ```Expressions```.
```
matekelemen:
```suggestion
The underlying `Expression` instance of ```NodalExpression```, ```ConditionExpression``` and ```ElementExpression``` can be accessed directly, if necessary. The advantage of working
with bare ```Expression```s is that they are not bound to a specific model part or entity type, which means that they are completely interchangable. The following code snippet shows how to use bare ```Expression```s.
```

docs/pages/Kratos/Expressions/IOs/Literal_Expression_Input.md
```diff
@@ -0,0 +1,140 @@
...
+title: Literal Expression Input
+keywords: 
+tags: [literal values, expressions, scalars]
+sidebar: kratos_expressions
+summary: 
+---
+
+## Introduction
+
+```LiteralExpressionIO``` is used to represent literal values. Following literal types are supported:
```
matekelemen:
```suggestion
```LiteralExpressionIO``` is used to represent literal values. The following literal types are supported:
```

docs/pages/Kratos/Expressions/IOs/Literal_Expression_Input.md
```diff
@@ -0,0 +1,140 @@
...
+* double
+* array_1d<double, 3>
+* array_1d<double, 4>
+* array_1d<double, 6>
+* array_1d<double, 9>
+* Vector
+* Matrix
+
+This IO will create an expression with the specified literal for all entities. Since it is going to be constant for all entities, the literal will be stored only once for
+all entities making the memory occupation the least. When the expression is evaluated, this stored literal will be read by all entities.
```
matekelemen:
```suggestion
This IO will create an expression with the specified literal for all entities.
```

docs/pages/Kratos/Expressions/IOs/Literal_Expression_Input.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Setting literal expressions
+
+Following code snippet shows how to set literal expressions
```
matekelemen:
```suggestion
The following code snippet shows how to set literal expressions
```

docs/pages/Kratos/Expressions/IOs/Literal_Expression_Input.md
```diff
@@ -0,0 +1,140 @@
...
+
+# now check
+for node in model_part.Nodes:
+    velocity = node.GetValue(Kratos.VELOCITY)
+    print(f"node_id: {node.Id}, velocity = [{velocity[0]}, {velocity[1]}, {velocity[2]}]")
+```
+
+## Setting literal expression to zero
+
+The ```LiteralExpressionIO``` also can be used to set entity data to zero for a given variable. **In the case of dynamic data types, it will create empty data sets [Such as Vector with size 0, Matrix with shape (0,0)**. The passed variable is only used to determine the entity data shape. Following code-snippet shows an example:
```
matekelemen:
```suggestion
The ```LiteralExpressionIO``` also can be used to set entity data to zero for a given variable. **In the case of dynamically sized types, it will create empty instances [Such as Vector with size 0, Matrix with shape (0,0)**. The passed variable is only used to determine the entity shape. Example:
```

docs/pages/Kratos/Expressions/IOs/Literal_Expression_Input.md
```diff
@@ -0,0 +1,140 @@
...
+Kratos.Expression.VariableExpressionIO.Read(nodal_expression, Kratos.PRESSURE, False)
+
+print("PRESSURE before resetting:\n", nodal_expression.Evaluate())
+
+scalar_expression = Kratos.Expression.NodalExpression(model_part)
+Kratos.Expression.LiteralExpressionIO.SetData(scalar_expression, 1.0)
+inverse_expression = scalar_expression / nodal_expression
+
+# now write the value to non-historical velocity
+Kratos.Expression.VariableExpressionIO.Write(inverse_expression, Kratos.TEMPERATURE, False)
```
matekelemen, line 139: > now write the value to non-historical velocity

please don't do blind copy-pasta; this is everywhere so please go through your examples and correct the variable names

docs/pages/Kratos/Expressions/IOs/Literal_Expression_Input.md
```diff
@@ -0,0 +1,140 @@
...
+
+# now check
```
matekelemen, line 141:
```suggestion
```

docs/pages/Kratos/Expressions/IOs/Literal_Expression_Input.md
```diff
@@ -0,0 +1,140 @@
...
+for node in model_part.Nodes:
+    velocity = node.GetValue(Kratos.VELOCITY)
+    print(f"node_id: {node.Id}, pressure = {node.GetValue(Kratos.PRESSURE)}, temperature = {node.GetValue(Kratos.TEMPERATURE)}")
```
matekelemen: again, it's very confusing if you do copy-pasta in the docs


```suggestion
for node in model_part.Nodes:
    print(f"node_id: {node.Id}, pressure = {node.GetValue(Kratos.PRESSURE)}, temperature = {node.GetValue(Kratos.TEMPERATURE)}")
```

docs/pages/Kratos/Expressions/IOs/Literal_Expression_Input.md
```diff
@@ -0,0 +1,140 @@
...
+```
+
+## Using expressions without the model parts
+The ```NodalExpression```, ```ConditionExpression``` and ```ElementExpression``` has an expression which can be directly used if required. The advantage of working
+with the ```Expression``` directely is, then it is not bound to a model part of a ```DataValueContainer```. Hence, these expressions can be interchanged if required in
+advanced use cases. Following code snippet shows how to use bare ```Expressions```.
```
matekelemen, line 166: please copy my other suggestion

docs/pages/Kratos/Expressions/IOs/Nodal_Position_Expression_IO.md
```diff
@@ -0,0 +1,78 @@
...
+title: Nodal Position Expression IO
+keywords: 
+tags: [nodal position, initial, current, expressions]
+sidebar: kratos_expressions
+summary: 
+---
+
+## Introduction
+
+```NodalPositionExpressionIO``` allows writing and reading nodal positions of a given model part. This can be used on following configurations:
```
matekelemen:
```suggestion
```NodalPositionExpressionIO``` allows writing and reading nodal positions of a given model part. This can be used on following the configurations:
```

docs/pages/Kratos/Expressions/IOs/Nodal_Position_Expression_IO.md
```diff
@@ -0,0 +1,78 @@
...
+* Initial -> This uses the initial nodal position to read and write.
+* Current -> This uses the current position of the nodes to read and write.
```
matekelemen:
```suggestion
* Initial -> initial positions of the node
* Current -> positions of the nodes in the current configuration
```

docs/pages/Kratos/Expressions/IOs/Nodal_Position_Expression_IO.md
```diff
@@ -0,0 +1,78 @@
...
+
+# now read nodal positions
+for node in model_part.Nodes:
+    print(f"node_id: {node.Id}, initial: [{node.X0}, {node.Y0}, {node.Z0}], current: [{node.X}, {node.Y}, {node.Z}]")
+```
+
+## Using expressions without the model parts
+The ```NodalExpression```, ```ConditionExpression``` and ```ElementExpression``` has an expression which can be directly used if required. The advantage of working
+with the ```Expression``` directely is, then it is not bound to a model part of a ```DataValueContainer```. Hence, these expressions can be interchanged if required in
+advanced use cases. Following code snippet shows how to use bare ```Expressions```.
```
matekelemen: same

docs/pages/Kratos/Expressions/IOs/Variable_Expression_IO.md
```diff
@@ -0,0 +1,153 @@
...
+title: Variable Expression IO
+keywords: 
+tags: [variable expression io, expressions, variable]
+sidebar: kratos_expressions
+summary: 
+---
+
+## Introduction
+
+```VariableExpressionIO``` is used to read ```Kratos::Variable``` data in a data value container to a expression or write an expression data in to a ```Kratos::Variable``` in a specified data value container. The expression will only read or write to the entities (i.e. nodes/conditions/elements) in the local mesh. In the case of writing expression data to a variable in data value container, if it is used in ***distributed memory** parallelized environment, proper synchronization will be automatically done. Hence, all of these operations are compatible with **shared memory** and **distributed memory** parallelizations.
```
matekelemen:
```suggestion
```VariableExpressionIO``` is used to read ```Kratos::Variable``` from a container to an expression vice versa. The expression will only read or write to the entities (i.e. nodes/conditions/elements) in the local mesh. If writing an expression to a variable in a container during an MPI run, synchronization between ranks is automatically done.
```

docs/pages/Kratos/Expressions/IOs/Variable_Expression_IO.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Variable types
+
+Following variable types are allowed to be read in or write to from an expression.
```
matekelemen:
```suggestion
The following Kratos variable types are supported:
```

docs/pages/Kratos/Expressions/IOs/Variable_Expression_IO.md
```diff
@@ -0,0 +1,153 @@
...
+* ```Kratos::Variable<double>```
+* ```Kratos::Variable<array_1d<double, 3>>```
+* ```Kratos::Variable<array_1d<double, 4>>```
+* ```Kratos::Variable<array_1d<double, 6>>```
+* ```Kratos::Variable<array_1d<double, 9>>```
+* ```Kratos::Variable<Vector>```
+* ```Kratos::Variable<Matrix>```
+
+### Shape of the resulting expression
+In the case of static data types given by static data type variables (such as ```Kratos::Variable<double>```, ```Kratos::Variable<array_1d<double, 3>>```), the shape of the expression will be always have number of local entities in the first dimension, rest will be followed by the static size of the data type. If the variable type being used is of dynamic data type (such as ```Kratos::Variable<Vector>```, ```Kratos::Variable<Matrix>```), then it will represent still the number of local entities in the first dimension. Rest of the dimensions will represent the dimenions of the first data value found in the respective container. [In the case of **distributed memory** parallelized runs, first data value shape synchronization is done to identify the shape of the expression].
```
matekelemen:
```suggestion
If the data type has static size (such as ```Kratos::Variable<double>```, ```Kratos::Variable<array_1d<double, 3>>```), the result's shape is the number of local entities, followed by the static size of the type. However, if the data type has dynamic size (such as ```Kratos::Variable<Vector>```, ```Kratos::Variable<Matrix>```), then all entries in the container are assumed to have the same shape as the first entry. [In MPI runs, the first entry's shape is synchronized across all ranks to determine the final shape of the expression].
```

docs/pages/Kratos/Expressions/IOs/Variable_Expression_IO.md
```diff
@@ -0,0 +1,153 @@
...
+
+# now print and see
+for node in model_part.Nodes:
+    velocity = node.GetValue(Kratos.VELOCITY)
+    acceleration = node.GetSolutionStepValue(Kratos.ACCELERATION)
+    print(f"node_id: {node.Id}, velocity: [{velocity[0]},{velocity[1]},{velocity[2]}], acceleration: [{acceleration[0]},{acceleration[1]},{acceleration[2]}]")
+```
+
+## Reading/Writing condition values
+Reading and writing to condition values have the same interface within ```VariableExpressionIO```. The first argument shall be a `ConditionExpression` and the second argumen shall be a ```Kratos::Variable```.
```
matekelemen:
```suggestion
```

docs/pages/Kratos/Expressions/IOs/Variable_Expression_IO.md
```diff
@@ -0,0 +1,153 @@
...
+
+
+## Reading/Writing nodal values
+Reading and writing to nodal values have the same interface within ```VariableExpressionIO```. The first argument shall be a `NodalExpression`, the second argumen shall be a ```Kratos::Variable``` and the last argument shall be a boolean indicating true if it is required to read from historical nodal data value container or false if it is required read from non-historical nodal data value container. In the case of **distributed memory** parallelized runs, a proper synchronization between ghost nodes are done when writing is executed.
```
matekelemen:
```suggestion
```

docs/pages/Kratos/Expressions/IOs/Variable_Expression_IO.md
```diff
@@ -0,0 +1,153 @@
...
+
+# now print and see
+for condition in model_part.Conditions:
+    velocity = condition.GetValue(Kratos.VELOCITY)
+    acceleration = condition.GetValue(Kratos.ACCELERATION)
+    print(f"condition_id: {condition.Id}, velocity: [{velocity[0]},{velocity[1]},{velocity[2]}], acceleration: [{acceleration[0]},{acceleration[1]},{acceleration[2]}]")
+```
+
+## Reading/Writing element values
+Reading and writing to element values have the same interface within ```VariableExpressionIO```. The first argument shall be a `ElementExpression` and the second argumen shall be a ```Kratos::Variable```.
```
matekelemen:
```suggestion
```

docs/pages/Kratos/Expressions/IOs/Variable_Expression_IO.md
```diff
@@ -0,0 +1,153 @@
...
+for element in model_part.Elements:
+    velocity = element.GetValue(Kratos.VELOCITY)
+    acceleration = element.GetValue(Kratos.ACCELERATION)
+    print(f"element_id: {element.Id}, velocity: [{velocity[0]},{velocity[1]},{velocity[2]}], acceleration: [{acceleration[0]},{acceleration[1]},{acceleration[2]}]")
+```
+
+## Using expressions without the model parts
+The ```NodalExpression```, ```ConditionExpression``` and ```ElementExpression``` has an expression which can be directly used if required. The advantage of working
+with the ```Expression``` directely is, then it is not bound to a model part of a ```DataValueContainer```. Hence, these expressions can be interchanged if required in
+advanced use cases. Following code snippet shows how to use bare ```Expressions```.
```
matekelemen, line 169: same

docs/pages/Kratos/Expressions/Utilities/Reshape.md
sunethwarna: done :)

docs/pages/Kratos/Expressions/IOs/Literal_Expression_Input.md
sunethwarna: Sorry about that. I checked the examples, and removed the faulty ones.

docs/pages/Kratos/Expressions/IOs/Literal_Expression_Input.md
sunethwarna: resolved.

sunethwarna:
@matekelemen @Igarizza I have addressed your comments. Could you take a look?

Igarizza (dismissed):

docs/pages/Kratos/Expressions/Utilities/Abs.md
```diff
@@ -0,0 +1,116 @@
...
+title: Abs
+keywords: 
+tags: [abs, expression]
+sidebar: kratos_expressions
+summary: 
+---
+
+## Introduction
+
+This computes absolute value of each component of the given expression. Assume the input expression is given by $$\underline{\mathbb{u}} = \left\lbrace u_{ij},  \forall (i,j)\in\left[0, M\right)\times\left[0, N\right)\right\rbrace$$ where $$i^{th}$$ entity's $$j^{th}$$ component is represented by $$u_{ij}$$ with $$i\in \left[0, M\right)$$ for each entity and $$j\in \left[0, N\right)$$ for each component in each entity. Following equation illustrates the formulation of the resulting expression.
```
matekelemen:
```suggestion
This computes the componentwise absolute value of the given expression. Assume the input expression is given by $$\underline{\mathbf{u}} = \left\lbrace u_{ij},  \forall (i,j)\in\left[0, M\right)\times\left[0, N\right)\right\rbrace$$ where the $$i^{th}$$ entity's $$j^{th}$$ component is represented by $$u_{ij}$$. The following equation illustrates the formulation of the resulting expression.
```

docs/pages/Kratos/Expressions/Utilities/Abs.md
```diff
@@ -0,0 +1,116 @@
...
+
+<p align="center">$$ Abs\left(\underline{\mathbb{u}}\right) = \left\lbrace \left|u_{ij}\right|,  \forall (i,j)\in\left[0, M\right)\times\left[0, N\right)\right\rbrace$$</p>
```
matekelemen:
```suggestion
<p align="center">$$ abs\left(\underline{\mathbf{u}}\right) = \left\lbrace \left|u_{ij}\right|,  \forall (i,j)\in\left[0, M\right)\times\left[0, N\right)\right\rbrace$$</p>
```

docs/pages/Kratos/Expressions/Utilities/Abs.md
```diff
@@ -0,0 +1,116 @@
...
+node_id: 3, velocity=[-7.0, -8.0, -9.0], acceleration = [7.0, 8.0, 9.0]
+ |  /           |                  
+ ' /   __| _` | __|  _ \   __|    
+ . \  |   (   | |   (   |\__ \  
+_|\_\_|  \__,_|\__|\___/ ____/
+           Multi-Physics 9.4."3"-docs/expression_documentation-6de5f1a499-Release-x86_64
+           Compiled for GNU/Linux and Python3.11 with GCC-13.2
+Compiled with threading and MPI support.
+Maximum number of threads: 30.
+Running without MPI.
```
matekelemen:
```suggestion
 |  /           |                  
 ' /   __| _` | __|  _ \   __|    
 . \  |   (   | |   (   |\__ \  
_|\_\_|  \__,_|\__|\___/ ____/
           Multi-Physics 9.4."3"-docs/expression_documentation-6de5f1a499-Release-x86_64
           Compiled for GNU/Linux and Python3.11 with GCC-13.2
Compiled with threading and MPI support.
Maximum number of threads: 30.
Running without MPI.
node_id: 1, velocity=[-1.0, -2.0, -3.0], acceleration = [1.0, 2.0, 3.0]
node_id: 2, velocity=[-4.0, -5.0, -6.0], acceleration = [4.0, 5.0, 6.0]
node_id: 3, velocity=[-7.0, -8.0, -9.0], acceleration = [7.0, 8.0, 9.0]
```

docs/pages/Kratos/Expressions/Utilities/Abs.md
```diff
@@ -0,0 +1,116 @@
...
+node_id: 3, velocity=[-7.0, -8.0, -9.0], pressure = 9.0
+ |  /           |                  
+ ' /   __| _` | __|  _ \   __|    
+ . \  |   (   | |   (   |\__ \  
+_|\_\_|  \__,_|\__|\___/ ____/
+           Multi-Physics 9.4."3"-docs/expression_documentation-6de5f1a499-Release-x86_64
+           Compiled for GNU/Linux and Python3.11 with GCC-13.2
+Compiled with threading and MPI support.
+Maximum number of threads: 30.
+Running without MPI.
```
matekelemen:
```suggestion
 |  /           |                  
 ' /   __| _` | __|  _ \   __|    
 . \  |   (   | |   (   |\__ \  
_|\_\_|  \__,_|\__|\___/ ____/
           Multi-Physics 9.4."3"-docs/expression_documentation-6de5f1a499-Release-x86_64
           Compiled for GNU/Linux and Python3.11 with GCC-13.2
Compiled with threading and MPI support.
Maximum number of threads: 30.
Running without MPI.
node_id: 1, velocity=[-1.0, -2.0, -3.0], pressure = 3.0
node_id: 2, velocity=[-4.0, -5.0, -6.0], pressure = 6.0
node_id: 3, velocity=[-7.0, -8.0, -9.0], pressure = 9.0
```

docs/pages/Kratos/Expressions/Utilities/Collapse.md
```diff
@@ -0,0 +1,58 @@
...
+title: Collapse
+keywords: 
+tags: [collapse, lazy expression, expressions]
+sidebar: kratos_expressions
+summary: 
+---
+
+## Introduction
+```Collapse``` method is used to collapse the lazy expression tree. **This is useful when the lazy expression tree is larger which occupies or exceeds the RAM
+limitations.**.
```
matekelemen:
```suggestion
The ```Collapse``` method forces an evaluation of the input expression into a flat array. **This is useful when the input expression represents a complex/deep/chained operation with many operands that occupy considerable memory. In such a scenario, `Collapse` allows the user to replace the original expression hierarchy with a single array (freeing up RAM), at the expense of immediate evaluation.**.
```

docs/pages/Kratos/Expressions/Utilities/Collapse.md
```diff
@@ -0,0 +1,58 @@
...
+
+Using this method on an expression which will always result in an expression having the same shape, having the same evaluated values. But the resulting expression will be not having the tree structure since it is collapsed.
```
matekelemen:
```suggestion
The resulting expression retains the original's shape.
```

docs/pages/Kratos/Expressions/Utilities/Collapse.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Use case
+The max depth of an expression can be checked to identify the max depth of the lazy expression as explained in the following code snippet. It also explains how to collapse the tree hierarchy of the lazy expressions.
```
matekelemen:
```suggestion
The following example shows how to compute the maximum depth of the expression hierarchy, and collapse it down to a single array.
```

docs/pages/Kratos/Expressions/Utilities/Collapse.md
```diff
@@ -0,0 +1,58 @@
...
+# read the VELOCITY from non-historical nodal container
+Kratos.Expression.VariableExpressionIO.Read(nodal_expression, Kratos.VELOCITY, False)
+
+nodal_expression *= 2
+nodal_expression += 1
+
+print(nodal_expression.GetMaxDepth())
+
+collapsed_exp = Kratos.Expression.Utils.Collapse(nodal_expression)
+print(collapsed_exp.GetMaxDepth())
```
matekelemen:
```suggestion
collapsed_exp = Kratos.Expression.Utils.Collapse(nodal_expression)

# Destroy the original expression to free up memory
del nodal_expression

print(collapsed_exp.GetMaxDepth())
```

docs/pages/Kratos/Expressions/Utilities/Collapse.md
```diff
@@ -0,0 +1,58 @@
...
+1
+ |  /           |                  
+ ' /   __| _` | __|  _ \   __|    
+ . \  |   (   | |   (   |\__ \  
+_|\_\_|  \__,_|\__|\___/ ____/
+           Multi-Physics 9.4."3"-docs/expression_documentation-6de5f1a499-Release-x86_64
+           Compiled for GNU/Linux and Python3.11 with GCC-13.2
+Compiled with threading and MPI support.
+Maximum number of threads: 30.
+Running without MPI.
```
matekelemen:
```suggestion
 |  /           |                  
 ' /   __| _` | __|  _ \   __|    
 . \  |   (   | |   (   |\__ \  
_|\_\_|  \__,_|\__|\___/ ____/
           Multi-Physics 9.4."3"-docs/expression_documentation-6de5f1a499-Release-x86_64
           Compiled for GNU/Linux and Python3.11 with GCC-13.2
Compiled with threading and MPI support.
Maximum number of threads: 30.
Running without MPI.
3
1
```

docs/pages/Kratos/Expressions/Utilities/Comb.md
```diff
@@ -0,0 +1,72 @@
+---
+title: Comb
+keywords: 
+tags: [combing, expressions]
+sidebar: kratos_expressions
+summary: 
+---
+
+## Introduction
+This allows combining multiple expressions in to one expression. **The number of entities on all the combining expressions should have the same.**
```
matekelemen:
```suggestion
`Comb` merges multiple expressions into one. **All input expressions must have the same number of entities.**
```

docs/pages/Kratos/Expressions/Utilities/Comb.md
```diff
@@ -0,0 +1,72 @@
...
+
+Example:
+All provided expressions in the list_of_expression.
```
matekelemen:
```suggestion
```

docs/pages/Kratos/Expressions/Utilities/Comb.md
```diff
@@ -0,0 +1,72 @@
...
+For example, let the list_of_expression contain the following expressions:
+exp_list[0] = data{1, 2, 3} with 3 items, and item shape = []
+                   -  -  -
+exp_list[1] = data{4, 5, -1, 6, 7, -1, 8, 9, -1} with 3 items, and item shape = [3]
+                   --------  --------  --------
+The resulting expression has item shape [3] with 4 items:
+output_exp = [1, 4, 5, -1, 2, 6, 7, -1, 3, 8, 9, -1]
+                ---------     --------     --------
+```
+This creates a lazy expression, hence it has a constant cost complexity irrespective of the data size (The expression won't be evaluated unless @ref Expression::Evaluate is called).
```
matekelemen:
```suggestion
The expression won't be evaluated unless @ref Expression::Evaluate is called.
```

docs/pages/Kratos/Expressions/Utilities/Comb.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Use case
+Following code snippet illustrates how the above example can be replicated.
```
matekelemen:
```suggestion
```

docs/pages/Kratos/Expressions/Utilities/Comb.md
```diff
@@ -0,0 +1,72 @@
...
+ [ 3.  8.  9. -1.]]
+ |  /           |                  
+ ' /   __| _` | __|  _ \   __|    
+ . \  |   (   | |   (   |\__ \  
+_|\_\_|  \__,_|\__|\___/ ____/
+           Multi-Physics 9.4."3"-docs/expression_documentation-6de5f1a499-Release-x86_64
+           Compiled for GNU/Linux and Python3.11 with GCC-13.2
+Compiled with threading and MPI support.
+Maximum number of threads: 30.
+Running without MPI.
```
matekelemen: Please fix the rest of the expected outputs as well.

```suggestion
 |  /           |                  
 ' /   __| _` | __|  _ \   __|    
 . \  |   (   | |   (   |\__ \  
_|\_\_|  \__,_|\__|\___/ ____/
           Multi-Physics 9.4."3"-docs/expression_documentation-6de5f1a499-Release-x86_64
           Compiled for GNU/Linux and Python3.11 with GCC-13.2
Compiled with threading and MPI support.
Maximum number of threads: 30.
Running without MPI.
[[ 1.  4.  5. -1.]
 [ 2.  6.  7. -1.]
 [ 3.  8.  9. -1.]]
```

docs/pages/Kratos/Expressions/Utilities/EntityMax.md
```diff
@@ -0,0 +1,62 @@
...
+title: EntityMax
+keywords: 
+tags: [entity max, max, expression]
+sidebar: kratos_expressions
+summary: 
+---
+
+## Introduction
+
+This get the maximum value from each component in each entity. Assume the input expression is given by $$\underline{\mathbb{u}} = \left\lbrace u_{ij},  \forall (i,j)\in\left[0, M\right)\times\left[0, N\right)\right\rbrace$$ where $$i^{th}$$ entity's $$j^{th}$$ component is represented by $$u_{ij}$$ with $$i\in \left[0, M\right)$$ for each entity and $$j\in \left[0, N\right)$$ for each component in each entity. Following equation illustrates the formulation of the resulting expression which is always a scalar expression.
```
matekelemen:
```suggestion
`EntityMax` finds the largest component of each entity. Assume the input expression is given by $$\underline{\mathbf{u}} = \left\lbrace u_{ij},  \forall (i,j)\in\left[0, M\right)\times\left[0, N\right)\right\rbrace$$ where $$i^{th}$$ entity's $$j^{th}$$ component is represented by $$u_{ij}$$. The result is always a scalar expression.
```

docs/pages/Kratos/Expressions/Utilities/EntityMax.md
```diff
@@ -0,0 +1,62 @@
...
+
+<p align="center">$$ EntityMax\left(\underline{\mathbb{u}}\right) = \left\lbrace v_i, \text{where} \quad v_i = \max_{j\in\left[0, N\right)} {u_{ij}} \quad  \forall i\in\left[0, M\right)\right\rbrace$$</p>
```
matekelemen:
```suggestion
<p align="center">$$ EntityMax\left(\underline{\mathbf{u}}\right) = \left\lbrace v_i | \quad v_i = \max_{j\in\left[0, N\right)} {u_{ij}} \quad  \forall i\in\left[0, M\right)\right\rbrace$$</p>
```

docs/pages/Kratos/Expressions/Utilities/EntityMax.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Use cases
+Following code snippet illustrates how to use ```EntityMax```.
```
matekelemen:
```suggestion
```

matekelemen:
the `docs/_site` directory needs a `.gitignore` because the files generated by jekyll get tracked by git.

docs/pages/Kratos/Expressions/Utilities/Abs.md
sunethwarna: done :)

docs/pages/Kratos/Expressions/Utilities/Abs.md
sunethwarna: done :)

docs/pages/Kratos/Expressions/Utilities/Abs.md
sunethwarna: fixed :)

docs/pages/Kratos/Expressions/Utilities/Abs.md
sunethwarna: fixed

docs/pages/Kratos/Expressions/Utilities/Collapse.md
sunethwarna: done :)

docs/pages/Kratos/Expressions/Utilities/Collapse.md
sunethwarna: done :)

docs/pages/Kratos/Expressions/Utilities/Collapse.md
sunethwarna: done :)

docs/pages/Kratos/Expressions/Utilities/Collapse.md
sunethwarna: done :)

docs/pages/Kratos/Expressions/Utilities/Collapse.md
sunethwarna: done :)

docs/pages/Kratos/Expressions/Utilities/Comb.md
sunethwarna: done :)

docs/pages/Kratos/Expressions/Utilities/Comb.md
sunethwarna: done :)

docs/pages/Kratos/Expressions/Utilities/Comb.md
sunethwarna: done :)

docs/pages/Kratos/Expressions/Utilities/Comb.md
sunethwarna: done :)

docs/pages/Kratos/Expressions/Utilities/EntityMax.md
sunethwarna: done :)

docs/pages/Kratos/Expressions/Utilities/EntityMax.md
sunethwarna: done :)

docs/pages/Kratos/Expressions/Utilities/EntityMax.md
sunethwarna: done :)

sunethwarna:
@matekelemen I have addressed your comments, and went through the copy pasta stuff also. Could you take a look?

docs/pages/Kratos/Expressions/Utilities/Comb.md
sunethwarna: done :)

matekelemen (approved):

PR #11998, opened 2024-01-30
[GeoMechanicsApplication] Settlement output file is not closed on error

### LINKED ISSUE #11994: [GeoMechanicsApplication] Output file is not closed when settlement calculation throws an exception
When a settlement calculation is run using the custom settlement workflow, and when this calculation throws an exception, the output file is not properly closed. The issue can be reproduced by running the custom settlement workflow with the attached input files. Stage 2 raises an error which results in the reported behavior.

**Note:** It appears that `.json` and `.mdpa` cannot be directly attached to an issue on GitHub. The corresponding input files can be found at the following network location (accessible to Deltares staff only): `N:\Deltabox\Postbox\Graaf,van de Anne\vanWalter`.

avdg81:
**📝 Description**
Fixed an issue where the output file was not properly closed when an exception was thrown from a settlement calculation.

**🆕 Changelog**
- Added a class for scoped output file access, which opens the output file on construction, and which closes it on destruction. An instance of this class now ensures that any output file access is limited to the runtime of the `Run` method of the time loop executor. That is, the output file is now always closed, regardless whether `Run` is left through a `return` statement or through an exception.
- Split the implementation of the `Initialize` member function of the solver strategy wrapper. Initializing the output has been moved to a separate member function (i.e. the interface for initializing and finalizing the output is now symmetric).
- Added two unit tests that cover the improvements.

### DIFF
--- a/applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
+++ b/applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
@@ -17,6 +17,7 @@
 #include <memory>
 #include <vector>
 
+#include "scoped_output_file_access.h"
 #include "strategy_wrapper.hpp"
 #include "time_incrementor.h"
 #include "time_loop_executor_interface.h"
@@ -63,6 +64,7 @@ public:
         mStrategyWrapper->SaveTotalDisplacementFieldAtStartOfTimeLoop();
         std::vector<TimeStepEndState> result;
         TimeStepEndState              NewEndState = EndState;
+        ScopedOutputFileAccess        limit_output_file_access_to_this_scope{*mStrategyWrapper};
         while (mTimeIncrementor->WantNextStep(NewEndState) && !IsCancelled()) {
             mStrategyWrapper->IncrementStepNumber();
             // clone without end time, the end time is overwritten anyway
@@ -73,8 +75,6 @@ public:
             mStrategyWrapper->OutputProcess();
             result.emplace_back(NewEndState);
         }
-
-        mStrategyWrapper->FinalizeOutput();
 
         return result;
     }
--- a/applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
+++ b/applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
@@ -101,7 +101,16 @@ public:
         return mCountAccumulateTotalDisplacementFieldCalled;
     }
 
-    void OutputProcess() override { ++mCountOutputProcessCalled; }
+    void OutputProcess() override
+    {
+        ++mCountOutputProcessCalled;
+        if (mOutputProcessCallback) mOutputProcessCallback();
+    }
+
+    [[nodiscard]] std::size_t GetCountInitializeOutputCalled() const
+    {
+        return mCountInitializeOutputCalled;
+    }
 
     [[nodiscard]] std::size_t GetCountOutputProcessCalled() const
     {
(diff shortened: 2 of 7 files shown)


WPK4FEM (approved):
Nice and clean solution.
I could find and follow all components and got matching explanation in our phone call.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
```diff
@@ -358,4 +386,37 @@ KRATOS_TEST_CASE_IN_SUITE(ExpectFinalizeSolutionStepCalledOnceForEveryStep, Krat
...
+    KRATOS_EXPECT_EQ(solver_strategy->GetCountFinalizeOutputCalled(), 1);
+}
+
+KRATOS_TEST_CASE_IN_SUITE(ExpectOutputIsInitializedAndFinalizedWhenRunThrows, KratosGeoMechanicsFastSuite)
+{
+    TimeLoopExecutor executor;
+    const auto       wanted_num_of_cycles_per_step = std::size_t{1};
+    executor.SetTimeIncrementor(std::make_unique<FixedCyclesTimeIncrementor>(wanted_num_of_cycles_per_step));
+    auto solver_strategy = std::make_shared<DummySolverStrategy>(TimeStepEndState::ConvergenceState::converged);
+    solver_strategy->SetOutputProcessCallback([](){throw Exception{"Test exception"};});
```
rfaasse, line 407: Since we always use KRATOS_ERROR in the code, it might be nice to have that here as well (instead of throw).

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
```diff
@@ -63,6 +64,7 @@ class TimeLoopExecutor : public TimeLoopExecutorInterface
         mStrategyWrapper->SaveTotalDisplacementFieldAtStartOfTimeLoop();
         std::vector<TimeStepEndState> result;
         TimeStepEndState              NewEndState = EndState;
+        ScopedOutputFileAccess        limit_output_file_access_to_this_scope{*mStrategyWrapper};
```
rfaasse, line 67: Do you think we need to add a comment somewhere to refer to the RAII principle? I can imagine it could be puzzling when seeing it for the first time (although the naming also already gives quite a clear view on what this should do).

rfaasse (approved):
Nice and clean solution! I have two very minor comments/questions. It's of course up to you if you think it's worth making small changes for.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
avdg81: What I like about explicitly writing `throw Exception` is there is no doubt about what kind of exception needs to be caught later in the test. In other words, when you would use `KRATOS_ERROR` you would need to know that it throws a `Kratos::Exception`. In production code we use `KRATOS_TRY` and `KRATOS_CATCH` for exception handling, but those macros hide the exception type that is being thrown. I can't use `KRATOS_CATCH` here, since it **rethrows** the exception after having caught it. For clarity, I prefer how it is currently expressed, but if you have convincing arguments to change it, I would like to learn about them.

applications/GeoMechanicsApplication/custom_workflows/time_loop_executor.hpp
avdg81: I don't think that is necessary. For instance, one of our colleagues was not aware of the name of this idiom (RAII), but he perfectly understood how this class works. It would, however, be a good candidate for our internal knowledge sharing program.

applications/GeoMechanicsApplication/tests/cpp_tests/test_time_loop_executor.cpp
rfaasse: Good point, no follow-up needed, thanks!

PR #12008, opened 2024-02-02
[OptApp] Cleanup old files

sunethwarna:
**📝 Description**
Untill now the `OptApp` had twoversions overlapping with each other. This PR removes the old implementations encouraging to go forward with the new implementation. There will be upcoming PRs where we will port the old responses to the new implementation.

**🆕 Changelog**
- Removes old implementations


Igarizza (approved):
Looks correct

PR #12009, opened 2024-02-02
[Docs] Add python processing locally

sunethwarna:
This PR adds capability to add python snippet output to the docs automatically. Following is the documentation on how to use it.

1. Add `## POST_PROCESS_PAGES_PYTHON_OUTPUT_GENERATION` anywhere in the python snippet. It should not have any leading or trailing spaces/tabs.
2. Run the local build in your computer [Make sure to initialize kratos environment or any other library environments in the terminal which is being used by the snippet]. This will run the snippet and capture the output. It will also remove the `## POST_PROCESS_PAGES_PYTHON_OUTPUT_GENERATION` tag line from the python snippet.
3. Now you will see some block after the python snippet which contains the python output.

This generation is only done if ```process_pages.py``` is passed with `-t local` flag.

** This will not change any existing python snippets. If the automated output is required, the above tag needs to be placed within the python snippet. **

**🆕 Changelog**
- Adds capability to automatically add python snippet output.

### DIFF
--- a/docs/process_pages.py
+++ b/docs/process_pages.py
@@ -251,7 +251,7 @@ def AddPythonSnippetOutputs(file_path: Path) -> None:
                     with open(temp_file_path, "w") as temp_file_output:
                         temp_file_output.writelines(snippet_lines)
 
-                    subprocess_run = subprocess.run([GetPython3Command(), temp_file_path], stdout=subprocess.PIPE, universal_newlines=True, check=True)
+                    subprocess_run = subprocess.run([GetPython3Command(), "-u", temp_file_path], stdout=subprocess.PIPE, universal_newlines=True, check=True)
                     output_lines.append("\n")
                     output_lines.append("Expected output:\n")
                     output_lines.append("```bash\n")
--- a/docs/process_pages.py
+++ b/docs/process_pages.py
@@ -273,7 +273,7 @@ def AddPythonSnippetOutputs(file_path: Path) -> None:
             found_python_snippet_block = True
             snippet_lines = []
 
-    with open(str(file_path), "w") as file_output:
+    with open(file_path, "w") as file_output:
         file_output.writelines(output_lines)
 
 if __name__ == "__main__":
--- a/docs/process_pages.py
+++ b/docs/process_pages.py
@@ -3,7 +3,6 @@ from argparse import ArgumentParser
 from collections import OrderedDict
 from pathlib import Path
 import subprocess
-import os
 
 from entry_utilities import default_header_dict
 from entry_utilities import file_navigation_levels
@@ -260,7 +259,7 @@ def AddPythonSnippetOutputs(file_path: Path) -> None:
                     output_lines.append("```\n")
 
                     # remove the temp file
-                    os.remove(temp_file_path)
+                    Path(temp_file_path).unlink()
 
                     if is_existing_output_found:
                         index += temp_index + lines[temp_index:].index("```\n") + 1
--- a/docs/process_pages.py
+++ b/docs/process_pages.py
@@ -248,16 +248,15 @@ def AddPythonSnippetOutputs(file_path: Path) -> None:
                     # existing is not found or force re-write is enabled
 
                     # create a temp file
-                    temp_file_path = str(file_path.absolute()) + ".temp.py"
+                    temp_file_path = f"{file_path.name}.temp.py"
                     with open(temp_file_path, "w") as temp_file_output:
                         temp_file_output.writelines(snippet_lines)
 
-                    popen = subprocess.Popen(["python3", temp_file_path], stdout=subprocess.PIPE, universal_newlines=True)
+                    subprocess_run = subprocess.run([GetPython3Command(), temp_file_path], stdout=subprocess.PIPE, universal_newlines=True)
                     output_lines.append("\n")
                     output_lines.append("Expected output:\n")
                     output_lines.append("```bash\n")
-                    for stdout_line in iter(popen.stdout.readline, ""):
-                        output_lines.append(stdout_line)
+                    output_lines.append(subprocess_run.stdout)
                     output_lines.append("```\n")
 
                     # remove the temp file
@@ -269,7 +268,7 @@ def AddPythonSnippetOutputs(file_path: Path) -> None:
         if found_python_snippet_block:
             snippet_lines.append(line)
 
-        if line.startswith("```python"):
+        if line == "```python\n":
             found_python_snippet_block = True
             snippet_lines = []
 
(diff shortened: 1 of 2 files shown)


docs/process_pages.py
```diff
@@ -215,6 +217,65 @@ def CreateNavigatonBar(root_path: str, max_levels: int, default_header_dict: dic
     list_of_strings = GenerateStrings(list_of_entries, is_locally_built)
     return list_of_strings
 
+def AddPythonSnippetOutputs(file_path: Path) -> None:
+    with open(str(file_path), "r") as file_input:
```
philbucher:
```suggestion
    with open(file_path, "r") as file_input:
```

docs/process_pages.py
```diff
@@ -215,6 +217,65 @@ def CreateNavigatonBar(root_path: str, max_levels: int, default_header_dict: dic
...
+                        index += temp_index + lines[temp_index:].index("```\n") + 1
+
+        if found_python_snippet_block:
+            snippet_lines.append(line)
+
+        if line.startswith("```python"):
+            found_python_snippet_block = True
+            snippet_lines = []
+
+    with open(str(file_path), "w") as file_output:
```
philbucher:
```suggestion
    with open(file_path, "w") as file_output:
```

philbucher (commented):
Some comments :)

docs/process_pages.py
```diff
@@ -215,6 +217,65 @@ def CreateNavigatonBar(root_path: str, max_levels: int, default_header_dict: dic
...
+                        is_existing_output_found  = lines[temp_index] == "Expected output:\n"
+                        is_existing_output_found &= lines[temp_index+1] == "```bash\n"
+                        break
+                    temp_index += 1
+
+                if not is_existing_output_found:
+                    # existing is not found or force re-write is enabled
+
+                    # create a temp file
+                    temp_file_path = str(file_path.absolute()) + ".temp.py"
```
philbucher: You can use with_name

No need to convert to str

docs/process_pages.py
```diff
@@ -215,6 +217,65 @@ def CreateNavigatonBar(root_path: str, max_levels: int, default_header_dict: dic
...
+                    with open(temp_file_path, "w") as temp_file_output:
+                        temp_file_output.writelines(snippet_lines)
+
+                    popen = subprocess.Popen(["python3", temp_file_path], stdout=subprocess.PIPE, universal_newlines=True)
```
philbucher: Please use subprocess.run, it is the modern alternative

philbucher: Also, use GwtPyExe from the kratos testing utils

docs/process_pages.py
```diff
@@ -215,6 +217,65 @@ def CreateNavigatonBar(root_path: str, max_levels: int, default_header_dict: dic
...
+                    output_lines.append("\n")
+                    output_lines.append("Expected output:\n")
+                    output_lines.append("```bash\n")
+                    for stdout_line in iter(popen.stdout.readline, ""):
+                        output_lines.append(stdout_line)
+                    output_lines.append("```\n")
+
+                    # remove the temp file
+                    os.remove(temp_file_path)
```
philbucher: I think there is also a fct for this in pathlib

docs/process_pages.py
philbucher: Or with_suffix

sunethwarna: good idea :)

docs/process_pages.py
sunethwarna: done :)

docs/process_pages.py
sunethwarna: Hmm... I could not find it. It has a method to remove dirs, but coudn't find to remove files ... :/

docs/process_pages.py
sunethwarna: done :)

docs/process_pages.py
philbucher: I use `file.unlink()` for this

docs/process_pages.py
```diff
@@ -215,6 +217,64 @@ def CreateNavigatonBar(root_path: str, max_levels: int, default_header_dict: dic
...
+
+                if not is_existing_output_found:
+                    # existing is not found or force re-write is enabled
+
+                    # create a temp file
+                    temp_file_path = f"{file_path.name}.temp.py"
+                    with open(temp_file_path, "w") as temp_file_output:
+                        temp_file_output.writelines(snippet_lines)
+
+                    subprocess_run = subprocess.run([GetPython3Command(), temp_file_path], stdout=subprocess.PIPE, universal_newlines=True)
```
philbucher:
```suggestion
                    subprocess_run = subprocess.run([GetPython3Command(), temp_file_path], stdout=subprocess.PIPE, universal_newlines=True, check=True)
```

otherwise failures go undetected

docs/process_pages.py
```diff
@@ -215,6 +217,64 @@ def CreateNavigatonBar(root_path: str, max_levels: int, default_header_dict: dic
...
+            output_lines.append(line)
+
+        if found_python_snippet_block and line.startswith("```"):
+            found_python_snippet_block = False
+            if "## POST_PROCESS_PAGES_PYTHON_OUTPUT_GENERATION\n" in snippet_lines:
+                # check whether existing expected output is found
+                temp_index = index
+                is_existing_output_found = False
+                while temp_index < len(lines):
+                    if lines[temp_index].strip() != "":
```
philbucher:
```suggestion
                    if lines[temp_index].strip():
```

technically sufficient

docs/process_pages.py
```diff
@@ -258,6 +318,10 @@ def CreateNavigatonBar(root_path: str, max_levels: int, default_header_dict: dic
                         file_output.write("entries:\n")
                         file_output.writelines(list_of_strings)
 
+    if is_locally_built:
+        from KratosMultiphysics.testing.utilities import GetPython3Command
```
philbucher, line 321:
```suggestion
```

unused

sunethwarna: It is used within the method following this line, I did the import their so, specially in the CI, you don't need to have Kratos environment to depoly the pages.

docs/process_pages.py
sunethwarna: thanks for the hint... and it is done :)

philbucher (dismissed):

sunethwarna:
Fixed the pushing of cpp outputs by adding "-u" arg. @philbucher could you check it ? Thanks :)

philbucher (approved):

PR #12012, opened 2024-02-05
[Core] Adding `VectorIntegerHash` to `key_hash.h`

loumalouomega:
**📝 Description**

This PR introduces a new `VectorIntegerHash` struct, providing a custom hash function for `std::vector<TIntegerType>`.

**🆕 Changelog**

- [Adding `VectorIntegerHash` to `key_hash.h`](https://github.com/KratosMultiphysics/Kratos/commit/421edaa744ff5c3b932059054d29819089ff21c0)

### DIFF
--- a/kratos/includes/key_hash.h
+++ b/kratos/includes/key_hash.h
@@ -90,6 +90,29 @@ namespace Kratos
 ///@{
 
     class VariableData; // forward declaration
+    
+    /**
+     * @brief A custom hash function for std::vector<TIntType>.
+     * @details This struct defines a custom hash function for std::vector<TIntType>. It computes the hash value of a vector by iterating through its elements and combining their hash values.
+     * @tparam TIntType The integer type
+     */
+    template<class TIntegerType = int>
+    struct VectorIntegerHash 
+    {
+        /**
+         * @brief Calculate the hash value for a std::vector<int>.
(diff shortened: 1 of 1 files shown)


kratos/includes/key_hash.h
```diff
@@ -90,6 +90,29 @@ namespace Kratos
...
 
     class VariableData; // forward declaration
+    
+    /**
+     * @brief A custom hash function for std::vector<TIntType>.
+     * @details This struct defines a custom hash function for std::vector<TIntType>. It computes the hash value of a vector by iterating through its elements and combining their hash values.
+     * @tparam TIntType The integer type
+     */
+    template<class TIntegerType = int>
+    struct VectorIntegerHash
```
sunethwarna, line 100: Jus curious, why call this `VectorIntegerHash` when it can be done with any `std::vector<Type>`?

loumalouomega: Nope, because the only type which is a hash itself is an integer, other type should be cast to integer and that may not always work

loumalouomega:
Hello, any reason to block this?

loumalouomega:
> I have no reason to block this but why do you need to use a vector as a hash key?
> 
> It looks really inefficient. Even if its used to index let's say elements using the id of their nodes I would go with tuples

Of course, but when it is the only option...

PR #12015, opened 2024-02-05
[GeoMechanicsApplication] Eliminated duplicated custom element utilities

### LINKED ISSUE #12014: [GeoMechanicsApplication] Eliminate duplicated custom element utilities
Apparently, in the past the custom element utilities of the StructuralMechanicsApplication had been copied to the GeoMechanicsApplication. To avoid duplication, we'd better eliminate the copy, which is nearly identical to the original. The two major differences between the copy and its original are:
1. The StructuralMechanicsApplication uses nonmember functions which are encapsulated in a dedicated namespace, whereas the GeoMechanicsApplication uses a class with static member functions. The function signatures do match, so switching from one implementation to the other should be straightforward.
2. When calculating the Rayleigh damping matrix, the StructuralMechanicsApplication takes into account a tiny tolerance for the 'alpha' and 'beta' coefficients, whereas the GeoMechanicsApplication accepts any positive value for them.

avdg81:
**📝 Description**
Apparently, in the past the custom element utilities of the StructuralMechanicsApplication had been copied to the GeoMechanicsApplication. To eliminate this duplication, we have removed the (near) copy and replaced its usages by calling the corresponding functions from the StructuralMechanicsApplication. To make this work, we had to export two utility functions.

Note that there were some minor differences which we deemed insignificant:
- The StructuralMechanicsApplication uses nonmember functions which are encapsulated in a dedicated namespace, whereas the GeoMechanicsApplication used a class with static member functions.
- When calculating the Rayleigh damping matrix, the StructuralMechanicsApplication takes into account a tiny tolerance for the 'alpha' and 'beta' coefficients, whereas the GeoMechanicsApplication accepted any positive value for them.


applications/StructuralMechanicsApplication/custom_utilities/structural_mechanics_element_utilities.h
```diff
@@ -195,7 +195,7 @@ double GetRayleighBeta(
  * @param rElement The Element for which the mass-matrix should be computed
  * @return The density after apply the mass factor to the element
  */
-double GetDensityForMassMatrixComputation(const Element& rElement);
+double KRATOS_API(STRUCTURAL_MECHANICS_APPLICATION) GetDensityForMassMatrixComputation(const Element& rElement);
```
philbucher, line 198: nice catch

philbucher (approved):

WPK4FEM (approved):
Looks consistent to me. Away with the copies.

rfaasse (approved):
Very nice removal of duplication!

